PR #2084, opened 2020-07-22
Add remote coinbase transaction handling to the wallet

philipr-za:
## Description

This PR adds the functionality for an external application to request a complete Coinbase transaction from the wallet by providing the reward, fees and block height at which this coinbase will appear. The Transaction Service will then request a coinbase key from the Output Manger Service which will generate a private key from the “coinbase” branch of the current Key Manager using the Block Height as an index. The transaction service will then produce and return the Coinbase transaction to the requester and start monitoring the chain for the Coinbase output to be mined. If the chain advances to a height beyond the Coinbase transactions block height and it has not appeared in the chain it is cancelled.

If multiple requests are made for a coinbase transaction of the same height the previous transaction and outputs are cancelled and the new output and transaction becomes the active Coinbase transaction for that block height.

## How Has This Been Tested?
Tests are provided that test the Output Manager and Transaction manager logic in the database backend and from the Service API perspectives.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/mining/mod.rs
+++ b/base_layer/core/src/mining/mod.rs
@@ -21,9 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 mod blake_miner;
-mod coinbase_builder;
 mod error;
 mod miner;
 
-pub use coinbase_builder::CoinbaseBuilder;
 pub use miner::Miner;
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -81,6 +81,7 @@ where TBackend: OutputManagerBackend + Clone + 'static
 {
     resources: OutputManagerResources<TBackend>,
     key_manager: Mutex<KeyManager<PrivateKey, KeyDigest>>,
+    coinbase_key_manager: Mutex<KeyManager<PrivateKey, KeyDigest>>,
     request_stream:
         Option<reply_channel::Receiver<OutputManagerRequest, Result<OutputManagerResponse, OutputManagerError>>>,
     base_node_response_stream: Option<BNResponseStream>,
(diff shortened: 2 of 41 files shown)


base_layer/core/src/mining/mod.rs
```diff
@@ -20,10 +20,18 @@
...
 mod coinbase_builder;
+#[cfg(feature = "base_node")]
 mod error;
+#[cfg(feature = "base_node")]
 mod miner;
 
+#[cfg(any(feature = "base_node", feature = "transactions"))]
+pub use coinbase_builder::CoinbaseBuildError;
+#[cfg(any(feature = "base_node", feature = "transactions"))]
 pub use coinbase_builder::CoinbaseBuilder;
```
stringhandler: <Suggestion> Maybe this struct should move to the `transactions` mod?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -436,6 +449,45 @@ where
...
+    {
+        let mut key = PrivateKey::default();
+        {
+            let km = self.coinbase_key_manager.lock().await;
+            key = km.derive_key(block_height)?.k;
+        }
+
+        self.resources
+            .db
+            .cancel_pending_transaction_below_block_height(block_height)
```
stringhandler: Are all transactions cancelled or only coinbase transactions?

base_layer/wallet/src/output_manager_service/storage/database.rs
```diff
@@ -519,6 +525,18 @@ where T: OutputManagerBackend + Clone + 'static
             .and_then(|inner_result| inner_result)
     }
 
+    pub async fn cancel_pending_transaction_below_block_height(
```
stringhandler: This method name mismatches the one actually called, where it only cancels __at__ the block height. Is there something going to be added in future?

stringhandler (commented):
Looking good. Just some questions

base_layer/core/src/mining/mod.rs
philipr-za: Hmm, ja I think it could. I will see if I can do it neatly.

base_layer/wallet/src/output_manager_service/service.rs
philipr-za: Any transactions that have a blockheight specified which will only be coinbase transactions. Perhaps that column should be explicitly `coinbase_block_height` so as not to confuse with other types of maturity that might come later.

base_layer/wallet/src/output_manager_service/storage/database.rs
philipr-za: Oh I actually moved from this to just `at_block_height` because I realized it should only relate to the current block_height and let the monitoring protocols handle the ones at lower blockheight. Will update this.

SWvheerden (dismissed):

stringhandler (approved):
Did a code review, but didn't test it locally. Looks fine. Any problems should come out when we start calling this from the miner code. The missing transactions are logged in another issue, as mentioned

PR #2086, opened 2020-07-24
Implement optimised outbound messaging

sdbondi:
## Description

This PR optimises the outbound messaging by flushing less regularly as
suggested by the `Sink` trait documentation.

The previous implementation would receive a message, write it and flush.
This is sub-optimal because a buffer may not be full and so many smaller
"chunks" are being sent.

This implementation reads a message, writes it and continues to read
until there is a "pause" (`Poll::Pending`) at which point is flushes any
remaining data still to be sent. Ensuring that the buffer is utilised.

This implementation is made much more complex (and slightly slower) by the current
requirement to report on if each message was put in a buffer (considered sent)
which allows the message send to be reattempted.

## Motivation and Context

Performance improvements for outbound messaging.

## How Has This Been Tested?

Public behaviour is extensively tested in existing tests/examples/applications. All of these function as expected.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3657,7 +3657,6 @@ dependencies = [
  "serde_json",
  "tari_crypto",
  "tari_infra_derive",
- "tari_storage",
  "tari_utilities",
 ]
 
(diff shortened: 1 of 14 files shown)


SWvheerden (approved):

SWvheerden (approved):

stringhandler (approved):
Happy with the concept and approach. Looked over the code and it seems fine. Didn't test it locally though

PR #2090, opened 2020-07-24
Implement interface for `tari_console_wallet` application

philipr-za:
## Description
This PR implements the layout and UX of the `tari_console_wallet`. To run the wallet in the console type `cargo run --bin tari_console_wallet` in the root of the repo.

The data in this PR is currently all dummy data to demonstrate the layout and UX of the app. The next PR will be to plumb a full LibWallet backend into the app.

Missing UI features to be added in later PRs are:
- Fully console to handle commands like those in the `tari_base_node` app

I included adding a new feature to the Output Manager Service to include a time-locked balance if the chain tip is provided.

## How Has This Been Tested?
Unit tests provided for components

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,11 +8,11 @@ checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"
 
 [[package]]
 name = "aead"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3672ba0d12aaf256aabcdab516ebed87b5cec4b602316d7a3035fac9b7a9567b"
+checksum = "7fc95d1bdb8e6666b2b217308eeeb09f2d6728d104be3e31916cc74d15420331"
 dependencies = [
- "generic-array 0.14.2",
+ "generic-array 0.14.3",
 ]
 
 [[package]]
@@ -86,9 +86,9 @@ checksum = "34fde25430d87a9388dadbe6e34d7f72a462c8b43ac8d309b42b0a8505d7e2a5"
 
 [[package]]
 name = "anyhow"
-version = "1.0.31"
+version = "1.0.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85bb70cc08ec97ca5450e6eba421deeea5f172c0fc61f78b5357b2a8e8be195f"
+checksum = "6b602bfe940d21c130f3895acd65221e8a61270debe89d628b9cb4e3ccb8569b"
 
 [[package]]
 name = "arc-swap"
@@ -145,9 +145,9 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25f9db3b38af870bf7e5cc649167533b493928e50744e2c30ae350230b414670"
 dependencies = [
- "proc-macro2 1.0.18",
+ "proc-macro2 1.0.19",
  "quote 1.0.7",
- "syn 1.0.34",
+ "syn 1.0.36",
 ]
 
 [[package]]
@@ -156,9 +156,9 @@ version = "0.1.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a265e3abeffdce30b2e26b7a11b222fe37c6067404001b434101457d0385eb92"
 dependencies = [
- "proc-macro2 1.0.18",
+ "proc-macro2 1.0.19",
  "quote 1.0.7",
- "syn 1.0.34",
+ "syn 1.0.36",
 ]
 
 [[package]]
@@ -239,7 +239,7 @@ dependencies = [
  "lazycell",
  "log",
  "peeking_take_while",
- "proc-macro2 1.0.18",
+ "proc-macro2 1.0.19",
  "quote 1.0.7",
  "regex",
  "rustc-hash",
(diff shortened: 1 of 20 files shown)


CjS77:
Look great! Some thoughts:

Transactions:
1. Balance breakdown should also have an entry for time-locked / maturing balance
2. The transaction detail should auto-update while moving through the tx list

Send/Receive
1. Hotkeys to copy my pubkey to the clipboard maybe?

Other:
1. Sync status / Get chain meta data at top of every screen.
2. Is it possible to have a console page as well with the current prompt / advanced cmd set?

Other features like UTXO set etc can come later

philipr-za:
@CjS77 Thanks for the feedback will definitely work on all these items, just some comments on some of the suggestions.

1. On the sync status, I am guessing you mean whether the base node is synced? There is currently no way to request this via the comms stack so I will show chain meta data but will add the proper sync status later by either adding an extra field to the Get Chain Meta data response or by adding the gRPC interface to this wallet (or both).
2. I was going to add a console using the parsing logic from the Base Node though initially it will only be able to support the commands directly related to the wallet and anything that can be done over the comms layer to the base node.

hansieodendaal:
Hi there, I think it is important that OSX, Linux and Windows be supported from the get-go.

philipr-za:
> Hi there, I think it is important that OSX, Linux and Windows be supported from the get-go.

The tui-rs library support different backends but the one for windows is apparently not as performant as termion. Adding in another backend won't be difficult but I feel can be its own PR to avoid this one getting too big.

hansieodendaal:
> The tui-rs library support different backends but the one for windows is apparently not as performant as termion. Adding in another backend won't be difficult but I feel can be its own PR to avoid this one getting too big.

Agreed, as long as multi-platform support is not handled as an afterthought, with the same functionality and feature set accross.

stringhandler:
Tested it at commit 98e432a and it really looks great. Just need a way to quit (Maybe Q?)
I haven't looked through the code yet, but will dig into it when it's out of WIP

SWvheerden:
Is it possible to show/create a help/index page explaining what each term means?
For example I am making an educated guess by saying red = outbound, green = inbound. And then also what the status is.

hansieodendaal:
Hi, got it working on an Ubuntu virtual. Same comment as above, _**looks great**_. 

Some specifics below and I assume this console wallet will replace the base node wallet we currently have:
- For `Completed Transactions`, could we perhaps have a maturity column alongside `Status`?
- Colour in the IntelliJ IDEA terminal has some problem. The transaction details are not legible; it appears to be a very dark colour on black, looks like black on black. Much better in a normal terminal, but then the pink is not easy to read. Maybe some standardized theme across terminals?
- In the current base node wallet we do not have borders. This helps a lot when using `list-transactions`, which outputs texts strings much wider than the borders.
- I also miss the command set and a terminal to input commands. Is the plan not to support all commands we have currently, or could we have the ability to spawn a terminal for power users?

philipr-za:
> Is it possible to show/create a help/index page explaining what each term means?
> For example I am making an educated guess by saying red = outbound, green = inbound. And then also what the status is.

Probably will do this, or else the design might include more tooltips. Going to leave this till the Design team has had a say.

philipr-za:
Thanks for the feedback.

> Hi, got it working on an Ubuntu virtual. Same comment as above, _**looks great**_.
> 
> Some specifics below and I assume this console wallet will replace the base node wallet we currently have:
> 
> * For `Completed Transactions`, could we perhaps have a maturity column alongside `Status`?

We are very limited in horizontal real-estate but I will see if I can fit it in.

> * Colour in the IntelliJ IDEA terminal has some problem. The transaction details are not legible; it appears to be a very dark colour on black, looks like black on black. Much better in a normal terminal, but then the pink is not easy to read. Maybe some standardized theme across terminals?

So I am using a Color abstraction that should be platform independent and I am only using magenta or white, so will need to check out the color issues but I am not super worried about making it work inside of IDEA.

> * In the current base node wallet we do not have borders. This helps a lot when using `list-transactions`, which outputs texts strings much wider than the borders.

This is a windowed gui so we are not just printing boundless strings like in the CLI app. I might leave out the borders in the Console tab if it makes sense.

> * I also miss the command set and a terminal to input commands. Is the plan not to support all commands we have currently, or could we have the ability to spawn a terminal for power users?

I am going to add in a console to the app that can support the parser is in the Base node app but bear in mind this is just a wallet so only a small subset of those commands will be relevant.

hansieodendaal:
> I am going to add in a console to the app that can support the parser is in the Base node app but bear in mind this is just a wallet so only a small subset of those commands will be relevant.

This is my take on it, what needs to be carried over. From the current list:
- `list-utxos`
- `list-transactions`
- `list-completed-transactions`
- `cancel-transaction`
- `send-tari`
- `list-peers`
- `list-connections`
- `discover-peer`
- `make-it-rain`
- `coin-split`

I also think we can do with this new functionality:
- `cancel-all-transactions`
- `resend-all-pending-transactions`
- `validate-all-txos`

philipr-za:
So I think that I am going to put off the console functionality until after I have integrated LibWallet to support the functionality in the current UI due to some time pressure and it being an easy feature to bolt on after this PR.

hansieodendaal:
> So I think that I am going to put off the console functionality until after I have integrated LibWallet to support the functionality in the current UI due to some time pressure and it being an easy feature to bolt on after this PR.

That makes sense, yes. We can run the base node wallet and this wallet concurrently until this wallet has matured sufficiently to replace all power-user features.

stringhandler (approved):
Tested locally. Looks great!

PR #2091, opened 2020-07-25
Added burst send mode to comms stress test

hansieodendaal:
## Description
Added the ability to send multiple bursts of messages in batches of size `1` to `n`.

## Motivation and Context
Adding more options to do comms stress testing.

## How Has This Been Tested?
Tested with 2x comms stress test executables over Tor.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/examples/stress/service.rs
+++ b/comms/examples/stress/service.rs
@@ -82,6 +82,7 @@ pub enum StressTestServiceRequest {
 pub struct StressProtocol {
     pub kind: StressProtocolKind,
     pub num_messages: u32,
+    pub burst_size: u32,
     pub message_size: u32,
 }
 
@@ -89,13 +90,17 @@ pub struct StressProtocol {
 pub enum StressProtocolKind {
     ContinuousSend,
     AlternatingSend,
+    BurstSend,
 }
 
 impl StressProtocol {
-    pub fn new(kind: StressProtocolKind, num_messages: u32, message_size: u32) -> Self {
+    const PROTOCOL_BYTES_SIZE: usize = 13;
+
+    pub fn new(kind: StressProtocolKind, num_messages: u32, burst_size: u32, message_size: u32) -> Self {
         Self {
             kind,
             num_messages,
+            burst_size,
             message_size,
         }
     }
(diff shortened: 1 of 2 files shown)


comms/examples/stress/service.rs
```diff
@@ -389,6 +435,40 @@ async fn start_responder_protocol(peer: NodeId, mut substream: Substream) -> Res
...
         },
+        StressProtocolKind::BurstSend => {
+            let mut received = vec![];
+            let start = Instant::now();
+
+            for _i in 0..(protocol.num_messages / protocol.burst_size) {
+                while let Some(Ok(msg)) = framed.next().await {
+                    received.push(decode_msg(msg));
+                }
+            }
```
sdbondi: `framed.next().await` will not return None after a burst. It will just `await` the next message. `next()` returns None once the stream has ended (e.g. disconnected substream). So your while loop will read all the messages, and subsequent loops will exit immediately (including the while loop below). 

  ```suggestion
                while let Some(Ok(msg)) = framed.next().await {
                    received.push(decode_msg(msg));
                }
            
```

File: comms/examples/stress/service.rs [hunk 2]
```diff
@@ -389,6 +435,40 @@ async fn start_responder_protocol(peer: NodeId, mut substream: Substream) -> Res
...
+            if protocol.num_messages % protocol.burst_size > 0 {
+                while let Some(Ok(msg)) = framed.next().await {
+                    received.push(decode_msg(msg));
+                }
+            }
```
sdbondi: Not needed

comms/examples/stress/service.rs
```diff
@@ -290,13 +301,48 @@ async fn start_initiator_protocol(mut conn: PeerConnection, protocol: StressProt
...
+
+            println!(
+                "Sending {} messages in bursts of {} and finally {} messages ({} MiB)",
+                protocol.num_messages,
+                protocol.burst_size,
+                protocol.num_messages % protocol.burst_size,
+                protocol.num_messages * protocol.message_size / 1024 / 1024,
+            );
+            for _i in 0..(protocol.num_messages / protocol.burst_size) {
+                let mut iter = stream::iter(
```
sdbondi, line 327: Given that there are no pauses between "bursts", is this not almost exactly the same as continuous send? A test I would like to see added is concurrent reading and writing.

sdbondi (commented):
Looks good - I feel a better stress test to add may be a concurrent reading/writing test.

comms/examples/stress/service.rs
hansieodendaal: Thanks, did not realize.

comms/examples/stress/service.rs
hansieodendaal: > A test I would like to see added is concurrent reading and writing.

This is easily simulated by starting a longish test from both ends within a short time span; no additional code changes needed.

hansieodendaal: > Given that there are no pauses between "bursts", is this not almost exactly the same as continuous send?

I guess the only difference will be finalization given by `send_all` after each burst, i.e. `A future that completes after the given stream has been fully processed into the sink, including flushing.`

sdbondi:
I don't think this adds any value over the existing continuous send test.

PR #2094, opened 2020-07-29
Integrate LibWallet into the `tari_console_wallet` application

philipr-za:
## Description
This PR integrates the tari_wallet backend into the tari_console_wallet application. The application needed to reused a lot of the utility functions from `tari_base_node` so it moves those utility functions into a new crate called `tari_app_utilities`. 

The wallet will use the same config location as `tari_base_node` and the same identity and data files. It will create a separate log config and log to a separate folder within ~/.tari.

This PR is still in progress and the following is what I think still needs to be done to flesh out all the functionality present in the current UI:

- Add way to specific wallet base node in config (Currently just using the first Seed Node)
- Add way to edit base node pub key and address in console UI
- Read Balance from wallet
- Read Contacts from wallet
- Update above on relevant wallet events by subscribing to the wallet Event Streams.
- Read Chain tip from Base Node in wallet (this is a whole thing) and plumb into UI with updates via wallet events
- Add calling Send transaction in wallet
- Add contacts from UI
- Read current connections from Wallet for Network Tab
- Optimize the cacheing and display of the Complete Transaction list which lags when it is very long. Currently it caches the full completed transactions but only needs a few fields which should help a lot.
- Determine if running this app with the old BN will create issues with a single database file in terms of crashes and corruption (Perhaps the WAL journal mode is good enough to allow this operation in Sqlite?)

After this we still need to integrate gRPC and add a Console page.

## How Has This Been Tested?
Tests provided for features added to support application.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


applications/tari_app_grpc/src/conversions/aggregate_body.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::generated::tari_rpc as grpc;
+use tari_core::transactions::transaction::Transaction;
+
+impl From<AggregateBody> for grpc::AggregateBody {
+    fn from(source: AggregateBody) -> Self {
+        Self {
+            inputs: source
+                .inputs()
+                .iter()
+                .map(|input| grpc::TransactionInput {
```
sdbondi: Should rather implement `impl Into<TransactionInput> for grpc::TransactionInput` and use `.map(Into::into)` because this snippet of code occurs in a few places.

applications/tari_app_grpc/src/conversions/new_block_template.rs
```diff
@@ -0,0 +1,125 @@
...
+                target_difficulty: block.header.pow.target_difficulty.as_u64(),
+            }),
+        };
+        Self {
+            body: Some(grpc::AggregateBody {
+                inputs: block
+                    .body
+                    .inputs()
+                    .iter()
+                    .map(|input| grpc::TransactionInput {
```
sdbondi: Reuse `impl Into<grpc::TransactionInput>`, same with outputs and kernels

applications/tari_app_grpc/src/conversions/aggregate_body.rs
sdbondi: Same with `grpc::TransactionOutput` and `grpc::TransactionKernel`

applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -0,0 +1,189 @@
...
+/// ## Parameters
+/// `identity_file` - Reference to file path
+/// `public_address` - Network address of the base node
+/// `create_id` - Whether an identity needs to be created or not
+/// `peer_features` - Enables features of the base node
+///
+/// # Return
+/// A NodeIdentity wrapped in an atomic reference counter on success, the exit code indicating the reason on failure
+pub fn setup_node_identity(
+    identity_file: &PathBuf,
```
sdbondi: nit: While we're moving utility functions around I'd recommend making them generic on `AsRef<Path>` so that you can write e.g. `setup_node_identity("/tmp/file.tmp", ...)`. Same for `load_identity` - just for the sake of ease of use and consistency

```suggestion
pub fn setup_node_identity<P: AsRef<Path>>(
    identity_file: P,
```

applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// Tries to construct a node identity by loading the secret key and other metadata from disk and calculating the
+/// missing fields from that information.
+/// ## Parameters
+/// `path` - Reference to a path
+///
+/// ## Returns
+/// Result containing a NodeIdentity on success, string indicates the reason on failure
+pub fn load_identity(path: &Path) -> Result<NodeIdentity, String> {
+    if !path.exists() {
```
sdbondi: nit:
```suggestion
pub fn load_identity<P: AsRef<Path>>(path: P) -> Result<NodeIdentity, String> {
    if !path.as_ref().exists() {
```

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -0,0 +1,66 @@
...
+                    TransactionServiceSqliteDatabase,
+                    OutputManagerSqliteDatabase,
+                    ContactsServiceSqliteDatabase,
+                >,
+            >,
+        >,
+    ) -> Self
+    {
+        Self { wallet }
+    }
```
sdbondi, line 41: You'll be able to avoid the lock entirely by cloning the `transaction_service` each time you need to use it. Essentially you're cloning an `mpsc::UnboundedSender` which is very cheap. 

```suggestion
    pub fn new(
        wallet: Arc<
                Wallet<
                    WalletSqliteDatabase,
                    TransactionServiceSqliteDatabase,
                    OutputManagerSqliteDatabase,
                    ContactsServiceSqliteDatabase,
                >,
        >,
    ) -> Self
    {
        Self { wallet }
    }
```

Depending on how `TransactionService` is implemented, you should also be able to get some concurrency gains.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -0,0 +1,66 @@
...
+    ) -> Result<Response<GetCoinbaseResponse>, Status>
+    {
+        let request = request.into_inner();
+        let response = self
+            .wallet
+            .write()
+            .await
+            .transaction_service
+            .generate_coinbase_transaction(request.reward.into(), request.fee.into(), request.height)
+            .await;
```
sdbondi, line 57:
```suggestion
       let mut transaction_sevice = self
            .wallet
            .transaction_service.clone();
         let response = transaction_service
            .generate_coinbase_transaction(request.reward.into(), request.fee.into(), request.height)
            .await;
```

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -0,0 +1,263 @@
...
+        pending_transactions.extend(
+            self.wallet
+                .write()
+                .await
+                .transaction_service
+                .get_pending_inbound_transactions()
+                .await?
+                .values()
+                .map(|t| CompletedTransaction::from(t.clone()))
+                .collect::<Vec<CompletedTransaction>>(),
```
sdbondi, line 186: Same here - rather than using a RwLock - it's cheaper and more concurrent to clone the handle and use it mutably

applications/tari_console_wallet/src/ui/stateful_list.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn select_first(&mut self) {
+        self.state.select(Some(0));
+    }
+
+    pub fn selected_item(&self) -> Option<T> {
+        match self.state.selected() {
+            None => None,
+            Some(i) => Some(self.items[i].clone()),
+        }
+    }
```
sdbondi: nit: It feels more correct to return a reference from this function i.e. `Option<&T>` and leave the cloning to the consumer if necessary - looking at usages of this function, I see `selected_item().as_ref()` which is wasteful. Granted most other places seem to need a `selected_item().cloned()`, but I prefer the explicitness anyway.

sdbondi:
```rust
self.state.selected().and_then(|i| self.items.get(i))
```

applications/tari_console_wallet/src/ui/ui_contact.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub alias: String,
+    pub public_key: String,
+    pub emoji_id: String,
+}
+
+impl From<Contact> for UiContact {
+    fn from(c: Contact) -> Self {
+        Self {
+            alias: c.alias,
+            public_key: format!("{}", c.public_key),
```
sdbondi: nit: Clippy would probably complain - `c.public_key.to_string()` 🤷‍♂️

sdbondi (commented):
Looks awesome! I'm pretty sure that you can do away with the `RwLock` if you simply clone the handle and use it mutably as needed.

I'm going to run this branch and do some UI testing.

applications/tari_app_grpc/src/conversions/aggregate_body.rs
sdbondi: You may run into issues with this (conflicting implementations) - in which case, I say leave it as is

sdbondi:
Looks like we need to get rid of some warnings before this can be considered to go in. Some of the warnings are more serious e.g. unused results and futures but generally all warnings should be fixed before being merged in.

applications/tari_app_grpc/src/conversions/aggregate_body.rs
stringhandler: This code was actually just moved from another file, not changed in this implementation. So maybe something to look at at a later stage

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
stringhandler: Perhaps worth looking into this at a later stage. I don't disagree that it may be more optimal, but there is still quite a lot to implement, and at this stage it's a fairly heavy lock. Conceptually I was thinking: The GRPC server and TUI will be using the wallet. Some parts of the GRPC are only going to need read access to it, some need write. In the TUI, the display cycle only reads from the wallet, and when you press a button, or invoke an action you need write access to it. You may be able to go into finer grain locking, or use a non-locking strategy, but as I said, not all of the functions are implemented yet, so you may find that you constantly changing these locks.

philipr-za: As discussed this architectural comment will be addressed in subsequent PRs

philipr-za:
Ok this is ready for a review. Cleaned up all the warnings and addressed most the comment. The comments that were not directly addressed will be dealt with in subsequent PRs

stringhandler (approved):

sdbondi (approved):

PR #2095, opened 2020-07-30
Change SQLite RENAME to DROP and CREATE

stringhandler:
Older SQLite version do not support RENAME for columns
so use DROP and CREATE to change the name of the table
instead


sdbondi (approved):
LGTM

hansieodendaal:
Works on Ububtu

SWvheerden (approved):

PR #2096, opened 2020-07-30
Fix mining ON/OFF toggle delayed response

hansieodendaal:
## Description
- Fixed mining ON/OFF delayed response when `toggle-mining` was requested from the CLI, by introducing an event channel that directly interrupts the blocking mining loop.
- Provided better user feedback by displaying the global mining enabled state as well as the actual mining state.

```
>> get-mining-state
Mining is DISABLED by the user
Mining state is currently OFF
Mining hash rate is: 0.000000 MH/s
>>
>> toggle-mining
Mining requested to be turned ON
>> Mining is ON
>>
>> get-mining-state
Mining is ENABLED by the user
Mining state is currently ON
Mining hash rate is: 0.000000 MH/s
>>
>> toggle-mining
Mining requested to be turned OFF
>> Mining is OFF
>>
>> get-mining-state
Mining is DISABLED by the user
Mining state is currently OFF
Mining hash rate is: 0.000000 MH/s
>>
```

## Motivation and Context
When mining was disabled from the CLI whilst in progress, changing the global mining enabled state, it only stopped after the miner received an interrupt based on mempool events and blockchain status. Consequently, requesting mining status from the CLI also reported the global mining enabled state instead of the actual mining state.

## How Has This Been Tested?
Tested with the base node runtime on Windows 10 and Ubuntu Linux

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/mining/miner.rs
+++ b/base_layer/core/src/mining/miner.rs
@@ -28,7 +28,7 @@ use crate::{
     blocks::{Block, BlockHeader, NewBlockTemplate},
     consensus::ConsensusManager,
     mempool::MempoolStateEvent,
-    mining::{blake_miner::CpuBlakePow, error::MinerError},
+    mining::{blake_miner::CpuBlakePow, error::MinerError, MinerInstruction},
     proof_of_work::PowAlgorithm,
     transactions::{
         transaction::UnblindedOutput,
@@ -40,7 +40,7 @@ use core::sync::atomic::{AtomicBool, AtomicU64};
 use futures::{
     channel::{
         mpsc,
-        mpsc::{Receiver, Sender},
+        mpsc::{Receiver as mpscReceiver, Sender as mpscSender},
     },
     pin_mut,
     StreamExt,
@@ -51,7 +51,14 @@ use std::sync::{atomic::Ordering, Arc};
 use tari_broadcast_channel::Subscriber;
 use tari_crypto::keys::SecretKey;
 use tari_shutdown::ShutdownSignal;
-use tokio::{task, task::spawn_blocking};
+use tokio::{
+    sync::{
+        broadcast,
+        broadcast::{Receiver as syncReceiver, Sender as syncSender},
+    },
+    task,
+    task::spawn_blocking,
+};
 
 pub const LOG_TARGET: &str = "c::m::miner";
 
@@ -60,11 +67,13 @@ pub struct Miner {
     stop_mining_flag: Arc<AtomicBool>,
     consensus: ConsensusManager,
     node_interface: LocalNodeCommsInterface,
-    utxo_sender: Sender<UnblindedOutput>,
+    utxo_sender: mpscSender<UnblindedOutput>,
     node_state_event_rx: Option<Subscriber<StateEvent>>,
     mempool_state_event_rx: Option<Subscriber<MempoolStateEvent>>,
+    miner_instruction_events: syncSender<MinerInstruction>,
     threads: usize,
-    enabled: Arc<AtomicBool>,
+    mining_enabled_by_user: Arc<AtomicBool>,
+    mining_status: Arc<AtomicBool>,
     hashrate: Arc<AtomicU64>,
 }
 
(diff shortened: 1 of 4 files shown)


hansieodendaal:
@SWvheerden please have a look? Thanks

base_layer/core/src/mining/miner.rs
```diff
@@ -77,7 +86,9 @@ impl Miner {
         threads: usize,
     ) -> Miner
     {
-        let (utxo_sender, _): (Sender<UnblindedOutput>, Receiver<UnblindedOutput>) = mpsc::channel(1);
+        let (utxo_sender, _): (mpscSender<UnblindedOutput>, mpscReceiver<UnblindedOutput>) = mpsc::channel(1);
+        let (miner_instruction_events, _): (syncSender<MinerInstruction>, syncReceiver<MinerInstruction>) =
+            broadcast::channel(10);
```
SWvheerden, line 91: Is there a reason we want to que more than 1 instruction. Dont we just want the last instruction?

base_layer/core/src/mining/miner.rs
```diff
@@ -281,6 +318,25 @@ impl Miner {
...
                 },
+                instruction = mining_instruction_event.select_next_some() => {
+                    use MinerInstruction::*;
+                    match instruction.unwrap_or_else(|_| (IgnoreInstruction)) {
+                        PauseMining => {
+                            info!(target: LOG_TARGET, "Mining pause event received from the CLI");
+                            stop_mining_flag.store(true, Ordering::Relaxed);
+                            mining_enabled_by_user.store(false, Ordering::Relaxed);
+                            start_mining = false;
+                            wait_for_miner = true;
```
SWvheerden: If this is true, an event van turn the miner on again. 
Imagine this:
1: user sets pause.
2: miner stops
3: node goes to lagging state. 
4: node goes to BlocksSynchronized state
5: node starts mining again. 

But just setting this to false means the miner wont be able to listen again here.

SWvheerden (changes requested):
I dont believe this will work as intended. See my comment

base_layer/core/src/mining/miner.rs
hansieodendaal: Strictly speaking not necessary, but this is just to make sure we do not drop any commands from the CLI, a bit of resilience

base_layer/core/src/mining/miner.rs
SWvheerden: Nevermind here, missed the ! on line 291

base_layer/core/src/mining/miner.rs
```diff
@@ -86,21 +97,35 @@ impl Miner {
...
 
+    /// This function returns the sender portion of the mining instruction event channel so that start and
+    /// shutdown events can be sent to the miner while mining
+    pub fn get_miner_instruction_events_sender_channel(&self) -> syncSender<MinerInstruction> {
+        self.miner_instruction_events.clone()
+    }
+
+    /// This function returns the receiver portion of the mining instruction event channel so that start and
+    /// shutdown events can be received to interact with the miner while mining
+    pub fn subsrcibe_to_instruction_events(&self) -> syncReceiver<MinerInstruction> {
```
hansieodendaal: @SWvheerden remarked that this can be called directly in `pub async fn mine(mut self)`. Agreed.

base_layer/core/src/mining/miner.rs
hansieodendaal: I understand the confusion, as it is really difficult to get a clear meaning here.
- Propose to change `start_mining` to `spawn_mining_task`
- Propose to change `let mut wait_for_miner = false;` and `while !wait_for_miner {` to positive logic
- Propose to change `wait_for_miner ` to `wait_for_mining_event`
`

SWvheerden (dismissed):

hansieodendaal:
@mikethetike when to merge this one?

stringhandler (approved):
Seems a bit complicated, but tested and works as expected

SWvheerden (approved):
LGTM

PR #2097, opened 2020-07-30
Edge case/race condition fix for incorrect connectivity connection state 

sdbondi:
## Description

Recognise if a dial is cancelled due to an inbound connection beating the dial.
In this case, the `PeerConnectionFailed(DialCancelled)` event should not be published.
This fixes an issue which causes connectivity manager to mistakenly
assign a `Failed` state to a connection in some cases.

## Motivation and Context

If the connectivity connection state says that the connection has failed, it prevents that peer from receiving propagation messages destined for it. 

## How Has This Been Tested?

Added a dial cancellation test and tested on base nodes (running two base nodes and attempting wallet discovery can reproduce this condition) 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/src/connection_manager/manager.rs
+++ b/comms/src/connection_manager/manager.rs
@@ -318,7 +318,7 @@ where
         match request {
             DialPeer(node_id, reply_tx) => match self.get_active_connection(&node_id) {
                 Some(conn) => {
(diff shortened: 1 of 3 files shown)


SWvheerden (approved):

stringhandler (approved):
Untested, but looks good

PR #2098, opened 2020-07-31
Merged Mining - Tari Conduit Application

SWvheerden:
## Description

Added application which XMRig can be connected to which forwards requests back and forth between XMRig and MoneroD.
Added parsing of Monero requests from XMRig and adding of merge mining data to response from MoneroD.
Added integration to Tari Base Node for merged mining.
Added ReadMe.
Added GRPC refector

Code by @StriderDM  and @SWvheerden

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -0,0 +1,49 @@
...
+
+import "types.proto";
+import "google/protobuf/wrappers.proto";
+import "google/protobuf/timestamp.proto";
+
+
+// The gRPC interface for interacting with the wallet.
+service Wallet {
+    // This returns a coinbase transction
+    rpc GetCoinbase(CoinbaseRequest) returns (stream CoinbaseSupply);
```
stringhandler: The choice to use a stream here seems weird. Won't it be a single result?

stringhandler (dismissed):
I ran this locally, but it's difficult to test it fully. I think there may be some breaking changes on the GRPC interface for the explorer etc, but I think we can merge it and then fix those. Will test it for backwards compatibility, but otherwise I'm happy to merge it

applications/tari_app_grpc/proto/wallet.proto
SWvheerden: haha good catch. It should not be a stream. This is what you get if you write the code without implementing it yet.

stringhandler:
Tested this locally, and synced fine with the current chain

PR #2103, opened 2020-08-04
Update `make-it-rain` CLI command

hansieodendaal:
## Description
- Added (1) blocking peer discovery to the make-it-rain CLI command to ensure a valid connection exists before starting to send Txs, (2) test start time, (3) transaction rate. 
- ~~Peer disconnect timeout changed for the messaging protocol.~~

## Motivation and Context
- More functionality was needed for Tx stress testing
- Forced peer discovery before the test starts provides more focussed results
- ~~The messaging protocol disconnected the receiving peer if it was inactive for 5 min, while the test was still in progress.~~

## How Has This Been Tested?
Tested between two base nodes in Windows

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/parser.rs
+++ b/applications/tari_base_node/src/parser.rs
@@ -150,9 +150,7 @@ include!(concat!(env!("OUT_DIR"), "/consts.rs"));
 
 const MAKE_IT_RAIN_USAGE: &str = "\nmake-it-rain [Txs/s] [duration (s)] [start amount (uT)] [increment (uT)/Tx] \
                                   [\"start time (UTC)\" / 'now' for immediate start] [public key or emoji id to send \
-                                  to] [message]\n       or\nmake-it-rain [Txs/s] [duration (s)] [start amount (uT)] \
-                                  [increment (uT)/Tx] [\"start time (UTC)\" / 'now' for immediate start] --file \
-                                  [\"path to file\" containing list of 'public key or emoji id' 'message']\n";
+                                  to] [message]\n";
 
 /// This will go through all instructions and look for potential matches
 impl Completer for Parser {
@@ -755,14 +753,14 @@ impl Parser {
         // let command_arg = args.take(4).collect::<Vec<&str>>();
         let height = args.next();
         if height.is_none() {
-            self.print_help("get-block".split(" "));
+            self.print_help("get-block".split(' '));
             return;
         }
         let height = match height.unwrap().parse::<u64>().ok() {
             Some(height) => height,
             None => {
                 println!("Invalid block height provided. Height must be an integer.");
-                self.print_help("get-block".split(" "));
+                self.print_help("get-block".split(' '));
                 return;
             },
         };
@@ -776,7 +774,7 @@ impl Parser {
             None => Format::Text,
             Some(_) => {
                 println!("Unrecognized format sspecifier");
-                self.print_help("get-block".split(" "));
+                self.print_help("get-block".split(' '));
                 return;
             },
         };
@@ -798,7 +796,7 @@ impl Parser {
                         historical_block
                             .block
                             .to_json()
-                            .unwrap_or("Error deserializing block".into())
+                            .unwrap_or_else(|_| "Error deserializing block".into())
                     ),
                     (None, _) => println!("Block not found at height {}", height),
                 },
@@ -811,14 +809,14 @@ impl Parser {
         // let command_arg = args.take(4).collect::<Vec<&str>>();
         let hex = args.next();
         if hex.is_none() {
-            self.print_help("search-utxo".split(" "));
+            self.print_help("search-utxo".split(' '));
             return;
         }
         let commitment = match Commitment::from_hex(&hex.unwrap().to_string()) {
             Ok(v) => v,
             _ => {
                 println!("Invalid commitment provided.");
-                self.print_help("search-utxo".split(" "));
+                self.print_help("search-utxo".split(' '));
                 return;
             },
         };
@@ -849,14 +847,14 @@ impl Parser {
         // let command_arg = args.take(4).collect::<Vec<&str>>();
         let hex = args.next();
         if hex.is_none() {
-            self.print_help("search-stxo".split(" "));
+            self.print_help("search-stxo".split(' '));
             return;
         }
         let commitment = match Commitment::from_hex(&hex.unwrap().to_string()) {
             Ok(v) => v,
             _ => {
                 println!("Invalid commitment provided.");
-                self.print_help("search-stxo".split(" "));
+                self.print_help("search-stxo".split(' '));
                 return;
             },
         };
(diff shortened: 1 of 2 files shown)


applications/tari_base_node/src/parser.rs
```diff
@@ -275,7 +273,7 @@ impl Parser {
                 self.process_cancel_transaction(args);
             },
             SendTari => {
-                self.process_send_tari(args);
+                Self::process_send_tari(args, self.executor.clone(), self.wallet_transaction_service.clone());
```
sdbondi: I think you did this because you were using `self` in the spawned task (`&'a self` where 'a does not meet the 'static lifetime requirement of `spawn`) 

```suggestion
                self.process_send_tari(args);
```

The rest of the modification follows

applications/tari_base_node/src/parser.rs
```diff
@@ -1488,7 +1486,12 @@ impl Parser {
     }
 
     /// Function to process the send transaction command
-    fn process_send_tari<'a, I: Iterator<Item = &'a str>>(&mut self, mut args: I) {
+    fn process_send_tari<'a, I: Iterator<Item = &'a str>>(
+        mut args: I,
+        executor: runtime::Handle,
+        mut wallet_transaction_service: TransactionServiceHandle,
+    )
+    {
```
sdbondi:
```suggestion
fn process_send_tari<'a, I: Iterator<Item = &'a str>>(&mut self, mut args: I) {
```

applications/tari_base_node/src/parser.rs
```diff
@@ -1517,10 +1520,9 @@ impl Parser {
         let msg = args.collect::<Vec<&str>>().join(" ");
 
         let fee_per_gram = 25 * uT;
-        let mut txn_service = self.wallet_transaction_service.clone();
-        self.executor.spawn(async move {
-            let event_stream = txn_service.get_event_stream_fused();
-            match txn_service
+        executor.spawn(async move {
+            let event_stream = wallet_transaction_service.get_event_stream_fused();
+            match wallet_transaction_service
```
sdbondi: Clone `self.wallet_transaction_service` and move (i.e. pass ownership) the cloned version into the async block. A reference to `self` cannot be used with `executor.spawn` because the `spawn` requires a `'static` lifetime (i.e. all state that is used must either be owned or exist as long as the program execution).  

```suggestion
     let wallet_transaction_service = self.wallet_transaction_service.clone();

        self.executor.spawn(async move {
            let event_stream = wallet_transaction_service.get_event_stream_fused();
            match wallet_transaction_service
```

sdbondi (commented):
LGTM - Think you just need to clone wallet_transaction_service and you can keep `process_send_tari` the same as all the other functions

applications/tari_base_node/src/parser.rs
```diff
@@ -1686,16 +1690,64 @@ impl Parser {
...
+                            " " +
+                            &msg;
+                        let args = command_str.split_whitespace();
+                        // Execute
+                        let millis_actual_i = (Utc::now() - start).num_milliseconds() as u64;
+                        let millis_target_i = (i as f64 / (tx_per_s / 1000.0)) as u64;
+                        if millis_target_i - millis_actual_i > 0 {
+                            tokio::time::delay_for(Duration::from_millis(millis_target_i - millis_actual_i)).await;
+                        }
+                        Self::process_send_tari(args, executor.clone(), wallet_transaction_service.clone());
```
SWvheerden: Would it not be more ergonomic and more intuitive to rather move the send logic out from `process_send_tari(..)`
And then let make_it_rain and process_send_tari call that function. 

Currently here you have to create "fake" arguments and then reparse them again to actual arguments.

hansieodendaal: Hi thank you for the comment. ~~I do not agree for the following reasons:~~
~~- This is a specific variation for `process_send_tari`, and a custom command must be created and passed to the function, thus no duplication.~~
~~- W.r.t. the send logic, the `process_make_it_rain` function uses all the send logic of the `process_send_tari`, but not vice versa.~~

Will make a separate function.

applications/tari_base_node/src/parser.rs
hansieodendaal: Thank you for the comments. As discussed, this was a secondary change due to `self.process_send_tari` being called lower down in a spawned task inside a loop. Will change the interface to be consistent again and create a separate async function that can be called by both `process_send_tari` and `process_make_it_rain`, similar to @SWvheerden's suggestion.

applications/tari_base_node/src/parser.rs
hansieodendaal: As above

applications/tari_base_node/src/parser.rs
hansieodendaal: As above

sdbondi (dismissed):
Nice - LGTM!

SWvheerden (dismissed):

applications/tari_base_node/src/parser.rs
```diff
@@ -1516,79 +1514,22 @@ impl Parser {
...
-                    warn!(target: LOG_TARGET, "Error communicating with wallet: {:?}", e);
-                    return;
-                },
-                Ok(_) => println!("Sending {} Tari to {} ", amount, dest_pubkey),
-            };
+            send_tari(
+                amount,
+                dest_pubkey.clone(),
+                msg.clone(),
+                wallet_transaction_service.clone(),
```
stringhandler: nit: double clone here

stringhandler (dismissed):
Untested, but looks good. The timeout on the config may be a bit excessive, but lets see how it fares

stringhandler:
Looks like `rust-fmt` failed though...

applications/tari_base_node/src/parser.rs
hansieodendaal: Done

hansieodendaal:
> Looks like `rust-fmt` failed though...

Ran `cargo fmt --all` twice now :-)

applications/tari_base_node/src/parser.rs
```diff
@@ -1686,16 +1622,72 @@ impl Parser {
...
+                        let msg = msg.clone();
+                        executor.spawn(async move {
+                            send_tari(
+                                start_amount + amount_inc * (i as u64),
+                                dest_pubkey,
+                                msg,
+                                wallet_transaction_service,
+                            )
+                            .await;
+                        });
```
sdbondi, line 1674:
```suggestion
                        executor.spawn(
                            send_tari(
                                start_amount + amount_inc * (i as u64),
                                dest_pubkey,
                                msg,
                                wallet_transaction_service,
                            )
                           
                        );
```

It feels like spawning `n` tasks is redundant because send_transaction already does this internally after preparing the transaction. So the gains in tps should be minimal - that being said, I don't think there's much harm either so let's see.

sdbondi (approved):
LGTM

PR #2105, opened 2020-08-04
Rollback on pruned sync validation failure

sdbondi:
## Description

This PR adds an `EffectivePrunedHeight` chain metadata entry and
`InProgressPrunedSyncState` metadata item to the blockchain db.
`InProgressPrunedSyncState` is constructed and persisted once pruned sync
starts and is used to keep track of the start index of "temporary"
checkpoints created during sync.

If a horizon state validation error occurs the sync data is rolled back
by fetching each temporary checkpoint and reversing the operations and
deleting the checkpoints. A bit more intelligence can be added here depending 
on the kind of validation failure, certain state could not be rolled back and therefore 
not need to be redownloaded. 

If horizon sync is interrupted, the state is loaded again and the sync is resumed.
_EDIT: If a request times out, or some other non-validation
failure occurs the sync state is kept and the sync is resumed later ( WAITING>LISTENING>HEADER_SYNC>HORIZON_SYNC)_ 

Once the sync is complete, the InProgressPrunedSyncState is deleted and
the chain metadata updated accordingly.

Other changes:

- Sync peers now contains each peer's chain metadata so that an appropriate
  peer can be selected for block sync i.e. a peer with an effective
  pruned height lower than the syncing nodes current height.
- Header sync does not reset the chain metadata to genesis anymore as a
  that should be in pruned sync.
- Prefixed all database functions that ony are applicable for horizon
  sync with `horizon_sync_` to indicate that they are specialized and
  shouldn't used in other areas of the system at this time.
- DRY'd up all the checkpoint "utility" functions. It is no longer
  required to maintain separate functions for lmdb and memorydb.
- Removed `n` clones when storing a block
- Added `set_metadata` and `delete_metadata` convenience calls to
  `DbTransaction`
- Optimised chain balance validation to not have to hold every
  kernel/UTXO in memory while summing commitments.
- A few error improvements and migration to `thiserror`
- New basic test for horizon sync rollback

## Motivation and Context

Previously, if a horizon state validation error occurs the database needed to be deleted to recover. This PR adds some simple metadata to allow the database to rollback in this case.  

## How Has This Been Tested?

Added new test for behaviour when validation fails. Admittedly, this test is a little simple and more effort needs to be put into testing. A few scenarios were tested on a 2 node network and current testnet:

1. Sync from 0 to tip
1. Interrupt sync and test that it resumes on restart
1. Artificially fail validation and test that the sync rolls back, resyncs, fails, rollsback. Allow validation to succeed, and check that sync succeeds
1. Set a large pruned height, sync until block sync has started - stop the base node, set a small pruned height - begin horizon sync and check that it succeeds without redownloading state that was obtained during block sync.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/chain_metadata_service/service.rs
+++ b/base_layer/core/src/base_node/chain_metadata_service/service.rs
@@ -294,6 +294,7 @@ mod test {
             height_of_longest_chain: Some(1),
             best_block: Some(vec![]),
             pruning_horizon: 64,
+            effective_pruned_height: 0,
             accumulated_difficulty: 1.into(),
         }
     }
--- a/base_layer/core/src/base_node/states/sync_peers.rs
+++ b/base_layer/core/src/base_node/states/sync_peers.rs
@@ -20,45 +20,8 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use crate::{
-    base_node::{comms_interface::CommsInterfaceError, states::block_sync::BlockSyncError},
-    chain_storage::{ChainStorageError, MmrTree},
-    transactions::transaction::TransactionError,
-    validation::ValidationError,
-};
-use thiserror::Error;
-use tokio::task;
+use crate::base_node::chain_metadata_service::PeerChainMetadata;
 
-#[derive(Debug, Error)]
-pub enum HorizonSyncError {
-    #[error("Peer sent an empty response")]
-    EmptyResponse,
-    #[error("Peer sent an invalid response")]
-    IncorrectResponse,
-    #[error("Exceeded maximum sync attempts")]
-    MaxSyncAttemptsReached,
-    #[error("Chain storage error: {0}")]
-    ChainStorageError(#[from] ChainStorageError),
-    #[error("Comms interface error: {0:?}")]
-    CommsInterfaceError(#[from] CommsInterfaceError),
-    #[error("Block sync error: {0:?}")]
-    BlockSyncError(#[from] BlockSyncError),
-    #[error("Final state validation failed: {0}")]
-    FinalStateValidationFailed(ValidationError),
-    #[error("Join error: {0}")]
-    JoinError(#[from] task::JoinError),
-    #[error("Invalid kernel signature: {0}")]
-    InvalidKernelSignature(TransactionError),
-    #[error("Validation failed for {0} MMR")]
-    InvalidMmrRoot(MmrTree),
-}
-
-impl HorizonSyncError {
-    pub fn is_recoverable(&self) -> bool {
-        use HorizonSyncError::*;
-        match self {
-            FinalStateValidationFailed(_) | InvalidMmrRoot(_) => false,
-            _ => true,
-        }
-    }
-}
+pub type SyncPeer = PeerChainMetadata;
+/// Type alias for a collection of PeerChainMetadata
+pub type SyncPeers = Vec<SyncPeer>;
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -24,19 +24,20 @@ use crate::{
     chain_storage::{
         consts::{
             BLOCKCHAIN_DATABASE_ORPHAN_STORAGE_CAPACITY,
-            BLOCKCHAIN_DATABASE_PRUNED_MODE_CLEANUP_INTERVAL,
+            BLOCKCHAIN_DATABASE_PRUNED_MODE_PRUNING_INTERVAL,
             BLOCKCHAIN_DATABASE_PRUNING_HORIZON,
         },
-        db_transaction::{DbKey, DbKeyValuePair, DbTransaction, DbValue, MetadataKey, MetadataValue, MmrTree},
+        db_transaction::{DbKey, DbTransaction, DbValue, MetadataKey, MetadataValue, MmrTree},
         error::ChainStorageError,
         ChainMetadata,
         HistoricalBlock,
+        InProgressHorizonSyncState,
     },
     consensus::ConsensusManager,
     proof_of_work::{Difficulty, PowAlgorithm, ProofOfWork},
     transactions::{
         transaction::{TransactionInput, TransactionKernel, TransactionOutput},
-        types::{Commitment, HashOutput, Signature},
+        types::{Commitment, HashOutput, PublicKey, Signature},
     },
     validation::{StatelessValidation, StatelessValidator, Validation, ValidationError, Validator},
 };
@@ -45,6 +46,7 @@ use log::*;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::VecDeque,
+    convert::TryFrom,
     sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 use strum_macros::Display;
@@ -58,7 +60,7 @@ const LOG_TARGET: &str = "c::cs::database";
 pub struct BlockchainDatabaseConfig {
     pub orphan_storage_capacity: usize,
     pub pruning_horizon: u64,
-    pub pruned_mode_cleanup_interval: u64,
+    pub pruning_interval: u64,
 }
 
 impl Default for BlockchainDatabaseConfig {
@@ -66,7 +68,7 @@ impl Default for BlockchainDatabaseConfig {
         Self {
             orphan_storage_capacity: BLOCKCHAIN_DATABASE_ORPHAN_STORAGE_CAPACITY,
             pruning_horizon: BLOCKCHAIN_DATABASE_PRUNING_HORIZON,
-            pruned_mode_cleanup_interval: BLOCKCHAIN_DATABASE_PRUNED_MODE_CLEANUP_INTERVAL,
+            pruning_interval: BLOCKCHAIN_DATABASE_PRUNED_MODE_PRUNING_INTERVAL,
         }
     }
 }
@@ -157,9 +159,16 @@ pub trait BlockchainBackend: Send + Sync {
     ) -> Result<HashOutput, ChainStorageError>;
     /// Constructs a merkle proof for the specified merkle mountain range and the given leaf position.
     fn fetch_mmr_proof(&self, tree: MmrTree, pos: usize) -> Result<MerkleProof, ChainStorageError>;
-    /// Fetches the checkpoint corresponding to the provided height, the checkpoint consist of the list of nodes
-    /// added & deleted for the given Merkle tree.
-    fn fetch_checkpoint(&self, tree: MmrTree, height: u64) -> Result<MerkleCheckPoint, ChainStorageError>;
+    /// Fetches the `MerkleCheckPoint` corresponding to the given height. In pruned mode, the underlying database may
+    /// not be able to provide the checkpoint as it has been merged into the base checkpoint. In this case a
+    /// `BeyondPruningHorizon` error is returned.
+    fn fetch_checkpoint_at_height(&self, tree: MmrTree, height: u64) -> Result<MerkleCheckPoint, ChainStorageError>;
+    /// Fetches the `MerkleCheckPoint` at the given index
+    fn fetch_checkpoint_at_index(
+        &self,
+        tree: MmrTree,
+        index: usize,
+    ) -> Result<Option<MerkleCheckPoint>, ChainStorageError>;
     /// Fetches the total merkle mountain range node count upto the specified height.
     fn fetch_mmr_node_count(&self, tree: MmrTree, height: u64) -> Result<u32, ChainStorageError>;
     /// Fetches the leaf node hash and its deletion status for the nth leaf node in the given MMR tree. The height
@@ -210,7 +219,7 @@ pub trait BlockchainBackend: Send + Sync {
     /// Returns the stored header with the highest corresponding height.
     fn fetch_last_header(&self) -> Result<Option<BlockHeader>, ChainStorageError>;
     /// Returns the stored chain metadata.
-    fn fetch_metadata(&self) -> Result<ChainMetadata, ChainStorageError>;
+    fn fetch_chain_metadata(&self) -> Result<ChainMetadata, ChainStorageError>;
     /// Returns the set of target difficulties for the specified proof of work algorithm.
     fn fetch_target_difficulties(
         &self,
@@ -223,6 +232,8 @@ pub trait BlockchainBackend: Send + Sync {
     fn count_utxos(&self) -> Result<usize, ChainStorageError>;
     /// Returns the kernel count
     fn count_kernels(&self) -> Result<usize, ChainStorageError>;
+    /// Returns the checkpoint count for the given MmrTree
+    fn count_checkpoints(&self, tree: MmrTree) -> Result<usize, ChainStorageError>;
     /// Validate the Merkle root for the given `MmrTree` matches the header at the given height
     fn validate_merkle_root(&self, tree: MmrTree, height: u64) -> Result<bool, ChainStorageError>;
 }
@@ -268,36 +279,39 @@ macro_rules! fetch {
 /// let db = BlockchainDatabase::new(db_backend, &rules, validators, BlockchainDatabaseConfig::default()).unwrap();
 /// // Do stuff with db
 /// ```
-pub struct BlockchainDatabase<T> {
-    db: Arc<RwLock<T>>,
-    validators: Validators<T>,
+pub struct BlockchainDatabase<B> {
+    db: Arc<RwLock<B>>,
+    validators: Validators<B>,
     config: BlockchainDatabaseConfig,
 }
 
-impl<T> BlockchainDatabase<T>
-where T: BlockchainBackend
+impl<B> BlockchainDatabase<B>
+where B: BlockchainBackend
 {
     /// Creates a new `BlockchainDatabase` using the provided backend.
     pub fn new(
-        db: T,
+        db: B,
         consensus_manager: &ConsensusManager,
-        validators: Validators<T>,
+        validators: Validators<B>,
         config: BlockchainDatabaseConfig,
     ) -> Result<Self, ChainStorageError>
     {
+        debug!(
+            target: LOG_TARGET,
+            "Initializing database pj={}", config.pruning_horizon
+        );
         let blockchain_db = BlockchainDatabase {
             db: Arc::new(RwLock::new(db)),
             validators,
             config,
         };
-        let metadata = blockchain_db.get_metadata()?;
+        let metadata = blockchain_db.get_chain_metadata()?;
         if metadata.height_of_longest_chain.is_none() {
             let genesis_block = consensus_manager.get_genesis_block();
             blockchain_db.store_new_block(genesis_block)?;
             blockchain_db.store_pruning_horizon(config.pruning_horizon)?;
-        } else if (metadata.is_archival_node() && (config.pruning_horizon != metadata.pruning_horizon)) ||
-            (metadata.is_pruned_node() && (config.pruning_horizon < metadata.pruning_horizon))
-        {
+        }
+        if config.pruning_horizon != metadata.pruning_horizon {
             debug!(
                 target: LOG_TARGET,
                 "Updating pruning horizon from {} to {}.", metadata.pruning_horizon, config.pruning_horizon,
@@ -309,7 +323,7 @@ where T: BlockchainBackend
 
     // Be careful about making this method public. Rather use `db_and_metadata_read_access`
     // so that metadata and db are read in the correct order so that deadlocks don't occur
-    pub fn db_read_access(&self) -> Result<RwLockReadGuard<T>, ChainStorageError> {
+    pub fn db_read_access(&self) -> Result<RwLockReadGuard<B>, ChainStorageError> {
         self.db.read().map_err(|e| {
             error!(
                 target: LOG_TARGET,
@@ -319,7 +333,7 @@ where T: BlockchainBackend
         })
     }
 
-    pub fn db_write_access(&self) -> Result<RwLockWriteGuard<T>, ChainStorageError> {
+    fn db_write_access(&self) -> Result<RwLockWriteGuard<B>, ChainStorageError> {
         self.db.write().map_err(|e| {
             error!(
                 target: LOG_TARGET,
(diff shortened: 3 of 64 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -157,9 +159,12 @@ pub trait BlockchainBackend: Send + Sync {
...
     /// Constructs a merkle proof for the specified merkle mountain range and the given leaf position.
     fn fetch_mmr_proof(&self, tree: MmrTree, pos: usize) -> Result<MerkleProof, ChainStorageError>;
-    /// Fetches the checkpoint corresponding to the provided height, the checkpoint consist of the list of nodes
-    /// added & deleted for the given Merkle tree.
+    /// Fetches the `MerkleCheckPoint` corresponding to the given height. In pruned mode, the underlying database may
+    /// not be able to provide the checkpoint as it has been merged into the base checkpoint. In this case a
+    /// `BeyondPruningHorizon` error is returned.
     fn fetch_checkpoint(&self, tree: MmrTree, height: u64) -> Result<MerkleCheckPoint, ChainStorageError>;
+    /// Fetches the `MerkleCheckPoint` at the given index
+    fn fetch_checkpoint_at(&self, tree: MmrTree, index: usize) -> Result<Option<MerkleCheckPoint>, ChainStorageError>;
```
SWvheerden: Whats the diff here. 
Would checkpoint at always reference the first checkpoint the node has, while height is the chain height?

sdbondi: Yup exactly, this fetches by index because the "temporary" checkpoints don't correspond to any particular height during horizon sync.

SWvheerden:
I have not finished the going over the PR yet. 
But I have on big question. Why are we doing roll back on sync validation?

When syncing pruned mode, we have a single big validation check already built in the MMRs. The MMR's of all synced nodes will be identical. Pruned and archival nodes DONT deleted from the MMR's. The MMR's themselves wont be too big. 
We must download only the MMR's and verify them according to the header checkpoints. 

If thats done, you can ask nodes for individual utxo's and the like. If that download fails, just ask for that item again. Because you have a synced MMR's you know what data you dont have and what you have. You check your own data and fill in whats required.

sdbondi:
In order for sync to be fast (i.e reduce the number of request - wait - response cycles), I think we should download the kernels, UTXOs etc and verify the merkle roots at the sync height. We don't necessarily need to download the MMRs separately as all the hashes can be obtained from the downloaded items themselves. My reasoning: downloading a bunch of hashes, then sending those same hashes back to request the items you need is wasteful and slow (`n` kernels it will be `2n` hashes down and up the wire, for UTXOs it obviously depends but worst case is `2n`).

Note: currently, we DO download the hashes and then reupload them but that may/will/can change when a supporting network call to fetch/stream the kernels/UTXOs/RPs from the requested indexes.

I do see what you are saying about downloading small hashes and being able to validate that you have the "strongest" state. But either way we're going to have validate the final chain balance.  
If that fails, what should happen? Before this PR, you would need to delete your database. Now the database can recover from this. It could potentially recover without nuking all the state downloaded in the last horizon sync, but I feel this is a good start and better than having to delete your db and start from scratch.  

> If that download fails, just ask for that item again. 

This is what happens already :) The state is only rolled back if the final state validation or merkle root validation fails.

> The MMR's of all synced nodes will be identical.

I don't think we can make that assumption, we only know if this is true once we've downloaded up to the sync height and we must still handle if that is not the case - currently, this is handled by removing all the (bad) state. 

> We must download only the MMR's and verify them according to the header checkpoints

The one issue with verifying against "header checkpoints" is that pruned nodes can't give you MMRs at an arbitrary height.
If a peer has an effective PH of 1000, and we want to sync to height 900 (because our PH is different) - as things currently stand - we can't sync from that peer because they have lost the height mappings <= 1000. I do think that we can remove this limitation though, either way we need the state so perhaps block sync shouldn't care of we have some of that state already (headers, kernels etc)

In principle, you could build up the merkle tree one by one and map it to a particular header by checking the merkle roots but I don't think this is trivially implemented with the current backend and may not be particularly efficient.

SWvheerden:
> In order for sync to be fast (i.e reduce the number of request - wait - response cycles), I think we should download the kernels, UTXOs etc and verify the merkle roots at the sync height. We don't necessarily need to download the MMRs separately as all the hashes can be obtained from the downloaded items themselves. My reasoning: downloading a bunch of hashes, then sending those same hashes back to request the items you need is wasteful and slow (`n` kernels it will be `2n` hashes down and up the wire, for UTXOs it obviously depends but worst case is `2n`).

The thing is you have to download the MMR's first as you cant construct an mmr is the utxo set is not complete. Which it wont be for a pruned node. If a utxo is spent, it deleted. So the only remaining data is the mmr. So you cant construct the mmr from the utxo's you deleted. And if you download and construct your mmr with the utxo's in a different order then your mmr's will not match the header tip. So you have to download those mmr nodes to get the "missing" pieces anyway. 

I get the fact that for the kernels this data will be 2n * 8 bytes extra downloading. Download this way is just way quicker to get to a state thats validate-able.


> I do see what you are saying about downloading small hashes and being able to validate that you have the "strongest" state. But either way we're going to have validate the final chain balance.
> If that fails, what should happen? Before this PR, you would need to delete your database. Now the database can recover from this. It could potentially recover without nuking all the state downloaded in the last horizon sync, but I feel this is a good start and better than having to delete your db and start from scratch.

Well before downloading anything you would need to check the header's pow. If the chain balance fails. You know that the longest chain does not follow your consensus code. So you need to look to another node that can supply you a different chain. 


> > The MMR's of all synced nodes will be identical.
> 
> I don't think we can make that assumption, we only know if this is true once we've downloaded up to the sync height and we must still handle if that is not the case - currently, this is handled by removing all the (bad) state.

The MMR's for all nodes on that chain you're syncing to will have to be identical. If they are not their mmr roots wont match. The fact that you are downloading to a specific sync height matched to a specific header, means you can easily verify the mmr data and see that its correct before downloading any data. 


> 
> > We must download only the MMR's and verify them according to the header checkpoints
> 
> The one issue with verifying against "header checkpoints" is that pruned nodes can't give you MMRs at an arbitrary height.
> If a peer has an effective PH of 1000, and we want to sync to height 900 (because our PH is different) - as things currently stand - we can't sync from that peer because they have lost the height mappings <= 1000. I do think that we can remove this limitation though, either way we need the state so perhaps block sync shouldn't care of we have some of that state already (headers, kernels etc)
> 
> In principle, you could build up the merkle tree one by one and map it to a particular header by checking the merkle roots but I don't think this is trivially implemented with the current backend and may not be particularly efficient.

Lets clarify this. Are you talking about header's or the mmr_checkpoint struct in the code. Irrespective of what height you're downloading to and what you're downloading. You can never download to a height lower than the PH of a pruned node you are downloading from.  And the problem is the utxos. The can be spent and then deleted. But he can still provide you with a header of how the root's look like at block 900. He can only provide you with data for 1000 and onwards. If you have the mmr's downloaded for block 900 and verified. That node can provide you with the utxo info was well. 



The biggest problem with syncing is syncing from pruned nodes. And the problem syning form them is the utxo's. Apart form the height issue with PH. The dont have a complete list state of the utxo for the mmr. But they all have a complete mmr. And that MMR will be in sync for every node on a specific chain.

base_layer/core/src/chain_storage/blockchain_database.rs
stringhandler: It might be best to rename `fetch_checkpoint` to `fetch_checkpoint_at_height`

sdbondi: Agreed - done

base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -294,6 +294,7 @@ mod test {
             height_of_longest_chain: Some(1),
             best_block: Some(vec![]),
             pruning_horizon: 64,
```
stringhandler, line 296: This value can probably come out now

base_layer/core/src/base_node/states/sync_peers.rs
```diff
@@ -0,0 +1,27 @@
...
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use crate::base_node::chain_metadata_service::PeerChainMetadata;
+
+pub type SyncPeer = PeerChainMetadata;
+/// Type alias for a collection of PeerChainMetadata
+pub type SyncPeers = Vec<SyncPeer>;
```
stringhandler, line 27: I don't think these aliases add much value, and it can be confusing to try find these types. `SyncPeer` is perhaps ok, but I think it's more clear to have `Vec<SyncPeer>` than to have `SyncPeers`. It implies that there is some extra logic that is available in SyncPeer. I actually think having the original types (`Vec<PeerChainMetadata>`) is the most clear

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -268,29 +275,33 @@ macro_rules! fetch {
 /// let db = BlockchainDatabase::new(db_backend, &rules, validators, BlockchainDatabaseConfig::default()).unwrap();
 /// // Do stuff with db
 /// ```
-pub struct BlockchainDatabase<T> {
-    db: Arc<RwLock<T>>,
-    validators: Validators<T>,
+pub struct BlockchainDatabase<B> {
```
stringhandler, line 282: nit: I think it's more common to use `T` or `TBackend` in generics, but I don't mind this

stringhandler (commented):
Some nice code clean up in this PR. It's quite involved, so I can't really say much without testing it

stringhandler:
When testing this, I found that if I changed the `pruning_horizon` in the config, it still had the old value. 
config:
```
pruning_horizon = 60000

```
`get_chain_metadata`
```
>> get-chain-metadata
Height of longest chain : 0
Geometric mean of longest chain : 0
Best block : 4a5af9e5dd60c2ab012ef8d75a465c6864b0e59c258f3ea76ea45ae93a332fac
Pruning horizon : 10000
Effective pruned height : 0
```

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
```diff
@@ -98,7 +99,25 @@ where
...
+            error!(
+                target: LOG_TARGET,
+                "Could not add delete value into lmdb transaction: {:?}", e
+            );
+            ChainStorageError::AccessError(e.to_string())
+        })
+}
+
+/// Deletes the given key. An error is returned if the key does not exist
+pub fn lmdb_delete_checked<K>(txn: &WriteTransaction, db: &Database, key: &K) -> Result<(), ChainStorageError>
```
SWvheerden: Should the names not be swapped around?
I would take it that *_checked would returned the result of the checked which is a bool. And delete would just return a result.

base_layer/core/src/chain_storage/checkpoint_utils.rs
```diff
@@ -0,0 +1,140 @@
...
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//! ## Common blockchain utility functions for ArrayLike impls
+
+use crate::chain_storage::ChainStorageError;
+use std::cmp;
+use tari_mmr::{ArrayLike, ArrayLikeExt, MerkleCheckPoint};
+
+/// Calculate the total leaf node count upto a specified height. If the height is less than the effective pruning
+/// horizon the total number of nodes up to the effective pruned height is returned.
```
SWvheerden, line 30: I am against moving this out to chain storage. It leaks Checkpoints to the higher level. Checkpoints are only required for a key-value store. 
I know out current implementation already leaks it out. But we dont have to increase that. I would keep this in lmdb.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -136,23 +146,24 @@ where D: Digest + Send + Sync
             orphans_db: get_database(&store, LMDB_DB_ORPHANS)?,
             utxo_mmr: MmrCache::new(MemDbVec::new(), utxo_checkpoints.clone(), mmr_cache_config)?,
             curr_utxo_checkpoint: {
-                let acc_count = fetch_last_mmr_node_added_count(&utxo_checkpoints)?;
+                let acc_count = checkpoint_utils::fetch_last_mmr_node_added_count(&utxo_checkpoints)?;
```
SWvheerden, line 140: see comment about checkpoint_utils above

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -600,24 +632,43 @@ where D: Digest + Send + Sync
...
+            },
+            MmrTree::RangeProof => {
+                checkpoint_utils::fetch_checkpoint(&self.range_proof_checkpoints, pruned_mode, tip_height, height)
+            },
         }
         .map_err(|e| ChainStorageError::AccessError(format!("Checkpoint error: {}", e.to_string())))?
         .ok_or_else(|| ChainStorageError::OutOfRange)
     }
 
+    fn fetch_checkpoint_at(&self, tree: MmrTree, index: usize) -> Result<Option<MerkleCheckPoint>, ChainStorageError> {
```
SWvheerden: should we not rather make this at_height. 
That way its very clear which cp youre fetching

base_layer/core/src/chain_storage/lmdb_db/lmdb_vec.rs
```diff
@@ -96,6 +99,47 @@ where
...
+    //     let iter = CursorIter::new(cursor, &access, head, lmdb::lmdb_iter_next).map_err(|e| {
+    //         error!(target: LOG_TARGET, "Could not get next cursor from lmdb: {:?}", e);
+    //         ChainStorageError::AccessError(e.to_string())
+    //     })?;
+    //     let iter = match range.end_bound() {
+    //         Bound::Included(v) => iter.take(*v),
+    //         Bound::Excluded(v) => Ok(iter.take(*v - 1).map(|(_, v)| v).collect()),
+    //         Bound::Unbounded => Ok(iter.map(|(_, v)| v).collect()),
+    //     }
+    // }
```
SWvheerden: ? Commented out code?
Is there a reason for this?

SWvheerden: If its unused, but we are not sure why its there and we would like to keep it for a short while as test. Put a comment about that in, with the data this was done.

base_layer/core/src/chain_storage/memory_db/memory_db.rs
```diff
@@ -306,21 +319,22 @@ where D: Digest + Send + Sync
                 },
                 WriteOperation::RewindMmr(tree, steps_back) => match tree {
                     MmrTree::Kernel => {
-                        let last_cp = rewind_checkpoints(&mut db.kernel_checkpoints, steps_back)?;
+                        let last_cp = checkpoint_utils::rewind_checkpoints(&mut db.kernel_checkpoints, steps_back)?;
```
SWvheerden, line 322: See comment above about checkpoint utils.

base_layer/core/src/chain_storage/memory_db/memory_db.rs
```diff
@@ -691,6 +726,25 @@ where D: Digest + Send + Sync
             ),
         }
     }
+
+    fn fetch_checkpoint_at(&self, tree: MmrTree, index: usize) -> Result<Option<MerkleCheckPoint>, ChainStorageError> {
```
SWvheerden: Same comment about at_index than above?

base_layer/core/src/proof_of_work/blake_pow.rs
```diff
@@ -38,12 +38,12 @@ pub fn blake_difficulty(header: &BlockHeader) -> Difficulty {
 
 pub fn blake_difficulty_with_hash(header: &BlockHeader) -> (Difficulty, Vec<u8>) {
     let bytes = header.hash();
-    let hash = Blake2b::digest(&bytes).to_vec();
-    let hash = Blake256::digest(&hash).to_vec();
+    let hash = Blake2b::digest(&bytes);
+    let hash = Blake256::digest(&hash);
```
SWvheerden, line 42: I think this just made mining a lot faster :)

base_layer/core/src/validation/helpers.rs
```diff
@@ -127,3 +126,20 @@ pub fn check_achieved_and_target_difficulty<B: BlockchainBackend>(
     }
     Ok(())
 }
+
+pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, ValidationError> {
```
SWvheerden, line 146: Should we not put this in the chain_storage rather?

SWvheerden (commented):
Looks good
I left a few questions

base_layer/core/src/chain_storage/checkpoint_utils.rs
sdbondi: It was duplicated in lmdb_db and memory_db. The duplication causes time wasted working out why tests don't work (because you change something in one and not the other) and it's a recipe for bugs. This makes the leak problem smaller by containing it in one place - we need only look at this file to know what operations are used RE checkpoints which should make it easier to refactor later.
This module is not public to anything outside of chain storage. Where should I put code that is common to both dbs?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: at_index you mean?

base_layer/core/src/chain_storage/lmdb_db/lmdb_vec.rs
sdbondi: Nope - was just left in by mistake

base_layer/core/src/proof_of_work/blake_pow.rs
sdbondi: 🎉

base_layer/core/src/chain_storage/memory_db/memory_db.rs
sdbondi: :+1:

base_layer/core/src/validation/helpers.rs
sdbondi: As in add it to BlockchainDatabase? I do agree that validators should not really care about how is_stxo is found

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
sdbondi: Hmm, so this was a bit lazy - I should just fix the reason why we need two delete functions: the error (including "value not found" is squashed into a string) which makes it hard to know if the value you deleted didn't exist in the first place (which is fine in many cases) or some other error happened.

sdbondi: Fixed this by removing the extra function and adding an error variant that we can handle (ye it's not perfect)

base_layer/core/src/chain_storage/checkpoint_utils.rs
SWvheerden: I would put this with lmdb, and give memory db access to this. 
The reasoning behind this. Is that memory db is suppose to work like the lmdb, thus sharing it should be okay.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
SWvheerden: yes

base_layer/core/src/validation/helpers.rs
SWvheerden: yes

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Ye just keeping it consistent with the existing code (i.e `B: BlockchainBackend` everywhere)

base_layer/core/src/base_node/states/sync_peers.rs
sdbondi: This is a placeholder for a SyncPeers type that we may need to improve peer selection (i.e. it may allow us to ask the chain metadata service for another set of peers that satisfy some requirements, e.g have a low enough effective pruned height)

base_layer/core/src/chain_storage/checkpoint_utils.rs
sdbondi: I disagree

base_layer/core/src/validation/helpers.rs
sdbondi: We'd need to add to the `BlockchainBackend` trait (validators only know about the trait (which is probably not correct). So I vote for not doing that in this PR and making that part of a refactor for validators. At least this "helper" is confined to validators - so you know that it only is used there

base_layer/core/src/chain_storage/checkpoint_utils.rs
sdbondi: I disagree - but @mikethetike what do you think - I'll do whatever others in the community agree on

sdbondi: To be clear, I agree with what you're saying, but I disagree with how to solve it - and with the issue with putting it in the `chain_storage::checkpoint_utils` module (though I guess it depends on if a developer decides to "just use" these functions where ever within `chain_storage` because "it's there").

We should then move `lmdb` and `memory` into `backends`, move the `checkpoint_utils` stuff into `backends` and have it private to `backends`. I'm sure you can see why I chose this way, this PR is already large enough without moving large swaths of code around.

sdbondi:
@mikethetike 
> When testing this, I found that if I changed the `pruning_horizon` in the config, it still had the old value.
> config:

This was by design (i.e. you can't have a pruning horizon larger than you had before until more blocks come in) but changing it to a smaller PH would work. However, since this PR introduces effective pruning horizon which is always the source of truth for the pruned height, the pruning horizon should just be set to whatever is in the config. I've made this change.

base_layer/core/src/base_node/chain_metadata_service/service.rs
sdbondi: Agreed, I'll see if there are any consequences to taking it out.

base_layer/core/src/chain_storage/checkpoint_utils.rs
stringhandler: I think having it in `checkpoint_utils` is fine. I don't think it's leaking anything that isn't available already. 

> I would put this with lmdb, and give memory db access to this.

I don't agree with this approach. `lmdb` and `memory` should be distinct to the point where we could theoretically have `memory` and `lmdb` feature flags.

base_layer/core/src/base_node/chain_metadata_service/service.rs
sdbondi: I looked at this, and it's used to determine if the node / an external peer is in archival/pruned mode - technically, we should only care that the node's effective pruning height is low enough for us to sync. There are a number of tests and area of the system that use `is_pruned_mode` so think it won't be a super tiny change - maybe left for another PR?

stringhandler (approved):
I'm happy to merge this in. I haven't tested it again since my last review, but happy with the code

PR #2107, opened 2020-08-04
Add instructions for running in pruned mode

stringhandler:


applications/tari_base_node/README.md
```diff
@@ -36,3 +36,9 @@ cargo install tari_base_node
 ```
 
 ## Configuration
+### Running in pruned mode
+To run your node in pruned mode, add a `pruning_horizon` value
```
sdbondi, line 40: Should probably mention that it should go under the `base_node.rincewind` section.

Issue #1718, opened 2020-04-16
Migrate from `derive-error` to `thiserror`/`anyhow`


CjS77:
`derive-error` served us well, but it's a little long in the tooth and does not seem to be very actively maintained anymore.

We've decided to standardize on [`thiserror`](https://crates.io/crates/thiserror) as the error handling library going forward. 

This entails migrating alllllllll the current error types over to `thiserror` and dropping the derive-error dependency as a finla goal.

delta1:
issue can be closed @mikethetike @CjS77 #2112

PR #2112, opened 2020-08-06
Migrate to thiserror and remove dep on derive-error

delta1:
## Description
All error types migrated to use `thiserror` crate. Dependency on `derive-error` removed. 

## Motivation and Context
#1718 motivates the migration of error types to `thiserror` crate in place of `derive-error`.
Also bumped `thiserror` to version `1.0.20` across the repo. 
In places where previous error messages were defined for `derive-error` I have used those. 
Where not previously defined, I have used generic descriptions of the errors - please advise if these are not suitable?

## How Has This Been Tested?
* [X] cargo test 
* [X] circle ci pipeline
* [X] manual sanity test

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/service/error.rs
+++ b/base_layer/core/src/base_node/service/error.rs
@@ -21,15 +21,17 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::base_node::comms_interface::CommsInterfaceError;
-use derive_error::Error;
 use tari_comms_dht::outbound::DhtOutboundError;
+use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum BaseNodeServiceError {
-    CommsInterfaceError(CommsInterfaceError),
-    DhtOutboundError(DhtOutboundError),
-    #[error(msg_embedded, no_from, non_std)]
+    #[error("Comms interface error: `{0}`")]
+    CommsInterfaceError(#[from] CommsInterfaceError),
+    #[error("DHT outbound error: `{0}`")]
+    DhtOutboundError(#[from] DhtOutboundError),
+    #[error("Invalid request error: `{0}`")]
     InvalidRequest(String),
-    #[error(msg_embedded, no_from, non_std)]
+    #[error("Invalid response error: `{0}`")]
     InvalidResponse(String),
 }
--- a/base_layer/core/src/mempool/error.rs
+++ b/base_layer/core/src/mempool/error.rs
@@ -30,21 +30,26 @@ use crate::{
     },
     transactions::transaction::TransactionError,
 };
-use derive_error::Error;
+use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum MempoolError {
-    UnconfirmedPoolError(UnconfirmedPoolError),
-    OrphanPoolError(OrphanPoolError),
-    PendingPoolError(PendingPoolError),
-    ReorgPoolError(ReorgPoolError),
-    TransactionError(TransactionError),
-    ChainStorageError(ChainStorageError),
-    /// The Blockchain height is undefined
+    #[error("Unconfirmed pool error: `{0}`")]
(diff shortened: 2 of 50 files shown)


base_layer/core/src/base_node/chain_metadata_service/error.rs
```diff
@@ -21,20 +21,23 @@
...
-    /// Peer did not send any chain metadata
+    #[error("Failed to decode chain metadata")]
+    DecodeError(#[from] DecodeError),
+    #[error("Peer did not send any chain metadata")]
     NoChainMetadata,
-    LivenessError(LivenessError),
-    CommsInterfaceError(CommsInterfaceError),
-    MessageError(MessageError),
-    /// Failed to publish `ChainMetadataEvent`
+    #[error("Liveness Error")]
```
sdbondi: Few other places where the actual error is not being output.
 
```suggestion
    #[error("Liveness Error: {0}")]
```

base_layer/core/src/base_node/service/error.rs
```diff
@@ -21,15 +21,17 @@
...
-use derive_error::Error;
 use tari_comms_dht::outbound::DhtOutboundError;
+use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum BaseNodeServiceError {
-    CommsInterfaceError(CommsInterfaceError),
-    DhtOutboundError(DhtOutboundError),
-    #[error(msg_embedded, no_from, non_std)]
+    #[error("Comms interface error")]
```
sdbondi:
```suggestion
    #[error("Comms interface error: {0}")]
```

same for most of the rest

base_layer/core/src/mempool/error.rs
```diff
@@ -30,21 +30,26 @@ use crate::{
...
 #[derive(Debug, Error)]
 pub enum MempoolError {
-    UnconfirmedPoolError(UnconfirmedPoolError),
-    OrphanPoolError(OrphanPoolError),
-    PendingPoolError(PendingPoolError),
-    ReorgPoolError(ReorgPoolError),
-    TransactionError(TransactionError),
-    ChainStorageError(ChainStorageError),
-    /// The Blockchain height is undefined
+    #[error("Unconfirmed pool error")]
```
sdbondi:
```suggestion
    #[error("Unconfirmed pool error: {0}")]
```

sdbondi (commented):
LGTM 🍸  - think that this PR should just make sure the info contained in each variant on the enum is output otherwise we could get a few useless errors like "Database Error" and "Message error"

```rust
 #[error("Message error: {0}")] // Add the `{0}`
    MessageError(#[from] MessageError),
```

base_layer/core/src/base_node/chain_metadata_service/error.rs
delta1: thanks @sdbondi - I thought that was only necessary for the "String" types, and #[from] took care of the others? 

How do I know which errors need that interpolation - is it all of them?

sdbondi: `#[from]` generates a [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) implementation to make it easy to convert say a `LivenessError` to a `ChainMetadataError`.

What `#[error("Liveness Error: {0}")]` is doing is calling the [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) implementation on `LivenessError` (which itself is implemented using `thiserror` and so has a `Display` implementation.

sdbondi: So I think most of them should have the "inner error" output - basically it depends on the meaning of the error, which is hard to evaluate in bulk, but we definitely want to err on the side of displaying the info vs hiding it.

delta1:
Perfect thanks for the feedback @sdbondi I'll get that done

sdbondi (dismissed):
:shipit:

delta1:
@sdbondi @mikethetike this PR is updated after merging all the current stuff from dev and fighting with git for a while

stringhandler (approved):

PR #2114, opened 2020-08-07
RFC: Tari Script - first draft

CjS77:
Provide a description of Tari Script, its implementation and its
applications.

Add tari script semantics

Describe extensions

### DIFF
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -157,7 +157,7 @@ malicious actors could take the UTXO for themselves.
 Therefore, it's imperative that the UTXO creator sign the script.
 
 Further, this signature must be present in the _kernel_ in some form, otherwise miners will be able to remove the script
-via cut-through.
+via cut-through, whereas kernels are never pruned.
 
 One approach to commit to the script hashes is to modify the output commitments using the [data commitments] approach
 suggested by [Phyro](https://github.com/phyro). In this approach, when creating a new UTXO, the owner also calculates
@@ -198,7 +198,7 @@ is modified with a commitment to the script hash, so
 
 $$ \hat{C} = C + \mathrm{H}(C \Vert s).G $$
 
-and wallets will sign the kernel with $$ k + \mathrm{H}(C_i \Vert s_i) $$ rather than just _k_.
+and wallets will sign the kernel with $$ k + \mathrm{H}(C \Vert s) $$ rather than just _k_.
 
 The overall and block balance checks must also be modified to use \\( \hat{C} \\) rather than _C_.
 
@@ -213,7 +213,7 @@ $$
   \sum(C_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(C_j + \mathrm{H}(C_j \Vert s_j).G) + \sum(\mathrm{fee}.G)  \\\\
   \text{If the accounting is correct, all values will cancel} \\\\
   \sum(k_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(k_j + \mathrm{H}(C_j \Vert s_j).G)  \\\\
-  \text{The sum of all the blinding factors is the definition of the excess,}\\; x_s\cdot G = X_s \\\\
+  \text{The sum of all the blinding factors (times G) is the definition of the excess,}\\; x_s\cdot G = X_s \\\\
   X_s + \sum(\mathrm{H}(C_i \Vert s_i).G) - \sum(\mathrm{H}(C_j \Vert s_j).G) \\\\
   \text{Define}\\, \Lambda = \sum(\mathrm{H}(C_i \Vert s_i).G) - \sum(\mathrm{H}(C_j \Vert s_j).G) \\\\
   X_s + \Lambda \\\\
@@ -254,9 +254,9 @@ In particular, UTXOs can still be pruned because the \\( \Lambda \\) values chan
 cancel out in the overall balance in the same way that the pruned out excesses are.
 
 However, a problem arises now in that as it stands, the UTXOs _cannot_ be pruned because we would lose some data needed
-to verify the kernel signatures, i.e. \\ \mathrm{H}(C_i \Vert s_i) \\) and that data only exists in the UTXOs. However,
+to verify the kernel signatures, i.e. \\( \mathrm{H}(C_i \Vert s_i) \\) and that data only exists in the UTXOs. However,
 we can salvage this situation fairly easily by noticing that we only need that _hash_ of the commitment and script hash.
-If we track an MMR of \\ C_i \Vert s_i \\), then those hashes are always available, even after the UTXOs themselves have
+If we track an MMR of \\( C_i \Vert s_i \\), then those hashes are always available, even after the UTXOs themselves have
 been discarded. In terms of additional block space required, this amounts to a single 32 byte hash per header (the MMR
 root).
 
@@ -269,9 +269,9 @@ The main properties of Tari script are
 * The scripting language is stack-based. At redeem time, the UTXO spender must supply an input stack. The script runs by
   operating on the the stack contents.
 * If an error occurs during execution, the script fails.
-* After the script completes, the script fails if the stack contains zero, or more than one element.
-* The script passes if and only if the script executes to completion with no errors, the resulting stack has a single
-  element, and the value of that element is equivalent to zero.
+* After the script completes, it is successful if and only if it has not aborted, and there is exactly a single element
+  on the stack with a value of zero. In other words, the script fails if the stack is empty, or contains more than one
+  element, or aborts early.
 * It is not Turing complete, so there are no loops or timing functions.
 * The Rust type system ensures that only compatible data types can be operated on, e.g. A public key cannot be added to
   an integer scalar. Errors of this kind cause the script to fail.
@@ -329,10 +329,22 @@ which maps to
 
 ```text
 71  b0           7a       ae2337ce44f9ebb6169c863ec168046cb35ab4ef7aa9ed4f5f1f669bb74b09e5  81          70   ac
-Dup HashBlake256 PushHash(ae2337ce44f9ebb6169c863ec168046cb35ab4ef7aa9ed4f5f1f669bb74b09e5) EqualVerify Drop CheckSig"
+Dup HashBlake256 PushHash(ae2337ce44f9ebb6169c863ec168046cb35ab4ef7aa9ed4f5f1f669bb74b09e5) EqualVerify Drop CheckSig
 ```
 
 Input parameters are serialised in an analogous manner.
+
+The types of input parameters that are accepted are:
+
+```rust,ignore
+pub enum StackItem {
+    Number(i64),
+    Hash(HashValue),
+    Commitment(PedersenCommitment),
+    PublicKey(RistrettoPublicKey),
+    Signature(RistrettoSchnorr),
+}
+```
 
 ## Extensions
 
@@ -371,9 +383,10 @@ Alice then locks the output with the following script:
 Dup PushPubkey(P_B) EqualVerify CheckSig Add
 ```
 
-where `P_B` is Bob's public key. Tari script is similar, but not identical to Bitcoin script, and has slightly different
-semantics. The interpretation of this script is, "Given a Public key, and a signature of this script, the public key
-must be equal to the one in the locking script, and the signature must be valid using the same public key".
+where `P_B` is Bob's public key. As one can see, this Tari script is very similar to Bitcoin script.  
+The interpretation of this script is, "Given a Public key, and a signature of this
+script, the public key must be equal to the one in the locking script, and the signature must be valid using the same
+public key".
 
 This is in effect the same as Bitcoin's P2PK script. To increase privacy, Alice could also lock the UTXO with a P2PKH
 script:
@@ -386,6 +399,10 @@ where `HB` is the hash of Bob's public key.
 
 In either case, only someone with the knowledge of Bob's private key can generate a valid signature, so Alice will not
 be able to unlock the UTXO to spend it.
+
+Since the script is committed to and cannot be cut-through, only Bob will be able to spend this UTXO unless someone is
+able to discover the private key from the public key information (the discrete log assumption), or if the majority of
+miners collude to not honour the consensus rules governing the successful evaluation of the script (the 51% assumption).
 
 [data commitments]: https://phyro.github.io/grinvestigation/data_commitments.html
 [LIP-004]: https://github.com/DavidBurkett/lips/blob/master/lip-0004.mediawiki
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -1,0 +1,393 @@
+# RFC-0201/TariScript
+
+## Tari Script
+
+![status: draft](theme/images/status-draft.svg)
+
+**Maintainer(s)**: [Cayle Sharrock](https://github.com/CjS77)
+
+# Licence
+
+[The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright 2020 The Tari Development Community
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
+
+1. Redistributions of this document must retain the above copyright notice, this list of conditions and the following
+   disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+   derived from this software without specific prior written permission.
+
+
+THIS DOCUMENT IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS", AND ANY EXPRESS OR IMPLIED WARRANTIES,
+INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+## Language
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT
+RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in
+[BCP 14](https://tools.ietf.org/html/bcp14) (covering RFC2119 and RFC8174) when, and only when, they appear in all
+capitals, as shown here.
+
+## Disclaimer
+
+This document and its content are intended for information purposes only and may be subject to change or update without
+notice.
+
+This document may include preliminary concepts that may or may not be in the process of being developed by the Tari
+community. The release of this document is intended solely for review and discussion by the community of the
+technological merits of the potential system outlined herein.
+
+## Goals
+
+This Request for Comment (RFC) presents a proposal for introducing Tari Script into the Tari base layer protocol. Tari
+Script aims to provide a general mechanism for enabling further extensions such as side chains, the DAN, one-sided
+payments and atomic swaps.
+
+## Related Requests for Comment
+
+* [RFC-0200: Base Layer Extensions](BaseLayerExtensions.md)
+* [RFC-0300: The Tari Digital Assets Network](RFC-0300_DAN.md)
+
+## Introduction
+
+It is hopefully clear to anyone reading these RFCs that the ambitions of the Tari project extend beyond a
+Mimblewimble-clone-coin.  
+It should also be fairly clear that basic Mimblewimble does not have the feature set to provide functionality such as:
+
+* One-sided payments
+* Atomic swaps (possible with scriptless scripts, but not easy)
+* Hash time-locked contracts (possible with scriptless scripts, but not easy)
+* Multiparty side-chain peg outs and peg-ins
+* Generalised smart contracts
+
+Extensions to Mimblewimble have been proposed for most of these features, for example, David Burkett's one-sided payment
+proposal for LiteCoin ([LIP-004]), this project's [HTLC RFC](RFC-0230_HTLC.md), the pegging proposals for the
+Clacks side-chain, and [Scriptless script]s.
+
+This RFC makes the case that if Tari were to implement a scripting language similar to Bitcoin script, then all of these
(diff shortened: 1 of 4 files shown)


RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+not concerned that every transaction in history was valid from an inflation perspective, but that the net effect of all
+transactions lead to zero inflation. This sounds worse than it is, since locally, every individual transaction is
+checked for validity at the time of inclusion in the blockchain.
+
+We presume however, that the intervening transactions were all valid, since they all passed consensus at the time of
+inclusion.
+
+This is why we can discard range proofs after UTXOs are spent. We presume that the UTXO chain that led to the current
+set were all checked and had positive values, but we only really care that the current UTXO set contains all
+non-negative values.
```
SWvheerden: This is technically not completely true., we never presume that the UTXO chain that led to the current
set were all checked and had positive valuesAltho currently in the our implementation in normal archival (not pruned) operation a node will only check the incoming utxo's rangeproofs. And assume that all others where checked by it self, which in effect it did by processing the chain block by block.  Pruned node's will only download the current utxo set. But they will check the entire chain state. 

Even is somehow , someone managed to sneak in a "inflated" utxo at height 1, and by height 100 validly spent that utxo. By height 1 Million, a node checking the utxo set, can still see that there was inflation somewhere in the chain. 

The reason we only "check" the current rangeproofs are because its only possible to inflate supply by either supplying a fake current rangeproof or by changing the total supply.  And total supply is fixed at height.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+
+This argument is independent of the nature of the additional restrictions. Specifically, if these restrictions are
+manifested as a script that provides additional constraints over whether a UTXO may be spent, the same arguments apply:
+We assume that the rules were properly enforced up until now, but it's not really our concern. What we care about is
```
SWvheerden: This is also not really true. Most if not all rules currently reside in the kernels. They are not pruned and can be checked forever. 

I think the only 2 rules thats not able to check for ever is: coinbase spent height, and utxo spent limit.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+
+The overall and block balance checks must also be modified to use \\( \hat{C} \\) rather than _C_.
+
+### Transaction balance
+
+The new transaction balance is thus
+
+$$  
+\begin{array}
+  .\sum(\mathrm{Inputs}) - \sum(\mathrm{Outputs}) + \sum(\mathrm{fee}_i.G)  \\\\
```
SWvheerden:
```suggestion
  .\sum(\mathrm{Inputs}) - \sum(\mathrm{Outputs}) - \sum(\mathrm{fee}_i.G)  \\\\
```

Inputs = outputs + fee

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+  \sum\hat{C_i} - \sum\hat{C_j} + \sum(\mathrm{fee}_i.G)  \\\\
+  \sum(C_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(C_j + \mathrm{H}(C_j \Vert s_j).G) + \sum(\mathrm{fee}.G)  \\\\
```
SWvheerden:
```suggestion
  \sum(C_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(C_j + \mathrm{H}(C_j \Vert s_j).G) - \sum(\mathrm{fee}.G)  \\\\
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+The overall and block balance checks must also be modified to use \\( \hat{C} \\) rather than _C_.
+
+### Transaction balance
+
+The new transaction balance is thus
+
+$$  
+\begin{array}
+  .\sum(\mathrm{Inputs}) - \sum(\mathrm{Outputs}) + \sum(\mathrm{fee}_i.G)  \\\\
+  \sum\hat{C_i} - \sum\hat{C_j} + \sum(\mathrm{fee}_i.G)  \\\\
```
SWvheerden:
```suggestion
  \sum\hat{C_i} - \sum\hat{C_j} - \sum(\mathrm{fee}_i.G)  \\\\
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+}
+```
+
+Now when calculating the transaction or block balance, we calculate a different set of commitments. The current commitment,
+
+$$ C = v.H + k.G $$
+
+is modified with a commitment to the script hash, so
+
+$$ \hat{C} = C + \mathrm{H}(C \Vert s).G $$
```
SWvheerden, line 204: This is not explicitly stated, but I take it s = script hash?

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+Thus all three requirements are satisfied and Tari Script, using this formulation should offer the same security
+guarantees that range proofs do.
+
+In particular, UTXOs can still be pruned because the \\( \Lambda \\) values change sign when used as inputs and will  
+cancel out in the overall balance in the same way that the pruned out excesses are.
+
+However, a problem arises now in that as it stands, the UTXOs _cannot_ be pruned because we would lose some data needed
+to verify the kernel signatures, i.e. \\( \mathrm{H}(C_i \Vert s_i) \\) and that data only exists in the UTXOs. However,
+we can salvage this situation fairly easily by noticing that we only need that _hash_ of the commitment and script hash.
+If we track an MMR of \\( C_i \Vert s_i \\), then those hashes are always available, even after the UTXOs themselves have
```
SWvheerden: Would it not be easier to commit to \\( C_i \Vert s_i \\) in the kernel it self than the header. 

if we add a mmr for \\( C_i \Vert s_i \\), data transfered by block would be:
+ 32 bytes for every utxo
+ 32 bytes for header. 

If we add it to the kernel it would be:
+ 32 bytes per tx. 
+ 32 bytes per utxo

So a block is a bit bigger. 

But storage is:
if we add a mmr for \\( C_i \Vert s_i \\), data stored is
 + 2^(k+1)-1  * 32 bytes where k = total number of utxos
If you just want to store the leaves its 
+ k* 32 bytres. 

If we add it to the kernel, storage is:
+ 32 bytes per kenel.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+further, that pruning spent outputs (and possibly the scripts associated with them) is no different from pruning range
+proofs.
+
+### Additional requirements
+
+The assumptions that broadly equate scripting with range proofs in the above argument are:
+
+* The script (hash) must be committed to in the blockchain.
+* The script must not be malleable in any way without invalidating the transaction.
+* The creator of the UTXO must commit to and sign the script (hash).
```
SWvheerden, line 153: Cant we just incorporate it in the challange?
That should ensure its not malleable, is signed and is commit in the blockchain. 
I made a case we I think its better to put this in the kernel

RFC/src/RFC-0201_TariScript.md
SWvheerden: I think we might get away to make this field optional as well. That way the extra data is only stored if its required

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+```
+
+where `HB` is the hash of Bob's public key.
+
+In either case, only someone with the knowledge of Bob's private key can generate a valid signature, so Alice will not
+be able to unlock the UTXO to spend it.
+
+Since the script is committed to and cannot be cut-through, only Bob will be able to spend this UTXO unless someone is
+able to discover the private key from the public key information (the discrete log assumption), or if the majority of
+miners collude to not honour the consensus rules governing the successful evaluation of the script (the 51% assumption).
```
SWvheerden, line 451: The only problem is you would need an archival node to prove this. 

I think to increase trust we might need to put some limit on how long the script + result is stored. 
Take for example we have a case where a node receives a string of 10 valid headers.
 Current MW states that f you receive that 10 headers + state, you can 100% verify that nothing bad happend there. If we include tari script we cannot say with 100% surety that  nothing bad happend. This is especially important if we want to have one sided tx's where we more rely on the security of the script. 
In bitcoin you dont have pruned node state so all nodes will have to provide you those 10 blocks. 

In practice I think our current nodes won't accept that and will just continue mining on the current chain. But for network integrity and locking this down,  I think we need to add a consensus to node that forces a limit on what the minimum pruning horizon is. 

On coins like grin, you can run with a pruning horizon of 1, and altho you will have to reorg a lot, you can alway be 100% sure that nothing bad happend. 
If you do that here with tari_script you cannot say that.  If we enforce a limit of say coinbase_lockheight blocks or so pruning_horizon you know with a fair amount of security that you will see that tx and its script. It is checked and  was available to be checked for a while. And we lock this as part of consensus if a node provides a fork of less than coinbase_height, and he cant provide the blocks so you can check the transction's scripts. You ignore that fork.

SWvheerden (commented):
Looks good.

RFC/src/RFC-0201_TariScript.md
CjS77: I didn't see your point until the comment below about holding onto scripts until the pruning horizon. 
Essentially, you're saying that you can still detect irregularities even if we don't check the individual range proofs because they'll manifest in the balance equation; whereas with script you don't have that double check feature.

The rest of the argument still stands though.. if you had two invalid txs somehow but the second one reversed the first one, you wouldn't detect that in the pruning balance.

RFC/src/RFC-0201_TariScript.md
CjS77: I was still really talking about range proofs, which are clearly not in the kernel.

RFC/src/RFC-0201_TariScript.md
CjS77: Thanks

RFC/src/RFC-0201_TariScript.md
CjS77: Storage is definitely higher, but that's not great, but ok.

The problem with committing the hash in the kernel is that you permanently link kernels to outputs, breaking privacy which is not great at all. I saw the extra storage as the price we pay to maintain privacy.

I realised that fields can't really be optional, because then miners could go and remove everything, and nodes would allow it because missing data is valid.

RFC/src/RFC-0201_TariScript.md
CjS77: Same issue. It's bad for privacy.

RFC/src/RFC-0201_TariScript.md
CjS77: It is L#164

RFC/src/RFC-0201_TariScript.md
CjS77: Yes, that's a valid point.

We could require keeping scripts until the pruning horizon, which is a few thousand blocks. So an attack is still possible if there's a very, very deep re-org.

The best way to block that attack is to re-spend coins to yourself with a default script (aka normal Tari tx) where possible. This is simple for one-sided transactions, but easier said than done, for say peg-ins.

CjS77: I'm also thinking that we don't, as a rule have scripts where there is no blinding factor, or where the blinding factor is known, but use DH shared secrets in nearly all cases. This removes the honeypot where miners can say "if only the script was gone, we could spend this coin"

RFC/src/RFC-0201_TariScript.md
SWvheerden: o cwl, I was under the impression this moved on to some of the spend rules of kernels like the mined height etc. Those are locked away and provable for ever.

RFC/src/RFC-0201_TariScript.md
SWvheerden: That can be locked up in the challenge removing the malleability.  But you're right, the privacy leak is big.

RFC/src/RFC-0201_TariScript.md
SWvheerden: Agreed.

RFC/src/RFC-0201_TariScript.md
SWvheerden: That would help. Its going to help make it harder to break. Its going to take some colluding to get the sender + miner in on it.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+This means that from a philosophical viewpoint, there ought to be no reason that Tari Script is not workable, and
+further, that pruning spent outputs (and possibly the scripts associated with them) is no different from pruning range
+proofs.
+
+### Additional requirements
+
+The assumptions that broadly equate scripting with range proofs in the above argument are:
+
+* The script (hash) must be committed to in the blockchain.
+* The script must not be malleable in any way without invalidating the transaction.
```
philipr-za, line 152: On a related issue we still have not solved this point for the simple mechanism of TransactionOutput Maturity. As far as I know it is not committed too and while it is enforced in the consensus rules it is malleable by a miner without invalidating the transaction.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+malicious actors could take the UTXO for themselves.
+
+Therefore, it's imperative that the UTXO creator sign the script.
+
+Further, this signature must be present in the _kernel_ in some form, otherwise miners will be able to remove the script
+via cut-through, whereas kernels are never pruned.
+
+One approach to commit to the script hashes is to modify the output commitments using the [data commitments] approach
+suggested by [Phyro](https://github.com/phyro). In this approach, when creating a new UTXO, the owner also calculates
+the hash of the locking script, _s_, such that `s = H(script)`. The hash script gets stored in the UTXO itself.
```
philipr-za:
```suggestion
the hash of the locking script, _s_, such that `s = H(script)`. The script hash gets stored in the UTXO itself.
```

RFC/src/RFC-0201_TariScript.md
philipr-za: If we include the Output Features in the hash this solves the Maturity malleability too.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+  \sum\hat{C_i} - \sum\hat{C_j} - \sum(\mathrm{fee}_i.G)  \\\\
+  \sum(C_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(C_j + \mathrm{H}(C_j \Vert s_j).G) - \sum(\mathrm{fee}.G)  \\\\
+  \text{If the accounting is correct, all values will cancel} \\\\
+  \sum(k_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(k_j + \mathrm{H}(C_j \Vert s_j).G)  \\\\
+  \text{The sum of all the blinding factors (times G) is the definition of the excess,}\\; x_s\cdot G = X_s \\\\
+  X_s + \sum(\mathrm{H}(C_i \Vert s_i).G) - \sum(\mathrm{H}(C_j \Vert s_j).G) \\\\
+  \text{Define}\\, \Lambda = \sum(\mathrm{H}(C_i \Vert s_i).G) - \sum(\mathrm{H}(C_j \Vert s_j).G) \\\\
+  X_s + \Lambda \\\\
+\end{array}  
+$$
```
philipr-za:
```suggestion
$$  
    \begin{eqnarray}
        \mathrm{TransactionSum} &=&\sum(\mathrm{Inputs}) - \sum(\mathrm{Outputs}) - \sum(\mathrm{fee}_i.G)  \\
        &=& \sum\hat{C_i} - \sum\hat{C_j} - \sum(\mathrm{fee}_i.G)  \\
        &=& \sum(C_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(C_j + \mathrm{H}(C_j \Vert s_j).G) - \sum(\mathrm{fee}.G)  \\
   \end{eqnarray}  
$$
  If the accounting is correct, all values will cancel,
$$
    \mathrm{TransactionSum} = \sum(k_i + \mathrm{H}(C_i \Vert s_i).G) - \sum(k_j + \mathrm{H}(C_j \Vert s_j).G)  
$$
The sum of all the blinding factors (times G) is the traditional Mimblewimble definition of the excess,
$$
    X_s =  x_s\cdot G
$$
If we define,
$$
    \Lambda = \sum(\mathrm{H}(C_i \Vert s_i).G) - \sum(\mathrm{H}(C_j \Vert s_j).G)
$$
then the new transaction excess can be written as
$$
    X_\mathrm{new} = X_s + \Lambda \\.
$$
```

philipr-za: Nit: Just an editorial suggestion to make the maths easier to read. An equation needs some equals strictly speaking.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+
+The kernels are unmodified, except that the excess will now include \\( \Lambda \\), representing the sum of all the
+commitments to the UTXO script hashes. This also means that the kernel signatures are calculated slightly differently:
+
+$$  
+\begin{array}
+  .s_i &= r_i + e.\bigl(k_i + \mathrm{H}(C_i \Vert s_i) \bigr) \\\\
+  s_i.G &= r_i.G + e.\bigl(k_i + \mathrm{H}(C_i \Vert s_i) \bigr).G \\\\
+  s_i.G &= R_i + e.\bigl(P_i + \mathrm{H}(C_i \Vert s_i)\bigr) \\\\
```
philipr-za:
```suggestion
  .s_i &= r_i + e.\bigl(k_i + \mathrm{H}(C_i \Vert s_i) \bigr) \\
  s_i.G &= r_i.G + e.\bigl(k_i + \mathrm{H}(C_i \Vert s_i) \bigr).G \\
  s_i.G &= R_i + e.\bigl(P_i + \mathrm{H}(C_i \Vert s_i)\bigr) \\
```

philipr-za: Nit: when you see this rendered in Mathjax there are big space in between the lines

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+Tari Script opcodes are enumerated from 0 to 255 and are represented as a single unsigned byte. The opcode set is
+initially limited to allow for the applications specified in this RFC, but can be expanded in future.
+
+```rust,ignore
+pub enum Opcode {
+    /// Push the current chain height onto the stack
+    PushHeight,
+    /// Push the associated 32-byte value onto the stack
+    PushHash(Box<HashValue>),
+    /// Hash to top stack element with the Blake256 hash function and push the result to the stack
```
philipr-za:
```suggestion
    /// Hash the top stack element with the Blake256 hash function and push the result to the stack
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,410 @@
...
+
+### One-sided transactions
+
+One-sided transactions are Mimblewimble payments that do not require the receiver to interact in the transaction
+process. [LIP-004] describes how this will be implemented in Litecoin's Mimblewimble implementation. The main thrust is
+that the sender uses Diffie-Hellman exchange to generate a shared private key that is used as the receiver's blinding
+factor.
+
+To prevent the sender from spending the coins (since both parties now know the spending key), there is an additional
+balance that is performed on the block and transaction that requires the spender to know the receiver's private key.
```
philipr-za:
```suggestion
check that is performed on the block and transaction that requires the spender to know the receiver's private key.
```

philipr-za (commented):
Looking good. I made a couple observations regarding a related problem we have yet to solve about the malleability of an output's Maturity which can also be solved using the data commitment described here. 

Pardon my editorial suggestions, I can't help it.

RFC/src/RFC-0201_TariScript.md
CjS77: I believe you're right. Nice touch!

RFC/src/RFC-0201_TariScript.md
CjS77: It's actually an additional balance equation that they add to the consensus rules.

RFC/src/RFC-0201_TariScript.md
CjS77: 🤔 Not on my machine. Without double escapes it renders everything on one line.

RFC/src/RFC-0201_TariScript.md
CjS77: Reworded this section to take this comment into account.

RFC/src/RFC-0201_TariScript.md
philipr-za: Oh, I was using MathJax in CodiMD so your render will reflect what will appear on the website

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,462 @@
...
+
+If it somehow happened that two illegal transactions made it into the blockchain (perhaps due to a bug), and the two
+cancelled each other out, such that the global coin supply was still correct, one would never know this when doing a
+chain synchronisation in pruned mode.
+
+But if there was a steady inflation bug due to invalid range proofs making it into the blockchain, a pruned mode sync
+would still detect that _something_ was awry, because the global coin supply balance acts as another check.
+
+With Tari script, once the script has been pruned away, and then there is a re-org to an earlier point on the chain,
+then there's no way way to ensure that the script was honoured.
```
stringhandler, line 142:
```suggestion
then there's no way to ensure that the script was honoured.
```

stringhandler: Is the case here:
1. UTXO is spent in original chain
2. Chain is re-org'ed, and UTXO is spent to a different output?

or 

1. UTXO is not spent in original chain
2. Chain is re-org'ed and spent in new chain

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,462 @@
...
+
+However, a single honest archival node would be able to detect any fraud on the same chain and provide a simple proof
```
stringhandler, line 144: I don't think it's possible to have an archival node on the same chain if the UTXO was spent to different commitments. If the pruned chain is falsely generated, no archival nodes will be on that chain

RFC/src/RFC-0201_TariScript.md
```diff
@@ -0,0 +1,462 @@
...
+that a transaction did not honour the redeem script.
+
+### Additional requirements
+
+The assumptions that broadly equate scripting with range proofs in the above argument are:
+
+* The script (hash) must be committed to in the blockchain.
+* The script must not be malleable in any way without invalidating the transaction.
+* The creator of the UTXO must commit to and sign the script (hash).
```
stringhandler, line 153: What key do they use to sign the script? The blinding factor is not a good candidate here, as it may be known to multiple parties if the true spending criteria is solely in the script. For example, say there are 3 of 3 signatures needed to spend an output, all three would know the blinding factor and a single party could replace the script and sign it.

SWvheerden (approved):
LGTM

stringhandler (approved):
Looks good. I am worried that there are some edge cases that are not obvious and that distributing the blinding factor of the UTXO when there are multiple parties involve could be difficult. I am also not 100% sure about the balance equation working out, but I can work through that on my own and see if there's a case where we need more data. 

It's still an RFC here so happy to merge it

RFC/src/RFC-0201_TariScript.md
CjS77: Possibly either. Where it's beyond a standard double-spend attack, it's typically in say, one-sided payments where the spender knows the blinding factor, and then uses the fact that the script won't be enforced to spend the output somewhere else. If nodes had the script, that spend would be blocked.

RFC/src/RFC-0201_TariScript.md
CjS77: True, what I'm trying to say here is that the archival node could demonstrate to pruned nodes that they're on the wrong chain.

RFC/src/RFC-0201_TariScript.md
CjS77: The signature is discussed in detail below. LMK if it's not covering your concern.

Issue #2092, opened 2020-07-27
Proposal: RPC protocol for tari_comms


sdbondi:
This proposal is (tick applicable):

* [x] A new feature
* [ ] An extension to an existing RFC/proposal <LINK>
* [ ] An alternative approach to an existing RFC or proposal <LINK>

### Change log

| Date | Changes |
|:-----|:--------|
| 28/07/2020     | Initial        |

### Goals
 
_Fire-and-forget_ messaging is used throughout the current system. While some areas of the system suit this approach perfectly (transaction/block/DHT propagation), it falls short for many current use-cases.

This proposal aims to address the shortcomings of the current messaging approach, namely:
- excessive complexity (which causes bugs and code that is hard to maintain)
- lack of separation of concerns (nodes have to worry about messaging semantics)
- performance is difficult (the current protocol lends itself to implementers constructing large messages and makes _streaming_ many small messages difficult)
- visibility is difficult (not knowing if a message has actually been received. Knowing if a peer is actually connected requires holding a handle and querying connectivity manager)
- Writing protocols (conversations) is difficult. Any type of message can arrive at any time. This means that the domain layer has to implement a mapping between requests and responses. Additionally, defending against unexpected, potentially malicious messages is something that has to be actively reasoned about for every incoming message type (Did I request this message? Did this message come from the peer I expected it to?). 
- Unbounded tasks are often spawned per message, which can be an attack vector
- Testing is difficult. Outbound and inbound messaging is difficult to stub out.

### Assumptions

- A reliable transport such as TCP is used
- Many devs are familiar with gRPC. 
- Implementing a RPC interface similar to tonic (except p2p) using the current comms infrastructure will not be prohibitively difficult.

### Abstract

The proposes a more bounded and structured approach to protocols on the Tari p2p network heavily influenced by tonic RPC.

What the proposed RPC protocol does do:
- mapping requests to responses
- managing timeouts and disconnects
- message framing
- Rate limiting
- consistent `RpcError` api
- (de)serialization of requests/responses (protobuf)
- stream requests and/or responses
- async service and client trait (async_trait crate)

What an RPC protocol does not do / limitations:
- Peer discovery, store and forward, message propagation/broadcasting (use the DHT/outbound messaging handles)
- Reconnecting disconnected peers (Use connectivity manager)
- Resending messages or any "retry behaviour" (the request either works or doesnt)
- Requests without responses

Future extensions/nice to haves 
[...]

PR #2121, opened 2020-08-12
RPC protocol for tari comms

sdbondi:
## Description

This PR is the first iteration on the tari comms RPC protocol. This protocol is
WIP and not currently in use in any part of the system.

Details here: https://github.com/tari-project/tari/issues/2092

- The tari comms connections are already multiplexed (like HTTP2)
  and so RPC itself carries a small footprint of a few bytes (<~ 10 bytes) per request.
- Server-side streaming supported. Client-side streaming left as a TODO as it's generally less common.
- Statically-typed service routing (tonic-RPC knock off 🙏 )
- Different from tonic however, the router is used to select the matching service factory (impl MakeService) which in turn is used to produce a service that is handed over to the executing task. 
   This service factory can be generated by the proc macro and so is transparent to the API user. 
- RPC protocol version is negotiated once ahead of time, so the version
  does not need to be sent with every message.

Few notes on concurrency:
- All service impls must be Send + 'static. 
- A SINGLE request/response can be performed per peer per protocol at a time. You may clone the client and send multiple 
  requests at once, however they will not be furnished until the previous client requests have completed. Apart from greatly 
  simplifying the code, this also reduces the potential to boundlessly exploit resources. It is expected that each client 
  "session" will only be making one request at a time anyway. If a client peer wishes to make concurrent requests to a 
  service, they are free to open more than one channel at a time.
- None of the RPC service traits may take mutable references to self (i.e. if mutation is required, some method of interior mutability must be used - that satisfies Send).

In progress:
- [ ] Integration with comms (establish client session, register services, peer info etc)
- [ ] Implement proc macro to generate domain-specific client/server code

## Motivation and Context

See: https://github.com/tari-project/tari/issues/2092

Example usage (with possible proc macro API): 
```rust
#[tari_rpc(module = generated, protocol_name = "/tari/greeting/1.0")]
pub trait GreetingRpc: Send + Sync + 'static {
    #[rpc(method = 1)]
    async fn say_hello(&self, request: Request<SayHelloRequest>) -> Result<Response<SayHelloResponse>, RpcStatus>;
    #[rpc(method = 2)]
    async fn return_error(&self, request: Request<()>) -> Result<Response<()>, RpcStatus>;
    #[rpc(streaming, method = 3)]
    async fn get_greetings(&self, request: Request<u32>) -> Result<Streaming<String>, RpcStatus>;
    #[rpc(streaming, method = 4)]
    async fn streaming_error(&self, request: Request<String>) -> Result<Streaming<String>, RpcStatus>;
    #[rpc(streaming, method = 5)]
    async fn streaming_error2(&self, _: Request<()>) -> Result<Streaming<String>, RpcStatus>;
}

pub struct GreetingService;

#[async_trait]
impl GreetingRpc for GreetingService {
    async fn say_hello(&self, _: Request<SayHelloRequest>) -> Result<Response<SayHelloResponse>, RpcStatus> {
        Ok(Response::new(SayHelloResponse {
            greeting: "Yo".to_string(),
        }))
    }

    async fn return_error(&self, _: Request<()>) -> Result<Response<()>, RpcStatus> {
        unimplemented!()
    }
// (...)
}

async fn client() {
  let mut client = generated::client::GreetingServiceClient::builder(framed)
        .with_deadline(Duration::from_secs(5))
        .connect()
        .await
        .unwrap();

    let resp = client
        .say_hello(SayHelloRequest {
            name: "Yathvan".to_string(),
            language: 1,
        })
        .await
        .unwrap();
}
```

## How Has This Been Tested?

Relatively comprehensive unit and smoke testing.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/src/proto/rpc.proto
+++ b/comms/src/proto/rpc.proto
@@ -1,0 +1,49 @@
+syntax = "proto3";
+
+package tari.comms.rpc;
+
+// Message type for all RPC requests
+message RpcRequest {
+    // An identifier that is unique per request per session. This value is not strictly needed as
+    // requests and responses alternate on the protocol level without the need for storing and matching
+    // request IDs. However, this value is useful for logging.
+    uint32 request_id = 1;
+    // The method identifier. The matching method for a given value is defined by each service.
+    uint32 method = 2;
+    // Message flags. Currently this is not used for requests.
+    uint32 flags = 3;
+    // The length of time in seconds that a client is willing to wait for a response
+    uint64 deadline = 4;
+
+    // The message payload
+    bytes message = 10;
+}
+
+// Message type for all RPC responses
+message RpcResponse {
+    // The request ID of a prior request.
+    uint32 request_id = 1;
+    // The status of the response. A non-zero status indicates an error.
+    uint32 status = 2;
+    // Message flags. Currently only used to indicate if a stream of messages has completed.
+    uint32 flags = 3;
+
+    // The message payload. If the status is non-zero, this contains additional error details.
+    bytes message = 10;
+}
+
+// Message sent by the client when negotiating an RPC session. A server may close the substream if it does
+// not agree with the session parameters.
+message RpcSession {
+    // The RPC versions supported by the client
+    repeated uint32 supported_versions = 1;
+}
+
+message RpcSessionReply {
+    oneof session_result {
+        // The RPC version selected by the server
+        uint32 accepted_version = 1;
+        // Indicates the server rejected the session
+        bool rejected = 2;
+    }
+}
--- a/comms/src/protocol/rpc/client.rs
+++ b/comms/src/protocol/rpc/client.rs
@@ -1,0 +1,459 @@
+//  Copyright 2020, The Tari Project
+//
+//  Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+//  following conditions are met:
+//
+//  1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+//  disclaimer.
+//
+//  2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
+//  following disclaimer in the documentation and/or other materials provided with the distribution.
+//
+//  3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
+//  products derived from this software without specific prior written permission.
+//
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+//  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+//  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use super::message::MethodId;
+use crate::{
+    framing::CanonicalFraming,
+    message::MessageExt,
+    proto,
+    protocol::rpc::{body::ClientStreaming, Request, Response, RpcError, RpcStatus},
+    runtime::task,
+};
+use bytes::Bytes;
+use futures::{
+    channel::{mpsc, oneshot},
+    future::Either,
+    task::{Context, Poll},
+    AsyncRead,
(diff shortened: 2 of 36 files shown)


comms/src/protocol/rpc/client.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    /// Negotiates and establishes a session to the peer's RPC service using the given substream
+    pub async fn connect(self) -> Result<TClient, RpcError> {
+        RpcClient::connect(self.config, self.framed).await.map(Into::into)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct RpcClientConfig {
+    pub deadline: Option<Duration>,
```
philipr-za, line 188: So being pedantic here but isn't this a timeout. I always thought a Deadline referred to an absolute time that a task should be done by rather than a relative duration in which it should be completed i.e. a deadline = finish by 15:45:12.123 as opposed to finished before 10 seconds passes from now.

comms/src/protocol/rpc/client.rs
```diff
@@ -0,0 +1,459 @@
...
+            flags: 0,
+            message: request.message.to_vec(),
+        };
+
+        debug!(target: LOG_TARGET, "Sending request: {}", req);
+
+        let start = Instant::now();
+        self.framed.send(req.to_encoded_bytes().into()).await?;
+
+        let (mut response_tx, response_rx) = mpsc::channel(10);
```
philipr-za, line 363: is the buffer size of 10 just for safety?

comms/src/protocol/rpc/server.rs
```diff
@@ -0,0 +1,431 @@
...
+
+struct ActivePeerRpcService<TSvc, TSubstream> {
+    config: RpcServer,
+    node_id: NodeId,
+    service: TSvc,
+    framed: Option<CanonicalFraming<TSubstream>>,
+    shutdown_signal: OptionalShutdownSignal,
+}
+
+impl<TSvc, TSubstream> ActivePeerRpcService<TSvc, TSubstream>
```
philipr-za, line 301: Nit: This guy needs a `fn new(...)` I think

comms/src/protocol/rpc/test.rs
```diff
@@ -0,0 +1,540 @@
...
+    );
+    (notif_tx, server_hnd, shutdown)
+}
+
+async fn setup<T: GreetingRpc>(service: T) -> (MemorySocket, task::JoinHandle<Result<(), RpcError>>, Shutdown) {
+    let (mut notif_tx, server_hnd, shutdown) = setup_service(service).await;
+    let (inbound, outbound) = MemorySocket::new_pair();
+    let node_identity = build_node_identity(Default::default());
+
+    // Notify that a peer wants to speak the greeting RPC protocol
```
philipr-za, line 78: Looking forward to see how this is plumbed into the current inbound pipeline. This will come after Yamux?

philipr-za (dismissed):
Looking good, pretty crazy bit of work already though can see there is still much to do.

comms/src/protocol/rpc/client.rs
sdbondi: Ye it allows some buffering to occur which may speed things up - say if the consumer is slow to read, the buffer allows a cushion for messages to be received without creating immediate backpressure on the substream, meaning the data is at the receiving end and can be processed without waiting for the sender to send more.

comms/src/protocol/rpc/client.rs
sdbondi: Maybe - using RPC terminology here (deadline is a RPC thing). Deadline refers to the time that a server has to complete its work before it should expect that the client has given up and isn't listening anymore. Relative or absolute, I think deadline is fine.

comms/src/protocol/rpc/test.rs
sdbondi: Correct (although there is no "inbound pipeline" here), this is why the RPC components are "socket-agnostic". In this test we use a memory socket, but in the real world, the PeerConnection will negotiate and provide a substream. 

The basic idea is that you somehow get hold of a `PeerConnection` (i.e a handle to the underlying connection to a peer - there are a few ways that one would do that depending on the use case) and you would call something like `peer_connection.connect_rpc_client::<MySuperClient>().await`  or `MySuperClient::connect(&mut peer_connection)`.

comms/src/protocol/rpc/server.rs
sdbondi: Maybe - it's private and basically a glorified internal function, so I'd say no.

comms/src/proto/rpc.proto
```diff
@@ -0,0 +1,49 @@
...
+    // An identifier that is unique per request per session. This value is not strictly needed as
+    // requests and responses alternate on the protocol level without the need for storing and matching
+    // request IDs. However, this value is useful for logging.
+    uint32 request_id = 1;
+    // The method identifier. The matching method for a given value is defined by each service.
+    uint32 method = 2;
+    // Message flags. Currently this is not used for requests.
+    uint32 flags = 3;
+    // The length of time in seconds that a client is willing to wait for a response
+    uint64 deadline = 4;
```
stringhandler, line 16: Does this go over the wire? I don't think that there will be many cases where a server will react differently with this parameter. It will generally respond as soon as it can

stringhandler (approved):
Really slick. Didn't run locally, but code looks solid. The true test will be if it's easy to use, but looks like it will be. This doesn't affect any existing code from what I can see, so I think we can merge it

comms/src/proto/rpc.proto
sdbondi: Ye it's over the wire - the client tells the server how long they're willing to wait. If the "processing part" (i.e. the part that a dev implements) completes before the duration no problem. Otherwise, the server should give up because the client is not going to care about the response anyway. Obviously there is a guard where if the client sets this below some minimum, the server immediately rejects the request (to prevent the client setting some silly deadline like 10ms) - a deadline would typically be 30 seconds.

sdbondi: In reality client waits a little longer than the deadline (deadline + grace period) to account for latency (we wouldn't want a server to send a response just within the deadline, but the client has gone away just before the response can be received)

PR #2125, opened 2020-08-13
Remove debug tests on Circle CI

stringhandler:
Removed debug tests from Circle CI checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,11 +24,11 @@ commands:
           name: Build the project
           command: cargo build --all --all-features --jobs=3 <<#parameters.release>>--release<</parameters.release>>
       - run:
-          name: Cargo fmt (ignored in release mode)
+          name: Cargo fmt
           command: |
             TOOLCHAIN=$(cat rust-toolchain)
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
LGTM - `debug_assertions` could be misused/change behaviour in some unforeseen way so maybe this is not ideal. Did you try adding `lib: true` to the test step in debug? This may skip the doctests (educated guess based on `cargo test --lib`) which I think is the cause of the failing builds. Otherwise, I think that this can go in as a stop gap.

.circleci/config.yml
```diff
@@ -55,15 +55,6 @@ jobs:
...
           paths: book
 
-  test-tari-debug:
-    docker:
-      - image: *rust_image
-    resource_class: medium
-    steps:
-      - checkout
-      - test:
-          release: false
```
sdbondi, line 65: Would adding `lib: true` skip the docs tests?

stringhandler:
I think that debug tests may be worth doing infrequently, but doing both on every PR seems excessive anyway. It can be added back at a later stage

PR #2127, opened 2020-08-14
Simplified messaging protocol, fix for excessive delays

sdbondi:
## Description

This PR simplifies the outbound messaging and allows it to use
`StreamExt::forward` instead of a custom `Stream` impl.

This simplification means that the MessageSent event is no longer
sent and that in some rare cases a message may be "marked" as
sent even if the peer will not receive the message (e.g. sudden disconnect)

It should be acknowleged that previously this was the case anyway,
and "successfully sent" is more accurately "successfully buffered".

- Removed MessageSent event
- Removed retries from protocol
- Outbound messaging will retry once to connect and establish a
  substream, this is to handle the case where a pending dial is cancelled
  due to tie breaking - so one and only one retry is required.
- Retries happen per peer and not per message.
- Removed peer manager from messaging completely which could/would slightly delay messaging.
- A fix for a large outbound delay that can experienced in high contention situations.
- Ensure that sinks are closed if the workers exit for any reason.
- Added messaging stress test

## Motivation and Context

This PR acknowledges that guaranteed deliverability is neither easy nor desirable (RE tradeoffs) 
in a p2p system. The direct fire and forget messaging protocol was always "best effort"
and never provided guarantees around deliverability - and with those limitations, the benefits 
(simplification, performance, bug fixed by removing code) can be obtained. That being said
in the vast majority of cases with a stable connection, the message will arrive (reliability matches that of TCP/Tor)

The upcoming RPC protocol will provide the "either it worked or it didn't" semantic which has 
been sorely missing.

## How Has This Been Tested?

Tests updated as necessary. Every test that uses comms (i.e. the majority of the current test suite) is also testing this
component. Ran a base node for a few hours.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/outbound/serialize.rs
+++ b/comms/dht/src/outbound/serialize.rs
@@ -95,6 +95,12 @@ where S: Service<OutboundMessage, Response = (), Error = PipelineError> + Clone 
 
             let body = Bytes::from(envelope.to_encoded_bytes());
 
+            trace!(
+                target: LOG_TARGET,
+                "Serialized outbound message {} for peer `{}`. Passing onto next service",
+                tag,
+                destination_node_id.short_str()
+            );
             next_service
                 .oneshot(OutboundMessage {
                     tag,
--- a/applications/tari_base_node/src/parser.rs
+++ b/applications/tari_base_node/src/parser.rs
@@ -981,6 +981,7 @@ impl Parser {
         self.executor.spawn(async move {
             let start = Instant::now();
             println!("🌎 Peer discovery started.");
+
             match dht
                 .discover_peer(dest_pubkey.clone(), NodeDestination::PublicKey(dest_pubkey))
                 .await
--- a/base_layer/core/tests/node_service.rs
+++ b/base_layer/core/tests/node_service.rs
@@ -530,7 +530,6 @@ fn propagate_and_forward_invalid_block_hash() {
     // Create unknown block hash
     block1.header.height = 0;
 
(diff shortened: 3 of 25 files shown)


hansieodendaal:
Nice one

comms/dht/src/outbound/serialize.rs
```diff
@@ -95,6 +95,11 @@ where S: Service<OutboundMessage, Response = (), Error = PipelineError> + Clone
 
             let body = Bytes::from(envelope.to_encoded_bytes());
 
+            trace!(
+                target: LOG_TARGET,
+                "Serialized outbound message {:?}. Passing onto next service",
```
hansieodendaal, line 100: `"Serialized outbound message {:?} for peer `{}`. Passing onto next service",`

hansieodendaal:
```suggestion
                "Serialized outbound message {:?} for peer `{}`. Passing onto next service",
```

comms/dht/src/outbound/serialize.rs
```diff
@@ -95,6 +95,11 @@ where S: Service<OutboundMessage, Response = (), Error = PipelineError> + Clone
...
+                message.tag
```
hansieodendaal:
```suggestion
                message.tag,
                destination_node_id.short_str()
```

hansieodendaal (commented):
Nice, looks promising, will use this version to do some Tx level stress testing.

stringhandler (approved):

Issue #2122, opened 2020-08-12
Transaction broadcast to mempool protocol loop triggered unnecessarily


hansieodendaal:
**Describe the bug**

The `wallet::transaction_service::protocols::broadcast_protocol` broadcasts a Tx to the mempool more than once without waiting the prescribed time-out period, and also fails to detect a `MempoolServiceResponse TxStorage(UnconfirmedPool)` response. This happened in ~827 of ~1287 cases in a Tx stress test,

For both cases below, the broadcast protocol time-outs was set to 3600s:
```
  136564: 2020-08-12 09:50:56.688705400 [wallet::transaction_service::config] TRACE Timeouts - Base node monitoring: 3600s, Direct send: 3600s, Broadcast send: 3600s
```

Example 1: The consecutive broadcast took place ~11.3s later.
```
  137461: 2020-08-12 09:51:28.363111100 [wallet::transaction_service::protocols::broadcast_protocol] INFO  Attempting to Broadcast Transaction (TxId: 9676091740453170075 and Kernel Signature: 6e5db94c34704dfd71fd0fd71113896c9e6996e6ad185be5efb97cc1ee19870d) to Mempool
  141491: 2020-08-12 09:51:30.136417700 [wallet::output_manager_service] TRACE Could not publish Base Node Response, no subscribers to receive. (Err SendError(BaseNodeServiceResponse { request_key: 9676091740453170075, response: Some(TransactionOutputs(TransactionOutputs { outputs: [] })) }))
  149427: 2020-08-12 09:51:37.877918100 [wallet::transaction_service::service] TRACE Received Mempool Response: MempoolServiceResponse { request_key: 9676091740453170075, response: TxStorage(UnconfirmedPool) }
  151300: 2020-08-12 09:51:39.406070000 [wallet::transaction_service::protocols::broadcast_protocol] INFO  Mempool broadcast timed out for Transaction with TX_ID: 9676091740453170075
  151301: 2020-08-12 09:51:39.406070000 [wallet::transaction_service::protocols::broadcast_protocol] TRACE Error sending event, usually because there are no subscribers: SendError(MempoolBroadcastTimedOut(9676091740453170075))
[...]

PR #2129, opened 2020-08-14
Improved trace logging for the transaction broadcast protocol, some logic improvement

hansieodendaal:
## Description
Added more trace logs and improved logic of what is logged when for the transaction broadcast protocol, to enable improved system-level debugging and fault finding. Related to #2122.

## Motivation and Context
Illogical results were seen in the log files, where Tx broadcast to the mempool would happen again after it was detected in the mempool.

## How Has This Been Tested?
Tested in a Windows base node with Tx stress testing

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
@@ -191,11 +191,21 @@ where TBackend: TransactionBackend + Clone + 'static
             let mut delay = delay_for(self.timeout).fuse();
             futures::select! {
                 mempool_response = mempool_response_receiver.select_next_some() => {
+                    trace!(
+                        target: LOG_TARGET,
+                        "Transaction monitoring event 'mempool_response' triggered ({:?})",
+                        mempool_response,
+                    );
                     if self.handle_mempool_response(mempool_response).await? {
                         break;
                     }
                 },
                 base_node_response = base_node_response_receiver.select_next_some() => {
+                    trace!(
+                        target: LOG_TARGET,
+                        "Transaction monitoring event 'base_node_response' triggered ({:?})",
+                        base_node_response,
+                    );
                     if self.handle_base_node_response(base_node_response).await? {
                         break;
                     }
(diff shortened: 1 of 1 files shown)


philipr-za (approved):
LGTM

base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
```diff
@@ -205,31 +215,36 @@ where TBackend: TransactionBackend + Clone + 'static
                         self.timeout = to;
                         info!(
                             target: LOG_TARGET,
-                            "Broadcast monitoring protocol (Id: {}) timeout updated to {:?}", self.id ,self.timeout
+                            "Transaction monitoring event 'updated_timeout' triggered (Id: {}), timeout updated to {:?}", self.id ,self.timeout
+                        );
+                    } else {
+                        trace!(
```
stringhandler, line 221: Is this not an error? Perhaps add a warning here?

base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
```diff
@@ -205,31 +215,36 @@ where TBackend: TransactionBackend + Clone + 'static
...
+                            self.id,
+                            updated_timeout,
                         );
                     }
                 },
                 () = delay => {
-                },
-            }
-
-            info!(
```
stringhandler, line 216: This also feels like it's an important event. Perhaps even should be a `warn!`

base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
philipr-za: Na, it occurs if somehow this worker has lagged behind the pubsub buffer which can happen if there is a lot going on and the Send process above takes a while.

base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
philipr-za: This was just moved to only log if it was a timeout where as before anything that exited the select! would fire this log. Don't think this is a warn because it is not a show stopper and base nodes can be busy so we try again?

hansieodendaal: > This also feels like it's an important event. Perhaps even should be a `warn!`

It is only informative for trace level debugging, as the broadcast-to-mempool-and-wait loop will be repeated.

stringhandler (approved):

PR #2133, opened 2020-08-14
Add support for optional rolling log files to FFI wallet library

### LINKED ISSUE #2124: Add support to wallet_ffi to having rolling log files
Add the ability to set whether the Wallet_ffi logging uses rolling log files based on size.

Potential options: 
1. Size of files
2. Number of files in fixed window or else no fixed window

philipr-za:
## Description
Currently LibWallet accepts a single file path for logging and write all logs to that file. This PR adds the ability to optionally specify whether you want the logs to use multiple rolling files and at what size of file to trigger the roll over. If non zero values are provided for both `num_rolling_log_files` and `size_per_log_file_bytes` then the fixed window rolling functionality of log4rs will be used. If this is not desired 0 values should be passed for these arguments.

The rolled over archived files will be numbered version of the original filename. For example if the logging file path that is provided is `/User/me/wallet/log_170820_14382455.log`, 2 roll over files are specified and the size threshold is 10_000 bytes. 
1. The most current log file will always be `log_170820_14382455.log` 
2. When that reaches 10_000 bytes in size it will be renamed to `log_170820_14382455.0.log` and logging will continue in `log_170820_14382455.log`.
3. When the next roll over limit is reached: 
    1. `log_170820_14382455.0.log` will be renamed to `log_170820_14382455.1.log` 
    2. the current `log_170820_14382455.log` will be renamed to `log_170820_14382455.0.log`
    3. logging will continue in `log_170820_14382455.log`.
4. When the next roll over occurs because we specified only 2 roll over files the oldest one (index 1) will be deleted, 0 will be renamed to 1 and the current file renamed to index 0.

This can be used to limit the amount of logs LibWallet generates per session.

@kukabi @Jasonvdb for changes to `wallet_create(...)`

## Motivation and Context
Closes https://github.com/tari-project/tari/issues/2124

## How Has This Been Tested?
Both old and new logging is now used in LibWallet tests to a temporary directory to confirm they are working.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -195,6 +195,13 @@ use tari_wallet::{
     wallet::WalletConfig,
 };
 
+use log4rs::append::{
+    rolling_file::{
+        policy::compound::{roll::fixed_window::FixedWindowRoller, trigger::size::SizeTrigger, CompoundPolicy},
+        RollingFileAppender,
+    },
+    Append,
+};
 use tari_core::consensus::Network;
 use tokio::runtime::Runtime;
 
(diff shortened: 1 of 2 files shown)


StriderDM (approved):
LGTM

StriderDM:
Note: This is a breaking change.

kukabi:
@philipr-za Log files reach 10_000 bytes quite quick, do you think it's better if we had 10 megs per file? So the total is 30 megs max. Or did you mean 10_000_000 bytes?

philipr-za:
> @philipr-za Log files reach 10_000 bytes quite quick, do you think it's better if we had 10 megs per file? So the total is 30 megs max. Or did you mean 10_000_000 bytes?

It was just a number given as an example. You will provide the size, in bytes, when you call `wallet_create(...)`

kukabi:
ah ok 🤦  thanks: )

SWvheerden (approved):

stringhandler (approved):
Untested, but looks good

PR #2138, opened 2020-08-17
Added support for OS-assigned proxied listening address (tor)

sdbondi:
## Description

The tor proxy is configured to proxy a hidden service to a local
address. Previously, that local address would have to be configured to
some non-zero value so that comms could listen on it. This PR allows the
local port to be assigned by the OS (i.e. port `0`). The comms listener will listen on
an OS-assigned port and the hidden service will be automatically configured to forward
traffic to the local address. This does not effect the onion address port, typically
this would be set to a hardcoded value (e.g. 18100).

- FFI updated (no public (or other) breaking changes)
- Tari base node configuration now uses an OS-assigned port by default

## Motivation and Context

In the vast majority of use-cases, the tari_comms api user does not care about the proxied listener port and just wants some appropriate open port to be selected.

## How Has This Been Tested?

Existing tests pass. Tested on base nodes. 
Untested on mobile but this can be tested by checking for app crashes on startup.

To test a base node:
1. Remove `tor_forward_address` setting from the config if present (so that the default is used) 
1. Start up a wallet/base node
1. Find the listening port in the logs e.g `INFO  Listening for peer connections on '/ip4/127.0.0.1/tcp/63746'`
1. Check that `list-connections` has inbound connections
1. Restart the node and check that a different listening port is used as per step 3 and we have inbound connections

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -914,8 +914,7 @@ fn setup_wallet_transport_type(config: &GlobalConfig) -> TransportType {
                     .unwrap()
             );
 
-            let mut forward_addr = multiaddr_to_socketaddr(&forward_address).expect("Invalid tor forward address");
-            forward_addr.set_port(forward_addr.port() + 1);
+            let forward_addr = multiaddr_to_socketaddr(&forward_address).expect("Invalid tor forward address");
             TransportType::Tor(TorConfig {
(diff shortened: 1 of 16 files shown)


stringhandler (approved):
Untested, but looks good

PR #2143, opened 2020-08-18
Fix coinbase vulnerability

SWvheerden:
## Description

This introduces a fix for a coinbase vulnerability by ensure that the coinbase is flagged correctly and the amount is exactly as required. 
Refactor block and aggregate body checks

## Motivation and Context

Currently the coinbase is only checked on the amount added to the block. So in essence this is checked by checking that a block emits the correct emission reward.  But its possible to manipulate the block to make the coinbase transaction very small, and add a second "coinbase" that does not have the lock heighjt. This means the block aoumt is till correct and the chain has no inflation. But the coinbase is incorrect. 
This PR forces the coinbase to be the exact amount as it should be. It also requires that coinbase transactions not have an offset in their blinding factors. Out current code also does not add this. The use of the offset is only their so that you cannot link a kernel to a output. But in the case of the coinbase, both contain flags and the consensus code check that there is only 1 utxo and 1 kernel flagged as coinbase. So there is no hiding the tx kernel here.  

The refactor of the block checks to aggregated body has been done as those test only run on the aggregated body it self. This refactor has the benefit of allowing transactions to access those checks as well as make the checks easier to test. 

WARNING
This test has the potential to break consensus if there was any shenanigans in the chain in the past. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -19,7 +19,6 @@
 // SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
 use crate::transactions::{
     fee::Fee,
     tari_amount::*,
@@ -79,6 +78,11 @@ impl AggregateBody {
     /// Provide read-only access to the output list
     pub fn outputs(&self) -> &Vec<TransactionOutput> {
         &self.outputs
+    }
+
+    /// Should be used for tests only. Get a mutable reference to the outputs
+    pub fn outputs_mut(&mut self) -> &mut Vec<TransactionOutput> {
+        &mut self.outputs
     }
 
     /// Provide read-only access to the kernel list
@@ -191,6 +195,90 @@ impl AggregateBody {
             fee += kernel.fee;
         }
         fee
+    }
+
+    /// This function will check spent kernel rules like tx lock height etc
+    pub fn check_kernel_rules(&self, height: u64) -> Result<(), TransactionError> {
+        for kernel in self.kernels() {
+            if kernel.lock_height > height {
+                warn!(target: LOG_TARGET, "Kernel lock height was not reached: {}", kernel);
+                return Err(TransactionError::InvalidKernel);
+            }
+        }
+        Ok(())
+    }
+
+    /// Run through the outputs of the block and check that
+    /// 1. There is exactly ONE coinbase output
+    /// 1. The output's maturity is correctly set
+    /// 1. The amount is correct.
+    pub fn check_coinbase_output(
+        &self,
+        reward: MicroTari,
+        coinbase_lock_height: u64,
+        factories: &CryptoFactories,
+        height: u64,
+    ) -> Result<(), TransactionError>
+    {
+        let mut coinbase_utxo = None;
+        let mut coinbase_kernel = None;
+        let mut coinbase_counter = 0; // there should be exactly 1 coinbase
+        for utxo in self.outputs() {
+            if utxo.features.flags.contains(OutputFlags::COINBASE_OUTPUT) {
+                coinbase_counter += 1;
+                if utxo.features.maturity < (height + coinbase_lock_height) {
+                    warn!(target: LOG_TARGET, "Coinbase {} found with maturity set too low", utxo);
+                    return Err(TransactionError::InvalidCoinbaseMaturity);
+                }
+                coinbase_utxo = Some(utxo.clone());
+            }
+        }
+        if coinbase_counter != 1 {
+            warn!(
+                target: LOG_TARGET,
+                "{} coinbases found in body. Only a single coinbase is permitted.", coinbase_counter,
+            );
+            return Err(TransactionError::InvalidCoinbaseCount);
+        }
+
+        let mut coinbase_counter = 0; // there should be exactly 1 coinbase kernel as well
+        for kernel in self.kernels() {
+            if kernel.features.contains(KernelFeatures::COINBASE_KERNEL) {
+                coinbase_counter += 1;
+                coinbase_kernel = Some(kernel.clone());
+            }
+        }
+        if coinbase_counter != 1 {
+            warn!(
+                target: LOG_TARGET,
+                "{} coinbase kernels found in body. Only a single coinbase kernel is permitted.", coinbase_counter,
+            );
+            return Err(TransactionError::InvalidCoinbaseCount);
+        }
+        // Unwrap used here are fine as they should have an amount in them by here. If the coinbase's are missing the
+        // counters should be 0 and the fn should have returned an error by now.
+        let utxo = coinbase_utxo.unwrap();
+        let rhs =
+            &coinbase_kernel.unwrap().excess + &factories.commitment.commit_value(&BlindingFactor::default(), reward.0);
+        if rhs != utxo.commitment {
+            warn!(target: LOG_TARGET, "Coinbase {} amount validation failed", utxo);
+            return Err(TransactionError::InvalidCoinbase);
+        }
+        Ok(())
+    }
+
+    /// This function will check all stxo to ensure that feature flags where followed
+    pub fn check_stxo_rules(&self, height: u64) -> Result<(), TransactionError> {
+        for input in self.inputs() {
+            if input.features.maturity > height {
+                warn!(
+                    target: LOG_TARGET,
+                    "Input found that has not yet matured to spending height: {}", input
+                );
+                return Err(TransactionError::InputMaturity);
+            }
+        }
+        Ok(())
     }
 
     /// Validate this transaction by checking the following:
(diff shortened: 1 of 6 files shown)


base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
         fee
     }
 
+    /// This function will check spent kernel rules like tx lock height etc
+    pub fn check_kernel_rules(&self, height: u64) -> Result<(), TransactionError> {
+        for kernel in self.kernels() {
+            if kernel.lock_height > height {
+                debug!(target: LOG_TARGET, "Kernel lock height was not reached: {}", kernel);
```
CjS77: Breaking rules should be a WARN, and it'll make more sense in the log context to include the `height` as well.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
...
+        height: u64,
+    ) -> Result<(), TransactionError>
+    {
+        let mut aggregated_body = AggregateBody::empty();
+        let mut coinbase_counter = 0; // there should be exactly 1 coinbase
+        for utxo in self.outputs() {
+            if utxo.features.flags.contains(OutputFlags::COINBASE_OUTPUT) {
+                coinbase_counter += 1;
+                if utxo.features.maturity < (height + consensus_constants.coinbase_lock_height()) {
+                    warn!(target: LOG_TARGET, "Coinbase found with maturity set too low");
```
CjS77: You could include some info to id the utxo too in a noisy log context.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
...
+                    return Err(TransactionError::InvalidCoinbase);
```
CjS77: Maybe we should have more granular Coinbase errors. => CoinbaseNotMatured

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
...
+                }
+                aggregated_body.add_output(utxo.clone());
+            }
+        }
+        if coinbase_counter != 1 {
+            warn!(
+                target: LOG_TARGET,
+                "{} coinbases found in body. Only a single coinbase is permitted.", coinbase_counter,
+            );
+            return Err(TransactionError::InvalidCoinbase);
```
CjS77: More granular Coinbase errors => IncorrectCoinbaseCount

base_layer/core/src/validation/block_validators.rs
```diff
@@ -378,3 +387,49 @@ fn check_cut_through(block: &Block) -> Result<(), ValidationError> {
     }
     Ok(())
 }
+
+fn check_coinbase_output_dont_check_amount(
```
CjS77: This is only called from statelessValidator -- but you could just call check_coinbase_output, so this function seems totally unnecessary.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -238,14 +241,20 @@ fn check_block_weight(block: &Block, consensus_constants: &ConsensusConstants) -
...
+{
     trace!(
         target: LOG_TARGET,
         "Checking coinbase output on block with hash {}",
         block.hash().to_hex()
     );
+    let total_coinbase = rules.calculate_coinbase_and_fees(block);
     block
-        .check_coinbase_output(consensus_constants)
+        .check_coinbase_output(total_coinbase, &rules.consensus_constants(), factories)
```
CjS77:
```suggestion
        .check_coinbase_output(total_coinbase, rules.consensus_constants(), factories)
```

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
...
+    }
+
+    /// Run through the outputs of the block and check that
+    /// 1. There is exactly ONE coinbase output
+    /// 1. The output's maturity is correctly set
+    /// 1. The amount is correct.
+    pub fn check_coinbase_output(
+        &self,
+        reward: MicroTari,
+        consensus_constants: &ConsensusConstants,
```
CjS77: Rather take lock_height as an argument here, then you can drop the namespace import of ConsensusConstants AND it would make testing easier since you wouldn't have to create a whole CC object to test this fn.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -193,6 +195,86 @@ impl AggregateBody {
...
+            }
+        }
+        if coinbase_counter != 1 {
+            warn!(
+                target: LOG_TARGET,
+                "{} coinbase kernels found in body. Only a single coinbase kernel is permitted.", coinbase_counter,
+            );
+            return Err(TransactionError::InvalidCoinbase);
+        }
+        aggregated_body
```
CjS77: `validate_internal_consistency` is a very expensive call and it doesn't look to be saving any lines of code -- what with constructing a whole new AggregateBody --since you only really want to check the CB amount is correct. The signature check & Range proof check is going to be called again at some point when the full body is validated.

So just calculate and check Is CBUtxo.commitment == CBExcess + (Sum(fees) + reward).H? 
Then the checks above can be dramatically simplified too, e.g. something like

`counter = self.kernels().map_and_filter(|k| if k.features.contains(KernelFeatures::COINBASE_KERNEL) {1} else {0}).sum()`

base_layer/core/src/validation/block_validators.rs
```diff
@@ -190,7 +193,7 @@ impl StatelessValidation<Block> for MockStatelessBlockValidator {
     /// 1. Where all the rules for the spent outputs followed?
     /// 1. Was cut through applied in the block?
     fn validate(&self, block: &Block) -> Result<(), ValidationError> {
-        check_coinbase_output(block, &self.rules.consensus_constants())?;
+        check_coinbase_output_dont_check_amount(block, self.rules.clone(), &self.factories)?;
```
CjS77: You can just use the usual  check_coinbase_output fn here. Don't see why you're creating a new version

base_layer/core/src/validation/block_validators.rs
SWvheerden: This test is here to let some tests pass. 
Running it normally seems a lot of tests use the check_coinbase_output tests, but they do not have the correct output. One test actually uses this as a method to get an error.

base_layer/core/src/validation/block_validators.rs
SWvheerden: check comment above

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: Done

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: done

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: done

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: Done

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: done

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: Done. I think. 
You still need to find and map out the coinbase utxo and kernel to use its values.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -378,3 +387,44 @@ fn check_cut_through(block: &Block) -> Result<(), ValidationError> {
     }
     Ok(())
 }
+
+fn check_coinbase_output_dont_check_amount(block: &Block, rules: ConsensusManager) -> Result<(), ValidationError> {
```
stringhandler: This method seems unnecessary, can it be removed?

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -255,5 +255,82 @@ mod test {
...
+                42
+            ),
+            Err(TransactionError::InvalidCoinbase)
+        );
+        // testing that "block" is still valid
+        assert_eq!(
+            tx.body
+                .validate_internal_consistency(&BlindingFactor::default(), block_reward, &factories),
+            Ok(())
+        );
```
stringhandler, line 340: For sanity, can we add a call to one of the validators that catches the inflation created here?

stringhandler (dismissed):
Looks good. I will pull and test it locally. I would have liked to see a unit test that mirrors the exact vulnerability, but I think if you create an issue to create the test later, I'd be happy with that. Ideally there should be at least one test for each of the errors `InvalidKernel`, `InvalidCoinbaseMaturity`, `InvalidCoinbaseCount`, `InputMaturity`. 

The test that you've added covers the case where the incorrect fee is added, which is ok, but the block in that test also inflates the supply, which would hopefully is caught by another validator. A test that shows this vulnerability more clearly may be: 
1. Create a coinbase, set its value to 0.
2. Create another UTXO (as you have done) with the correct emission and fees that should have been in the coinbase.
3. Verify that the test finds it.

SWvheerden:
I added more tests to cover every single coinbase validation error there is

stringhandler (approved):
Happy with the new tests. (Didn't run locally, but will merge if CI passes)

PR #2144, opened 2020-08-18
[Merged Mining] Proxy Update

StriderDM:
## Description
Changed base url address of stagenet to one which has hash power assigned to it, moved constants into globalconfig.
Modified error handling so an error is not just logged and consumed, but that a result is always returned to XMRig whenever possible. Should not panic.
Split handle_connection function into handle_get and handle_post functions.
Added request_id for post methods.
Fixed clippy lints.
Improved logging further.

## Motivation and Context
Returns proxy to a functional state.

## How Has This Been Tested?
Have had it running for at least an hour with the below in 3 terminals.
```
cargo run --bin tari_base_node
cargo run --bin tari_merge_mining_proxy
./xmrig
```

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/main.rs
+++ b/applications/tari_merge_mining_proxy/src/main.rs
@@ -26,12 +26,14 @@ extern crate arrayref;
 extern crate chrono;
 extern crate jsonrpc;
 use chrono::Local;
+use config;
 use curl::easy::{Auth, Easy, List};
 use structopt::StructOpt;
-use tari_common::{ConfigBootstrap, ConfigError};
+use tari_common::{ConfigBootstrap, ConfigError, GlobalConfig};
 
 // TODO: Log to file
 use crate::error::MmProxyError;
+use json::JsonValue;
 use log::*;
 use monero::{
     blockdata::Block,
@@ -68,14 +70,6 @@ pub type PrivateKey = RistrettoSecretKey;
 
 pub const LOG_TARGET: &str = "tari_mm_proxy::app";
 
-// TODO: Refactor into a configuration file
-const MONEROD_URL: &str = "monero-stagenet.exan.tech:38081";
-const MONEROD_USER: &str = "user";
-const MONEROD_PASS: &str = "pass";
-const TARI_GRPC_URL: &str = "http://127.0.0.1:18142";
-const LOCALHOST: &str = "127.0.0.1:7878";
-const USE_AUTH: bool = false;
-
 #[derive(Clone, Debug)]
 pub struct TransientData {
     tari_block: Option<grpc::GetNewBlockResult>,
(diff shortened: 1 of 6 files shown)


applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -69,7 +69,8 @@ pub type PrivateKey = RistrettoSecretKey;
 pub const LOG_TARGET: &str = "tari_mm_proxy::app";
 
 // TODO: Refactor into a configuration file
-const MONEROD_URL: &str = "monero-stagenet.exan.tech:38081";
+// Address taken from https://community.xmr.to/xmr-seed-nodes/stagenet
+const MONEROD_URL: &str = "http://192.110.160.146:38081";
 const MONEROD_USER: &str = "user";
 const MONEROD_PASS: &str = "pass";
 const TARI_GRPC_URL: &str = "http://127.0.0.1:18142";
```
SWvheerden: We should probably move this toward a config file, but thats a future PR.

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -85,121 +86,113 @@ pub struct TransientData {
 }
 
 // setup curl authentication
-fn base_curl_auth(curl: &mut Easy) -> Result<(), MmProxyError> {
-    curl.username(MONEROD_USER)
-        .map_err(|_| MmProxyError::CurlError("Invalid username".to_string()))?;
-    curl.password(MONEROD_PASS)
-        .map_err(|_| MmProxyError::CurlError("Invalid password".to_string()))?;
+fn base_curl_auth(curl: &mut Easy) {
+    curl.username(MONEROD_USER).unwrap();
```
SWvheerden: this will cause an error right? Should we not rather fail gracefully with an error than an outright panic?

SWvheerden: And the same for all of the unwraps

SWvheerden (commented):
I flagged some of them. But I have a big problem with using unwrap() all of the place. If this gets an error , it will just panic and exit the thread.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Agreed.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Haven't managed to get it to panic on these unwraps, even with an invalid config.

The errors for curl are presented on the perform() function, which will be caught by the match.

There was an unchecked unwrap on line 657 which is fixed now and appears to be the source of the poisoned lock error.

stringhandler: Try to avoid unwraps in general. Even if they are 100% fine, it makes it harder and harder to review because you then have to check everywhere to see if the unwrap is ok

StriderDM: Good point. Am updating the code.

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -231,220 +206,406 @@ fn stringify_request(buffer: &[u8]) -> String {
...
-                                            .pow
-                                            .clone()
-                                            .ok_or_else(|| MmProxyError::MissingDataError("Pow data".to_string()))?;
-                                        powdata.pow_data = bincode::serialize(&pow_data).map_err(|_| {
-                                            MmProxyError::ParseError("Failure to serialize block".to_string())
-                                        })?;
+                                    Some(pow_data) => {
+                                        let mut tariheader: grpc::BlockHeader = block.header.unwrap();
+                                        let mut powdata = tariheader.pow.clone().unwrap();
+                                        powdata.pow_data = bincode::serialize(&pow_data).unwrap();
```
SWvheerden: We need to replace these unwraps. 
If there is a problem with malformed data, the thread will panic here

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -231,220 +206,406 @@ fn stringify_request(buffer: &[u8]) -> String {
...
                             if transient.tari_block.is_some() && transient.monero_seed.is_some() {
-                                let blockresult: grpc::GetNewBlockResult = transient
-                                    .clone()
-                                    .tari_block
-                                    .ok_or_else(|| MmProxyError::MissingDataError("New Block result".to_string()))?;
-                                let mut block: grpc::Block = blockresult
-                                    .block
-                                    .ok_or_else(|| MmProxyError::MissingDataError("Block".to_string()))?;
+                                let blockresult: grpc::GetNewBlockResult = transient.clone().tari_block.unwrap();
+                                let mut block: grpc::Block = blockresult.block.unwrap();
```
SWvheerden: unwraps

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -454,101 +615,232 @@ fn handle_connection(
...
+            match grpcclientresult {
+                Ok(mut grpcclient) => {
+                    match rt.block_on(grpcclient.get_new_block_template(grpc::PowAlgo { pow_algo: 0 })) {
+                        Ok(new_block_template_response) => {
+                            let new_template = new_block_template_response.into_inner().new_block_template.unwrap();
+                            match add_coinbase(consensus, new_template) {
+                                Some(coinbased_block) => match rt.block_on(grpcclient.get_new_block(coinbased_block)) {
+                                    Ok(newblock) => {
+                                        let block = newblock.into_inner();
+                                        let hash = block.clone().mining_data.unwrap().mergemining_hash;
```
SWvheerden: unwrap

SWvheerden (commented):
there is still a lot of unwraps here

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Removed unwraps.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Removed unwraps.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Removed unwraps.

common/config/presets/rincewind-simple.toml
```diff
@@ -33,3 +33,8 @@ peer_seeds = [
 enable_mining = false
 grpc_enabled = true
 grpc_address = "127.0.0.1:18142"
+monerod_url = "http://192.110.160.146:38081"
```
stringhandler, line 36: I think these should go into their own section. Perhaps `merge_mining.monero`?

stringhandler (approved):
Tested locally. Config settings can be moved at a later stage.

PR #2145, opened 2020-08-19
Tidy up balance equation

CjS77:
The sum commitments can be tidied up considerably by adding the fees in
the calling function. It also makes the overall balance code easier to
read.


stringhandler (approved):
Nice, seems neater

PR #2150, opened 2020-08-19
Correctly setup wallet default onion and local forwarding ports

sdbondi:
## Description

This PR removes a port collision if the user has an explicitly set
`tor_forward_address` set in their config. This option now only applies
to the base node. The wallet is always configured to use an OS-assign
forwarding port and a pre-defined wallet onion port of 18101 (the same
as the Aurora wallet).

Commented out `tor_forward_address` in config templates.

## Motivation and Context

Base node would fail to start up if `tor_forward_address` was set 
because both the base node and its wallet were configured to use this port.
In the vast majority of cases, `tor_forward_address` would never need to be set (since #2138 )  

## How Has This Been Tested?

Set `tor_forward_address` in base node and checked in logs that the listener is setup correctly and that inbound connections work.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/common/config/presets/windows.toml
+++ b/common/config/presets/windows.toml
@@ -215,7 +215,7 @@ tor_control_auth = "none" # or "password=xxxxxx"
 # The onion port to use.
 tor_onion_port = 18150
 # The address to which traffic on the node's onion address will be forwarded
-tor_forward_address = "/ip4/127.0.0.1/tcp/18150"
+# tor_forward_address = "/ip4/127.0.0.1/tcp/0"
 # Instead of attemping to get the SOCKS5 address from the tor control port, use this one. The default is to
 # use the first address returned by the tor control port (GETINFO /net/listeners/socks).
 #tor_socks_address_override=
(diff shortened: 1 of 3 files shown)


stringhandler (dismissed):

common/config/presets/windows.toml
```diff
@@ -215,7 +215,7 @@ tor_control_auth = "none" # or "password=xxxxxx"
 # The onion port to use.
 tor_onion_port = 18150
 # The address to which traffic on the node's onion address will be forwarded
-tor_forward_address = "/ip4/127.0.0.1/tcp/18150"
+# tor_forward_address = "/ip4/127.0.0.1/tcp/0"
```
hansieodendaal, line 218: Please change line 304 as well:
`#tor_forward_address = "/ip4/127.0.0.1/tcp/18141"` -> `#tor_forward_address = "/ip4/127.0.0.1/tcp/0"`

sdbondi: Didn't see that one - thanks - done

stringhandler (approved):

PR #2155, opened 2020-08-20
Re-enable artifacts and another s3 upload GHA

leet4tari:
## Motivation and Context
s3-sync-action only uploads with Linux GHA. Enabled GHA artifact uploads, so we have a backup plan for binaries
build by the GHA. Also replaced S3 GHA with ```aws s3``` bash command and tested on Linux/OSX/Win.  
  
Also updated filename as per @mikethetike request.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -32,6 +32,13 @@ jobs:
     - name: Checkout source code
       uses: actions/checkout@v2
 
+    - name: Declare Global Variables
+      id: vars
+      shell: bash
+      run: |
+        echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
     - name: Setup Rust toolchain
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/base_node_binaries.yml
```diff
@@ -143,3 +142,13 @@ jobs:
         AWS_REGION: '${{ secrets.AWS_REGION }}'
         SOURCE_DIR: '$GITHUB_WORKSPACE/dist'
         DEST_DIR: '${{ env.S3DESTDIR }}'
+
+    - name: Sync dist to S3 - Other
+      uses: shallwefootball/s3-upload-action@master
```
sdbondi: Since there are secrets being handed over - perhaps it would be safer to either use a fork or lock to a commit

sdbondi (commented):
LGTM - just  concerned about handing over secrets to a master branch that's not part of the Tari project

.github/workflows/base_node_binaries.yml
stringhandler: Yeah I agree. A fork would probably be the safest

leet4tari: Not sure we can lock to a commit, but we could possible use a tag. A fork could work too.

leet4tari: Replaced the GHA actions with ```aws s3``` cmd, was tricky, but works on Linux/OSX/Win.

stringhandler (commented):
Approved

PR #2157, opened 2020-08-24
Add grace period to SAF message request time

stringhandler:
Add minimum request time for requesting SAF messages

## Motivation and Context

On iOS specifically, there are cases where the DHT is shut down before the SAF request can finish. The next time it requests SAF messages, it will use the DHT shut down time. I suspect there may be windows where a message arrives after a node has requested SAF messages and before the DHT shuts down. This can result in missing messages. While it may be better to have a more robust timing mechanism, AFAIK there is no danger in receiving a message twice and so for now a simple fix is to just ask for all messages inside of the high priority TTL period. At a later stage it might be worth looking into this problem again.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/config.rs
+++ b/comms/dht/src/config.rs
@@ -93,6 +93,10 @@ pub struct DhtConfig {
     pub connectivity_random_pool_refresh: Duration,
     /// The active Network. Default: TestNet
     pub network: Network,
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Makes sense even if it's just to see if it improves reliability

SWvheerden (approved):
LGTM

PR #2161, opened 2020-08-25
Implement transaction negotiation message resending in the wallet

### LINKED ISSUE #1804: Monitor transaction counterparties with Liveness service and resend Tx Replies and Tx Finalized messages
Currently a Wallet that is a Transaction Sender that sends a transaction but has not received a Reply will monitor the Recipient for liveness using the Liveness service (PR https://github.com/tari-project/tari/pull/1777). If the Recipient comes online and hasn't sent a reply yet the transaction will be sent again.

However, a wallet that is the Recipient of a Transaction will only send the Transaction Reply once via Store and Forward and attempt to send it once directly. 

It would be more robust if the Recipient wallet uses the Liveness service to monitor if their transaction counterparty's come online. If they do come online and they have not yet sent the Transaction Finalized messages then the Recipient wallet should resend the Tx Reply to have the Sender wallet complete the transactions.

It is also possible that that the Transaction Finalized message did not make it back to the Tx Recipient wallet so if a Transaction Sender wallet receives a Tx Reply for a completed transaction it should still send the Tx Finalized message so that the Recipient can close out the transaction.

### LINKED ISSUE #2137: Improve Transaction Service robustness by periodically resending messages during negotiation
Currently the logic for sending transactions and negotiating them between parties is not very robust. Each stage in the negotiation is sent only once. When sending we first try to send the message directly and if that fails, or discovery needs to occur, the message is sent indirectly via Store and Forward (SAF). There was an initial hope that we would be able to get the reliability of Store and Forward to be 100%. But, we have realized that due to the unpredictable app shutdowns that occur so often in the mobile apps and the long timeouts and other issues that Tor presents that it is actually not possible to be 100% sure that a given message made it to the counter-party.

We did start to explore a strategy of resending incomplete transaction negotiation messages, but it was during the time where we were experiencing a lot of networking flooding. So we decided to hold back on that strategy to rather try focus on indirect communication reliability. We have come to the conclusion that there are some unsolvable issues with the current communications stack that we just have to live with. We have also solved the bugs with message propogation that resulted in the previous flooding. Based on these conclusions we propose to implement an automatic transaction message resending strategy to improve the robustness of the transaction negotiation process with the limitations of the SAF messaging in mind.

Below we summarize the messages sent between Sender and Receiver during a Transaction negotiation:

| Message  | Sender | | Receiver  |
|---|---|---|---|
| 1 | Sends `SenderPartialTransaction` message  | `>`  |    |
| 2 | | `<` | Sends `RecipientSignedMessage`|
|   | Completes the transaction and can now start the broadcast protocol |  |  |
| 3 | Sends `TransactionFinalizedMessage` | `>` |   |
|   |  | | Once received the Receiver can now start the broadcast and chain monitoring protocols |

So the strategy I propose is that a wallet will respond to a repeated message from the counterparty and when expecting a message a wallet will periodically resend the last message from the negotiation to illicit the response. For example, once Message 1 is sent if the Sender has not received Message 2 it will periodically resend Message 1 in order to illict Message 2 from the counter party. Once the Receiver has received Message 1 they will send Message 2 but until they receive Message 3 they will periodically resend Message 2. The resending will be governed by the following rules:

1. The party will resend when the wallet starts up if it has been more than 60 minutes since the last time a message was sent.
2. If the wallet has been open continuously it will resend the message every 60 minutes.
3. If the transaction is not completed after 3 days it is cancelled.

Currently, the mobile wallet frontends cancel the transaction after 3 days. I propose that this cancellation be done by the backend (with the period being configurable) so that this behaviour is consistent between mobile apps and the desktop clients. @Kukabi and @Jasonvdb do you agree with this change?

Informing the counter-party of a manual cancellation has been on the cards for a while but with this new resend strategy it will be even more important in order to reduce unnecessary traffic on the network. So part of this implementation will be to implement a new `CancelTransaction` message that can be used to inform the counterparty of a manual cancellation and also as a response when a counterparty resend a message for a transaction that is already cancelled.

One last element of this proposal are some rules about how often a response will be expected. This is to reduce the potential for DoS by an attacker. A party will only respond to resent messages once every 5 minutes per transaction. This should be plenty of time when messages are only being resent every 60 minutes.

Thoughts @sdbondi and @mikethetike?

philipr-za:
## Description
Currently the logic for sending transactions and negotiating them between parties is not very robust. Each stage in the negotiation is sent only once. When sending we first try to send the message directly and if that fails, or discovery needs to occur, the message is sent indirectly via Store and Forward (SAF). We hoped to get SAF 100% reliable but due to reasons laid out in https://github.com/tari-project/tari/issues/2137 it is not tenable. 

This PR adds in the logic to periodically resends the initial Transaction Sender Message and the Transaction Reply in order to illicit the next step in the negotiation protocol from the counterparty. The PR also adds the logic to respond to repeated messages but with a cool down period in which repeats will be ignored to protect against DoS attacks. 

The PR also adds in a new comms level message to inform the counterparty if a transaction protocol has been cancelled. Initially this is sent if the Sender cancels an in progress transaction so that the receiver knows to stop expecting it. This message is also sent if a Transaction Reply is received for a cancelled message to let the Received know about the cancellation.

In order to stop a wallet resending these message indefinitely a transaction will be cancelled if it hasn’t resolved after a long timeout period (Default is being set to 3 days).

A large portion of the PR is tests and some code reorganization so its not as big as it looks.

## Motivation and Context
Closes https://github.com/tari-project/tari/issues/2137
Closes https://github.com/tari-project/tari/issues/1804

## How Has This Been Tested?
Tests are provided that test:
1. A sender repeats sending the initial message if no response is received from the receiver
2. A receiver repeats sending the Reply in no response is received from the receiver
3. For both parties that the repeated message is sent on startup if the period has elapsed.
4. For both parties that they can handle receiving the repeated messages and respond appropriately
5. Check that both parties ignore repeated messages sent before the cool down has elapsed
6. Check the cancellation messages are sent by the Sender when a transaction is cancelled
7. Check that receiver acts correctly on receiving a cancellation message
8. Check that the transaction is cancelled by both parties after the timeout is reached during operation or on startup.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/p2p/src/proto/tari.p2p.message_type.rs
+++ b/base_layer/p2p/src/proto/tari.p2p.message_type.rs
@@ -15,8 +15,9 @@ pub enum TariMessageType {
     BaseNodeResponse = 70,
     MempoolRequest = 71,
     MempoolResponse = 72,
-    /// -- DAN Messages --
     TransactionFinalized = 73,
+    /// -- DAN Messages --
+    TransactionCancelled = 74,
     // -- Extended --
     Text = 225,
     TextAck = 226,
--- a/base_layer/wallet/src/lib.rs
+++ b/base_layer/wallet/src/lib.rs
@@ -5,7 +5,7 @@
 #![deny(unused_must_use)]
 #![deny(unreachable_patterns)]
 #![deny(unknown_lints)]
-#![recursion_limit = "1024"]
+#![recursion_limit = "2048"]
 #![feature(drain_filter)]
 #![feature(type_alias_impl_trait)]
 
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
@@ -26,13 +26,11 @@ use crate::{
         error::{TransactionServiceError, TransactionServiceProtocolError},
         handle::TransactionEvent,
         service::TransactionServiceResources,
-        storage::database::{
-            CompletedTransaction,
-            InboundTransaction,
-            TransactionBackend,
-            TransactionDirection,
-            TransactionStatus,
+        storage::{
+            database::TransactionBackend,
+            models::{CompletedTransaction, InboundTransaction, TransactionDirection, TransactionStatus},
         },
+        tasks::send_transaction_reply::send_transaction_reply,
     },
 };
 use chrono::Utc;
@@ -44,20 +42,16 @@ use futures::{
 use log::*;
 use rand::rngs::OsRng;
 use std::sync::Arc;
-use tari_comms::{peer_manager::NodeId, types::CommsPublicKey};
-use tari_comms_dht::{
-    domain_message::OutboundDomainMessage,
-    envelope::NodeDestination,
-    outbound::{MessageSendStates, OutboundEncryption, SendMessageResponse},
-};
+use tari_comms::types::CommsPublicKey;
+
 use tari_core::transactions::{
     transaction::{OutputFeatures, Transaction},
-    transaction_protocol::{proto, recipient::RecipientState, sender::TransactionSenderMessage},
+    transaction_protocol::{recipient::RecipientState, sender::TransactionSenderMessage},
     types::PrivateKey,
     ReceiverTransactionProtocol,
 };
 use tari_crypto::keys::SecretKey;
-use tari_p2p::tari_message::TariMessageType;
+use tokio::time::delay_for;
 
 const LOG_TARGET: &str = "wallet::transaction_service::protocols::receive_protocol";
 
@@ -146,7 +140,6 @@ where TBackend: TransactionBackend + Clone + 'static
             }
 
             let amount = data.amount;
-
             let spending_key = self
                 .resources
                 .output_manager_service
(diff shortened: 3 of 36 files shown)


base_layer/p2p/src/proto/tari.p2p.message_type.rs
```diff
@@ -15,8 +15,9 @@ pub enum TariMessageType {
     BaseNodeResponse = 70,
     MempoolRequest = 71,
     MempoolResponse = 72,
-    /// -- DAN Messages --
     TransactionFinalized = 73,
+    /// -- DAN Messages --
+    TransactionCancelled = 74,
```
sdbondi, line 20:
```suggestion
    TransactionCancelled = 74,
     // -- DAN Messages --
```

sdbondi: Clicked "Add single comment" instead of "Start a review" :P

philipr-za: I keep trying to do that but that is the generated file

base_layer/wallet/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![recursion_limit = "1024"]
```
sdbondi, line 8: Oof - I think that this may [be resolved in rust >=1.45 (July 2020)](https://github.com/rust-lang/futures-rs/issues/1917#issuecomment-548135409) - so a compiler upgrade will eventually remove the need to set recursion limit.

base_layer/p2p/src/proto/tari.p2p.message_type.rs
sdbondi: Ah I missed that - nevermind then :)

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
```diff
@@ -406,10 +250,75 @@ where TBackend: TransactionBackend + Clone + 'static
...
+                warn!(
+                    target: LOG_TARGET,
+                    "Error resending Transaction Reply (TxId: {}): {:?}", self.id, e
+                );
+            }
+            self.resources
+                .db
+                .increment_send_count(self.id)
+                .await
+                .map_err(|e| TransactionServiceProtocolError::new(self.id, TransactionServiceError::from(e)))?;
```
sdbondi, line 313: Shouldn't `increment_send_count` only be called if the `send_transaction_reply` succeeds?

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
```diff
@@ -430,6 +339,29 @@ where TBackend: TransactionBackend + Clone + 'static
...
+                                target: LOG_TARGET,
+                                "Error resending Transaction Reply (TxId: {}): {:?}", self.id, e
+                            );
+                        } else {
+                            self.resources
+                                .db
+                                .increment_send_count(self.id)
+                                .await
+                                .map_err(|e| TransactionServiceProtocolError::new(self.id, TransactionServiceError::from(e)))?;
+                        }
```
sdbondi: `match` is more idiomatic

```suggestion
                        match send_transaction_reply(
                            inbound_tx.clone(),
                            self.resources.outbound_message_service.clone(),
                            self.resources.config.direct_send_timeout,
                        )
                        .await
                        {
                        Ok(_) => {
                           self.resources
                                .db
                                .increment_send_count(self.id)
                                .await
                                .map_err(|e| TransactionServiceProtocolError::new(self.id, TransactionServiceError::from(e)))?;
                        },
                        Err(err) => {
                            warn!(
                                target: LOG_TARGET,
                                "Error resending Transaction Reply (TxId: {}): {:?}", self.id, e
                            );
                        
                        }
```

sdbondi (dismissed):
LGTM - will run a few tests on the branch before approving

base_layer/wallet/src/lib.rs
philipr-za: Ja, might also be the trigger to split up the Tx Service a bit though need to figure out how to do it without adding more wiring

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
philipr-za: I considered it as an attempt was made to send and we will only try again (succeed or fail) again in the next period. Wanted to be careful not to get stuck attempting too many sends as it would thrash a mobile's battery.

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
philipr-za: Damnit Stanley, why do you hate the classic C syntax so much?

philipr-za: But ok will update but honestly its because the classic syntax is easier to format manually inside a select! macro

sdbondi: Yeess .. submit to the will of rust

sdbondi (dismissed):
All my stuck transactions on the base node went through after one retry 🎉

PR #2162, opened 2020-08-25
Added tari_rpc proc macro trait attribute

sdbondi:
## Description

This PR adds a proc macro attribute to generate the "harness code" for
a service that implements the trait.

```rust
#[tari_rpc(protocol_name = "/t/greeting/1.0", server_struct = GreetingService, client_struct = GreetingClient)]
pub trait GreetingRpc: Send + Sync + 'static {
  #[rpc(method = 1)]
  async fn say_hello(&self, request: Request<SayHelloRequest>) -> Result<Response<SayHelloResponse>, RpcStatus>;
  #[rpc(method = 2)]
  async fn return_error(&self, request: Request<()>) -> Result<Response<()>, RpcStatus>;
  #[rpc(method = 3)]
  async fn get_greetings(&self, request: Request<u32>) -> Result<Streaming<String>, RpcStatus>;
}
```

`tari_rpc` options
- `protocol_name` is the value used during protocol negotiation
- `server_struct` is the name of the "server" struct that is generated
- `client_struct` is the name of the client struct that is generated

`rpc` attribute
- `method` is a unique number assigned to each function. Once a method
id is used it should never be reused (similar to protobuf field
numbers)

## Motivation and Context

Provides simple to use code generation for RPC service and client "harness" code - allowing the user to define Tari comms RPC services in a similar way to .proto `service` definitions.

## How Has This Been Tested?

Generated code compiles and behaviour tested.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


stringhandler (dismissed):
Looks good. I think it might be best to merge it and then add issues when it's used, because it might change before then

sdbondi:
Updated/rebased the branch

stringhandler (approved):

Issue #2101, opened 2020-07-31
Mempool never cleared of invalid transaction, so miner can never mine a block


stringhandler:
**Describe the bug**

A miner asks a base node for a block, but the base node has an invalid transaction in the mempool, for example a double spend or an already mined transaction. Any blocks that the miner mine will be invalid and whenever the miner asks the base node for a block, it includes the transaction.
It's not clear how the mempool gets the invalid transaction, but it could possibly be that it syncs from another base node that is on an older tip. 

The following logs are present (see attached):
```
2020-07-30 09:28:15.275142524 [c::val::block_validators] WARN  Block validation failed because the block has invalid input: bcd8c25773c0c28d2ad63cdf288b9b1882f2a8fc719d5c8bdd7a22fad7ca1338 [OutputFeatures { flags: (empty), maturity: 0 }]

2020-07-30 09:28:15.275156299 [c::cs::database] WARN  Orphan block 65517 (d7ee2885566cd7891fca45a30e26eb86d695cec27d23684499657374fb9304f7) failed validation during chain reorg.
2020-07-30 09:28:15.281637759 [c::cs::data
```

[mining-logs.txt.zip](https://github.com/tari-project/tari/files/5007246/mining-logs.txt.zip)

stringhandler:
I have a feeling something like this is happening:
```
1. Node 1 has Tx with excess sig xs_a in its mempool
2. Node 2 starts up, and starts syncing and gets xs_a in its mempool
3. Node 1 receives a block and removes xs_a from the mempool
4. Node 2 eventually syncs, but somehow still has xs_a
5. Node 3 starts, and receives xs_a from Node 2
6. Node 2 and 3 continue to receive blocks, but will never remove xs_a because none of the blocks it receives will contain xs_a
7. In addition, any nodes that sync with Node 2 and 3 will receive xs_a
```

Could be wrong though

SWvheerden:
Thats sound very likely, I agree with this 100%. 

We either have to clear/check the mempool after sync is done, or we have check each transaction as we put in in the mempool

sdbondi:
Sounds very plausible - a call to `txn.validate_internal_consistency` in the mempool sync protocol before inserting would fix it.

EDIT: Actually the `Mempool::insert` function already runs the validators. So we do check each transaction as they are inserted.

delta1:
Firstly my apologies if I'm asking obvious stuff :) 

Going through that log @mikethetike at the first signs of the issue: 

[...]

PR #2166, opened 2020-08-26
Fix issue where mempool is not cleared of invalid tx

delta1:
## Description

This commit fixes a logical error where a mempool reorg is not
being processed when more than one block is added and _no_ blocks are
removed. Previously when no blocks were removed, `handle_reorg` would
return `BlockAddResult::Ok` - even if multiple blocks were added. This
introduced a bug where transactions from the added blocks could be left
in the mempool. A test was added to `chain_storage` to ensure a reorg in
this case.

As a consequence of now forcing a reorg even if no blocks are removed,
an additional issue was uncovered with `Mempool::process_reorg` causing
a panic when `removed_blocks` was empty. A test has been added to catch
this case, and the code refactored to remove calls to `expect` in favor
of explicitly handling the `Option`s.

## Motivation and Context
Issue #2101 

## How Has This Been Tested?
* Existing tests pass
* New tests catch the issues they are fixing
* Manually ran the node 

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1582,25 +1582,38 @@ fn handle_reorg<T: BlockchainBackend>(
         .height -
         1;
     let removed_blocks = reorganize_chain(db, block_validator, fork_height, reorg_chain)?;
-    if removed_blocks.is_empty() {
-        Ok(BlockAddResult::Ok)
-    } else {
-        debug!(
+    let num_removed_blocks = removed_blocks.len();
+    let num_added_blocks = added_blocks.len();
+
+    // reorg is required when any blocks are removed or more than one are added
+    // see https://github.com/tari-project/tari/issues/2101
+    if num_removed_blocks > 0 || num_added_blocks > 1 {
+        info!(
             target: LOG_TARGET,
-            "Chain reorg processed from (accum_diff:{}, hash:{}) to (accum_diff:{}, hash:{})",
+            "Chain reorg required from {} to {} (accum_diff:{}, hash:{}) to (accum_diff:{}, hash:{}). Number of \
+             blocks to remove: {}, to add: {}.
+            ",
+            tip_header,
+            fork_tip_header,
             tip_header.pow,
             tip_header.hash().to_hex(),
             fork_tip_header.pow,
-            fork_tip_hash.to_hex()
-        );
-        info!(
-            target: LOG_TARGET,
-            "Reorg from ({}) to ({})", tip_header, fork_tip_header
+            fork_tip_hash.to_hex(),
+            num_removed_blocks,
+            num_added_blocks,
         );
         Ok(BlockAddResult::ChainReorg((
             Box::new(removed_blocks),
             Box::new(added_blocks),
         )))
+    } else {
+        trace!(
+            target: LOG_TARGET,
+            "No reorg required. Number of blocks to remove: {}, to add: {}.",
+            num_removed_blocks,
+            num_added_blocks,
+        );
+        Ok(BlockAddResult::Ok)
     }
 }
 
--- a/base_layer/core/src/mempool/mempool_storage.rs
+++ b/base_layer/core/src/mempool/mempool_storage.rs
@@ -166,32 +166,35 @@ where T: BlockchainBackend
             );
         }
 
-        let prev_tip_height = removed_blocks
-            .last()
-            .expect("Added empty set of blocks on reorg.")
-            .header
-            .height;
-        let new_tip_height = new_blocks
-            .last()
-            .expect("Removed empty set of blocks on reorg.")
-            .header
-            .height;
+        let previous_tip = removed_blocks.last().map(|block| block.header.height);
+        let new_tip = new_blocks.last().map(|block| block.header.height);
+
         self.insert_txs(
             self.reorg_pool
                 .remove_reorged_txs_and_discard_double_spends(removed_blocks, &new_blocks)?,
         )?;
         self.process_published_blocks(new_blocks)?;
 
-        if new_tip_height < prev_tip_height {
-            debug!(
-                target: LOG_TARGET,
-                "Checking for time locked transactions in unconfirmed pool as chain height was reduced from {} to {} \
-                 during reorg.",
-                prev_tip_height,
-                new_tip_height,
-            );
-            self.pending_pool
-                .insert_txs(self.unconfirmed_pool.remove_timelocked(new_tip_height))?;
+        if let (Some(previous_tip_height), Some(new_tip_height)) = (previous_tip, new_tip) {
+            if new_tip_height < previous_tip_height {
+                debug!(
+                    target: LOG_TARGET,
+                    "Checking for time locked transactions in unconfirmed pool as chain height was reduced from {} to \
+                     {} during reorg.",
+                    previous_tip_height,
+                    new_tip_height,
+                );
+                self.pending_pool
+                    .insert_txs(self.unconfirmed_pool.remove_timelocked(new_tip_height))?;
+            } else {
+                debug!(
+                    target: LOG_TARGET,
+                    "No need to check for time locked transactions in unconfirmed pool. Previous tip height: {}. New \
+                     tip height: {}.",
+                    previous_tip_height,
+                    new_tip_height,
+                );
+            }
         }
 
         Ok(())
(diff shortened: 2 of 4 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1582,9 +1582,16 @@ fn handle_reorg<T: BlockchainBackend>(
         .height -
         1;
     let removed_blocks = reorganize_chain(db, block_validator, fork_height, reorg_chain)?;
-    if removed_blocks.is_empty() {
-        Ok(BlockAddResult::Ok)
-    } else {
+    let num_removed_blocks = removed_blocks.len();
+    let num_added_blocks = added_blocks.len();
+
+    if reorg_required(num_removed_blocks, num_added_blocks) {
```
philipr-za: This is a nit but isn't this a little verbose including the function below for a single logical statement? It feels like it's perfectly readable if the logical statement and the comment are up here rather than in their own function.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1601,9 +1608,41 @@ fn handle_reorg<T: BlockchainBackend>(
...
 }
 
+fn reorg_required(num_removed_blocks: usize, num_added_blocks: usize) -> bool {
+    // if any blocks are being removed
+    // or more than one block is being added
+    // then a reorg is required
+    num_removed_blocks > 0 || num_added_blocks > 1
+}
+
+#[test]
```
philipr-za: We tend to keep unit tests like this inside of a module called test with a cfg flag to show it must not be compiled unless we are testing. We also keep this module at the bottom of these files so they are not interleaved with actual code.

```Rust
#[cfg(test)]
mod test {
    #[test] 
    fn my_test() {...}
}

File: base_layer/core/src/chain_storage/blockchain_database.rs [hunk 3]
```diff
@@ -1601,9 +1608,41 @@ fn handle_reorg<T: BlockchainBackend>(
...
+fn test_reorg_required() {
+    // no blocks removed, only one block added - no reorg
+    assert!(reorg_required(0, 1) == false);
+
+    // more than one block removed - should reorg
+    assert!(reorg_required(1, 0) == true);
```
philipr-za: the `==true` is superfluous
```suggestion
    assert!(reorg_required(1, 0));
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1601,9 +1608,41 @@ fn handle_reorg<T: BlockchainBackend>(
...
 
+fn reorg_required(num_removed_blocks: usize, num_added_blocks: usize) -> bool {
+    // if any blocks are being removed
+    // or more than one block is being added
+    // then a reorg is required
+    num_removed_blocks > 0 || num_added_blocks > 1
+}
+
+#[test]
+fn test_reorg_required() {
```
philipr-za: So I appreciate explaining the behaviour in the comments of this test but this test is pretty trivial, you are essentially testing a single logical statement, which is easy to read, with all this code and in the end it makes the codebase noisier. I would say explain the desired behaviour in a comment perhaps? It would be helpful to test the actual behaviour in a more functional way?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1582,9 +1582,16 @@ fn handle_reorg<T: BlockchainBackend>(
...
         1;
     let removed_blocks = reorganize_chain(db, block_validator, fork_height, reorg_chain)?;
-    if removed_blocks.is_empty() {
-        Ok(BlockAddResult::Ok)
-    } else {
+    let num_removed_blocks = removed_blocks.len();
+    let num_added_blocks = added_blocks.len();
+
+    if reorg_required(num_removed_blocks, num_added_blocks) {
+        trace!(
```
stringhandler: I think just add this to the `info!`. The debug could probably even be removed

base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -166,32 +166,27 @@ where T: BlockchainBackend
...
+                debug!(
+                    target: LOG_TARGET,
+                    "Checking for time locked transactions in unconfirmed pool as chain height was reduced from {} to \
+                     {} during reorg.",
+                    previous_tip_height,
+                    new_tip_height,
+                );
+                self.pending_pool
+                    .insert_txs(self.unconfirmed_pool.remove_timelocked(new_tip_height))?;
+            }
```
stringhandler, line 197:
```suggestion
            }
            else {
             info!(.... "Processed reorg, but no reorg was required")
            }
```

stringhandler: This means we can see in the logs if this branch is ever hit. Could also be at `debug` level

base_layer/core/tests/mempool.rs
```diff
@@ -397,6 +397,13 @@ fn test_reorg() {
     assert_eq!(stats.unconfirmed_txs, 2);
     assert_eq!(stats.timelocked_txs, 1);
     assert_eq!(stats.published_txs, 3);
+
+    // test that process_reorg can handle the case when removed_blocks is empty
+    // see https://github.com/tari-project/tari/issues/2101#issuecomment-680726940
+    let template = chain_block(&blocks[3], vec![], consensus_manager.consensus_constants());
+    let reorg_block4 = db.calculate_mmr_roots(template).unwrap();
+
+    assert_eq!(mempool.process_reorg(vec![], vec![reorg_block4]).unwrap(), ());
```
stringhandler: Out of interest, this would also be ok, since a panic would fail the test
```suggestion
    mempool.process_reorg(vec![], vec![reorg_block4]).unwrap();
```

stringhandler (commented):
Happy to merge, but I think put the data in the `trace!()` into the `info!()` statement.

philipr-za (commented):
Nice catch, some comments that are mostly stylistic nits but otherwise LGTM

base_layer/core/tests/mempool.rs
delta1: Yeah I did previously have it that way but wanted to make it more explicit. Will revert

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: I agree it's a bit verbose but it is explicit. ~Instead of creating new variables I'll just pass in the lengths.~ I respectfully disagree about the function though, given that there's currently no integration test for this the `handle_reorg` function (afaik). At least this way the intent is clear, explicit, and has a (admittedly simple) test to catch a regression in case the `reorg_required` function is changed. Deferring to @mikethetike - keep the `reorg_required` function and test for now?

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: Gotcha - will fix

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: I knew someone would catch this :) Will revert

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: Agreed that if we had a better test of the `handle_reorg` function end-to-end that it's unnecessary. I'll take a look at adding that. In the meanwhile it seems dumb and overly verbose, but it's a fixing an issue that was difficult to uncover and is only a handful of lines. I'll neaten it up to the minimum. Will fix

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: Thanks - will fix

base_layer/core/src/mempool/mempool_storage.rs
delta1: Good idea - will add a debug to indicate that the new tip was >= previous tip and so checking for time-locked in unconfirmed was not necessary

SWvheerden (dismissed):
LGTM

base_layer/core/src/chain_storage/blockchain_database.rs
philipr-za: Fair enough, it was a nit and as you say perhaps the verbosity is justified here.

philipr-za (approved):
LGTM

SWvheerden (approved):
LGTM

PR #2170, opened 2020-08-27
Control POW algorithms via block percentage

SWvheerden:
## Description

Adds in a consensus constant to control how the block chain should control the percentage of blocks going to which algorithm. This is added in as an array of u8. This specifies the amount as a percentage of 100. 
Also removed the algorithm count constant as the array can be used to calculate this as well. Its better to have it being controlled by a single constant than two. 

When doing hybrid mining its important to adjust the target time of the algorithm in respect to the percentage of blocks you want. Normally if you want 2 algorithms with a 50-50 split the individual target time becomes `Target = target_time/0.5` which looks like it should be `target_time+algo_count `. This PR fixes this algorithm as well. 

## Motivation and Context

As part of the discussion on IRC it was noted that there was a desire to favour monero blocks over pure tari blocks. This PR allows the user to specify what percentage the POW should be divided as in 60-40, 50-50 etc. 

## How Has This Been Tested?

New tests added showing how the split works even with 3 algo's. More is supported. The array just needs to increase. 
The values have been matched to simulations done. Rounding issues do result in a second or two difference. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -42,6 +42,7 @@ pub struct ConsensusConstants {
     /// This is the our target time in seconds between blocks
     target_block_interval: u64,
     /// When doing difficulty adjustments and FTL calculations this is the amount of blocks we look at
+    /// https://github.com/zawy12/difficulty-algorithms/issues/14
     difficulty_block_window: u64,
     /// When doing difficulty adjustments, this is the maximum block time allowed
     difficulty_max_block_interval: u64,
@@ -141,7 +142,7 @@ impl ConsensusConstants {
             // 0
             return 0;
         }
-        ((self.target_block_interval as f64) / self.algo_split[algo] as f64 * 100.0) as u64
+        self.target_block_interval * 100 / (self.algo_split[algo] as u64)
     }
 
     /// The maximum time a block is considered to take. Used by the difficulty adjustment algorithms
@@ -154,7 +155,7 @@ impl ConsensusConstants {
             // 0
             return 0;
         }
-        ((self.difficulty_max_block_interval as f64) / self.algo_split[algo] as f64 * 100.0) as u64
+        self.difficulty_max_block_interval * 100 / (self.algo_split[algo] as u64)
     }
 
     /// This is how many blocks we use to count towards the median timestamp to ensure the block chain moves forward.
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -47,8 +47,6 @@ pub struct ConsensusConstants {
     difficulty_max_block_interval: u64,
     /// Maximum transaction weight used for the construction of new blocks.
     max_block_transaction_weight: u64,
-    /// The amount of PoW algorithms used by the Tari chain.
-    pow_algo_count: u64,
     /// This is how many blocks we use to count towards the median timestamp to ensure the block chain moves forward
     median_timestamp_count: usize,
     /// This is the initial emission curve amount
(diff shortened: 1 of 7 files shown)


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -63,6 +61,9 @@ pub struct ConsensusConstants {
     genesis_coinbase_value_offset: MicroTari,
     /// This is the maximum age a monero merge mined seed can be reused
     max_randomx_seed_height: u64,
+    /// This keeps track of the block split targets and which algo is accepted
+    /// Ideally this should count up to 100. If this does not you will reduce your target time.
+    algo_split: [u8; 3],
```
CjS77, line 67: Hmm, you're combining 2 functions in one field here. It's confusing and makes all the code below 5x more complicated than it needs to be.

I would actually just have an array of target block times here, and a function that returns the aggregate target block time, instead of explicitly targeting a given split.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -121,19 +122,39 @@ impl ConsensusConstants {
 
     /// The amount of PoW algorithms used by the Tari chain.
     pub fn get_pow_algo_count(&self) -> u64 {
-        self.pow_algo_count
+        let mut count = 0;
```
CjS77, line 126: Case in point, if it were just the target block times, or straight splits, this would be one line:
`self.algo_split.len()`

CjS77 (commented):
I would recommend take a different approach this, as outlined in my review.

base_layer/core/src/consensus/consensus_constants.rs
SWvheerden: I am replaying to this below, as the single replay is covers both

base_layer/core/src/consensus/consensus_constants.rs
SWvheerden: I am replaying to this below, as the single replay is covers both

SWvheerden: But more specifically, I dont think an array of target block times will just be able to say .len() As you might have 3 algorithms defined in your pow struct. Or you are running local_net with purely a single pow. Then you need to make it an vec, and make sure it only matches your single algo, meaning len = 0. 

But that does not solve the case we you are running blake , which is algo 2, how do you know which split to use.

If I understand your think correct here, if you only want to run blake. You would have an `algo_split = [120]`
But the get_target time function needs to know which index to look at, `algo_split[1] != [120]`, its going to do index out of bounds. So you have to do `algo_split = [0,120]`
But then `algo_split.len() != algo_count `

SWvheerden:
The reason I merged the fields, is because the fields are related. Currently we have:
 `Block_chain_target_time`
`Algo count`

Now we are adding an `algo split as well`. All of these fields are related. Changing one will change the effect the other one has. They all influence the target time each algorithm which will in turn influence the chain_block_time regardless of what the target in the consensus is set there. 

I get it that we could technically just use the target_times in the algo_split array. But to me thats more confusing. 
I know from the simulations and fact that I know a the formule, so I calculated this, that an array split of [200,300] means that you would have an algorithm split of 60% for the first, 40% for the second. But I have no idea what the block_chain_target_time is.  Thats not possible to get here. 

Except you can leave that field, so you would have it. But it would not be used. If `block_chain_target = 120` and you would then change `algo_split= [300,300]` the actual block chain time would be 150, not the specified 120.  That would cause confusion. 

The main reason I coded it as follows is that everyone can read and understand wat the block_chain_target_time is, and everyone can read and understand what a split of [60,40] means if its percentages. They can easily change this for local_net and use different results and know what they will be getting out. There is no reason to go look at simulations and or do calculations to understand what the constants are by looking at them.  

The way its defined in this PR I believe you only change the 2 things you as user really care about:
1. The blockchain block interval
2. The Pow algo split. 

You as a user doesnt really care about individual target times. The block_chain cares but not the user.

stringhandler:
I'm also in favour of just having an array of target block times for each algo.

SWvheerden:
I removed the floating pointing point math here. 
I also created PR #2231 to use just the target time as the input parameters. The two methods can now be compared.

stringhandler:
Closed in favour of #2231

PR #2160, opened 2020-08-25
Peer manager and SAF code cleanup

sdbondi:
## Description

- Remove "bubble sort" code from peer manager in favour of Rust native
  optimized merge sort in `closest` and `in_network_region` calls
- Removed extra database gets in `closest_peers`, `random_peers` and `in_network_region` calls
- Removed unused get_region_stats call
- Removed unused `distance_threshold` SAF field
- Removed unused `InRegion` SAF message request.
- Added tests for `in_network_region`
- _memorynet_ uses SAF requests on connect for reliability
- _memorynet_ added some more details to output about the connectivity state of a
  wallet and it's neighbours when requesting SAF messages

## Motivation and Context

Removed some "crufty" code from peer manager. SAF `InRegion` call is not used at all and can be removed - this call required the `distance_threshold` field and so that was also removed. This should make SAF requests a little lighter (locally and over the wire)  

## How Has This Been Tested?

Added tests for `in_network_region` + existing tests pass. Tested memorynet (best test for SAF and DHT behaviour)

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #2175, opened 2020-08-27
Implement DHT event and Wallet FFI callback to be triggered when “enough” SAF responses are received

philipr-za:
## Description

This PR implements a DHT event bus with 1 new event that indicates that the DHT has received “enough” Store and Forward responses from its connected neighbours that we are relatively confident we have received any SAF responses due for us.

The SAF Handler on the inbound pipeline now sends a signal every time it handles a SAF response to the StoreAndForwardService. The StoreAndForwardService counts these signals and waits for the Event from the Connectivity manager to indicate when the node is considered Online. Online means that a proportion of neighbouring peers have been connected too. The SAF service then waits until it has received SAF response for at least the number of connected peers returned by the ConnectivityEvent and then it will fire the DhtEvent::StoreAndForwardMessagesReceived.

Currently a Node will not send a SAF response if it has no SAF messages to send but to enable this functionality this PR changes that logic to send an empty response so that the requesting node knows for sure there were no messages waiting for it.

The Wallet FFI Callback handler was extended to accept a new callback for to be called on this event. The Callback handler now subscribes to the DHT event bus and calls this new callback when the DhtEvent::StoreAndForwardMessagesReceived is received.

**NOTE: This PR will only start to work when the Peers in the network are also running this code because until then they will not respond if there are not SAF messages to be delivered.**

@kukabi and @Jasonvdb Please note the change to `wallet_create(…)`

## How Has This Been Tested?
Tests updated to test:
1. That the SAF middleware does signal when it receives a SAF response
2. That the SAF service does emit the event 
3. That the Callback handler does fire when it receives the event
4. That the SAF service does respond when it has no SAF messages to return.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/tests/dht.rs
+++ b/comms/dht/tests/dht.rs
@@ -39,6 +39,7 @@ use tari_comms::{
 use tari_comms_dht::{
     domain_message::OutboundDomainMessage,
     envelope::NodeDestination,
+    event::DhtEvent,
     inbound::DecryptedDhtMessage,
     outbound::{OutboundEncryption, SendMessageParams},
     DbConnectionUrl,
@@ -299,6 +300,7 @@ async fn dht_discover_propagation() {
 #[tokio_macros::test]
 #[allow(non_snake_case)]
 async fn dht_store_forward() {
+    let _ = env_logger::try_init();
     let node_C_node_identity = make_node_identity(PeerFeatures::COMMUNICATION_NODE);
     // Node B knows about Node C
     let node_B = make_node(PeerFeatures::COMMUNICATION_NODE, None).await;
@@ -351,12 +353,19 @@ async fn dht_store_forward() {
     collect_stream!(node_B_msg_events, take = 2, timeout = Duration::from_secs(20));
 
     let mut node_C = make_node_with_node_identity(node_C_node_identity, Some(node_B.to_peer())).await;
+    let mut node_C_dht_events = node_C.dht.subscribe_dht_events();
     let mut node_C_msg_events = node_C.comms.subscribe_messaging_events();
     // Ask node B for messages
     node_C
         .dht
         .store_and_forward_requester()
         .request_saf_messages_from_peer(node_B.node_identity().node_id().clone())
+        .await
+        .unwrap();
+    node_C
+        .dht
+        .store_and_forward_requester()
+        .request_saf_messages_from_peer(node_A.node_identity().node_id().clone())
         .await
         .unwrap();
     // Wait for node C to and receive a response from the SAF request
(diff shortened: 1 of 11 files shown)


stringhandler (dismissed):
Looks good. Didn't test locally though

comms/dht/src/builder.rs
```diff
@@ -75,6 +75,11 @@ impl DhtBuilder {
         self
     }
 
+    pub fn enable_auto_store_and_forward_requests(mut self) -> Self {
```
sdbondi: I've actually added something similar here: https://github.com/tari-project/tari/pull/2160/files#diff-91f3ccfba0c0a74bd4bac16bb43b8813 - might be better if use one (I vote for a single set function rather than enable/disable) - feel free to leave it as is and I'll clean it up when merging my PR, since it probably comes after this one

comms/dht/tests/dht.rs
```diff
@@ -485,6 +500,7 @@ async fn dht_propagate_dedup() {
     let received = filter_received(collect_stream!(node_A_messaging, timeout = Duration::from_secs(20)));
     // Expected race condition: If A->(B|C)->(C|B) before A->(C|B) then (C|B)->A
     if received.len() > 0 {
+        println!("Events: {:?}", received);
```
sdbondi:
```suggestion

```

comms/dht/tests/dht.rs
```diff
@@ -134,22 +142,23 @@ async fn setup_comms_dht(
...
-    .disable_auto_store_and_forward_requests()
     .with_database_url(DbConnectionUrl::MemoryShared(random::string(8)))
     .with_discovery_timeout(Duration::from_secs(60))
-    .with_num_neighbouring_nodes(8)
-    .finish()
-    .await
-    .unwrap();
+    .with_num_neighbouring_nodes(8);
 
+    if with_saf {
```
sdbondi: Did it break any tests in this suite to have it always enabled?

If not - I say that you can just always have it on and avoid having an extra boolean param

sdbondi (commented):
LGTM - few inconsequential comments - going to test and approve if all goes well

comms/dht/tests/dht.rs
philipr-za: It did break a few tests that made assumptions on how many messages are sent and I was not comfortable updating them because I wasn't 100% sure what their intent was.

comms/dht/tests/dht.rs
philipr-za: Woops

comms/dht/tests/dht.rs
sdbondi: Ah right, ye it makes sense to remove all the noise so that tests can remain focused

sdbondi: Though now that I think about it - I think you should add an explicit call from node C to Node A (currently there is one for Node C -> B) - then I believe you wont need auto saf requests enabled

philipr-za: Cool, done

sdbondi (approved):
Ship it!

PR #2176, opened 2020-08-28
[Merged Mining] Bug fixes and additional tests.

StriderDM:
## Description
Moved test to monero_rx.
Added test to check verify header is working.
Fixed potential bug in merge_mining_proxy.
Removed asserts from all functions except tests.

## How Has This Been Tested?
cargo test --all

## Types of changes
* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/proof_of_work/monero_rx.rs
+++ b/base_layer/core/src/proof_of_work/monero_rx.rs
@@ -20,7 +20,13 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use crate::{blocks::BlockHeader, proof_of_work::Difficulty, tari_utilities::ByteArray, U256};
+use crate::{
+    blocks::BlockHeader,
+    proof_of_work::Difficulty,
+    tari_utilities::ByteArray,
+    validation::ValidationError,
+    U256,
+};
 use monero::{
     blockdata::{
         block::{Block as MoneroBlock, BlockHeader as MoneroBlockHeader},
@@ -42,8 +48,8 @@ pub enum MergeMineError {
     SerializeError(String),
     #[error("Error deserializing Monero data")]
     DeserializeError,
-    #[error("Hashing of Monero data failed")]
-    HashingError,
+    #[error("Hashing of Monero data failed: {0}")]
+    HashingError(String),
     #[error("RandomX error: {0}")]
     RandomXError(#[from] RandomXError),
     #[error("Validation error: {0}")]
@@ -75,31 +81,46 @@ pub fn cn_fast_hash(data: &[u8]) -> Vec<u8> {
 }
 
 // Tree hash count in monero
-fn tree_hash_cnt(count: usize) -> usize {
-    assert!(count >= 3);
-    assert!(count <= 0x10000000);
+fn tree_hash_cnt(count: usize) -> Result<usize, MergeMineError> {
+    if count < 3 {
+        return Err(MergeMineError::HashingError(
+            "Cannot calculate Monero root, algorithm path error".to_string(),
+        ));
+    }
+
+    if count > 0x10000000 {
+        return Err(MergeMineError::HashingError(
+            "Cannot calculate Monero root, hash count too large".to_string(),
+        ));
+    }
 
     let mut pow: usize = 2;
     while pow < count {
         pow <<= 1;
     }
-    pow >> 1
+
+    Ok(pow >> 1)
 }
 
 /// Tree hash algorithm in monero
 #[allow(clippy::needless_range_loop)]
-pub fn tree_hash(hashes: Vec<Hash>) -> Vec<u8> {
-    assert!(!hashes.is_empty());
+pub fn tree_hash(hashes: Vec<Hash>) -> Result<Vec<u8>, MergeMineError> {
+    if hashes.is_empty() {
+        return Err(MergeMineError::HashingError(
+            "Cannot calculate Monero root, hashes is empty".to_string(),
+        ));
+    }
+
     match hashes.len() {
-        1 => hashes[0].0.to_vec(),
+        1 => Ok(hashes[0].0.to_vec()),
         2 => {
             let mut buf: [u8; 64] = [0; 64];
             buf[..32].copy_from_slice(&hashes[0].0.to_vec());
             buf[32..].copy_from_slice(&hashes[1].0.to_vec());
-            cn_fast_hash(&buf)
+            Ok(cn_fast_hash(&buf))
         },
         _ => {
-            let mut cnt = tree_hash_cnt(hashes.len());
+            let mut cnt = tree_hash_cnt(hashes.len())?;
             let mut buf: Vec<u8> = Vec::with_capacity(cnt * 32);
 
             for i in 0..(2 * cnt - hashes.len()) {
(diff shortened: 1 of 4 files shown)


base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
-        .coinbase_tx
-        .prefix
-        .extra
-        .0
-        .contains(&merged_mining_subfield(header)))
-    {
+    let mut found = false;
+    for item in monero_data.coinbase_tx.prefix.extra.0.clone() {
+        match item {
+            SubField::MergeMining(x, y) => {
```
sdbondi: Please name these more descriptively - according to the `SubField` docs: `SubField::MergeMining(depth, merkle_root)` - but maybe since we're using that field for the block header hash: `SubField::MergeMining(depth, hash)`.

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
+                let compare = Hash::from(from_slice(&header.merged_mining_hash()));
+                if (x, y) == (VarInt(0), compare) {
+                    found = true;
```
sdbondi: Can we exit early if the field is found - or is it important to fail if there are any other MergeMining fields which don't match?

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
+        let serialized = bincode::serialize(&monero_data).unwrap();
+        let pow = ProofOfWork {
+            accumulated_monero_difficulty: Default::default(),
+            accumulated_blake_difficulty: Default::default(),
+            target_difficulty: Default::default(),
+            pow_algo: PowAlgorithm::Monero,
+            pow_data: serialized,
+        };
+        block_header.pow = pow;
+        let monero_data_check = MoneroData::new(&block_header).unwrap();
```
sdbondi: Unused variable

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
+        let pow = ProofOfWork {
+            accumulated_monero_difficulty: Default::default(),
+            accumulated_blake_difficulty: Default::default(),
+            target_difficulty: Default::default(),
+            pow_algo: PowAlgorithm::Monero,
+            pow_data: serialized,
+        };
+        block_header.pow = pow;
+        let monero_data = MoneroData::new(&block_header).unwrap();
+        verify_header(&block_header, &monero_data).unwrap();
```
sdbondi, line 562: This checks the happy path - but I think some tests that check the failure behaviour - i.e given inputs `bad_block_header` and `monero_data` I expect error X etc. are valuable.

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
-        .extra
-        .0
-        .contains(&merged_mining_subfield(header)))
-    {
+    let mut found = false;
+    for item in monero_data.coinbase_tx.prefix.extra.0.clone() {
+        match item {
+            SubField::MergeMining(x, y) => {
+                let compare = Hash::from(from_slice(&header.merged_mining_hash()));
+                if (x, y) == (VarInt(0), compare) {
```
sdbondi: I don't hate this "trick" - but if you don't have to use a trick in the first place usually results in clearer code.

```suggestion
                if depth == VarInt(0) && merge_mining_hash == hash {
```

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +231,343 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
-        .prefix
-        .extra
-        .0
-        .contains(&merged_mining_subfield(header)))
-    {
+    let mut found = false;
+    for item in monero_data.coinbase_tx.prefix.extra.0.clone() {
+        match item {
+            SubField::MergeMining(x, y) => {
+                let compare = Hash::from(from_slice(&header.merged_mining_hash()));
```
sdbondi: What is `compare`? Name it `merge_mining_hash`. This should be moved out of the loop so that it doesn't have to be recalculated every iteration.

sdbondi (commented):
Looks good - few code comments - have not tested it

base_layer/core/src/proof_of_work/monero_rx.rs
StriderDM: Thanks, should exit as soon as a match occurs.

base_layer/core/src/proof_of_work/monero_rx.rs
StriderDM: The problem was that I needed to match a `typeof()` on a `enum` which is why I did it this way.

base_layer/core/src/proof_of_work/monero_rx.rs
StriderDM: Fair comment, will add more tests. Thanks.

base_layer/core/src/proof_of_work/monero_rx.rs
sdbondi: In that case I suggest using the `Iterator::any` call

```rust
let monero_data = &monero_data.coinbase_tx.prefix.extra.0;
let is_found = monero_data.iter().any(|item| {
        match item {
            SubField::MergeMining(x, y) => depth == VarInt(0) && merge_mining_hash == hash,
             _ => false 
}});
```

base_layer/core/src/proof_of_work/monero_rx.rs
sdbondi: Awesome - thanks

StriderDM: Have added more tests, think I have covered most cases now.

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -231,25 +257,538 @@ fn verify_root(monero_data: &MoneroData) -> Result<(), MergeMineError> {
...
+            pow_algo: PowAlgorithm::Monero,
+            pow_data: serialized,
+        };
+        block_header.pow = pow;
+        let monero_data = MoneroData::new(&block_header).unwrap();
+        verify_header(&block_header, &monero_data).unwrap();
+    }
+
+    #[test]
+    #[should_panic]
```
sdbondi: Generally speaking, the `#[should_panic]` attribute should almost never be needed. Rather than allowing that the test to pass when panicking (which can happen in all sorts of cases where yo'd want the test to fail) make assertions which will panic if they are not met.

The `unpack_enum!` utility macro in `tari_test_utils` helps a lot with error enums.

```rust
use tari_test_utils::unpack_enum;

// (....)

let err =  verify_header(&block_header, &monero_data).unwrap_err();
// I may prefer adding a new error variant to describe this specific error e.g MergeMineError::MoneroCoinbaseTagNotFound - but up to you
unpack_enum!(MergeMineError::ValidationError(details) = err);

 assert!(details.contains("merge mining tag was not found"));


```

sdbondi (commented):
Rather assert something in the positive than using `#[should_panic]`

stringhandler (approved):

PR #2179, opened 2020-08-28
Fix geometric mean comparison halting the blockchain

hansieodendaal:
## Description
When the geometric mean of a new block is compared to that of the network, it must be accepted if it is equal, due to the number being rounded. `geo_mean = sqrt(achieved_difficulty_01 ^ 2 x achieved_difficulty_02 ^ 2)`

## Motivation and Context

If achieved difficulty is `49999999` in the example below, the absolute rounded up geometric mean stays the same:
|           | diff_01 | diff_02             | geo_mean   |
| --------- | ------- | ------------------- | ---------- |
| chain tip | 1       | 740 390 525 623 642 | 27 210 119 |
| new block | 1       | 740 390 575 623 641 | 27 210 119 |

## How Has This Been Tested?
On two Windows nodes

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/validation/accum_difficulty_validators.rs
+++ b/base_layer/core/src/validation/accum_difficulty_validators.rs
@@ -35,7 +35,7 @@ impl<B: BlockchainBackend> Validation<Difficulty, B> for AccumDifficultyValidato
         let tip_header = db
             .fetch_last_header()?
             .ok_or_else(|| ValidationError::custom_error("Cannot retrieve tip header. Blockchain DB is empty"))?;
-        if *accum_difficulty <= tip_header.total_accumulated_difficulty_inclusive() {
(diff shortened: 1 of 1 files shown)


stringhandler (approved):

SWvheerden (changes requested):
I am for the code, 
But as it is here it will allow a block of difficulty 0 to come in. 
We need to check that the incoming block is greater than the minimum of that algo.  There is a consensus constant call for that

stringhandler:
@SWvheerden the difficulty can not be 0 because the target difficulty will be validated, and there is a minimum difficulty. The method `total_accumulated_difficulty_inclusive` is actually badly named because it is the geometric mean of the accumulated difficulty.

SWvheerden:
I double checked, yes you are correct. That is checked elsewhere in the code

PR #2183, opened 2020-08-31
Fix restoration of previous blocks after failed reorg

### LINKED ISSUE #2182: Reorg failure can cause database corruption in some cases
**Describe the bug**
When a reorg occurs where more than one block is reorg'd _out_ and, for whatever reason, the reorg fails - the previous state before the reorg should be recovered. However, during a recent bug with the block validator which caused a false negative, blocks that were reorged out could not be reinserted. The restoration process itself failed and caused the database to contain state that the base node software did not expect (a.k.a corrupted).

**To Reproduce**
Somehow cause a _multiblock_ reorg with blocks that will fail stateful validation.

**Expected behavior**
The blockchain database should correctly revert the changes and be in a consistent state. 

**Additional context**
This is caused by the "failed reorg restoration code" attempting to reinsert blocks in height _descending_ order instead of _ascending_. 

Additionally, the "failed reorg restoration code" expects that reinsertion of a (previously validated and inserted) blocks cannot fail, however if the blocks are inserted in the incorrect order - this assumption is no longer correct.

e.g. 

This chain has been detached and must be readded to the main chain
  
` -- 51000 -- 51001 -- 51002`

The code currently adds 50164 first (using `store_new_block`, which expects the next mainchain tip block), this causes the chain metadata to be set to 50164, the header to be inserted, but then may fail to spend the transactions contained in 50164 and so fails leaving chain metadata set incorrectly and a "dangling" header. `-- 51000 -x- 51002` 

It should be noted that `store_new_block` should be atomic - so it should be investigated as to why the atomicity was not observed here. Because of this, it is possible that this is not the source of the corruption observed - however, other than the seeming failure of the atomicity guarantees, this fits very well - moreover, the logs point unequivocally to this happening as described.

---

Following this, after the base node state machine finishes `WAITING` and attempts to add another block, the following error/warn is observed: 

`WARN  Block #85161 (db19bce67c11a2338a9141e407830466129577fd26a47468048f934f709b5bc7) validation failed: InvalidQuery("Cannot retrieve header. Blockchain DB is empty"`

This is due to the "dangling" header and the way that `fetch_last_header` works: it expects `header_db.count() - 1` to always equal current block height. This is no longer true due to the gaps in the header db.

sdbondi:
## Description

This PR reverses the order in which blocks that were reorged out are put
back after a failed reorg. During a reorg, the main chain is rewound to
a height where the reorged blocks link to the main chain. These rewound
blocks are kept in an in-memory buffer and are ordered in height
_decending_ order. In the event of a failed reorg, the rewound blocks
should be readded to the main chain from the current mainchain tip
upwards. This PR reverses the order that the in-memory blocks are added
to the main chain.

- Added `restore_reorged_chain` function, called in the event of a
  failure (validation, failure to commit) when reorging. [link](https://github.com/tari-project/tari/pull/2183/files#diff-e3daff4e22512430954760b7e8d4b6c6R1672)
-  `store_new_block` now adds the required operations to a mutable `DbTransaction`.
  This allows other operations to be performed atomically when storing a
  block.
- [bugfix] Restore chain on failed database commit and simplified `reorganize_chain` a bit 
  (previously it was assumed that the commit cannot fail because the
  validators had already checked the block) [link](https://github.com/tari-project/tari/pull/2183/files#diff-e3daff4e22512430954760b7e8d4b6c6R1658)
- [possible bugfix] Added `lmdb_txn_get` and `lmdb_txn_exists` functions. While a write
  transcation is in progress, any read operations should be using that
  transaction to obtain the most current state. As a contrived example:
  UTXO A exists. The db transaction `Txn[Spend(UTXO(A)), Unspend(UTXO(A))]` is performed.
  An `UnspendError` would result because the `lmdb_get` on the
  `stxos_db` would not know about the previous uncommitted spend
  operation. This is no longer the case as read operations are aware of the transaction.
- Removed all occurrences where references to `RwLock(Read|Write)Guard`s are passed around.
  This is bad form (basically it's bad because it's verbose and makes the function needlessly less flexible, 
   testing the function a little harder) and does not change anything functionally - 
   references and mutable references are now passed around where applicable.   
- [bugfix] Fixed incorrect order of operations for UTXO state rewinding (cf: [blockchain_database.rs lines: 1385-1404](https://github.com/tari-project/tari/pull/2183/files#diff-e3daff4e22512430954760b7e8d4b6c6R1378)). 
  When rewinding the UTXO checkpoints, the TXOs corresponding to nodes_added would be deleted _and then_ the STXOs 
   would be unspent. This means that if a UTXO was added and spent within the same block the rewind would error (ValueNotFound) - the rewind (therefore reorg) would rollback and fail. This changes the order to first unspend the STXO (moving it into the `utxo_db`) and then delete the added UTXOs. handle_reorg_failure_recovery only passes with this change
- [tests] Added `handle_reorg_failure_recovery` test to check 1) recovery from failure
  to commit reorged blocks, and 2) recovery from failed block validation and 3) That the UTXO rewinding is done correctly.
  This test uses the LMDB backend to increase parity between production
  and automated testing. [link](https://github.com/tari-project/tari/pull/2183/files#diff-0ec739352cf5afda478eca4db8610fc2R917)
- Allow the pass/fail flag to be set in the same test in `MockValidator`
- Various logging improvements

## Motivation and Context

Closes #2182.

## How Has This Been Tested?

New LMDB chain storage integration test. Basic checks on my local base nodes (still syncs and adds blocks, tested with full blocks on rincewind).

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
@@ -25,6 +25,7 @@ use lmdb_zero::{
     error::{self, LmdbResultExt},
     put,
     ConstAccessor,
+    ConstTransaction,
     Cursor,
     CursorIter,
     Database,
@@ -63,7 +64,7 @@ where T: DeserializeOwned {
         .map_err(|e| error::Error::ValRejected(e.to_string()))
 }
 
-pub fn lmdb_insert<K, V>(txn: &WriteTransaction, db: &Database, key: &K, val: &V) -> Result<(), ChainStorageError>
+pub fn lmdb_insert<K, V>(txn: &WriteTransaction<'_>, db: &Database, key: &K, val: &V) -> Result<(), ChainStorageError>
 where
     K: Serialize,
     V: Serialize,
@@ -81,7 +82,7 @@ where
         })
 }
 
-pub fn lmdb_replace<K, V>(txn: &WriteTransaction, db: &Database, key: &K, val: &V) -> Result<(), ChainStorageError>
+pub fn lmdb_replace<K, V>(txn: &WriteTransaction<'_>, db: &Database, key: &K, val: &V) -> Result<(), ChainStorageError>
 where
     K: Serialize,
     V: Serialize,
@@ -100,19 +101,18 @@ where
 }
 
 /// Deletes the given key. An error is returned if the key does not exist
-pub fn lmdb_delete<K>(txn: &WriteTransaction, db: &Database, key: &K) -> Result<(), ChainStorageError>
+pub fn lmdb_delete<K>(txn: &WriteTransaction<'_>, db: &Database, key: &K) -> Result<(), ChainStorageError>
 where K: Serialize {
     let key_buf = serialize(key)?;
     txn.access().del_key(&db, &key_buf)?;
     Ok(())
 }
 
-pub fn lmdb_get<K, V>(env: &Environment, db: &Database, key: &K) -> Result<Option<V>, ChainStorageError>
+pub fn lmdb_get_txn<K, V>(txn: &ConstTransaction<'_>, db: &Database, key: &K) -> Result<Option<V>, ChainStorageError>
 where
     K: Serialize,
     V: DeserializeOwned,
 {
-    let txn = ReadTransaction::new(env).map_err(|e| ChainStorageError::AccessError(e.to_string()))?;
     let access = txn.access();
     let key_buf = serialize(key)?;
     match access.get(&db, &key_buf).to_opt() {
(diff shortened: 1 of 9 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
```diff
@@ -134,9 +134,17 @@ where
...
+pub fn lmdb_get<K, V>(env: &Environment, db: &Database, key: &K) -> Result<Option<V>, ChainStorageError>
+where
+    K: Serialize,
+    V: DeserializeOwned,
+{
     let txn = ReadTransaction::new(env).map_err(|e| ChainStorageError::AccessError(e.to_string()))?;
+    lmdb_txn_get(&txn, db, key)
+}
+
+pub fn lmdb_txn_exists<K>(txn: &ConstTransaction<'_>, db: &Database, key: &K) -> Result<bool, ChainStorageError>
```
stringhandler: I originally thought this meant "check if a transactions exists", so maybe `lmdb_exists_txn` better. It'll also show up in intellisense nearer the original method

stringhandler (dismissed):
Nice. I think this is much neater and I like the use of transactions. I think the naming would be better, as I mentioned, but that's not important for now. It's difficult to test this, but the test seems to be correct.

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
sdbondi: Ye ideally I think we should get rid of `lmdb_txn_exists` and just have `lmdb_exists` that takes a transaction (move the responsibility of creating db transactions in lmdb_db) - just didnt want to make too many changes

sdbondi: But I agree with the name change - will resolve conflicts and rename

stringhandler (approved):

PR #2187, opened 2020-09-01
Fix hash estimation for block explorer

SWvheerden:
## Description

Changed the network hash estimation to use the target difficulty and not the total network accumulated difficulty. 

## Motivation and Context

Currently the hash estimate of the network hash rate is looking at accumulated difficulty. This means that the calculation of the network hash rate is dependant on how old the chain is, and not the target difficulty that needed to be solved for that specific block. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


hansieodendaal (dismissed):
LGTM

stringhandler (commented):
I think we should actually add two fields here:
1. achieved_hash_rate
2. target_hash_rate

This will make it more robust and clear in future

stringhandler (approved):

PR #2188, opened 2020-09-01
Added basic RPC interface (get_peers) to DHT

sdbondi:
## Description

- Added `DhtRpcService` trait with `get_peers` call
- `get_peers` returns peers that are closer to the caller's `NodeId`
- Added tests for `get_peers`
- Added some basic test utils/mocks to make testing RPC easier
- Added infallible `NodeId::from_public_key` function

## Motivation and Context

Part of an implementation of _outbound only_  peer discovery which is more in-line with kademlia.

## How Has This Been Tested?

Number of unit tests testing all code paths in `get_peers` call

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/proto/mod.rs
+++ b/comms/dht/src/proto/mod.rs
@@ -22,8 +22,13 @@
 
 use crate::proto::{dht::JoinMessage, envelope::Network};
 use rand::{rngs::OsRng, RngCore};
-use std::fmt;
-use tari_comms::NodeIdentity;
+use std::{convert::TryInto, fmt};
+use tari_comms::{
+    multiaddr::Multiaddr,
+    peer_manager::{NodeId, Peer, PeerFeatures, PeerFlags},
+    types::CommsPublicKey,
+    NodeIdentity,
+};
 use tari_utilities::{hex::Hex, ByteArray};
 
 #[path = "tari.dht.envelope.rs"]
@@ -31,6 +36,9 @@ pub mod envelope;
 
 #[path = "tari.dht.rs"]
 pub mod dht;
+
+#[path = "tari.dht.rpc.rs"]
+pub mod rpc;
 
 #[path = "tari.dht.store_forward.rs"]
 pub mod store_forward;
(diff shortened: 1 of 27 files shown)


comms/dht/src/proto/mod.rs
```diff
@@ -88,3 +96,42 @@ impl fmt::Display for dht::JoinMessage {
...
+                .iter()
+                .map(|addr| addr.address.to_string())
+                .collect(),
+            peer_features: peer.features.bits(),
+        }
+    }
+}
+
+impl rpc::Peer {
+    pub fn to_native_peer(&self) -> Option<Peer> {
```
philipr-za: Not a job for a try_into impl?

comms/dht/src/rpc/service.rs
```diff
@@ -0,0 +1,109 @@
...
+            peers.len(),
+            message.n,
+            node_id.short_str()
+        );
+
+        if peers.is_empty() {
+            return Ok(Streaming::empty());
+        }
+
+        let (mut tx, rx) = mpsc::channel(cmp::min(10, peers.len() as usize));
```
philipr-za, line 96: What is this 10 based on? Should perhaps be a const explaining or else a config item?

comms/dht/src/proto/mod.rs
sdbondi: 🤦 You're right! I think I wanted to `impl TryFrom<rpc::Peer> for Peer` which would obviously not compile

comms/dht/src/rpc/service.rs
sdbondi: Nothing in particular - it's just a bit of space to keep things flowing and the `cmp::min` is probably being a little obsessive about optimising this/quick win - I should add a comment in the code actually. The size is arbitrary really (how much space do you want to use per request to get_peers) so I think a const is overkill (brings too much attention to something which is not important).

philipr-za (dismissed):
LGTM, the RPC stuff is looking great.

philipr-za (approved):

PR #2190, opened 2020-09-02
Add chain strength comparer for dealing with ties

stringhandler:
- Add chain strength comparer
- Update reorg logic to use strength comparer
- Add limit to find_header_with_achieved_difficulty
- Add test blockchain for cleaner tests
- Add fetch all orphans

## Description

Add chain strength comparer for dealing with ties in difficulty. The order can be configured in the `ConsensusManager`. Example:
```
  chain_strength_comparer: 
                strongest_chain()
                    .by_accumulated_difficulty()
                    .then()
                    .by_monero_difficulty()
                    .then()
                    .by_blake_difficulty()
                    .then()
                    .by_height()
                    .build(),
            
```

Also add a test blockchain class that makes tests easier to read and write. I've taken the `handle_reorg` test in `base_layer/core/tests/chain_storage_tests/chain_storage.rs` and rewritten it as follows. The test failed locally, so I still need to investigate.

```rust
#[test]
fn blockchain_reorgs_to_stronger_chain() {
    let mut blockchain = TestBlockchain::with_genesis("GB");
    let blocks = blockchain.builder();
    blockchain.add_block(blocks.new_block("A1").child_of("GB").difficulty(1));
    blockchain.add_block(blocks.new_block("A2").child_of("A1").difficulty(3));
    blockchain.add_block(blocks.new_block("A3").child_of("A2").difficulty(1));
    blockchain.add_block(blocks.new_block("A4").child_of("A3").difficulty(1));

    assert_eq!(Some(blockchain.tip()), blockchain.get_block("A4"));
    assert!(blockchain.orphan_pool().is_empty());

    blockchain.add_block(blocks.new_block("B2").child_of("A1").difficulty(1));
    assert_eq!(Some(blockchain.tip()), blockchain.get_block("A4"));
    blockchain.add_block(blocks.new_block("B3").child_of("B2").difficulty(1));
    assert_eq!(blockchain.chain(), ["GB", "A1", "A2", "A3", "A4"]);
    assert_eq!(Some(blockchain.tip()), blockchain.get_block("A4"));

    blockchain.add_block(blocks.new_block("B4").child_of("B3").difficulty(5));
    // Should reorg
    assert_eq!(Some(blockchain.tip()), blockchain.get_block("B4"));

    blockchain.add_block(blocks.new_block("C4").child_of("B3").difficulty(20));
    assert_eq!(Some(blockchain.tip()), blockchain.get_block("C4"));

    assert_eq!(blockchain.chain(), ["GB", "A1", "B2", "B3", "C4"]);
}
```

## Motivation and Context

To make tests simple to write and visually inspect for errors

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -482,6 +482,19 @@ where B: BlockchainBackend
         fetch_orphan(&*db, hash)
     }
 
+    pub fn fetch_all_orphans(&self) -> Result<Vec<Block>, ChainStorageError> {
+        let db = self.db_read_access()?;
+        let mut result = vec![];
+        // TODO: this is a bit clumsy in order to safely handle the results. There should be a cleaner way
```
sdbondi, line 483: I did some work that never got finished on lazy-loaded iterators (for LMDB but with the intention to do it in a db agnostic way) - which should replace the need for the for_each functions and having to allocate space for all items - and since `Result` implements `FromIterator`, errors can be handled cleanly.

sdbondi (commented):
Really clean API 💯  Some test infrastructure that make blockchain tests easier to write and reason about are badly needed. Will approve if tests pass

base_layer/core/src/chain_storage/blockchain_database.rs
stringhandler: Yeah, this is a bit clumsy. I think this method will be temporary. Looking forward to using you new interface

hansieodendaal (commented):
Just commenting on the `ChainStrengthComparer `.

I like this tie-breaker approach suggested by @stringhandler in `#tar-dev IRC`:

"_In the case of the ~~difficulties~~ **geometric mean** being identical, couldn't we apply modulo to the hash and take the greatest result of the two? E.g. hash % 2^n, where n gives us the range of numbers (n=8, modulo will be 0-255)_"

We have discussed something similar in the past, using a hash of the header, but could not reach an agreement at the time. This will be simple and easy to verify. To make things a bit fairer, we could use u128 for geometric mean to include more significant digits, instead of u64.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1508,16 +1520,24 @@ fn handle_reorg<T: BlockchainBackend>(
...
+        return Err(ChainStorageError::InvalidOperation("No chain tips found in orphan pool".to_string()));
+    }
+    let fork_header = fork_header.unwrap();
+    let fork_tip_hash = fork_header.hash();
+
     let tip_header = db
         .fetch_last_header()?
         .ok_or_else(|| ChainStorageError::InvalidQuery("Cannot retrieve header. Blockchain DB is empty".into()))?;
-    if fork_tip_hash == new_block_hash {
+    if fork_header.hash() == new_block_hash {
```
SWvheerden: why hash it here again? You stored it in a var 5 lines up

```suggestion
    if fork_tip_hash == new_block_hash {
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1561,26 +1581,13 @@ fn handle_reorg<T: BlockchainBackend>(
...
-        },
+        }
     }
 
     // We've built the strongest orphan chain we can by going backwards and forwards from the new orphan block
     // that is linked with the main chain.
-    let fork_tip_block = fetch_orphan(db, fork_tip_hash.clone())?;
-    let fork_tip_header = fork_tip_block.header.clone();
+    let fork_tip_block = fetch_orphan(db, fork_header.hash())?;
+   // let fork_tip_header = fork_tip_block.header.clone();
```
SWvheerden: why leave this here?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1561,26 +1581,13 @@ fn handle_reorg<T: BlockchainBackend>(
...
-            return Err(err.into());
-        },
+        }
     }
 
     // We've built the strongest orphan chain we can by going backwards and forwards from the new orphan block
     // that is linked with the main chain.
-    let fork_tip_block = fetch_orphan(db, fork_tip_hash.clone())?;
-    let fork_tip_header = fork_tip_block.header.clone();
+    let fork_tip_block = fetch_orphan(db, fork_header.hash())?;
```
SWvheerden: Dont you have the value stored?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1850,19 +1857,23 @@ fn find_orphan_chain_tips<T: BlockchainBackend>(db: &T, parent_height: u64, pare
...
     db: &T,
     orphan_chain_tips: Vec<BlockHash>,
-) -> Result<(Difficulty, BlockHash), ChainStorageError>
+    chain_strength_comparer: &dyn ChainStrengthComparer
+) -> Result<Option<BlockHeader>, ChainStorageError>
 {
-    let mut best_accum_difficulty = Difficulty::min();
-    let mut best_tip_hash: Vec<u8> = vec![0; 32];
+    let mut best_block_header : Option<BlockHeader> =None;
+    //let mut best_tip_hash: Vec<u8> = vec![0; 32];
```
SWvheerden: commented out code?

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -174,6 +190,17 @@ impl ConsensusManagerBuilder {
...
+                strongest_chain()
+                    .by_accumulated_difficulty()
+                    .then()
+                    .by_monero_difficulty()
+                    .then()
+                    .by_blake_difficulty()
+                    .then()
+                    .by_height()
+                    .build(),
+            ),
```
SWvheerden: I have a suspicion this might increase the incentive for a selfish mine. Especially from the monero side as you know that if your difficulty is higher, it will be favoured more. 

I would put height as the fist tie breaker, then monero then blake. Height should break most ties and favour the chain that moves the block_chain to most forward in terms of mined.

SWvheerden (commented):
Looks good. One or two small things. 

I like the direction of this and I think this will improve things going forward. 
But, I made a comment about improving this in the review. But I think this might increase re-orgs. While a re-org in itself is not bad, we should not actively encourage or provide behaviour to increase re-orgs. Thats also the reason we went from a 1 min to a 2 min block time. To decrease re-orgs. 

This code as it stands, is an improvement to the current code, but I think we can do better. 
I made a comment on how I think we should improve it. But I think we can improve it better by:
Comparing the accumulated geometric mean. 
If this is equal, we look at the height, and the previous block hash. 
We then only accept the new tip, if its height is greater than the other tip, and its child of the other tip. Meaning its the same chain. Other wise we dont re-org and we keep our current chain.

SWvheerden:
> We have discussed something similar in the past, using a hash of the header, but could not reach an agreement at the time. This will be simple and easy to verify. To make things a bit fairer, we could use u128 for geometric mean to include more significant digits, instead of u64.

I personally dont like looking at the modulo of the hash or anything related to that. Its overcomplicating the problem. We only need to ensure the chain moves forward. We dont need to find the biggest chain by the smallest bit. As long as the chain keeps moving forward, we will have a eventual winner. 

As to the u128. That will only work if we multiply the difficulties by a factor. We currently have 0 significant bits as we ceil the difficulty. As this is a local only comparison, we can more easily solve this by just keeping it f64. The SQRT equation is already f64.

hansieodendaal:
> 
> As to the u128. That will only work if we multiply the difficulties by a factor. We currently have 0 significant bits as we ceil the difficulty. As this is a local only comparison, we can more easily solve this by just keeping it f64. The SQRT equation is already f64.

I think we should keep it as simple as possible, and I agree @SWvheerden  just using an f64 for geometric mean comparison would solve >99 % of cases where the u64 would have given a tie, as it would give at least 10 predictable significant digits on any architecture.

SWvheerden:
> > As to the u128. That will only work if we multiply the difficulties by a factor. We currently have 0 significant bits as we ceil the difficulty. As this is a local only comparison, we can more easily solve this by just keeping it f64. The SQRT equation is already f64.
> 
> I think we should keep it as simple as possible, and I agree @SWvheerden just using an f64 for geometric mean comparison would solve >99 % of cases where the u64 would have given a tie, as it would give at least 10 predictable significant digits on any architecture.

yip. Cases not solved by this we use as a preference of not reorg the chain. The next block will provide a definite winner. Doesnt matter which comparison we use, if its relatively close on one metric, it will incentives miners to keep their local chain and mine on it. So the next block is always going to decide the clear winner.

base_layer/core/src/consensus/network.rs
```diff
@@ -35,7 +35,7 @@ pub enum Network {
 }
 
 impl Network {
-    pub fn create_consensus_constants(self) -> ConsensusConstants {
+    pub fn create_consensus_constants(&self) -> ConsensusConstants {
```
sdbondi: Clippy would actually complain about this because in this case it's actually cheaper to pass in a copy of the value than a new reference. https://rust-lang.github.io/rust-clippy/master/#trivially_copy_pass_by_ref - super pedantic though.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1861,19 +1862,23 @@ fn find_orphan_chain_tips<T: BlockchainBackend>(db: &T, parent_height: u64, pare
...
     db: &T,
     orphan_chain_tips: Vec<BlockHash>,
-) -> Result<(Difficulty, BlockHash), ChainStorageError>
+    chain_strength_comparer: &dyn ChainStrengthComparer,
+) -> Result<Option<BlockHeader>, ChainStorageError>
 {
-    let mut best_accum_difficulty = Difficulty::min();
-    let mut best_tip_hash: Vec<u8> = vec![0; 32];
+    let mut best_block_header: Option<BlockHeader> = None;
+    // let mut best_tip_hash: Vec<u8> = vec![0; 32];
```
SWvheerden: remove this

base_layer/core/src/proof_of_work/proof_of_work.rs
```diff
@@ -108,13 +108,18 @@ impl ProofOfWork {
     ///
     /// The total accumulated difficulty is most often used to decide on which of two forks is the longest chain.
     pub fn total_accumulated_difficulty(&self) -> Difficulty {
```
SWvheerden, line 116: Is this function still used?

SWvheerden (commented):
Looks good. 
Theirs some code thats just commented out. 
In the main reorg code, you store the hash locally in a var, but you keep hashing it anyway.

base_layer/core/src/proof_of_work/proof_of_work.rs
stringhandler: This still gets stored in the chain metadata. Changing this would cause a reorgs in old blocks

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -166,6 +181,9 @@ impl ConsensusManagerBuilder {
             network: self.network,
             emission,
             gen_block: self.gen_block,
+            chain_strength_comparer: self
+                .chain_strength_comparer
+                .unwrap_or(strongest_chain().by_accumulated_difficulty().build()),
```
sdbondi: `strongest_chain()....` should only run if required.
```suggestion
                .unwrap_or_else(|| strongest_chain().by_accumulated_difficulty().build()),
```

base_layer/core/tests/helpers/nodes.rs
```diff
@@ -190,11 +187,9 @@ impl BaseNodeBuilder {
...
     pub fn start(self, runtime: &mut Runtime, data_path: &str) -> (NodeInterfaces, ConsensusManager) {
         let mmr_cache_config = self.mmr_cache_config.unwrap_or(MmrCacheConfig { rewind_hist_len: 10 });
-        let validators = self.validators.unwrap_or(Validators::new(
-            MockValidator::new(true),
-            MockValidator::new(true),
-            MockAccumDifficultyValidator {},
-        ));
+        let validators = self
+            .validators
+            .unwrap_or(Validators::new(MockValidator::new(true), MockValidator::new(true)));
```
sdbondi, line 192:
```suggestion
            .unwrap_or_else(|| Validators::new(MockValidator::new(true), MockValidator::new(true)));
```

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Agreed, no need to rehash as far as I can see

base_layer/core/src/consensus/chain_strength_comparer.rs
```diff
@@ -0,0 +1,79 @@
...
+        let a_val = a.total_accumulated_difficulty_inclusive_squared();
+        let b_val = b.total_accumulated_difficulty_inclusive_squared();
+        if a_val < b_val {
+            Ordering::Less
+        } else {
+            // f64's can never really be equal, so there is no `cmp` for f64
+            // there are also weird NaN edge cases
+            Ordering::Greater
+        }
+    }
```
sdbondi: It's not true that f64's can never be equal but I understand what you were getting at here. 
PartialOrd (partial_cmp) exists for this case i.e `NaN != NaN`. 

```suggestion
        let a_val = a.total_accumulated_difficulty_inclusive_squared();
        let b_val = b.total_accumulated_difficulty_inclusive_squared();
        a_val.partial_cmp(b_val).unwrap_or(Ordering::Greater)
```

If we explicitly write it out the logic, it may not make complete sense what we are doing.

```rust
if a_val.is_nan() || b_val.is_nan() {
  return Ordering::Greater;
}
```
However, we know that `total_accumulated_difficulty_inclusive_squared`'s implementation can never return NaN.

Given: 

```rust
    let a = u64::MAX as f64;
    let b = u64::MAX as f64;
    let c = a * b;
    assert_eq!(c.is_nan(), false);
    assert_eq!(c.is_infinite(), false);
    assert_eq!(c.is_finite(), true);
    assert_eq!(c.is_normal(), true);
    assert_eq!(c.is_sign_positive(), true);
```
if it ever does return NaN, then the implementation is incorrect.  For these cases, failing loudly (i.e. panic!) is typically the best option (unit tests should be written to catch that case).  

I'm more concerned about FP precision when multiplying two large numbers and comparing them:

```rust
    let a = u64::MAX as f64;
    let b = u64::MAX as f64;
    let c = a * b;
   // This passes! Add one more zero to make it fail
    assert_eq!(c - 10_000_000_000_000_000_000_000.0, c);
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0b25ababfa755fba06b3fe88d4956b09

With more typical numbers
```rust
    let a = 77765317093172601.0f64;
    let b = 77765317093172600.0f64;
    assert_eq!(a*a, a*b); // <--- this passes
```

As the accumulated difficulty grows, the precision becomes worse. I'm honestly not sure how big of an issue this is but I thought I'd mention it.

sdbondi: Might want to use `total_cmp`
https://doc.rust-lang.org/beta/std/primitive.f64.html#method.total_cmp
 since the comparison is part of a floating point standard.

sdbondi (commented):
I still have reservations about the loss of precision when multiplying and comparing acc. difficulties in floating point however the loss of precession should be much less than the current `ceil(sqrt(f64*f64))` implementation.

SWvheerden (dismissed):
LGTM

Issue #2181, opened 2020-08-31
LMDB `mapsize` (environment size) does not increase as the database is filling up.


hansieodendaal:
**Describe the bug**

LMDB `mapsize` (environment size) does not increase as the database is filling up. This results in db write failures when syncing from scratch at ~76,000 block height, when the db size reaches 1GiB. (_#1817 changed the 50GiB initial size down to 1GiB for reasons discussed there._)

**To Reproduce**

Sync from scratch on a blockchain that is larger than 1 GiB, without restarting the basenode.

**Expected behavior**

- LMDB environment size should increase automatically. See related sizing logic in #1817.
- A unit test should be added to test automatic resizing if the limits are approached.

**Screenshots**

n/a

**Desktop (please complete the following information):**

Windows 10, Ubuntu Linux

**Smartphone (please complete the following information):**

n/a

**Additional context**

See related issues

delta1:
closed by #2193

PR #2193, opened 2020-09-04
Implement LMDB automatic resizing

delta1:
## Description

This commit implements the functionality to automatically resize LMDB environment mapsize as it fills up. This should ensure that new data can always be inserted as long as there is disk space available. 

There were a few minor issues fixed: spelling, docs, and removing an unused `use` declaration. 

Thanks to @hansieodendaal for his contributions

## Motivation and Context
Issue #2181 

## How Has This Been Tested?
* [X] Existing tests pass 
* [X] New tests created 
* [X] Tested manually on macOS 
* [ ] Testing on Windows/Linux still required

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/error.rs
+++ b/base_layer/core/src/chain_storage/error.rs
@@ -25,6 +25,7 @@ use crate::{
     validation::ValidationError,
 };
 use tari_mmr::{error::MerkleMountainRangeError, MerkleProofError};
+use tari_storage::lmdb_store::LMDBError;
 use thiserror::Error;
 use tokio::task;
 
@@ -85,6 +86,11 @@ pub enum ChainStorageError {
     OutOfRange,
     #[error("Value not found: {0}")]
     LmdbValueNotFound(lmdb_zero::Error),
+    #[error("LMDB error: {source}")]
+    LmdbError {
+        #[from]
+        source: LMDBError,
+    },
 }
 
 impl From<LMDBVecError> for ChainStorageError {
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
@@ -45,21 +45,19 @@ pub fn serialize<T>(data: &T) -> Result<Vec<u8>, ChainStorageError>
 where T: Serialize {
     let size = bincode::serialized_size(&data).map_err(|e| ChainStorageError::AccessError(e.to_string()))?;
     let mut buf = Vec::with_capacity(size as usize);
-    bincode::serialize_into(&mut buf, data)
-        .or_else(|e| {
-            error!(target: LOG_TARGET, "Could not serialize lmdb: {:?}", e);
-            Err(e)
-        })
-        .map_err(|e| ChainStorageError::AccessError(e.to_string()))?;
+    bincode::serialize_into(&mut buf, data).map_err(|e| {
+        error!(target: LOG_TARGET, "Could not serialize lmdb: {:?}", e);
+        ChainStorageError::AccessError(e.to_string())
+    })?;
     Ok(buf)
 }
 
 pub fn deserialize<T>(buf_bytes: &[u8]) -> Result<T, error::Error>
 where T: DeserializeOwned {
     bincode::deserialize(buf_bytes)
-        .or_else(|e| {
+        .map_err(|e| {
             error!(target: LOG_TARGET, "Could not deserialize lmdb: {:?}", e);
-            Err(e)
+            e
         })
         .map_err(|e| error::Error::ValRejected(e.to_string()))
 }
--- a/infrastructure/storage/src/lmdb_store/store.rs
+++ b/infrastructure/storage/src/lmdb_store/store.rs
@@ -34,22 +34,72 @@ use std::{
 };
 
 const LOG_TARGET: &str = "lmdb";
+const BYTES_PER_MB: usize = 1024 * 1024;
 
 /// An atomic pointer to an LMDB database instance
 type DatabaseRef = Arc<Database<'static>>;
 
+#[derive(Debug, Clone)]
+pub struct LMDBConfig {
+    init_size_bytes: usize,
+    grow_size_bytes: usize,
+    resize_threshold_bytes: usize,
+}
+
+impl LMDBConfig {
+    /// Specify LMDB config in bytes.
+    pub fn new(init_size_bytes: usize, grow_size_bytes: usize, resize_threshold_bytes: usize) -> Self {
+        Self {
+            init_size_bytes,
+            grow_size_bytes,
+            resize_threshold_bytes,
+        }
+    }
+
+    /// Specify LMDB config in megabytes.
+    pub fn new_from_mb(init_size_mb: usize, grow_size_mb: usize, resize_threshold_mb: usize) -> Self {
+        Self {
+            init_size_bytes: init_size_mb * BYTES_PER_MB,
+            grow_size_bytes: grow_size_mb * BYTES_PER_MB,
+            resize_threshold_bytes: resize_threshold_mb * BYTES_PER_MB,
+        }
+    }
+
+    /// Get the initial size of the LMDB environment in bytes.
+    pub fn init_size_bytes(&self) -> usize {
+        self.init_size_bytes
+    }
+
+    /// Get the grow size of the LMDB environment in bytes. The LMDB environment will be resized by this amount when
+    /// `resize_threshold_bytes` are left.
+    pub fn grow_size_bytes(&self) -> usize {
+        self.grow_size_bytes
+    }
+
+    /// Get the resize threshold in bytes. The LMDB environment will be resized when this much free space is left.
+    pub fn resize_threshold_bytes(&self) -> usize {
+        self.resize_threshold_bytes
+    }
+}
+
+impl Default for LMDBConfig {
+    fn default() -> Self {
+        Self::new_from_mb(64, 16, 4)
+    }
+}
+
 /// A builder for [LMDBStore](struct.lmdbstore.html)
 /// ## Example
 ///
-/// Create a new LMDB database of 500MB in the `db` directory with two named databases: "db1" and "db2"
+/// Create a new LMDB database of 64MB in the `db` directory with two named databases: "db1" and "db2"
 ///
 /// ```
-/// # use tari_storage::lmdb_store::LMDBBuilder;
+/// # use tari_storage::lmdb_store::{LMDBBuilder, LMDBConfig};
 /// # use lmdb_zero::db;
 /// # use std::env;
 /// let mut store = LMDBBuilder::new()
 ///     .set_path(env::temp_dir())
-///     .set_environment_size(500)
+///     .set_env_config(LMDBConfig::default())
 ///     .set_max_number_of_databases(10)
 ///     .add_database("db1", db::CREATE)
 ///     .add_database("db2", db::CREATE)
@@ -59,9 +109,9 @@ type DatabaseRef = Arc<Database<'static>>;
 #[derive(Default)]
 pub struct LMDBBuilder {
     path: PathBuf,
-    db_size_mb: usize,
     max_dbs: usize,
     db_names: HashMap<String, db::Flags>,
+    env_config: LMDBConfig,
 }
 
 impl LMDBBuilder {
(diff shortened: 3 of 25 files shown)


infrastructure/storage/src/lmdb_store/store.rs
```diff
@@ -359,11 +368,61 @@ impl LMDBStore {
...
+            previous_size,
+            new_size_bytes,
+            new_size_bytes - previous_size
+        );
+
+        Ok(())
+    }
+
+    /// Returns a result tuple indicating if the LMDB Environment should be resized, along with the new size in bytes.
+    /// Currently uses a contant resize threshold.
```
hansieodendaal:
```suggestion
    /// Currently uses a constant resize threshold.
```

delta1: Thanks Hansie nice catch - fixed

infrastructure/storage/src/lmdb_store/store.rs
```diff
@@ -359,11 +368,61 @@ impl LMDBStore {
...
+            target: LOG_TARGET,
+            "LMDB mapsize: {}, bytes used: {}, bytes remaining: {}, used: {:.*}, threshold: {}, resize required: {}.",
+            env_info.mapsize,
+            size_used_bytes,
+            env_info.mapsize - size_used_bytes,
+            4, // decimal places for "used"
+            size_used_bytes as f64 / env_info.mapsize as f64,
+            RESIZE_THRESHOLD,
+            needs_resize.to_string()
+        );
```
hansieodendaal, line 416:
```suggestion
        let needs_resize = should_resize(size_used_bytes, env_info.mapsize, RESIZE_THRESHOLD);

        let new_size = match needs_resize {
            true => {
                debug!(
                    target: LOG_TARGET,
                    "LMDB mapsize resize required: {}, bytes used: {}, bytes remaining: {}, used: {:.*}, threshold: \
                     {}.",
                    env_info.mapsize,
                    size_used_bytes,
                    env_info.mapsize - size_used_bytes,
                    4, // decimal places for "used"
                    size_used_bytes as f64 / env_info.mapsize as f64,
                    RESIZE_THRESHOLD,
                );
                size_used_bytes + init_size_bytes
            },
            false => init_size_bytes,
        };
```

infrastructure/storage/tests/lmdb.rs
```diff
@@ -330,3 +330,47 @@ fn lmbd_resize_on_create() {
...
+                }
+            }
+            let env_info = store.env().info().unwrap();
+            let psize = store.env().stat().unwrap().psize as usize;
+            let page_size_total = psize * env_info.last_pgno;
+            let percent_left = 1.0 - page_size_total as f64 / env_info.mapsize as f64;
+
+            // check the allocated size is now greater than it was initially
+            assert!(page_size_total > 1024 * 1024);
+            assert!(percent_left > 0.0);
```
hansieodendaal, line 374: I think in this instance we should also assert that `space left <= (1 + RESIZE_THRESHOLD ) * initial capacity`, to ensure it does not grow out of bounds.

hansieodendaal (commented):
Nice one! 

- My one concern is that I think it will be overkill to call the trace log in `pub fn resize_required` for every db write operation, which will happen as currently implemented. See proposed.
- Also added a test suggestion to check that the db does not grow out of bounds.

Regards

infrastructure/storage/src/lmdb_store/store.rs
delta1: Thanks for the suggestion Hansie - I'll certainly take a look at this. It shouldn't return `init_size_bytes` though - as the env size may have already grown beyond that and so this would actually try to decrease the mapsize.

infrastructure/storage/tests/lmdb.rs
delta1: Thanks I'll take a look at that

hansieodendaal: I will post new suggestions; ran a 'sync from scratch' test and the resize logic has some problem.

hansieodendaal (commented):
See https://github.com/delta1/tari/pull/1 for updated review suggestions; I have redone my previous suggestions.

delta1:
Thanks @hansieodendaal - I'll merge that and test it out! Appreciate your assistance

infrastructure/storage/src/lmdb_store/store.rs
```diff
@@ -359,11 +352,40 @@ impl LMDBStore {
...
+    pub fn get_init_size(&self) -> usize {
+        self.env_init_size_bytes
+    }
+
+    /// Resize the LMDB environment if the resize threshold is breached.
+    pub fn resize_if_required(env: &Environment, init_size_bytes: usize) -> Result<(), LMDBError> {
+        let env_info = env.info()?;
+        let size_used_bytes = env.stat()?.psize as usize * env_info.last_pgno;
+        let needs_resize = env_info.mapsize - size_used_bytes <= (init_size_bytes as f64 * RESIZE_THRESHOLD) as usize;
+        let path = env.path().unwrap().to_str().unwrap();
```
hansieodendaal:
```suggestion
```

infrastructure/storage/src/lmdb_store/store.rs
```diff
@@ -359,11 +352,40 @@ impl LMDBStore {
...
+        if needs_resize {
+            unsafe {
+                env.set_mapsize(size_used_bytes + init_size_bytes)?;
+            }
+            debug!(
+                target: LOG_TARGET,
+                "({}) LMDB size used {:?} MB, environment space left {:?} MB, increased by {:?} MB.",
+                path,
```
hansieodendaal:
```suggestion
                env.path().unwrap().to_str().unwrap(),
```

infrastructure/storage/src/lmdb_store/store.rs
hansieodendaal: Realized this is an unnecessary call each time, could be simplified

infrastructure/storage/src/lmdb_store/store.rs
hansieodendaal: As per above

base_layer/core/src/chain_storage/error.rs
```diff
@@ -85,6 +86,11 @@ pub enum ChainStorageError {
     OutOfRange,
     #[error("Value not found: {0}")]
     LmdbValueNotFound(lmdb_zero::Error),
+    #[error("LMDB error: {source}")]
+    LMDBError {
```
stringhandler:
```suggestion
    LmdbError {
```

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
```diff
@@ -46,9 +46,9 @@ where T: Serialize {
     let size = bincode::serialized_size(&data).map_err(|e| ChainStorageError::AccessError(e.to_string()))?;
     let mut buf = Vec::with_capacity(size as usize);
     bincode::serialize_into(&mut buf, data)
-        .or_else(|e| {
+        .map_err(|e| {
             error!(target: LOG_TARGET, "Could not serialize lmdb: {:?}", e);
```
stringhandler: nit: Could probably just move the log into the `map_err()` below, instead of having 2 `map_err`s

stringhandler (commented):
One or two nits, but overall happy with how you've done it. I think that RESIZE_THRESHOLD should be moved to the config, with a default value as well as initial size and a new variable for size to grow by. This allows users to control disk usage. I think that RESIZE_THRESHOLD should also be a number in MB and not a percentage. The reason for that is the DB is always growing by a constant size which is `max_block_size * block_time`, so we could use the resize threshold to always add enough space for `x` number of blocks, rather than keeping 50% for growth.
I suggest `init_db_size = 1GB`, `grow_db_size = 500Mb`, `resize_threshold=100Mb`

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
delta1: Good idea - will fix

hansieodendaal:
> One or two nits, but overall happy with how you've done it. I think that RESIZE_THRESHOLD should be moved to the config, with a default value as well as initial size and a new variable for size to grow by. This allows users to control disk usage. I think that RESIZE_THRESHOLD should also be a number in MB and not a percentage. The reason for that is the DB is always growing by a constant size which is `max_block_size * block_time`, so we could use the resize threshold to always add enough space for `x` number of blocks, rather than keeping 50% for growth.
> I suggest `init_db_size = 1GB`, `grow_db_size = 500Mb`, `resize_threshold=100Mb`

@mikethetike and @delta1, I do not think these should be user-configurable, or to change it as suggested, as it would then be prone to errors, overly complicating the code to guard against that. Currently, we have 3 LMDB databases, each with their own size. We could just choose settings that would be small enough to preserve disk space. For example, the blockchain database could be 250MB (currently 1000MB), the peer database and wallet peer database could each be 10 MB (currently 50 MB), all with with RESIZE_THRESHOLD at 50%, increasing by init size.

stringhandler:
@hansieodendaal all user configs are prone to error. To be fair, the two peer_db databases don't need to be very configurable, but the main block chain database should be. We can guard against configuration values that don't make sense simply by checking the values provided (e.g. `resize_threshold` > `grow_db_size`) and throw a `ConfigurationError`. One user may decide to allocate the full DB size before syncing, and another may prefer to only grow the db by 1mb at a time, keeping it as small as possible. A user also may not have enough space to grow the db to its full extent (e.g. if we set it to 10gb), but would like to run it for a few more blocks before moving to different hardware.

hansieodendaal:
In this case, maybe we can implement a minimum init and grow size to be something like `min_size = 10 x largest block size`
Also `resize_threshold < grow_db_size` and `resize_threshold < min_size`

stringhandler:
Personally I like the option of being able to set `grow_size` (or whatever we call it) to `0`, which should error if it ever tries to resize and must be attended to manually. I also think that `resize_threshold` should have no minimum to give the user the most control. They may want to resize on every block. As long as they could not put in a value that would result in an infinite loop, or other crash, I don't think we need to restrict the YOLO users. We should however, choose responsible defaults that don't affect performance but also save on disk space

hansieodendaal:
> Personally I like the option of being able to set `grow_size` (or whatever we call it) to `0`, which should error if it ever tries to resize and must be attended to manually. 

Do not think this is wise; will be a user support nightmare. Do not invite what can go wrong. The grow size must be safely larger than the size of any write operation.

> I also think that `resize_threshold` should have no minimum to give the user the most control. 

We must guard against writing a block to the database without enough space left, similar to the above comment, so there must be a minimum value.

delta1:
Personally I think these should be user configurable, with sensible defaults, and guards in place when reading them from config. We can hash out the finer details.

delta1:
The initial 1000 MB gets us to approx 73000 blocks which is about 13 to 14 KB blocks on average. 

With a maximum block size of 1 MB, and 720 blocks per day, it's *possible* that the database might require an extra 720 MB per day (although likely to be lower than that). 

So I agree with @mikethetike for *defaults* - start at 1000 MB, resize when there is 100 MB left, and increase the db size by 500 MB. 

When reading in the configuration values, I suggest to check that the following are true: 
- initial size must be at least 100 MB
- growth size must be at least 10 MB 
- resize threshold (resize with this much space left) must be at least 1 MB, and must be **less than** growth size. ~50% of the initial db size.~ 

@mikethetike @hansieodendaal can we please finalise these figures for implementation?

stringhandler:
Agree with those figures, just not sure about the reason for this one: 

> and must be less than 50% of the initial db size.

I think resize threshold should be smaller than growth size otherwise it will constantly be resizing

delta1:
It was just an arbitrarily chosen limit. Happy with your suggestion - will check that resize threshold is less than growth size.

hansieodendaal:
Just keep in mind we may very well increase the maximum block size, as the current sizing (maximum block weight) was done for 1 minute target block times, which have doubled to 2 minutes. Therefore, 1 MB is too tight and must include a safety margin. My proposal:

- initial size must be at least 100 MB (_as above_)
- growth size must be at least 20 MB
- resize threshold (resize with this much space left) must be at least 10 MB

delta1:
Good point. Implementing as such.

stringhandler (approved):

PR #2194, opened 2020-09-04
Add `is_synced` flag to BaseNodeServiceResponse & update UTXO validator

### LINKED ISSUE #1808: Only sync the wallet's output status from Base nodes that are synced to the chain tip
Once we have the ability to monitor a base node's chain tip and sync status as described in https://github.com/tari-project/tari/issues/1806 the wallet should only check the status of its spendable outputs from base nodes that are synced to the chain tip.

Currently the wallet will immediately ask a base node about the wallet's spendable outputs and if the BN doesn't know about them the wallet will consider them invalid. This could be done in error if the BN being queried is not up to date.

philipr-za:
## Description
This PR adds a Boolean flag to all BaseNodeServiceResponse’s so that the node querying can know if the queried node is synced to the chain tip. The wallet is then updated to abort a UTXO validation operation if this flag is false. This is because if a node is not synced it could incorrectly declare UTXO invalid that it is not yet aware of.

To do this the Base Node Service needed to know whether the Base Node State machine had synced at least once and was currently in the Listening state. To allow the Base Node Service to speak to the Base Node State Machine the state machine had to be refactored into a Service with a Handle that could be registered in the Service Stack.

Additional the base node was updated to not process new blocks or transactions until it had synced at least once. This determination was done using the flag made available via the new BaseNodeStateMachineHandle.

## Motivation and Context
Closes https://github.com/tari-project/tari/issues/1808

## How Has This Been Tested?
Tests are provided that test:
1. That the flag is set in the base node state machine once it has synced
2. That the Output Manager Service UTXO validation is aborted if the `is_synced` flag is false.

A test was not written to check that the `is_synced` flag in BaseNodeServiceResponse is set based on the info provided by the BaseNodeStateMachineHandle because the infrastructure to write this test does not exist and will be a huge project to write. This was however tested by compiling and running on a live node.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


SWvheerden (dismissed):
LGTM

hansieodendaal (commented):
@philipr-za, it seems this is not working entirely as intended.

1. I started up a base node and wallet with many UTXOs, that was lagging the blockchain by 135 blocks. The UTXO validation protocol was started with the base node not synced, and it then came back with `Error completing UTXO Validation Protocol (Id: 2253227295055671284): BaseNodeNotSynced`. The UTXO validation protocol was never attempted again after sync was achieved.

2. Many of these in the log files: 
   ```
   [tari_broadcast_channel::async_channel] DEBUG Subscriber with ID '40012' has 231 buffer messages dropped.
   ```
   Created  in `base_layer\core\src\base_node\state_machine_service\initializer.rs`:
   ```
   let (status_event_publisher, status_event_subscriber): (Publisher<_>, Subscriber<_>) = bounded(1, 4);
   ```

See log files attached:
1. [base_layer.log](https://github.com/tari-project/tari/files/5182640/base_layer.log)
2. [other.log](https://github.com/tari-project/tari/files/5182642/other.log)

philipr-za:
In responses to point 1 that is as intended. If the wallet attempts a UTXO validation and the base node is not synced then the process is aborted. Restarting the process is not the focus of this PR because it is non-trivial. The wallet does not have access to the Base Node's event streams so we cannot just restart once the Base Node is synced. Currently the wallet will do the validation again on a restart or when a coinbase is mined.

philipr-za:
In response to point 2:
This is actually a result of the way this pubsub channel is being used. It has a buffer size of 1 by design so that when a new status is publisher the old one is dropped. Previously we didn't get these debug messages because there was only 1 subscriber at all times but with the refactor we allow multiple subscribers being cloned from the one held by the Handle. So I would argue that either this debug message should be trace OR we need to change this channel into a Tokio Watch channel rather than using the tari_broadcast_channel in this weird way. @mikethetike thoughts?

hansieodendaal:
> In responses to point 1 that is as intended. If the wallet attempts a UTXO validation and the base node is not synced then the process is aborted. Restarting the process is not the focus of this PR because it is non-trivial. The wallet does not have access to the Base Node's event streams so we cannot just restart once the Base Node is synced. Currently the wallet will do the validation again on a restart or when a coinbase is mined.

I think the problem here would be that the UTXO query will never succeed if the base node is lagging even a couple of blocks and the base node is not mining. So a next PR will fix that? 

Just a thought, maybe the process could be prevented from starting by polling the initial syn status inside `fn validate_outputs(` or the functions that call it?

philipr-za:
> > In responses to point 1 that is as intended. If the wallet attempts a UTXO validation and the base node is not synced then the process is aborted. Restarting the process is not the focus of this PR because it is non-trivial. The wallet does not have access to the Base Node's event streams so we cannot just restart once the Base Node is synced. Currently the wallet will do the validation again on a restart or when a coinbase is mined.
> 
> I think the problem here would be that the UTXO query will never succeed if the base node is lagging even a couple of blocks and the base node is not mining. So a next PR will fix that?
> 
> Just a thought, maybe the process could be prevented from starting by polling the initial syn status inside `fn validate_outputs(` or the functions that call it?

We definitely don't want to be polling from the wallet if we can help it, we already have a major battery usage issue. 

If we feel we need a more intelligent way to decide to restart a UTXO validation then we should make an issue and address it in a subsequent PR.

Validating the UTXO's is not something that is super critical for a mobile wallet in my opinion because it is very rare that is will be an issue for them and they restart fairly often. For mining wallets it is an issue but they already have a heuristic that will revalidate UTXOs every time a new coinbase is mined.

hansieodendaal:
> In response to point 2:
> This is actually a result of the way this pubsub channel is being used. It has a buffer size of 1 by design so that when a new status is publisher the old one is dropped. Previously we didn't get these debug messages because there was only 1 subscriber at all times but with the refactor we allow multiple subscribers being cloned from the one held by the Handle. So I would argue that either this debug message should be trace OR we need to change this channel into a Tokio Watch channel rather than using the tari_broadcast_channel in this weird way. @mikethetike thoughts?

Cool, no untoward problem then. So I would vote to use the Tokio infrastructure, as changing the debug message level will not really change anything, it might just obscure the other channels we might want to monitor for dropped messages; we are still using 4 of those.

hansieodendaal:
> Validating the UTXO's is not something that is super critical for a mobile wallet in my opinion because it is very rare that is will be an issue for them and they restart fairly often. For mining wallets it is an issue but they already have a heuristic that will revalidate UTXOs every time a new coinbase is mined.

Point taken about the mobile wallets' battery life; my comments were actually directed at a base node wallet attached to a base node (that is not mining). So we would need an issue for that, as it is not really going to work as currently implemented.

philipr-za:
> > Validating the UTXO's is not something that is super critical for a mobile wallet in my opinion because it is very rare that is will be an issue for them and they restart fairly often. For mining wallets it is an issue but they already have a heuristic that will revalidate UTXOs every time a new coinbase is mined.
> 
> Point taken about the mobile wallets' battery life; my comments were actually directed at a base node wallet attached to a base node (that is not mining). So we would need an issue for that, as it is not really going to work as currently implemented.

The wallet is in the process of being separated from the base node and this issue shows why that is important. That is not to say that we shouldn't have a retry strategy but just that it is beyond the scope of this PR.

philipr-za:
> > In response to point 2:
> > This is actually a result of the way this pubsub channel is being used. It has a buffer size of 1 by design so that when a new status is publisher the old one is dropped. Previously we didn't get these debug messages because there was only 1 subscriber at all times but with the refactor we allow multiple subscribers being cloned from the one held by the Handle. So I would argue that either this debug message should be trace OR we need to change this channel into a Tokio Watch channel rather than using the tari_broadcast_channel in this weird way. @mikethetike thoughts?
> 
> Cool, no untoward problem then. So I would vote to use the Tokio infrastructure, as changing the debug message level will not really change anything, it might just obscure the other channels we might want to monitor for dropped messages; we are still using 4 of those.

Alright, I have migrated the tari_broadcast_channel to a tokio watch channel.

base_layer/core/src/base_node/state_machine_service/handle.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn get_status_event_stream_fused(&self) -> Fuse<watch::Receiver<StatusInfo>> {
+        self.get_status_event_stream().fuse()
+    }
+
+    pub fn shutdown_signal(&self) -> ShutdownSignal {
+        self.shutdown_signal.clone()
+    }
+
+    /// This lets any interested party update the the initial blockchain sync status
+    pub fn set_initial_blockchain_sync_success_status(&self, status: bool) {
```
stringhandler: I'm not sure about this, but I'll have to see where it's used. Seems like this field should only ever be readonly and determined by the base node

base_layer/core/src/base_node/state_machine_service/state_machine.rs
```diff
@@ -100,26 +106,28 @@ impl<B: BlockchainBackend + 'static> BaseNodeStateMachine<B> {
         metadata_event_stream: Subscriber<ChainMetadataEvent>,
         config: BaseNodeStateMachineConfig,
         sync_validators: SyncValidators,
-        shutdown_signal: ShutdownSignal,
+        interrupt_signal: ShutdownSignal,
+        shutdown_trigger: Shutdown,
```
stringhandler: What is the difference between `interrupt_signal` and `shutdown_trigger`. They sound really similar

base_layer/core/src/base_node/state_machine_service/handle.rs
```diff
@@ -0,0 +1,90 @@
...
+    status_event_receiver: watch::Receiver<StatusInfo>,
+    shutdown_signal: ShutdownSignal,
+    initial_blockchain_sync_success: Arc<AtomicBool>,
+}
+
+impl StateMachineHandle {
+    pub fn new(
+        state_change_event_subscriber: Subscriber<StateEvent>,
+        status_event_receiver: watch::Receiver<StatusInfo>,
+        initial_blockchain_sync_success: Arc<AtomicBool>,
```
stringhandler: The `AtomicBool` seems out of place here. Is it not better to use channels similar to the rest of the struct?

base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -86,6 +91,11 @@ impl Listening {
                         let sync_peers = select_sync_peers(local_tip_height, &best_metadata, &peer_metadata_list);
 
                         let sync_mode = determine_sync_mode(&local, best_metadata, sync_peers);
+                        if sync_mode == SyncStatus::UpToDate &&
```
stringhandler: I don't think this step is necessary. The base node being in `Listening` state is IMHO sufficient to determine that the base node is synced. If `sync_mode.is_lagging()` is true in the next statement, the state will transition out of `Listening`

base_layer/core/src/base_node/service/service.rs
```diff
@@ -412,16 +428,30 @@ async fn handle_incoming_request<B: BlockchainBackend + 'static>(
...
+    let mut status_event_stream = state_machine_handle.get_status_event_stream();
+
+    let in_listening_state = match status_event_stream.recv().await {
+        None => false,
+        Some(s) => match s {
+            StatusInfo::Listening(_) => true,
+            _ => false,
+        },
+    };
+    let is_synced = in_listening_state && state_machine_handle.get_initial_blockchain_sync_success_status();
```
stringhandler: As I mention later, this extra check doesn't seem necessary. Simply being in `Listening` is probably enough of a check. In fact, I'm not sure if you could ever logically have `in_listening_state == true` and `get_initial_blockchain_sync_success_status()` be false. Unless I am missing another place where is it set outside of the `Listening` state.

stringhandler (commented):
Some nice refactorings here and clean up. I haven't read the other review comments yet, but I'm not a fan of the `get_initial_blockchain_sync_success_status` `AtomicBool` approach. I think that it is sufficient to check if the base node is in `Listening` state and it looks like if it is not, the flag result will be false. Obviously this is up for discussion, but I'm in favour of removing `get(set)_initial_blockchain_sync_success_status` entirely

stringhandler:
I like the use of `tokio::sync::watch`. Regarding the UTXO validation, without knowing the exact details, I would argue for this process:
1. If the UTXO would be deleted or detected as reorged away, it should check `is_synced` and only delete it if it's true
2. If the UTXO was unspent but now detected as spent, `is_synced` is ignored and it is marked spent regardless. A future reorg should resolve this. This should be the majority of cases
3. If the UTXO was spent and is now detected as unspent, `is_synced` must be true before changing

I'm not sure if that introduces any vulnerabilities, but that should handle most cases

base_layer/core/src/base_node/state_machine_service/handle.rs
philipr-za: I agree with this, was intended to be only read-only in the handle. Will remove.

base_layer/core/src/base_node/state_machine_service/state_machine.rs
philipr-za: The interrupt signal is an external signal that tells the state machine it should shutdown. The Shutdown Signal is how the state machine indicates to the Application that it has shutdown. Suggestions for a clearer name?

base_layer/core/src/base_node/state_machine_service/handle.rs
philipr-za: I disagree on this one. This is an event that occurs exactly once and is not a stream of changes so I believe a read-only AtomicBool is appropriate.

base_layer/core/src/base_node/state_machine_service/states/listening.rs
philipr-za: Unfortunately it is because the State Machine starts in StartUp transitions to Listening and then only transitions to Lagging after a full round of ChainMetadata pings has been received from the neighbourhood peers. This means that for a short but not insignificant time the State Machine is in Listening state but not synced.

base_layer/core/src/base_node/service/service.rs
philipr-za: Explained above why we need both these checks. After the initial sync has been completed we still need to check if we are in listening in case the node has become out of sync after the initial sync.

philipr-za:
> I like the use of `tokio::sync::watch`. Regarding the UTXO validation, without knowing the exact details, I would argue for this process:
> 
> 1. If the UTXO would be deleted or detected as reorged away, it should check `is_synced` and only delete it if it's true
> 2. If the UTXO was unspent but now detected as spent, `is_synced` is ignored and it is marked spent regardless. A future reorg should resolve this. This should be the majority of cases
> 3. If the UTXO was spent and is now detected as unspent, `is_synced` must be true before changing
> 
> I'm not sure if that introduces any vulnerabilities, but that should handle most cases

Just for clarity, currently the wallet Output Manager Service only validates the first point because we have yet to implement point 3. It is on the cards but until we have point 3 we cannot implement point 2 in the way you describe but I will make an issue for point 2 and 3.

base_layer/core/src/base_node/state_machine_service/handle.rs
hansieodendaal: Agree with @philipr-za

base_layer/core/src/base_node/state_machine_service/states/listening.rs
hansieodendaal: Agreed with @philipr-za

base_layer/core/src/base_node/service/service.rs
hansieodendaal: Agreed with @philipr-za. When the node starts up, `in_listening_state == true` and `get_initial_blockchain_sync_success_status() == false`. Also true for a fresh node starting up for the first time.

hansieodendaal:
> > I like the use of `tokio::sync::watch`. Regarding the UTXO validation, without knowing the exact details, I would argue for this process:
> > 
> > 1. If the UTXO would be deleted or detected as reorged away, it should check `is_synced` and only delete it if it's true
> > 2. If the UTXO was unspent but now detected as spent, `is_synced` is ignored and it is marked spent regardless. A future reorg should resolve this. This should be the majority of cases
> > 3. If the UTXO was spent and is now detected as unspent, `is_synced` must be true before changing
> > 
> > I'm not sure if that introduces any vulnerabilities, but that should handle most cases
> 
> Just for clarity, currently the wallet Output Manager Service only validates the first point because we have yet to implement point 3. It is on the cards but until we have point 3 we cannot implement point 2 in the way you describe but I will make an issue for point 2 and 3.

As a side note, something I do manually with a base node wallet to clean it up is to change the status of all UTXOs to unspent, start up the base node, and then rely on the UTXO query to invalidate all UTXOs that are not unspent. I know this method is not elegant, but it is effective.

hansieodendaal:
Currently, the UTXO Validation protocol will return `UTXO Validation protocol has ended with result Ok(Err(OutputManagerProtocolError { id: 4724786466859331499, error: BaseNodeNotSynced }))` if a bade node is not synced, but only after the `fetch_utxo` request has been handled in the base node. I think the base node should return an error upon a `fetch_utxo` request if the initial sync has not been done.

philipr-za:
> Currently, the UTXO Validation protocol will return `UTXO Validation protocol has ended with result Ok(Err(OutputManagerProtocolError { id: 4724786466859331499, error: BaseNodeNotSynced }))` if a bade node is not synced, but only after the `fetch_utxo` request has been handled in the base node. I think the base node should return an error upon a `fetch_utxo` request if the initial sync has not been done.

I don't understand the issue here? The `fetch_utxo` is the only interaction the Output Manager has with the base node so it is the only time that it finds out that the base node is not synced?

hansieodendaal:
> I don't understand the issue here? The `fetch_utxo` is the only interaction the Output Manager has with the base node so it is the only time that it finds out that the base node is not synced?

The base node handles the fetch UTXO request and returns a result irrespective; I think it should just return an error if initial sync has not been done, seeing that the Output Manager ignores the result that are returned anyway.

philipr-za:
> > I don't understand the issue here? The `fetch_utxo` is the only interaction the Output Manager has with the base node so it is the only time that it finds out that the base node is not synced?
> 
> The base node handles the fetch UTXO request and returns a result irrespective; I think it should just return an error if initial sync has not been done, seeing that the Output Manager ignores the result that are returned anyway.

There are no errors in the comms level Base Node Responses and there are also going to be cases where we are not concerned if the base node is synced or not. We can make that determination based on the `is_synced` flag.

hansieodendaal:
> > > I don't understand the issue here? The `fetch_utxo` is the only interaction the Output Manager has with the base node so it is the only time that it finds out that the base node is not synced?
> > 
> > 
> > The base node handles the fetch UTXO request and returns a result irrespective; I think it should just return an error if initial sync has not been done, seeing that the Output Manager ignores the result that are returned anyway.
> 
> There are no errors in the comms level Base Node Responses and there are also going to be cases where we are not concerned if the base node is synced or not. We can make that determination based on the `is_synced` flag.

Ok agreed, this is something for a separate issue, not related to this PR.

hansieodendaal:
LGTM

base_layer/core/src/base_node/state_machine_service/state_machine.rs
stringhandler: Perhaps `should_shutdown` and `on_shutdown`?

philipr-za: changed it to `service_shutdown`.

base_layer/core/src/base_node/state_machine_service/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pub fn get_state_change_event_stream_fused(&self) -> Fuse<Subscriber<StateEvent>> {
+        self.get_state_change_event_stream().fuse()
+    }
+
+    /// This clones the receiver end of the channel and gives out a copy to the caller
+    /// This allows multiple subscribers to this channel by only keeping one channel and cloning the receiver for every
+    /// caller.
+    pub fn get_status_event_stream(&self) -> watch::Receiver<StatusInfo> {
```
stringhandler:
```suggestion
    pub fn get_latest_status_info(&self) -> watch::Receiver<StatusInfo> {
```

base_layer/core/src/base_node/state_machine_service/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.status_event_receiver.clone()
+    }
+
+    pub fn get_status_event_stream_fused(&self) -> Fuse<watch::Receiver<StatusInfo>> {
```
stringhandler: Is there a use case for not using the fused stream? Perhaps we should only expose the fused stream

base_layer/core/src/base_node/state_machine_service/handle.rs
stringhandler: Or perhaps even `get_current_status_info()`

stringhandler (commented):
So the more I think about it, I really think that the sync state should be determined from the current state of the state machine. It greatly simplifies things and prevents future mistakes that contributors might make if they had to make two calls to determine it.

Listening is the "everything is fine" or normal state that the node is in, and at that point it can process. Unfortunately we may be in this state and are actually not able to process normally. Conceptually, there should be an `Uncertain` state where we are listening for metadata and an `UpToDate` state where we are listening, but we would end up with a lot of duplicate code. To simplify things, I think let's add some data to the `ListeningInfo` state_info struct about why we are in a listening state. 

Example:
```rust
/// This struct contains info that is use full for external viewing of state info
pub struct ListeningInfo{
    is_synced: bool
}
```

Here is the state transition:
```rust
 match (state, event) {
            (Starting(s), Initialized) => Listening(s.into()),
            (HeaderSync(s), HeadersSynchronized(local, sync_height)) => HorizonStateSync(
                states::HorizonStateSync::new(local, s.network_metadata, s.sync_peers, sync_height),
            ),
            (HeaderSync(s), HeaderSyncFailure) => Waiting(s.into()),
            // TODO: Simplify block sync and implement From<HorizonStateSync>
            (HorizonStateSync(s), HorizonStateSynchronized) => BlockSync(
                self.config.block_sync_config.sync_strategy,
                s.network_metadata,
                s.sync_peers,
            ),
            (HorizonStateSync(s), HorizonStateSyncFailure) => Waiting(s.into()),
            (BlockSync(s, _, _), BlocksSynchronized) => Listening(s.into()),
            (BlockSync(s, _, _), BlockSyncFailure) => Waiting(s.into()),
            (Listening(_), FallenBehind(Lagging(network_tip, sync_peers))) => {
                BlockSync(self.config.block_sync_config.sync_strategy, network_tip, sync_peers)
            },
            (Listening(_), FallenBehind(LaggingBehindHorizon(network_tip, sync_peers))) => {
                HeaderSync(states::HeaderSync::new(network_tip, sync_peers))
            },
            (Waiting(s), Continue) => Listening(s.into()),
            (_, FatalError(s)) => Shutdown(states::Shutdown::with_reason(s)),
            (_, UserQuit) => Shutdown(states::Shutdown::with_reason("Shutdown initiated by user".to_string())),
            (s, e) => {
                warn!(
                    target: LOG_TARGET,
                    "No state transition occurs for event {:?} in state {}", e, s
                );
                s
            },
        }
```


If we have entered `Listening` from `Starting` or from `Waiting`, we are in a Listening, but not synched state (`is_synced == false`, but if we enter from  `BlocksSynchronized`, we are in a Listening and synched state (`is_synced ==true`).

Inside of the `Listening` state, if we are found to be lagging, we change to a lagging state, but otherwise, we can move to a Listening and synched state:

```rust
 let sync_mode = determine_sync_mode(&local, best_metadata, sync_peers);
                        if sync_mode.is_lagging() {
                            debug!(target: LOG_TARGET, "{}", sync_mode);
                            return StateEvent::FallenBehind(sync_mode);
                        }
                        else {
                            shared
                                .set_status_info(StatusInfo::Listening(ListeningInfo::new(true)))
                                .await;
                        }
```

This way any component in the system can easily determine the state of the system by watching the `latest_state()` of the `watch` on the `StateMachineHandle`

hansieodendaal:
Maybe these refactoring suggestions should be taken to a consecutive PR?

base_layer/core/src/base_node/state_machine_service/handle.rs
philipr-za: In tests you don't want it fused but in places with a select you do.

base_layer/core/src/base_node/state_machine_service/handle.rs
philipr-za: How about `get_status_info_watch()`? Reasoning is that this doesn't actually return the current status info but a watch channel.

philipr-za:
As discussed I have updated the state machine to hold the state of if the initial bootstrap sync has completed internally and it is presented to clients within the Listening state's ListeningInfo via the state machines current status watch channel.

hansieodendaal:
Thanks @philipr-za

hansieodendaal:
Looking good

stringhandler (approved):
Awesome! Thanks

PR #2196, opened 2020-09-07
[Merge Mining] Select lowest difficulty

StriderDM:
## Description
Select lowest difficulty between monero and tari to be passed to XMRig.
When submitting a block, if difficulty is lower than current monero difficulty, submit only to Tari, otherwise submit to  both tari and monero.

Merge https://github.com/tari-project/tari/pull/2176 first.

## Motivation and Context
Allows blocks to be mined at the lowest difficulty between parent and child chain.

## How Has This Been Tested?
cargo test --all

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/main.rs
+++ b/applications/tari_merge_mining_proxy/src/main.rs
@@ -26,7 +26,6 @@ extern crate arrayref;
 extern crate chrono;
 extern crate jsonrpc;
 use chrono::Local;
-use config;
 use curl::easy::{Auth, Easy, List};
 use structopt::StructOpt;
 use tari_common::{ConfigBootstrap, ConfigError, GlobalConfig};
@@ -43,6 +42,7 @@ use monero::{
 use rand::rngs::OsRng;
 use regex::Regex;
 use std::{
+    cmp::min,
     convert::TryFrom,
     io::{prelude::*, Read},
     net::{TcpListener, TcpStream},
@@ -74,8 +74,23 @@ pub const LOG_TARGET: &str = "tari_mm_proxy::app";
 pub struct TransientData {
     tari_block: Option<grpc::GetNewBlockResult>,
     monero_seed: Option<String>,
+    monero_difficulty: Option<u64>,
     tari_height: Option<u64>,
+    tari_difficulty: Option<u64>,
     tari_prev_submit_height: Option<u64>,
+}
+
+impl Default for TransientData {
+    fn default() -> Self {
+        Self {
+            tari_block: None,
+            monero_seed: None,
+            monero_difficulty: None,
+            tari_height: None,
+            tari_difficulty: None,
+            tari_prev_submit_height: None,
+        }
+    }
 }
 
 // setup curl authentication
@@ -328,6 +343,28 @@ fn get_seed_hash(data: &[u8]) -> Option<String> {
     }
 }
 
+fn get_monero_difficulty(data: &[u8]) -> Option<u64> {
+    debug!(target: LOG_TARGET, "Data {}", &stringify_request(data));
+    let parsed = json::parse(&stringify_request(data));
+    match parsed {
+        Ok(parsed) => {
+            if parsed["result"]["difficulty"].is_null() {
+                return None;
+            }
+            let difficulty = &parsed["result"]["difficulty"].as_u64();
+            difficulty.to_owned()
+        },
+        Err(e) => {
+            error!(
+                target: LOG_TARGET,
+                "{:#}",
+                MmProxyError::ParseError(format!("Failure to parse json to get monero difficulty, {:?}", e))
+            );
+            None
+        },
+    }
+}
+
 // TODO: Temporary till RPC call is in place
 fn add_coinbase(consensus: ConsensusManager, grpc_block: grpc::NewBlockTemplate) -> Option<grpc::NewBlockTemplate> {
     let block = NewBlockTemplate::try_from(grpc_block);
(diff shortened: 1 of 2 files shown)


applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -376,6 +424,50 @@ fn get_spending_key() -> (PrivateKey, PrivateKey) {
...
+        Ok(mut parsed) => {
+            if parsed["result"]["difficulty"].is_null() {
+                error!(
+                    target: LOG_TARGET,
+                    "{:#}",
+                    MmProxyError::ParseError("Monero response invalid, cannot determine difficulty to use".to_string())
+                );
+                return data.to_vec();
+            }
+            let monero_difficulty = &parsed["result"]["difficulty"].as_u64();
```
SWvheerden: Would it not be better to use the function above that already parses the diff as a u64?
It seems that the logic is double here. Or am I missing something here?

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -605,7 +697,17 @@ fn handle_post(
...
+                        transient.tari_block = None;
+                        transient.monero_seed = None;
+                        transient.monero_difficulty = None;
+                        transient.tari_difficulty = None;
                     } else {
+                        // Failure here means XMRig wont submit since it already succeeded to monero
+                        transient.tari_block = None;
+                        transient.monero_seed = None;
+                        transient.tari_difficulty = None;
+                        transient.monero_difficulty = None;
```
SWvheerden: Why not put this in front of the if?Then you only need to have this once

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -724,40 +831,53 @@ fn handle_connection(
...
-                            debug!(target: LOG_TARGET, "Handling POST Method: {}", method);
-                            debug!(
-                                target: LOG_TARGET,
-                                "Response: {:#}",
-                                String::from_utf8_lossy(&data).to_string()
-                            );
-                            match handle_post(&method, &json, data.clone(), transient, consensus, rt, config) {
-                                Ok(result) => {
-                                    data = result;
+            if let "submitblock" = method.as_str() {
```
SWvheerden, line 814: We need to handle the 2 cases depending on the achieved diff. 
We currently target the lower one, which is correct. But we could have found one that hit both. 

for example. Tari is the lower on at 100K, while monero is 200K, but somehow we found a block at 500K, which is higher than both. So we need to submit to both. 

I would just get the achieved. And check the achieved against both tari and monero. If its higher, submit

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -853,142 +968,35 @@ fn stream_handler(
...
-                return Err(MmProxyError::OtherError(format!("Failed to acquire lock, {:?}", e)));
+                Err(e) => Err(MmProxyError::OtherError(format!("Failed to acquire lock, {:?}", e))),
             },
+            Err(e) => Err(MmProxyError::OtherError(format!("Failed to acquire lock, {:?}", e))),
         });
     }
     Ok(())
 }
 
-#[cfg(test)]
```
SWvheerden: Dont these test pass anymore?

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -388,6 +424,50 @@ fn get_spending_key() -> (PrivateKey, PrivateKey) {
...
+                Some(monero_difficulty) => {
+                    let difficulty = min(monero_difficulty, &tari_difficulty).to_owned();
+                    parsed["result"]["difficulty"] = difficulty.into();
+                    parsed.dump().into()
+                },
+                None => {
+                    error!(
+                        target: LOG_TARGET,
+                        "{:#}",
+                        MmProxyError::ParseError(
```
stringhandler: Why not rather return this error and log it in one place in the caller?

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -388,6 +424,50 @@ fn get_spending_key() -> (PrivateKey, PrivateKey) {
     (key, r)
 }
 
+// Select lowest difficulty
+fn select_difficulty(data: &[u8], tari_difficulty: u64) -> Vec<u8> {
```
stringhandler: nit: Just a style thing, but since this method alters `data` and then returns it, I would say it's better to `move` `data` in as a `Vec<u8>` so that the old version is not used by code elsewhere. The name of the method also implies that it is only reading and not altering the data. A better name might be: `select_and_update_lowest_difficulty`

SWvheerden (commented):
looks good, just one or two comments

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -742,38 +831,53 @@ fn handle_connection(
...
-                            debug!(
-                                target: LOG_TARGET,
-                                "Response: {:#}",
-                                String::from_utf8_lossy(&data).to_string()
-                            );
-                            match handle_post(&method, &json, data.clone(), transient, consensus, rt, config) {
-                                Ok(result) => {
-                                    data = result;
+            if let "submitblock" = method.as_str() {
+                if transient.tari_difficulty.unwrap_or_default() < transient.monero_difficulty.unwrap_or_default() {
```
stringhandler, line 815: I think it would be better to calculate the difficulty of the block here and then have two if statements:
```
if diff > monero_diff {
  submit_to_monero();
}
if diff > tari_diff {
  submit_to_tari();
}
```
At some point these should be done in parallel in order to minimise the time to submit and hence increase the chance of winning the reward. Having the two if statements also handles the case where `tari_difficulty` is higher than `monero_difficulty`, which is missing

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: These tests were moved to monero_rx.

stringhandler (commented):
Happy with the concept. Some nits, but mainly the submit should also handle the case of tari_diff > monero_diff

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: It logs the error and the continues, returning the data unmodified. This will in turn be sent back to XMRig.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: We need to revisit `monero_difficulty_calculation(header: &BlockHeader)` for this, think it is incorrect as it currently is.

Based on this:
https://github.com/monero-project/monero/blob/master/src/cryptonote_core/blockchain.cpp#L4065-L4072

Monero validates the RandomX hash (using `check_hash`) against the current difficulty (using `get_difficulty_for_next_block`). 

What I think we have to do here is write a rust equivalent of `check_hash`, embed the current_difficult in MoneroData, validate the hash in the same way and then just return the current_difficulty as achieved difficulty  (`monero_data.current_difficulty`).

StriderDM:
Found a bug in this PR and have gone back to previous commit. Will reopen a new PR.

PR #2200, opened 2020-09-08
Simply trace log message

hansieodendaal:
## Description

Simplified trace log message for "Could not publish Base Node Response, no subscribers to receive.".

## Motivation and Context

The error contents do not need to be included in the trace log, for example, 5,000 UTXOs that are returned if there no more subscribers left.

## How Has This Been Tested?

Tested on a base node in Windows with a UTXO query that is returned but not handled.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3814,7 +3814,7 @@ dependencies = [
  "futures 0.3.5",
  "prost",
  "quote 1.0.7",
- "syn 1.0.38",
+ "syn 1.0.39",
  "tari_comms",
  "tari_test_utils",
  "tokio",
--- a/base_layer/core/src/proto/generated/tari.base_node.rs
+++ b/base_layer/core/src/proto/generated/tari.base_node.rs
@@ -20,82 +20,6 @@ pub struct ChainMetadata {
     #[prost(uint64, tag = "6")]
     pub effective_pruned_height: u64,
 }
-/// Response type for a received BaseNodeService requests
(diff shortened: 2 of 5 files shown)


stringhandler:
Seems ok at first glance, but I need a bit more context about the problem that this causes? Is the way that the error is printed causing trouble somewhere? Seems irresponsible to swallow the error here. Maybe a comment should be added why it's ok to swallow the error

>for example,  5,000 UTXOs that are returned if there no more subscribers left.

I don't understand this example, is "5000 UTXOs" the error message?

hansieodendaal:
Good point, let me give some more context. A typical response, as changed, where this trace is printed in response to a 5000 sized UTXO query that cannot be handled anymore looks like this:
```
2020-09-08 11:22:28.537227400 [wallet::output_manager_service] TRACE Select Loop end
2020-09-08 11:22:48.750649500 [wallet::output_manager_service] TRACE Handling Base Node Response, Trace: Tag#2789948446666732764
2020-09-08 11:22:48.750649500 [wallet::output_manager_service] TRACE Could not publish Base Node Response, no subscribers to receive.
```
Without the change the trace for those 3 lines is attached below, 16MB worth of data, which makes looking at log files unruly:
[PR-2020.log](https://github.com/tari-project/tari/files/5187613/PR-2020.log)

stringhandler:
I see. Thanks

stringhandler (approved):

PR #2202, opened 2020-09-08
Add user-configurable chain monitoring timeout

hansieodendaal:
## Description
Added a user-configurable chain monitoring timeout, to be different from the base node monitoring timeout.

## Motivation and Context
Currently, the timeout used to re-transmit a transaction to the mempool is the same as used to monitor the blockchain if a transaction has been mined. For better control during testing and in operation,  a user-configurable chain monitoring timeout is now added.

## How Has This Been Tested?
Tested in a base node on Windows.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -506,7 +506,8 @@ where
         wallet_subscriptions,
         factories.clone(),
         config.base_node_query_timeout,
-        config.transaction_base_node_monitoring_timeout,
+        config.transaction_broadcast_monitoring_timeout,
+        config.transaction_chain_monitoring_timeout,
         config.transaction_direct_send_timeout,
         config.transaction_broadcast_send_timeout,
         network,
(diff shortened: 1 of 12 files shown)


applications/tari_base_node/src/builder.rs
```diff
@@ -507,6 +507,7 @@ where
         factories.clone(),
         config.base_node_query_timeout,
         config.transaction_base_node_monitoring_timeout,
+        config.transaction_chain_monitoring_timeout,
```
stringhandler, line 510: This sounds very similar to the setting above. Perhaps a better name is `transaction_mined_monitoring_timeout`?

stringhandler (dismissed):
Happy with the concept, just think the name is too similar to "Base node monitoring timeout".

base_layer/wallet/src/transaction_service/config.rs
```diff
@@ -28,6 +28,7 @@ const LOG_TARGET: &str = "wallet::transaction_service::config";
 #[derive(Clone)]
 pub struct TransactionServiceConfig {
     pub base_node_monitoring_timeout: Duration,
+    pub chain_monitoring_timeout: Duration,
```
philipr-za, line 31: If we are separating these two timeouts into distinct config values then you should rename the `base_node_monitoring_timeout` to `broadcast_monitoring_timeout`

hansieodendaal: Saw that relation just now

philipr-za (approved):
LGTM

PR #2207, opened 2020-09-09
[Merge Mining] Adjust difficulty

StriderDM:
## Description
Selects lowest difficulty between Monero and Tari.
Simulates accept message for XMRig when only submitting to Tari.

## Motivation and Context
Allows XMRig to mine at the lowest difficulty.

## How Has This Been Tested?
Manual testing.

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -25,7 +25,7 @@ use bytes::BytesMut;
 use futures::StreamExt;
 use hyper::{
     body::Bytes,
-    http::{header, response::Parts},
+    http::{header, response::Parts, HeaderValue},
     service::Service,
     Body,
     Method,
@@ -33,13 +33,15 @@ use hyper::{
     Response,
     StatusCode,
     Uri,
+    Version,
 };
 use jsonrpc::error::StandardError;
 use log::*;
-use monero::blockdata;
 use reqwest::{ResponseBuilderExt, Url};
 use serde_json as json;
+use serde_json::Value;
 use std::{
+    cmp::min,
     convert::TryInto,
     future::Future,
     net::SocketAddr,
@@ -111,7 +113,7 @@ impl Service<Request<Body>> for MergeMiningProxyService {
             match inner.handle(req).await {
                 Ok(resp) => Ok(resp),
                 Err(err) => {
-                    debug!(target: LOG_TARGET, "Error handling GET request: {}", err);
+                    debug!(target: LOG_TARGET, "Error handling request: {}", err);
 
                     Ok(Response::builder()
                         .status(StatusCode::INTERNAL_SERVER_ERROR)
@@ -172,6 +174,7 @@ impl InnerService {
             if let Some(current_height) = transient.tari_height {
                 if height != current_height {
                     json["height"] = current_height.into();
+                    debug!(target: LOG_TARGET, "Tari tip changed.");
                 } else if let Some(submit_height) = transient.tari_prev_submit_height {
                     if submit_height >= current_height {
                         debug!(target: LOG_TARGET, "Already submitted for current Tari height");
@@ -196,9 +199,6 @@ impl InnerService {
         let mut transient = self.state.transient_data.write().await;
         let resp = monerod_resp.body();
         if resp["result"]["status"] != "OK" {
-            // Failure here means XMRig wont submit since it already succeeded to monero
-            transient.tari_block = None;
-            transient.monero_seed = None;
             return Err(MmProxyError::InvalidMonerodResponse(format!(
                 "Response status failed: {:#}",
                 resp["result"]
@@ -240,7 +240,8 @@ impl InnerService {
             .clone()
             .ok_or_else(|| MmProxyError::MissingDataError("Invalid transient proof of work".to_string()))?;
 
-        let monero_block = helpers::deserialize_from_hex(params.to_string())?;
+        let params_string = params.to_string().replace("\"", "");
+        let monero_block = helpers::deserialize_monero_block_from_hex(params_string)?;
         let monero_data = helpers::construct_monero_data(monero_block, monero_seed)?;
         let pow_data = bincode::serialize(&monero_data)?;
         pow.pow_data = pow_data;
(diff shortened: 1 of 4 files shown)


applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -226,6 +242,60 @@ fn get_request_id(json: &[u8]) -> i64 {
...
+}
+
+fn select_and_update_lowest_difficulty(data: &[u8], tari_difficulty: u64) -> Vec<u8> {
+    let monero_difficulty = get_monero_difficulty(data);
+    match monero_difficulty {
+        Some(monero_difficulty) => {
+            let parsed = json::parse(&stringify_request(data));
+            match parsed {
+                Ok(mut parsed) => {
+                    let difficulty = min(monero_difficulty, tari_difficulty).to_owned();
```
SWvheerden: This will favour the min. 
You are comparing Option<u64> here. 
None is less than some(1). 
```suggestion
                    let difficulty = min(monero_difficulty.unwrap_or(u64::max), tari_difficulty.unwrap_or(u64::max));
```

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -226,6 +242,60 @@ fn get_request_id(json: &[u8]) -> i64 {
...
+                target: LOG_TARGET,
+                "{:#}",
+                MmProxyError::ParseError(format!("Failure to parse json to get monero difficulty, {:?}", e))
+            );
+            None
+        },
+    }
+}
+
+fn select_and_update_lowest_difficulty(data: &[u8], tari_difficulty: u64) -> Vec<u8> {
```
SWvheerden: I take it  if this function fails we should start over. 
We cant mine without a difficulty. Should this not return an error.

applications/tari_merge_mining_proxy/src/main.rs
SWvheerden: Ideally if both at broken, we cant mine so we should return error.

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -736,30 +806,59 @@ fn handle_connection(
...
                 "{", "{", 418, "Curl failed to brew the request", "}", request_id, "}"
             )
             .as_bytes()
             .to_vec();
-            match base_curl(json.len() as u64, &url, true, config) {
-                Ok(mut curl) => {
-                    match do_curl(&mut curl, &json) {
-                        // request_id
+
+            if transient.tari_difficulty.unwrap_or_default() < transient.monero_difficulty.unwrap_or_default() &&
```
SWvheerden: We should look at the achieved difficulty here.  Not comparing the mined difficulty as is here currently. 
This should be a simple if :
`if achieved_block_difficulty >= tari_target_diff
submit tari

if achieved_block_difficulty >= monero_target_diff
submit monero
`
We mine to the target, but we submit/check to the achieved.

SWvheerden (commented):
I think a lot of the problems we have can be solved with proper error handling which is not implemented here. We are suppressing and or hiding errors, 

I am taking the example of the difficulty.  The handle post should know about the fact that the difficulty failed. Currently it does not. It just assumed it worked correctly and there is only a log entry that it failed.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: It will either return an error (already in 'data' because the monero request failed) or it will still have a valid monero difficulty.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: Neither are options here, tari_difficulty is passed in as u64,  monero_difficulty is matched before it reaches this line.

applications/tari_merge_mining_proxy/src/main.rs
StriderDM: While I understand this, we can assume that submitting to monero would have succeeded when only submitting to tari. But what happens if it had to fail when submitting to monero, wouldn't we be presenting an incorrect response to XMRig (which displays success and failure stats)? Unfortunately, if it succeeds to monero and fails to tari there is not much we could do about it in that case.

We also do need to revisit the achieved difficulty calculation as previously mentioned.

applications/tari_merge_mining_proxy/src/main.rs
SWvheerden: O sorry, youre right, I missed the match above.

applications/tari_merge_mining_proxy/src/main.rs
SWvheerden: Are you saying at this point in time in the code `data: &[u8]`has already been checked and the that 
` parsed = json::parse(&stringify_request(data));` cannot fail and will succeed 100% of the time.

applications/tari_merge_mining_proxy/src/main.rs
SWvheerden: What do you mean we need to revisit the achieved difficulty? 

`we can assume that submitting to monero would have succeeded when only submitting to tari`
I completely disagree with this statement. What if tari has a higher difficulty? Then we need to submit only to monero. 

We dont care about submitting an incorrect block to monero, or tari for that matter. We care about getting as much as possible coins for the miner. If we can get Tari, or Monero, or both, we need to claim. If we can submit a valid block for one, we need to do that asp. We cant wait till we have both.

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -381,30 +403,76 @@ impl InnerService {
...
-            .body(request.body().clone())
-            .send()
-            .await
-            .map_err(MmProxyError::MonerodRequestFailed)?;
-        let json_response = convert_reqwest_response_to_hyper_json_response(resp).await?;
+        let mut should_proxy = true;
+        let body: Bytes = request.body().clone();
+        let json = json::from_slice::<json::Value>(&body[..]).unwrap_or_default();
+
+        if transient.current_difficulty.unwrap_or_default() <= transient.monero_difficulty.unwrap_or_default() &&
```
stringhandler: AFAICT this is always true

StriderDM: Thanks, `<=` should have been `<`

stringhandler (approved):
Looks good. Will test and merge

PR #2214, opened 2020-09-10
Add recovery mode

SWvheerden:
## Description

Add recovery mode to re-add blocks to the db one by one.

## Motivation and Context

In the past we have had occurrences where the db got corrupted at a certain point. This PR adds in a recovery mode to allow a node to "re-sync" from its own database without adding any new data from syncing while also clearing out the current db ensuring that only correct and whole information is added to the db.  

By making the state machine do the work of re-adding the data to the database the node can be stopped from asking for blocks/data from other nodes. It is also possible to use this code to port the node from one database-to another as the source db does not necessarily have to match the end database. It reads from one database, adds to another. 

## How Has This Been Tested?

By locally rebuilding the node database. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/main.rs
+++ b/applications/tari_base_node/src/main.rs
@@ -103,6 +103,8 @@ mod parser;
 mod tasks;
 /// Utility functions
 mod utils;
+// recovery mode
+mod recovery;
 
 use log::*;
 use parser::Parser;
@@ -181,14 +183,20 @@ fn main_inner() -> Result<(), ExitCodes> {
         );
         return Ok(());
     }
+    // This is the main and only shutdown trigger for the system.
+    let shutdown = Shutdown::new();
+
+    if bootstrap.rebuild_db {
+        info!(target: LOG_TARGET, "Node is in recovery mode, entering recovery");
+        recovery::initiate_recover_db(&node_config)?;
+        let _ = rt.block_on(recovery::run_recovery(&node_config));
+        return Ok(());
+    };
 
     if bootstrap.init {
         info!(target: LOG_TARGET, "Default configuration created. Done.");
         return Ok(());
     }
-
-    // This is the main and only shutdown trigger for the system.
-    let shutdown = Shutdown::new();
 
     // Build, node, build!
     let ctx = rt
@@ -202,7 +210,6 @@ fn main_inner() -> Result<(), ExitCodes> {
             error!(target: LOG_TARGET, "{}", err);
             ExitCodes::UnknownError
         })?;
-
     // Run, node, run!
     let parser = Parser::new(rt.handle().clone(), &ctx, &node_config);
 
@@ -222,9 +229,7 @@ fn main_inner() -> Result<(), ExitCodes> {
         target: LOG_TARGET,
         "Node has been successfully configured and initialized. Starting CLI loop."
     );
-
     cli_loop(parser, shutdown);
-
     match rt.block_on(base_node_handle) {
         Ok(_) => info!(target: LOG_TARGET, "Node shutdown successfully."),
         Err(e) => error!(target: LOG_TARGET, "Node has crashed: {}", e),
--- a/applications/tari_base_node/src/recovery.rs
+++ b/applications/tari_base_node/src/recovery.rs
@@ -1,0 +1,166 @@
+// Copyright 2020. The Tari Project
+//
+// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+// following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+// disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
+// following disclaimer in the documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
+// products derived from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+// INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+// USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+use anyhow::anyhow;
+use log::*;
+use std::{
+    io::{self, Write},
+    path::Path,
+    sync::Arc,
+};
+
+use tari_app_utilities::utilities::ExitCodes;
+use tari_common::{DatabaseType, GlobalConfig};
+use tari_core::{
+    chain_storage::{
+        async_db,
+        create_lmdb_database,
+        create_recovery_lmdb_database,
+        remove_lmdb_database,
+        BlockchainBackend,
+        BlockchainDatabase,
+        BlockchainDatabaseConfig,
+        Validators,
+    },
+    consensus::{ConsensusManagerBuilder, Network as NetworkType},
+    transactions::types::CryptoFactories,
+    validation::{
+        block_validators::{FullConsensusValidator, StatelessBlockValidator},
+        mocks::MockValidator,
+    },
+};
+use tari_mmr::MmrCacheConfig;
+pub const LOG_TARGET: &str = "base_node::app";
+
+pub fn initiate_recover_db(node_config: &GlobalConfig) -> Result<(), ExitCodes> {
+    // create recovery db
+    match &node_config.db_type {
+        DatabaseType::LMDB(p) => {
+            let _backend = create_recovery_lmdb_database(&p).map_err(|err| {
+                error!(target: LOG_TARGET, "{}", err);
+                ExitCodes::UnknownError
+            })?;
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            return Err(ExitCodes::UnknownError);
+        },
+    };
+    Ok(())
+}
+
+pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), anyhow::Error> {
+    println!("Starting recovery mode");
+    let (temp_db, main_db) = match &node_config.db_type {
+        DatabaseType::LMDB(p) => {
+            let backend =
+                create_lmdb_database(&p, node_config.db_config.clone(), MmrCacheConfig::default()).map_err(|e| {
+                    error!(target: LOG_TARGET, "Error opening db: {}", e);
+                    anyhow!("Could not open DB: {}", e)
+                })?;
+            let new_path = Path::new(&p).join("temp_recovery");
+
+            let temp = create_lmdb_database(&new_path, node_config.db_config.clone(), MmrCacheConfig::default())
+                .map_err(|e| {
+                    error!(target: LOG_TARGET, "Error opening recovery db: {}", e);
+                    anyhow!("Could not open recovery DB: {}", e)
+                })?;
+            (temp, backend)
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            return Err(anyhow!("Recovery mode is only available for LMDB"));
+        },
+    };
+    let rules = ConsensusManagerBuilder::new(node_config.network.into()).build();
+    let factories = CryptoFactories::default();
+    let validators = Validators::new(
+        FullConsensusValidator::new(rules.clone()),
+        StatelessBlockValidator::new(rules.clone(), factories.clone()),
+    );
+    let db_config = BlockchainDatabaseConfig {
+        orphan_storage_capacity: node_config.orphan_storage_capacity,
+        pruning_horizon: node_config.pruning_horizon,
+        pruning_interval: node_config.pruned_mode_cleanup_interval,
+    };
+    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config)?;
+    do_recovery(db, temp_db).await?;
+
+    info!(
+        target: LOG_TARGET,
+        "Node has completed recovery mode, it will try to cleanup the db"
+    );
+    match &node_config.db_type {
+        DatabaseType::LMDB(p) => {
+            let new_path = Path::new(p).join("temp_recovery");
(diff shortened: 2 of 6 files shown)


applications/tari_base_node/src/builder.rs
```diff
@@ -352,6 +356,7 @@ async fn build_node_context<B>(
 ) -> Result<BaseNodeContext, String>
 where
     B: BlockchainBackend + 'static,
+    D: BlockchainBackend + 'static,
```
stringhandler: Why not just reuse B here? It's going to be a lot of testing effort to try different combinations

SWvheerden: At on of the places, rust complained along the lines it has no guaranteed, that the two backends are the same when just reusing` <B>`. By changing this to `<D>` it does not complain, and we gain the benefit that you swop db's by using this command.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -537,6 +536,29 @@ pub fn create_lmdb_database<P: AsRef<Path>>(
...
+    let lock_file = Path::new(path.as_ref()).join("lock.mdb");
+
+    let new_data_file = Path::new(&new_path).join("data.mdb");
+    let new_lock_file = Path::new(&new_path).join("lock.mdb");
+
+    fs::rename(data_file, new_data_file)
+        .map_err(|err| ChainStorageError::CriticalError(format!("Could not copy LMDB store:{}", err)))?;
+    fs::rename(lock_file, new_lock_file)
+        .map_err(|err| ChainStorageError::CriticalError(format!("Could not copy LMDB store:{}", err)))?;
+    Ok(())
```
sdbondi: `Path::join` creates a new `PathBuf` without consuming or mutating the current path object.

```suggestion
     let new_path = path.as_ref().join("temp_recovery");
    let _ = fs::create_dir_all(&new_path);

    let data_file = path.as_ref().join("data.mdb");
    let lock_file = path.as_ref().join("lock.mdb");

    let new_data_file = new_path.join("data.mdb");
    let new_lock_file = new_path.join("lock.mdb");

    fs::rename(data_file, new_data_file)
        .map_err(|err| ChainStorageError::CriticalError(format!("Could not copy LMDB store:{}", err)))?;
    fs::rename(lock_file, new_lock_file)
        .map_err(|err| ChainStorageError::CriticalError(format!("Could not copy LMDB store:{}", err)))?;
    Ok(())
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -537,6 +536,29 @@ pub fn create_lmdb_database<P: AsRef<Path>>(
...
+}
+
+pub fn remove_lmdb_database<P: AsRef<Path>>(path: P) -> Result<(), ChainStorageError> {
+    std::fs::remove_dir_all(&path)
```
sdbondi: nit:

```suggestion
    fs::remove_dir_all(&path)
```

base_layer/core/src/base_node/state_machine_service/state_machine.rs
```diff
@@ -215,6 +223,39 @@ impl<B: BlockchainBackend + 'static> BaseNodeStateMachine<B> {
         let _ = self.service_shutdown.trigger();
     }
 
+    // Function to handle the recovery attempt of the db
+    fn recover_db<D: BlockchainBackend + 'static>(self, temp_db: D) -> Result<(), String> {
```
sdbondi: I don't see why the recover_db function has to be part of the base node state machine.

Can this not all happen in the base_node_application? All you need to db access.

sdbondi (commented):
This will be a useful tool which I hope will not be needed :)

I cannot see why BaseNodeStateMachine is involved here. The recovery can all take place within the base_node_application code.

base_layer/core/src/base_node/state_machine_service/state_machine.rs
SWvheerden: You need the backend to run as normal as it needs to do all the validation that would normally run on a new block. 
By starting up the state_machine, we know the internal backend is run normally, and we have full control over the state, and can ensure that the statemachine does not enter a sync state.

sdbondi: > You need the backend to run as normal as it needs to do all the validation that would normally run on a new block.

This happens independently of the state machine so I don't see this as a good argument for putting this code here.

> By starting up the state_machine, we know the internal backend is run normally, and we have full control over the state, and can ensure that the statemachine does not enter a sync state.

Do we ensure that the state machine does not enter a sync state? Do we need to? None of the code does what you are saying here.

base_layer/core/src/base_node/state_machine_service/state_machine.rs
```diff
@@ -215,6 +223,39 @@ impl<B: BlockchainBackend + 'static> BaseNodeStateMachine<B> {
...
+            .height_of_longest_chain
+            .unwrap_or(0);
+        info!(target: LOG_TARGET, "Starting recovery",);
+        // we start at height 1
+        let mut counter = 1;
+        loop {
+            trace!(target: LOG_TARGET, "Asking for block with height: {}", counter);
+            let block = temp_db_backend.fetch_block(counter).map_err(|e| e.to_string())?.block;
+            trace!(target: LOG_TARGET, "Adding block: {}", block);
+            self.db.add_block(block).map_err(|e| e.to_string())?;
```
sdbondi: You should use async_db - this function will synchronously block an a tokio runtime thread (preventing task multiplexing on that thread)

applications/tari_base_node/src/main.rs
```diff
@@ -174,6 +176,36 @@ fn main_inner() -> Result<(), ExitCodes> {
...
+                )
+            },
+            _ => {
+                error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+                return Ok(());
+            },
+        }
+    } else {
+        None
+    };
```
CjS77: This should all go into a separate module.

Recovery could be a completely separate executable, so it doesn't go it the main node event loop.

applications/tari_base_node/src/main.rs
```diff
@@ -213,12 +246,34 @@ fn main_inner() -> Result<(), ExitCodes> {
         "Node has been successfully configured and initialized. Starting CLI loop."
     );
 
-    cli_loop(parser, shutdown);
+    if !bootstrap.rebuild_db {
```
CjS77: This is pretty messy since you do another check a few lines later.
and unscalable (what if there's a `bootstrap.foo` option in future?)

applications/tari_base_node/src/builder.rs
CjS77: Actually, this feels like we're violating the single-responsibility and the Open-Closed principle here.

If Recovery was in it's own module, you could call builder twice, once for the current DB and once for the new DB, and respect both SOLID practices.

applications/tari_base_node/src/main.rs
CjS77: All these issues go away if recovery is its own separate module

base_layer/core/src/base_node/state_machine_service/state_machine.rs
CjS77: Agree with @sdbondi. Essentially, the idea (recovery) is a good one, but this implementation is a bit hacky that violates several SOLID principles. If it were refactored into a separate module that reused the existing mechanisms, I would approve it.

CjS77 (changes requested):
Support and like the idea, but disagree with the specific approach.

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            return Err(ExitCodes::UnknownError);
+        },
+    };
+    Ok(())
+}
+
+pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), String> {
+    println!("string recovery mode");
```
sdbondi:
```suggestion
    println!("Starting recovery mode");
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            return Err(ExitCodes::UnknownError);
+        },
+    };
+    Ok(())
+}
+
+pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), String> {
```
sdbondi: I added `anyhow` as a dep for handling misc errors in applications

```suggestion
pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), anyhow::Error> {
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+    println!("string recovery mode");
+    let (temp_db, main_db) = match &node_config.db_type {
+        DatabaseType::LMDB(p) => {
+            let backend =
+                create_lmdb_database(&p, node_config.db_config.clone(), MmrCacheConfig::default()).map_err(|e| {
+                    error!(target: LOG_TARGET, "Error opening db: {}", e);
+                    format!("Could not open DB: {}", e.to_string())
```
sdbondi:
```suggestion
                    anyhow!("Could not open DB: {}", e)
```

`use anyhow::anyhow` at the top

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+                })?;
+            let new_path = Path::new(&p).join("temp_recovery");
+
+            let temp = create_lmdb_database(&new_path, node_config.db_config.clone(), MmrCacheConfig::default())
+                .map_err(|e| {
+                    error!(target: LOG_TARGET, "Error opening recovery db: {}", e);
+                    format!("Could not open recovery DB: {}", e.to_string())
```
sdbondi: The `{}` formatter calls the same code as `to_string`

```suggestion
                    anyhow!("Could not open recovery DB: {}", e)
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+                })?;
+            (temp, backend)
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            Err(format!("Recovery mode is only available for LMDB"))?
```
sdbondi:
```suggestion
            Err(anyhow!("Recovery mode is only available for LMDB"))?
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        target: LOG_TARGET,
+        "Node has completed recovery mode, it will try to cleanup the db"
+    );
+    match &node_config.db_type {
+        DatabaseType::LMDB(p) => {
+            let new_path = Path::new(p).join("temp_recovery");
+            remove_lmdb_database(&new_path).map_err(|e| {
+                error!(target: LOG_TARGET, "Error removing recovery db: {}", e);
+                format!("Could not remove recovery DB: {}", e.to_string())
+            })
```
sdbondi: We log any errors anyway - so your code could be cleaned up in a number of places by just using `?` and `anyhow`.

```suggestion
            remove_lmdb_database(&new_path)?;
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+    );
+    let db_config = BlockchainDatabaseConfig {
+        orphan_storage_capacity: node_config.orphan_storage_capacity,
+        pruning_horizon: node_config.pruning_horizon,
+        pruning_interval: node_config.pruned_mode_cleanup_interval,
+    };
+    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config).map_err(|e| {
+        error!(target: LOG_TARGET, "Error could not create BlockchainDatabase {}", e);
+        format!("Could not create BlockchainDatabase: {}", e.to_string())
+    })?;
```
sdbondi:
```suggestion
    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config)?;
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            Ok(())
+        },
+    }
+}
+
+// Function to handle the recovery attempt of the db
+async fn do_recovery<D: BlockchainBackend + 'static>(db: BlockchainDatabase<D>, temp_db: D) -> Result<(), String> {
```
sdbondi:
```suggestion
async fn do_recovery<D: BlockchainBackend + 'static>(db: BlockchainDatabase<D>, temp_db: D) -> Result<(), anyhow::Error> {
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+    // We dont care about the values, here, so we just use mock validators, and a mainnet CM.
+    let rules = ConsensusManagerBuilder::new(NetworkType::LocalNet).build();
+    let validators = Validators::new(MockValidator::new(true), MockValidator::new(true));
+    let temp_db_backend = BlockchainDatabase::new(temp_db, &rules, validators, BlockchainDatabaseConfig::default())
+        .map_err(|e| e.to_string())?;
```
sdbondi:
```suggestion
    let temp_db_backend = BlockchainDatabase::new(temp_db, &rules, validators, BlockchainDatabaseConfig::default())?;
```

Return the error using `?` everywhere else too

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        trace!(target: LOG_TARGET, "Adding block: {}", block);
+        db.add_block(Arc::new(block)).map_err(|e| {
+            info!(
+                target: LOG_TARGET,
+                "Stopped recovery at height {}, reason: {}",
+                counter,
+                e.to_string()
+            );
+            e.to_string()
+        })?;
```
sdbondi: Use async db so to not block a tokio thread (yes, it's not a big deal in this case but we should get in the habit of doing it)

Use `anyhow!` to provide extra info.

```suggestion
        async_db::add_block(db.clone(), Arc::new(block)).await.map_err(|e| {
            anyhow!(
                "Stopped recovery at height {}, reason: {}",
                counter,
                e
            )
        })?;
```

sdbondi (commented):
Looks great now that it doesn't touch the state machine.

The rest of my comments are basically nits about error handling and code neatness (think they also fix some clippy errors)

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -537,6 +536,29 @@ pub fn create_lmdb_database<P: AsRef<Path>>(
     LMDBDatabase::<HashDigest>::new(lmdb_store, mmr_cache_config)
 }
 
+pub fn create_recovery_lmdb_database<P: AsRef<Path>>(path: P) -> Result<(), ChainStorageError> {
+     let new_path = path.as_ref().join("temp_recovery");
```
sdbondi: Formatting
```suggestion
    let new_path = path.as_ref().join("temp_recovery");
```

applications/tari_base_node/src/recovery.rs
```diff
@@ -0,0 +1,166 @@
...
+            let temp = create_lmdb_database(&new_path, node_config.db_config.clone(), MmrCacheConfig::default())
+                .map_err(|e| {
+                    error!(target: LOG_TARGET, "Error opening recovery db: {}", e);
+                    anyhow!("Could not open recovery DB: {}", e)
+                })?;
+            (temp, backend)
+        },
+        _ => {
+            error!(target: LOG_TARGET, "Recovery mode is only available for LMDB");
+            Err(anyhow!("Recovery mode is only available for LMDB"))?
```
sdbondi: 📎 
```suggestion
            return Err(anyhow!("Recovery mode is only available for LMDB"));
```

sdbondi (dismissed):
Looks good 🚢   (not explicitly tested)

PR #2216, opened 2020-09-10
Delay UTXO validation till after initial blockchain sync

hansieodendaal:
## Description
- Delay UTXO validation till the initial blockchain sync has been concluded.
- Fix clippy warnings.

## Motivation and Context
If not delayed, UTXO validation will almost never succeed, as the validation is aborted if the node is not synced.

## How Has This Been Tested?

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/service/service.rs
+++ b/base_layer/core/src/base_node/service/service.rs
@@ -33,7 +33,7 @@ use crate::{
         generate_request_key,
         proto,
         service::error::BaseNodeServiceError,
-        state_machine_service::states::StatusInfo,
+        state_machine_service::states::StateInfo,
         RequestKey,
         StateMachineHandle,
         WaitingRequests,
@@ -354,12 +354,7 @@ where B: BlockchainBackend + 'static
         // Determine if we are bootstrapped
         let status_watch = self.state_machine_handle.get_status_info_watch();
 
-        let bootstrapped = match *(status_watch.borrow()) {
-            StatusInfo::Listening(li) => li.is_bootstrapped(),
-            _ => false,
-        };
-
-        if !bootstrapped {
+        if !(*status_watch.borrow()).bootstrapped {
             debug!(
                 target: LOG_TARGET,
                 "Propagated block `{}` from peer `{}` not processed while busy with initial sync.",
@@ -437,13 +432,10 @@ async fn handle_incoming_request<B: BlockchainBackend + 'static>(
         .await?;
 
     // Determine if we are synced
-    let mut status_watch = state_machine_handle.get_status_info_watch();
-    let is_synced = match status_watch.recv().await {
-        None => false,
-        Some(s) => match s {
-            StatusInfo::Listening(li) => li.is_synced(),
-            _ => false,
-        },
+    let status_watch = state_machine_handle.get_status_info_watch();
+    let is_synced = match (*status_watch.borrow()).state_info {
+        StateInfo::Listening(li) => li.is_synced(),
+        _ => false,
     };
 
     let message = proto::BaseNodeServiceResponse {
--- a/base_layer/core/src/base_node/state_machine_service/initializer.rs
+++ b/base_layer/core/src/base_node/state_machine_service/initializer.rs
@@ -97,7 +97,7 @@ where B: BlockchainBackend + 'static
     ) -> Self::Future
     {
         let (state_event_publisher, state_event_subscriber): (Publisher<_>, Subscriber<_>) = bounded(10, 3);
-        let (status_event_sender, status_event_receiver) = tokio::sync::watch::channel(StatusInfo::StartUp);
+        let (status_event_sender, status_event_receiver) = tokio::sync::watch::channel(StatusInfo::new());
 
         let shutdown = Shutdown::new();
         let handle = StateMachineHandle::new(state_event_subscriber, status_event_receiver, shutdown.to_signal());
--- a/base_layer/core/src/base_node/state_machine_service/state_machine.rs
+++ b/base_layer/core/src/base_node/state_machine_service/state_machine.rs
@@ -30,6 +30,7 @@ use crate::{
                 BlockSyncConfig,
                 HorizonSyncConfig,
                 StateEvent,
+                StateInfo,
                 StatusInfo,
                 SyncPeerConfig,
                 SyncStatus,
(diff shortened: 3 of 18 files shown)


delta1:
LGTM built and ran tests successfully

applications/tari_base_node/src/builder.rs
```diff
@@ -473,19 +471,42 @@ where
...
+                    .validate_utxos(UtxoValidationRetry::UntilSuccess)
+                    .await
+                    .expect("Problem starting the Output Manager Service Utxo Valdation process");
+                break;
+            }
+            debug!(
+                target: LOG_TARGET,
+                "UTXO validation process not started while busy with initial sync.",
+            );
+            tokio::time::delay_for(Duration::from_millis(30_000)).await;
```
philipr-za: might as well be `from_secs(30)`

philipr-za (dismissed):
LGTM

applications/tari_base_node/src/builder.rs
```diff
@@ -473,19 +471,42 @@ where
...
+        loop {
+            let bootstrapped = match status_watch.recv().await {
+                None => false,
+                Some(s) => match s {
+                    StatusInfo::Listening(li) => li.is_bootstrapped(),
+                    _ => false,
+                },
+            };
+            if bootstrapped {
+                let mut oms_handle = wallet_handles_clone
```
philipr-za: This is slightly inefficient, I would grab this handle outside the closure and pass it in instead of having to get it each loop. Same as you do with the state machine handle

hansieodendaal: I disagree with this one: `oms_handle` is used directly after and then the code breaks out of the loop, thus called only once. It is defined in the context where it is used.

applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -49,7 +49,7 @@ pub fn setup_node_identity<P: AsRef<Path>>(
     peer_features: PeerFeatures,
 ) -> Result<Arc<NodeIdentity>, ExitCodes>
 {
-    match load_identity(identity_file.as_ref().clone()) {
+    match load_identity(<&std::path::Path>::clone(&identity_file.as_ref())) {
```
sdbondi: `load_identity` takes a `P: AsRef<Path>` which is what `identity_file` already is, so this will work:

```suggestion
    match load_identity(identity_file) {
```

applications/tari_base_node/src/builder.rs
sdbondi: `delay_for` is imported in builder.rs. 
```suggestion
            delay_for(Duration::from_secs(30)).await;
```

sdbondi (commented):
LGTM - few small changes

applications/tari_app_utilities/src/identity_management.rs
hansieodendaal:
```suggestion
     match load_identity(identity_file.as_ref()) {
```

sdbondi: :+1:  I see it's used later in the code - `identity_file.as_ref()` is also acceptable

sdbondi: or changing the others below to `&identity_file`

hansieodendaal: Needed a reference, as the value is borrowed further down

applications/tari_base_node/src/builder.rs
```diff
@@ -473,19 +471,42 @@ where
...
+                    .get_handle::<OutputManagerHandle>()
+                    .expect("OutputManagerService is not registered");
```
stringhandler: I might be wrong, but the thread that is spawned here is never joined, so this panic will never be caught. The program will continue and this error will never be logged.

The main problem here is that a thread is getting started in the background, and there's no logical place for it to be joined. This makes me think it would be better to have the caller of this method be responsible for starting and joining this thread. It might also be better to listen for a state change event instead of polling the state each time

stringhandler (commented):
I'm happy with the idea, but I think the approach of moving the code into a thread that is never joined is going to swallow errors and panics

applications/tari_base_node/src/builder.rs
hansieodendaal: We could possibly add further messages to the log file to determine if an exception occurred? Apart from that, this code is simple and works well under real base node and stress conditions. 

My proposal:

``` rust
            if bootstrapped {
                debug!(
                    target: LOG_TARGET,
                    "UTXO validation to start: Get OutputManagerHandle.",
                );
                let mut oms_handle = wallet_handles_clone
                    .get_handle::<OutputManagerHandle>()
                    .expect("OutputManagerService is not registered");
                debug!(
                    target: LOG_TARGET,
                    "UTXO validation to start: Set BaseNodePublicKey.",
                );
                oms_handle
                    .set_base_node_public_key(base_node_public_key)
                    .await
                    .expect("Problem setting local base node public key for output manager service.");
                debug!(
                    target: LOG_TARGET,
                    "UTXO validation to start: Attempting to start the Output Manager Service UTXO Valdation process.",
                );
                oms_handle
                    .validate_utxos(UtxoValidationRetry::UntilSuccess)
                    .await
                    .expect("Problem starting the Output Manager Service UTXO Valdation process");
                debug!(
                    target: LOG_TARGET,
                    "UTXO validation to start: UTXO Valdation process concluded successfully .",
                );
                break;
            }
```

philipr-za: The expect messages will still not be logged. I would suggest that the getting of the handle happen before the spawn so that will panic and for the other two you can use a map error before the expect to log the problem e.g.
```Rust
oms_handle
                    .set_base_node_public_key(base_node_public_key)
                    .await
                    .map_err(|e| {
                            error!(target: LOG_TARGET, "An error occured: {}", e);
                            e
                     }).expect("Problem setting local base node public key for output manager service.");
```

hansieodendaal: Done

philipr-za:
Will need to be updated in context of https://github.com/tari-project/tari/pull/2239

applications/tari_base_node/src/builder.rs
```diff
@@ -473,19 +471,59 @@ where
...
+                        e
+                    })
+                    .expect("An error occurred with Output Manager Service starting the UTXO Validation process");
+                break;
+            }
+            debug!(
+                target: LOG_TARGET,
+                "UTXO validation process not started while busy with initial sync.",
+            );
+            delay_for(Duration::from_secs(30)).await;
```
philipr-za: This is actually unnecessary, the recv() above will block until a new status is published.

hansieodendaal: Updated

hansieodendaal:
Now dependent on #2239, rebased on top of that branch

PR #2217, opened 2020-09-10
Fix LWMA issues

SWvheerden:
## DO NOT MERGE IN CURRENT TESTNET
This will break current consensus. 

## Description

Fix difficulty adjustment bug in amount of timestamps required
Fix difficulty adjustment bug in the constants
fix difficulty adjustment bug in timestamp correction

## Motivation and Context

This fixes 3 bugs in the difficulty adjustment. 

The first bug fixed is that currently the piece of code that feeds the lwma, only asks the `constants.block_window`. The LWMA uses more that the block_window to do the calculations. Before this PR it required `constants.block_window +1 `. This PR lifts it up to `+2`. 

The second bug related to the first is that the lwma "fixes" the timestamps to ensure they are in order. 
For example if the lwma has the following 3 timestamps: `[50][49][52]`. It will fix the timestamps and use` [50][51][52]`. 
The lwma will use the timestamps ad calculate the solvetimes. But if it fixed a timestamp is not used in the solvetime anymore, it the solvetime will look higher than it was. This can be a significant number. Going back to the previous example of `[50][49][52]`, the solve times will be made `[1][1][1]`. This is all still fine, but one block later, the sequence looks as follows: `[49][52][53]`. No the solvetime looks like `[3][1][1]` This is a sudden increase. We need to know we corrected that first sample. So we need to increase the looked at timestamps by +1. 

The 3rd bug was that the constant for the maximum block window was set during testing to by 60 * target_time. This is too large as  lwma has a tendency to drop too much for long solvetimes.  It can motivates cycles, which can be seen in the current lwma. We need to to limit this to 6. See: https://github.com/zawy12/difficulty-algorithms/issues/3 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/validators/headers.rs
+++ b/base_layer/core/src/base_node/validators/headers.rs
@@ -79,8 +79,8 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
             Difficulty::from(1)
         } else {
             let target_difficulties = self.fetch_target_difficulties(block_header)?;
-
             let constants = self.rules.consensus_constants();
+            dbg!(constants.get_difficulty_block_window());
             get_target_difficulty(
                 target_difficulties,
                 constants.get_difficulty_block_window() as usize,
@@ -98,7 +98,8 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
                 ))
             })?
         };
-
+        dbg!(&block_header.pow.target_difficulty);
+        dbg!(&target);
         if block_header.pow.target_difficulty != target {
             warn!(
                 target: LOG_TARGET,
@@ -134,7 +135,9 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
         block_header: &BlockHeader,
     ) -> Result<Vec<(EpochTime, Difficulty)>, ValidationError>
     {
-        let block_window = self.rules.consensus_constants().get_difficulty_block_window();
+        // We need to request + 2 timestamps as the lwma uses the first 2 only for timestamp verification and not
+        // calculation
+        let block_window = self.rules.consensus_constants().get_difficulty_block_window() + 2;
         let start_height = block_header.height.saturating_sub(block_window);
         if start_height == block_header.height {
             return Ok(vec![]);
@@ -144,7 +147,7 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
             target: LOG_TARGET,
             "fetch_target_difficulties: new header height = {}, block window = {}",
             block_header.height,
-            block_window
+            block_window - 2
         );
 
         let block_window = block_window as usize;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,18 +147,18 @@ checksum = "25f9db3b38af870bf7e5cc649167533b493928e50744e2c30ae350230b414670"
 dependencies = [
  "proc-macro2 1.0.21",
  "quote 1.0.7",
- "syn 1.0.39",
+ "syn 1.0.40",
 ]
 
 [[package]]
 name = "async-trait"
-version = "0.1.38"
+version = "0.1.40"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e1a4a2f97ce50c9d0282c1468816208588441492b40d813b2e0419c22c05e7f"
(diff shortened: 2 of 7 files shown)


base_layer/core/src/base_node/validators/headers.rs
```diff
@@ -79,8 +79,9 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
             Difficulty::from(1)
         } else {
             let target_difficulties = self.fetch_target_difficulties(block_header)?;
-
+            dbg!(&target_difficulties);
```
stringhandler: These should be removed or have a target added to them

stringhandler:
I don't know if you've made a strong enough case to justify this. I'm struggling to follow your example.

>The first bug fixed is that currently the piece of code that feeds the LWMA, only asks the constants.block_window. The LWMA uses more that the block_window to do the calculations. Before this PR it required constants.block_window +1 . This PR lifts it up to +2. 

Adding and subtracting to this number should be controlled inside of the the LWMA functions only. Otherwise there will be places where the 2 was not added. 

I do not understand the second bug at all. Perhaps if you used realistic timestamps it would be more clear?

I agree with the third bug.

SWvheerden:
Lets rephrase how lwma works.  
The formula for LWMA is:  `(N*ST{N} + [N-1]*ST{N-1} + [N-2] *ST{N-2}) / N + [N-1] + [N-2]`
where ST = Solve_time and N = block height. And the window = 3

But we have 2 potential problems here, we dont store ST, and we allow negative ST's to a degree. In our code, we essentially feed the LWMA a tuple of (Block_time, Difficulty). From the block_time we can calculate the ST. As the block_times are unix timestamps.

So lets look at take an example here. 
We populate the time data of the lwma with the following data:
`Block_times[1598918400, 1598918460, 1598918120] and difficulties: [10_000,10_000,10_000]`

Now we can calculate the ST by taking `Block_time{N} - Block_time{N-1}`
We can then construct a array of ST: `[60,60]`. 
Now we see we need N-3 to calculate the solve_time ST{N-2}  as the `ST{N-2} = BT{N-2} - BT{N-3}`
So we need an extra tuple of data:
So we should rather have :
`Block_times[1598918340, 1598918400, 1598918460, 1598918120] `and `difficulties: [10_000,10_000,10_000, 10_000]`
Now we can have a` ST: [60,60,60]`. Which is exactly the block window of 3.
Now because LWMA cannot use zero or negative ST, but calculated ST's can have a negative or zero value we need to "fix" them be fore using them. 

So in the case where we have the following tuple of data:
` Block_times[1598918400, 1598918300, 1598918460, 1598918120]` and `difficulties: [10_000,10_000,10_000, 10_000]`
we have `ST's of: [-100,60,60]`. So before using the values, we "fix" the ST's to this:` [1,59,60]` by making using a fixed timestamp data of:` [1598918400, 1598918401, 1598918460, 1598918120]`. The ST's area again calculated as above from the timestamps. 

This is all how it currently works. 
Now lets add another block to the chain, the LWMA's data looks as follows:
` Block_times[1598918300, 1598918460, 1598918120, 1598918180]`. Now the ST's are: `[160,60,60]`.
That first solvetime is not technically correct. as the timestamp "1598918300" is not correct. It was fixed with the previous calculation, but now the lwma does not have the information to know it is incorrect. So calculates an incorrect ST from it. 
So we add another blockimestamp data to it, so the LWMA data looks as follows now:
`Block_times[1598918400, 1598918300, 1598918460, 1598918120, 1598918180] `and `difficulties: [10_000,10_000,10_000, 10_000,10_000]`

Now it can see  that this timestamp "1598918300" is wrong and fix it. and the resulting ST's:` [59,60,60]` which is correct.

stringhandler:
I see where you are going with this, but AFAICS, the only blocks that would be affected by this is really quick blocks at the beginning of the chain. Once the window hits its full size, the blocks at this end of the window contribute so little (`1/block_window!`) to the end result, I'm not sure that it's worth the extra complexity. As long all nodes agree, I don't see the need for this extra implementation detail.

SWvheerden:
Closing this PR for now. 
bug: `Fix difficulty adjustment bug in the constants` was fixed. 
Due to the complicity of the other we are going to see if this is needed or not. Can this work well enough if this is created sub-optimally.

PR #2221, opened 2020-09-11
Allow client application to control when transaction broadcast protocols are restarted

philipr-za:
## Description
Previously the wallet restarted its transaction broadcast and negotiation protocols as soon as a base node pubkey was set. This caused problems in the base node where on a restart the mempool was empty but the chain had yet to sync so validly mined transactions were cancelled because their chain monitoring protocol started before the blocks were synced.

This PR separates the functions to restart the transaction protocols from the set_base_node_public_key function in the transaction service. This allows the transaction negotiation protocols to be started ASAP as they don’t need a base node and then the client application can choose when to start the broadcast protocols.

The base node application was updated to wait for the local node to be synced before starting the transaction broadcast protocols. The FFI library was updated to start these protocol right after the base node public key was supplied to retain the previous behaviour in the mobile wallets.

## How Has This Been Tested?
tests updated

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -467,12 +467,59 @@ where
 
     // Set the base node for the wallet to the 'local' base node
     let base_node_public_key = base_node_comms.node_identity().public_key().clone();
-    wallet_handles
+    let mut transaction_service_handle = wallet_handles
         .get_handle::<TransactionServiceHandle>()
-        .expect("TransactionService is not registered")
+        .expect("TransactionService is not registered");
+    transaction_service_handle
         .set_base_node_public_key(base_node_public_key.clone())
         .await
         .expect("Problem setting local base node public key for transaction service.");
+    transaction_service_handle
+        .restart_transaction_protocols()
+        .await
+        .expect("Problem restarting transaction protocols in the Transaction Service");
+    // Only start the transaction broadcast protocols once the local node is synced
+    let state_machine = base_node_handles
+        .get_handle::<StateMachineHandle>()
+        .expect("Could not get State Machine handle");
+    task::spawn(async move {
+        let mut status_watch = state_machine.get_status_info_watch();
+        debug!(
+            target: LOG_TARGET,
+            "Waiting for initial sync before restarting transaction protocols."
+        );
+        loop {
+            let bootstrapped = match status_watch.recv().await {
+                None => false,
+                Some(s) => s.bootstrapped,
+            };
+
+            if bootstrapped {
+                let _ = transaction_service_handle
+                    .restart_broadcast_protocols()
+                    .await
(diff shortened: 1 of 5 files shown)


applications/tari_base_node/src/builder.rs
```diff
@@ -467,12 +467,45 @@ where
...
-    wallet_handles
+    let mut transaction_service_handle = wallet_handles
         .get_handle::<TransactionServiceHandle>()
-        .expect("TransactionService is not registered")
+        .expect("TransactionService is not registered");
+    transaction_service_handle
         .set_base_node_public_key(base_node_public_key.clone())
         .await
         .expect("Problem setting local base node public key for transaction service.");
+    transaction_service_handle
```
hansieodendaal, line 477: I think this should also be delayed until the initial sync has been confirmed, which should be the main priority at startup. For example, **has the Tx been mined** is dependent on an up to date blockchain.

hansieodendaal (commented):
Nice refactoring! This logic makes a lot of sense.

Just one comment on when to start the transaction service for a base node wallet.

applications/tari_base_node/src/builder.rs
philipr-za: So with this refactor all this does is set the base node pub key, the protocols are only started when the explicit functions are called which is done in the task that waits for the sync to complete. The service itself needs to be started so that its handles are available to the other services.

applications/tari_base_node/src/builder.rs
```diff
@@ -467,12 +467,45 @@ where
...
+        .restart_transaction_protocols()
```
hansieodendaal, line 478: This is the actual context; it looks like the transaction protocols are started before initial sync has been achieved, unlike `.restart_broadcast_protocols()`

applications/tari_base_node/src/builder.rs
hansieodendaal: Please see comment below

applications/tari_base_node/src/builder.rs
philipr-za: Transaction protocols are the negotiation protocols with other wallets so this means that the wallet is immediately available to receive/complete negotiation of more transactions, any broadcast and chain monitoring protocols that start from scratch before the BN is synced are fine, its the restart of those protocols before the bn is synced that is the problem.

philipr-za: Ok point taken, to avoid any potential issues in the BN wallet I have made this change.

hansieodendaal:
LGTM

philipr-za:
needs to be updated in light of https://github.com/tari-project/tari/pull/2239

stringhandler (approved):

PR #2226, opened 2020-09-14
Fully async, refactored merge mining proxy

sdbondi:
## Description

This PR is a refactor of the merge mining proxy. It provides a structure
to the code that should make it much easier to maintain and understand
as well as massive performance and stability improvements.

- Used hyper for accepting HTTP connections (it is a dependency of tonic
  anyway, and it pretty great for creating a http server at a low level)
- All connections are spawned on their own tokio task
- Removed all manual HTTP header parsing
- Use async RwLocks for shared state (maybe we could remove locks
  altogether in future).
- Fixed a bug with the ordering of the transaction hashes which would
  cause the monero root validation to fail if more than the coinbase was
  in the Monero block.
- Many small optimisations (extra clones, inefficient allocations, early
  function exits etc)
- Removed some unnecessary type conversions and byte wrangling
- Removed CURL dependency
- Added some very basic tests (todo: write more tests!)
- Fixed a few clippy lints

## Motivation and Context

The previous threading would cause crashes.  
```
Merged Mining Proxy started.
thread 'main' panicked at 'failed to spawn thread: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', /rustc/feb3536eba10c2e4585d066629598f03d5ddc7c6/src/libstd/thread/mod.rs:619:5
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
```

## How Has This Been Tested?

It's not tested - more automated tests need to be written (perhaps for another PR).

Merge mining with this proxy has not been tested, but is a TODO

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


applications/tari_app_grpc/proto/types.proto
```diff
@@ -68,9 +68,11 @@ message ProofOfWork {
 
 //This is used to request the which pow algo should be used with the block template
 message PowAlgo{
-    // 0 = Monero
-    // 1 = Blake
-    uint32 pow_algo = 1;
+    enum PowAlgo {
+        PowAlgoMonero = 0;
```
stringhandler:
```suggestion
        POW_ALGO_MONERO = 0;
```

applications/tari_app_grpc/proto/types.proto
```diff
@@ -68,9 +68,11 @@ message ProofOfWork {
...
+        PowAlgoBlake = 1;
```
stringhandler:
```suggestion
        POW_ALGO_BLAKE = 1;
```

applications/tari_app_grpc/proto/types.proto
stringhandler: As per https://developers.google.com/protocol-buffers/docs/style

applications/tari_merge_mining_proxy/src/error.rs
```diff
@@ -20,27 +20,58 @@
...
 pub enum MmProxyError {
-    #[error("Chain storage error: {source}")]
-    ChainStorageError {
-        #[from]
-        source: ChainStorageError,
-    },
+    #[error("Configuration error: {0}")]
+    ConfigurationError(#[from] ConfigurationError),
+    #[error("Configuration error: {0}")]
+    ConfigError(#[from] ConfigError),
```
stringhandler: Weird that these two similarly named errors exist. I've created #2228 to merge them

stringhandler (dismissed):
Happy from reviewing the code, but I'll need to test it locally to make sure

applications/tari_app_grpc/proto/types.proto
sdbondi: `Use CamelCase (with an initial capital) for enum type names and CAPITALS_WITH_UNDERSCORES for value names`

Sounds like you're right, though I'm a little confused about the difference between a "enum type name" and "value name" - updated.

base_layer/core/src/proof_of_work/monero_rx.rs
```diff
@@ -190,45 +197,62 @@ fn monero_difficulty_calculation(header: &BlockHeader) -> Result<Difficulty, Mer
...
 /// Utility function to transform array to fixed array
-pub fn from_slice(bytes: &[u8]) -> [u8; 32] {
+///
+/// ## Panics
+///
+/// Panics if the given slice has a length of less than 32
+pub fn from_slice_to_array(bytes: &[u8]) -> [u8; 32] {
     let mut array = [0; 32];
-    let bytes = &bytes[..array.len()]; // panics if not enough data
+    // TODO: Is a panic wise here? Does every caller know (invariant) the length before calling this function?
```
SWvheerden: On principle I am totally against panics, as they cause all shorts of issues in incorrect error handling. 
I would personally put an result here and check it. 

If we look at where its called from:
`let expected_merge_mining_hash = Hash::from(from_slice_to_array(&header.merged_mining_hash()));`
This function should not panic has it purely creates a hash, but then again if that fails, I think it returns an empty vec. But the function that calls this, does return a result.

SWvheerden (commented):
LGTM

This a much needed improvement, thanks

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    /// Proxy a request received by this server to Monerod
+    async fn proxy_request_to_monerod(
+        &self,
+        mut req: Request<Body>,
+    ) -> Result<(Request<Bytes>, Response<json::Value>), MmProxyError>
+    {
+        let monerod_uri = self.get_fully_qualified_monerod_url(req.uri())?;
+        debug!(target: LOG_TARGET, "Proxying request to {}", monerod_uri);
```
StriderDM: If the request is a post request, we would want to log the `method` in the request as well, since they all go to the same place (/json_rpc).

`id` would be useful as well.

sdbondi: Good idea!

base_layer/core/src/proof_of_work/monero_rx.rs
sdbondi: I think these conversions can be removed with the other changes I've made anyhow. They are only used in a few places in tests - will remove

sdbondi: Removed this conversion :)

stringhandler (dismissed):

stringhandler (approved):
Reviewed clippy fixes

PR #2227, opened 2020-09-14
Log number and time it takes to search for orphan chain tips

sdbondi:
This PR adds a log line that displays the number of orphan blocks
searched when finding orphan chain tips.
For example

```
[c::cs::database] DEBUG Searched 719 orphan(s), found 0 parent(s) in 22s
```

 This is a inefficient algorithm, `O(n)` (best case). Typically, this does not matter
 so much because the number of orphan blocks is low. However, on my node this step
 would take ~22s during add_block (719 orphans searched). After manually clearing
 my orphan db, each add_block call would take ~1s.


stringhandler (approved):

stringhandler:
That's pretty wild. Definitely a place that can be improved

PR #2230, opened 2020-09-16
Expose kernel excess, public nonce, and signature via wallet FFI

### LINKED ISSUE #2201: Expose kernel excess and excess signature via FFI
Currently, the Aurora wallet doesn't show any details about the transaction in the transaction details view. It would be useful to have a link to a block explorer or the excess or excess signature in order to prove to the receiver that I have sent it. 

The FFI currently doesn't expose these fields (AFAIK)

delta1:
## Description
This commit adds functions to the wallet FFI in order to get the transaction excess, public nonce, and excess signature of a `CompletedTransaction`. This will allow the mobile apps to link to the block explorer to search for the transaction and display the block once tari-project/block-explorer-frontend#107 is implemented.

## Motivation and Context
Resolves #2201 

## How Has This Been Tested?

Existing tests pass, new functions utilised in existing tests, wallet FFI builds successfully. The actual FFI needs to be tested from the mobile wallet side. 

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet_ffi/src/error.rs
+++ b/base_layer/wallet_ffi/src/error.rs
@@ -100,7 +100,7 @@ impl From<InterfaceError> for LibWalletError {
     }
 }
 
-/// This implementation maps the internal WalletError to a set of LibWalletErrors. The mapping is explicitly manager
+/// This implementation maps the internal WalletError to a set of LibWalletErrors. The mapping is explicitly managed
 /// here and error code 999 is a catch-all code for any errors that are not explicitly mapped
 impl From<WalletError> for LibWalletError {
     fn from(w: WalletError) -> Self {
@@ -257,8 +257,8 @@ impl From<WalletError> for LibWalletError {
     }
 }
 
-/// This implementation maps the internal HexError to a set of LibWalletErrors. The mapping is explicitly managed
-/// here and error code 999 is a catch-all code for any errors that are not explicitly mapped
+/// This implementation maps the internal HexError to a set of LibWalletErrors.
+/// The mapping is explicitly managed here.
 impl From<HexError> for LibWalletError {
     fn from(h: HexError) -> Self {
         error!(target: LOG_TARGET, "{}", format!("{:?}", h));
@@ -279,8 +279,8 @@ impl From<HexError> for LibWalletError {
     }
 }
 
-/// This implementation maps the internal ByteArrayError to a set of LibWalletErrors. The mapping is explicitly manager
-/// here and error code 999 is a catch-all code for any errors that are not explicitly mapped
+/// This implementation maps the internal ByteArrayError to a set of LibWalletErrors.
+/// The mapping is explicitly managed here.
 impl From<ByteArrayError> for LibWalletError {
     fn from(b: ByteArrayError) -> Self {
         error!(target: LOG_TARGET, "{}", format!("{:?}", b));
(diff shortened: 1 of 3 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1365,6 +1422,159 @@ pub unsafe extern "C" fn completed_transaction_get_destination_public_key(
...
+    let mut error = 0;
+    ptr::swap(error_out, &mut error as *mut c_int);
+    if transaction.is_null() {
+        error = LibWalletError::from(InterfaceError::NullError("transaction".to_string())).code;
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+
+    // check the tx is not in pending state
+    if matches!((*transaction).status, TransactionStatus::Pending) {
```
philipr-za: I just realized that you must also check if it has the Imported status, that means a UTXO was imported and faux transaction was created to document the event but it wont have a formal transaction behind it.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1365,6 +1422,159 @@ pub unsafe extern "C" fn completed_transaction_get_destination_public_key(
...
```
philipr-za: Check for imported status as discussed above

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1365,6 +1422,159 @@ pub unsafe extern "C" fn completed_transaction_get_destination_public_key(
...
```
philipr-za: Check for Imported as above

philipr-za (commented):
Looks really good, I just realized we missed an edge case to check for when we initially discussed this.

base_layer/wallet_ffi/src/error.rs
```diff
@@ -363,3 +363,29 @@ impl From<SchnorrSignatureError> for LibWalletError {
...
+
+/// This implementation maps the internal TransactionError to a set of LibWalletErrors.
+/// The mapping is explicitly managed here.
+impl From<TransactionError> for LibWalletError {
+    fn from(v: TransactionError) -> Self {
+        error!(target: LOG_TARGET, "{}", format!("{:?}", v));
+        match v {
+            TransactionError::StatusError(_) => Self {
+                code: 640,
+                message: format!("{:?}", v),
```
sdbondi: Since `TransactionError` derives from `thiserror`, use Display format.

```suggestion
                message: v.to_string()
```

(trait `ToString` is implemented for all `T: Display`)

base_layer/wallet_ffi/src/error.rs
```diff
@@ -363,3 +363,29 @@ impl From<SchnorrSignatureError> for LibWalletError {
...
+    StatusError(String),
+    #[error("The transaction has the wrong number of kernels: `{0}`")]
+    KernelError(String),
+}
+
+/// This implementation maps the internal TransactionError to a set of LibWalletErrors.
+/// The mapping is explicitly managed here.
+impl From<TransactionError> for LibWalletError {
+    fn from(v: TransactionError) -> Self {
+        error!(target: LOG_TARGET, "{}", format!("{:?}", v));
```
sdbondi:
```suggestion
        error!(target: LOG_TARGET, "{}", v);
```
However, I don't think an `error!` log level makes sense for all calls to this function. It feels like an unexpected side effect to log in a conversion function, so I'd say rather log at the call site.

sdbondi (commented):
Looks good (untested) - just some minor comments

base_layer/wallet_ffi/src/lib.rs
delta1: Thanks Phil - I'll add that `Imported` status

base_layer/wallet_ffi/src/lib.rs
delta1: will fix

base_layer/wallet_ffi/src/lib.rs
delta1: will fix

base_layer/wallet_ffi/src/error.rs
delta1: Thanks @sdbondi will do

base_layer/wallet_ffi/src/error.rs
delta1: Thanks, it's done similarly for other errors. @philipr-za what do you suggest to do here?

delta1: It's done the same way for all the other errors here in wallet_ffi, so will keep it consistent for now. If they all need to be refactored then I'll add that in a separate PR.

philipr-za: So the format was left over from the old error derive macro we used so that can definitely go. 

I am not sure I agree about the logging here, these errors are all show stoppers for the FFI call that elicits them and indicate that the client library is doing something wrong, I think those should be logged at the highest level and are not worth trying to get too specific about.

philipr-za (dismissed):
LGTM

delta1:
sorry - `cargo fmt`

delta1:
RIP cargo clippy 😅

philipr-za (approved):

PR #2231, opened 2020-09-16
Implement algo split as hashmap of timestamps

SWvheerden:
## Description

This is a additional implementation of PR #2170 
While 2170 uses the percentage target and the chain target as input parameters. This PR just uses individual target_time of each algorithm doing away with chain target block time entirely. This PR does not use any math but only provides the values as set in code. 

## Motivation and Context

As part of the discussion on IRC it was noted that there was a desire to favour monero blocks over pure tari blocks. This PR allows the user to specify what percentage the POW should be divided as in 60-40, 50-50 etc.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -57,12 +52,24 @@ pub struct ConsensusConstants {
...
+    /// This keeps track of the block split targets and which algo is accepted
+    /// Ideally this should count up to 100. If this does not you will reduce your target time.
+    algo_split: HashMap<PowAlgorithm, Algo_Time>,
+}
+
+// This is just a convenience  wrapper to put all the info into a hashmap per diff algo
+#[derive(Clone)]
+pub struct Algo_Time {
+    // NB this is very important to set this as 6 * the target time
+    pub max_target_time: u64,
```
stringhandler, line 70: I'm assuming this is for the LWMA. Maybe it should be removed here and put into the LWMA logic. I have a feeling I was the one who moved it from there to the consensus constants, but I think calling it max_target_time sounds like it has meaning outside of just the LWMA

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -57,12 +52,24 @@ pub struct ConsensusConstants {
...
     /// This is the maximum age a monero merge mined seed can be reused
     max_randomx_seed_height: u64,
+    /// This keeps track of the block split targets and which algo is accepted
+    /// Ideally this should count up to 100. If this does not you will reduce your target time.
+    algo_split: HashMap<PowAlgorithm, Algo_Time>,
+}
+
+// This is just a convenience  wrapper to put all the info into a hashmap per diff algo
+#[derive(Clone)]
+pub struct Algo_Time {
```
stringhandler: nit: I think it's better not to have inconsistent abbreviations (Algo vs Algoritm)
```suggestion
pub struct PowAlgorithmConstants {
```

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -57,12 +52,24 @@ pub struct ConsensusConstants {
...
     pub(in crate::consensus) emission_tail: MicroTari,
-    /// This is the initial min difficulty for the difficulty adjustment
-    min_pow_difficulty: (Difficulty, Difficulty),
     /// The offset relative to the expected genesis coinbase value
     genesis_coinbase_value_offset: MicroTari,
     /// This is the maximum age a monero merge mined seed can be reused
     max_randomx_seed_height: u64,
+    /// This keeps track of the block split targets and which algo is accepted
+    /// Ideally this should count up to 100. If this does not you will reduce your target time.
+    algo_split: HashMap<PowAlgorithm, Algo_Time>,
```
stringhandler: nit: `algo_split` implies that this struct only contains information about the split, when it contains other info as well
```suggestion
    proof_of_work: HashMap<PowAlgorithm, PowAlgorithmConstants>,
```

stringhandler (commented):
Some nits. `Algo_Time` is not standard naming (should be `AlgoTime`), so even if you don't use the name I suggested that needs to be fixed. I think the repeated comment warnings (e.g. `// NB this is very important to set this as 6 * the algo target time`) would be more effective as a single doc comment in the module or for the struct explaining how to choose good settings. This also allows more space to explain the reason for setting it to e.g. `target_time * block_window / 20`. That reasoning is not clear here at all.

Also, if the comments are repeated, there is always a chance that some of them will become stale and then will be misleading and confusing when there are two versions of what it is very important to set it to.

As I stated, the max_target_time is a field that is only used in LWMA, and should perhaps be calculated inside the LWMA as `target_time * 6`. Alternatively, add a constant `lwma_max_block_time_factor = 6` as a consensus constant, and then calculate it as `target_time * lwma_max_block_time_factor`

base_layer/core/src/consensus/consensus_constants.rs
SWvheerden: Its totally there for LWMA. But we cant change it here, as its currently used as a *6 for monero and *60 for blake. So we need to keep it separate if we want the code to work for the current test net

stringhandler (dismissed):

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -161,69 +169,91 @@ impl ConsensusConstants {
...
+        algos.insert(PowAlgorithm::Blake, PowAlgorithmConstants {
+            max_target_time: 720,
+            min_difficulty: 1.into(),
+            target_time: 120,
+        });
+        algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+            max_target_time: 720,
+            min_difficulty: 1.into(),
+            target_time: 120,
+        });
```
hansieodendaal, line 308: I think these values should be based on a simple calculation to show where they come from, as it would make future code maintenance a lot easier, i.e. 
```
target_block_interval = 60

// Algo 1 - Blake
block_percentage = 50
target_time = target_block_interval * 100 / block_percentage // 120
max_target_time = target_time * 6 // 720

// Algo 2 - Monero
block_percentage = 50
target_time = target_block_interval * 100 / block_percentage // 120
max_target_time = target_time * 6 // 720
```

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -161,69 +169,91 @@ impl ConsensusConstants {
...
+        algos.insert(PowAlgorithm::Blake, PowAlgorithmConstants {
+            max_target_time: 1800,
+            min_difficulty: 60_000_000.into(),
+            target_time: 300,
+        });
+        algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+            max_target_time: 800,
+            min_difficulty: 60_000_000.into(),
+            target_time: 200,
+        });
```
hansieodendaal, line 339: In line with the above comment:
```
target_block_interval = 120

// Algo 1 - Blake
block_percentage = 40
target_time = target_block_interval * 100 / block_percentage // 300
max_target_time = target_time * 6 // 1800

// Algo 2 - Monero
block_percentage = 60
target_time = target_block_interval * 100 / block_percentage // 200
max_target_time = target_time * 6 // 1200
```

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -161,69 +169,91 @@ impl ConsensusConstants {
...
         let difficulty_block_window = 90;
+        let difficulty_block_window = 90;
+        let mut algos = HashMap::new();
+        algos.insert(PowAlgorithm::Blake, PowAlgorithmConstants {
+            max_target_time: 1800,
+            min_difficulty: 60_000_000.into(),
+            target_time: 300,
+        });
+        algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+            max_target_time: 800,
```
hansieodendaal, line 336: I believe this value is wrong; should it not be `200 * 6 = 1200`?
(case in point)

hansieodendaal (commented):
I think some of these key values should be based on a simple calculation to show where they come from, as it would make future code maintenance a lot easier and minimise errors.

stringhandler (approved):

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: Hi @mikethetike and @SWvheerden, I see this PR has been merged without fixing the error above or commenting on it?

SWvheerden: These are place holder mainnet values, and are not used currently. 
If you try to use the mainnet network type, the code gives out a panic

hansieodendaal: The point I am trying to make is that the value is wrong and should be corrected

PR #1986, opened 2020-06-12
Propagate block hashes instead of full blocks (network breaking)

sdbondi:
## Description

- Nodes propagate a block hash (`NewBlock` message = 33 bytes) instead
  of the full block.  This will significantly reduce network bandwidth
  usage for block propagation.
- On receipt of a block hash a node checks if it has the block.
  If so, it simply ignores the message. Otherwise, it requests the
  full block from the peer that sent the `NewBlock` message. If the
  block is valid and has been added to the node's blockchain db,
  the node propagates the block hash message to other peers. 
- Change block propagate test to tests a few block hashes being propagated rather
  than a single block.
- Changed invalid block test to check that an invalid block is not
  propagated.
- New test to check that an invalid block hash is not propagated
- Cleaned up some necessary generics on structs

## Motivation and Context

Significantly reduce the bandwidth required for block propagation. 
The benefits are twofold:
1. Outgoing: The size of the message that is propagated in one round is many orders of magnitude smaller
1. Incoming: Previously, a node may receive a full block many times. With this change a node will receive the block exactly once.

Possible issues:
1. The `InboundNodeCommsHandler::handle_new_block_message` function is able to correctly handle duplicates by sequentially querying all nodes that sent the block hash message. However, due to message dedup, the message is only received once. If the full block is not returned from the source node (for whatever reason) then the node will miss the block and have to transition to block sync state to catch up.  

## How Has This Been Tested?

- Invalid block propagation test changed to propagate a hash that exists and have Alice return an invalid block (or rather a block that is invalid to Bob and Carol).
- Test the behaviour when a block hash is propagated for which Alice cannot return a block. 

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [x] Breaking change (fix or feature that would cause existing functionality to change)
* [x] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #2232, opened 2020-09-16
Propagate block hashes to all connected nodes

sdbondi:
## Description

This PR changes the (previously unused) `Flood` broadcast strategy to send a message to
every _connected_ peer node. Since PR #1986, block hashes
are propagated across the network and the full block is only downloaded
once per peer (push became push-pull gossip). This large drop in propagation
overhead means that we can afford to spread the hash a little more
widely and use the modified `Flood` strategy. Previously, the hash propagation 
used the propagate strategy which only sends to 4 random connected peers.

## Motivation and Context

The flood strategy is changed from sending _all_ valid peers in the peer list, 
to only sending to currently connected peers (inbound and outbound connections). 
As before, client nodes are exempt from flood propagation messages.

Using this strategy for block hash propagation may improve block propagation 
across the network. 

## How Has This Been Tested?

The flood strategy is already partially tested and those tests pass. Existing base node tests pass.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/service/service.rs
+++ b/base_layer/core/src/base_node/service/service.rs
@@ -630,7 +630,7 @@ async fn handle_outbound_block(
 ) -> Result<(), CommsInterfaceError>
 {
     outbound_message_service
-        .propagate(
+        .flood(
             NodeDestination::Unknown,
             OutboundEncryption::None,
             exclude_peers,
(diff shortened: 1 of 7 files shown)


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #2242, opened 2020-09-17
Only update mempool after block sync is complete

sdbondi:
## Description

In block sync:
- Removed the call to submit_block in favour of calling
  `BlockchainDatabase::add_block` directly.
- Don't emit block events during sync.
- Emit BlockSyncCompleted after block sync is completed

In chain storage:
- A single read only atomic reference of a Block is passed around
  instead of multiple in-memory clones of the same Block.

Misc.
- Updated one existing test to check for the new
  `BlockEvent::BlockSyncComplete` event.

**NOTE: The chain metadata that is advertised to the network will not change until sync is complete**

## Motivation and Context

During block sync we don't particularly care maintaining mempool state (@SWvheerden is this correct?). Some of the algorithms in the mempool are expensive to run repeatedly during block sync. With this PR they are only run once block sync is complete.

## How Has This Been Tested?

Base node test. Tested for the correct emitted event. Existing block sync tests pass.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
@@ -21,7 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 use crate::{
     base_node::{
-        comms_interface::{Broadcast, CommsInterfaceError},
+        comms_interface::{BlockEvent, CommsInterfaceError},
         state_machine_service::{
             states::{
                 helpers::{ban_all_sync_peers, ban_sync_peer, request_headers, select_sync_peer},
@@ -36,13 +36,14 @@ use crate::{
         },
     },
     blocks::{blockheader::BlockHeader, Block},
-    chain_storage::{async_db, BlockchainBackend, ChainMetadata, ChainStorageError},
+    chain_storage::{async_db, BlockAddResult, BlockchainBackend, ChainMetadata, ChainStorageError},
 };
 use core::cmp::min;
 use log::*;
 use std::{
     fmt::{Display, Formatter},
     str::FromStr,
+    sync::Arc,
 };
 use tari_comms::{connectivity::ConnectivityError, peer_manager::PeerManagerError};
 use tari_crypto::tari_utilities::{hex::Hex, Hashable};
@@ -106,7 +107,7 @@ impl Display for BlockSyncInfo {
 impl Default for BlockSyncConfig {
     fn default() -> Self {
         Self {
-            sync_strategy: BlockSyncStrategy::ViaBestChainMetadata(BestChainMetadataBlockSyncInfo),
+            sync_strategy: BlockSyncStrategy::ViaBestChainMetadata(BestChainMetadataBlockSync),
             max_metadata_request_retry_attempts: MAX_METADATA_REQUEST_RETRY_ATTEMPTS,
             max_header_request_retry_attempts: MAX_HEADER_REQUEST_RETRY_ATTEMPTS,
             max_block_request_retry_attempts: MAX_BLOCK_REQUEST_RETRY_ATTEMPTS,
@@ -119,7 +120,7 @@ impl Default for BlockSyncConfig {
 
 #[derive(Clone, Copy, Debug)]
 pub enum BlockSyncStrategy {
-    ViaBestChainMetadata(BestChainMetadataBlockSyncInfo),
+    ViaBestChainMetadata(BestChainMetadataBlockSync),
     ViaRandomPeer(ForwardBlockSyncInfo),
 }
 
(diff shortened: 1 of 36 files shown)


base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -335,6 +336,11 @@ async fn synchronize_blocks<B: BlockchainBackend + 'static>(
                 request_and_add_blocks(shared, sync_peers, block_nums).await?;
                 sync_height += block_nums_count;
             }
+            let metadata = async_db::get_chain_metadata(shared.db()).await?;
+            let last_block = async_db::fetch_block(shared.db(), metadata.height_of_longest_chain()).await?;
+            shared
+                .local_node_interface
+                .publish_block_event(BlockEvent::BlockSyncComplete(Arc::new(last_block.block)));
```
SWvheerden, line 343: This is a problem for clearing out the mempool. 
We need to send through all the blocks we added, not just the last one. 
If we downloaded more than 1 block, only the last one will be cleared out. 

From the workings of the chain, a chain_reorg happend, that logic handles more than one block. We need to follow that logic here.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -438,43 +444,67 @@ async fn request_and_add_blocks<B: BlockchainBackend + 'static>(
...
             }
 
             shared.publish_event_info();
-            match shared
-                .local_node_interface
-                .submit_block(block.clone(), Broadcast::from(false))
-                .await
-            {
-                Ok(_) => {
+            match async_db::add_block(shared.db.clone(), block.clone()).await {
```
SWvheerden, line 462: I think this is a move in the wrong direction for code maintainability. The code had this in the past, and we moved away from this because there is certain logic that needs to be handled when a block is added. 
Now we just add blocks and its up to the "user" to ensure the logic is handled correctly. As far as I read the code, now I think we already have a bug in the logic. I think we should just have fn submit_block(), and all logic should reside in there. That way we cannot forgot to notify a client, notify them in the correct way etc.

SWvheerden (commented):
I like the refactor here of the block, But I think the move here is wrong way. I left a comment way.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
sdbondi: Ye I thought so. This mempool logic really kills the node during block sync - perhaps we can reconcile the mempool after sync.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
stringhandler: Disagree. The handling of the result below is an indication that different parts of the code will handle this event differently. Which parts of the code require handling on a block added? Perhaps those are too tightly coupled.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
stringhandler: I think reconciling the mempool afterwards makes more sense. Events are not guaranteed delivery. A more robust approach would be to only reconcile when you know the best chain data is in the correct state

base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -119,10 +119,10 @@ make_async!(fetch_mmr_proof(tree: MmrTree, pos: usize) -> MerkleProof, "fetch_mm
 make_async!(fetch_mmr_root(tree: MmrTree) -> HashOutput, "fetch_mmr_root");
 make_async!(insert_mmr_node(tree: MmrTree, hash: Hash, deleted: bool) -> (), "insert_mmr_node");
 make_async!(validate_merkle_root(tree: MmrTree, height: u64) -> bool, "validate_merkle_root");
-make_async!(rewind_to_height(height: u64) -> Vec<Block>, "rewind_to_height");
+make_async!(rewind_to_height(height: u64) -> Vec<Arc<Block>>, "rewind_to_height");
 
 //---------------------------------- Block --------------------------------------------//
-make_async!(add_block(block: Block) -> BlockAddResult, "add_block");
+make_async!(add_block(block: Arc<Block>) -> BlockAddResult, "add_block");
```
stringhandler, line 125:
```suggestion
make_async!(add_block(block: &Block) -> BlockAddResult, "add_block");
```

stringhandler (commented):
Looks good. Will test locally

stringhandler (dismissed):

base_layer/core/src/chain_storage/async_db.rs
sdbondi: The lifetime of the reference won't be 'static and that is required by `task::spawn`

sdbondi:
Fixed conflict

stringhandler (approved):
Test locally and working

PR #2243, opened 2020-09-18
Delay UTXO validation process until after initial blockchain sync has been achieved

hansieodendaal:
## Description
- Delay UTXO validation process until after initial blockchain sync has been achieved, monitor OMS events to determine success. 
- Squash double event emitted for `OutputManagerError::BaseNodeNotSynced`.
- Fix clippy warnings in the same file.

_**Note:** See review comments in #2216_

## Motivation and Context
If not delayed, UTXO validation will almost never succeed, as the validation is aborted if the node is not synced.

## How Has This Been Tested?

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -21,7 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::miner;
-use futures::future;
+use futures::{future, stream::Fuse, FutureExt, StreamExt};
 use log::*;
 use std::{
     path::Path,
@@ -49,6 +49,7 @@ use tari_comms::{
     tor,
     tor::TorIdentity,
     transports::SocksConfig,
+    types::CommsPublicKey,
     utils::multiaddr::multiaddr_to_socketaddr,
     CommsNode,
     ConnectionManagerEvent,
@@ -108,7 +109,7 @@ use tari_shutdown::ShutdownSignal;
 use tari_wallet::{
     output_manager_service::{
         config::OutputManagerServiceConfig,
-        handle::OutputManagerHandle,
+        handle::{OutputManagerEvent, OutputManagerEventReceiver, OutputManagerHandle},
         protocols::utxo_validation_protocol::UtxoValidationRetry,
         storage::sqlite_db::OutputManagerSqliteDatabase,
         OutputManagerServiceInitializer,
@@ -123,7 +124,6 @@ use tari_wallet::{
 };
 use tokio::{
     runtime,
-    stream::StreamExt,
     sync::{broadcast, broadcast::Sender as syncSender, watch},
     task,
     time::delay_for,
@@ -162,7 +162,7 @@ impl BaseNodeContext {
         let mut rx = miner.get_utxo_receiver_channel();
         rt.spawn(async move {
             info!(target: LOG_TARGET, " ⚒️ Mining wallet ready to receive coins.");
-            while let Some(utxo) = rx.next().await {
+            while let Some(utxo) = StreamExt::next(&mut rx).await {
                 match wallet_output_handle.add_output(utxo).await {
                     Ok(_) => {
                         info!(
@@ -303,7 +303,7 @@ pub async fn configure_and_initialize_node(
     let result = match &config.db_type {
         DatabaseType::Memory => {
             let backend = MemoryDatabase::<HashDigest>::default();
-            let ctx = build_node_context(
+            build_node_context(
                 backend,
                 network,
                 node_identity,
@@ -311,13 +311,12 @@ pub async fn configure_and_initialize_node(
                 config,
                 interrupt_signal,
             )
-            .await?;
-            ctx
+            .await?
         },
         DatabaseType::LMDB(p) => {
             let backend = create_lmdb_database(&p, config.db_config.clone(), MmrCacheConfig::default())
                 .map_err(|e| e.to_string())?;
-            let ctx = build_node_context(
+            build_node_context(
                 backend,
                 network,
                 node_identity,
@@ -325,8 +324,7 @@ pub async fn configure_and_initialize_node(
                 config,
                 interrupt_signal,
             )
-            .await?;
-            ctx
+            .await?
         },
     };
     Ok(result)
@@ -477,64 +475,20 @@ where
         .set_base_node_public_key(base_node_public_key.clone())
         .await
         .expect("Problem setting local base node public key for transaction service.");
-    transaction_service_handle
-        .restart_transaction_protocols()
-        .await
-        .expect("Problem restarting transaction protocols in the Transaction Service");
-    // Only start the transaction broadcast protocols once the local node is synced
+    let oms_handle = wallet_handles
+        .get_handle::<OutputManagerHandle>()
+        .expect("OutputManagerService is not registered");
+    // Only start the transaction broadcast and UTXO validartion protocols once the local node is synced
     let state_machine = base_node_handles
         .get_handle::<StateMachineHandle>()
         .expect("Could not get State Machine handle");
-    task::spawn(async move {
-        let mut status_watch = state_machine.get_status_info_watch();
-        debug!(
-            target: LOG_TARGET,
-            "Waiting for initial sync before restarting transaction protocols."
-        );
-        loop {
-            let bootstrapped = match status_watch.recv().await {
-                None => false,
-                Some(s) => s.bootstrapped,
-            };
-
-            if bootstrapped {
-                let _ = transaction_service_handle
-                    .restart_broadcast_protocols()
-                    .await
-                    .map_err(|e| {
-                        error!(
-                            target: LOG_TARGET,
-                            "Problem restarting broadcast protocols in the Transaction Service"
-                        );
-                        e
-                    });
-
-                let _ = transaction_service_handle
-                    .restart_transaction_protocols()
-                    .await
-                    .map_err(|e| {
-                        error!(
-                            target: LOG_TARGET,
-                            "Problem restarting transaction negotiation protocols in the Transaction Service"
-                        );
-                        e
-                    });
-                break;
-            }
-        }
-    });
-    let mut oms_handle = wallet_handles
-        .get_handle::<OutputManagerHandle>()
-        .expect("OutputManagerService is not registered");
-    oms_handle
-        .set_base_node_public_key(base_node_public_key)
-        .await
-        .expect("Problem setting local base node public key for output manager service.");
-    // Start the Output Manager UTXO Validation
-    oms_handle
-        .validate_utxos(UtxoValidationRetry::UntilSuccess)
-        .await
-        .expect("Problem starting the Output Manager Service Utxo Valdation process");
+    task::spawn(start_transaction_protocols_and_utxo_validation(
+        state_machine,
+        transaction_service_handle,
+        oms_handle,
+        config.base_node_query_timeout.clone(),
+        base_node_public_key.clone(),
+    ));
 
     //---------------------------------- Mining --------------------------------------------//
 
@@ -596,7 +550,7 @@ async fn sync_peers(
     wallet_peer_manager: Arc<PeerManager>,
 )
 {
-    while let Some(Ok(event)) = events_rx.next().await {
+    while let Some(Ok(event)) = StreamExt::next(&mut events_rx).await {
         if let ConnectionManagerEvent::PeerConnected(conn) = &*event {
             if !wallet_peer_manager.exists_node_id(conn.peer_node_id()).await {
                 match base_node_peer_manager.find_by_node_id(conn.peer_node_id()).await {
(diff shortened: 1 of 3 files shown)


applications/tari_base_node/src/builder.rs
```diff
@@ -473,18 +471,60 @@ where
...
-    oms_handle
-        .validate_utxos(UtxoValidationRetry::UntilSuccess)
-        .await
-        .expect("Problem starting the Output Manager Service Utxo Valdation process");
+    task::spawn(async move {
+        let mut status_watch = state_machine.get_status_info_watch();
+        let mut once_only = true;
+        loop {
+            let bootstrapped = match status_watch.recv().await {
+                Some(s) => match s {
```
philipr-za: This inner match is superfluous you can just return s.bootstrapped directly

applications/tari_base_node/src/builder.rs
```diff
@@ -473,18 +471,60 @@ where
...
+                    .set_base_node_public_key(base_node_public_key)
+                    .await
+                    .map_err(|e| {
+                        error!(
+                            target: LOG_TARGET,
+                            "An error occurred with Output Manager Service setting the base node public key: {}", e
+                        );
+                        e
+                    })
+                    .expect("An error occurred with Output Manager Service setting the base node public key.");
```
philipr-za: I am concerned about a panic here, feels unnecessarily dangerous to have tasks panic. So I think log the error and then squash it.

applications/tari_base_node/src/builder.rs
```diff
@@ -473,18 +471,60 @@ where
...
+                    .validate_utxos(UtxoValidationRetry::UntilSuccess)
+                    .await
+                    .map_err(|e| {
+                        error!(
+                            target: LOG_TARGET,
+                            "An error occurred with Output Manager Service starting the UTXO Validation process: {}", e
+                        );
+                        e
+                    })
+                    .expect("An error occurred with Output Manager Service starting the UTXO Validation process");
```
philipr-za: same as above.

applications/tari_base_node/src/builder.rs
```diff
@@ -473,18 +471,60 @@ where
...
+                break;
+            }
+            if once_only {
```
philipr-za: Is a negative log useful? I would just do one log at the top of the task saying that its starting to monitor for sync and then the log that says it was successful tells you everything you need to know.

hansieodendaal: This logic will tell you if the blockchain was not synced when UTXO validation was attempted the first time, as opposed to having that log message every time at startup regardless.

hansieodendaal:
Some flaky tests here:

On local: `test transaction_service::service::test_transaction_cancellation_sqlite_db ... FAILED`
```
---- transaction_service::service::test_transaction_cancellation_sqlite_db stdout ----
thread 'transaction_service::service::test_transaction_cancellation_sqlite_db' panicked at 'alice call wait 1: "wait_call_count timed out before before receiving the expected number of calls. (Expected =
2, Got = 0)"', base_layer\wallet\tests\transaction_service\service.rs:2753:5
```
In GitHub: `test dht_propagate_dedup ... FAILED`
```
---- dht_propagate_dedup stdout ----
thread 'dht_propagate_dedup' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', comms/dht/tests/dht.rs:513:5
```

applications/tari_base_node/src/builder.rs
stringhandler: Seems like this flag is unnecessary, but it's a nit.

stringhandler (dismissed):
Happy in principle, but there are conflicts with #2221 which does a similar thing

hansieodendaal:
> Happy in principle, but there are conflicts with #2221 which does a similar thing

Reworking this to get rid of the conflict. Also picked up another edge case whereby initial sync has been achieved, but by the time the UTXO query is returned, the base node is not synced anymore. Applying a fix for this to only exit the task if a `UtxoValidationSuccess` event is received.

``` 
2020-09-22 13:10:47.760597300 [c::bn::initialization] DEBUG Initial sync achieved and starting with transaction and UTXO validation protocols.
2020-09-22 13:10:47.974489300 [wallet::output_manager_service::protocols::utxo_validation_protocol] DEBUG Starting UTXO validation protocol (Id: 6169002742224074600) for Unspent Outputs Validation
2020-09-22 13:11:11.669274800 [wallet::output_manager_service::protocols::utxo_validation_protocol] WARN  Assigned Base Node is not synced to chain tip, aborted UTXO Validation protocol (id: 6169002742224074600)
```

applications/tari_base_node/src/builder.rs
```diff
@@ -122,7 +122,7 @@ use tari_wallet::{
 };
 use tokio::{
     runtime,
-    stream::StreamExt,
+    stream::StreamExt as tokioStreamExt,
```
philipr-za: pretty sure we dont need to use both this and the one from futures

philipr-za (commented):
This has gotten long enough that it definitely deserves its own function rather than being in the middle of the builder.

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
                         error!(
                             target: LOG_TARGET,
-                            "Problem restarting transaction negotiation protocols in the Transaction Service"
+                            "Problem restarting transaction negotiation protocols in the Transaction Service: {}", e
                         );
                         e
                     });
+
+                #[derive(PartialEq, Debug)]
+                enum ValidationTask {
```
stringhandler: I think having an enum here in the method is a sign that this logic should be separated into simpler methods.

stringhandler (commented):
IMHO This seems like a lot of code for a specific case of the wallet that bundled with the base node, which is not always going to be the case. I am happy to merge it if you feel strongly that it improves, but I must insist on the logic being separated into, perhaps, two smaller methods to reduce the complexity.

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
+                                        trace!(
+                                            target: LOG_TARGET,
+                                            "UTXO validation unknown event, waiting.",
+                                        );
+                                        continue;
+                                    },
+                                }
+                            },
+                            // Restart the protocol if it timed out
+                            () = delay => {
```
philipr-za: This is actually unnecessary, the UTXO validation protocols will timeout on their own and you can do what you do with this branch above in the `OutputManagerEvent::UtxoValidationTimedOut =>` branch

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
+                    InvalidOutputs,
+                    UnspentOutputs,
+                }
+                let mut event_stream = oms_handle.get_event_stream_fused();
+                let mut validation = ValidationTask::InvalidOutputs;
+                let mut end_protocol = false;
+                let mut delay = delay_for(config_base_node_query_timeout_clone * 2).fuse();
+                loop {
+                    match validation {
+                        ValidationTask::InvalidOutputs => {
```
philipr-za: Revalidating the invalid UTXO's is not a critical task. This is why it doesn't have its own explicit API just yet. I don't think you need to explicitly wait for it to complete, this is partly why this code is so complex. If you want to do it explicitly then it should be removed from the `set_base_node_public_key` method and made explicit so you can monitor the event bus for it but I do not think this is needed. Don't wait for it to complete after setting the public key.

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
+                                    error!(
+                                        target: LOG_TARGET,
+                                        "Problem with Output Manager Service setting the base node public key: {}", e
+                                    );
+                                    e
+                                });
+                            trace!(target: LOG_TARGET, "Attempting UTXO validation for Invalid Outputs.",);
+                        },
+                        ValidationTask::UnspentOutputs => {
+                            let _ = oms_handle
```
philipr-za: So you are throwing away the ID of the validation request here. The number returned by this call is what you should use to differentiate between the OutputManagerEvents while you wait this protocol to resolve. This way you don't have to be careful to only start one protocol at a time because you can tell which events are the ones you care about using this ID.

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
+                                        );
+                                        break;
+                                    },
+                                    // Exit upon success
+                                    OutputManagerEvent::UtxoValidationSuccess(s) => {
+                                        trace!(
+                                            target: LOG_TARGET,
+                                            "UTXO validation event 'UtxoValidationSuccess' ({}).", s,
+                                        );
+                                        if validation == ValidationTask::InvalidOutputs {
```
philipr-za: I don't think you need to do this. You only care able the Unspent outputs validation.

applications/tari_base_node/src/builder.rs
```diff
@@ -495,43 +493,155 @@ where
...
+                                            validation = ValidationTask::UnspentOutputs;
+                                        } else {
+                                            end_protocol = true;
+                                        }
+                                        break;
+                                    },
+                                    // Wait for the next event if timed out (several can be attempted)
+                                    OutputManagerEvent::UtxoValidationTimedOut(s) => {
```
philipr-za: Use this to decide when to restart and also check that `s` is the ID of the validation protocol you care about.

applications/tari_base_node/src/builder.rs
```diff
@@ -474,19 +472,19 @@ where
...
-    // Only start the transaction broadcast protocols once the local node is synced
+    let oms_handle = wallet_handles
+        .get_handle::<OutputManagerHandle>()
+        .expect("OutputManagerService is not registered");
+    // Only start the transaction broadcast and UTXO validartion protocols once the local node is synced
     let state_machine = base_node_handles
         .get_handle::<StateMachineHandle>()
         .expect("Could not get State Machine handle");
+    let config_base_node_query_timeout_clone = config.base_node_query_timeout.clone();
     task::spawn(async move {
```
philipr-za: So instead of spawning an `async{ ... }` block here put all this code into an async function and then you can just call spawn on that function.

applications/tari_base_node/src/builder.rs
hansieodendaal: Moving this logic out to a separate function as @philipr-za suggested.

Note that the enum is to prevent duplicating code just below starting here.
``` rust
                loop {
                    futures::select! {
```

philipr-za: I think the point is that with the correct use of a function the code would not be duplicated, but also I think you only need to monitor for the Unspent validation anyway.

applications/tari_base_node/src/builder.rs
hansieodendaal: This is a safety net. In one of my system-level tests, the protocol never returned with another event and the task got stuck in `futures::select! {`

applications/tari_base_node/src/builder.rs
hansieodendaal: I do not agree that only the unspent UTXOs validation should be monitored; both tasks are important, one never knows how many invalid UTXOs could be valid due to a reorg, and can be perfectly well monitored as the code stands. If removing revalidating invalid UTXOs from the `set_base_node_public_key` method is important, that could be another PR.

applications/tari_base_node/src/builder.rs
hansieodendaal: See comment above.

applications/tari_base_node/src/builder.rs
hansieodendaal: See comment above

applications/tari_base_node/src/builder.rs
hansieodendaal: No, the monitoring loop should not restart with this event, as it is a valid event that can occur in the validation protocol. The code should just wait for the next event.

applications/tari_base_node/src/builder.rs
hansieodendaal: Agreed

applications/tari_base_node/src/builder.rs
philipr-za: That is because you are ignoring the ID that is contained in `s` and are worried that you will get a timeout event for a different protocol. You are creating your own timeout when a timeout already exists for the protocol.

applications/tari_base_node/src/builder.rs
philipr-za: It should not. The validation protocols have a built in timeout and the issue is you are not using this instance of the validation protocol's timeout.

hansieodendaal:
> IMHO This seems like a lot of code for a specific case of the wallet that bundled with the base node, which is not always going to be the case. I am happy to merge it if you feel strongly that it improves,

UTXO validation for base node wallets almost never succeed after startup with the current implementation due to 'BaseNodeNotSynced' / `OutputManagerEvent::UtxoValidationAborted`. So this PR fixes a system-level issue.

> but I must insist on the logic being separated into, perhaps, two smaller methods to reduce the complexity.

Logic moved to a separate function

applications/tari_base_node/src/builder.rs
philipr-za: This code is clunky because of this issue. If you really feel that it is important enough to be monitored here then you should make the change and use the protocol IDs to properly monitor these instances of the validation protocols. It is part of the reason why you have had to introduce an extraneous delay to this code that is absolutely not required.

applications/tari_base_node/src/builder.rs
hansieodendaal: Even if the `futures::select! {` delay is removed, the monitoring loop should still not be restarted if an `OutputManagerEvent::UtxoValidationTimedOut` event occurs, it should just wait for the next event.

The logic is:
- Restart the validation protocol:
  - `OutputManagerEvent::UtxoValidationAborted`
  - `OutputManagerEvent::UtxoValidationFailure`
- Exit (success) if
  - `OutputManagerEvent::UtxoValidationSuccess`
- Wait for the next event:
  - `OutputManagerEvent::UtxoValidationTimedOut`
  - `OutputManagerEvent::Error`
  - _

philipr-za: but the UtxoValidationTimedOut event is exactly the same thing as your timeout but its built-in to the process being monitored? Why ignore it rather than just use it to restart the process?

applications/tari_base_node/src/builder.rs
hansieodendaal: I would feel uneasy about removing the safety net. The `OutputManagerEvent::UtxoValidationTimedOut` event is used in the logic, see below.

applications/tari_base_node/src/builder.rs
hansieodendaal: I can change the logic, but it will not any more efficient. The safety net delay for the `futures::select! {` loop only comes into play if the validation protocol stops sending events altogether.

applications/tari_base_node/src/builder.rs
hansieodendaal: > I think the point is that with the correct use of a function the code would not be duplicated

Yes, I see the point now, this can be done

applications/tari_base_node/src/builder.rs
hansieodendaal: Fixed

applications/tari_base_node/src/builder.rs
hansieodendaal: Fixed - event timeout restart added in the correct location

philipr-za (dismissed):

stringhandler (approved):

PR #2245, opened 2020-09-18
[GRPC] Implement SearchKernels

delta1:
## Description

This commit adds the ability to search for blocks containing transaction kernels via GRPC.

## Motivation and Context

As part of tari-project/tari#2201 and tari-project/block-explorer-frontend#107 we need a way to search for blocks containing specific transaction kernels. 

## How Has This Been Tested?

Existing tests pass, new GRPC call tested manually 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/src/generated/tari.rpc.rs
+++ b/applications/tari_app_grpc/src/generated/tari.rpc.rs
@@ -399,7 +399,7 @@ pub struct ListHeadersRequest {
 }
 /// The request used for querying blocks in the base node's current best chain. Currently only querying by height is
 /// available. Multiple blocks may be queried.e.g. [189092,100023,122424]. The order in which they are returned is not
-/// guarenteed.
+/// guaranteed.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct GetBlocksRequest {
     #[prost(uint64, repeated, tag = "1")]
@@ -430,7 +430,7 @@ pub struct MetaData {
 /// This is the message that is returned for a miner after it asks for a new block.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct GetNewBlockResult {
-    /// This is the header hash of the complated block
+    /// This is the header hash of the completed block
     #[prost(bytes, tag = "1")]
     pub block_hash: std::vec::Vec<u8>,
     /// This is the completed block
@@ -451,6 +451,12 @@ pub struct MinerData {
     pub reward: u64,
     #[prost(bytes, tag = "4")]
     pub mergemining_hash: std::vec::Vec<u8>,
+}
+/// This is the request type for the Search Kernels rpc
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct SearchKernelsRequest {
+    #[prost(message, repeated, tag = "1")]
+    pub signatures: ::std::vec::Vec<Signature>,
 }
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
 #[repr(i32)]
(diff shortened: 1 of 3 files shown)


applications/tari_app_grpc/src/generated/tari.rpc.rs
```diff
@@ -683,6 +697,20 @@ pub mod base_node_client {
             let path = http::uri::PathAndQuery::from_static("/tari.rpc.BaseNode/GetTipInfo");
             self.inner.unary(request.into_request(), path, codec).await
         }
+
+        #[doc = " Search for blocks containing the specified kernels"]
```
sdbondi, line 693: `///` is the same as `doc` attr 
```suggestion
        /// Search for blocks containing the specified kernels
```

Same thing in a number of other places

sdbondi (commented):
Code looks like it'll work, though we should look into writing tests for our grpc endpoints.

applications/tari_app_grpc/proto/types.proto
```diff
@@ -199,6 +199,12 @@ message Signature {
     bytes signature = 2;
 }
 
+// Excess signature identified by hex strings for the public nonce and signature.
+message ExcessSignature {
```
SWvheerden: Cant you just re-use signature above?
Its the same signature, here you use string, where above it just uses the bytes.

delta1:
agreed on the tests

applications/tari_app_grpc/src/conversions/signature.rs
```diff
@@ -25,6 +25,7 @@ use tari_crypto::tari_utilities::ByteArray;
 
 use crate::tari_rpc as grpc;
 use tari_core::transactions::types::{PrivateKey, PublicKey, Signature};
+use tari_utilities::hex::Hex;
```
SWvheerden: If you reuse signature you should be able to remove this dependancy.

applications/tari_app_grpc/src/generated/tari.rpc.rs
delta1: ah that was generated, i'll check to make sure it's consistent with the rest

SWvheerden (commented):
Looks good, I just have a question regarding the use of the sigature message in the proto file

applications/tari_app_grpc/proto/types.proto
delta1: So I did try that, and was getting some ugly mismatch where the bytes received by the GRPC server would fail the conversion to the keys

delta1: That could be where I was using `grpcurl` and probably testing it incorrectly previously - this did make my testing much easier. 

Besides that, in general this call is going to take the hex strings as inputs - is there a big difference to accepting the hex strings instead of forcing the client to pass in a byte array?

applications/tari_app_grpc/src/generated/tari.rpc.rs
delta1: All the generated implementations in the file end up with the doc attr https://github.com/tari-project/tari/blob/86b4e66b262de07c9f54f6372f7e2240974c0b96/applications/tari_app_grpc/src/generated/tari.rpc.rs#L519

applications/tari_app_grpc/proto/types.proto
SWvheerden: There is going to be a str-to-pubkey conversion on the base_node side doing it this way. 
But if the client is not an actual human,like a wallet he is going to use the bytes. 
Thus you will have a pubkey-to-str-to-pubkey conversion. 
To me that sound like we should only do that if required, hence why I would recommend doing the conversion client side before it sends it over, so there is always only one conversion

delta1: Cool that makes sense - then I'll just need some help figuring out why the keys can't be created from the bytes GRPC is receiving

delta1: for the provided public nonce `72ae4e7972263c031ce1f1569abcc0e953670bbbefa29c196e4d0afcb46f4819 `, GRPC is receiving this byte array (48 items)

```
[239, 102, 158, 225, 238, 253, 239, 109, 186, 221, 205, 55, 213, 199, 181, 127, 94, 122, 245, 166, 220, 115, 71, 189, 231, 126, 187, 209, 182, 219, 121, 246, 182, 245, 205, 125, 233, 238, 29, 209, 167, 220, 111, 142, 159, 227, 205, 125]
```

where our PublicKey as_bytes function produces this byte array (32 items)

```
[114, 174, 78, 121, 114, 38, 60, 3, 28, 225, 241, 86, 154, 188, 192, 233, 83, 103, 11, 187, 239, 162, 156, 25, 110, 77, 10, 252, 180, 111, 72, 25]
```

delta1:
The supplied `public_nonce` and `signature` hashes need to be converted from hex to base64 before submission with GRPC - thanks @mikethetike !

stringhandler (approved):

PR #2251, opened 2020-09-21
Fix possible inflation bug

SWvheerden:
## Description

PR fixes possible inflation bug. 

## Motivation and Context

Currently its possible to create non existent inputs, flag those with coinbase flags and the base_node's will expect them as valid and dont check them. By doing this it is possible to inflate the supply of xTR. Pruned syncing should detect this when they sync, if this happened past  their pruning horizon. Archival nodes wont detect this however. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,7 +36,7 @@ dependencies = [
  "aes",
  "block-cipher",
  "ghash",
- "subtle 2.2.3",
+ "subtle 2.3.0",
 ]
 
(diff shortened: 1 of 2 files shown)


sdbondi (dismissed):
Agree with this - this conditional seems to have been added in error

stringhandler (dismissed):
Good find

PR #2252, opened 2020-09-21
DHT network discovery (peer syncing)

sdbondi:
## Description

The PR introduces peer syncing and network discovery to improve a node's
overall knowledge of the network.

- Added a state machine that is responsible for syncing peers in rounds
  according to the current node state and previous round.
- If the number of peers in the peer manager is below a threshold, the sync will attempt 
  to sync (streamed by RPC) many peers at once. Once the threshold has been reached, 
 a small number of peers are periodically requested from random peers.  
- Event notifies other components when a peer sync round has completed.
- The DHT connectivity actor responds to this event by "refreshing" the
  neighbouring and random peer pools.
- Created base for mocking out RPC calls
- Created DHT mock
- Added some basic tests for network discovery

## Motivation and Context

Although analysis is ongoing, the network segregation issues that have been experienced are likely partially or completely a result of nodes not being aware of other nodes outside of their "area" in metric space. In addition to neighbouring peers, each node maintains a random pool of peers that are used for message propagation and gossip as well as helping to make an eclipse attack more difficult. 

The current join procedure only allows peers to know about their immediate neighbourhood (by design). The nodes selected for the random pool should be selected from outside of the nodes neighbourhood.

## How Has This Been Tested?

Fairly basic functional tests. Tested on local base node. 

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


comms/dht/src/event.rs
```diff
@@ -20,13 +20,19 @@
...
 
+use crate::network_discovery::DhtNetworkDiscoveryRoundInfo;
 use std::sync::Arc;
 use tokio::sync::broadcast;
 
 pub type DhtEventSender = broadcast::Sender<Arc<DhtEvent>>;
 pub type DhtEventReceiver = broadcast::Receiver<Arc<DhtEvent>>;
 
 #[derive(Debug)]
+#[non_exhaustive]
```
stringhandler, line 31: I'm not sure about this. Is it not safer to force a breaking change when new variants are added?

comms/dht/src/network_discovery/mod.rs
```diff
@@ -0,0 +1,762 @@
...
+            "DiscoveryParams({} peer(s) selected, num_peers_to_request = {}, max_accept_closer_peers = {})",
+            self.peers.len(),
+            self.num_peers_to_request,
+            self.max_accept_closer_peers
+        )
+    }
+}
+
+#[derive(Debug)]
+struct Initializing<'a> {
```
stringhandler: It would be helpful to have these in different files to help future source control diffs

comms/dht/src/network_discovery/mod.rs
```diff
@@ -0,0 +1,762 @@
...
+                if stats.has_new_peers() {
+                    self.publish_event(DhtEvent::NetworkDiscoveryPeersAdded(stats.clone()));
+                }
+
+                State::Ready(DiscoveryReady::new(self.context.clone(), stats))
+            },
+            (State::Ready(_), StateEvent::Idle) => State::Waiting(Waiting::new(config.idle_period)),
+            (_, StateEvent::Shutdown) => State::Shutdown,
+            (_, StateEvent::Errored(err)) => {
+                debug!(
```
stringhandler:
```suggestion
                error!(
```

stringhandler: This is the kind of thing you'd want to set up an alert for in the logs. Error makes it easier to find. Warn is probably also fine

stringhandler (dismissed):
Really cool. Just some nits

comms/dht/src/event.rs
sdbondi: I don't think it makes any kind of difference that matters in our case because this library forms part of one codebase - tari_comms_dht is a library, and we're saying here that client code should handle the case where they don't know a particular event (client code must include a `_ => {}` branch) so that more events can be added later without making a breaking change . Generally, users will do this anyway as their component is only interested in a few events but not all - this just forces a user to handle that case (typically by just ignoring the event). 

Does the use of non_exhaustive here increase the likelihood of bugs being introduced down the line? I'd say it's neutral.

comms/dht/src/network_discovery/mod.rs
sdbondi: This was bothering me too :) Will make the change

comms/dht/src/network_discovery/mod.rs
sdbondi: Agree - I think initially this branch would execute of no sync peers could be selected, but the behaviour had since changed to select other peers and try them.

sdbondi:
Fixed conflicts and a broken test

PR #2257, opened 2020-09-22
Decouple Mempool and BlockchainDatabase

sdbondi:
## Description

The Mempool should be independent of the blockchain database.
The only reason a coupling currently exsits is because the Mempool requires a generic
`B: BlockchainBackend` just to satisfy the `impl Validator<B>` interface
for its validators.  The `Validator` trait leaks `B: BlockchainBackend`.

To fix this:
- `Validator` renamed to `StatefulValidator`
- `StatelessValidator` renamed to `Validator`
- `FullTxValidator` renamed to `TxInternalConsistencyValidator` and
  performs the internal consistency check only (no db required)
- TxInputAndMaturityValidator implements `Validator`
  (previously `StatelessValidator`) and now takes a
  `BlockchainDatabase<B>`
- The two validators are composed (using the `and_then` combinator)
- Removed the `B` generic from `Mempool` and `OrphanPool`.
- Because of this change, a few mempool tests no longer require the creation
  of a blockchain database.

Extras:
- Fixed a panic if the node receives a transaction with zero kernels.
- Updated some async tests to use the tokio test runtime instead of a mix of futures-rs and the test_async helper

## Motivation and Context

Code simplification and decoupling

## How Has This Been Tested?

Rust's type system provides the guarantees that this is correct. Existing tests pass.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/mempool/orphan_pool/orphan_pool_storage.rs
+++ b/base_layer/core/src/mempool/orphan_pool/orphan_pool_storage.rs
@@ -21,7 +21,6 @@
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::{
-    chain_storage::{BlockchainBackend, BlockchainDatabase},
     mempool::orphan_pool::{error::OrphanPoolError, orphan_pool::OrphanPoolConfig},
     transactions::{transaction::Transaction, types::Signature},
     validation::{ValidationError, Validator},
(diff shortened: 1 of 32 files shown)


delta1 (dismissed):
LGTM - `cargo test` passed

delta1 (dismissed):

base_layer/core/src/mempool/orphan_pool/orphan_pool_storage.rs
```diff
@@ -37,50 +36,46 @@ pub const LOG_TARGET: &str = "c::mp::orphan_pool::orphan_pool_storage";
...
             "Inserting tx into orphan pool: {}",
             tx_key.get_signature().to_hex()
         );
         trace!(target: LOG_TARGET, "Transaction inserted: {}", tx);
         let _ = self.txs_by_signature.insert(tx_key, tx, self.config.tx_ttl);
+        Ok(())
     }
 
     /// Insert a set of new transactions into the OrphanPoolStorage
     #[allow(dead_code)]
```
CjS77: Can this be removed now?

CjS77 (dismissed):
Superb.

General comments:
* Don't sneak in other bug fixes unrelated to this PR - those could have been merged quickly while the bigger PR is being reviewed.
* I don't necessarily agree that the name changes of the Validation structs are an improvement, and it made the PR so much bigger.

sdbondi:
@CjS77 Completely agree :) Takes a bit of discipline when you find an unrelated bug to stop where you are and start a new branch from development and make the fix, but important to do so.

The name changes just reflect that the (previously) `StatelessValidator` is the trait that abstracts validation only and doesn't leak anything. i.e. it is not the special case so should be called `Validator` - IMO the (now) `StatefulValidator`  can be deprecated and replaced, but that is up for discussion.

EDIT: The validation structs are renamed to reflect that they now perform a single (as in concern) validation rather than a "full" validation. Validators are composable using `and_then` to achieve the same thing as a "full" validator

base_layer/core/src/validation/traits.rs
```diff
@@ -22,29 +22,29 @@
 
 use crate::validation::{and_then::AndThenValidator, error::ValidationError};
 
-pub type Validator<T, B> = Box<dyn Validation<T, B>>;
-pub type StatelessValidator<T> = Box<dyn StatelessValidation<T>>;
+pub type StatefulValidator<T, B> = Box<dyn StatefulValidation<T, B>>;
+pub type Validator<T> = Box<dyn Validation<T>>;
```
SWvheerden, line 26: I dont think we should rename `StatelessValidator` to just `validator`. I like renaming of `validator` to `StatefulValidator` thou.  Should we not rather stick to exactly clear calls of the names. Ei `StatelessValidator` and `StatefulValidator`

SWvheerden (dismissed):
Looks good. I just think we should rethink the naming of the validator traits

base_layer/core/src/validation/traits.rs
sdbondi: Did you read my comment above? `StatelessValidator` is the "purest" form of a `Validator`- kind of like naming a trait `AnimalessVehicle` :P Stateless has no meaning here, state can be used within a `StatelessValidator`

base_layer/core/src/mempool/orphan_pool/orphan_pool_storage.rs
sdbondi: Think it may as well be - removed

stringhandler (approved):
I'm happy with this. I moved back and forth on where the database reference should be. Having a reference passed into the `validate()` method means that the caller can control the use of the read/write locks and specifically reuse a single read lock. Whereas each `Validator` holding a reference to the `BlockchainDatabase` and requesting `read_access` might end up in unpredictable deadlocks and changes to the database in between validators. Another approach would be to wrap many `StatefulValidators` in a single `(Stateless)Validator` which gets the read lock and passes it to the inner validators. 
In the end I settled on the idea that the MemPool does not directly depend on the database, and that  only the validators do, so it makes more sense to move that dependency into the validator as you have. So untested, but I agree with the approach. 

BTW, I also agree with the naming

PR #2260, opened 2020-09-23
Refactor transaction error return information

hansieodendaal:
## Description
- Refactored transaction error return information to get the actual reason a transaction failed.
- Fixed unit tests.

## Motivation and Context
During stress tests ~10% transactions are not finalized and fail for some reason, without proper insight into the problem. Invrestigating this issue to was found that the `pub fn finalize(&mut self, features: KernelFeatures, factories: &CryptoFactories) -> Result<bool, TPE>` function returned a `false` upon error instead of the error itself.

## How Has This Been Tested?
Unit tests

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/helpers.rs
+++ b/base_layer/core/src/transactions/helpers.rs
@@ -285,8 +285,7 @@ pub fn create_tx(
 
     let mut stx_protocol = stx_builder.build::<Blake256>(&factories).unwrap();
     match stx_protocol.finalize(KernelFeatures::empty(), &factories) {
-        Ok(true) => (),
-        Ok(false) => panic!("{:?}", stx_protocol.failure_reason()),
+        Ok(_0) => (),
         Err(e) => panic!("{:?}", e),
     }
     (
@@ -332,8 +331,7 @@ pub fn spend_utxos(schema: TransactionSchema) -> (Transaction, Vec<UnblindedOutp
     };
     outputs.push(change_output);
     match stx_protocol.finalize(KernelFeatures::empty(), &factories) {
-        Ok(true) => (),
-        Ok(false) => panic!("{:?}", stx_protocol.failure_reason()),
+        Ok(_0) => (),
         Err(e) => panic!("{:?}", e),
     }
     let txn = stx_protocol.get_transaction().unwrap().clone();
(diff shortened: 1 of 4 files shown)


base_layer/core/src/transactions/helpers.rs
```diff
@@ -332,8 +331,7 @@ pub fn spend_utxos(schema: TransactionSchema) -> (Transaction, Vec<UnblindedOutp
     };
     outputs.push(change_output);
     match stx_protocol.finalize(KernelFeatures::empty(), &factories) {
-        Ok(true) => (),
-        Ok(false) => panic!("{:?}", stx_protocol.failure_reason()),
+        Ok(_0) => (),
         Err(e) => panic!("{:?}", e),
```
philipr-za, line 335: I know its outside of your changes but a panic here seems a bit drastic? Shouldn't this error be propagated

base_layer/wallet/tests/transaction_service/service.rs
```diff
@@ -930,7 +930,7 @@ fn test_accepting_unknown_tx_id_and_malformed_reply<T: TransactionBackend + Clon
             futures::select! {
                 event = alice_event_stream.select_next_some() => {
                     if let TransactionEvent::Error(s) = &*event.unwrap() {
-                        if s == &"TransactionError(ValidationError(\"Transaction could not be finalized\"))".to_string() {
+                        if s == &"TransactionProtocolError(TransactionBuildError(InvalidSignatureError))".to_string() {
```
philipr-za, line 933: This is a sign that whoever wrote this initially (yes it was me :P) didn't do this right for maintainability.

philipr-za (approved):
LGTM, one comment on something adjacent to this PR's changes

base_layer/wallet/tests/transaction_service/service.rs
hansieodendaal: All good actually :-)

base_layer/core/src/transactions/helpers.rs
hansieodendaal: This helper function is only used in tests, so maybe all good?

PR #2261, opened 2020-09-23
Improve block validation

SWvheerden:
## Description

Improve validation speed by removing a redundant check
Fix hash check in code where we are suppose to check that there is not a hash clash. 

## Motivation and Context

The first fix fixes a check in the outputs. We already check that the output is in the unspend_txo list. If the output exists there it cannot exist in the spend_txo list. So we dont have to check that it does not. This removes an o(n) loop check on the outputs. 

The second fix fixes the hash_clash check. We need to ensure that a utxo has a unique hash, a clash here can be caused by many factors including but not limited to a actual hash clash, reused blinding factor etc. To do this we need a check that looks that the hash is unique and it needs to be consistent across pruned and archival modes. We therefor search the mmr that pruned nodes also keep to know if the hash is unique or not. Currently the test would only say it exists if it was already spent. This is incorrect, it should always be unique. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/validation/block_validators.rs
+++ b/base_layer/core/src/validation/block_validators.rs
@@ -31,7 +31,7 @@ use crate::{
     consensus::{ConsensusConstants, ConsensusManager},
     transactions::types::CryptoFactories,
     validation::{
-        helpers::{check_achieved_and_target_difficulty, check_median_timestamp, is_stxo},
+        helpers::{check_achieved_and_target_difficulty, check_median_timestamp, is_utxo_in_mmr},
         StatefulValidation,
         Validation,
         ValidationError,
@@ -115,7 +115,7 @@ impl<B: BlockchainBackend> StatefulValidation<Block, B> for FullConsensusValidat
     fn validate(&self, block: &Block, db: &B) -> Result<(), ValidationError> {
         let block_id = format!("block #{} ({})", block.header.height, block.hash().to_hex());
         check_inputs_are_utxos(block, db)?;
-        check_not_stxos(block, db)?;
+        check_outputs_are_not_existing_hashes(block, db)?;
         trace!(
             target: LOG_TARGET,
             "Block validation: All inputs and outputs are valid for {}",
@@ -299,25 +299,15 @@ fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result
     Ok(())
 }
 
-// This function checks that the inputs and outputs do not exist in the STxO set.
-fn check_not_stxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
-    for input in block.body.inputs() {
-        if is_stxo(db, input.hash())? {
-            // we dont want to log this as a node or wallet might retransmit a transaction
-            debug!(
-                target: LOG_TARGET,
-                "Block validation failed due to already spent input: {}", input
-            );
-            return Err(ValidationError::ContainsSTxO);
-        }
-    }
+// This function checks that the outputs do not exist in the STxO set.
+fn check_outputs_are_not_existing_hashes<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
     for output in block.body.outputs() {
-        if is_stxo(db, output.hash())? {
+        if is_utxo_in_mmr(db, output.hash())? {
             debug!(
                 target: LOG_TARGET,
-                "Block validation failed due to previously spent output: {}", output
+                "Block validation failed due to hash already in MMR: {}", output
             );
-            return Err(ValidationError::ContainsSTxO);
+            return Err(ValidationError::ContainsKnownTxO);
         }
     }
     Ok(())
--- a/base_layer/core/src/validation/helpers.rs
+++ b/base_layer/core/src/validation/helpers.rs
@@ -161,6 +161,16 @@ pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, V
     }
 }
 
+pub fn is_utxo_in_mmr<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, ValidationError> {
+    // Check if the UTXO MMR contains the specified UTXO hash, the backend stxo_db is not used for this task as
+    // archival nodes and pruning nodes might have different STXOs in their stxo_db as horizon state STXOs are
+    // discarded by pruned nodes.
+    match db.fetch_mmr_leaf_index(MmrTree::Utxo, &hash)? {
+        Some(_v) => Ok(true),
+        None => Ok(false),
+    }
+}
+
 pub fn is_utxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, ValidationError> {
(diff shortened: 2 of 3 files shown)


sdbondi (dismissed):
Agree with this change (not tested, reviewed and understood)

This adds a requirement/guarantee to implementions of `BlockChainBackend`, namely
`A transaction output MUST be either be spent OR unspent (never both)`

An obvious requirement and one that is currently satisfied.

These changes should provide a good speed boost to `add_block`

base_layer/core/src/validation/block_validators.rs
```diff
@@ -296,19 +296,9 @@ fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result
 }
 
 // This function checks that the inputs and outputs do not exist in the STxO set.
```
stringhandler: Comment looks like it's out of date

base_layer/core/src/validation/helpers.rs
```diff
@@ -156,6 +156,17 @@ pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, V
     }
 }
 
+// We dont care about if its spend or not, we only want to know if its in the mmr.
```
stringhandler: nit: This comment is unnecessary. The name of the method is sufficient

base_layer/core/src/validation/block_validators.rs
```diff
@@ -296,19 +296,9 @@ fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result
...
-fn check_not_stxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
-    for input in block.body.inputs() {
```
stringhandler, line 304: While it's true that in the current order of validators, this check is unnecessary, I don't think this shouldn't be removed. Instead it should be called before `check_inputs_are_utxos`. The reason is that the block may be put into different places depending on the validation result. Example: If a block contains a known double spend, then it is clearly invalid and should be discarded. If the UTXO in the input does not exist, it may be put in the orphan pool.

stringhandler (commented):
I agree with the more concise of checking for duplicate UTXOs, but I think there's still place for checking for STXOs vs non-existent UTXOs and callers may be running different logic depending on the validation error given. I think the input check proportion of `check_not_stxos` should remain and be moved before `check_inputs_are_utxos`

base_layer/core/src/validation/block_validators.rs
SWvheerden: Currently if any of them fail it results in the same location. Double spend and an unknown input result in orphan pool. 
The chance of having the same tx in an orphan pool with a chain split, is close 100%. In this case its the same tx and its not really an invalid block. Its just invalid for this chain. 

All of these checks are for the statefull validator, which checks on the current chain. Orphan pool results are already checked by now so we know these blocks might be valid. If any of these checks fail, we know the blocks are not valid for the current chain.

sdbondi: Just thinking this through: There are two stages to validation: the orphan validators (if validation fails the block will not enter the orphan pool) and the "full" block validator which comes after. A failure here will result in the block not being entered into the main chain (however it is already in the orphan pool at this point).
 
A double spend within a block is checked at the orphan stage by `check_duplicate_transactions_inputs` - if that fails it will not enter the orphan pool. The "inputs are stxos" check happens on the "main chain" stage but I think it's redundant because the database implementations ensure that a given commitment cannot be in two spend states at once, so the UTXO check is sufficient. 

Any optimisations here are a boon for block sync times.

base_layer/core/src/validation/helpers.rs
```diff
@@ -156,6 +156,16 @@ pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, V
     }
 }
 
+pub fn is_utxo_in_mmr<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, ValidationError> {
```
hansieodendaal, line 164: I think this function name is non-intuitive; we may do better to rename it to mimic the functional aspect rather than the implementation aspect.

Update: Retracting comment

base_layer/core/src/validation/helpers.rs
```diff
@@ -156,6 +156,16 @@ pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, V
...
+    // Check if the UTXO MMR contains the specified deleted UTXO hash, the backend stxo_db is not used for this task as
+    // archival nodes and pruning nodes might have different STXOs in their stxo_db as horizon state STXOs are
+    // discarded by pruned nodes.
```
hansieodendaal: This comment does not seem to reflect what happens below; it is an exact copy of the comment for `pub fn is_stxo<T: BlockchainBackend>(db: &T, hash: HashOutput) -> Result<bool, ValidationError> {`

hansieodendaal (commented):
Definitely an efficiency improvement. The only question in my mind is if a double-spend can be detected with cut-through in the same block?

SWvheerden:
> my mind is if a double-spend can be detected with cut-through in the same block?

Yes it can, but thats in stateless validation

hansieodendaal (dismissed):
LGTM

base_layer/core/src/validation/block_validators.rs
```diff
@@ -299,20 +299,10 @@ fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result
...
-                "Block validation failed due to already spent input: {}", input
-            );
-            return Err(ValidationError::ContainsSTxO);
-        }
-    }
+// This function checks that the outputs do not exist in the STxO set.
+fn check_outputs_duplicate_hashes<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
     for output in block.body.outputs() {
-        if is_stxo(db, output.hash())? {
+        if is_utxo_in_mmr(db, output.hash())? {
```
stringhandler, line 305: There error message in line 310 (`return Err(ValidationError::ContainsSTxO);`)  and debug message should be updated to reflect what this check is actually doing, otherwise it will be confusing to debug in future

base_layer/core/src/validation/block_validators.rs
```diff
@@ -299,20 +299,10 @@ fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result
...
-            // we dont want to log this as a node or wallet might retransmit a transaction
-            debug!(
-                target: LOG_TARGET,
-                "Block validation failed due to already spent input: {}", input
-            );
-            return Err(ValidationError::ContainsSTxO);
-        }
-    }
+// This function checks that the outputs do not exist in the STxO set.
+fn check_outputs_duplicate_hashes<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
```
stringhandler:
```suggestion
fn check_outputs_are_not_existing_hashes<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
```

SWvheerden:
This has been fixed in PR #2431

PR #2262, opened 2020-09-24
Implement config option to enable transaction fee to exceed the amount for spending dust

hansieodendaal:
## Description

Preventing fee to be larger than the amount will also prevent spending dust in some cases. Added a config option as a general setting to allow the fee to be larger than the amount in order to enable spending dust for small valued transactions.

## Motivation and Context

See console output where transactions end up pending because a fee larger than the amount is not allowed below. Note the number of inputs (dust) collected to make up the amount and fee.

```
>> list-transactions
Inbound Transactions
>> No pending inbound transactions found.

Outbound Transactions
Transaction ID       | Dest Public Key                                                  | Amount    | Fee       | Status  | Sender State                                          | Timestamp           | Message
4219066494340707534  | 7ad67f1c608d25bf6126f5ce5e44cabc5f1b0464ae14efd1969d52f2dfb5c518 | 10000 µT | 36100 µT | Pending | CollectingSingleSignature(1415 input(s), 1 output(s)) | 2020-09-23 14:34:03 | #Test dust
11827765445046070496 | 7ad67f1c608d25bf6126f5ce5e44cabc5f1b0464ae14efd1969d52f2dfb5c518 | 10000 µT | 10800 µT | Pending | CollectingSingleSignature(416 input(s), 0 output(s))  | 2020-09-23 14:35:33 | #Test dust
```
Error messages form log files (as per #2260):
```
2020-09-23 16:34:03.916777200 [wallet::transaction_service::protocols::send_protocol] INFO  Pending Outbound Transaction TxId: 4219066494340707534 added. Waiting for Reply or Cancellation
2020-09-23 16:34:07.902492900 [wallet::transaction_service::protocols::send_protocol] ERROR Transaction (TxId: 4219066494340707534) could not be finalized. Failure error: ValidationError("Fee is greater than amount")
2020-09-23 16:34:07.903493300 [wallet::transaction_service::service] TRACE Send Protocol for Transaction has ended with result Ok(Err(TransactionServiceProtocolError { id: 4219066494340707534, error: TransactionProtocolError(ValidationError("Fee is greater than amount")) }))
2020-09-23 16:34:07.903493300 [wallet::transaction_service::service] WARN  Error completing Send Transaction Protocol (Id: 4219066494340707534): TransactionProtocolError(ValidationError("Fee is greater than amount"))


2020-09-23 16:35:33.836955500 [wallet::transaction_service::protocols::send_protocol] INFO  Pending Outbound Transaction TxId: 11827765445046070496 added. Waiting for Reply or Cancellation
2020-09-23 16:35:35.183538000 [wallet::transaction_service::protocols::send_protocol] ERROR Transaction (TxId: 11827765445046070496) could not be finalized. Failure error: ValidationError("Fee is greater than amount")
2020-09-23 16:35:35.183538000 [wallet::transaction_service::service] TRACE Send Protocol for Transaction has ended with result Ok(Err(TransactionServiceProtocolError { id: 11827765445046070496, error: TransactionProtocolError(ValidationError("Fee is greater than amount")) }))
2020-09-23 16:35:35.183538000 [wallet::transaction_service::service] WARN  Error completing Send Transaction Protocol (Id: 11827765445046070496): TransactionProtocolError(ValidationError("Fee is greater than amount"))
```

## How Has This Been Tested?

Tested between two base nodes in Windows.

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/transaction_protocol/sender.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/sender.rs
@@ -78,13 +78,6 @@ pub(super) struct RawTransactionInfo {
     pub message: String,
 }
 
-impl RawTransactionInfo {
-    pub fn calculate_total_amount(&self) -> MicroTari {
-        let to_others: MicroTari = self.amounts.iter().sum();
-        to_others + self.amount_to_self
-    }
-}
-
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct SingleRoundSenderData {
     /// The transaction id for the recipient
@@ -342,14 +335,7 @@ impl SenderTransactionProtocol {
     /// Performs sanitary checks on the collected transaction pieces prior to building the final Transaction instance
     fn validate(&self) -> Result<(), TPE> {
         if let SenderState::Finalizing(info) = &self.state {
-            let total_amount = info.calculate_total_amount();
             let fee = info.metadata.fee;
-            // The fee should be less than the amount. This isn't a protocol requirement, but it's what you want 99.999%
-            // of the time, and our users will thank us if we reject a tx where they put the amount in the fee field by
-            // mistake!
-            if fee > total_amount {
-                return Err(TPE::ValidationError("Fee is greater than amount".into()));
-            }
             // The fee must be greater than MIN_FEE to prevent spam attacks
             if fee < MINIMUM_TRANSACTION_FEE {
                 return Err(TPE::ValidationError("Fee is less than the minimum".into()));
@@ -773,5 +759,56 @@ mod test {
                 TransactionProtocolError::ValidationError("Recipient output range proof failed to verify".into())
             ),
         }
+    }
+
+    fn get_fee_larger_than_amount_values() -> (MicroTari, MicroTari, MicroTari) {
+        (MicroTari(2500), MicroTari(51), MicroTari(500))
+    }
+
+    #[test]
+    fn disallow_fee_larger_than_amount() {
+        let factories = CryptoFactories::default();
+        // Alice's parameters
+        let alice = TestParams::new();
+        let (utxo_amount, fee_per_gram, amount) = get_fee_larger_than_amount_values();
+        let (utxo, input) = make_input(&mut OsRng, utxo_amount, &factories.commitment);
+        let mut builder = SenderTransactionProtocol::builder(1);
+        builder
+            .with_lock_height(0)
+            .with_fee_per_gram(fee_per_gram)
+            .with_offset(alice.offset.clone())
+            .with_private_nonce(alice.nonce.clone())
+            .with_change_secret(alice.change_key.clone())
+            .with_input(utxo.clone(), input)
+            .with_amount(0, amount);
+        // Verify that the initial 'fee greater than amount' check rejects the transaction when it is constructed
+        match builder.build::<Blake256>(&factories) {
+            Ok(_) => panic!("'BuildError(\"Fee is greater than amount\")' not caught"),
+            Err(e) => assert_eq!(e.message, "Fee is greater than amount".to_string()),
+        };
+    }
+
+    #[test]
+    fn allow_fee_larger_than_amount() {
+        let factories = CryptoFactories::default();
+        // Alice's parameters
+        let alice = TestParams::new();
+        let (utxo_amount, fee_per_gram, amount) = get_fee_larger_than_amount_values();
+        let (utxo, input) = make_input(&mut OsRng, utxo_amount, &factories.commitment);
+        let mut builder = SenderTransactionProtocol::builder(1);
+        builder
+            .with_lock_height(0)
+            .with_fee_per_gram(fee_per_gram)
+            .with_offset(alice.offset.clone())
+            .with_private_nonce(alice.nonce.clone())
+            .with_change_secret(alice.change_key.clone())
+            .with_input(utxo.clone(), input)
+            .with_amount(0, amount)
+            .with_prevent_fee_gt_amount(false);
+        // Test if the transaction passes the initial 'fee greater than amount' check when it is constructed
+        match builder.build::<Blake256>(&factories) {
+            Ok(_) => assert!(true),
+            Err(e) => panic!("Unexpected error: {:?}", e),
+        };
     }
 }
--- a/base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
@@ -28,6 +28,7 @@ use crate::transactions::{
         TransactionOutput,
         UnblindedOutput,
         MAX_TRANSACTION_INPUTS,
+        MAX_TRANSACTION_OUTPUTS,
         MINIMUM_TRANSACTION_FEE,
     },
     transaction_protocol::{
@@ -38,12 +39,15 @@ use crate::transactions::{
     types::{BlindingFactor, CryptoFactories, PrivateKey, PublicKey},
 };
 use digest::Digest;
+use log::*;
 use std::{
     cmp::max,
     collections::HashMap,
     fmt::{Debug, Error, Formatter},
 };
 use tari_crypto::{keys::PublicKey as PublicKeyTrait, tari_utilities::fixed_set::FixedSet};
+
+pub const LOG_TARGET: &str = "c::tx::tx_protocol::tx_initializer";
 
 /// The SenderTransactionInitializer is a Builder that helps set up the initial state for the Sender party of a new
 /// transaction Typically you don't instantiate this object directly. Rather use
@@ -67,6 +71,7 @@ pub struct SenderTransactionInitializer {
     excess_blinding_factor: BlindingFactor,
     private_nonce: Option<PrivateKey>,
     message: Option<String>,
+    prevent_fee_gt_amount: bool,
 }
 
 pub struct BuildError {
@@ -95,6 +100,7 @@ impl SenderTransactionInitializer {
             private_nonce: None,
             excess_blinding_factor: BlindingFactor::default(),
             message: None,
+            prevent_fee_gt_amount: true,
         }
     }
 
@@ -158,6 +164,12 @@ impl SenderTransactionInitializer {
         self
     }
 
+    /// Enable or disable spending of an amount less than the fee
+    pub fn with_prevent_fee_gt_amount(&mut self, prevent_fee_gt_amount: bool) -> &mut Self {
+        self.prevent_fee_gt_amount = prevent_fee_gt_amount;
+        self
+    }
+
     /// Tries to make a change output with the given transaction parameters and add it to the set of outputs. The total
     /// fee, including the additional change output (if any) is returned along with the amount of change.
     /// The change output **always has default output features**.
@@ -211,6 +223,11 @@ impl SenderTransactionInitializer {
         })
     }
 
+    fn calculate_total_amount(&self, amount_to_self: &MicroTari) -> MicroTari {
+        let to_others: MicroTari = self.amounts.clone().into_vec().iter().sum();
+        to_others + amount_to_self
+    }
+
     /// Construct a `SenderTransactionProtocol` instance in and appropriate state. The data stored
     /// in the struct is _moved_ into the new struct. If any data is missing, the `self` instance is returned in the
     /// error (so that you can continue building) along with a string listing the missing fields.
@@ -223,20 +240,20 @@ impl SenderTransactionInitializer {
         Self::check_value("Missing Fee per gram", &self.fee_per_gram, &mut message);
         Self::check_value("Missing Offset", &self.offset, &mut message);
         Self::check_value("Missing Private nonce", &self.private_nonce, &mut message);
-        if !self.amounts.is_full() {
-            message.push(format!("Missing all {} amounts", self.amounts.size()));
-        }
-        if self.inputs.is_empty() {
-            message.push("Missing Input".to_string());
-        }
-        // Prevent overflow attacks by imposing sane limits on some key parameters
-        if self.inputs.len() > MAX_TRANSACTION_INPUTS {
-            message.push("Too many inputs".into());
-        }
         if !message.is_empty() {
             return self.build_err(&message.join(","));
         }
-        // Everything is here. Let's send some Tari!
+        if !self.amounts.is_full() {
+            let size = self.amounts.size();
+            return self.build_err(&*format!("Missing all {} amounts", size));
+        }
+        if self.inputs.is_empty() {
+            return self.build_err("A transaction cannot have zero inputs");
+        }
+        // Prevent overflow attacks by imposing sane limits on inputs
+        if self.inputs.len() > MAX_TRANSACTION_INPUTS {
+            return self.build_err("Too many inputs in transaction");
+        }
         // Calculate the fee based on whether we need to add a residual change output or not
         let (total_fee, change) = match self.add_change_if_required() {
             Ok((fee, change)) => (fee, change),
@@ -246,7 +263,7 @@ impl SenderTransactionInitializer {
         if total_fee < MINIMUM_TRANSACTION_FEE {
             return self.build_err("Fee is less than the minimum");
         }
-
+        // Create transaction outputs
         let outputs = match self
             .outputs
             .iter()
@@ -258,11 +275,15 @@ impl SenderTransactionInitializer {
                 return self.build_err(&e.to_string());
             },
         };
+        // Prevent overflow attacks by imposing sane limits on outputs
+        if outputs.len() > MAX_TRANSACTION_OUTPUTS {
+            return self.build_err("Too many outputs in transaction");
+        }
 
-        let nonce = self.private_nonce.unwrap();
+        let nonce = self.private_nonce.clone().unwrap();
         let public_nonce = PublicKey::from_secret_key(&nonce);
-        let offset = self.offset.unwrap();
-        let excess_blinding_factor = self.excess_blinding_factor;
+        let offset = self.offset.clone().unwrap();
+        let excess_blinding_factor = self.excess_blinding_factor.clone();
         let offset_blinding_factor = &excess_blinding_factor - &offset;
         let excess = PublicKey::from_secret_key(&offset_blinding_factor);
         let amount_to_self = self.outputs.iter().fold(MicroTari::from(0), |sum, o| sum + o.value);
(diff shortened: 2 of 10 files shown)


base_layer/core/src/transactions/transaction_protocol/proto/transaction_sender.proto
```diff
@@ -17,6 +17,8 @@ message SingleRoundSenderData {
     TransactionMetadata metadata = 5;
     // Plain text message to receiver
     string message = 6;
+    // Allow fee greater than amount
```
CjS77: These feel like two different things (allowing fee > amount and dust collection). Specifically, preventing dust collection is typically a special case of the former, so rather just name and comment using the former.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -99,6 +104,9 @@ pub struct SingleRoundSenderData {
     pub metadata: TransactionMetadata,
     /// Plain text message to receiver
     pub message: String,
+    /// Dust collector enabled/disabled
+    #[serde(default)]
+    pub dust_collector_enabled: bool,
```
CjS77: Is this field something we want to be sending over the wire? It feels like it should be a private setting to each wallet.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -345,10 +354,15 @@ impl SenderTransactionProtocol {
             let total_amount = info.calculate_total_amount();
             let fee = info.metadata.fee;
             // The fee should be less than the amount. This isn't a protocol requirement, but it's what you want 99.999%
-            // of the time, and our users will thank us if we reject a tx where they put the amount in the fee field by
-            // mistake!
+            // of the time, however, preventing this will also prevent spending dust.
             if fee > total_amount {
-                return Err(TPE::ValidationError("Fee is greater than amount".into()));
+                if !info.dust_collector_enabled {
```
CjS77: ` if (!info.dust_collector_enabled && fee > total_amount)..`

and save yourself a nested if

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -777,4 +791,124 @@ mod test {
             ),
         }
     }
+
+    fn get_dust_collector_test_values() -> (MicroTari, MicroTari, MicroTari) {
+        (MicroTari(2500), MicroTari(51), MicroTari(500))
+    }
+
+    #[test]
+    fn dust_collector_disallow_fee_larger_than_amount() {
```
CjS77: Tests could be DRYd up

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -159,6 +159,7 @@ mod test {
             public_nonce: pub_rs.clone(),
             metadata: m.clone(),
             message: "".to_string(),
+            dust_collector_enabled: false,
```
CjS77: Doesn't seem to be used, so would advise removing it as per pvs comment.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -158,6 +160,12 @@ impl SenderTransactionInitializer {
         self
     }
 
+    /// Enable or disable spending of dust less than the fee
+    pub fn with_dust_collector(&mut self, dust_collector_enabled: bool) -> &mut Self {
```
CjS77: You don't need an `Option` here. There is going to be a default, so set the default in the builder constructor and this function can override it if it's called.

common/config/presets/windows.toml
```diff
@@ -85,6 +85,9 @@ base_node_query_timeout = 120
 #transaction_direct_send_timeout = 20
 # This is the timeout period that will be used for sending transactions via broadcast mode (default = 30)
 #transaction_broadcast_send_timeout = 30
+# If a large amount of tiny valued uT UTXOs are used as inputs to a transaction, the fee may be larger than the
+# transaction amount. Enable this setting to allow effective spending of "dust" UTXOs (default = false).
+#dust_collector_enabled = true
```
CjS77: The real reason this setting exists is to prevent "fat fingering" -- if a user accidentally sets the fee in the output value and vice versa, and so you almost always want this to be false. The Dust issue is a special instance of when you'd want to override the setting.

But note that most rational actors are not going to net spend money just to clean out dust UTXOs.

So I would modify the description to reflect the main use case and then indicate the special case, and set the default to true.

common/config/tari_config_sample.toml
```diff
@@ -85,6 +85,9 @@ base_node_query_timeout = 120
 #transaction_direct_send_timeout = 20
 # This is the timeout period that will be used for sending transactions via broadcast mode (default = 30)
 #transaction_broadcast_send_timeout = 30
+# If a large amount of tiny valued uT UTXOs are used as inputs to a transaction, the fee may be larger than the
+# transaction amount. Enable this setting to allow effective spending of "dust" UTXOs (default = false).
+#dust_collector_enabled = true
```
CjS77: As above.

common/src/configuration/global.rs
```diff
@@ -84,6 +84,7 @@ pub struct GlobalConfig {
     pub transaction_chain_monitoring_timeout: Duration,
     pub transaction_direct_send_timeout: Duration,
     pub transaction_broadcast_send_timeout: Duration,
+    pub dust_collector_enabled: bool,
```
CjS77: The global config struct is actually deprecated. There's a better config setup that @dunnock introduced, but iirc only the mempool config has migrated to that structure. Just FYI.

common/src/configuration/utils.rs
```diff
@@ -80,6 +80,7 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
     cfg.set_default("wallet.transaction_direct_send_timeout", 600).unwrap();
     cfg.set_default("wallet.transaction_broadcast_send_timeout", 600)
         .unwrap();
+    cfg.set_default("wallet.dust_collector_enabled", false).unwrap();
```
CjS77: The default and sample configs should align. But `false` is the correct answer here. :)

CjS77 (commented):
Like and support the concept. Had a few comments and minor issues.

SWvheerden (dismissed):
Looks good, I agree with the statements made by @CjS77. I dont think we need to send these settings over the wire. It seems to me that this is something we really only want to set privately. 
The remote wallet should have this set privately. I am not 100% fav with the rest of the code here. But we should be mindfull of dust attacks.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Agreed, had the wrong understanding here. The flag only needs to go from the sender to the base node for final validation, not to the receiver.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Going with this:
``` rust
            if fee > total_amount {
                warn!(
                    target: LOG_TARGET,
                    "Fee ({}) is greater than amount ({}) for Transaction (TxId: {}).", fee, total_amount, info.ids[0]
                );
                if info.prevent_fee_gt_amount{
                    return Err(TPE::ValidationError("Fee is greater than amount".into()));
                }
            }
```

base_layer/core/src/transactions/transaction_protocol/proto/transaction_sender.proto
CjS77: Is github being github again?

I'm still seeing `dust_collector_enabled` when `allow_fee_gt_amount` was the agreed new name

hansieodendaal: Sorry, have not pushed at the time yet.

stringhandler:
This really seems like a check that the UI would do. I think the check should be removed entirely

CjS77:
> This really seems like a check that the UI would do. I think the check should be removed entirely

At Coinbase we saw multiple cases (and possibly missed more) where people got this wrong. So it happens. And UIs aren't where this happens. It's in handcrafted txs. You'll thank the failsafe when it saves you money.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -158,6 +160,12 @@ impl SenderTransactionInitializer {
         self
     }
 
+    /// Enable or disable spending of an amount less than the fee
+    pub fn with_fee_guard(&mut self, prevent_fee_gt_amount: bool) -> &mut Self {
```
stringhandler: nit: Rather keep the name the same so that it's easy to find
```suggestion
    pub fn with_prevent_fee_gt_amount(&mut self, prevent_fee_gt_amount: bool) -> &mut Self {
```

stringhandler (commented):
I think the problem here is that the validation only happens at finalize. I would suggest a better strategy would be to not store `prevent_fee_gt_amount` on `RawTransactionInfo` and instead adding the check in `SenderTransactionInitializer::build`. (There should be a check at that point added anyway). Then remove the check from the validation (finalize) stage. This does two things: 

1. The transaction doesn't go across the wire at all if it's incorrect. This prevent encumbering outputs while you are waiting for negotiation.
2. Anything that is already sent can be received. As it currently stands, you may have the setting set to true, send the transaction and upon finalize, even though the config setting is false, you cannot finalize that transaction.

I'm not sure if it's possible, but the receiver may be able to alter this value in transit, which is a problem I'd rather not invite.

hansieodendaal:
> I think the problem here is that the validation only happens at finalize. I would suggest a better strategy would be to not store `prevent_fee_gt_amount` on `RawTransactionInfo` and instead adding the check in `SenderTransactionInitializer::build`. (There should be a check at that point added anyway). Then remove the check from the validation (finalize) stage.

This is an excellent proposal, saw many remnant pending Txs at the receiver due to this. Will implement. 

_**Edit:** Elected to keep the check in the validation (finalize) stage as well._

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -342,13 +340,18 @@ impl SenderTransactionProtocol {
...
-            let total_amount = info.calculate_total_amount();
+            let total_amount = calculate_total_amount(info.amount_to_self, info.amounts.clone());
             let fee = info.metadata.fee;
             // The fee should be less than the amount. This isn't a protocol requirement, but it's what you want 99.999%
-            // of the time, and our users will thank us if we reject a tx where they put the amount in the fee field by
-            // mistake!
+            // of the time, however, always preventing this will also prevent spending dust in some edge cases.
             if fee > total_amount {
-                return Err(TPE::ValidationError("Fee is greater than amount".into()));
+                warn!(
```
CjS77: Why do this twice?

Are there any code paths that let you do the check here but not in initialisation? (in which case, drop the other check)
or in init but not here (in which case drop this check)

Otherwise drop one of them. If dropping this one, then you can also move the `calculate_total_amount` function to the other file and drop exporting a very niche function, or keep it as a method as it was.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -259,10 +270,10 @@ impl SenderTransactionInitializer {
             },
         };
 
-        let nonce = self.private_nonce.unwrap();
+        let nonce = self.private_nonce.clone().unwrap();
```
CjS77, line 283: Why clone here? And the other places?

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -76,13 +79,8 @@ pub(super) struct RawTransactionInfo {
...
     pub message: String,
-}
-
-impl RawTransactionInfo {
-    pub fn calculate_total_amount(&self) -> MicroTari {
-        let to_others: MicroTari = self.amounts.iter().sum();
-        to_others + self.amount_to_self
-    }
+    #[serde(default)]
+    pub prevent_fee_gt_amount: bool,
```
CjS77: You should be able to remove this entirely.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -76,13 +79,8 @@ pub(super) struct RawTransactionInfo {
     pub recipient_info: RecipientInfo,
     pub signatures: Vec<Signature>,
     pub message: String,
-}
-
-impl RawTransactionInfo {
-    pub fn calculate_total_amount(&self) -> MicroTari {
-        let to_others: MicroTari = self.amounts.iter().sum();
-        to_others + self.amount_to_self
-    }
```
CjS77, line 85: You have access to this function everywhere you have access to `self.amounts` so why remove it? There's no other use case for "adding a MicroTari to a FixedSet of MicroTari"

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -342,13 +340,18 @@ impl SenderTransactionProtocol {
     /// Performs sanitary checks on the collected transaction pieces prior to building the final Transaction instance
     fn validate(&self) -> Result<(), TPE> {
         if let SenderState::Finalizing(info) = &self.state {
-            let total_amount = info.calculate_total_amount();
+            let total_amount = calculate_total_amount(info.amount_to_self, info.amounts.clone());
```
CjS77: The original code is cleaner and faster.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -774,4 +782,86 @@ mod test {
...
+            .unwrap();
+        // Transaction should be complete
+        assert!(alice.is_finalizing());
+        // Fee larger than amount should be allowed
+        assert!(alice.get_fee_amount().unwrap() > utxo_amount - alice.get_fee_amount().unwrap());
+        match alice.finalize(KernelFeatures::empty(), &factories) {
+            Ok(_) => (),
+            Err(e) => panic!("Unexpected error: {:?}", e),
+        };
+    }
```
CjS77: You've already tested what you need to test afaict, so remove all this

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -482,6 +485,11 @@ pub fn calculate_tx_id<D: Digest>(pub_nonce: &PublicKey, index: usize) -> u64 {
     u64::from_le_bytes(bytes)
 }
 
+pub(crate) fn calculate_total_amount(amount_to_self: MicroTari, amounts: Vec<MicroTari>) -> MicroTari {
```
CjS77: Assuming this function lives on (and moves), you don't want to take ownership of `amount` here, otherwise there's no much unnecessary memory copying. Take a reference, either `&[u8]` or `&FixedSet<MicroTari>`

CjS77 (commented):
Getting there.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: There is a specific feature to test in `fn validate(&self)` that is called by `pub fn finalize(&mut self, features: KernelFeatures, factories: &CryptoFactories)`, so that is why I need to test that the code does not panic.
``` rust
   fn validate(&self) -> Result<(), TPE> {
        if let SenderState::Finalizing(info) = &self.state {
            let total_amount = calculate_total_amount(info.amount_to_self, info.amounts.clone());
            let fee = info.metadata.fee;
            // The fee should be less than the amount. This isn't a protocol requirement, but it's what you want 99.999%
            // of the time, however, always preventing this will also prevent spending dust in some edge cases.
            if fee > total_amount {
                warn!(
                    target: LOG_TARGET,
                    "Fee ({}) is greater than amount ({}) for Transaction (TxId: {}).", fee, total_amount, info.ids[0]
                );
                if info.prevent_fee_gt_amount {
                    return Err(TPE::ValidationError("Fee is greater than amount".into()));
                }
            }
```

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: This was a point of contention in my mind if it is possible to construct a transaction without calling ` pub fn build<D: Digest>(mut self, factories: &CryptoFactories)` (as in some tests), or alternatively, to somehow alter the data after the initial send to the receiver, so I opted to be conservative and keep the sanity check in both places similar to the other sanity checks (`MAX_TRANSACTION_INPUTS`, `MINIMUM_TRANSACTION_FEE`).

I am quite happy to drop the sanity check in `fn validate(&self)` if you feel strongly about it (and I guess I then need to drop the other duplicate sanity checks as well).

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: The final call in line 307 to `return self.build_err("Fee is greater than amount");` creates use of moved value to `self` in all those places.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Pending your answer on my initial comment

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Agreed, reverting to that

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Agreed, reverting to that

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Noted, thanks. Reverting to the former.

base_layer/core/src/transactions/transaction_protocol/sender.rs
CjS77: I think because it's a fat-finger check and not a consensus critical check, we should just do it once, and before things hit the wire as Mike said (I thought that both calls were before transmission initially, so I didn't bring it up in my previous reviews).

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
CjS77: Oh, then see if you can move that check earlier before the the first reference to `self`.

CjS77:
Not for this PR, but I think some of the other checks in `validation` could be done before anything hits the wire as well,  minimum fee for instance.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Keeping it in the builder only then

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: Optimized

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Removed latter part of the test in line with change in code

hansieodendaal:
> Not for this PR, but I think some of the other checks in `validation` could be done before anything hits the wire as well, minimum fee for instance.

Sanity checks for these values were already done - `MAX_TRANSACTION_INPUTS` and `MINIMUM_TRANSACTION_FEE` - added `MAX_TRANSACTION_OUTPUTS`.

stringhandler (approved):

PR #2264, opened 2020-09-28
Implement STXO validation in the wallet’s Output Manager Service

philipr-za:
## Description
Currently the Output Manager Service in the wallet has 2 types of Output validation:
1. Validation Unspent outputs to see if they are still valid spendable outputs
2. Revalidate outputs marked as Invalid (from the previous validation

This PR adds a third type of validation which checks the wallet’s Spent outputs to see if any of them have become Unspent since the last time the wallet started up. This is done because it is possible that during a Reorg a transaction that the wallet thought was successful could be rolled back. In that case currently the wallet will never restore those funds. This validation will be able to restore the outputs spent in such a situation to spendable, returning them to the wallets Available balance.

A third kind TxoValidationType is added to the Txo Validation Protocol and it is started at the same time the Invalid UTXO revalidation protocol is started (i.e. on startup for the Base Node Wallet and when the Base Node Public key is set for the Wallet FFI library).

This PR does not do anything to the transactions associated with the outputs so the Wallets will still show the affected transaction as successful but at least the funds will become available again. Fixing the presentation of the affect Transaction will be the focus of a subsequent PR.

## How Has This Been Tested?
Tests added for the Output Manager database layers

Test added that tests the new validation in the Output Manager Service.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -3455,7 +3457,27 @@ pub unsafe extern "C" fn wallet_add_base_node_peer(
         ptr::swap(error_out, &mut error as *mut c_int);
         return false;
     }
-
+    // Start once off Output manager validation protocols that need the base node peer to be set
+    if let Err(e) = (*wallet).runtime.block_on(
```
stringhandler, line 3461: These `block_on` might cause unresponsive UI. Just something to look out for

stringhandler (dismissed):
Looks good. Untested. I think there may be some performance implications but we can keep an eye out for that

base_layer/wallet_ffi/src/lib.rs
philipr-za: In the FFI calls all calls through to the wallet use block_on because the FFI library calls needs to resolve before returning. That's how error states are passed back to the FFI client. It would be up to the client to manage these calls in an async way.

philipr-za:
> Looks good. Untested. I think there may be some performance implications but we can keep an eye out for that

Indeed with all three validation modes when the wallet has a lot of outputs this can become slow. However I agree that the validation for invalid and spent UTXO's specifically we should probably have some more intelligent ways to decide which outputs have a chance of becoming valid or unspent again otherwise it's not worth the effort. I will make an issue for it

applications/tari_base_node/src/builder.rs
```diff
@@ -532,7 +534,15 @@ where
         .expect("Problem setting local base node public key for output manager service.");
     // Start the Output Manager UTXO Validation
     oms_handle
-        .validate_utxos(UtxoValidationRetry::UntilSuccess)
+        .validate_txos(TxoValidationType::Unspent, TxoValidationRetry::UntilSuccess)
+        .await
+        .expect("Problem starting the Output Manager Service Utxo Valdation process");
```
hansieodendaal:
```suggestion
        .expect("Problem starting the Output Manager Service Unspent Txo Validation process");
```

hansieodendaal: Note typo in `Valdation` vs `Validation`. Also below.

applications/tari_base_node/src/builder.rs
```diff
@@ -532,7 +534,15 @@ where
...
+    oms_handle
+        .validate_txos(TxoValidationType::Invalid, TxoValidationRetry::Limited(5))
+        .await
+        .expect("Problem starting the Output Manager Service Utxo Valdation process");
```
hansieodendaal:
```suggestion
        .expect("Problem starting the Output Manager Service Invalid Txo Validation process");
```

applications/tari_base_node/src/builder.rs
```diff
@@ -532,7 +534,15 @@ where
...
+    oms_handle
+        .validate_txos(TxoValidationType::Unspent, TxoValidationRetry::Limited(5))
         .await
         .expect("Problem starting the Output Manager Service Utxo Valdation process");
```
hansieodendaal:
```suggestion
        .expect("Problem starting the Output Manager Service Spent Txo Validation process");
```

applications/tari_base_node/src/builder.rs
```diff
@@ -532,7 +534,15 @@ where
...
-        .validate_utxos(UtxoValidationRetry::UntilSuccess)
+        .validate_txos(TxoValidationType::Unspent, TxoValidationRetry::UntilSuccess)
+        .await
+        .expect("Problem starting the Output Manager Service Utxo Valdation process");
+    oms_handle
+        .validate_txos(TxoValidationType::Invalid, TxoValidationRetry::Limited(5))
+        .await
+        .expect("Problem starting the Output Manager Service Utxo Valdation process");
+    oms_handle
+        .validate_txos(TxoValidationType::Unspent, TxoValidationRetry::Limited(5))
```
hansieodendaal:
```suggestion
        .validate_txos(TxoValidationType::Spent, TxoValidationRetry::Limited(5))
```

hansieodendaal: `TxoValidationType::Unspent` is called twice.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -322,8 +322,17 @@ impl OutputManagerHandle {
         }
     }
 
-    pub async fn validate_utxos(&mut self, retries: UtxoValidationRetry) -> Result<u64, OutputManagerError> {
-        match self.handle.call(OutputManagerRequest::ValidateUtxos(retries)).await?? {
+    pub async fn validate_txos(
+        &mut self,
+        valdiation_type: TxoValidationType,
```
hansieodendaal:
```suggestion
        validation_type: TxoValidationType,
```

hansieodendaal: This typo can be corrected throughout

base_layer/wallet/src/output_manager_service/protocols/txo_validation_protocol.rs
```diff
@@ -500,28 +511,50 @@ where TBackend: OutputManagerBackend + Clone + 'static
                     "Handled Base Node response (Id: {}) for Invalidated Outputs Query {}", request_key, self.id
                 );
             },
+            TxoValidationType::Spent => {
+                // Go through the response outputs and check if they are currently Spent, if they are then they can be
+                // marked as Unspent because they exist in the UTXO set. Horray!
```
hansieodendaal:
```suggestion
                // marked as Unspent because they exist in the UTXO set. Hurray!
```

hansieodendaal (commented):
Really a nice addition to TXO validation. Just picked up some spelling errors and one small issue that looks like it is a result of a copy-paste operation.

**Note:** Doing a system-level test, after I applied the suggested fix, where I deliberately changed some Unspent TXOs to being Spent TXOs in the desktop wallet. I do not expect any problems :-)

hansieodendaal (dismissed):
LGTM, system-level test also passed

stringhandler (dismissed):

philipr-za:
Conflicts resolved

stringhandler (dismissed):

philipr-za:
Rebase to pull in fixes to Development build

PR #2265, opened 2020-09-28
Minor optimisation for MMR + clippy fixes

sdbondi:
## Description

- Optimisation for `MerkleMountainRange::push` function. (Save one length and one get query per peak iteration)
- `push` functions take an owned version of the hash instead of
  internally cloning a reference to the hash. This means that in a
  number of cases (especially in real/non-test usage) a clone is avoided.
- Removed the `ArrayLike::get_or_panic` function. The reasons why a
  panic is allowed in a given case should be explicit. i.e
  `mmr.get(1)?.expect("we expected an index of 1 to work but it did not")`.
  Moreover, `get_or_panic` does not differentiate between an error
  returned from `get()` (DB IO error?) and the value not being found
  (Invariant violation where we may want to panic)
- Used `&[u8]` instead of `&Hash` as per clippy recommendations. It is
  reasonable that any implementation of a hash should be representable
  as a byte slice.

## Motivation and Context

MMR code improvements

## How Has This Been Tested?

Existing tests pass. Two local base nodes synced blocks. MMR validation passes.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/mmr/benches/bench.rs
+++ b/base_layer/mmr/benches/bench.rs
@@ -22,29 +22,28 @@
 //
 
 use blake2::Blake2b;
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use digest::Digest;
 use std::time::Duration;
 use tari_mmr::MerkleMountainRange;
 
 fn get_hashes(n: usize) -> Vec<Vec<u8>> {
-    let mut result = Vec::with_capacity(n);
-    for i in 0..n {
-        let h = Blake2b::digest(&i.to_le_bytes()).to_vec();
-        result.push(h);
-    }
-    result
+    (0..n).map(|i| Blake2b::digest(&i.to_le_bytes()).to_vec()).collect()
 }
 
 fn build_mmr(c: &mut Criterion) {
     c.bench_function("Build MMR", move |b| {
         let hashes = get_hashes(1000);
         let mut mmr = MerkleMountainRange::<Blake2b, _>::new(Vec::default());
-        b.iter(|| {
-            for i in 0..1000 {
-                let _ = mmr.push(&hashes[i]);
-            }
-        });
+        b.iter_batched(
+            || hashes.clone(),
+            |hashes| {
+                hashes.into_iter().for_each(|hash| {
+                    mmr.push(hash).unwrap();
+                });
+            },
+            BatchSize::SmallInput,
+        );
     });
 }
 
--- a/base_layer/mmr/src/merkle_mountain_range.rs
+++ b/base_layer/mmr/src/merkle_mountain_range.rs
@@ -66,7 +66,7 @@ where
             .clear()
             .map_err(|e| MerkleMountainRangeError::BackendError(e.to_string()))?;
         for hash in hash_iter.into_iter() {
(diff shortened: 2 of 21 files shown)


base_layer/mmr/src/merkle_mountain_range.rs
```diff
@@ -137,35 +137,48 @@ where
...
+                "left_sibling was greater than the reported number of elements contained in the impl of ArrayLike"
+            );
+            let left_hash = self
+                .hashes
+                .get(left_sibling)
+                .map_err(MerkleMountainRangeError::backend_error)?
+                .unwrap();
             peak *= 2;
             pos += 1;
-            let hash_count = self
```
sdbondi, line 163: 👁️  Please check my logic here. 

last_hash (@ hash_count - 1) should always be the previous hash we pushed to the MMR. So instead of getting the length of the hashes from `ArrayLike` and then fetching `len - 1` to get the last hash - just use the last hash we added.

CjS77: I think this is the real optimisation.

~Removing get_or_panic is a slowdown. I'd leave it in, or else `impl Index` so that you can do a direct lookup (which is basically the same thing under the hood for ArrayLikes that aren't backed by Index)~

I suppose in practice we're not backed by something that can do a direct lookup, except in tests, so this is fine.

base_layer/mmr/benches/bench.rs
```diff
@@ -22,29 +22,25 @@
...
+        b.iter_batched(
+            || {
+                counter += 1;
+                Blake2b::digest(&counter.to_le_bytes()).to_vec()
+            },
+            |hash| {
+                mmr.push(hash).unwrap();
+            },
+            BatchSize::SmallInput,
+        );
```
CjS77: You changed the benchmark here from one that measures the aggregate performance of many inserts to one that measures a single insert.

But not all inserts are equal. Some just add the hash, while others have to a bunch of peak calculations. So this means that the benchmark cannot create an apples for apples set of results and will always return a lot of outliers.

A more accurate bench would be use use the same idea as before (a 1000 hash MMR), but still use `iter_batch`

```
fn get_hashes(n: usize) -> Vec<Vec<u8>> {
    let mut result = Vec::with_capacity(n);
    for i in 0..n {
        let h = Blake2b::digest(&i.to_le_bytes()).to_vec();
        result.push(h);
    }
    result
}

fn build_mmr(c: &mut Criterion) {
    c.bench_function("Build MMR", move |b| {
        let hashes = get_hashes(1000);
        let mut mmr = MerkleMountainRange::<Blake2b, _>::new(Vec::default());
        b.iter_batched(
            || hashes.clone(),
            |hashes| {
                hashes.into_iter().for_each(|hash| { mmr.push(hash).unwrap(); });
            },
            BatchSize::SmallInput,
        );
    });
}
```

CjS77 (commented):
Looks good, but I'd prefer the benchmark to stay consistent

base_layer/mmr/benches/bench.rs
sdbondi: Yes that makes sense - code updated

stringhandler (approved):
This should help get clippy to 0

PR #2272, opened 2020-10-01
Peer banning reason

SWvheerden:
## Description

PR adds in a way for keep track in the peer database why a peer was banned. 

## Motivation and Context

Currently the peer db only keeps track of if they are banned and until when. This adds in a a reason why the peer was banned. This will allow retrospective inspection of why peers are getting banned

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/src/peer_manager/peer.rs
+++ b/comms/src/peer_manager/peer.rs
@@ -71,6 +71,7 @@ pub struct Peer {
     /// Flags for the peer.
     pub flags: PeerFlags,
     pub banned_until: Option<NaiveDateTime>,
+    pub banned_reason: String,
     pub offline_at: Option<NaiveDateTime>,
     /// Features supported by the peer
     pub features: PeerFeatures,
@@ -105,6 +106,7 @@ impl Peer {
             flags,
             features,
             banned_until: None,
+            banned_reason: "".to_string(),
             offline_at: None,
             connection_stats: Default::default(),
             added_at: Utc::now().naive_utc(),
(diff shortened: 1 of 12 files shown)


comms/src/peer_manager/peer.rs
```diff
@@ -71,6 +71,7 @@ pub struct Peer {
     /// Flags for the peer.
     pub flags: PeerFlags,
     pub banned_until: Option<NaiveDateTime>,
+    pub banned_reason: String,
```
sdbondi, line 74: This is a breaking change - we can write migrations for this (the infrastructure is there - see [here](https://github.com/tari-project/tari/blob/690b9ee09750c8e8f4d44c1a916a476c082a727f/comms/src/peer_manager/migrations/v1.rs#L68)) so users (esp mobile) are not disrupted, or we need to require that peer DBs are deleted.

sdbondi (commented):
Happy with this change - it is a breaking change for the peer db so I think writing a migration is the path of least resistance (esp for mobile).

delta1:
I'm getting some compile errors in comms when running `cargo test` eg: https://github.com/SWvheerden/tari/blob/sw_add_ban_reason/comms/src/peer_manager/manager.rs#L329

SWvheerden:
> I'm getting some compile errors in comms when running `cargo test` eg: https://github.com/SWvheerden/tari/blob/sw_add_ban_reason/comms/src/peer_manager/manager.rs#L329

haha I know what happend...
I never compiled the test modules

stringhandler (commented):
Happy with the code. Is it outputted anywhere? If you can add it to the `list-peers` command I'm happy to merge it

SWvheerden:
> outputted anywhere? If you can add it to the `list-peers` command I'm happy to merge it

Its added to the peer display

stringhandler (approved):
Sorry, must have missed that

PR #2280, opened 2020-10-01
Refactor basenode/wallet/p2p initialization

sdbondi:
## Description

The main thrust of this PR is to allow services assembled using the StackBuilder
to add protocols to comms before comms is spawned. These protocols will
typically be RPC services.

There existed a _chicken-and-egg_ problem between initializing the
`StackBuilder` services and RPC services which may depend on those services,
which in turn may depend on comms services.

The solution in this PR is to initialize comms itself using the stack
builder (`tari_p2p::P2pInitializer`) allowing objects to be registered
on the `StackBuilder` before comms has spawned. A `SharedCommsContext`
is provided to services that wish to register comms protocols and some
comms components which are available during early intialization (e.g.
PeerManager).

- Created `P2pInitializer` and added it to the base node and wallet
  `StackBuilder`
- Refactored generics out of the `CommsBuilder` and `UnspawnedCommsNode`
- Improved ergonomics of `StackBuilder`
- Added `spawn_when_ready` convinience function to `ServiceInitializer` context
  that allows a service to wait for all handles to become ready before
  starting.
- Comms no longer provides `ShutdownSignals` but has to have a shutdown
  signal provided to it. This allows for a single shutdown signal per
  app.
- Cleanup of `tari_base_node::builder` module
- Removed extra `Shutdown` trigger. A single shutdown trigger is
  used throughout
- Refactored tests and examples as necessary

## Motivation and Context

Base node initialisation was a bit of a mess (could still be improved further, but this PR is massive).  This PR allows a developer to expose backend services as tari_comms RPC services.

## How Has This Been Tested?

Existing tests updated where required. Base node initialisation manually tested a number of times. Base node functionality  operate as expected.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


base_layer/p2p/examples/pingpong.rs
```diff
@@ -176,39 +173,38 @@ mod pingpong {
...
+        let mut handles = StackBuilder::new(shutdown.to_signal())
+            .add_initializer(P2pInitializer::new(comms_config, publisher, vec![]))
+            .add_initializer(LivenessInitializer::new(
+                LivenessConfig {
+                    auto_ping_interval: None, // Some(Duration::from_secs(5)),
+                    ..Default::default()
+                },
+                Arc::clone(&subscription_factory),
+            ))
+            .finish()
```
stringhandler: nit: why not call this method `build()`

base_layer/p2p/src/initialization.rs
```diff
@@ -148,17 +166,17 @@ where
...
         .with_listener_address(node_identity.public_address())
         .with_listener_liveness_max_sessions(1)
         .with_node_identity(node_identity)
         .with_user_agent("/test/1.0")
         .with_peer_storage(peer_database)
         .with_dial_backoff(ConstantBackoff::new(Duration::from_millis(500)))
         .with_min_connectivity(1.0)
-        .build()?;
+        .with_shutdown_signal(shutdown_signal)
+        .finish()?;
```
stringhandler: nit: Why finish and not build?

base_layer/p2p/src/initialization.rs
```diff
@@ -176,54 +194,40 @@ where
...
+        .with_outbound_pipeline(outbound_rx, |sink| {
+            ServiceBuilder::new().layer(dht_outbound_layer).service(sink)
+        })
+        .max_concurrent_inbound_tasks(10)
+        .with_inbound_pipeline(
+            ServiceBuilder::new()
+                .layer(dht.inbound_middleware_layer())
+                .service(SinkService::new(connector)),
+        )
+        .finish();
```
stringhandler: Same here

base_layer/service_framework/src/context/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// Call the given function with the final handles once this future is ready (`notify_ready` is called).
+    pub fn lazy_service<F, S>(&self, service_fn: F) -> LazyService<F, Self, S>
+    where F: FnOnce(FinalServiceHandles) -> S {
+        LazyService::new(self.clone(), service_fn)
+    }
+
+    /// Spawn a task once handles are ready. The resolved handles are passed into this closure.
+    pub fn spawn_when_ready<F, Fut>(self, f: F) -> task::JoinHandle<Fut::Output>
```
stringhandler: Suggestion: maybe `on_started` or `on_ready` is a more clear name?

base_layer/service_framework/src/context/mod.rs
```diff
@@ -50,16 +54,18 @@ macro_rules! acquire_lock {
 }
 
 /// Simple collection for named handles
-#[derive(Default)]
-pub struct ServiceHandles {
-    handles: Mutex<HashMap<TypeId, Box<dyn Any + Sync + Send>>>,
+#[derive(Clone)]
+pub struct FinalServiceHandles {
```
stringhandler: nit: I don't think there is a `NonFinalServiceHandles`, so I think `ServiceHandles` is fine. 

Suggestion: This is a bit of an AbstractFactory even. You could even add a member per Trait it exposes. E.g. 
`ServiceHandleFactory::get_local_node_interface() ->  &LocalNodeInterface`

base_layer/service_framework/src/utillities.rs
```diff
@@ -0,0 +1,48 @@
+//  Copyright 2020, The Tari Project
```
stringhandler: spelling error in name of file

base_layer/service_framework/src/context/future.rs
```diff
@@ -0,0 +1,136 @@
...
+        task::spawn(self.wait_ready().then(f))
+    }
+
+    pub async fn wait_ready(self) -> FinalServiceHandles {
+        let _ = self.ready_signal.await;
+        self.inner
+    }
+
+    /// Spawns the task immediately without waiting for all initializers to complete
+    pub fn spawn_immediately<T>(&self, task: T) -> task::JoinHandle<T::Output>
```
stringhandler: Why would a caller use this instead of just calling `task::spawn()` directly? The method seems unnecessary and could lead to dead code

stringhandler (commented):
Nice. Some nits and suggestions since we are renaming and refactoring anyway. They are provided as a discussion starting point, so let me know what you think. On the whole it's pretty difficult to review the whole thing, but the new interfaces seem neater and easier to understand. I think we must choose either `finish` or `build` across the board for `Buidlers`.

base_layer/service_framework/src/context/future.rs
sdbondi: Do you not think it's important to know that you are spawning a new task?

base_layer/p2p/examples/pingpong.rs
sdbondi: No strong opinions - `finish` is widely used for builders and used consistently in comms. I don't mind changing it at all.

base_layer/service_framework/src/context/mod.rs
sdbondi: Agree - this was a product of figuring out what needs to change and I meant to change the name back.

The service framework is not coupled to core. It should stay that way. I think if we wanted to add a tari_core level abstraction that belongs in another PR.

base_layer/service_framework/src/context/future.rs
sdbondi: The reasons could be:
1. You don't need to have tokio as a dependency or import it (meh - it's fine I guess)
1. If we want to change the way services spawn, the change happens here (probably not going to happen)
1. We use `spawn_when_ready` to spawn tasks in the context, so this is a counterpart which doesn't wait until ready (kind of pedantic)

I think what I'll do is remove this and change the caller to just spawn when ready (even though it doesn't depend on any handles) - there is practically no difference.

stringhandler (approved):

PR #2282, opened 2020-10-01
Fix flaky test in `test_wallet_ffi`

hansieodendaal:
## Description

Handle runtime error by ~retrying~ _removing an unnecessary_ dial-peer _command_, otherwise, _when sometimes_ ~if~ it is not successful, the code panics.

_UPDATE: As per @sdbondi review comments below._

## Motivation and Context

```
---- test::test_wallet_ffi stdout ----
thread 'test::test_wallet_ffi' panicked at 'called `Result::unwrap()` on an `Err` value: DialCancelled', base_layer\wallet_ffi\src\lib.rs:5812:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## How Has This Been Tested?

Running `cargo test --release --all-features --no-fail-fast` in Windows and Ubuntu Linux

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -5831,16 +5831,6 @@ mod test {
             assert_eq!(peer_added, true);
             peer_added = wallet_add_base_node_peer(bob_wallet, public_key_alice.clone(), address_alice_str, error_ptr);
             assert_eq!(peer_added, true);
-            (*alice_wallet)
-                .runtime
-                .block_on(
-                    (*alice_wallet)
-                        .wallet
-                        .comms
-                        .connection_manager()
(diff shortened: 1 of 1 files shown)


philipr-za (dismissed):
Worth a try. Wonder why it does that.

CjS77 (dismissed):
LGTM

sdbondi:
Sometimes the dial is cancelled because both peers dial each other at the same time (this is made worse when using a memorytransport) - my suggestion is to avoid doing this in tests as it can make tests take longer / be flaky. I think we can take this dial out completely.

EDIT: ran the test a few times and it passes with the dial taken out.

hansieodendaal:
Removed the `dial_peer` as suggested, works just fine.

stringhandler (approved):

sdbondi (approved):

PR #2283, opened 2020-10-02
Update `test_transaction_cancellation_sqlite_db` to track down CI flakiness

philipr-za:
## Description
Added a check for the cancellation event which will help nail down where the test fails on CI. The test is not using a real comms stack and the operation its conducting is not time intensive so I cannot understand what is different on CI compared to local.

## How Has This Been Tested?
Is a test update

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/tests/transaction_service/service.rs
+++ b/base_layer/wallet/tests/transaction_service/service.rs
@@ -2760,9 +2760,28 @@ fn test_transaction_cancellation<T: TransactionBackend + Clone + 'static>(backen
 
     runtime.block_on(alice_ts.cancel_transaction(tx_id)).unwrap();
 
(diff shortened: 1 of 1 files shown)


delta1:
Nice I'll check this out now - I have had this error a number of times locally even, so it's pretty random and not just related to CI

hansieodendaal:
> 
> 
> Nice I'll check this out now - I have had this error a number of times locally even, so it's pretty random and not just related to CI

ditto

philipr-za:
> > Nice I'll check this out now - I have had this error a number of times locally even, so it's pretty random and not just related to CI
> 
> ditto

I will try run the whole test suite then, running the test on its own 100 times in a row showed no failures.

hansieodendaal (approved):
LGTM

PR #2284, opened 2020-10-02
Fix target difficulty for block 1 on rincewind

stringhandler:
## Description

Previously there was a bug where only the first block would apply the `min_target_difficulty`. This means that in the current rincewind testnet chain, block at height 1 has a target difficulty of 60k, and it then drops after that. A new chain cannot currently sync until this fix is applied

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


SWvheerden (approved):
LGTM

PR #2290, opened 2020-10-02
Improve Closest broadcast strategy and use it for saf txn broadcast

sdbondi:
## Description

Closest broadcast strategy changed as follow:
1. A predefined number of peers is used instead of an arbitrary `n`
   (broadcast_factor)
1. A minimum of 2 peers are dialled that are closer to the message
   destination. This is in addition to closer peers with existing
   connections.

- Transaction protocols use the new closest broadcast strategy
- The orders of the store and forward layers are switched as forward
depends on a message flag from store to determine how to propagate the
message (enabling a 'bloom' of messages around a region)

## Motivation and Context

This increases the likelihood that directed broadcast messages are propagated to the correct peers. The new connections that could possibly be established should reduce the number of hops to the destination. 

## How Has This Been Tested?

Existing tests pass - sent transactions via network propagation / SAF to mobile.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/service/service.rs
+++ b/base_layer/core/src/base_node/service/service.rs
@@ -574,7 +574,7 @@ async fn handle_outbound_block(
     outbound_message_service
         .flood(
             NodeDestination::Unknown,
-            OutboundEncryption::None,
+            OutboundEncryption::ClearText,
             exclude_peers,
             OutboundDomainMessage::new(
                 TariMessageType::NewBlock,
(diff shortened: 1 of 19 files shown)


philipr-za (approved):
LGTM, makes the Outbound Requester methods clearer.

stringhandler (approved):
Looks good. Untested

PR #2291, opened 2020-10-03
Console Wallet Automation for `send-tari` and `make-it-rain`

delta1:
## Description

Add command parsing and executing logic to enable automation of the console
wallet. By passing `--input path/to/file`, the lines of a text file can
be parsed and executed as wallet commands. A single command can be
executed by using `--command <command to execute>`.

Currently only `get-balance`, `send-tari`, and `make-it-rain` are
implemented, with more planned. In script mode, all commands are parsed before being
executed. If any commands fail to parse, the wallet exits immediately with
an appropriate error message.

New configuration options `command_send_wait_timeout`  and
`command_send_wait_stage` are added. These allow the command/script mode
to wait until the sent transactions are detected at the configured "wait
stage" within a given timeout value. Defaults to "Broadcast" and 300
second timeout.

- `DirectSendOrSaf` - The transaction was initiated and was accepted via
Direct Send or Store And Forward.
- `Negotiated` - The recipient replied and the transaction was negotiated.
- `Broadcast` - The transaction was broadcast to the base node mempool.
- `Mined` - The transaction was successfully detected as mined on the
blockchain. 

Example invocations: 

`tari_console_wallet --command get-balance`
`tari_console_wallet --command send-tari 420T <public key or emoji id> optional message`

`tari_console_wallet --input /path/to/file`

where "file" consists of: 

```
# a line beginning with '#' is skipped, as are blank lines 

get-balance
send-tari 420T <public key or emoji id> optional message

# make it rain arguments: [tx/s] [duration] [start amount] [increment amount] [datetime to start] [address] [message]
make-it-rain 1 120 420T 69T now <public key or emoji id> message
``` 

## Motivation and Context

Lays the groundwork for further automation of the console wallet. 

## How Has This Been Tested?

- cargo test
- new tests (could use more integration testing)
- manually 

## Types of changes

* [x] New Tests
* [x] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/automation/command_parser.rs
+++ b/applications/tari_console_wallet/src/automation/command_parser.rs
@@ -1,0 +1,243 @@
+// Copyright 2020. The Tari Project
+//
+// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+// following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+// disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
+// following disclaimer in the documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
+// products derived from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+// INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+// USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use crate::automation::{commands::WalletCommand, error::ParseError};
+
+use chrono::{DateTime, Utc};
+use chrono_english::{parse_date_string, Dialect};
+use core::str::SplitWhitespace;
+use std::{
+    fmt::{Display, Formatter},
+    str::FromStr,
+};
+use tari_app_utilities::utilities::parse_emoji_id_or_public_key;
+
+use tari_core::transactions::{tari_amount::MicroTari, types::PublicKey};
+
+#[derive(Debug)]
+pub struct ParsedCommand {
+    pub command: WalletCommand,
+    pub args: Vec<ParsedArgument>,
+}
+
+impl Display for ParsedCommand {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let command = match self.command {
+            WalletCommand::GetBalance => "get-balance",
+            WalletCommand::SendTari => "send-tari",
+            WalletCommand::MakeItRain => "make-it-rain",
+        };
+
+        let args = self
+            .args
+            .iter()
+            .map(|a| a.to_string())
+            .collect::<Vec<String>>()
+            .join(" ");
+
+        write!(f, "{} {}", command, args)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ParsedArgument {
+    Amount(MicroTari),
+    PublicKey(PublicKey),
+    Text(String),
+    Float(f64),
+    Int(u64),
+    Date(DateTime<Utc>),
+}
+
+impl Display for ParsedArgument {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ParsedArgument::Amount(v) => write!(f, "{}", v.to_string()),
+            ParsedArgument::PublicKey(v) => write!(f, "{}", v.to_string()),
+            ParsedArgument::Text(v) => write!(f, "{}", v.to_string()),
+            ParsedArgument::Float(v) => write!(f, "{}", v.to_string()),
+            ParsedArgument::Int(v) => write!(f, "{}", v.to_string()),
+            ParsedArgument::Date(v) => write!(f, "{}", v.to_string()),
+        }
+    }
+}
+
+pub fn parse_command(command: &str) -> Result<ParsedCommand, ParseError> {
+    let mut args = command.split_whitespace();
+    let command_str = args.next().ok_or_else(|| ParseError::Empty("command".to_string()))?;
+
+    let command =
+        WalletCommand::from_str(command_str).map_err(|_| ParseError::WalletCommand(command_str.to_string()))?;
+
+    use WalletCommand::*;
+    let args = match command {
+        GetBalance => Vec::new(),
+        SendTari => parse_send_tari(args)?,
+        MakeItRain => parse_make_it_rain(args)?,
+    };
+
+    Ok(ParsedCommand { command, args })
+}
+
+fn parse_make_it_rain(mut args: SplitWhitespace) -> Result<Vec<ParsedArgument>, ParseError> {
+    let mut parsed_args = Vec::new();
+
+    // txs per second
+    let txps = args.next().ok_or_else(|| ParseError::Empty("Txs/s".to_string()))?;
+    let txps = txps.parse::<f64>().map_err(ParseError::Float)?;
+    if txps > 25.0 {
+        println!("Maximum transaction rate is 25/sec");
+        return Err(ParseError::Invalid);
+    }
+    parsed_args.push(ParsedArgument::Float(txps));
+
+    // duration
+    let duration = args.next().ok_or_else(|| ParseError::Empty("duration".to_string()))?;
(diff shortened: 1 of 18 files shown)


applications/tari_console_wallet/src/main.rs
```diff
@@ -111,30 +124,188 @@ fn main_inner() -> Result<(), ExitCodes> {
...
+        WalletMode::Tui => tui_mode(runtime, grpc, node_config, node_identity, wallet),
+        WalletMode::Grpc => grpc_mode(runtime, grpc, node_config),
+        WalletMode::Script(path) => script_mode(runtime, path, wallet),
+        WalletMode::Invalid => Err(ExitCodes::InputError(
+            "Can't run GRPC and file input mode simultaneously!".to_string(),
+        )),
+    }
+}
+
+fn script_mode(mut runtime: Runtime, path: PathBuf, wallet: WalletThreadSafe) -> Result<(), ExitCodes> {
```
stringhandler: The script might also be a once off string

applications/tari_console_wallet/src/main.rs
```diff
@@ -111,30 +124,188 @@ fn main_inner() -> Result<(), ExitCodes> {
...
+        println!("{:?}", parsed);
+
+        use ParsedArgument::*;
+        match parsed.command {
+            BaseNodeCommand::GetBalance => match output_service.get_balance().await {
+                Ok(balance) => println!("Balance: {:?}", balance),
+                Err(e) => println!("output manager error! {:?}", e),
+            },
+            // todo: refactor to function
+            BaseNodeCommand::SendTari => {
```
stringhandler: There should probably be two enums: BaseNodeCommand, WalletCommand

applications/tari_console_wallet/src/main.rs
delta1: sounds good, instead of an `input_file` give it an `input_command`

applications/tari_console_wallet/src/main.rs
delta1: Do you suggest a completely separate enum or some way of inheriting or reusing?

stringhandler: I think completely separate. The base node will eventually not accept WalletCommand commands

applications/tari_app_utilities/src/commands.rs
```diff
@@ -0,0 +1,373 @@
...
+};
+use thiserror::Error;
+use tokio::{
+    runtime::Handle,
+    stream::StreamExt,
+    task::JoinError,
+    time::{delay_for, timeout},
+};
+
+/// Enum representing commands used by the wallet
```
philipr-za: So these are the commands supported by the Script interface of the `tari_console_wallet` but there are a number of extra wallet specific commands supported by the current parser in the Base Node. Shouldn't this then rather be in the `tari_console_wallet` rather than in this more generic app utilities crate as its specific to that app?

philipr-za: The more I see of this file I feel like it should all be in the `tari_console_wallet` under its own module. This is app specific logic not common between the apps.

applications/tari_app_utilities/src/commands.rs
```diff
@@ -0,0 +1,373 @@
...
+
+pub async fn monitor_transactions(
+    transaction_service: TransactionServiceHandle,
+    tx_ids: Vec<u64>,
+    wait_stage: TransactionStage,
+) -> Vec<SentTransaction>
+{
+    let mut event_stream = transaction_service.get_event_stream_fused();
+    let mut results = Vec::new();
+    debug!("monitor transactions wait_stage: {:?}", wait_stage);
```
philipr-za: applies to all the log statements but you should have a target.

applications/tari_app_utilities/src/commands.rs
```diff
@@ -0,0 +1,373 @@
...
+    println!(
+        "Monitoring {} sent transactions to {:?} stage...",
+        tx_ids.len(),
+        wait_stage
+    );
+
+    loop {
```
philipr-za: Probably need to work in some sort of timeout and/or way to show the the process is alive and ongoing.

applications/tari_console_wallet/src/main.rs
```diff
@@ -104,36 +102,36 @@ fn main_inner() -> Result<(), ExitCodes> {
         .build()
         .unwrap();
 
-    let wallet = runtime.block_on(setup_wallet(&node_config, wallet_identity))?;
+    let wallet = runtime.block_on(setup_wallet(&config, wallet_identity))?;
     debug!(target: LOG_TARGET, "Starting app");
 
     let node_identity = wallet.comms.node_identity().as_ref().clone();
     let wallet = Arc::new(RwLock::new(wallet));
     let grpc = crate::grpc::WalletGrpcServer::new(wallet.clone());
```
philipr-za, line 111: We can just copy the wallet object directly, we do not need the arc or the rwlock I believe.

base_layer/wallet/src/lib.rs
```diff
@@ -41,9 +44,12 @@ use crate::{
...
 };
+
 pub type WalletSqlite = Wallet<
     WalletSqliteDatabase,
     TransactionServiceSqliteDatabase,
     OutputManagerSqliteDatabase,
     ContactsServiceSqliteDatabase,
 >;
+
+pub type WalletThreadSafe = Arc<RwLock<WalletSqlite>>;
```
philipr-za: I don't think we need this. The WalletSqlite struct is a handle that is meant to be cloneable and its interfaces are async and threadsafe. I believe we can just copy the wallet handle directly for each component of this app (grpc server gets a clone of the wallet object and the other component, TUI, script etc, gets its own copy).

Taking this out means we remove the chance for weird deadlocks and also don't need the extra boiler plate of asking for a lock before interacting with the wallet.

applications/tari_app_utilities/src/commands.rs
```diff
@@ -0,0 +1,373 @@
...
+#[derive(Clone, PartialEq, Debug, Display, EnumIter, EnumString)]
+#[strum(serialize_all = "kebab_case")]
+pub enum WalletCommand {
+    GetBalance,
+    SendTari,
+    MakeItRain,
+}
+
+#[derive(Debug, Error)]
+pub enum CommandError {
```
philipr-za: Nit: but we tend to put these in their own error.rs file so they are easy to find. This will be easier when moving the logic in this file into a module in the `tari_console_wallet` app I think.

philipr-za (commented):
Looking good so far, I really like the way you genericized the parsing of the arguments.

applications/tari_app_utilities/src/commands.rs
delta1: Sure that makes sense, I wanted to refactor the base node to utilise the same parsing and executing logic, but my rust-fu isn't quite there yet :) Happy to move it into a console wallet module

applications/tari_app_utilities/src/commands.rs
delta1: Cool - will do

applications/tari_app_utilities/src/commands.rs
delta1: There is a timeout one level up where monitor_transactions is called, and the timeout and wait stage are configurable

applications/tari_console_wallet/src/main.rs
delta1: Cool, I was just keeping what was already there. I'll try that out

base_layer/wallet/src/lib.rs
delta1: As above, I'll try this out

applications/tari_app_utilities/src/commands.rs
delta1: derp I missed this one, thanks I'll double-check that

applications/tari_app_utilities/src/commands.rs
delta1: https://github.com/tari-project/tari/pull/2291/files#diff-0aab53ebe00309fc20e74b8823646775R358

applications/tari_app_utilities/src/commands.rs
philipr-za: I believe in the end the wallet functionality will be completely removed from the Base Node and only live in the wallet app which is primarily why I believe this belongs in that crate. If this parsing ends up being used in a second app then I think it can move back here though if possible I would want to keep the dependency on the Transaction Service Handles out of here as that implies this is more than just parsing.

delta1: That makes sense, thanks @philipr-za

delta1:
Still working on comments provided by @philipr-za 🙏

base_layer/wallet/src/lib.rs
delta1: Discussed offline, this is going to require a bit more of a refactoring due to the shutdown signals from various clients

applications/tari_console_wallet/src/automation/command_parser.rs
```diff
@@ -0,0 +1,208 @@
...
+    let parsed = parse_command(command_str);
+    assert!(parsed.is_err());
+
+    let command_str = "send-tari 999T 6a493210f7499cd17fecb510ae0cea23a110e8d5b901f8acadd3095c73a3b919";
+    let parsed = parse_command(command_str);
+    assert!(parsed.is_ok());
+
+    let command_str = "send-tari 999T 6a493210f7499cd17fecb510ae0cea23a110e8d5b901f8acadd3095c73a3b919 message";
+    let parsed = parse_command(command_str);
+    assert!(parsed.is_ok());
```
philipr-za: Nit: actually test that the parsed values are correct and also test uT and no unit on the tari amount.

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -0,0 +1,374 @@
...
+#[derive(Debug)]
+pub struct SentTransaction {
+    id: u64,
+    stage: TransactionStage,
+}
+
+pub async fn send_tari(
+    mut wallet_transaction_service: TransactionServiceHandle,
+    args: Vec<ParsedArgument>,
+) -> Result<u64, CommandError>
```
philipr-za: Very small nit: I aliased this as TxId in the Wallet crate to make it clear what the number means. I actually wonder if in the future we shouldn't make it a hard type.

philipr-za (dismissed):
Looks great, couple nits

applications/tari_console_wallet/src/automation/command_parser.rs
delta1: Agreed, and I'd like to add the other commands too

applications/tari_console_wallet/src/automation/commands.rs
delta1: That's a good idea, will do

philipr-za (approved):
LGTM

PR #2292, opened 2020-10-05
Improve TXO validation, broadcast protocols and transaction protocols restart logic at base node startup.

hansieodendaal:
## Description

At base node startup, TXO validation, broadcast protocols and transaction protocols restart only restart after initial blockchain sync has been achieved. However, broadcast and transaction protocols should only be restarted after TXO validation has been concluded, as UTXOs used in those unconcluded transactions may be invalid. This effect is pronounced if there are many unconcluded transactions at base node startup.

## Motivation and Context

Preventing broadcast of invalid transactions that would live in the base node network for a really long time and that would waste resources.

## How Has This Been Tested?

Tested with base node startup in Windows.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -331,7 +331,6 @@ where
         .await
         .expect("Problem setting local base node public key for transaction service.");
     let oms_handle = wallet_handles.expect_handle::<OutputManagerHandle>();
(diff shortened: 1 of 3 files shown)


stringhandler (approved):
Not going to be too fussy on this code because the wallet will eventually be removed from the base node and this will no longer be necessary

PR #2293, opened 2020-10-05
Prevent tari_app_grpc rebuild on every build

sdbondi:
## Description

`tari_app_grpc` rebuilds each time because the tonic output is written to
the `src` directory and the generated file is written on each build.
This causes rust to recompile.

This PR builds to `OUT_DIR` which avoids the rebuild.

This comes at a cost of developer experience  however, support 
is already offered by some IDEs
https://blog.jetbrains.com/clion/2019/12/recent-updates-in-rust/


delta1 (approved):
LGTM. I also stumbled across [sccache](https://github.com/mozilla/sccache) which looks like it could help with build caching

sdbondi:
@delta1 Interesting crate - though not sure if the use case applies. Rust's existing incremental build cache works great, but it being broken by writes to src from build.rs.

delta1:
yep sorry, nothing to do with this PR just wanted to mention it

stringhandler (approved):

PR #2298, opened 2020-10-05
Improve block sync speed by scaling block window

stringhandler:
## Description

Increase or decrease the number of blocks requested during block sync according to timeouts. If the request times out, halve the number of blocks requested. If the request returns in less than half of the timeout time, double the blocks requested. 
Nodes were also banned if they timed out. This has been changed to only ban if they time out when the block window requested is 1.
Note: I reduced the `BASE_NODE_SERVICE_REQUEST_TIMEOUT` to 30 seconds. This may impact other calls.

Special thanks to @hansieodendaal for his PR into this PR. His changes got combined with mine in a squash merge.

## Motivation and Context

The previous code would only request 5 blocks at a time (window). This number can be increased, but there are some blocks that take a long time to process, so a window of  128 would be a problem because a single block may timeout. Changing this to a dynamic window solves the problem.  

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_utilities/src/utilities.rs
+++ b/applications/tari_app_utilities/src/utilities.rs
@@ -189,23 +189,10 @@ pub fn convert_socks_authentication(auth: SocksAuthentication) -> socks::Authent
 ///
 /// ## Returns
 /// A result containing the runtime on success, string indicating the error on failure
-pub fn setup_runtime(config: &GlobalConfig) -> Result<Runtime, String> {
-    let num_core_threads = config.core_threads;
-    let num_blocking_threads = config.blocking_threads;
-    let num_mining_threads = config.num_mining_threads;
-
-    info!(
-        target: LOG_TARGET,
-        "Configuring the node to run on {} core threads, {} blocking worker threads and {} mining threads.",
-        num_core_threads,
-        num_blocking_threads,
-        num_mining_threads
-    );
+pub fn setup_runtime() -> Result<Runtime, String> {
     tokio::runtime::Builder::new()
         .threaded_scheduler()
         .enable_all()
-        .max_threads(num_core_threads + num_blocking_threads + num_mining_threads)
-        .core_threads(num_core_threads)
         .build()
         .map_err(|e| format!("There was an error while building the node runtime. {}", e.to_string()))
 }
--- a/applications/tari_base_node/src/main.rs
+++ b/applications/tari_base_node/src/main.rs
@@ -154,7 +154,7 @@ fn main_inner() -> Result<(), ExitCodes> {
     debug!(target: LOG_TARGET, "Using configuration: {:?}", node_config);
 
     // Set up the Tokio runtime
-    let mut rt = setup_runtime(&node_config).map_err(|err| {
+    let mut rt = setup_runtime().map_err(|err| {
         error!(target: LOG_TARGET, "{}", err);
         ExitCodes::UnknownError
     })?;
--- a/base_layer/core/src/base_node/consts.rs
+++ b/base_layer/core/src/base_node/consts.rs
@@ -23,6 +23,12 @@
 use std::time::Duration;
 
 /// The allocated waiting time for a request waiting for service responses from remote base nodes.
-pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
+pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+/// The minimum allocated waiting time for a request waiting for service responses from remote base nodes.
+pub const BASE_NODE_SERVICE_REQUEST_MIN_TIMEOUT: Duration = Duration::from_secs(10);
+/// The allocated waiting time for a fetch blocks request waiting for service responses from remote base nodes.
+pub const FETCH_BLOCKS_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+/// The minimum allocated waiting time for a fetch blocks request waiting for service responses from remote base nodes.
+pub const FETCH_BLOCKS_SERVICE_REQUEST_MIN_TIMEOUT: Duration = Duration::from_secs(10);
 /// The fraction of responses that need to be received for a corresponding service request to be finalize.
 pub const BASE_NODE_SERVICE_DESIRED_RESPONSE_FRACTION: f32 = 0.6;
--- a/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
@@ -22,6 +22,7 @@
 use crate::{
     base_node::{
         comms_interface::{BlockEvent, CommsInterfaceError},
+        consts::FETCH_BLOCKS_SERVICE_REQUEST_TIMEOUT,
         state_machine_service::{
             states::{
                 helpers::{ban_all_sync_peers, ban_sync_peer, request_headers, select_sync_peer},
@@ -44,6 +45,7 @@ use std::{
     fmt::{Display, Formatter},
     str::FromStr,
     sync::Arc,
+    time::{Duration, Instant},
 };
 use tari_comms::{connectivity::ConnectivityError, peer_manager::PeerManagerError};
 use tari_crypto::tari_utilities::{hex::Hex, Hashable};
@@ -54,14 +56,14 @@ const LOG_TARGET: &str = "c::bn::state_machine_service::states::block_sync";
 // The maximum number of retry attempts a node can perform to request a particular block from remote nodes.
 const MAX_METADATA_REQUEST_RETRY_ATTEMPTS: usize = 3;
 const MAX_HEADER_REQUEST_RETRY_ATTEMPTS: usize = 5;
-const MAX_BLOCK_REQUEST_RETRY_ATTEMPTS: usize = 5;
+const MAX_BLOCK_REQUEST_RETRY_ATTEMPTS: usize = 20;
 // The maximum number of retry attempts for attempting to validly request and add the block at a specific block height
 // to the chain.
 const MAX_ADD_BLOCK_RETRY_ATTEMPTS: usize = 3;
 // The number of headers that can be requested in a single query.
-const HEADER_REQUEST_SIZE: usize = 100;
+const HEADER_REQUEST_SIZE: usize = 1024;
 // The number of blocks that can be requested in a single query.
-const BLOCK_REQUEST_SIZE: usize = 5;
+const BLOCK_REQUEST_SIZE: usize = 256;
 
 /// Configuration for the Block Synchronization.
 #[derive(Clone, Copy)]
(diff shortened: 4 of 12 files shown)


delta1:
clippy warnings

delta1:
otherwise looks good I'm testing locally

base_layer/core/src/base_node/consts.rs
```diff
@@ -23,6 +23,6 @@
 use std::time::Duration;
 
 /// The allocated waiting time for a request waiting for service responses from remote base nodes.
-pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
+pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
```
hansieodendaal, line 26: I do not think this value should be changed at all, as too many things depend on this. Rather introduce a constant specifically for block sync that can be changed via the config file.

hansieodendaal (commented):
Just concerned about changing the global `BASE_NODE_SERVICE_REQUEST_TIMEOUT` constant. It seems this implementation warrants an exclusive configurable timeout for syncing.

Will run some system-level tests where I use an exclusive local timeout constant for this, keeping the original intact.

base_layer/core/src/base_node/consts.rs
hansieodendaal, line 26:
```suggestion
pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
```

hansieodendaal (commented):
~Proposed changes to make the sync service request timeout unique and not change the global `BASE_NODE_SERVICE_REQUEST_TIMEOUT`~

Please see proposed code changes in https://github.com/mikethetike/tari/pull/2

sdbondi (commented):
This has a big flaw:
If blocks are full, then requesting > 8 will result in the sending node assembling one huge message that exceeds the messaging protocols 8mb limit (EDIT: the message is discarded). So unless the size of the group of blocks is taken into account when packaging messages, you could never safely request more than 7/8 blocks at a time. The solution we have that solves this and deals with timeouts is to stream via a base node RPC interface.

I'm not a fan of more config for things that should just work.

stringhandler:
I don't know if I understand. If you request more than >8 and the blocks are full, it will timeout and request 4 the next time. This speeds up block syncing significantly when there are a lot of empty blocks, and slows down for larger blocks. It's important for the timeout to be small so that the request is resent

philipr-za:
I would also like to add a voice to do this operation by streaming. Downloading a large piece of data is usually done by streaming chunks of that file so feels like we should stream on a per block basis. Streaming means that you would be served the blocks as fast as the BN can serve blocks and a single error won’t mean losing a whole window’s worth of blocks just the block in progress. 

Making the window larger is an attempt to reduce this request/response + timeout overhead per block received but at the cost that a larger window is risky because many blocks worth of processing (on the server BN) and waiting time (on the client BN) could be lost. Streaming reduces this risk down to single blocks.

Edit: also the receving BN can adapt their sync strategy on a much finer grain basis i.e. connect to more BNs because this one is streaming slow.

sdbondi (dismissed):
@mikethetike Sure, this will speed things up - it's just not robust, efficient or a particularly good strategy to be frank.

EDIT: Apologies if this came off as aggressive, was certainly not my intention. I wanted to point out that we have better ways to get this right :) As others have pointed out (as did I) this is faster than what we have - so I have no problem with this going in and I understand why you did it, however I'd prefer to have a solution that does not throw large messages away. I also believe that the development effort is less to do it correctly once than keeping a more complex solution around which constantly needs to be patched when issues crop up.

stringhandler:
Lol, ok. Not getting much love for this. I think I'll get some benchmarks and then we can compare it with a streamed approach and then I'll close if need be

philipr-za:
The work is done and this is definitely an improvement over what we have at the moment so I don't see why it shouldn't go in as it is. 

However, with our evolving understanding of networking and with the insights we have gained by having to do block sync using this windowed approach (due to the limitations of our initial comms layer design) it is fairly clear that Streaming is a better approach for this process. Perhaps in the review for this PR is not the 100% ideal place to discuss.

delta1:
In some quick testing this looks significantly faster than what we currently have. I agree that it's done **and** an improvement and should be merged - with the goal to get refactored to streaming as soon as its viable.

hansieodendaal:
> The work is done and this is definitely an improvement over what we have at the moment so I don't see why it shouldn't go in as it is.
> 
> However, with our evolving understanding of networking and with the insights we have gained by having to do block sync using this windowed approach (due to the limitations of our initial comms layer design) it is fairly clear that Streaming is a better approach for this process. Perhaps in the review for this PR is not the 100% ideal place to discuss.

I agree that this should be merged as it is definitely an improvement (_ran some sync tests as well_), although not ideal. In the blockchain sync from scratch tests, many responses took >10 and >20 minutes, and that was due to the large timeout of 10 minutes. This PR will also improve that as the timeout is short.

_**Side note:** Seeing that messages larger than the messaging protocols 8mb limit are discarded, a possible improvement (for another PR) for a peer base node could be to only respond with the number of blocks that fits into the messaging protocols 8mb limit._

stringhandler:
So did some benchmarks, but it didn't sync all the way due to some weirdness. The average was about 200-250 blocks per minute for the first 50000 blocks. 

Another improvement in this PR is that peers only get banned if they timeout once they cannot respond with even one block. Previously it would ban on any time out. I think this will help prevent unnecessary banning.

delta1:
Clippy!

applications/tari_app_utilities/src/utilities.rs
```diff
@@ -189,23 +189,10 @@ pub fn convert_socks_authentication(auth: SocksAuthentication) -> socks::Authent
...
-        num_core_threads,
-        num_blocking_threads,
-        num_mining_threads
-    );
+pub fn setup_runtime() -> Result<Runtime, String> {
     tokio::runtime::Builder::new()
         .threaded_scheduler()
         .enable_all()
-        .max_threads(num_core_threads + num_blocking_threads + num_mining_threads)
-        .core_threads(num_core_threads)
```
hansieodendaal, line 208: ~~These changes should not be part of this PR~~

**Edit:** I am just concerned that this may break something, but I have not done the system tests or know the impact, so if this is a safe change to do, by all means.

applications/tari_base_node/src/main.rs
```diff
@@ -154,7 +154,7 @@ fn main_inner() -> Result<(), ExitCodes> {
     debug!(target: LOG_TARGET, "Using configuration: {:?}", node_config);
 
     // Set up the Tokio runtime
-    let mut rt = setup_runtime(&node_config).map_err(|err| {
+    let mut rt = setup_runtime().map_err(|err| {
```
hansieodendaal, line 157: ~~This change should not be part of this PR~~

**Edit:** See above related comment in this regard.

base_layer/core/src/base_node/consts.rs
```diff
@@ -23,6 +23,12 @@
...
```
hansieodendaal, line 26:
```suggestion
pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
```

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -327,21 +322,67 @@ async fn synchronize_blocks<B: BlockchainBackend + 'static>(
...
+                    // if we took less than half the timeout, try doubling the
+                    // blocks we request
+                    if timer.elapsed() < FETCH_BLOCKS_SERVICE_REQUEST_TIMEOUT / 2 {
+                        block_window = blocks_synced * 2;
+                    } else {
+                        // Otherwise, request the same number of blocks. `request_and_add_blocks` automatically
+                        // scales down
+                        block_window = blocks_synced;
+                    }
+                }
```
hansieodendaal, line 358:
```suggestion
```

hansieodendaal: There seems to be a merge error here

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -327,21 +322,67 @@ async fn synchronize_blocks<B: BlockchainBackend + 'static>(
...
+                    // if we took less than half the timeout, try doubling the
+                    // blocks we request
+                    if timer.elapsed() < shared.config.block_sync_config.fetch_blocks_timeout / 2 {
+                        block_window = blocks_synced * 2;
+                    } else {
+                        // Otherwise, request the same number of blocks. `request_and_add_blocks` automatically
+                        // scales down
+                        block_window = blocks_synced;
+                    }
+                }
```
hansieodendaal, line 384:
```suggestion
```

hansieodendaal: There seems to be a merge error here

common/src/configuration/global.rs
```diff
@@ -58,8 +58,6 @@ pub struct GlobalConfig {
     pub orphan_storage_capacity: usize,
     pub pruning_horizon: u64,
     pub pruned_mode_cleanup_interval: u64,
-    pub core_threads: usize,
-    pub blocking_threads: usize,
```
hansieodendaal, line 62: ~~Not for this PR~~

**Edit:** See above related comment in this regard.

common/src/configuration/global.rs
```diff
@@ -205,17 +205,6 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
...
-    let key = config_string("base_node", &net_str, "core_threads");
-    let core_threads = cfg
-        .get_int(&key)
-        .map_err(|e| ConfigurationError::new(&key, &e.to_string()))? as usize;
-
-    let key = config_string("base_node", &net_str, "blocking_threads");
-    let blocking_threads = cfg
-        .get_int(&key)
-        .map_err(|e| ConfigurationError::new(&key, &e.to_string()))? as usize;
-
```
hansieodendaal, line 218: ~~Not for this PR~~

**Edit:** See above related comment in this regard.

common/src/configuration/global.rs
```diff
@@ -422,8 +423,6 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         orphan_storage_capacity,
         pruning_horizon,
         pruned_mode_cleanup_interval,
-        core_threads,
-        blocking_threads,
```
hansieodendaal, line 426: ~~Not for this PR~~

**Edit:** See above related comment in this regard.

hansieodendaal (commented):
I like this, just two categories of comments:
- **Edited** Some code changes ~~should~~ _may_ not be _fit_ for this PR
- **Edited** We _may_ need to keep the `BASE_NODE_SERVICE_REQUEST_TIMEOUT` unchanged for now, _as I am not sure if all base node service requests can live with this short timeout? Maybe fine-tune the service request timeouts on an individual basis later on._

base_layer/core/src/base_node/consts.rs
hansieodendaal: I am not sure if all base node service requests can live with this short timeout? That is why I suggest keeping this one as it was and maybe fine-tune the service request timeouts on an individual basis later on.

PR #2316, opened 2020-10-08
Add meta_data field to peer

SWvheerden:
## Description

This PR adds in meta_data field to the peer that can be used by upstream clients. 

## Motivation and Context

I used a Hashmap<u8,Vec<u8>> to ensure future proofing of the field. This will also allow more than one struct of information to be serialised into the peer. By using a serialised piece of info the peer does not need to be aware of the application layer etc.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/src/peer_manager/migrations/v3.rs
+++ b/comms/src/peer_manager/migrations/v3.rs
@@ -48,7 +48,7 @@ use tari_storage::{
 const LOG_TARGET: &str = "comms::peer_manager::migrations::v2";
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
-pub struct PeerV2 {
+pub struct PeerV3 {
     id: Option<PeerId>,
     public_key: CommsPublicKey,
     #[serde(serialize_with = "serialize_to_hex")]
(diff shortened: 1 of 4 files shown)


sdbondi (approved):
Looks good!

comms/src/peer_manager/migrations/v3.rs
```diff
@@ -0,0 +1,115 @@
...
+use log::*;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use tari_crypto::tari_utilities::hex::serialize_to_hex;
+use tari_storage::{
+    lmdb_store::{LMDBDatabase, LMDBError},
+    IterationResult,
+};
+
+const LOG_TARGET: &str = "comms::peer_manager::migrations::v2";
```
delta1, line 48: I know this is merged now, but just wanted to check if this should be `::v3` instead?

SWvheerden: Yes This should be, thanks for pointing out. Fixing in my current code

delta1: Nice one 👍

SWvheerden: #2318

Issue #2247, opened 2020-09-18
`find_orphan_chain_tips` in BlockchainDatabase is very inefficient


sdbondi:
`find_orphan_chain_tips` requires searching though all orphans recursively. This causes large reorgs (+1000 block) to take hours.

Additionally, it is likely possible to slow a node to a crawl by pushing many large orphan blocks to it. These dummy blocks do not need to be mined because the difficulty is not checked before adding into the orphan block pool.

SWvheerden:
I think the biggest culprit inside there is the iterator over the orphans. Iterating over 1000 items should not take seconds.

sdbondi:
@SWvheerden Each iteration has to deserialise an entire block, which involves [decompression for every public key](https://github.com/tari-project/tari-crypto/blob/7ee2dfb24595e8a858ddd3f2f8de15793fa466a2/src/ristretto/ristretto_keys.rs#L340) and [other operations](https://github.com/dalek-cryptography/curve25519-dalek/blob/master/src/scalar.rs#L210) per input/output/kernel.

I suspect there is no perf issue with the iterator implementation itself. I think a better strategy is needed to remove the need to load and deserialise in the first place. Maybe tip tracking and/or only keeping the headers in the orphan db and only downloading the full blocks if needed for a reorg.

EDIT:
I have seen ~900 blocks take as long as 22 seconds on my macbook after a stress test. I suspect that is because there are many more expensive "crypto operations" for each full block. The problem isn't the "crypto operations", it's that we have to iterate over the whole orphan db when adding a block in the first place

hansieodendaal:
Some stats:

The log entries below shows measured time for each block to be added during a fairly large block sync. This can happen if a base node is started up when it last synced a couple of days ago or if a large chain reorg is happening.
``` rust
2020-10-09 07:15:19.044207800 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:23.087208500 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:27.297718300 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:32.445717800 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:36.493718400 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
[...]

PR #2325, opened 2020-10-09
Clean out orphan database at startup (optionally) and before a block sync

hansieodendaal:
## Description
This PR adds the ability to optionally clean out the orphan database at startup and as well as before block sync of a preset size. _(Note: As a side issue, this PR also removes `\n` characters from the `Display for TransactionOutput` to make it more readable in the log files.)_

The orphan database is used as an integral part of normal block sync and for reorgs. Every block that needs to be added to the blockchain database in normal sync operation is validated and submitted to the orphan database; if validation succeeds it is removed from the orphan database and added to the main blockchain database. As part of this process, the entire orphan database is searched to discover and validate links to other blocks, however, searching through the entire orphan database every time can be hugely inefficient, especially if there are many orphan blocks in the orphan database (see related issue #2247). 

Reorgs will still happen as per the current design as the orphan database will only be cleared once before missing blocks are requested and added to the database.

As can be seen below orphan search is done for every block added to the blockchain, and if a coupe of thousand blocks need to be synced, this will be hugely ineffective if there are many orphans. This change improves sync performance considerably.
``` rust
2020-10-09 07:15:19.044207800 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:23.087208500 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:27.297718300 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:32.445717800 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:36.493718400 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:40.719718700 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:44.851719100 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:48.913227100 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:53.902735900 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:15:58.066735500 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:16:02.222735400 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
2020-10-09 07:16:06.221737100 [c::cs::database] DEBUG Searched 1031 orphan(s) in 4s
```

## Motivation and Context
Blockchain sync needs to fast and efficient.

## How Has This Been Tested?
Tested in a base node in Windows

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/builder.rs
+++ b/applications/tari_base_node/src/builder.rs
@@ -230,16 +230,33 @@ pub async fn configure_and_initialize_node(
     node_identity: Arc<NodeIdentity>,
     wallet_node_identity: Arc<NodeIdentity>,
     interrupt_signal: ShutdownSignal,
+    cleanup_orphans_at_startup: bool,
 ) -> Result<BaseNodeContext, anyhow::Error>
 {
     let result = match &config.db_type {
         DatabaseType::Memory => {
             let backend = MemoryDatabase::<HashDigest>::default();
-            build_node_context(backend, node_identity, wallet_node_identity, config, interrupt_signal).await?
+            build_node_context(
+                backend,
+                node_identity,
+                wallet_node_identity,
+                config,
+                interrupt_signal,
+                cleanup_orphans_at_startup,
+            )
+            .await?
         },
         DatabaseType::LMDB(p) => {
             let backend = create_lmdb_database(&p, config.db_config.clone(), MmrCacheConfig::default())?;
-            build_node_context(backend, node_identity, wallet_node_identity, config, interrupt_signal).await?
+            build_node_context(
+                backend,
+                node_identity,
+                wallet_node_identity,
+                config,
+                interrupt_signal,
+                cleanup_orphans_at_startup,
+            )
+            .await?
         },
     };
     Ok(result)
@@ -262,6 +279,7 @@ async fn build_node_context<B>(
     wallet_node_identity: Arc<NodeIdentity>,
     config: &GlobalConfig,
     interrupt_signal: ShutdownSignal,
+    cleanup_orphans_at_startup: bool,
 ) -> Result<BaseNodeContext, anyhow::Error>
 where
     B: BlockchainBackend + 'static,
@@ -279,7 +297,7 @@ where
         pruning_horizon: config.pruning_horizon,
         pruning_interval: config.pruned_mode_cleanup_interval,
     };
-    let db = BlockchainDatabase::new(backend, &rules, validators, db_config)?;
+    let db = BlockchainDatabase::new(backend, &rules, validators, db_config, cleanup_orphans_at_startup)?;
     let mempool_validator = MempoolValidators::new(
         TxInternalConsistencyValidator::new(factories.clone()).and_then(TxInputAndMaturityValidator::new(db.clone())),
         TxInputAndMaturityValidator::new(db.clone()),
--- a/applications/tari_base_node/src/recovery.rs
+++ b/applications/tari_base_node/src/recovery.rs
@@ -103,7 +103,7 @@ pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), anyhow::Erro
         pruning_horizon: node_config.pruning_horizon,
         pruning_interval: node_config.pruned_mode_cleanup_interval,
     };
-    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config)?;
+    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config, true)?;
     do_recovery(db, temp_db).await?;
 
     info!(
@@ -134,7 +134,8 @@ async fn do_recovery<D: BlockchainBackend + 'static>(
     // We dont care about the values, here, so we just use mock validators, and a mainnet CM.
     let rules = ConsensusManagerBuilder::new(NetworkType::LocalNet).build();
     let validators = Validators::new(MockValidator::new(true), MockValidator::new(true));
-    let temp_db_backend = BlockchainDatabase::new(temp_db, &rules, validators, BlockchainDatabaseConfig::default())?;
+    let temp_db_backend =
+        BlockchainDatabase::new(temp_db, &rules, validators, BlockchainDatabaseConfig::default(), false)?;
     let max_height = temp_db_backend
         .get_chain_metadata()
         .map_err(|e| anyhow!("Could not get max chain height: {}", e))?
--- a/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
@@ -73,6 +73,7 @@ pub struct BlockSyncConfig {
     pub max_add_block_retry_attempts: usize,
     pub header_request_size: usize,
     pub block_request_size: usize,
+    pub orphan_db_clean_out_threshold: usize,
 }
 
 #[derive(Clone, Debug, PartialEq, Default)]
@@ -114,6 +115,7 @@ impl Default for BlockSyncConfig {
             max_add_block_retry_attempts: MAX_ADD_BLOCK_RETRY_ATTEMPTS,
             header_request_size: HEADER_REQUEST_SIZE,
             block_request_size: BLOCK_REQUEST_SIZE,
+            orphan_db_clean_out_threshold: 0,
         }
     }
 }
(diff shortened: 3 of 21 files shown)


SWvheerden (dismissed):
LGTM

base_layer/core/tests/chain_storage_tests/chain_backend.rs
```diff
@@ -109,7 +109,10 @@ fn lmdb_insert_contains_delete_and_fetch_header() {
 
     // Cleanup test data - in Windows the LMBD `set_mapsize` sets file size equals to map size; Linux use sparse files
     if std::path::Path::new(&temp_path).exists() {
-        std::fs::remove_dir_all(&temp_path).unwrap();
+        match std::fs::remove_dir_all(&temp_path) {
```
sdbondi, line 112: These should probably rather use `tempfile::tempdir()` which cleans up the temporary file on drop. Any reason why these changed?

sdbondi (dismissed):
Although this is fixing the symptom, I think it's worth trying this out.

base_layer/core/tests/chain_storage_tests/chain_backend.rs
hansieodendaal: Yes, this is a funny one, as something changed recently on my one Windows computer that changed the default behaviour for a while (but not on the other). For some reason, this call did not remove the db folder as intended, as it seems the handles were not released properly. The `.unwrap()` in `std::fs::remove_dir_all(&temp_path).unwrap();` caused a runtime exception, which in turn caused the unit tests to fail. However, it was possible to delete the test folder afterwards with File Explorer.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -322,6 +325,25 @@ async fn synchronize_blocks<B: BlockchainBackend + 'static>(
                 info.tip_height = network_tip_height;
             }
 
+            // Searching through the orphan database for every block to be added during a large block sync is not
+            // efficient at the moment
+            if network_tip_height as i64 - sync_height as i64 > MISSING_BLOCKS_THRESHOLD as i64 {
```
stringhandler: I don't think this is the correct place for this. 

If the large block sync was efficient, where would you do this? In other words, if that problem did not exist, would you still add this code? Let's rather add it in the correct place, if it's needed and fix the large block sync efficiency. Otherwise we end up with 2 pieces of code that need to be cleaned up later

stringhandler (commented):
I agree with the principle that there are cases where you would want to clean up the orphan database. I think it should be an option that you maybe start up with. If you wanted to clean the orphan db, then you could restart the node with that option. 

However I don't agree with the implementation. Checking and deleting the orphan db just before a large sync feels like a bandaid over a larger problem, and when that problem is eventually solved, this will cause unexpected behaviours

hansieodendaal:
> I agree with the principle that there are cases where you would want to clean up the orphan database. I think it should be an option that you maybe start up with. If you wanted to clean the orphan db, then you could restart the node with that option.
> 

This is a great comment. Added optional command-line argument to do orphans db cleanup.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
hansieodendaal: I have updated the PR description to motivate why this test and cleanup is in the correct place. It is also now a user-configurable option.

applications/tari_base_node/src/builder.rs
```diff
@@ -230,16 +230,33 @@ pub async fn configure_and_initialize_node(
...
         DatabaseType::Memory => {
             let backend = MemoryDatabase::<HashDigest>::default();
-            build_node_context(backend, node_identity, wallet_node_identity, config, interrupt_signal).await?
+            build_node_context(
+                backend,
+                node_identity,
+                wallet_node_identity,
+                config,
+                interrupt_signal,
+                cleanup_orphans_at_startup,
```
SWvheerden, line 245: Its already in the config? Rather than passing in a bool, you can just call `config. orphan_db_clean_out_threshold>0` where you need to pass it into the db.

applications/tari_base_node/src/recovery.rs
```diff
@@ -103,7 +103,7 @@ pub async fn run_recovery(node_config: &GlobalConfig) -> Result<(), anyhow::Erro
         pruning_horizon: node_config.pruning_horizon,
         pruning_interval: node_config.pruned_mode_cleanup_interval,
     };
-    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config)?;
+    let db = BlockchainDatabase::new(main_db, &rules, validators, db_config, false)?;
```
SWvheerden: I would either set this as true, or make it configurable. 
Recovery mode will essentially strip and clean you normal db leaving it only with the blocks passing validation in the db. Following this methodology, I would not set it hard coded to false.

applications/tari_base_node/src/recovery.rs
```diff
@@ -134,7 +134,8 @@ async fn do_recovery<D: BlockchainBackend + 'static>(
     // We dont care about the values, here, so we just use mock validators, and a mainnet CM.
     let rules = ConsensusManagerBuilder::new(NetworkType::LocalNet).build();
     let validators = Validators::new(MockValidator::new(true), MockValidator::new(true));
-    let temp_db_backend = BlockchainDatabase::new(temp_db, &rules, validators, BlockchainDatabaseConfig::default())?;
+    let temp_db_backend =
```
SWvheerden, line 137: This is fine, as this is not using its orphan pool at all.

base_layer/core/src/transactions/transaction.rs
```diff
@@ -399,7 +399,7 @@ impl Display for TransactionOutput {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
         let proof = self.proof.to_hex();
         fmt.write_str(&format!(
-            "{} [{:?}] Proof: {}..{}\n",
+            "({} [{:?}] Proof: {}..{})",
```
SWvheerden, line 402: see review comment

base_layer/core/tests/chain_storage_tests/chain_storage.rs
```diff
@@ -1289,7 +1301,10 @@ fn invalid_block() {
 
     // Cleanup test data - in Windows the LMBD `set_mapsize` sets file size equals to map size; Linux use sparse files
     if std::path::Path::new(&temp_path).exists() {
-        std::fs::remove_dir_all(&temp_path).unwrap();
+        match std::fs::remove_dir_all(&temp_path) {
```
SWvheerden, line 1304: see review comment

base_layer/core/tests/chain_storage_tests/chain_storage.rs
```diff
@@ -1157,7 +1160,10 @@ fn restore_metadata_and_pruning_horizon_update() {
 
     // Cleanup test data - in Windows the LMBD `set_mapsize` sets file size equals to map size; Linux use sparse files
     if std::path::Path::new(&path).exists() {
-        std::fs::remove_dir_all(&path).unwrap();
+        match std::fs::remove_dir_all(&path) {
```
SWvheerden, line 1163: see review comment

base_layer/core/tests/chain_storage_tests/chain_storage.rs
```diff
@@ -1019,7 +1019,10 @@ fn handle_reorg_failure_recovery() {
     }
     // Cleanup test data - in Windows the LMBD `set_mapsize` sets file size equals to map size; Linux use sparse files
     if std::path::Path::new(&temp_path).exists() {
-        std::fs::remove_dir_all(&temp_path).unwrap();
+        match std::fs::remove_dir_all(&temp_path) {
```
SWvheerden, line 1022: see review comment

SWvheerden (dismissed):
Looks good, I like it. I made one comment where I think we can improve this.

I also made a few comments: "see review comment"
There is refactoring done in this PR on code that was not part of the purpose of the PR. This PR, is for adding a cleanup of the orphan database. We should not "sneak" in fixes or improvements with other PRs. We should clearly label all changes. And preferably, keep them separate. This makes reviewing easier and will improve the quality of the review.

base_layer/core/tests/chain_storage_tests/chain_storage.rs
hansieodendaal: The tests failed in Windows under certain system curcimstances, so this had to be done to make the unit tests resilient. For some reason the test folders could not be deleted, so the code panicked at the `.unwrap()`, so this change was necessary.

base_layer/core/tests/chain_storage_tests/chain_storage.rs
hansieodendaal: As per above

base_layer/core/tests/chain_storage_tests/chain_storage.rs
hansieodendaal: As per above

base_layer/core/src/transactions/transaction.rs
hansieodendaal: Agreed, this should not have been changed without mentioning it. Can we live with this change as well, as it will make the log files more readable?

applications/tari_base_node/src/recovery.rs
hansieodendaal: Agreed, will change to `true`

applications/tari_base_node/src/builder.rs
hansieodendaal: No, `GlobalConfig` has an `orphan_db_clean_out_threshold` member that is fed in form the config file, which is different than `cleanup_orphans_at_startup` that is fed in from the command line.

applications/tari_base_node/src/recovery.rs
hansieodendaal: Agreed

base_layer/core/src/transactions/transaction.rs
hansieodendaal: Updated the PR description.

stringhandler (approved):

PR #2333, opened 2020-10-13
Fix off by one height error in target difficulty

stringhandler:
Fixes a subtle bug where the target difficulty would be validated against the current height, before a block was added, instead of the height at the block

### DIFF
--- a/base_layer/core/src/validation/helpers.rs
+++ b/base_layer/core/src/validation/helpers.rs
@@ -81,7 +81,6 @@ pub fn check_median_timestamp<B: BlockchainBackend>(
 pub fn check_achieved_and_target_difficulty<B: BlockchainBackend>(
     db: &B,
     block_header: &BlockHeader,
-    height: u64,
     rules: ConsensusManager,
 ) -> Result<(), ValidationError>
 {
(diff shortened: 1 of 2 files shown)


base_layer/core/src/validation/helpers.rs
```diff
@@ -81,7 +81,7 @@ pub fn check_median_timestamp<B: BlockchainBackend>(
 pub fn check_achieved_and_target_difficulty<B: BlockchainBackend>(
     db: &B,
     block_header: &BlockHeader,
-    height: u64,
+    _height: u64,
```
philipr-za: If it is not needed maybe just take it out?

sdbondi (dismissed):
Nice catch!

philipr-za (approved):
LGTM

sdbondi (approved):

SWvheerden:
> Fixes a subtle bug where the target difficulty would be validated against the current height, before a block was added, instead of the height at the block

This is wrong, the bug is the target_difficulty calculation of the block template creation. Not the validation. 
It needs to calculate the target difficulty using the current height. It cannot use the height of the block we are adding as this block does not yet exists in the database. the target difficulty for block of height N is calculated using the the heights of block N-1  up onto N-block window. 

PR #2335 fixes the bug

PR #2250, opened 2020-09-20
[Merged Mining] Difficulty Check

StriderDM:
## Description
Added 512-bit integer.
Added check_hash function.
Added difficulty to MoneroData.
Added additional helper methods.
Fixed endianess for monero hashes.
Fixed previous submit height for non-proxy submit.

## Motivation and Context
Monero hashes are stored in little endian and not big endian.

## Types of changes
* [x] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [ ] New Tests
* [ ] Documentation

## Checklist:
* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [ ] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #2335, opened 2020-10-13
Fix target difficulty

SWvheerden:
## Description

Fixes the target difficulty to use the correct height to construct the target difficulty. 
Bug was introduced in PR #2250. By calculating the difficulty using the incorrect block height. It was given the block height of the block the difficulty was for, but this is incorrect as the block height needs to be the tip. As the difficulty calculation uses the tip height and the calculates the target for the NEXT block, not the current block. 

Reverting PR #2333 as this changes the validation to match the creation which is wrong. Validation was right. This now means the validation and calculation of the target difficulty is off by one height. As 2333 calculates the difficulty as N-2, where N is the block_window.

This PR adjusts it back to it N-1. See PR #2217  for why this is currently the case. 

## Motivation and Context

Calculate Target difficulty always calculates the difficulty for the next block. So it assumes the height given is the tip.  The reason for this, is that you never use the current block in the target_difficulty, the target_difficulty is only determined by past blocks.  

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -523,6 +523,10 @@ where T: BlockchainBackend + 'static
         height: u64,
     ) -> Result<Difficulty, CommsInterfaceError>
     {
+        let height_of_longest_chain = async_db::get_chain_metadata(self.blockchain_db.clone())
+            .await?
+            .height_of_longest_chain
+            .ok_or_else(|| CommsInterfaceError::UnexpectedApiResponse)?;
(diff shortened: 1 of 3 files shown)


stringhandler (approved):
Seems counterintuitive to use the previous block and not the current one, but I suppose it makes it harder to manipulate.

SWvheerden:
> Seems counterintuitive to use the previous block and not the current one, but I suppose it makes it harder to manipulate.

The call thats changed is the code that fetches difficulties from the DB. You cant fetch a height that does not yet exist in the db.

PR #2337, opened 2020-10-14
Update TUI Transaction listing and Contact management

philipr-za:
## Description

WARNING: Do not run this app and the base node simultaneously, can corrupt your dbs

This PR firstly cleans up the TUI implementation a bit further from the initial refactor now that we are working with real data from the backend rather than dummy data. This involves a little bit or reorganisation. The main contributions of this PR are an updated StatefulList that can provide a windowed view into the list data and contact management.

The update to the StatefulList implementation was required because the naive implementation put forward by tui-rs will parse and format the entire list for every draw operation. This was a problem on a wallet with thousands of transactions which made the UI very laggy. This PR adds functionality to the Stateful list which will produce a windowed view into the list data based on what items are selected and the provided height of the list widget that will be rendered so only the minimum required items are formatted for rendering. To avoid rewriting the whole List widget from tui-rs this is achieved by outputting the windowed slice of the data and a generated ListState object that the native tui-rs render method expects.

The second part of this PR is adding the implementation to allow for managing Contacts in the Send/Receive tab. This involved completing the UI elements for listing contacts and adding, editing or deleting contacts. The initial UI logic for selecting a contact to send to, adding an amount and hitting Send Transaction is added though only the validation is done and not the actual sending yet. Errors are also displayed using an Error popup dialog.

## How Has This Been Tested?
Tests are provided for the updates to the StatefulList and the windowed view.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


delta1:
A few clippy warnings to fix up, which also caused the build to fail in Circle CI. 

Will start testing and review asap

philipr-za:
Yup, forgot clippy.

delta1 (approved):
This is really great work 👏 

The new windowed list works like a bomb, I could easily scroll through thousands of previous txs. 

Also tested the contacts management which worked great. My only suggestion there is that deleting a contact should ask for confirmation (or allow you to undo the delete) in case a cat jumps on your D key or something 😸 but that can go in the next PR.

applications/tari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -283,7 +284,7 @@ impl TransactionsTab {
 }
 
 impl<B: Backend> Component<B> for TransactionsTab {
-    fn draw(&mut self, f: &mut Frame<B>, area: Rect, app_state: &AppState) {
+    fn draw(&mut self, f: &mut Frame<B>, area: Rect, app_state: &mut AppState) {
```
stringhandler: `draw` was intended to only read from AppState. The idea was events change the state, and then the draw method only reads from them. That was the intention, but it's not always possible. Which part needs it to be mut?

philipr-za: the state that keeps track of which item is selected in the list based on the height of the rendered list which is only available at this point.

delta1: I think it's for the windowed items in the StatefulList? (edit: sorry this comment got posted before I saw Phil's reply!)

philipr-za: Correct, previously we kept the list selection state inside the component itself but for the list selection logic to work properly you would need to supply the full set of data into the render function call. To produce the windowed view into the list it had to move into the StatefulList.

delta1 (changes requested):
woops, seems there was a test failure

```
error[E0599]: no function or associated item named `with_items` found for struct `ui::widgets::stateful_list::StatefulList<_>` in the current scope
   --> applications/tari_console_wallet/src/ui/widgets/stateful_list.rs:128:39
```

philipr-za:
> woops, seems there was a test failure
> 
> ```
> error[E0599]: no function or associated item named `with_items` found for struct `ui::widgets::stateful_list::StatefulList<_>` in the current scope
>    --> applications/tari_console_wallet/src/ui/widgets/stateful_list.rs:128:39
> ```

Weird I ran `cargo test --all-features` locally. Wonder what is different? Will fix while I rebase.

Edit: might have been due to a clippy fix.

philipr-za:
> Also tested the contacts management which worked great. My only suggestion there is that deleting a contact should ask for confirmation (or allow you to undo the delete) in case a cat jumps on your D key or something 😸 but that can go in the next PR.

Nah, this app is a Touch Move environment and people need to learn to get used to it! 

I have added it to the To Do list for later PRs, a confirmation dialog will be useful in a number of places.

applications/tari_console_wallet/src/ui/components/transactions_tab.rs
philipr-za: Alright, I separated the ListState offset into its own struct so it can live in the component and AppState doesn't need to be mutable.

delta1 (dismissed):
LGTM

PR #2340, opened 2020-10-14
Expose latency for last call for client RPC connections

sdbondi:
Latency measurement of the last call comes for free in RPC. This PR
exposes the latency to users of the RPC client.


stringhandler (approved):
Looking good. Could have also been done with a tokio::watch, but this is great

PR #2343, opened 2020-10-15
Remove Arc<RwLock<>> from Wallet

delta1:
## Description

Removes the unnecessary `Arc<RwLock<_>>` from the Wallet, and updates the cascading changes. The reasoning for removal is that the Wallet object should already be a thread-safe and lightweight collection of handles, and so be easily cloned when necessary. 

For the Wallet to derive `Clone`, it is required that the WalletBackend and ContactsBackend implementors are also Clone. Additionally, the `Shutdown` type previously held in the Wallet is hoisted up a level into the client applications creating the wallet.  

## Motivation and Context

From feedback in [#2291](https://github.com/tari-project/tari/pull/2291#discussion_r503287026) 

## How Has This Been Tested?

- [x] cargo test
- [x] manually

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -1,42 +1,13 @@
-use std::sync::Arc;
 use tari_app_grpc::tari_rpc::{wallet_server, GetCoinbaseRequest, GetCoinbaseResponse};
-use tari_wallet::{
-    contacts_service::storage::sqlite_db::ContactsServiceSqliteDatabase,
-    output_manager_service::storage::sqlite_db::OutputManagerSqliteDatabase,
-    storage::sqlite_db::WalletSqliteDatabase,
-    transaction_service::storage::sqlite_db::TransactionServiceSqliteDatabase,
-    Wallet,
-};
-use tokio::sync::RwLock;
+use tari_wallet::WalletSqlite;
 use tonic::{Request, Response, Status};
 
 pub struct WalletGrpcServer {
-    wallet: Arc<
-        RwLock<
-            Wallet<
-                WalletSqliteDatabase,
-                TransactionServiceSqliteDatabase,
-                OutputManagerSqliteDatabase,
-                ContactsServiceSqliteDatabase,
-            >,
-        >,
-    >,
+    wallet: WalletSqlite,
 }
 
 impl WalletGrpcServer {
-    pub fn new(
-        wallet: Arc<
-            RwLock<
-                Wallet<
-                    WalletSqliteDatabase,
-                    TransactionServiceSqliteDatabase,
-                    OutputManagerSqliteDatabase,
-                    ContactsServiceSqliteDatabase,
-                >,
-            >,
-        >,
-    ) -> Self
-    {
+    pub fn new(wallet: WalletSqlite) -> Self {
         Self { wallet }
     }
 }
@@ -48,13 +19,12 @@ impl wallet_server::Wallet for WalletGrpcServer {
     ) -> Result<Response<GetCoinbaseResponse>, Status>
     {
         let request = request.into_inner();
-        let response = self
-            .wallet
-            .write()
-            .await
-            .transaction_service
+
+        let mut tx_service = self.wallet.transaction_service.clone();
+        let response = tx_service
             .generate_coinbase_transaction(request.reward.into(), request.fee.into(), request.height)
             .await;
+
         match response {
             Ok(resp) => Ok(Response::new(GetCoinbaseResponse {
                 transaction: Some(resp.into()),
--- a/applications/tari_console_wallet/src/main.rs
+++ b/applications/tari_console_wallet/src/main.rs
@@ -18,6 +18,7 @@ use tari_comms::{peer_manager::PeerFeatures, NodeIdentity};
 use tari_comms_dht::{DbConnectionUrl, DhtConfig};
 use tari_core::{consensus::Network as NetworkType, transactions::types::CryptoFactories};
 use tari_p2p::initialization::CommsConfig;
+use tari_shutdown::{Shutdown, ShutdownSignal};
 use tari_wallet::{
     error::WalletError,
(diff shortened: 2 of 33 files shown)


applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -7,33 +6,24 @@ use tari_wallet::{
...
-                TransactionServiceSqliteDatabase,
-                OutputManagerSqliteDatabase,
-                ContactsServiceSqliteDatabase,
-            >,
-        >,
+    wallet: Wallet<
+        WalletSqliteDatabase,
+        TransactionServiceSqliteDatabase,
+        OutputManagerSqliteDatabase,
+        ContactsServiceSqliteDatabase,
```
philipr-za:
```suggestion
    wallet: WalletSqlite
```

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -7,33 +6,24 @@ use tari_wallet::{
...
-                    TransactionServiceSqliteDatabase,
-                    OutputManagerSqliteDatabase,
-                    ContactsServiceSqliteDatabase,
-                >,
-            >,
+        wallet: Wallet<
+            WalletSqliteDatabase,
+            TransactionServiceSqliteDatabase,
+            OutputManagerSqliteDatabase,
+            ContactsServiceSqliteDatabase,
```
philipr-za:
```suggestion
        wallet: WalletSqlite
```

applications/tari_console_wallet/src/main.rs
```diff
@@ -43,17 +43,23 @@ mod utils;
 
 /// Application entry point
 fn main() {
-    match main_inner() {
-        Ok(_) => std::process::exit(0),
+    let mut shutdown = Shutdown::new();
+
+    match main_inner(shutdown.to_signal()) {
+        Ok(_) => {
+            shutdown.trigger().unwrap();
```
philipr-za: After we trigger the shutdown we need to wait on the wallet to complete its shutdown (which really just waits for the comms stack to finish its shutdown). 

I recommend we change the wallets shutdown (copied below) to a wait_until_shutdown so we let the client do the triggering.

```Rust
pub async fn wait_until_shutdown(mut self) {
        self.comms.wait_until_shutdown().await;
    }
```

Also whenever you are using an unwrap it is better to use an .expect("...") with a message saying what was supposed to happen.

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -60,15 +60,11 @@ pub const LOG_TARGET: &str = "wallet::ui::app";
...
-                TransactionServiceSqliteDatabase,
-                OutputManagerSqliteDatabase,
-                ContactsServiceSqliteDatabase,
-            >,
-        >,
+    pub wallet: Wallet<
+        WalletSqliteDatabase,
+        TransactionServiceSqliteDatabase,
+        OutputManagerSqliteDatabase,
+        ContactsServiceSqliteDatabase,
```
philipr-za:
```suggestion
    pub wallet: WalletSqlite
```

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -80,15 +76,11 @@ impl<B: Backend> App<B> {
...
-                    TransactionServiceSqliteDatabase,
-                    OutputManagerSqliteDatabase,
-                    ContactsServiceSqliteDatabase,
-                >,
-            >,
+        wallet: Wallet<
+            WalletSqliteDatabase,
+            TransactionServiceSqliteDatabase,
+            OutputManagerSqliteDatabase,
+            ContactsServiceSqliteDatabase,
```
philipr-za:
```suggestion
        wallet: WalletSqlite
```

base_layer/wallet/src/wallet.rs
```diff
@@ -221,8 +220,8 @@ where
 
     /// This method consumes the wallet so that the handles are dropped which will result in the services async loops
     /// exiting.
-    pub async fn shutdown(mut self) {
-        self.shutdown.trigger().expect(
+    pub async fn shutdown(self, mut shutdown: Shutdown) {
```
philipr-za: I propose that we rather change this into a wait_for_shutdown(...) and have the client's trigger the shutdown. So in the FFI lib you will need to do the following (In the Wallet FFI case we do not want to panic if the shutdown triggers are wrong so not using an expect() there)

```Rust
pub unsafe extern "C" fn wallet_destroy(wallet: *mut TariWallet) {
    if !wallet.is_null() {
        let mut m = Box::from_raw(wallet);
        match m.shutdown.trigger() {
            Err(_) => error!(target: LOG_TARGET, "No one is listening for the shutdown trigger!"),
            Ok(()) =>  m.runtime.block_on(m.wallet.wait_for_shutdown()),
            
        }
    }
}
```
And the equivalent in the console wallet and also the base node while there is still a wallet in there.

philipr-za (commented):
Getting there, you will also need to wire the Base Node shutdown into its wallet.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
delta1: Thanks was just busy with this

applications/tari_console_wallet/src/main.rs
delta1: Excellent thanks will do

base_layer/wallet/src/wallet.rs
delta1: Will do

delta1:
> Getting there, you will also need to wire the Base Node shutdown into its wallet.

~Cool will do - will have to investigate how the BN is starting its wallet~ 

As discussed, Base Node uses the WalletBootstrap and already wires up a shutdown signal.

stringhandler (approved):

delta1:
Thanks for the merge but I think they may be an issue with the tests...

Issue #2197, opened 2020-09-07
Improve initial blockchain sync with optimized receiving of blocks from peers


hansieodendaal:
This proposal is (tick applicable):

* [X] A new feature
* [X] An extension to an existing RFC/proposal (See #2113)
* [ ] An alternative approach to an existing RFC or proposal <LINK>

### Change log

| Date | Changes |
|:-----|:--------|
|      |         |

### Goals

To minimize delays between receiving blocks during the initial sync.

### Assumptions

This is dependent on RPC to be implemented.

### Abstract

Currently, one of the main reasons initial block sync is slow is due to the fact that blocks are requested in groups of 5 from a single peer, then when those blocks are received, they are processed, validated and added to the database before the next group of 5 blocks are requested form the same or another peer. This process can be enhanced by requesting and receiving blocks continuously in parallel to the validation processing.

## Description

Stream all block requests, without waiting for validation to be concluded for the next round to start. If a block in a group cannot be added to the database successfully, that group can be re-requested from a different host.

sdbondi:
My (incomplete) proposal for the RPC-based archival block sync is detailed here:

Reading: https://developer.bitcoin.org/devguide/p2p_network.html#blocks-first

## Initial Phase - Selecting Sync Peers

We start with the assumption that the node is online in the Tari comms connectivity sense, roughly it has a few connections to other base nodes. We select all those connections and attempt to establish RPC sessions with each base node.

If the RPC session succeeds, we know:
1. The peer supports the sync protocol we want to speak
1. The ping time for the peer
1. We have some assurance that the peer is prepared to offer syncing to us (is not overloaded etc)

At this point we have rooted out peers that will not respond to sync requests and peers with slow response times (although we do not have a measure of database fetch performance, it is conceivable that this can be implemented if ever needed). 

## Are we on the same chain?

Current strategy: Peers exchange chain metadata. The peer/s that claims to have the best metadata are chosen to sync. The peer's tip header is downloaded and validated. Once validated, it is assumed that the peer is not wasting our time and has the strongest chain. The disadvantage of this approach is that nodes can lie about their chain metadata and force a node to think that it has fallen behind, therefore entering "block sync mode". A potential DoS vector.

This may be mitigated by demanding partial proof that the strongest chain can be offered (some headers near the tip, Nakamoto consensus) and perhaps comparing that with other peers to increase confidence.

**Alternate strategy**

[...]

Issue #2113, opened 2020-08-07
Speed up sync times


stringhandler:
Testnet sync times are getting really slow

hansieodendaal:
# Blockchain sync profiling 2020/08/18

## Test Description

Two fresh base nodes were started (using Tor) in order to profile blockchain sync from scratch with trace logs being captured for both; one running on Windows and the other in an Ubuntu Linux virtual machine on the same host. Test conditions were not favorable as the host machine was used for many other high intensity tasks during this time. However, overall measurements still proved to be good enough.

Sync started when the blockchain tip was at block 74392; overview shown below:

| Basenode | Sync start time   | Blockchain  synced time | Duration | Duration (s) | Synced at block | Blocks/s |
| -------- | ----------------- | ----------------------- | -------- | ------------ | --------------- | -------- |
| Windows  | 2020/08/11  21:50 | 2020/08/12  15:52       | 18:01:28 | 64887.71     | 74919           | 1.15     |
| Linux    | 2020/08/11  21:51 | 2020/08/12  15:05       | 17:14:53 | 62093.44     | 74893           | 1.21     |

## Analysis

Analysis is based on detail time measurements in the code as well as differential time between specific consecutive trace logs. Differential analysis only uses log entries received with blocks in chronological order, as new blocks that were mined were also received via broadcast from peers.

### General Profiling

These general profiling measurements are extracted form the logs.

| #    | Log Entry                                                    | Description               |
| ---- | ------------------------------------------------------------ | ------------------------- |
| 1    | Requesting blocks [1, 2, 3, 4, 5] from 6b4773e599143ba0f87db39da2. | Block request time        |
| 2    | Received 5 blocks from peer                                  | Block receive time        |
| 3    | Response for HistoricalBlocks (request key: 5564039840279151865) received  after 734ms | Network response (direct) |
| 4    | [add_block] Exited blocking thread after 154ms. trace_id: '4048701038' | Block processing time     |

### Block Processing Profile

The table below shows a typical trace log of a new block being received for processing after the previous block has been successfully added to the database. The numbers in `Diff. Time` indicate which log messages are extracted for differential time measurements to be done. The first measurement would be between `0` and `1`, the next one between `1` and  `2`, and so on for all blocks received as part of block sync.

| Diff. Time | Log Entry                                                    | Description                       |
| ---------- | ------------------------------------------------------------ | --------------------------------- |
[...]

PR #2348, opened 2020-10-16
Base node block sync RPC service

sdbondi:
## Description

RPC service for faster block sync. Allows streaming of full blocks,
headers and a method for chain split detection. The client-side is next
up.

- Added BlockchainStateService. This service should eventually be the only service
  which provides a query interface to the blockchain database for other
  services.
- Added BaseNodeSyncRpcService. This RPC service exposes `sync_blocks`,
  `sync_headers`, `fetch_header_by_height` and `find_chain_split`. These
  functions will be used in the improved block sync.
- Added `NotFound` status to RpcStatus
- Added `RpcStatus::log_internal_error()` convenience function to make
  dealing with misc errors in RPC handler functions easier.
- Added `spawn_until_shutdown` method to the service initialiser context 
   that spawns a task until the shutdown is triggered. This removes the need for   
   implementers to worry about how to implement shutdown in initialisers.
- Cleaned up some test helpers
- Added unit tests for the Base node sync RPC handlers
- Added unit tests for the BlockchainStateService
- Remove unused `tokio-test` dependency

## Motivation and Context

Described here: https://github.com/tari-project/tari/issues/2197#issuecomment-708502628
Ref #2113

## How Has This Been Tested?

Extensive unit tests

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


philipr-za (dismissed):
Not hugely familiar with the context of this PR but LGTM

base_layer/core/src/base_node/service/blockchain_state/service.rs
```diff
@@ -0,0 +1,211 @@
...
+                reply,
+                async_db::fetch_header(self.db(), height).map_err(Into::into),
+            )),
+            GetChainMetadata(reply) => Box::pin(reply_or_cancel(
+                reply,
+                async_db::get_chain_metadata(self.db()).map_err(Into::into),
+            )),
+            FindFirstMatchingHeader(hashes, reply) => Box::pin(reply_or_cancel(
+                reply,
+                Self::find_first_matching_header(self.db(), hashes),
```
stringhandler: In my opinion this should start sending blocks or headers back immediately. The rational is that a malicious node could advertise a false tip height, then on request, it can return a valid header and then ghost you on the next call.

If the sender must immediately provide a chain, it must have created a fake chain in order to respond to this request. This reduces the attack surface significantly.

base_layer/core/src/base_node/service/blockchain_state/service.rs
```diff
@@ -0,0 +1,211 @@
...
+            )),
+        }
+    }
+
+    async fn fetch_blocks(
+        db: BlockchainDatabase<B>,
+        start: Bound<u64>,
```
stringhandler, line 147: Apart from local querying and GRPC, I don't see the value in using heights instead of hashes. The sync node's chain could change during this step or in between calls, making this query invalid. Perhaps pass in a starting hash and a number of blocks.

base_layer/core/src/base_node/service/blockchain_state/service.rs
```diff
@@ -0,0 +1,211 @@
...
+        // `HistoricalBlock`s are converted to `Block`s here because we're exposing a minimal required interface.
+        // If the backends are refactored and it becomes an issue to include the extra historical data along with
+        // the block, this will make that refactor easier because nothing can rely on that data if it isn't exposed.
+        // If however, the historical data is needed, then exposing it here is a much easier refactor.
+        Ok(blocks.into_iter().map(|b| b.block).collect())
+    }
+
+    async fn fetch_headers(
+        db: BlockchainDatabase<B>,
+        start: Bound<u64>,
```
stringhandler, line 181: Same note here on the heights vs using hashes

comms/src/protocol/rpc/status.rs
```diff
@@ -157,8 +171,10 @@ pub enum RpcStatusCode {
     MalformedResponse = 5,
     /// Misc. errors
     General = 6,
+    /// Entity not found
+    NotFound = 7,
     /// Unrecognised RPC status code
-    InvalidRpcStatusCode = 7,
+    InvalidRpcStatusCode = 100,
```
stringhandler:
```suggestion
    InvalidRpcStatusCode = 500,
```

stringhandler (commented):
Looks good. I think the use of heights instead of hashes could let to unexpected responses though. I would also prefer the FindChainSplit to immediately return headers to increase the cost of spoofing

comms/src/protocol/rpc/status.rs
sdbondi: So this variant actually isn't a specific code, it stands for "anything else other than the above". The only reason it's there at all is because ~~we have to put a number there~~(guess not - rust is happy ;)). I should add a comment for that.

That is why moving this number to > 7 doesnt actually break anything (other than older clients will say invalid status code instead of not found). So we can make this whatever we want really, it's more what makes the code clearer. I would prefer it to be within a u8 though, it doesn't matter at all, but if in the future it does, it's better to save the byte (varint encoding).

base_layer/core/src/base_node/service/blockchain_state/service.rs
sdbondi: Great idea! Will change the api.

comms/src/protocol/rpc/status.rs
sdbondi: I think it'll actually be clearer if we leave the number off completely. Rust will auto assign it, and it's clearer that the variant doesn't represent a single code.

base_layer/core/src/base_node/service/blockchain_state/service.rs
sdbondi: Is that a big problem? The RPC protocol will time out and we can move onto another peer (either way we must implement this). They could respond immediately with one header, then not stream the rest (essentially the same thing), so I don't think we can prevent peers from 'ghosting'.  

I'm not sure about this though, my thinking is that we establish the starting point for the sync and that the peer is willing and able to provide the blocks. Then begin the stream. I think that there is no reason in particular that we cannot start streaming a few headers from this call (well, the RPC frontend that uses this service call).

sdbondi: How about this?

```proto
message FindChainSplitRequest {
  repeated bytes block_hashes = 1;
  uint64 count = 2;
}

message FindChainSplitResponse {
  // An ordered list of headers starting from next header after the matching hash, up until `FindChainSplitRequest::count`
  repeated tari.core.BlockHeader headers = 1;
  // The index of the hash that matched from `FindChainSplitRequest::block_hashes`. This value could also be used to know how far back a split occurs.
  uint32 found_hash_index = 2;
}

```

stringhandler (approved):
Looks good. The tests are really nice and easy to read.

PR #2351, opened 2020-10-18
Implement transaction sending and event based updating to TUI app

philipr-za:
## Description
This PR implements transaction sending in the TUI. In addition to actually sending the Tx the PR adds monitoring of the send progress and updates to the transaction status in the transaction lists as the process progresses.

In order to allow events from the backend to update the state that is displayed by the TUI while also being able to update the state based on TUI input events the AppState has been refactored into a thread-safe pattern that manages its own RwLock and provides a rudimentary cached copy of the backend data for the front-end to be able to reference during every draw without cloning. The backend version of the data can be updated asynchronously by events from the backend and the frontend and only when the state has changed will the backend state be cloned to the cached copy. The update of the stale cache data is triggered by the frontend Tick event. The update of the stale cache is currently quite naive and there are a number of potential optimisations possible here to reduce the amount of data being cloned.

To update the state based on asynchronous events coming from the Wallet backend this PR introduces an Event Monitor which is a spawned task that listened for events that mean that transaction data has been updated. When an event occurs for a specific TxId the monitor triggers an update of just that transaction’s state in the AppState backend to minimise database reads.

For transaction sending a message field was added to the Send/Receive tab and popup dialogs were implemented to show success and errors that can be dismissed with a keypress. A “wait” dialog is also added that block further activity on that specific tab while a long operation is in progress (in this case the transaction send) but still allows the user to move and interact with the other tabs. In order to update the “wait” dialog with the current status of the send operation a Tokyo::watch channel is created and sent to a task which initiates the transaction send and then monitors the event bus for updates on its sending operation. The current stage of the operation is then broadcast on the watch channel for the wait dialog to monitor and use to update itself.

Other small improvements in this PR:
- Added the Excess Sig hex to the Detailed transaction view
- Transaction Status now shows if the transaction has been Cancelled

## How Has This Been Tested?
Tests provided for database interface updates.
App manually tested

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


stringhandler (approved):
Nice. Untested, but looking forward to trying it out

delta1 (approved):
Very nice! Tested sending a tx and it worked great. LGTM

PR #2353, opened 2020-10-19
Display connected peers in TUI

philipr-za:
## Description
This PR adds monitoring of the Connectivity Manager event stream and display of the currently connected Peers in the Network Tab of the TUI.

When a connectivity event is detected that would affect the connected peer list the list is updated in the backend.

## How Has This Been Tested?
Manually tested

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/ui/components/network_tab.rs
+++ b/applications/tari_console_wallet/src/ui/components/network_tab.rs
@@ -1,10 +1,10 @@
-use crate::ui::{components::Component, state::AppState};
+use crate::ui::{components::Component, state::AppState, widgets::MultiColumnList, MAX_WIDTH};
 use tui::{
     backend::Backend,
     layout::{Constraint, Layout, Rect},
-    style::{Color, Style},
+    style::{Color, Modifier, Style},
     text::Span,
-    widgets::{Block, Borders, Paragraph, Row, Table, Wrap},
+    widgets::{Block, Borders, ListItem, ListState, Paragraph, Wrap},
     Frame,
 };
 
@@ -14,10 +14,41 @@ impl NetworkTab {
     pub fn new() -> Self {
         Self {}
     }
+
+    pub fn draw_connected_peers_list<B>(&self, f: &mut Frame<B>, area: Rect, app_state: &AppState)
+    where B: Backend {
+        let block = Block::default().borders(Borders::ALL).title(Span::styled(
+            "Connected Peers",
+            Style::default().fg(Color::White).add_modifier(Modifier::BOLD),
+        ));
+        f.render_widget(block, area);
+
+        let list_areas = Layout::default()
+            .constraints([Constraint::Min(1)].as_ref())
+            .margin(1)
+            .split(area);
+
+        let peers = app_state.get_connected_peers();
+        let mut column0_items = Vec::with_capacity(peers.len());
+        let mut column1_items = Vec::with_capacity(peers.len());
+        let mut column2_items = Vec::with_capacity(peers.len());
+        for p in peers.iter() {
+            column0_items.push(ListItem::new(Span::raw(p.node_id.to_string())));
+            column1_items.push(ListItem::new(Span::raw(p.public_key.to_string())));
+            column2_items.push(ListItem::new(Span::raw(p.user_agent.clone())));
+        }
+        let column_list = MultiColumnList::new()
+            .heading_style(Style::default().fg(Color::Magenta))
+            .max_width(MAX_WIDTH)
+            .add_column(Some("NodeID"), Some(27), column0_items)
+            .add_column(Some("Public Key"), Some(65), column1_items)
+            .add_column(Some("User Agent"), Some(MAX_WIDTH.saturating_sub(93)), column2_items);
+        column_list.render(f, list_areas[0], &mut ListState::default());
+    }
 }
 
 impl<B: Backend> Component<B> for NetworkTab {
-    fn draw(&mut self, f: &mut Frame<B>, area: Rect, _app_state: &AppState) {
+    fn draw(&mut self, f: &mut Frame<B>, area: Rect, app_state: &AppState) {
         // This is all dummy content and layout for review
         let main_chunks = Layout::default()
             .constraints([Constraint::Length(1), Constraint::Length(8), Constraint::Min(10)].as_ref())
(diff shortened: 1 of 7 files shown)


applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -367,13 +380,32 @@ impl AppStateInner {
...
+        let connections = self.wallet.comms.connectivity().get_active_connections().await?;
+
+        let mut peers = Vec::new();
+        for c in connections.iter() {
+            if let Ok(p) = self.wallet.comms.peer_manager().find_by_node_id(c.peer_node_id()).await {
+                peers.push(p);
+            }
+        }
+
+        self.data.connected_peers = peers;
```
sdbondi: nit: Single allocation, clone peer manager once

```suggestion
        let peer_manager = self.wallet.comms.peer_manager();
        let mut peers = Vec::with_capacity(connections.len());
        for c in connections {
            let peer = peer_manager.find_by_node_id(c.peer_node_id()).await?;
            peers.push(peer);
        }

        self.data.connected_peers = peers;
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -195,6 +204,10 @@ impl AppState {
             None
         }
     }
+
+    pub fn get_connected_peers(&self) -> &Vec<Peer> {
```
sdbondi, line 208: nit: (thought this would be a 📎 , guess not)
```suggestion
    pub fn get_connected_peers(&self) -> &[Peer] {
```

applications/tari_console_wallet/src/ui/components/network_tab.rs
```diff
@@ -14,10 +14,41 @@ impl NetworkTab {
...
+        let list_areas = Layout::default()
+            .constraints([Constraint::Min(1)].as_ref())
+            .margin(1)
+            .split(area);
+
+        let peers = app_state.get_connected_peers();
+        let mut column0_items = Vec::new();
+        let mut column1_items = Vec::new();
+        let mut column2_items = Vec::new();
+        for p in peers.iter() {
```
sdbondi, line 35: nit:
```suggestion
        for p in peers {
```

applications/tari_console_wallet/src/ui/components/network_tab.rs
```diff
@@ -14,10 +14,41 @@ impl NetworkTab {
...
+
+        let list_areas = Layout::default()
+            .constraints([Constraint::Min(1)].as_ref())
+            .margin(1)
+            .split(area);
+
+        let peers = app_state.get_connected_peers();
+        let mut column0_items = Vec::new();
+        let mut column1_items = Vec::new();
+        let mut column2_items = Vec::new();
```
sdbondi: nit: single allocation each

```suggestion
        let mut column0_items = Vec::with_capacity(peers.len());
        let mut column1_items = Vec::with_capacity(peers.len());
        let mut column2_items = Vec::with_capacity(peers.len());
```

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -77,6 +80,26 @@ impl WalletEventMonitor {
...
+                                        ConnectivityEvent::PeerBanned(_) |
+                                        ConnectivityEvent::PeerOffline(_) |
+                                        ConnectivityEvent::PeerConnectionWillClose(_, _) => {
+                                            self.trigger_peer_state_refresh().await;
+                                        },
+                                        /// Only the above variants trigger state refresh
+                                        _ => (),
+                                     }
+                            },
+                            Err(e) => error!(target: LOG_TARGET, "Error reading from Connectivity event broadcast channel"),
```
sdbondi: Unused var `e`? (Is clippy even working?)

nit:
An error! saying "error reading" is not an ideal way to handle this case - the error from `Stream::next` on a tokio broadcast channel happens if 1) There are no senders (disconnected)(This cannot happen because that would be the end of the stream and this block would not execute), or 2) this loop could not read off the channel fast enough and some messages were dropped (lagged).
In this case, ignoring the error is the best option unless you can do something in response to missing some messages.

sdbondi (commented):
LGTM - some nits

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
sdbondi: So the only thing you can do is log that there is lagging happening - which could of course help make a slow select loop more noticeable in the logs. I'd make it debug and mention lagging, or not log it and make sure my loop isn't excessively slow.

applications/tari_console_wallet/src/ui/state/app_state.rs
philipr-za: Don't really see the benefit here? Makes it clear what collection is being stored inside and its returning a reference not a clone.

sdbondi: https://rust-lang.github.io/rust-clippy/master/#ptr_arg
 The caller doesnt care that it's a vec - it just needs a view into the data. It really doesn't matter - just mentioning that you should basically never do a `&Vec<_>` anywhere according to clippy style rules, I just follow them blindly :P

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -74,7 +77,27 @@ impl WalletEventMonitor {
...
+                                        ConnectivityEvent::PeerBanned(_) |
+                                        ConnectivityEvent::PeerOffline(_) |
+                                        ConnectivityEvent::PeerConnectionWillClose(_, _) => {
+                                            self.trigger_peer_state_refresh().await;
+                                        },
+                                        /// Only the above variants trigger state refresh
+                                        _ => (),
+                                     }
+                            },
+                            Err(e) => debug!(target: LOG_TARGET, "Lagging read on Connectivity event broadcast channel"),
```
sdbondi: Unused `e`

applications/tari_console_wallet/src/ui/state/app_state.rs
philipr-za: That lint is for arguments not returns. 🤷‍♂️ really don't see the benefit of hiding the fact that it's a vec.

sdbondi: There is no benefit either way - it's consistency - it doesnt matter

sdbondi (dismissed):
LGTM

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -367,13 +380,33 @@ impl AppStateInner {
...
+
     pub async fn get_shutdown_signal(&self) -> ShutdownSignal {
         self.wallet.comms.shutdown_signal()
     }
 
     pub async fn get_transaction_service_event_stream(&self) -> Fuse<TransactionEventReceiver> {
         self.wallet.transaction_service.get_event_stream_fused()
     }
+
+    pub async fn get_connectivity_event_stream(&self) -> Fuse<ConnectivityEventRx> {
```
sdbondi: async not needed
```suggestion
    pub fn get_connectivity_event_stream(&self) -> Fuse<ConnectivityEventRx> {
```

sdbondi (approved):

PR #2354, opened 2020-10-19
Fix transaction service bug with tx send/receive cancellation

philipr-za:
## Description
Fix a bug where the result of the cancellation oneshot in the send and receive protocols was not checked if it was Ok(()). This meant that on a shutdown when the sender side of that oneshot channel is dropped the receiver side resolves to a Err(_) and triggered an attempt to cancel the transaction. Generally the database was already broken down during the shutdown but depending on when stuff is dropped this could mistakenly cancel a pending transaction on wallet shutdown.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
@@ -333,12 +333,14 @@ where TBackend: TransactionBackend + 'static
                             break;
                         }
                     },
-                     _ = cancellation_receiver => {
-                        info!(target: LOG_TARGET, "Cancelling Transaction Receive Protocol for TxId: {}", self.id);
-                        return Err(TransactionServiceProtocolError::new(
-                            self.id,
-                            TransactionServiceError::TransactionCancelled,
-                        ));
+                    result = cancellation_receiver => {
+                        if result.is_ok() {
+                            info!(target: LOG_TARGET, "Cancelling Transaction Receive Protocol for TxId: {}", self.id);
+                            return Err(TransactionServiceProtocolError::new(
+                                self.id,
+                                TransactionServiceError::TransactionCancelled,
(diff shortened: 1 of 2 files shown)


delta1:
Nice catch - is there any test we could add for this?

delta1 (approved):

philipr-za:
> Nice catch - is there any test we could add for this?

Couldn't think of a way to test it as the bug occured during shutdown which breaks down all the interfaces you could use to do the test with.

delta1:
Yeah I figured it was difficult to test this one! Well spotted.

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
```diff
@@ -333,12 +333,14 @@ where TBackend: TransactionBackend + 'static
                             break;
                         }
                     },
-                     _ = cancellation_receiver => {
-                        info!(target: LOG_TARGET, "Cancelling Transaction Receive Protocol for TxId: {}", self.id);
-                        return Err(TransactionServiceProtocolError::new(
-                            self.id,
-                            TransactionServiceError::TransactionCancelled,
-                        ));
+                    result = cancellation_receiver => {
```
stringhandler, line 336: We should probably log the result via `warn!` if it's not `ok()`.

stringhandler (approved):

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
philipr-za: This can only happen if the other end of the channel is shutdown in which case all the channels in this select! will do the same and didn't feel like it needs to be logged as it will happen only in shutdown or else all the channels will show the same error (which are logged).

PR #2356, opened 2020-10-20
Combine validation code to use same function in pruned and archive mode.

SWvheerden:
## Description

Combine the pruned and archival validation function. 

## Motivation and Context

Currently pruned mode and archival mode currently uses 2 different function. Pruned mode header validation does not do monero block validation at all. It will do pow, but not checking the hashes etc. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/validators/headers.rs
+++ b/base_layer/core/src/base_node/validators/headers.rs
@@ -21,14 +21,17 @@
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::{
-    blocks::{BlockHeader, BlockHeaderValidationError},
+    blocks::BlockHeader,
     chain_storage::{BlockchainBackend, BlockchainDatabase},
     consensus::ConsensusManager,
-    proof_of_work::{get_median_timestamp, get_target_difficulty, Difficulty, PowError},
-    validation::{Validation, ValidationError},
+    validation::{
+        helpers::{check_achieved_and_target_difficulty, check_median_timestamp},
+        Validation,
+        ValidationError,
+    },
 };
 use log::*;
-use tari_crypto::tari_utilities::{epoch_time::EpochTime, hex::Hex, Hashable};
+use tari_crypto::tari_utilities::{hex::Hex, Hashable};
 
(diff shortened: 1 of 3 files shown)


sdbondi (dismissed):
LGTM - just need to fix clippy 📎

base_layer/core/src/base_node/validators/headers.rs
```diff
@@ -74,169 +76,19 @@ impl<B: BlockchainBackend> HeaderValidator<B> {
...
-
-        trace!(
-            target: LOG_TARGET,
-            "fetch_target_difficulties: #returned = {}",
-            target_difficulties.len()
-        );
-        Ok(target_difficulties.into_iter().rev().collect())
+        // We cant use the actual tip, as this is used by the header sync, the tip has not yet been downloaded, but we
+        // can assume that the previous header was added, so we use that as the tip.
+        let tip_height = block_header.height.saturating_sub(1);
```
sdbondi, line 81: The tip height passed in is incorrect. I suggest using the tip header.
```suggestion
             let tip_height = db.fetch_last_header()?.height;
```
From the usage in block validators:
```rust
     let tip_height = db.fetch_chain_metadata()?.height_of_longest_chain();
        check_median_timestamp(db, header, tip_height, self.rules.clone())?;
        trace!(
            target: LOG_TARGET,
            "BlockHeader validation: Median timestamp is ok for {} ",
            &header_id
        );
        check_achieved_and_target_difficulty(db, header, tip_height, self.rules.clone())?;
```

sdbondi (commented):
The height passed in should be the tip height

base_layer/core/src/base_node/validators/headers.rs
stringhandler: I think this is actually fine

stringhandler (approved):

base_layer/core/src/base_node/validators/headers.rs
SWvheerden: This should be fine. Header sync is cant ask for tip header, as the meta data is only set after sync, to this will stay at 0. We need to ask the previous header and work on the assumption that it was added. If it was not added it should have timed out

PR #2357, opened 2020-10-20
Return errors from inbound base node handler

stringhandler:
Instead of swallowing errors, return them and log them

## Motivation and Context

I saw a log on a node trying to sync with my node that timed out. I looked in the logs at that time and there was nothing. Looking at the code, there are a number of cases where the error is not logged if there is one encountered.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


sdbondi (approved):
Definitely needed - untested but will work pending passing tests (handle_request already returns an error result to the caller handle)

PR #2359, opened 2020-10-20
Set port + 1 for base node wallet in TCP mode

sdbondi:
This "temporary fix" only applies to TCP and SOCKS transport mode and
was refactored out at some point. This PR re-adds it (but the code is more explicit about what it's doing).


stringhandler (approved):
Looks good. Will test

PR #2306, opened 2020-10-06
DHT RPC get_peers returns all peers and "on connect" mode for NetworkDiscovery

sdbondi:
## Description

DHT RPC changes:
- `get_peers` now returns all peers
- `get_closest_peers` returns `n` peers closer to the given node id
- Added `include_clients` bool flag to indicate if client nodes should
  be returned
- New / updated unit tests

Network discovery changes:
- Added "on connect" mode. Initially, network discovery is "active" insofar
  as it actively connects to and seeks peers closer to the node's
  neighbourhood.  Once no more neighbouring nodes can be found,
  the network discovery enters "on connect" mode which syncs whenever a
  (non-client) peer connects.
- `DiscoveryReady` emits `OnConnectMode` event if no more peers are left
  to sync

Comms RPC changes:
- Changed default RPC client deadline to 30s from 100s

This change is backward compatible with previous `t/dht/1` clients. 

## Motivation and Context

This strategy is much more aggressive than the last iteration and allows peer lists to be synced more effectively.

## How Has This Been Tested?

Unit tests added/updated. Tested on base node with only seed peers 

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [x] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #2361, opened 2020-10-21
Remove DhtJoin reply and only send join when node comes online

sdbondi:
## Description

- Node will no longer reply to join messages as this method is
  deprecated in favour of #2306. Until #2306 is released to mobile,
  mobile wallets will fail to discover neighbouring nodes that include
  this PR.
- DHT connectivity actor only sends a join when online. Join is now more
  like an "announce" that optionally announces to the neighbourhood of the node that
  the node is online. `Join` could be renamed to `Announce` in future. 
- Only publish ConnectivityDegraded event only if the number of peers has changed
- Updated _memorynet_

## Motivation and Context

The primary suspect for why join messages are flooding is:
- Network discovery finds many new peers/neighbours, then
- Connectivity manager is notified and the new peers are added as "managed peers", then
- Each time this happens the connectivity status event (online, degraded) is published,
- This causes a Join message to be sent (as the node is now has a connection)

However there is a protection in place for this:
If a join was sent in the last 10 minutes, don't send another UNLESS there is only a single peer connected.
A node can only connect to a single peer + network discovery is doing its work and finding new peers = flooding.

It is currently not clear why the node may be in this state, however removing bug surface is the best approach.

## How Has This Been Tested?

Existing tests updated where necessary. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/p2p/src/initialization.rs
+++ b/base_layer/p2p/src/initialization.rs
@@ -341,7 +341,7 @@ where
         .node_identity()
         .has_peer_features(PeerFeatures::COMMUNICATION_NODE)
     {
-        comms = comms.add_rpc(RpcServer::new().add_service(dht.rpc_service()));
+        comms = comms.add_rpc_server(RpcServer::new().add_service(dht.rpc_service()));
     }
 
     // Hook up DHT messaging middlewares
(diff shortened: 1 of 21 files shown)


stringhandler (approved):
Looked in detail and I think it will work. Difficult to verify though

PR #2364, opened 2020-10-21
Revert change to fetch_mmr_nodes from #2357

delta1:
## Description
The change to `fetch_mmr_nodes` introduced in #2357 caused the `request_and_response_fetch_mmr_node_and_count` test to timeout and fail. This commit reverts that change, and makes the previous `debug` log into `warn` for greater visibility.

## Motivation and Context

This really depends on desired behaviour - the test indicates that an empty set should be returned if a peer requests "out of bounds" MMR nodes instead of an error result. The previous code comment also indicates _"another node might ask for mmr nodes we dont have, so we return ok([])"_.

Either this specific change introduced in #2357 needs to be reverted to maintain the existing invariant, or the test itself needs to be changed. I am happy to refactor the test if that's the correct response instead, please let me know.

## How Has This Been Tested?

cargo test

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -403,14 +403,27 @@ where T: BlockchainBackend + 'static
             NodeCommsRequest::FetchMmrNodes(tree, pos, count, hist_height) => {
                 let mut added = Vec::<Vec<u8>>::with_capacity(*count as usize);
                 let mut deleted = Bitmap::create();
-                let mmr_nodes =
-                    async_db::fetch_mmr_nodes(self.blockchain_db.clone(), *tree, *pos, *count, Some(*hist_height))
-                        .await?;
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -403,14 +403,27 @@ where T: BlockchainBackend + 'static
...
+                    // return ok([])
+                    Err(e) => warn!(
+                        target: LOG_TARGET,
+                        "Could not provide requested mmr nodes (pos:{},count:{}) to peer because: {}",
+                        pos,
+                        count,
+                        e.to_string()
+                    ),
                 }
                 deleted.run_optimize();
```
stringhandler, line 428: Looks like this should also be reverted?

delta1: I think that was common to both versions

stringhandler (approved):

PR #2369, opened 2020-10-22
Increase lagging broadcast channels' capacities

hansieodendaal:
## Description
During adverse conditions, it was found that the state events and transaction service broadcast channel capacities were inadequate to handle the load. Subscribers lagged up to 92 state events (capacity of 10) and 490 transaction events (capacity of 200). As a rule of thumb, the capacities were increased sufficiently to accommodate 5 to 10 times more events than the peak amount they lagged by. The tokio broadcast channels are also self-regulating i.t.o. memory usage.

```
2020-10-21 19:25:57.722798300 [c::bn::state_machine_service::states::listening]
    DEBUG Metadata event subscriber lagged by 92 item(s)
```
```
2020-10-21 22:44:13.858514800 [base_node::app::utils] WARN  Error reading from
    event broadcast channel Lagged(490)
```

**Edit**: Changes to the transaction service broadcast channel reverted.

## Motivation and Context
Subscribers to state events and transaction service events should never miss any messages.

## How Has This Been Tested?
Tested in 3 base nodes on Windows.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/initializer.rs
+++ b/base_layer/core/src/base_node/state_machine_service/initializer.rs
@@ -84,7 +84,7 @@ where B: BlockchainBackend + 'static
 
     fn initialize(&mut self, context: ServiceInitializerContext) -> Self::Future {
         trace!(target: LOG_TARGET, "init of base_node");
-        let (state_event_publisher, _) = broadcast::channel(10);
+        let (state_event_publisher, _) = broadcast::channel(500);
         let (status_event_sender, status_event_receiver) = watch::channel(StatusInfo::new());
 
         let handle = StateMachineHandle::new(


base_layer/core/src/base_node/state_machine_service/initializer.rs
```diff
@@ -81,7 +81,7 @@ where B: BlockchainBackend + 'static
 
     fn initialize(&mut self, context: ServiceInitializerContext) -> Self::Future {
         trace!(target: LOG_TARGET, "init of base_node");
-        let (state_event_publisher, _) = broadcast::channel(10);
+        let (state_event_publisher, _) = broadcast::channel(500);
```
philipr-za, line 87: This is not a huge change so I think its fine.

base_layer/wallet/src/transaction_service/mod.rs
```diff
@@ -192,7 +192,7 @@ where T: TransactionBackend + 'static
         let base_node_response_stream = self.base_node_response_stream();
         let transaction_cancelled_stream = self.transaction_cancelled_stream();
 
-        let (publisher, _) = broadcast::channel(200);
+        let (publisher, _) = broadcast::channel(10000);
```
philipr-za: This is a huge change. Bear in mind that this is impacting directly on the memory footprint of the app. I suspect this lagging is an artifact of having the wallet and base node running on the same Tokio runtime and won't be an issue when we split out the wallet. I wouldn't want to forget about this huge buffer then.

base_layer/core/src/base_node/state_machine_service/initializer.rs
hansieodendaal: Yes, still seing a lot of `[c::bn::state_machine_service::states::listening] DEBUG Metadata event subscriber lagged` in my testing.

base_layer/wallet/src/transaction_service/mod.rs
hansieodendaal: I think you are correct in this. I have not seen these subscribers lagging again after I picked it up initially, so I will revert this change.

base_layer/core/src/base_node/state_machine_service/initializer.rs
stringhandler: Sorry I disagree. If the base node has changed state 10 times and there is a part of the system that has not read the event, it is probably a sign that something is broken. I think leave it as 10

hansieodendaal: My argument is that these events should not be missed, as some parts of the system depends on all the state events. They could be missed for various reasons, one being if the base node has to defend against many spam messages, as happended just recently with join messages. It is much better to be resilient in this regard, as opposed to some part of the system not working properly. The default value of 10 was just a guess at the time, and now we have seen that it can sometimes be inadequate. As the tokio broadcast channels are also self-regulating i.t.o. memory usage, its use will be optimal.

hansieodendaal: The `ridcully` testnet also still has `[c::bn::state_machine_service::states::listening] DEBUG Metadata event subscriber lagged` log entries.

stringhandler (approved):

PR #2371, opened 2020-10-22
Add configurable BN service request timeouts

hansieodendaal:
## Description
- This change adds configurable base node service request timeouts, notably differentiating between general service requests (default at 180s), requests to fetch blocks for block sync (default at 30s) and requests for a complete current UTXO set included in the blockchain (default at 600s).
- Removed a mempool service request timeout setting in the sample config files that was not used.

## Motivation and Context
Different aspects of the base node operation require different timeouts to work optimally.

## How Has This Been Tested?

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/base_node/consts.rs
```diff
@@ -23,6 +23,12 @@
 use std::time::Duration;
 
 /// The allocated waiting time for a request waiting for service responses from remote base nodes.
-pub const BASE_NODE_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(180);
+pub const GENERAL_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(180);
```
stringhandler: These can be removed since they are defaulted in the config

base_layer/core/src/base_node/consts.rs
```diff
@@ -23,6 +23,12 @@
...
+/// The allocated waiting time for a fetch blocks request waiting for service responses from remote base nodes.
+pub const FETCH_BLOCKS_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+/// The allocated waiting time for a fetch UTXOs request waiting for service responses from remote base nodes.
+pub const FETCH_UTXOS_SERVICE_REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
+/// The minimum allocated waiting time for any request waiting for service responses from remote base nodes.
+pub const SERVICE_REQUEST_MINIMUM_TIMEOUT: Duration = Duration::from_secs(10);
 /// The fraction of responses that need to be received for a corresponding service request to be finalize.
 pub const BASE_NODE_SERVICE_DESIRED_RESPONSE_FRACTION: f32 = 0.6;
```
stringhandler: I think this should also be removed and added to the config file with a default

stringhandler (commented):
I like this. I think let's remove the constants entirely and just use the defaults in the config file (or the actual values in the config file)

base_layer/core/src/base_node/consts.rs
hansieodendaal: Constants removed in favour of moving the values directly into the default struct method.

base_layer/core/src/base_node/consts.rs
hansieodendaal: Constants removed in favour of moving the values directly into the default struct method, however, `desired_response_fraction` bit added to the config file, as we do not have a requirement for that to be user-configurable at the moment.

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -110,13 +100,14 @@ impl Default for BlockSyncConfig {
     fn default() -> Self {
         Self {
             sync_strategy: BlockSyncStrategy::ViaBestChainMetadata(BestChainMetadataBlockSync),
-            max_metadata_request_retry_attempts: MAX_METADATA_REQUEST_RETRY_ATTEMPTS,
-            max_header_request_retry_attempts: MAX_HEADER_REQUEST_RETRY_ATTEMPTS,
-            max_block_request_retry_attempts: MAX_BLOCK_REQUEST_RETRY_ATTEMPTS,
-            max_add_block_retry_attempts: MAX_ADD_BLOCK_RETRY_ATTEMPTS,
-            header_request_size: HEADER_REQUEST_SIZE,
-            block_request_size: BLOCK_REQUEST_SIZE,
+            max_metadata_request_retry_attempts: 3,
```
stringhandler, line 103: I don't think this is an improvement. Either leave the constants in or add a config setting for them. Perhaps a `default()` method doesn't make sense and all the values should be provided from config?

stringhandler (commented):
Happy with it in general, but I think the removal of unrelated constants (e.g. `MAX_METADATA_REQUEST_RETRY_ATTEMPTS`) without adding config values for them is a step backwards. If you revert those or add config values for them, I'm happy to merge

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
hansieodendaal: This default method was out of sorts with how others were done. Default values are generally passed directly, not via single use local constants, to improve readability it seems.
I do think we need the dafault method in this case. Generally the config settings overrided these where applicable.
I will revert to using the local constants here if you still think that is peferrable?

hansieodendaal:
@mikethetike, I have tested out the 30s default timeout to fetch blocks; this is not adequate for our current block sync strategy. Three out of eight nodes started banning sync peers because blocks were not returned in 30s, and could not sync to the tip anymore. I propose we change this default setting to 60s or 120s, do you agree?
```
2020-10-28 09:55:22.522676500 [c::bn::base_node_service::service] WARN  Request (request key 11880102790285126520) timed out after 30007ms
2020-10-28 09:55:22.522715000 [c::bn::base_node_service::service] ERROR Failed to send outbound request (request key: 11880102790285126520): RequestTimedOut
2020-10-28 09:55:22.522747100 [c::bn::state_machine_service::states::block_sync] DEBUG Failed to fetch blocks from peer: RequestTimedOut. Retrying.
2020-10-28 09:55:22.522755300 [c::bn::state_machine_service::states::block_sync] INFO  Banning peer Node ID: 1cc148eeaa18107edfe665a06b
```

**Edit:**
_It can take 30s for one block. So should be 5*30._
Changed to 150

base_layer/core/src/base_node/service/service.rs
```diff
@@ -71,17 +71,23 @@ const LOG_TARGET: &str = "c::bn::base_node_service::service";
...
     pub desired_response_fraction: f32,
 }
 
 impl Default for BaseNodeServiceConfig {
     fn default() -> Self {
         Self {
-            request_timeout: BASE_NODE_SERVICE_REQUEST_TIMEOUT,
-            desired_response_fraction: BASE_NODE_SERVICE_DESIRED_RESPONSE_FRACTION,
+            service_request_timeout: Duration::from_secs(180),
+            fetch_blocks_timeout: Duration::from_secs(30),
```
hansieodendaal:
```suggestion
            fetch_blocks_timeout: Duration::from_secs(150),
```

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -110,13 +100,14 @@ impl Default for BlockSyncConfig {
...
+            max_header_request_retry_attempts: 5,
+            max_block_request_retry_attempts: 5,
+            max_add_block_retry_attempts: 3,
+            header_request_size: 100,
+            block_request_size: 5,
             orphan_db_clean_out_threshold: 0,
+            fetch_blocks_timeout: Duration::from_secs(30),
```
hansieodendaal:
```suggestion
            fetch_blocks_timeout: Duration::from_secs(150),
```

common/config/presets/windows.toml
```diff
@@ -52,6 +52,15 @@
 # - Rate limit for the base node wallet (min value = 5, default value = 20).
 #buffer_rate_limit_base_node_wallet = 20
 
+# The timeout (s) for requesting blocks from a peer during blockchain sync (min value = 10 s, default value = 30 s).
```
hansieodendaal:
```suggestion
# The timeout (s) for requesting blocks from a peer during blockchain sync (min value = 10 s, default value = 150 s).
```

common/config/presets/windows.toml
```diff
@@ -52,6 +52,15 @@
...
+#fetch_blocks_timeout = 30
```
hansieodendaal:
```suggestion
#fetch_blocks_timeout = 150
```

common/src/configuration/utils.rs
```diff
@@ -63,6 +63,9 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
     cfg.set_default("common.buffer_rate_limit_base_node", 1_000).unwrap();
     cfg.set_default("common.buffer_rate_limit_base_node_wallet", 1_000)
         .unwrap();
+    cfg.set_default("common.fetch_blocks_timeout", 30).unwrap();
```
hansieodendaal:
```suggestion
    cfg.set_default("common.fetch_blocks_timeout", 150).unwrap();
```

hansieodendaal (commented):
~~If agreed, please commit suggestions before merge. Thanks~~
Will rebase

stringhandler (approved):

PR #2372, opened 2020-10-22
Increase short term ban duration

stringhandler:
## Description

Increase the ban time for short banning of peers

Also reduce log level of message when peer can't be connected

## Motivation and Context

Previously on a base node request timeout, a peer would be banned for 24 hours. This was excessively long, so it was reduced to the short banning period. This however is now shorter than the timeout itself (originally 10 mins, now 3 minutes), which means that if two peers timed out, the original peer would be unbanned again. This results in the same peers being tried over and over again.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


stringhandler:
Alternatively we could revert back to banning the peers for 24 hours on timeout, but that seems excessive

delta1 (approved):
LGTM

sdbondi (approved):

PR #2399, opened 2020-10-28
INFO log messages must be on a single line

CjS77:
Logging policy is that DEBUG and higher messages fit on one line.
Printing the entire block must be trace level.


stringhandler (dismissed):

delta1 (approved):
LGTM

PR #2403, opened 2020-10-28
Add exclusive file locks to Wallet, Chain and Peer db’s

philipr-za:
## Description

This PR adds an exclusive file locking mechanism to the wallet databases, the chain storage LMDB database and the Peer database. This is done to prevent two processes from simultaneously accessing this databases which can cause data corruption. 

This was done using the fs2 crate which contains a cross platform method for acquiring an OS level write lock on a file. For all these database the first step in initialization is to attempt to acquire a firelock on the lock file in the path where the database is located. If the file is not locked then a lock is acquired and the db initialisation proceeds as normal. If the file cannot be locked because the lock is already held by another instance an error is thrown. If the process holding the lock panics the OS level lock is lost so it is not needed to clean up the file explicitly for this method to work.

During testing this mechanism revealed that the wallet services were not being explicitly shutdown properly, we also plumb the shutdown signals into the services and their sub protocols.

The `test_store_and_forward_send_tx` test was being particularly flakey in the PR so I have opted to ignore it for now and come back and examine it properly in the future.

## How Has This Been Tested?
Tests were written for the wallet and chain storage db’s but it was not trivial to test the peer_manager addition as it is located in the p2p initialiser.

The code was manually tested on:
Mac OS X
Windows 10
Ubuntu 20.01
Android

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1196,6 +1196,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
+name = "fs2"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "640001e1bd865c7c32806292822445af576a6866175b5225aa2087ca5e3de551"
+dependencies = [
+ "kernel32-sys",
+ "libc",
+ "winapi 0.2.8",
+]
+
+[[package]]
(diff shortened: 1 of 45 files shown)


Jasonvdb:
Can confirm it works on iOS sim and device as well. 
Creates `.tari_wallet.sqlite3.lock` and `.p2p_file.lock`

delta1 (approved):
Tested on MacOS. LGTM.

stringhandler (approved):
Nice! This should help a lot

PR #2413, opened 2020-11-02
Validate previous header when calculating MMR roots

stringhandler:
## Description

When creating a new block, check that the node is at the tip and return an error if the block template is out of date

## Motivation and Context

I believe what is happening is that a node asks for a block template, then adds it's coin base, then the base node progresses to a new height or reorgs. Then when the base node asks for the mmr roots to be applied, it applies them without checking the previous header height.

## How Has This Been Tested?

Ran the integration tests manually

## Types of changes

> Possibly
* [ ] New Tests
* [ ] Documentation

## Checklist:


hansieodendaal:
1. `sync from scratch` did not work due to different unrelated error:
```
[c::bn::state_machine_service::states::block_sync] WARN  Banning all peers from local node, because they 
could not provide a valid chain link
```
2. sync from ~1000 blocks behind worked fine

**Note:** I think test case 1 should also pass when that problem is fixed.

hansieodendaal (approved):
LGTM, both sync test cases succeeded. Just some clippy warnings

PR #2415, opened 2020-11-03
Prevent loop in peer sync by storing all peers attempted

stringhandler:
## Motivation and Context

Using 3 base nodes, I found that the one node would hit 100% cpu. This was caused by it trying to sync peers with one peer and then the other peer in an infinite loop. The attempt here is to store all peers attempted in a single place. 

I think a better solution in future would be to store the last_peer_sync_succeeded time in the peer database so that it isn't retried again.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/network_discovery/ready.rs
+++ b/comms/dht/src/network_discovery/ready.rs
@@ -27,7 +27,7 @@ use super::{
 use crate::{network_discovery::DhtNetworkDiscoveryRoundInfo, DhtConfig};
 use log::*;
 use std::cmp;
-use tari_comms::peer_manager::{NodeId, PeerFeatures};
+use tari_comms::peer_manager::PeerFeatures;
 
 const LOG_TARGET: &str = "comms::dht::network_discovery::ready";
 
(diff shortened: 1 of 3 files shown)


hansieodendaal:
Running two tests: sync from scratch and sync from ~1000 blocks behind, both without pruning.

hansieodendaal:
**Question:** With selecting random sync peers from `all_attempted_peers` instead of `previous_sync_peers`, will it not result in random peers being selected for sync that is not available?

stringhandler:
The `all_attempted_peers` are passed into the `exclude_peers` parameter, so they are excluded and not tried again

hansieodendaal:
1. `sync from scratch` did not work due to different unrelated error:
```
[c::bn::state_machine_service::states::block_sync] WARN  Banning all peers from local node, because they 
could not provide a valid chain link
```
2. sync from ~1000 blocks behind worked fine

**Note:** I think test case 1 should also pass when that problem is fixed.

hansieodendaal (approved):
LGTM, both sync test cases succeeded.

PR #2420, opened 2020-11-04
[console wallet] show base node chain tip and sync status

delta1:
## Description

Replaces the placeholder data with live data from the base node service. Adds a configuration option to specify which base node the wallet should connect to, and then displays the chain tip and sync status.

## Motivation and Context

The console wallet is still using some placeholder information

## How Has This Been Tested?

Manually -  you can use the following setting to point the wallet to `tbn-oregon` base node in `config.toml`

```toml
[wallet]
base_node_service_peer = "e856839057aac496b9e25f10821116d02b58f20129e9b9ba681b830568e47c4d::/onion3/exe2zgehnw3tvrbef3ep6taiacr6sdyeb54be2s25fpru357r4skhtad:18141"
```

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/main.rs
+++ b/applications/tari_console_wallet/src/main.rs
@@ -7,7 +7,7 @@
 #![deny(unknown_lints)]
 #![recursion_limit = "512"]
 use log::*;
-use rand::{rngs::OsRng, RngCore};
+use rand::{rngs::OsRng, seq::SliceRandom};
 use std::{fs, sync::Arc};
 use structopt::StructOpt;
 use tari_app_utilities::{
@@ -15,12 +15,16 @@ use tari_app_utilities::{
     utilities::{parse_peer_seeds, setup_wallet_transport_type, ExitCodes},
 };
 use tari_common::{configuration::bootstrap::ApplicationType, ConfigBootstrap, GlobalConfig, Network};
-use tari_comms::{peer_manager::PeerFeatures, NodeIdentity};
+use tari_comms::{
+    peer_manager::{Peer, PeerFeatures},
+    NodeIdentity,
+};
 use tari_comms_dht::{DbConnectionUrl, DhtConfig};
 use tari_core::{consensus::Network as NetworkType, transactions::types::CryptoFactories};
 use tari_p2p::initialization::CommsConfig;
 use tari_shutdown::{Shutdown, ShutdownSignal};
 use tari_wallet::{
+    base_node_service::config::BaseNodeServiceConfig,
     error::WalletError,
     storage::sqlite_utilities::initialize_sqlite_database_backends,
     transaction_service::config::TransactionServiceConfig,
@@ -109,7 +113,14 @@ fn main_inner() -> Result<(), ExitCodes> {
         return Ok(());
     }
 
-    let wallet = runtime.block_on(setup_wallet(&config, node_identity, shutdown.to_signal()))?;
+    let base_node = get_base_node_peer(&config)?;
+
+    let wallet = runtime.block_on(setup_wallet(
+        &config,
+        node_identity,
+        base_node.clone(),
+        shutdown.to_signal(),
+    ))?;
 
     debug!(target: LOG_TARGET, "Starting app");
 
@@ -117,7 +128,7 @@ fn main_inner() -> Result<(), ExitCodes> {
 
     let handle = runtime.handle().clone();
     let result = match wallet_mode(bootstrap) {
-        WalletMode::Tui => tui_mode(handle, config, node_identity, wallet.clone()),
+        WalletMode::Tui => tui_mode(handle, config, node_identity, wallet.clone(), base_node),
         WalletMode::Grpc => grpc_mode(handle, wallet.clone(), config),
         WalletMode::Script(path) => script_mode(handle, path, wallet.clone(), config),
         WalletMode::Command(command) => command_mode(handle, command, wallet.clone(), config),
@@ -135,6 +146,32 @@ fn main_inner() -> Result<(), ExitCodes> {
     println!("Done.");
 
     result
+}
+
+fn get_base_node_peer(config: &GlobalConfig) -> Result<Peer, ExitCodes> {
+    // pick a random peer from peer seeds config
+    // todo: strategy for picking peer seed: random or lowest latency
+    let peer_seeds = parse_peer_seeds(&config.peer_seeds);
+    let peer_seed = peer_seeds.choose(&mut OsRng);
+
+    // get the configured base node service peer
+    let base_node_peers = parse_peer_seeds(&[config.wallet_base_node_service_peer.clone()]);
+    let base_node = base_node_peers.first();
+
+    let peer = match (peer_seed, base_node) {
+        // base node service peer is defined, so use that
+        (_, Some(node)) => node.clone(),
+        // only peer seeds were provided in config
+        (Some(seed), None) => seed.clone(),
+        // invalid configuration
+        _ => {
+            return Err(ExitCodes::ConfigError(
+                "No peer seeds or base node peer defined in config!".to_string(),
+            ))
+        },
+    };
+
+    Ok(peer)
 }
 
 fn wallet_mode(bootstrap: ConfigBootstrap) -> WalletMode {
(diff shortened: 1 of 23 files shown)


base_layer/wallet/src/base_node_service/config.rs
```diff
@@ -34,8 +34,8 @@ pub struct BaseNodeServiceConfig {
 impl Default for BaseNodeServiceConfig {
     fn default() -> Self {
         Self {
-            refresh_interval: Duration::from_secs(30),
-            request_keys_max_age: Duration::from_secs(120),
+            refresh_interval: Duration::from_secs(10),
+            request_keys_max_age: Duration::from_secs(60),
```
delta1: this should also come from config

applications/tari_console_wallet/src/main.rs
```diff
@@ -234,24 +233,24 @@ async fn setup_wallet(
...
-                seed_peers[random_seed_peer]
-                    .addresses
-                    .first()
-                    .expect("The seed peers should have an address")
-                    .to_string(),
-            )
-            .await
-            .map_err(|e| ExitCodes::WalletError(format!("Error setting wallet base node peer. {}", e)))?;
-    }
+    debug!(target: LOG_TARGET, "Setting base node peer");
```
philipr-za, line 282: Are we sure that if no base node is set that the wallet will error out on startup? A wallet can run and interact with other wallets without speaking to base nodes. It is not ideal but I am not sure its a full on error show stopper? Thoughts?

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -145,24 +142,42 @@ impl<B: Backend> App<B> {
 
         self.tabs.draw_titles(f, title_halves[0]);
 
-        let chain_meta_data = match get_dummy_base_node_status() {
+        let base_node_state = self.app_state.get_base_node_state();
```
philipr-za: This stuff should be moved into a proper self contained ui component that implements the `Component` trait and the drawing code should live in there I believe.

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -79,22 +85,37 @@ impl WalletEventMonitor {
...
+                    result = base_node_events.select_next_some() => {
+                        match result {
+                            Ok(msg) => {
+                                trace!(target: LOG_TARGET, "Wallet Event Monitor received base node event {:?}", msg);
+                                match (*msg).clone() {
+                                    BaseNodeEvent::BaseNodeState(state) => {
+                                        debug!(target: LOG_TARGET, "new base node state: {:?}", state);
+                                        self.trigger_base_node_state_refresh(state).await;
+                                    },
+                                    _ => warn!(target: LOG_TARGET, "Received an unhandled BaseNodeEvent"),
```
philipr-za: I don't think you log this, especially as a warn. This monitor just doesn't care about the other variants, its not an error if it receives them. See the comment in the match above for the variants of the ConnectivityEvent we don't care about.

philipr-za: In fact is this branch even possible on this enum? If this wasn't in the select! macro I believe clippy would have something to say.

applications/tari_console_wallet/src/main.rs
delta1: Good point, probably best to separate setting peer seeds from the base node peer? 

In that case, if no base node is set then default to one of the seeds?

applications/tari_console_wallet/src/ui/app.rs
delta1: Cool will do

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
delta1: Gotcha will do

applications/tari_console_wallet/src/main.rs
philipr-za: I think this is a point to discuss a bit further. I like this proposed strategy, I think it is the most plug and play path to the wallet just working out the box.

However based on our discussion yesterday we might not want to default to a seed peer if we feel that users will want to be careful about which base node they connect to we might only want to connect to explicitly chosen base nodes? I think perhaps that these will be a small percentage of power users,

delta1: For now, I'm going to implement the fallback to a random seed peer for ease of use, with logging info and documentation. Privacy-oriented users will of course be able to override the setting in config. We can then iterate on this later if necessary. 

Thanks for the feedback @philipr-za !

applications/tari_console_wallet/src/main.rs
```diff
@@ -137,6 +149,36 @@ fn main_inner() -> Result<(), ExitCodes> {
     result
 }
 
+fn get_peers(config: &GlobalConfig) -> Result<(Peer, Peer), ExitCodes> {
```
philipr-za: I don't think we should have 2 answers here? The base node we monitor should also be the base node we are using to broadcast TX and validate UTXOs

applications/tari_console_wallet/src/main.rs
```diff
@@ -234,24 +284,25 @@ async fn setup_wallet(
...
-                    .first()
-                    .expect("The seed peers should have an address")
-                    .to_string(),
-            )
-            .await
-            .map_err(|e| ExitCodes::WalletError(format!("Error setting wallet base node peer. {}", e)))?;
-    }
+    debug!(target: LOG_TARGET, "Setting base node service peer");
+    wallet
+        .set_base_node_service_peer(base_node)
```
philipr-za: As above I think this and the peer_seed below should be the same.

applications/tari_console_wallet/src/ui/components/network_tab.rs
```diff
@@ -45,48 +46,49 @@ impl NetworkTab {
...
+    where B: Backend {
+        let block = Block::default().borders(Borders::ALL).title(Span::styled(
+            "Base Node Peer",
+            Style::default().fg(Color::White).add_modifier(Modifier::BOLD),
+        ));
+        f.render_widget(block, area);
+
+        let base_node = app_state.get_base_node_peer();
+        let public_key = base_node.public_key.to_hex();
+        let public_address = base_node.addresses.first().unwrap().to_string();
```
philipr-za: This unwrap is risky, I would handle the case that for some reason there isn't a first address here and just default to a blank string rather than panic using unwrap_or_default(...)

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -79,22 +85,36 @@ impl WalletEventMonitor {
...
                             Err(_) => debug!(target: LOG_TARGET, "Lagging read on Connectivity event broadcast channel"),
                         }
                     },
+                    result = base_node_events.select_next_some() => {
+                        match result {
+                            Ok(msg) => {
+                                trace!(target: LOG_TARGET, "Wallet Event Monitor received base node event {:?}", msg);
+                                match (*msg).clone() {
+                                    BaseNodeEvent::BaseNodeState(state) => {
+                                        debug!(target: LOG_TARGET, "base node state: {:?}", state);
```
philipr-za: I don't think this log is valuable, you do log it as a trace a couple lines above and I would argue these should be trace anyway so I would drop this one.

base_layer/wallet/src/wallet.rs
```diff
@@ -232,8 +248,23 @@ where
         self.comms.clone().wait_until_shutdown().await;
     }
 
-    /// This function will set the base_node that the wallet uses to broadcast transactions and monitor the blockchain
-    /// state
+    /// This function will set the base_node service peer that the wallet uses to monitor the blockchain state
+    pub async fn set_base_node_service_peer(&mut self, peer: Peer) -> Result<(), WalletError> {
```
philipr-za: I think this should be merged with the set_base_node_peer() function below.

philipr-za (commented):
Looking good, couple small comments.

applications/tari_console_wallet/src/main.rs
delta1: Yeah that makes sense, and I did have it that way previously. I must have mistaken previous discussion saying they should be separate. Anyway, I'll update it.

philipr-za (approved):
👍

Issue #2426, opened 2020-11-10
list-headers n lists n+1 headers


delta1:
**Describe the bug**
Using `list-headers 1` shows 2 headers

**To Reproduce**
1. run tari base node
2. run command `list-headers 1`
3. output is last 2 headers instead of last 1

**Expected behavior**
Running `list-headers [n]` should only show *n* headers 

**current output example**

```
>> list-headers 1

Header hash: 106fe98c85348eb3bd6ff1604e7ea4b9c1d2ba1fd20bf5e6cabac2870aa22b2b
Version: 1
Block height: 2673
...

Header hash: b121130a967a2401507219ad97f920b67ba27a028b664639e5b56d598e04e61e
Version: 1
Block height: 2672
...

```

**Desktop (please complete the following information):**
 - OS & Version: all OS versions, TBN 0.6

sdbondi:
I've assigned myself, since I had to fix this to check headers in the upcoming block sync PR. Will reference this issue when that is ready to PR (ref #2428)

delta1:
nice one @sdbondi 👍

PR #2436, opened 2020-11-17
Fast block sync using RPC streaming

### LINKED ISSUE #2428: Fast "header-first" block sync
Tracking issue for fast block sync. 

Fast block sync will attempt to establish the strongest PoW chain by checking where a local chain and a remote peer's chain are linked and downloading headers from that point to the tip. Once headers are validated and downloaded, an archival node proceeds to block sync where full block bodies are downloaded. 

- [x] Base node sync RPC service (PR #2348)
- [ ] Various small bug fixes / enhancements in base node sync RPC
- [ ] Database backend support for partial state
  - [ ] Support rewinding with or without bodies
  - [ ] Support adding/associating block bodies with existing headers
- [ ] Change state machine to complete header sync before block sync
- [ ] Validate headers efficiently by building and maintaining an in-memory cache of target difficulties and timestamps for the block window
- [ ] Validate block bodies against header MMRs
- [ ] Chain metadata must adjust as necessary to reflect which full blocks the node can provide
- [ ] Sync peers should be selected at random for header sync
- [ ] Allow "force sync from this peer" to be configured
- [ ] Block sync needs to locate a peer that can provide the required full blocks

### Related issues

#2113

sdbondi:
## Description

Fast full block sync that uses the base node sync RPC interface.

**State machine**

- Updated base node state machine to first use a header sync, followed
  by full block body only sync (header-first sync)

**HeaderSync**

- Header sync determines the strongest proof of work from the pool of
  connected peers by downloading a set of 1000 headers, and validating chain
  strength relative the the local chain as well as the headers
  themselves.
- Header sync validator loads an initial state from the db once and then
  updates the in-memory state as validation of subsequent headers proceeds.

**Block body sync**

- Only block bodies are downloaded
- Validate block bodies and fail atomically

**Sync RPC**
- Various bug fixes
- `sync_block` uses start and end hashes instead of (hash, count)

**Config changes**

- Allow users to optionally specify sync peers in config

**Misc**
- Removed the blockchain state service in favour of using the AsyncBlockchainDb directly
- Only enter sync mode for the chain metadata of whitelisted sync peers, if they are specified
- TODO: Unit tests of header and block body sync 

## Motivation and Context

Closes #2428 #2113 #2426 
Ref #2348 

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -0,0 +1,148 @@
...
+        );
+        self.state = Some(State {
+            timestamps,
+            target_difficulties,
+        });
+
+        Ok(())
+    }
+
+    pub fn validate(&mut self, header: &BlockHeader) -> Result<(), BlockHeaderSyncError> {
```
stringhandler, line 91: This alters the internal state. Perhaps a better name would be  `validate_and_store`. It also assumes that the data passed to it is relative to the previous header, and because it is a public method, a check should be added to ensure that the new header is connected to the last header, otherwise throw an error

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -0,0 +1,429 @@
...
+                target: LOG_TARGET,
+                "Validating header #{} (Pow: {})",
+                header.height,
+                header.pow_algo(),
+            );
+            self.header_validator.validate(header)?;
+            debug!(target: LOG_TARGET, "Header #{} is VALID", header.height,)
+        }
+
+        self.header_validator
```
stringhandler: Does this work in the  case where the  chain is on a  weaker fork, but the headers  returned  for the stronger fork  are not  stronger?

Example:
NUM_HEADERS_TO_REQUEST is 1 in this example for clarity. We are on  `GEN...H.I1...Q1`, syncing to a  chain  on `GEN......H.I2...Q2...Z`.  We send our headers and receive `H` and `I2` from the peer node `I2` is the last header  that the sync peer  returns. `I2` is unlikely to be stronger than `Q1`

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -0,0 +1,429 @@
...
+            if let Some(h) = existing_header {
+                warn!(
+                    target: LOG_TARGET,
+                    "Received header #{} `{}` that we already have. Ignoring",
+                    h.height,
+                    h.hash().to_hex()
+                );
+                continue;
+            }
+            // self.header_validator.validate(&header)?;
```
stringhandler: Intentionally commented out?

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -0,0 +1,243 @@
...
+    async fn validate_block(
+        &self,
+        header: BlockHeader,
+        mut body: AggregateBody,
+    ) -> Result<(BlockHeader, AggregateBody), BlockSyncError>
+    {
+        let validator = self.block_validator.clone();
+        task::spawn_blocking(move || {
+            // Ensure that the block body is sorted as required by validation
+            body.sort();
```
stringhandler: Rather leave it as is. If the body is not sorted, it should fail validation

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -0,0 +1,243 @@
...
+            let block = block?;
+
+            let header = self
+                .blockchain_state
+                .get_header_by_hash(block.hash.clone())
+                .await?
+                .ok_or_else(|| {
+                    BlockSyncError::ReceivedInvalidBlockBody("Peer sent hash for block header we do not have".into())
+                })?;
+
```
stringhandler, line 173: We  should probably check the height matches at this point as well

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -0,0 +1,243 @@
...
+                target: LOG_TARGET,
+                "Storing block body #{} (PoW = {}, {} input(s), {} output(s), {} kernel(s))",
+                header.height,
+                header.pow_algo(),
+                block_body.inputs().len(),
+                block_body.outputs().len(),
+                block_body.kernels().len(),
+            );
+
+            let (header, block_body) = self.validate_block(header, block_body).await?;
```
stringhandler: If the blocks matches the header we had, but the  block fails validation, e.g. due to the chain balance being wrong, how  does this  recover? I'm worried that the Header sync part of  this process is not able to determine that chains including this header  hash cannot  sync and it will  repeatedly try to sync  to them. 

Perhaps  a bad header hash list can  prevent  that header from being synced? Or otherwise maybe we should  put all of the fields that could cause  validation to fail due to consensus in the header (e.g. total  sums  etc) and the only way this part of the process can fail is when a body cannot be  provided  that matches the value committed in the headers. If no valid body (or full block) can be found for a header from any peers, then that header should be marked as poison

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -233,12 +222,44 @@ impl LMDBDatabase {
...
+    {
+        debug!(
+            target: LOG_TARGET,
+            "height={} ({}), prev={}",
+            header.height,
+            header.hash().to_hex(),
+            header.prev_hash.to_hex()
+        );
+        let prev_data = if header.height == 0 {
+            BlockHeaderAccumulatedData::default()
```
stringhandler: This struct should not have a default, it should always have a specific value attached to a header.
```suggestion
            BlockHeaderAccumulatedData::new(header.hash(), total_kernel_offset)
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -379,7 +403,102 @@ impl LMDBDatabase {
...
+    fn insert_block_body(
+        &mut self,
+        txn: &WriteTransaction<'_>,
+        header_hash: HashOutput,
+        body: AggregateBody,
+    ) -> Result<(), ChainStorageError>
+    {
+        let (inputs, outputs, kernels) = body.dissolve();
+
+        let BlockAccumulatedData {
```
stringhandler: This  struct also does not make sense as a default. The values should be derived from a header or not exist at all.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -141,50 +140,54 @@ impl LMDBDatabase {
...
-                WriteOperation::Delete(delete) => self.op_delete(&write_txn, delete)?,
-                WriteOperation::UpdateBlockAccumulatedData(header_hash, data) => {
-                    self.op_update_block_accumulated_data(&write_txn, header_hash, data)?;
+                Insert(insert) => self.op_insert(&write_txn, insert)?,
+                Delete(delete) => self.op_delete(&write_txn, delete)?,
+                InsertBlockBody(header_hash, body) => {
+                    self.insert_block_body(&write_txn, header_hash, body)?;
                 },
-                WriteOperation::InsertKernel {
+                InsertKernel {
```
stringhandler: These ops may no longer  be necessary.  Will have to see when pruned mode is back though

base_layer/core/src/proof_of_work/target_difficulty.rs
```diff
@@ -20,28 +20,92 @@
...
-    let mut lwma = LinearWeightedMovingAverage::new(block_window, target_time, min_difficulty, max_block_time);
-    for (epoch, difficulty) in target_difficulties {
-        lwma.add(epoch, difficulty)?;
-    }
-    let mut target_difficulty = lwma.get_difficulty();
-    target_difficulty = cmp::min(max_difficulty, cmp::max(target_difficulty, min_difficulty));
-    Ok(target_difficulty)
+}
+
+impl TargetDifficulty {
```
stringhandler: Perhaps this describes the window rather than a single TargetDifficulty value 
```suggestion
impl TargetDifficultyWindow {
```

stringhandler (commented):
I'm not  able to go through the whole thing in one go. MW has  some difficulties that other open block chains do  not, specifically calculating  balances. This presents some problems when syncing headers vs bodies, and at which time we should validate. If we move some of these values into the header, we can valid the chain on the header alone, and then discard a block if the body does not match. This requires a bit more thought though

base_layer/core/src/base_node/sync/header_sync/validator.rs
sdbondi: This is why the validator is private to the header sync module - it's designed to work solely with header sync. Header sync must use it correctly. Header sync will always be validating in order. I have added the check for header connections today but have not pushed it yet :)

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: No just for quicker live testing -  will comment back in

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
sdbondi: We don't actually know if it is sorted - that flag is not sent to us, `is_sorted` simply says that `sorted` was called at least once.

sdbondi: Though I suppose if it isn't sent in canonical order the MMR check will fail, so this could save us the sort operation. I think that's a good optimisation.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: This check is not sufficient. In your example it checks `I2` against `I1` however this is not sufficient to know if the PoW is stronger between the two chains - the only way to know this AFAIK is to download all headers and compare the strength of each fork. I think I put a TODO comment just after this line about this. Header download and validation can be very quick (few seconds) so it may be worth asking the peer for their chain metadata, then if they say they have stronger PoW, downloading headers and checking.

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
sdbondi: The headers are trusted at this point (since header sync downloaded and validated them incl the heights) - I do check that the given hash forms a chain according to prev_hash.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
stringhandler: That is basically what is happening now, I think the only way to know is to download all of the headers that the peer has before doing this check

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
sdbondi: I think the solution may be to keep track of peer's that provided the incorrect information and avoid them (unless they are the only forced sync peer, in which case there is not much we can do except try that node again).

If a block body sync fails, the state machine will return to the header sync which can select another peer, rewind the state and resync headers. As long as we don't sync from that same peer. If all our peers are providing bad data, then I think all we can do is keep banning and trying someone else.

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
stringhandler: Yeah I see the prev_hash is  checked, so it's not  possible for the sync peer to send a block out  of order

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: This check is here as an attempt to ensure that the peer can provide some good headers before we make any changes to our local state (rewinding the headers/block bodies). We might need a "staging" area in the database for headers before they are accepted in so that tips can be compared.

Side note: header validation to ~6000 takes 5 minutes due to the current implementation of randomX validation - if that can be optimised ~6000 header download + validation would take around 30seconds over tor.

base_layer/core/src/proof_of_work/target_difficulty.rs
sdbondi: Yeah not opposed to that name

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: Yeah probably, for now they are the way we do atomicity. Likely we'll move to LMDB-backed atomicity

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: This is the `new` function's signature
```rust
 pub fn new(
        kernels: PrunedHashSet,
        outputs: PrunedHashSet,
        range_proofs: PrunedHashSet,
        deleted: Bitmap,
        total_kernel_sum: Commitment,
        total_utxo_sum: Commitment,
    ) -> Self
```

A default would populate the struct with zero/empty values. I stand corrected but I think that is intuitive to the Default trait. In this case, for the genesis block, the default is the start/initial state of the accumulated data. Perhaps `BlockAccumulatedData::empty()` is more descriptive.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -508,23 +627,77 @@ where B: BlockchainBackend
...
+        }
+
+        Ok(target_difficulties)
     }
 
-    /// `calculate_mmr_roots` takes a block template and calculates the MMR roots for a hypothetical new block that
-    /// would be built onto the chain tip. Note that _no checks_ are made to determine whether the template would
-    /// actually be a valid extension to the chain; only the new MMR roots are calculated
-    pub fn calculate_mmr_roots(&self, template: NewBlockTemplate) -> Result<Block, ChainStorageError> {
+    pub fn fetch_target_difficulties(&self, start_hash: HashOutput) -> Result<TargetDifficulties, ChainStorageError> {
```
stringhandler: When would you call this instead of the above method?

sdbondi: target_difficulties loads the Target difficulty windows for each PoW algo in a single transaction and is used by the header sync.

fetch_target_difficulty method fetches for a single algo and is used in the existing block validator when doing the target difficulty validation for a single header i.e. it loads the window for the given PoW then throws the info away.

stringhandler (dismissed):
Looks good. I think there are a few more things that can be done, but they can be done in follow up PRs

delta1:
Just some clippy warnings to take care of @sdbondi :)

PR #2430, opened 2020-11-11
Implement entry and persistence of custom base node in console wallet

philipr-za:
## Description
This PR adds the functionality to allow a user to specify a custom base node public key and address in the Network tab of the console wallet. If the public key and address can be parsed into a Comms Public Key and valid Net address this bae node is set to be used by all the wallet services and is saved into the wallet’s client key-value store.

On startup if there is a valid public key and address stored in the wallets client key-value store this is used as the base node for the wallet to connect to. If these values don’t exist or cannot be parsed into a valid public key or net address the wallet will fallback to the base node specified in the config.

A function is also added to clear the custom base node which, after a confirmation dialog, will delete the custom data from the database and revert to the config specified base node.

## How Has This Been Tested?
Manually tested

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [ ] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [x] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #2438, opened 2020-11-17
Wallet base node service fallback mode

delta1:
Requires PR #2430 (merged) 

## Description

Implement the ability for the wallet base node service to be configured with multiple peers, and ability to fallback to peer seeds as well (if configured). When the current base node is detected as being offline, the base node service will switch to the next in the list. The console wallet now also shows the latency of the connected base node.  
 

## Motivation and Context

 

Allows the console wallet to be configured with multiple base nodes, and automatic switching when not receiving a response within the configured request maximum age 

## How Has This Been Tested?

 

cargo test, and manually 

## Types of changes

* [ ] New Tests
* [x] Documentation 

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/main.rs
+++ b/applications/tari_console_wallet/src/main.rs
@@ -7,7 +7,6 @@
 #![deny(unknown_lints)]
 #![recursion_limit = "512"]
 use log::*;
-use rand::{rngs::OsRng, Rng};
 use std::{fs, str::FromStr, sync::Arc};
 use structopt::StructOpt;
 use tari_app_utilities::{
@@ -111,14 +110,10 @@ fn main_inner() -> Result<(), ExitCodes> {
         return Ok(());
     }
 
-    let base_node = get_base_node_peer(&config)?;
+    let base_nodes = get_base_node_peers(&config)?;
+    let base_node = base_nodes.first().expect("No base node service peer defined.").clone();
 
-    let wallet = runtime.block_on(setup_wallet(
-        &config,
-        node_identity,
-        base_node.clone(),
-        shutdown.to_signal(),
-    ))?;
+    let wallet = runtime.block_on(setup_wallet(&config, node_identity, base_nodes, shutdown.to_signal()))?;
 
     debug!(target: LOG_TARGET, "Starting app");
 
@@ -144,30 +139,36 @@ fn main_inner() -> Result<(), ExitCodes> {
     result
 }
 
-fn get_base_node_peer(config: &GlobalConfig) -> Result<Peer, ExitCodes> {
-    // base node service peer is defined, so use that
-    if let Ok(base_node) = SeedPeer::from_str(&config.wallet_base_node_service_peer).map(Peer::from) {
-        return Ok(base_node);
-    }
-
-    // pick a random peer from peer seeds config
-    // todo: strategy for picking peer seed: random or lowest latency
-    let mut peer_seeds = config
-        .peer_seeds
+fn get_base_node_peers(config: &GlobalConfig) -> Result<Vec<Peer>, ExitCodes> {
+    // get the configured base node service peers
+    let mut base_node_peers = config
+        .wallet_base_node_service_peers
         .iter()
         .map(|s| SeedPeer::from_str(s))
         .map(|r| r.map(Peer::from))
         .collect::<Result<Vec<_>, _>>()
-        .map_err(|err| ExitCodes::ConfigError(format!("Malformed seed peer: {}", err)))?;
+        .map_err(|err| ExitCodes::ConfigError(format!("Malformed base node service peer: {}", err)))?;
 
-    if !peer_seeds.is_empty() {
-        let idx = OsRng.gen_range(0, peer_seeds.len());
-        return Ok(peer_seeds.remove(idx));
+    // add the peer seeds, if fallback is configured
+    if config.wallet_base_node_service_fallback_enabled {
+        let peer_seeds = config
+            .peer_seeds
+            .iter()
+            .map(|s| SeedPeer::from_str(s))
+            .map(|r| r.map(Peer::from))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|err| ExitCodes::ConfigError(format!("Malformed seed peer: {}", err)))?;
+
+        base_node_peers.extend(peer_seeds);
     }
 
-    Err(ExitCodes::ConfigError(
-        "No peer seeds or base node peer defined in config!".to_string(),
-    ))
+    if base_node_peers.is_empty() {
+        Err(ExitCodes::ConfigError(
+            "No base node peers or peer seeds defined in config!".to_string(),
+        ))
+    } else {
+        Ok(base_node_peers)
+    }
 }
 
 fn wallet_mode(bootstrap: ConfigBootstrap) -> WalletMode {
@@ -189,7 +190,7 @@ fn wallet_mode(bootstrap: ConfigBootstrap) -> WalletMode {
 async fn setup_wallet(
     config: &GlobalConfig,
     node_identity: Arc<NodeIdentity>,
-    base_node: Peer,
+    base_nodes: Vec<Peer>,
     shutdown_signal: ShutdownSignal,
 ) -> Result<WalletSqlite, ExitCodes>
 {
@@ -264,7 +265,7 @@ async fn setup_wallet(
     let mut wallet = Wallet::new(
         wallet_config,
         wallet_backend,
-        transaction_backend.clone(),
+        transaction_backend,
         output_manager_backend,
         contacts_backend,
         shutdown_signal,
@@ -281,14 +282,7 @@ async fn setup_wallet(
     // TODO gRPC interfaces for setting base node
     debug!(target: LOG_TARGET, "Setting base node peer");
     wallet
-        .set_base_node_peer(
-            base_node.public_key.clone(),
-            base_node
-                .addresses
-                .first()
-                .expect("The base node peer should have an address!")
-                .to_string(),
-        )
+        .set_base_node_peers(base_nodes)
         .await
         .map_err(|e| ExitCodes::WalletError(format!("Error setting wallet base node peer. {}", e)))?;
 
--- a/base_layer/wallet/src/base_node_service/handle.rs
+++ b/base_layer/wallet/src/base_node_service/handle.rs
@@ -35,17 +35,20 @@ pub type BaseNodeEventReceiver = broadcast::Receiver<Arc<BaseNodeEvent>>;
 #[derive(Debug)]
 pub enum BaseNodeServiceRequest {
     GetChainMetadata,
+    SetBaseNodePeers(Vec<Peer>),
     SetBaseNodePeer(Box<Peer>),
 }
 /// API Response enum
 #[derive(Debug)]
 pub enum BaseNodeServiceResponse {
     ChainMetadata(Option<ChainMetadata>),
+    BaseNodePeersSet,
     BaseNodePeerSet,
 }
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub enum BaseNodeEvent {
     BaseNodeState(BaseNodeState),
+    BaseNodePeerSet(Box<Peer>),
 }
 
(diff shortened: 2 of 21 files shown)


delta1:
Rebased onto `development` - clippy etc will fail until #2446 is merged

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -312,8 +406,12 @@ where BNResponseStream: Stream<Item = DomainMessage<BaseNodeProto::BaseNodeServi
             "Handling Wallet Base Node Service Request: {:?}", request
         );
         match request {
+            BaseNodeServiceRequest::SetBaseNodePeers(peers) => {
```
philipr-za, line 409: Hmmm it feels a little confusing having these two requests named so similarly. From the first reading it might not be clear what the difference is? I think we need a little more descriptiveness in the naming. I discuss this a bit more in the comments in the handle.

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -92,6 +104,7 @@ pub struct BaseNodeService<BNResponseStream> {
     outbound_messaging: OutboundMessageRequester,
     event_publisher: BaseNodeEventSender,
     base_node_peer: Option<Peer>,
+    base_nodes: Vec<Peer>,
```
philipr-za, line 107: The `base_node_peer` and `base_nodes` are a little confusing perhaps `base_node_peer` should be called `current_base_node_peer`?

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -287,8 +358,31 @@ where BNResponseStream: Stream<Item = DomainMessage<BaseNodeProto::BaseNodeServi
...
+        // reset base node state
+        let state = BaseNodeState::default();
+        self.publish_event(BaseNodeEvent::BaseNodeState(state.clone()));
+        self.set_state(state);
+    }
+
+    fn set_base_node_peers(&mut self, peers: Vec<Peer>) -> Result<(), BaseNodeServiceError> {
+        self.base_nodes = peers;
+
+        if self.base_nodes.is_empty() {
```
philipr-za, line 374: I think you should check this before assigning the erroneous data to the internal state.

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -242,6 +260,51 @@ where BNResponseStream: Stream<Item = DomainMessage<BaseNodeProto::BaseNodeServi
...
+                 max request age ({}).",
+                never_connected,
+                timing_out
+            );
+            let state = self.state.set_online(false);
+            let event = BaseNodeEvent::BaseNodeState(state);
+            self.publish_event(event);
+
+            // switch to a new base node if we can
+            let base_nodes = self.base_nodes.clone();
```
philipr-za, line 292: This clone is not needed? You can work with references right until you get to line 296 where you just clone the peer you need.

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -242,6 +260,51 @@ where BNResponseStream: Stream<Item = DomainMessage<BaseNodeProto::BaseNodeServi
...
+            if base_nodes.len() > 1 {
```
philipr-za, line 293: what happens if there is just 1? You are assuming the set base node is the same as the 1 in the Vec? I think this can possibly not be true as setting the vec and the initial base node are done separately?

base_layer/wallet/src/base_node_service/service.rs
philipr-za: In fact I this these two request should maybe be merged into a single request where the base first base node in the vec becomes the one being monitored? Keeping them separate feels like it can lead to weirdness.

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -73,11 +83,13 @@ impl Default for BaseNodeState {
             chain_metadata: None,
             is_synced: None,
             updated: None,
+            latency: None,
+            online: true,
```
philipr-za, line 87: I believe his should be offline initially? Until you have received a response from the BN you don't know if its online or not so best to assume its offline until the response arrives?

base_layer/wallet/src/base_node_service/service.rs
philipr-za: I realized that this assumption comes from the way the service is initialized in `fn set_base_node_peers(...)` in the wallet container but I think there is a bug waiting to happen because the assumption relies on code outside of this service. We already have a situation where we set another current base node peer outside of that logic (when the console wallet sets a custom base node) and if there was only 1 peer in this list we would never fall back to it. (I also realize in my other comments I say we shouldn't fall back when a custom peer is specified but I think the point still stands that this assumption can be wrong if the public API is used to set a peer and there is only one fallback peer in the list).

base_layer/wallet/src/base_node_service/service.rs
philipr-za: I later found this exact logic living in `fn set_base_node_peers(...)` which I think should be brought into the service rather.

base_layer/wallet/src/wallet.rs
```diff
@@ -248,6 +253,31 @@ where
...
+    /// This function will set the list of base_nodes that the wallet base node service can use, and set the first one
+    /// as the current base node peer for all wallet services.
+    pub async fn set_base_node_peers(&mut self, peers: Vec<Peer>) -> Result<(), WalletError> {
+        info!("Wallet setting {} base node peers.", peers.len());
+
+        if let Some(mut base_node_service) = self.base_node_service.clone() {
+            base_node_service.set_service_peers(peers.clone()).await?;
+        }
+
+        let base_node = peers
```
philipr-za, line 265: I think this code and the logic from `set_base_node_peer(...)` should move into the service because there are some assumptions in the service that rely on this idea and also currently when we fallback to a seed peer we do not update the Transaction or Output manager services at all. We will need to keep the `set_base_node_peer(...)` function as it is used by the FFI, or else we just update the FFI function to use this function and construct a Peer vector with a single value in it.

base_layer/wallet/src/base_node_service/handle.rs
```diff
@@ -35,17 +35,20 @@ pub type BaseNodeEventReceiver = broadcast::Receiver<Arc<BaseNodeEvent>>;
 #[derive(Debug)]
 pub enum BaseNodeServiceRequest {
     GetChainMetadata,
+    SetBaseNodePeers(Vec<Peer>),
```
philipr-za, line 38: I think its a bit confusing why we have two requests with such similar names when they actually are going to end up being used for slightly different purposes. This new one is to provide a set of BNs of which the first one will be set as the current BN to be monitored and the rest can be fell back onto. `SetBaseNodePeer(Box<Peer>)` does not actually clear the set of BNs provided by the previous request and if this BN is offline we will actually fall back to those if they are set. I don't think the request name makes that 100% clear? In the end we are going to use this to set the Custom Base Node and in my other comments I suggest that we do not want to fall back when using a custom set Base Node so perhaps the `SetBaseNodePeer(Box<Peer>)` request name should reflect that use case more clearly?

philipr-za (commented):
An issue I spotted is that when a fallback occurs the base node peer being used by the Transaction Service and Output Manager Service is not updated to reflect it. Must just be careful to not do this is a Custom Base node is currently in use.

Another issue I think we have is that if we set a Custom Base Node and it turns out to be offline we will fallback to the base nodes in the configs secondary base node list. We can discuss this but I believe this is not desired behaviour, as a user if I set a custom base node I am doing it because I want to use that specific base node and I would like to be sure that I am definitely not all of a sudden using a config base node because my custom base node is not responding.

This is not a direct comment on this PR but I thought it worth putting it in the review as I must of missed it in the initial review of this service but if you look at the following line:

https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/base_node_service/service.rs#L275

I think this means if you receive a base node response with the correct key but not the correct type that you will never restore the “remaining” requests. So an erroneous base node response will clear the whole buffer of remaining requests.

applications/tari_console_wallet/src/main.rs
```diff
@@ -111,14 +110,10 @@ fn main_inner() -> Result<(), ExitCodes> {
         return Ok(());
     }
 
-    let base_node = get_base_node_peer(&config)?;
+    let base_nodes = get_base_node_peers(&config)?;
+    let base_node = base_nodes.first().expect("No base node service peer defined.").clone();
```
stringhandler, line 114:
```suggestion
    let base_node = base_nodes.first().ok_or(|| ExitCodes::ConfigError("No base node service peer defined.")).clone();
```

delta1:
> An issue I spotted is that when a fallback occurs the base node peer being used by the Transaction Service and Output Manager Service is not updated to reflect it. Must just be careful to not do this is a Custom Base node is currently in use.
> 
> Another issue I think we have is that if we set a Custom Base Node and it turns out to be offline we will fallback to the base nodes in the configs secondary base node list. We can discuss this but I believe this is not desired behaviour, as a user if I set a custom base node I am doing it because I want to use that specific base node and I would like to be sure that I am definitely not all of a sudden using a config base node because my custom base node is not responding.

Great point, will refactor 

> This is not a direct comment on this PR but I thought it worth putting it in the review as I must of missed it in the initial review of this service but if you look at the following line:
> 
> https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/base_node_service/service.rs#L275
> 
> I think this means if you receive a base node response with the correct key but not the correct type that you will never restore the “remaining” requests. So an erroneous base node response will clear the whole buffer of remaining requests.

That's a nice catch, thanks I'll put it after the match

base_layer/wallet/src/wallet.rs
delta1: It did previously work this way, I had a reason but seem to have forgotten right now. Will take another look.

base_layer/wallet/src/base_node_service/service.rs
stringhandler: Maybe  we should just  have one list of base_node_peers, with the list order being used as the fallback order? I think this  is  the simplest  option  that is also the most configurable.

base_layer/wallet/src/base_node_service/service.rs
delta1: I will take a look

base_layer/wallet/src/base_node_service/service.rs
delta1: In this case "erroneous" is just an empty vec, which is the same as the inital state. But the point is taken

base_layer/wallet/src/base_node_service/service.rs
delta1: Will take a look

stringhandler (commented):
I like it in concept, but it does seem like it would be simpler and cleaner to just maintain  one  list of peers and fallback in sequential order. If needs be, they can be loaded randomly,  except for putting the most preferred in front

comms/src/peer_manager/peer.rs
```diff
@@ -85,7 +90,7 @@ pub struct Peer {
     /// User agent advertised by the peer
     pub user_agent: String,
     /// Metadata field. This field is for use by upstream clients to record extra info about a peer.
-    /// We use a hashmap here so that we can use more than one "info set"
+    /// We use a btreemap here so that we can use more than one "info set"
```
delta1, line 93: this comment change should not have been included

base_layer/wallet/src/base_node_service/service.rs
delta1: @mikethetike that is how it's working

base_layer/wallet/src/base_node_service/service.rs
delta1: nice, I'll try that out

base_layer/wallet/src/base_node_service/handle.rs
delta1: IIRC the issue here was because the other services don't have the same public API. Will take a look

delta1:
Closing this PR in favour of a new one when ready

PR #2439, opened 2020-11-18
Provide initial sync status to merge mining proxy

hansieodendaal:
## Description
Initial sync status is now provided via gRPC to the merge mining proxy; this will enable more efficient use of computing resources when the initial blockchain sync of the linked base node has not been achieved yet. In this implementation, the merge mining proxy will wait for initial sync to be achieved before instructing XMRig to start mining, as otherwise, it cannot provide a valid Tari block template to mine on.

Typical user feedback in the merge mining proxy with XMRig running and if the base node is not synced at startup would be:
```
Listening on 127.0.0.1:7878...

Initial base node sync not achieved, current height at #7039, waiting...
Initial base node sync not achieved, current height at #7053, waiting...
Initial base node sync not achieved, current height at #7064, waiting...
Initial base node sync not achieved, current height at #7119, waiting...
Initial base node sync not achieved, current height at #7174, waiting...
Initial base node sync not achieved, current height at #7225, waiting...
Initial base node sync not achieved, current height at #7269, waiting...
Initial base node sync not achieved, current height at #7329, waiting...
Initial base node sync not achieved, current height at #7384, waiting...
Initial base node sync not achieved, current height at #7441, waiting...
Initial base node sync not achieved, current height at #7459, waiting...
Initial base node sync not achieved, current height at #7519, waiting...
Initial base node sync not achieved, current height at #7568, waiting...
Initial base node sync not achieved, current height at #7615, waiting...
Initial base node sync not achieved, current height at #7719, waiting...
Initial base node sync not achieved, current height at #7777, waiting...
Initial base node sync not achieved, current height at #7794, waiting...
Initial base node sync not achieved, current height at #7859, waiting...
Initial base node sync not achieved, current height at #7900, waiting...
Initial base node sync achieved, continuing

Listening on 127.0.0.1:7878...
```

and if the base node is synced at startup it would only be:
```
Listening on 127.0.0.1:7878...
```

## Motivation and Context
After a fresh installation, or after a restart after some time, if a user start all (base node, console wallet, merge mining proxy, XMrig), merge mining should not start until the base node has achieved initial synced.

## How Has This Been Tested?
Tested with all components in Ubuntu Linux and Windows 10.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/main.rs
+++ b/applications/tari_merge_mining_proxy/src/main.rs
@@ -49,7 +49,7 @@ async fn main() -> Result<(), MmProxyError> {
     let config = initialize()?;
 
     let addr = config.proxy_host_address;
-    println!("Listening on {}...", addr);
+    println!("\nListening on {}...\n", addr);
 
     let config = MergeMiningProxyConfig::from(config);
     let xmrig_service = MergeMiningProxyService::new(config, BlockTemplateRepository::new());
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -26,6 +26,7 @@ use crate::{
     helpers,
 };
 use bytes::BytesMut;
+use core::sync::atomic::AtomicBool;
 use futures::StreamExt;
 use hyper::{
     body::Bytes,
@@ -47,6 +48,7 @@ use std::{
     convert::TryFrom,
     future::Future,
     net::SocketAddr,
+    sync::{atomic::Ordering, Arc},
     task::{Context, Poll},
 };
 use tari_app_grpc::{tari_rpc as grpc, tari_rpc::GetCoinbaseRequest};
@@ -55,6 +57,7 @@ use tari_core::{
     blocks::{Block, NewBlockTemplate},
     proof_of_work::monero_rx,
 };
+use tokio::runtime::Handle;
 use tracing::{debug, error, info, instrument, trace, warn};
 
 pub const LOG_TARGET: &str = "tari_mm_proxy::xmrig";
@@ -68,6 +71,7 @@ pub struct MergeMiningProxyConfig {
     pub monerod_use_auth: bool,
     pub grpc_base_node_address: SocketAddr,
     pub grpc_console_wallet_address: SocketAddr,
+    pub proxy_host_address: SocketAddr,
 }
 
 impl From<GlobalConfig> for MergeMiningProxyConfig {
@@ -80,6 +84,7 @@ impl From<GlobalConfig> for MergeMiningProxyConfig {
             monerod_use_auth: config.monerod_use_auth,
             grpc_base_node_address: config.grpc_base_node_address,
             grpc_console_wallet_address: config.grpc_console_wallet_address,
+            proxy_host_address: config.proxy_host_address,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


delta1 (dismissed):
Tested and worked well

applications/tari_merge_mining_proxy/src/main.rs
```diff
@@ -49,7 +49,7 @@ async fn main() -> Result<(), MmProxyError> {
     let config = initialize()?;
 
     let addr = config.proxy_host_address;
-    println!("Listening on {}...", addr);
+    println!("\nListening on {}...\n", addr);
```
stringhandler, line 52: nit: This should be replaced with a log probably?

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -243,6 +301,23 @@ impl InnerService {
...
+                        "Initial base node sync not achieved, current height at #{}, waiting...",
+                        height
+                    );
+                    debug!(target: LOG_TARGET, "{}", msg);
+                    println!("{}", msg);
+                    return Err(MmProxyError::MissingDataError(" ".to_string() + &msg));
+                }
+                self.initial_sync_achieved.store(true, Ordering::Relaxed);
+                debug!(target: LOG_TARGET, "Initial base node sync achieved, continuing");
+                println!("Initial base node sync achieved, continuing\n");
```
stringhandler: These should be logs, not prints.

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -134,6 +140,7 @@ struct InnerService {
     config: MergeMiningProxyConfig,
     block_templates: BlockTemplateRepository,
     http_client: reqwest::Client,
+    initial_sync_achieved: Arc<AtomicBool>,
```
stringhandler, line 143: I don't think that this value needs to be stored. The base node provides it when called, and it doesn't need to be stored and used without calling the base node. Can't it be read from the base node response each time?

stringhandler (dismissed):
Agree with the idea to not mine during sync, but I think you are a bit heavy handed in storing the sync value in an Arc and having all the extra code around reading and writing it when it is available on every call

applications/tari_merge_mining_proxy/src/main.rs
hansieodendaal: This is intended user feedback; please see updated description

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: This is intended user feedback; please see updated description

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: Yes, it is possible to add the status to `async fn get_new_block_template` instead of to `async fn get_tip_info`, then the code can be simplified as suggested.

hansieodendaal: @mikethetike, this suggestion without having access to the `Arc<AtomicBool>` will result in a poorer user experience, as it will only be possible to inform the user that _"Initial base node sync not achieved, current height at #{}, waiting..."_ while not synced, and not a once-off message of _"Initial base node sync achieved, continuing"_

hansieodendaal:
Code simplification; reduced by 1 function.

stringhandler (dismissed):

hansieodendaal:
Just investigating the generated proto file change quickly

stringhandler (approved):

PR #2431, opened 2020-11-13
Remove checkpoints from BlockchainDatabase

stringhandler:
## Description

An attempt to remove the MMRCheckpoints from the database. 
Some highlights of main changes to review:
1. Storage of MMRPeakData instead of MMRCheckpoints
1. Calculation of MMR roots logic has changed, but outputted values are the same
1. Validation of Spent Inputs has changed
1. Finding link in chain for reorg `find_orphan_chain_tips` has changed to store and maintain the tips in the database.
1. Removal of MemoryDatabase
1. Change of serialization in keys in LMDB
1. All operations in LMDB require a transaction as opposed to creating a transaction implicitly
1. Removal of `for_each_xxxx` functions
1. Removal of `lmdb_vec`
1. `LocalNet` consensus values have been updated to match `ridcully` instead of `rincewind`
1. Removal of `spent_commitments` field in the `HistoricalBlock`. It has been left as an empty vector for compatibility with the current testnet nodes, but should be removed entirely as soon as it can be.
1. A fix to MutableMMR where the size was wrong if the MMRBackend had existing data in it

Areas which have been marked as unimplemented and need to be completed:
1. Pruned mode sync
1. Search for kernel/TXO

NOTE: There have been some updates to tests in #2412 and this branch includes code from that PR

Thanks to @SWvheerden and @sdbondi who I have taken snippets of code from their branches and supplemented code in this PR

## Motivation and Context

Future work: By removing the memory storage in the `BlockchainDatabase` struct we can also remove the `RwLock` on the `BlockchainDatabase`

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [ ] New Tests
* [ ] Documentation

## Checklist:

* [ ] I'm merging against the `development` branch.
* [ ] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [ ] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #2446, opened 2020-11-22
Fixes for cargo fmt, clippy, and test compilation

delta1:
## Description

Bunch of fixes to fix cargo fmt, clippy, and get tests compiling. Just a few tests that need to be updated to pass correctly.

## Motivation and Context

#2431 was a significant refactor and more work still needs to be done, this PR gets us one step closer 

## How Has This Been Tested?

cargo test compiles at least, a few tests still fail 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/metadata.rs
+++ b/base_layer/core/src/chain_storage/metadata.rs
@@ -169,15 +169,27 @@ impl Display for InProgressHorizonSyncState {
 mod test {
     use super::ChainMetadata;
 
+    impl ChainMetadata {
+        pub fn genesis() -> ChainMetadata {
+            ChainMetadata {
+                height_of_longest_chain: 0,
+                best_block: Vec::new(),
+                pruning_horizon: 0,
+                effective_pruned_height: 0,
+                accumulated_difficulty: 0,
+            }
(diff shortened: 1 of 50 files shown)


base_layer/core/src/chain_storage/metadata.rs
```diff
@@ -133,6 +133,18 @@ impl Display for ChainMetadata {
     }
 }
 
+impl Default for ChainMetadata {
```
stringhandler: I actually removed this default in a previous PR. Having the default led a number of checks to see if the values made sense. In cases where the data is not available, I used an Option instead. In test code, I'd suggest specify mock values.

stringhandler (commented):
Thanks for fixing these, however I would prefer to not have a Default for ChainMetadata. Happy to merge without that if you can revert those bits.

delta1:
> Thanks for fixing these, however I would prefer to not have a Default for ChainMetadata. Happy to merge without that if you can revert those bits.

Default removed - tests are still going to require work

stringhandler (approved):

PR #2450, opened 2020-11-24
Tari_script updates

SWvheerden:
## Description
This changes the proposal somewhat for tari script.

### DIFF
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -174,7 +174,8 @@ At a high level, Tari script works as follows:
 
 * The script commitment, which can be adequately represented by the hash of the canonical serialisation of the script in
 binary, is recorded in the transaction kernel. 
-* An additional two secret keys are provided by UTXO owners that signs, the script and the input data.
+* An additional secret key is provided by UTXO owner.
+* The script resolves to a public key that provides ownership and that signs the script input data to stop malleability and prove ownership of the corresponding private key.
 * These two keys are also used to created aggregate body owner offsets providing security against spending.
 
 In addition, the script must be attached to the UTXO. As will be demonstrated later, if this is not the case, the UTXO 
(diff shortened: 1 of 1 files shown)


RFC/src/RFC-0201_TariScript.md
```diff
@@ -269,164 +264,162 @@ pub struct TransactionInput {
...
-    script_signature: Signature
+    /// A signature with K_b, signing the script and input data
+    script_signature: Signature,
+    /// The sender pubkey, K_a
+    sender_pub_key: PublicKey
+    /// The receiver pubkey, K_b
+    receiver_pub_key: PublicKey
 }
 ```
+The input data to the script is sign with the receiver pub key of the Input, as the person spending the input should have control of it.
```
stringhandler: This doesn't sound right. The sender owns the previous output, how can the receiver sign it. Is "the  person spending the input" not the Sender?

SWvheerden: The names can be very confusing. But on the input, the previous owner is the receiver. 

For example  Alice > BoB
C1 > C2

On C1, Alice is the reciever, and on C2, Alice is the sender. 
Even thou  in this TX,  Alice is the spender, spending it by sending it to Bob.

stringhandler: I  see these values are actually the values in the previous output. It would probably be better for clarity to just reference the output via a hash

RFC/src/RFC-0201_TariScript.md
```diff
@@ -269,164 +264,162 @@ pub struct TransactionInput {
...
     /// The serialised script
     script: Vec<u8>,
-    /// The public script key, L
-    script_pub_key: PublicKey
     /// The script input data, if any
     input_data: Vec<u8>
-    /// A signature with k', signing the script and input data
-    script_signature: Signature
+    /// A signature with K_b, signing the script and input data
+    script_signature: Signature,
```
stringhandler, line 282: How does this differ from the owner_offset which already  proves knowledge of the receiver private key?

SWvheerden: Signs the script with the owner offset key. When locks the script and the script input data to not be malleable. The script might have extra spending constrictions on it, that is not in the normal owner offset.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -223,21 +216,22 @@ pub struct TransactionOutput {
     /// A proof that the commitment is in the right range
     proof: RangeProof,
     /// The serialised script
-    script: Vec<u8>,
-    /// The public script key, L
-    script_pub_key: PublicKey
+    script_hash: Vec<u8>,
```
stringhandler, line 220: if  this is not signed by the sender, the receiver can change it

SWvheerden:
Closing for #2489

PR #2456, opened 2020-11-26
Update wallet configuration defaults and in console wallet

philipr-za:
## Description

This PR syncs the default config values in the Wallet Config structs and the app config bootstrap defaults. The PR also updates the tari_console_wallet to actually use the values from the bootstrap config which was previously ignored.

## How Has This Been Tested?
Existing tests

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/wallet.rs
+++ b/base_layer/wallet/src/wallet.rs
@@ -85,6 +85,7 @@ pub struct WalletConfig {
     pub comms_config: CommsConfig,
     pub factories: CryptoFactories,
     pub transaction_service_config: Option<TransactionServiceConfig>,
+    pub output_manager_service_config: Option<OutputManagerServiceConfig>,
     pub buffer_size: usize,
     pub rate_limit: usize,
     pub network: Network,
@@ -92,22 +93,25 @@ pub struct WalletConfig {
 }
 
 impl WalletConfig {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         comms_config: CommsConfig,
         factories: CryptoFactories,
         transaction_service_config: Option<TransactionServiceConfig>,
+        output_manager_service_config: Option<OutputManagerServiceConfig>,
         network: Network,
         base_node_service_config: Option<BaseNodeServiceConfig>,
+        buffer_size: Option<usize>,
+        rate_limit: Option<usize>,
     ) -> Self
     {
         Self {
             comms_config,
             factories,
             transaction_service_config,
-            // This is the default buffer size for the pubsub_connector for the mobile wallet
-            buffer_size: 100,
-            // This is the default rate limit fot the pubsub_connector for the mobile wallet
-            rate_limit: 5,
+            output_manager_service_config,
+            buffer_size: buffer_size.unwrap_or_else(|| 1500),
+            rate_limit: rate_limit.unwrap_or_else(|| 50),
             network,
             base_node_service_config,
         }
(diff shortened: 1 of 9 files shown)


base_layer/wallet/src/wallet.rs
```diff
@@ -85,29 +85,33 @@ pub struct WalletConfig {
...
             comms_config,
             factories,
             transaction_service_config,
-            // This is the default buffer size for the pubsub_connector for the mobile wallet
-            buffer_size: 100,
-            // This is the default rate limit fot the pubsub_connector for the mobile wallet
-            rate_limit: 5,
+            output_manager_service_config,
+            buffer_size: buffer_size.unwrap_or_else(|| 1500),
+            rate_limit: rate_limit.unwrap_or_else(|| 1000),
```
hansieodendaal: I think we should be careful with these; it will impact the memory footprint in the mobile wallet. The rate limit of `1000` proposed here is not balanced with the buffer size; better to throttle it a bit and not have messages dropped under stress conditions. I would propose `50`, then the `1500` buffer will fill up in 30 s if nothing is processed.

common/src/configuration/utils.rs
```diff
@@ -80,13 +80,13 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
         default_subdir("wallet/console-wallet.dat", Some(&bootstrap.base_path)),
     )
     .unwrap();
-    cfg.set_default("wallet.base_node_query_timeout", 900).unwrap();
-    cfg.set_default("wallet.transaction_broadcast_monitoring_timeout", 600)
+    cfg.set_default("wallet.base_node_query_timeout", 60).unwrap();
+    cfg.set_default("wallet.transaction_broadcast_monitoring_timeout", 60)
```
hansieodendaal, line 88: These values should be in line with what is documented in the 2x `*.toml` files, or vice versa:
```
# This is the timeout period that will be used to monitor TXO queries to the base node (default = 30). Larger values
# are needed for wallets with many (>1000) TXOs to be validated.
base_node_query_timeout = 120
# This is the timeout period that will be used for base node broadcast monitoring tasks (default = 30)
#transaction_broadcast_monitoring_timeout = 30
# This is the timeout period that will be used for chain monitoring tasks (default = 30)
#transaction_chain_monitoring_timeout = 30
# This is the timeout period that will be used for sending transactions directly (default = 20)
#transaction_direct_send_timeout = 20
# This is the timeout period that will be used for sending transactions via broadcast mode (default = 30)
#transaction_broadcast_send_timeout = 30
```
Please align the `*.toml` files with the config defaults.

common/src/configuration/utils.rs
```diff
@@ -80,13 +80,13 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
...
         .unwrap();
-    cfg.set_default("wallet.transaction_chain_monitoring_timeout", 15)
+    cfg.set_default("wallet.transaction_chain_monitoring_timeout", 60)
```
hansieodendaal, line 90: See above

common/src/configuration/utils.rs
```diff
@@ -80,13 +80,13 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
...
         .unwrap();
-    cfg.set_default("wallet.transaction_direct_send_timeout", 600).unwrap();
-    cfg.set_default("wallet.transaction_broadcast_send_timeout", 600)
+    cfg.set_default("wallet.transaction_direct_send_timeout", 20).unwrap();
+    cfg.set_default("wallet.transaction_broadcast_send_timeout", 60)
```
hansieodendaal, line 93: See above

hansieodendaal (commented):
Looking good, definitely fixing a previous oversight. Just some specific comments belo.

hansieodendaal (approved):
LGTM - untested as it stands, but will put it through its paces in the next stress test.

PR #2466, opened 2020-11-30
Edge-case fixes for connection management and protocol negotiation

sdbondi:
## Description

- Bug fix for protocol negotiation would mistakenly be optimistic even if a peer did not advertise
  a particular protocol. This can cause long timeout delays when protocol negotiation fails.
- Connection tie breaking takes connection age into account. If a connection impolitely disconnects,
  the connection can linger on the other end for a long time. During that time, connection tie-breaking
  consensus could prevent a new connection from being established. A new connection cannot be made until
  the old one drops (could be a few minutes on tor). This fixes this by assuming that connection
  tie-breaking should only occur if two connections establish within less than a minute of each other.
- Discovery failure will always return the error to the caller instead of cancelling the reply
- `wait_for_connectivity` returns the number of connected peers when timing out, could be used by callers to proceed with limited connectivity

## Motivation and Context

Broken connections will not indefinitely prevent new connections from being established.
Protocol negotiation will fail quickly if a node does not accept a protocol.  
Ref #2436 

## How Has This Been Tested?

Tested on base node

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/ui/app.rs
+++ b/applications/tari_console_wallet/src/ui/app.rs
@@ -243,7 +243,7 @@ async fn get_custom_base_node_peer_from_db(wallet: &mut WalletSqlite) -> Option<
                 address.into(),
                 PeerFlags::default(),
                 PeerFeatures::COMMUNICATION_NODE,
-                &[],
+                Default::default(),
                 Default::default(),
             ))
         },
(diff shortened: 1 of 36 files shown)


philipr-za (approved):
LGTM

PR #2468, opened 2020-12-01
Blockchain database, target difficulty and validator improvements

sdbondi:
## Description

**PoW**

- Refactored target difficulty code to be more ergonomic
- Support a "rolling" window that can be updated with more data points
  as validation proceeds. Ref #2436
- Blake PoW algo removed (tests and supporting code), the enum still remains 
  to allow compatibility with the blockchain db

**Chain storage**

- Allow ranges of headers/blocks to be retrieved from storage
- Support for more headers than block bodies in blockchain database (Ref
  #2436)
- Various minor code improvements
- API changes in BlockchainDatabase reflect the blockchain state service api more closely, 
   and so the state service will be removed in future iterations. 
- Moved block accumulated data into its own module
- Split up inserting header and inserting block body functions. 
- Insert block body occurs within a single DB transaction

**Misc**

- Fixed list-headers n command (#2426)
- If a peer sends invalid chain metadata, the node will not accept it. Before it would
  silently treat accumulated difficulty as 0 if it is invalid.
 - Removed `delete_and_compress` call from MutableMmr with boolean compress flag in favour
   of existing `delete` and `compress` functions.

## Motivation and Context

Improvements which support faster validations and initial block sync (Ref #2436 )

## How Has This Been Tested?

Added unit tests for some blockchain db calls

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -364,11 +364,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         let cm = ConsensusManagerBuilder::new(self.node_config.network.into()).build();
         let block_hash = new_block.hash();
         let mining_hash = new_block.header.merged_mining_hash();
-        let pow = match new_block.header.pow.pow_algo {
-            PowAlgorithm::Monero => 0,
-            PowAlgorithm::Blake => 1,
-            PowAlgorithm::Sha3 => 2,
-        };
+        let pow = new_block.header.pow_algo() as i32;
```
SWvheerden, line 367: in grpc block template we are using u64, why i32 here?

applications/tari_base_node/src/parser.rs
```diff
@@ -834,25 +840,19 @@ impl Parser {
...
-            match handler.get_blocks(vec![height]).await {
+            match blockchain.fetch_blocks(height..=height).await {
                 Err(err) => {
-                    println!("Failed to retrieve blocks: {:?}", err);
-                    warn!(
-                        target: LOG_TARGET,
-                        "Error communicating with local base node: {:?}", err,
-                    );
+                    println!("Failed to retrieve blocks: {}", err);
+                    warn!(target: LOG_TARGET, "{}", err);
```
SWvheerden, line 848: I prefer debug printing in the log, normal print only prints the top error. Debug prints the entire error stack which is much more usefull for debugging.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -180,12 +179,16 @@ where T: BlockchainBackend + 'static
             },
             NodeCommsRequest::FetchHeadersAfter(header_hashes, stopping_hash) => {
                 // Send from genesis block if none match
```
SWvheerden:
```suggestion
              
```
 Invalid commit

base_layer/core/src/blocks/block_header.rs
```diff
@@ -346,12 +347,12 @@ mod test {
     use tari_crypto::tari_utilities::Hashable;
     #[test]
     fn from_previous() {
-        let mut h1 = crate::proof_of_work::blake_test::get_header();
+        let mut h1 = crate::proof_of_work::sha3_test::get_header();
```
SWvheerden, line 350: Did we remove the blake stuff?

I say either get a nonce with a non 0 for sha, or leave it as blake. 
We wrote the code so the blake/sha/monero/what ever algo will work

This test is nopt testing blake or sha, its testing if the difficulties are correctly added

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -186,7 +186,7 @@ impl ConsensusConstants {
     pub fn rincewind() -> Vec<Self> {
         let difficulty_block_window = 90;
         let mut algos1 = HashMap::new();
-        algos1.insert(PowAlgorithm::Blake, PowAlgorithmConstants {
+        algos1.insert(PowAlgorithm::Sha3, PowAlgorithmConstants {
```
SWvheerden: Might as well remove rincewind entirely then make these sha3. 
This will break the entire rincewind chain.

base_layer/core/src/chain_storage/target_difficulties.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use crate::{
+    blocks::BlockHeader,
+    consensus::ConsensusManager,
+    proof_of_work::{PowAlgorithm, TargetDifficultyWindow},
+};
+
+#[derive(Debug, Clone)]
+pub struct TargetDifficulties {
+    monero: TargetDifficultyWindow,
```
SWvheerden, line 31: why not use a hashmap like the constants. Then we dont have to worry about the unimplemented below. The code should be fully workable dynamically depending on what the constants are set up to be.

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -117,6 +117,7 @@ pub mod util {
     use crate::U256;
 
     /// This will provide the difficulty of the hash assuming the hash is big_endian
+    #[allow(dead_code)]
```
SWvheerden: This should not be dead code. This should be used by the sha3 pow

base_layer/core/src/proof_of_work/sha3_pow.rs
```diff
@@ -53,10 +51,9 @@ pub fn sha3_hash(header: &BlockHeader) -> Vec<u8> {
         .to_vec()
 }
 
-pub fn sha3_difficulty_with_hash(header: &BlockHeader) -> (Difficulty, Vec<u8>) {
+fn sha3_difficulty_with_hash(header: &BlockHeader) -> (Difficulty, Vec<u8>) {
     let hash = sha3_hash(header);
     let hash = Sha3_256::digest(&hash);
-    trace!(target: LOG_TARGET, "Sha3 hash result: {}", hash.to_vec().to_hex());
     let scalar = U256::from_big_endian(&hash); // Big endian so the hash has leading zeroes
```
SWvheerden: Should use the function above, thats flagged as dead

base_layer/core/src/validation/block_validators.rs
```diff
@@ -63,23 +68,41 @@ impl Validation<Block> for StatelessBlockValidator {
     /// 1. Is there precisely one Coinbase output and is it correctly defined with the correct amount?
     /// 1. Is the accounting correct?
     fn validate(&self, block: &Block) -> Result<(), ValidationError> {
-        let block_id = format!("block #{} ({})", block.header.height, block.hash().to_hex());
-        check_block_weight(block, &self.rules.consensus_constants(block.header.height))?;
+        if block.header.height == 0 {
+            warn!(target: LOG_TARGET, "Attempt to validate genesis block");
```
SWvheerden, line 73: Why have we added this?
I see no reason why cant check the genesis block. It should not fail?

base_layer/core/src/validation/block_validators.rs
```diff
@@ -63,23 +68,41 @@ impl Validation<Block> for StatelessBlockValidator {
...
         // Check that the inputs are are allowed to be spent
         block.check_stxo_rules()?;
         trace!(target: LOG_TARGET, "SV - Output constraints are ok for {} ", &block_id);
         check_cut_through(block)?;
         trace!(target: LOG_TARGET, "SV - Cut-through is ok for {} ", &block_id);
-        check_coinbase_output(block, self.rules.clone(), &self.factories)?;
+        check_coinbase_output(block, &self.rules, &self.factories)?;
         trace!(target: LOG_TARGET, "SV - Coinbase output is ok for {} ", &block_id);
-        check_accounting_balance(block, self.rules.clone(), &self.factories)?;
+        check_accounting_balance(block, &self.rules, &self.factories)?;
```
SWvheerden, line 106: You can safely and cheaply clone consensusmanager, its an arc.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -99,6 +122,77 @@ impl FullConsensusValidator {
     pub fn new(rules: ConsensusManager) -> Self {
         Self { rules }
     }
+
+    /// Calculates the achieved and target difficulties at the specified height and compares them.
+    fn check_achieved_and_target_difficulty<B: BlockchainBackend>(
```
SWvheerden, line 128: Why have you removed the help function?
This will be shared between block sync, pruning sync, header sync etc.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -99,6 +122,77 @@ impl FullConsensusValidator {
...
+            let mut target_difficulties = self.rules.new_target_difficulty(pow_algo, height);
+            for height in (0..=height).rev() {
+                let header = fetch_header(&*db, height)?;
+                if header.pow.pow_algo == pow_algo {
+                    target_difficulties.add_front(header.timestamp(), header.target_difficulty());
+                    if target_difficulties.is_full() {
+                        break;
+                    }
+                }
+            }
```
SWvheerden: Is this not duplicated here: pub fn fetch_target_difficulty(

base_layer/core/src/validation/helpers.rs
```diff
@@ -21,43 +21,75 @@
...
+) -> Result<(), ValidationError>
+{
+    if timestamps.is_empty() {
+        return Err(ValidationError::BlockHeaderError(
+            BlockHeaderValidationError::InvalidTimestamp,
+        ));
+    }
+
+    let median_timestamp = calc_median_timestamp(timestamps);
     if block_header.timestamp < median_timestamp {
```
SWvheerden: I believe this should be <=, not just <. 
If the block timestamp is allowed to be equal to the median. Then the block chain will never move forward in terms of timestamps.

base_layer/core/tests/median_timestamp.rs
```diff
@@ -1,138 +0,0 @@
...
-    let height = store.get_chain_metadata().unwrap().height_of_longest_chain();
-    median_timestamp = get_median_timestamp(get_header_timestamps(
-        &*store.db_read_access().unwrap(),
-        height,
-        timestamp_count,
-    ))
-    .expect("median returned an error");
-    // Median timestamp should be block 3 and not block 2
-    assert_eq!(median_timestamp, timestamps[2]);
-}
```
SWvheerden, line 138: Why remove all of these?
We are losing test coverage.

SWvheerden (commented):
Have a few comments I left. 
One reservation I have here is that there seems to be a lot of refactoring here. And while refactoring might improve the code quality. But we have also removed quite a bit of our code coverage to get it refactored. I dont believe the code quality gain is worth the loss in code coverage.

base_layer/core/tests/median_timestamp.rs
sdbondi: That function does not exist anymore. I agree that we need to add tests to cover a lot of the backend, but unfortunately most of the current tests create maintenance issues. I should be able to easily add tests for the refactor. Clean code is important. It makes testing easier, allows easier changes, easier onboarding for new devs, reduces bugs, performance (critical for block validation during sync!) etc. The benefits are manifold.

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
sdbondi: That is how it was code before - from looking at the definition in the proto file, its a proto enum type which are represented as i32

applications/tari_base_node/src/parser.rs
sdbondi: That is only true if `thiserror` isn't used properly. The `Display` impl should be descriptive enough that if a developer sees it, they will know what the problem is

base_layer/core/src/blocks/block_header.rs
sdbondi: I spoke to @mikethetike and we decided to remove blake to remove unused code from the codebase.

base_layer/core/src/chain_storage/target_difficulties.rs
sdbondi: I disagree:
1. A HashMap can have anything in it, so requires more code to deal with 
1. A HashMap is allocated on the heap and so requires allocation 
1. Needs to be initialized at runtime
1. Using fields makes it impossible at compile time to not have target difficulties for all algos (required for header sync) - each field must exist  (if you think about it, why should getting a Pow target difficulty window possibly not be available? We know what the Pow algos are at compile time)
1. This means that the `get` function does not have to return an `Option` or `Result` which makes client code easier
1. The unimplemented will be removed when we remove blake (or we can implement it) - so it's not a concern.

base_layer/core/src/consensus/consensus_constants.rs
sdbondi: Ah agree - I'll change these back, I just wanted to remove any reference to Blake to make it easier to remove when we can release a db breaking change. Can probably also remove rincewind then.

base_layer/core/src/proof_of_work/difficulty.rs
sdbondi: It isn't used

base_layer/core/src/proof_of_work/sha3_pow.rs
sdbondi: I'll fix that for you :)

base_layer/core/src/proof_of_work/difficulty.rs
sdbondi: Ah the code was duplicated inside the sha3 function - I'll use this function there

base_layer/core/src/validation/block_validators.rs
sdbondi: 1. We don't validate the genesis block anywhere in the code 
1. The genesis block is part of consensus and isn't a normal block (it doesn't necessarily follow the same rules as blocks do)
1. Many checks have to have some exception for the genesis block
e.g.

```rust
if block.height == 0 { return EverythingIsFine } else { do the real check }
```

There is a inherent security risk if a peer managed to pass a block of height 0 with whatever they wanted in, would that be added to the blockchain? If we just say that we cannot validate any block @ height 0, then that all goes away

base_layer/core/src/validation/block_validators.rs
sdbondi: Passing a reference does not require an atomic update of the Arc's counter and so I believe this is optimal. As a general habit, one should always pass in a reference if that is all that is needed (that is unless you are just going to clone inside the function, in which case pass an owned object, cloning at the call site if necessary).

base_layer/core/src/validation/block_validators.rs
sdbondi: It's self contained and will not be needed if the fast header / block sync is used.

base_layer/core/src/validation/block_validators.rs
sdbondi: Not sure if you are referring to `BlockchainDatabase::fetch_target_difficulty` - if so, the reason I duplicate code is because the validator only takes `B: BlockchainBackend` which does not have fetch_target_difficulty. The validators should be refactored have access to BlockchainDatabase, but then the BlockchainDatabase also needs to be refactored to allow the to perform its work within one read transaction. So this is a stop gap. TODO: refactor

base_layer/core/src/validation/helpers.rs
sdbondi: It sounds like that is possible after a few blocks with manipulated timestamps given enough mining control. Changing this shouldn't break consensus unless a node actually sets the same epoch time - I'll have to test if this is ok.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
sdbondi: Fixed 👍

base_layer/core/src/validation/helpers.rs
sdbondi: I've committed this change after testing it by synching a node from scratch - all blocks passed this validation

base_layer/core/src/chain_storage/target_difficulties.rs
stringhandler: Yeah,  this is more of a style thing. I don't mind either. The advantage of having a hashmap is that you can do `TargetDifficulties[pow.algo]` instead of matches everywhere,  but  that can be implemented here with a single method that does the match. Since it's done this way already, it's probably fine

base_layer/core/src/base_node/validators/headers.rs
```diff
@@ -70,25 +64,19 @@ impl<B: BlockchainBackend> Validation<BlockHeader> for HeaderValidator<B> {
 }
 
 impl<B: BlockchainBackend> HeaderValidator<B> {
-    pub fn is_genesis(&self, block_header: &BlockHeader) -> bool {
-        block_header.height == 0 && self.rules.get_genesis_block_hash() == block_header.hash()
-    }
-
     /// Calculates the achieved and target difficulties at the specified height and compares them.
     pub fn check_achieved_and_target_difficulty(&self, block_header: &BlockHeader) -> Result<(), ValidationError> {
         // We cant use the actual tip, as this is used by the header sync, the tip has not yet been downloaded, but we
```
stringhandler: Is this comment still relevant?

base_layer/core/src/blocks/block_header.rs
```diff
@@ -346,12 +347,12 @@ mod test {
...
         h1.nonce = 7600; // Achieved difficulty is 18,138;
```
stringhandler: Comment  needs an  update

base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -67,71 +72,192 @@ where F: FnOnce() -> R {
...
+    make_async_fn!(horizon_sync_spend_utxos(hash: Vec<HashOutput>) -> (), "horizon_sync_spend_utxos");
+
+    //---------------------------------- Misc. --------------------------------------------//
+    make_async_fn!(fetch_block_timestamps(start_hash: HashOutput) -> RollingVec<EpochTime>, "fetch_block_timestamps");
+
+    make_async_fn!(fetch_target_difficulties(start_hash: HashOutput) -> TargetDifficulties, "fetch_target_difficulties");
+
+    make_async_fn!(fetch_block_hashes_from_header_tip(n: usize, offset: usize) -> Vec<HashOutput>, "fetch_block_hashes_from_header_tip");
+
+    pub fn new(db: BlockchainDatabase<B>) -> Self {
```
stringhandler: nit: Strange place for a new()

base_layer/core/src/common/rolling_vec.rs
```diff
@@ -0,0 +1,175 @@
...
+//  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::ops::Deref;
+
+#[derive(Debug, Clone)]
+pub struct RollingVec<T>(Vec<T>);
```
stringhandler, line 28: nit: `VecDeque` is already a ring buffer, so might be better suited to this job

base_layer/core/src/validation/block_validators.rs
```diff
@@ -118,166 +212,64 @@ impl<B: BlockchainBackend> StatefulValidation<Block, B> for FullConsensusValidat
...
-fn check_duplicate_transactions_inputs(block: &Block) -> Result<(), ValidationError> {
-    trace!(
-        target: LOG_TARGET,
-        "Checking duplicate inputs and outputs on block with hash {}",
-        block.hash().to_hex()
-    );
-    for i in 1..block.body.inputs().len() {
-        if block.body.inputs()[i..].contains(&block.body.inputs()[i - 1]) {
+fn check_duplicate_transactions_inputs(body: &AggregateBody) -> Result<(), ValidationError> {
+    for i in 1..body.inputs().len() {
```
stringhandler: If the inputs are sorted, can't you just do one pass and see if it matches the previous input?

base_layer/core/src/validation/block_validators.rs
```diff
@@ -118,166 +212,64 @@ impl<B: BlockchainBackend> StatefulValidation<Block, B> for FullConsensusValidat
...
+        if body.inputs()[i..].contains(&body.inputs()[i - 1]) {
             return Err(ValidationError::custom_error("Duplicate Input"));
         }
     }
-    for i in 1..block.body.outputs().len() {
-        if block.body.outputs()[i..].contains(&block.body.outputs()[i - 1]) {
+    for i in 1..body.outputs().len() {
```
stringhandler: Same here

base_layer/core/src/validation/transaction_validators.rs
```diff
@@ -90,14 +91,21 @@ fn verify_timelocks(tx: &Transaction, current_height: u64) -> Result<(), Validat
...
-
-    let BlockAccumulatedData { deleted, .. } = db.fetch_block_accumulated_data(&tip.hash())?;
+    // `ChainMetadata::best_block` must always have the hash of the tip block.
+    // NOTE: the backend makes no guarantee that the tip header has a corresponding full body (interrupted header sync,
+    // pruned node) however the chain metadata MUST always be correct
+    let metadata = db.fetch_chain_metadata()?;
+    let data = db
+        .fetch_block_accumulated_data(metadata.best_block())?
+        .unwrap_or_else(|| {
+            panic!(
```
stringhandler, line 101: Panic seems a bit harsh here

stringhandler (commented):
Happy in general. There's one panic that needs to be sorted out  and  few  other comments that are less important

base_layer/core/src/chain_storage/async_db.rs
sdbondi: Yeah,  `cargo fmt` at work, but I'll create a separate impl block for these so that they can be just after the struct def

base_layer/core/src/common/rolling_vec.rs
sdbondi: Good idea, however after looking at it we lose `impl Deref<Target=[T]>` which has a few blanket impls for slices and some ergonomics with `&[T]`

base_layer/core/src/validation/block_validators.rs
sdbondi: You're right, and I think we must sort the inputs and outputs before validating even in initial block sync (we can't rely that the MR in the header is valid). Looking at Vec's built-in sort implementation, it is very efficient at sorting Vecs that are already mostly sorted, which is how they should be when received from a peer, so I think this is a great optimisation.

base_layer/core/src/validation/transaction_validators.rs
sdbondi: I'm in two minds about this. It _should_ be impossible for the best block in chain metadata to ever not be an existing block and have accumulated data. If that is not the case, something has gone wrong (the db can be considered "corrupt"). Thoughts? (happy to change this)

base_layer/core/src/base_node/validators/headers.rs
sdbondi: I've removed the comment and updated `fetch_target_difficulty` to be consistent.

```rust
    /// Returns the set of target difficulties for the specified proof of work algorithm. The calculated target
    /// difficulty will be for the given height i.e calculated from the previous header backwards until the target
    /// difficulty window is populated according to consensus constants for the given height.
    pub fn fetch_target_difficulty( pow_algo: PowAlgorithm, height: u64) ...
```

stringhandler (approved):
Nice. Let's try it out

PR #2469, opened 2020-12-01
Add initial sync override logic

hansieodendaal:
## Description
~~This PR adds the ability to override initial sync logic in a base node, by starting up with an `initial_sync_startup_state = true` status. This will enable UTXO queries to be handled, broadcast blocks and transactions to be processed and merge mining to proceed without having confirmation of having achieved initial sync with the blockchain.
Fixed compile errors in unit tests.~~

**Update:** This PR adds the ability to determine the initial sync logic in the merge mining proxy, by starting up with a `wait_for_initial_sync_at_startup` instruction either `true` (default value) or `false`. Setting this value to `false` will enable merge mining to proceed without having confirmation of the linked base node having achieved initial sync to the blockchain or not. Setting this value to `true` will result in merge mining only starting when initial sync has been achieved.

## Motivation and Context
~~This is primarily intended for edge cases, for example when a base node is isolated at startup or cannot sync for some reason, or if it does not care about having achieved initial sync or not.~~
~~_**Note:** It is also pretty handy in setting up a selfish mining attack using merge mining with XMRig, as otherwise the merge mining proxy will not proceed because it is checking the status._~~

**Update:** This gives a merge mining operator the ability to choose when merge mining should start after startup; immediately or only after base node initial sync has been reached.

## How Has This Been Tested?

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


SWvheerden:
The code looks fine. But I am against the idea of bypassing the initial sync code. This code is there to help ensure stability in the base node and network as a whole. 
The code as is it, should allow the node to run by itself if it looses touch with the network. It wont immediately, but it will after a while. 

And submitting code because it makes in attack easier seems like the wrong way for code security.

hansieodendaal:
> The code looks fine. But I am against the idea of bypassing the initial sync code. This code is there to help ensure stability in the base node and network as a whole.
> The code as is it, should allow the node to run by itself if it looses touch with the network. It wont immediately, but it will after a while.
> 
> And submitting code because it makes in attack easier seems like the wrong way for code security.

Philosophy of where to allow merge mining to proceed changed due to comment. Now solely determined within the merge mining proxy configuration settings.

stringhandler (approved):
Nice. Yeah I believe this approach is better than the previous one

PR #2470, opened 2020-12-01
Allow submitting of transactions via GRPC

stringhandler:
## Description

Adds a GRPC method and cucumber test for submitting transactions via GRPC 

## Motivation and Context

The main aim here is to allow non-rust clients to submit transactions, but also to expose this functionality to GRPC so that Cucumber tests can be written

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


applications/tari_app_grpc/src/conversions/transaction.rs
```diff
@@ -33,3 +35,18 @@ impl From<Transaction> for grpc::Transaction {
...
     }
 }
+
+impl TryFrom<grpc::Transaction> for Transaction {
+    type Error = String;
+
+    fn try_from(source: grpc::Transaction) -> Result<Self, Self::Error> {
+        Ok(Self {
+            offset: RistrettoSecretKey::from_bytes(&source.offset)
+                .map_err(|e| format!("Offset is not valid:{}", e.to_string()))?,
```
sdbondi, line 45: nit: 
`{}` in `format!` and `ToString::to_string` both call the `Display` impl so calling `to_string` with `format!` is superfluous.
```suggestion
                .map_err(|e| format!("Offset is not valid: {}", e))?,
```

We should probably at some point change the conversion error types to use `anyhow`.

sdbondi (dismissed):
Nice! LGTM

PR #2477, opened 2020-12-05
Integrate Rewindable outputs into Transactions

philipr-za:
## Description

This PR adds the ability for outputs generated in transactions to be generated using rewind keys that can allow for a partial or full rewinding or a range proof. The Output Manager Service now derives two extra keys using the master_seed key and two special branches in the key manager to be used as a `rewind_key` and `rewind_blinding_key`. All transactions that are produced by the wallet (Coinbases, Sender and Recipient transactions) will now be rewindable using these keys. The public versions allow for the first level of rewinding that reveals the proof message and value and the private versions allow for a full rewind to also reveal the spending key.

It is crucial to note that the private `rewind_key` and `rewind_blinding_key` are both extremely sensitive. Because we use the same keys for all outputs generated by a given wallet if these keys are compromised the entire wallet is compromised, not just the spending key for a single UTXO. Thus a focus of the updates to the Transaction negotiation protocols and CoinbaseBuilder in the `tari_core` crate was to ensure that the private rewind keys are only ever used in memory and are never serialised in the output of the protocols. They are only ever included in the serialised output as fully formed range proofs which are secure. 

A Public API call was added to the Output Manager Service to return the public rewind keys for use as view keys and to use in testing.

The PR does not provide wallet restoration or the functionality to use the view keys as yet.

Summary:
- Added `as_rewindable_transaction_output` to Unblinded Output
- Added two levels of rewinding to `TransactionOutput`
- Added being able to produce a rewindable output to `ReceiverTransactionProtocol`
- Added being able to produce a rewindable change output to `SenderTransactionInitializer` and the `SingleReceiverTransactionProtocol`
- Added being able to produce rewindable Coinbase outputs to the `CoinbaseBuilder`
- Updated OutputManagerService to produce the two rewind keys from its embedded KeyManager
- Updated requesting the Recipients half of a inbound transaction to use rewindable outputs
- Updated producing the Sender half of an outbound transaction to use rewindable outputs and change.
- Updated producing a Coinbase transaction to create a rewindable output
- Add API to get public rewind keys from Output Manager to allow for testing rewind and providing view keys.

## Motivation and Context
This is the first half of what is needed for wallet restoration, this PR will mean that outputs on the blockchain will now be rewindable with keys generated from the wallet's master seed.

## How Has This Been Tested?
Tests were provided for all the elements added to 'tari_core' and tests were added and updated in the wallet to check for rewindability

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/coinbase_builder.rs
+++ b/base_layer/core/src/transactions/coinbase_builder.rs
@@ -33,7 +33,7 @@ use crate::{
             TransactionBuilder,
             UnblindedOutput,
         },
-        transaction_protocol::{build_challenge, TransactionMetadata},
+        transaction_protocol::{build_challenge, RewindData, TransactionMetadata},
         types::{BlindingFactor, CryptoFactories, PrivateKey, PublicKey, Signature},
     },
 };
@@ -62,6 +62,7 @@ pub struct CoinbaseBuilder {
     fees: Option<MicroTari>,
     spend_key: Option<PrivateKey>,
     private_nonce: Option<PrivateKey>,
+    rewind_data: Option<RewindData>,
 }
 
 impl CoinbaseBuilder {
@@ -74,6 +75,7 @@ impl CoinbaseBuilder {
             fees: None,
             spend_key: None,
             private_nonce: None,
+            rewind_data: None,
         }
     }
 
@@ -98,6 +100,12 @@ impl CoinbaseBuilder {
     /// The nonce to be used for this transaction. This will usually be provided by a miner's wallet instance.
     pub fn with_nonce(mut self, nonce: PrivateKey) -> Self {
         self.private_nonce = Some(nonce);
+        self
+    }
+
+    /// Add the rewind data needed to make this coinbase output rewindable
+    pub fn with_rewind_data(mut self, rewind_data: RewindData) -> Self {
+        self.rewind_data = Some(rewind_data);
         self
     }
 
@@ -151,9 +159,15 @@ impl CoinbaseBuilder {
         let sig = Signature::sign(key.clone(), nonce, &challenge)
             .map_err(|_| CoinbaseBuildError::BuildError("Challenge could not be represented as a scalar".into()))?;
         let unblinded_output = UnblindedOutput::new(total_reward, key, Some(output_features));
-        let output = unblinded_output
-            .as_transaction_output(&self.factories)
-            .map_err(|e| CoinbaseBuildError::BuildError(e.to_string()))?;
+        let output = if let Some(rewind_data) = self.rewind_data.as_ref() {
+            unblinded_output
+                .as_rewindable_transaction_output(&self.factories, rewind_data)
+                .map_err(|e| CoinbaseBuildError::BuildError(e.to_string()))?
+        } else {
+            unblinded_output
+                .as_transaction_output(&self.factories)
+                .map_err(|e| CoinbaseBuildError::BuildError(e.to_string()))?
+        };
         let kernel = KernelBuilder::new()
             .with_fee(0 * uT)
             .with_features(kernel_features)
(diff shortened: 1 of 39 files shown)


delta1:
Really great work! 👏  I'm digging into it, but in the meanwhile there's a small clippy/build issue for an unused variable

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -266,6 +282,39 @@ mod test {
...
+        let rewind_data = RewindData {
+            rewind_key: rewind_key.clone(),
+            rewind_blinding_key: rewind_blinding_key.clone(),
+            proof_message: proof_message.clone(),
+        };
+
+        let p = TestParams::new();
+        let (builder, rules, factories) = get_builder();
+        let builder = builder
+            .with_block_height(42)
```
delta1, line 300: good ol' answer to everything 🤣

delta1 (dismissed):
LGTM

philipr-za:
Rebase and deal with a conflict

base_layer/core/src/transactions/transaction.rs
```diff
@@ -223,6 +231,39 @@ impl UnblindedOutput {
...
+        let proof = RangeProof::from_bytes(&proof_bytes)
+            .map_err(|_| TransactionError::RangeProofError(RangeProofError::ProofConstructionError))?;
+
+        let output = TransactionOutput {
+            features: self.features.clone(),
+            commitment,
+            proof,
+        };
+        // A range proof can be constructed for an invalid value so we should confirm that the proof can be verified.
+        if !output.verify_range_proof(&factories.range_proof)? {
```
stringhandler, line 260: Might be worth skipping this in Release mode. Is it not cheaper to verify `self.value` directly or is there another way of creating an invalid rangeproof?

base_layer/core/src/transactions/transaction.rs
```diff
@@ -353,7 +394,38 @@ impl TransactionOutput {
...
     pub fn verify_range_proof(&self, prover: &RangeProofService) -> Result<bool, TransactionError> {
-        Ok(prover.verify(&self.proof.to_vec(), &self.commitment))
+        Ok(prover.verify(&self.proof.0, &self.commitment))
+    }
+
+    /// Attempt to rewind the range proof to reveal the proof message and committed value
+    pub fn rewind_range_proof(
+        &self,
+        prover: &RangeProofService,
+        public_rewind_key: &PublicKey,
```
stringhandler: nit: Perhaps `rewind_public_key` and `rewind_blinding_public_key` is a better name? We don't want to suggest that these values are published somewhere

base_layer/core/src/transactions/transaction.rs
```diff
@@ -353,7 +394,38 @@ impl TransactionOutput {
 
     /// Verify that range proof is valid
     pub fn verify_range_proof(&self, prover: &RangeProofService) -> Result<bool, TransactionError> {
-        Ok(prover.verify(&self.proof.to_vec(), &self.commitment))
+        Ok(prover.verify(&self.proof.0, &self.commitment))
+    }
+
+    /// Attempt to rewind the range proof to reveal the proof message and committed value
+    pub fn rewind_range_proof(
```
stringhandler:
```suggestion
    pub fn rewind_range_proof_value_only(
```
Not sure what a better name would be, but my intention here is to verbalise how this method differs from `full_rewind_range_proof`

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -74,6 +87,10 @@ use tokio::{
 const LOG_TARGET: &str = "wallet::output_manager_service";
 const LOG_TARGET_STRESS: &str = "stress_test::output_manager_service";
 
+const KEY_MANAGER_COINBASE_BRANCH_KEY: &str = "coinbase";
+const KEY_MANAGER_REWIND_BRANCH_KEY: &str = "rewind";
```
```suggestion
const KEY_MANAGER_RECOVERY_VIEWONLY_BRANCH_KEY: &str = "recovery_viewonly";
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -74,6 +87,10 @@ use tokio::{
...
+const KEY_MANAGER_REWIND_BLINDING_BRANCH_KEY: &str = "rewind_blinding";
```
```suggestion
const KEY_MANAGER_RECOVERY_BLINDING_BRANCH_KEY: &str = "recovery_blinding";
```

stringhandler (commented):
Happy with the approach. I think there are some places where the naming could be changed for clarity, specifically using `recovery` instead of `rewind`. This is a subjective view and I don't want to go down the bike shedding path , so I'm not too fussed about it

philipr-za:
> Happy with the approach. I think there are some places where the naming could be changed for clarity, specifically using `recovery` instead of `rewind`. This is a subjective view and I don't want to go down the bike shedding path , so I'm not too fussed about it

My issue with that is that we are using the rewinding for recovery but we can also use the rewinding for other purposes when we put stuff into the proof message so these methods might not only be used for recovery purposes.

stringhandler:
Fair enough

base_layer/core/src/transactions/transaction.rs
philipr-za: 🤷‍♂️  This is from the original implementation of UnblindedOutput. If we are confident that the value in self.value cannot be invalid I guess we can remove this here and in 'as_transaction_output`

PR #2483, opened 2020-12-09
Keep a cached RandomXVM in memory to speed up verification

stringhandler:
## Description

- The main change in this PR is caching a RandomX virtual machine, dataset and cache inside of a RandomXFactory that will only create a new virtual machine for new seed (key). This change speeds up block sync significantly.
- The validator traits have been split into individual traits that describe how and when they are used.
- Added the concept of a `ChainBlock` and `ChainHeader` (credits here go to @SWvheerden) for a header or block that is attached to a chain, as opposed to a true orphan. In these cases, there is the idea of accumulated data that has been calculated from a previous block and accumulated data.

Note that a lot of this PR has been merged from the work by @SWvheerden in multiple other PRs

Note also that while I originally wanted to remove the `target_difficulty` and `accumulated_difficulty` fields from the headers, this would have required a reset. In the next reset, these fields should be removed.

## Motivation and Context

There's a fair amount of change to the validators. I found the generic approach confusing and I thought it might be less so by renaming them to describe the behaviour that they want to achieve. This way an implementer of the trait will know the job that needs to achieve and the place in the pipeline that the validation will be happening.


base_layer/core/src/chain_storage/accumulated_data.rs
```diff
@@ -216,8 +221,111 @@ impl<'de> Visitor<'de> for BlockAccumulatedDataVisitor {
...
+        self.achieved_difficulty = Some(achieved);
+        self
+    }
+
+    pub fn build(self) -> Result<BlockHeaderAccumulatedData, ChainStorageError> {
+        let monero_diff = self
+            .accumulated_monero_difficulty
+            .ok_or_else(|| ChainStorageError::InvalidOperation("difficulty not provided".to_string()))?;
+
+        let blake_diff = self
```
SWvheerden, line 282: We need to remember to rename this.

base_layer/core/src/proto/block.proto
```diff
@@ -67,9 +67,19 @@ message HistoricalBlock {
...
     Block block = 3;
+
+    BlockHeaderAccumulatedData accumulated_data = 4;
 }
 
 
+message BlockHeaderAccumulatedData {
+    uint64 achieved_difficulty = 1;
+    uint64 accumulated_monero_difficulty = 2;
+    uint64 accumulated_blake_difficulty = 3;
```
SWvheerden, line 78:
```suggestion
    uint64 accumulated_sha_difficulty = 3;
```

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -459,9 +464,9 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                 );
                 continue;
             }
-            self.header_validator.validate(&header)?;
-            let current_height = header.height;
-            self.db.write_transaction().insert_header(header).commit().await?;
+            let chain_header = self.header_validator.validate_and_calculate_metadata(header)?;
```
stringhandler, line 466: There's a bit of a name mismatch here. I have been toying with the idea of renaming `*AccumulatedData` to `*Metadata`. Thoughts?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -590,10 +493,33 @@ where B: BlockchainBackend
         fetch_header_by_block_hash(&*db, hash)
     }
 
+
+    /// Returns a connected header in the main chain by block hahs
```
```suggestion
    /// Returns a connected header in the main chain by block hash
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1524,9 +1477,9 @@ fn fetch_block_by_hash<T: BlockchainBackend>(
     if let Some(header) = fetch_header_by_block_hash(db, hash.clone())? {
         return Ok(Some(fetch_block(db, header.height)?));
     }
-    if let Ok(block) = fetch_orphan(db, hash) {
-        return Ok(Some(HistoricalBlock::new(block, 0)));
-    }
+    // if let Ok(block) = fetch_orphan(db, hash) {
```
stringhandler: Note here that blocks from the orphan pool are no longer returned

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -973,6 +1054,29 @@ impl BlockchainBackend for LMDBDatabase {
         })
     }
 
+    fn fetch_tip_header(&self) -> Result<ChainHeader, ChainStorageError> {
```
stringhandler, line 1052: Note this change. Previously would return the last header, now returns the header at tip according to metadata

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/state_sync.rs
```diff
@@ -43,7 +43,7 @@ use tari_crypto::tari_utilities::Hashable;
 
 const LOG_TARGET: &str = "c::bn::state_machine_service::states::horizon_state_sync";
 
-pub struct HorizonStateSynchronization<'a, B> {
+pub struct HorizonStateSynchronization<'a, B: BlockchainBackend> {
```
sdbondi, line 46: Trait bounds should (almost) never be needed on struct/enum definitions and are [generally undesirable](https://rust-lang.github.io/api-guidelines/future-proofing.html#data-structures-do-not-duplicate-derived-trait-bounds-c-struct-bounds). BlockchainBackend is added here because of the unnecessary trait bound on the `CandidateBlockBodyValidation` trait. I suggest removing the trait bound from `CandidateBlockBodyValidation`, `SyncValidators`, `BaseNodeStateMachine` and here.

sdbondi (commented):
Looks good - tested header sync and the speed up is as expected (i.e. ZOMG MUCH BETTER). 

Some issues that I encountered that need investigating:

Node that is behind is not initiating a sync because the remote acc diff is tiny:
```
We're at block 15977 with an accumulated difficulty of 1609903318819586512336656 and the network chain tip is at 16032 with an accumulated difficulty of 16
```

The node that is being synced FROM returns an RPC internal error (is there a db breaking change?) on the `sync_headers` call
```
 [c::base_node::sync_rpc] ERROR Internal error: The requested BlockHeaderAccumulatedData was not found via height:15977 in the database
```

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: I think `Metadata` is a little clearer

base_layer/core/src/proof_of_work/randomx_factory.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        result
+    }
+}
+
+#[derive(Clone)]
+pub struct RandomXVMInstance {
+    // Note: If the cache and dataset drops, the vm will be wonky, so have to store all
+    // three for now
+    instance: Arc<Mutex<(RandomXVM, RandomXCache, RandomXDataset)>>,
```
sdbondi, line 30: This Mutex shouldn't be needed, since no state is ever mutated.

```suggestion
    instance: Arc<(RandomXVM, RandomXCache, RandomXDataset)>,
```

stringhandler: I'm not 100% sure about that. The cache and the dataset don't change, but I'm not sure if the VM changes bits in FFI land. I would need to do a proper review of the `randomx_rs` crate. There are a few improvements that can be made there, e.g. pairing the lifetimes of the cache and dataset to the VM, which is not done right now. If the cache or dataset drop in rust, the memory will be released and the vm will continue happily.

I think the mutex is safer for now, and we can look at it again when I get some time to look at `randomx_rs`

sdbondi: I suppose it is possible that unsafe could have been abused to allow unsynchronized mutation - I did test it before but let's stick with the safer approach.

sdbondi (dismissed):
Tested block sync with a few nodes and this is a big speed up for header sync - Great stuff! 

Just noting here that this is a breaking change to blockchain db.

PR #2484, opened 2020-12-09
Exchange chain metadata with all connected nodes

sdbondi:
Include all connected nodes in ping rounds. This will reduce the time to
recognise a node advertising higher accumulated PoW in their chain
metadata.

This PR is up for discussion and is mainly to show where to make the change
should the community decide it is a good idea.

### DIFF
--- a/base_layer/p2p/src/services/liveness/service.rs
+++ b/base_layer/p2p/src/services/liveness/service.rs
@@ -245,10 +245,7 @@ where
     async fn start_ping_round(&mut self) -> Result<(), LivenessError> {
         let selected_peers = self
             .connectivity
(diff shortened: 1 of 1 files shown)


stringhandler:
I tested this, but it didn't seem to make a noticeable difference. I think it might be worth keeping this PR around and testing it when the network would benefit from it

SWvheerden:
We added that delay to entering sync, While in theory,  this code should notify you if you are behind. You will still be default wait till you are 6 blocks behind before you try to enter sync.

sdbondi:
I think we can close because I dont think it adds anything useful - feel free to reopen if you disagree

PR #2486, opened 2020-12-09
Add wallet build step to catch regressions

delta1:
## Description

The libwallet build requires that the `tari_wallet` crate can be built on independently of the base node. This commit adds `cargo build -p tari_wallet` to the circle ci build step to ensure that future commits don't break this requirement.

As a part of this, the `waiting_requests` module was extracted to tari_common_types since it is used in both the wallet and
base node (and was previously duplicated in both).

## Motivation and Context

We don't have visibility of a broken libwallet build until it is built on a new tag. 

## How Has This Been Tested?

- `cargo build`
- `cargo build -p tari_wallet`
- base node built and tested manually
- console wallet built and tested manually 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


stringhandler (commented):
I think this can also be achieved by building in the `wallet_ffi` folder. I originally had a PR that did a similar thing and decided to abandon it because build times were so long. I think we can probably put this in though.

Another option to speed up build times is running `cargo clippy` which is much faster and fails on compile errors and warnings. Perhaps trying that in the `wallet_ffi` folder might be an option?

delta1:
So I see this as an easy way to catch the issues with feature flags which go unnoticed, without trying to do all the actual libwallet building. This specific changes only adds ~2 mins to the build time. 

Screenshot 2020-12-09 at 12 40 36

delta1:
I think once the github actions build for libwallet is working again, it might be worth enabling it on push/pr ? But as you say that build process is very long.

stringhandler (approved):
Happy

PR #2488, opened 2020-12-09
Only clean orphans after syncing

stringhandler:
Instead of checking and deleting orphans on every block add, this changes the logic to only run clean up after batches have completed.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -799,14 +799,6 @@ where B: BlockchainBackend
             block,
         )?;
 
-        // Cleanup orphan block pool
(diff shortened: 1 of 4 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -827,6 +819,13 @@ where B: BlockchainBackend
         Ok(block_add_result)
     }
 
+    /// Clean out the entire orphan pool
```
sdbondi, line 821: Maybe something like: `/// Deletes older orphan entries as needed, ensuring that the number of stored orphan blocks never exceeds the configured orphan_storage_capacity`

sdbondi (dismissed):
LGTM

PR #2492, opened 2020-12-10
Implement Base Node Wallet RPC service

philipr-za:
## Description

Implement a Wallet RPC service in the Base Node

This PR implements an RPC service in the base node for the wallet to use when submitting and querying transactions. The RPC service provides two calls: 
`submit_transaction(tx)`: Will be used by a wallet to submit a transaction to the Base Node and the response informs the wallet if the tx was accepted into the mempool or it was rejected and why
`transaction_query(excess_sig)`: Lets the wallet request the status of a transaction via its excess_sig. The response will tell the wallet whether the transaction is Not Stored, in the Unconfirmed Pool or if it is Mined it will also return the block header and number of confirmations.

## Motivation

This RPC service represents the new interface that a wallet will use to interact with base nodes while trying to broadcast a transaction. This will replace the current fire-and-forget type messaging provided by the p2p comms layer which makes the logic for receiving the response to the query tricky.

## How Has This Been Tested?
The `test_base_node_wallet_rpc()` is provided that tests both the RPC calls using a test blockchain and mempool back end.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/rpc/service.rs
+++ b/base_layer/core/src/base_node/rpc/service.rs
@@ -1,0 +1,194 @@
+//  Copyright 2020, The Tari Project
+//
+//  Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+//  following conditions are met:
+//
+//  1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+//  disclaimer.
+//
+//  2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
+//  following disclaimer in the documentation and/or other materials provided with the distribution.
+//
+//  3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
+//  products derived from this software without specific prior written permission.
+//
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+//  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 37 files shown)


delta1:
This looks great, but `build wallet` failed in circle ci and it seems there are some test compilation failures too

philipr-za:
Urgh, those feature flags in core are a pain in the ass.

philipr-za:
woops

base_layer/core/src/base_node/rpc/service.rs
```diff
@@ -0,0 +1,194 @@
...
+            .await
+            .map_err(RpcStatus::log_internal_error(LOG_TARGET))?
+        {
+            TxStorageResponse::UnconfirmedPool => TxQueryResponse {
+                location: TxLocation::InMempool as i32,
+                block_hash: None,
+                confirmations: 0,
+            },
+            TxStorageResponse::ReorgPool |
+            TxStorageResponse::NotStoredOrphan |
```
stringhandler, line 182: Is it possible for the mempool to ever return these? Surely the mempool will not know the reason something is not stored, because it has not stored it

stringhandler (approved):
Looks good. I'd say that TxStorageResponse should not be reused for querying, and only used for submitting.

base_layer/core/src/base_node/rpc/service.rs
philipr-za: The validator tells it what the issue was, that's how it used to decide which pool to put it in.

philipr-za: oh on the query you are right its not possible

philipr-za: I agree on this change and I got started but turns out that the same struct is used for the service handle interface that is used by the proto of the mempool service so making an update to way the handle returns involves updating everything in the repo that touches the proto too which is extensive.  Should I do it in this PR?

PR #2494, opened 2020-12-11
Add console wallet passphrase encryption

delta1:
## Description

- When the console wallet starts up, check if per field encryption has been applied. If not, prompt the user for a password and confirmation, and then encrypt wallet. Subsequently the password must be provided to start and use the console wallet. 
- Allow the password to be provided by prompt, command line argument, config file, or environment variable (TARI_WALLET_PASSWORD) for automation purposes.
- Change the wallet password by passing `--change-password` at the command line
- Refactor console wallet initialization functions from `main.rs` to their own `init` module to clean up main.
- Remove emojis from base node status element.
- Fix bug where setting a custom base node did not update the base node detail element

## Motivation and Context

Security 

## How Has This Been Tested?

Manually in console wallet 

## Types of changes

* [ ] New Tests
* [x] Documentation

## Checklist:


applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+pub fn get_or_prompt_password(
+    arg_password: Option<String>,
+    config_password: Option<String>,
+) -> Result<Option<String>, ExitCodes>
+{
+    if arg_password.is_some() {
+        return Ok(arg_password);
+    }
+
+    let env = std::env::var_os(TARI_WALLET_PASSWORD);
```
stringhandler, line 62: This is not needed, I think the env var `TARI_WALLET__RIDCULLY__PASSWORD` should already handle it

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+    if let Some(p) = env {
+        let env_password = Some(
+            p.into_string()
+                .map_err(|_| ExitCodes::IOError("Failed to convert OsString into String".to_string()))?,
```
stringhandler, line 66: ConfigError maybe? This seems to be the only use of IOError and it's related to config

stringhandler (commented):
Happy in general. Will test it locally

applications/tari_console_wallet/src/init/mod.rs
delta1: Good point, it’s an extremely unlikely error and I’m not even sure what would cause it. But it’s not config, so maybe UnknownError?

applications/tari_console_wallet/src/init/mod.rs
delta1: Oh? I’ll take a look for that

stringhandler (approved):
Tested locally and I'm happy.

PR #2496, opened 2020-12-14
Remove Pending and Orphan pool from Mempool

philipr-za:
## Description
This PR removes the Pending and Orphan pools from the mempool. It has been decided that if a wallet submits a block that is time locked (would end up in Pending pool), has an input that is not on the chain yet (is an orphan tx) that these transactions should be rejected by the mempool outright. Furthermore, it was decided that if a submitted transaction would end up in the Reorg pool directly it is also rejected. A transaction can still move from the Unconfirmed pool to the Reorg pool but only during a chain reorg event.

## Motivation and Context
This simplifies how a Base Node handles submitted transactions significantly and thus puts the onus on the wallet to managed the timelocked and orphan cases themselves. Over all this simplifies the interaction between the wallet and base node.

## How Has This Been Tested?
Existing mempool test should cover this change with the updates made in this PR.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


philipr-za:
In terms of test I believe the following three tests are the most critical:
`test_insert_and_process_published_block`
`test_orphaned_mempool_transactions`
`block_event_and_reorg_event_handling`

They currently fail due to this helper function being broken: `generate_new_block(…)`

Though I believe all the base node and mempool tests will test this change.

stringhandler:
Tested and working

PR #2504, opened 2020-12-18
Fix resend bug in transaction send protocol

philipr-za:
## Description
When sending a transaction if the initial send fails to send both directly and via SAF the transaction is cancelled and the funds unencumbered.

We later added the resending functionality to this protocol and the bug that crept in was that if on a periodic resend both direct and SAF sending failed then the outputs would be unencumbered but the protocol would continue to wait for a reply instead of fully exiting.

This PR moves the cancellation to only happen on the initial send, this is so that if you try send a Tx on a device with no connectivity it will immediately fail and free the funds. The auto cancellation is removed from the repeat sends which will require the user to manually cancel.

This PR also includes some fixes to wallet tests that broke after we updated the Transaction Negotiation protocols to send both SAF and Direct every time.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
@@ -209,6 +209,18 @@ where TBackend: TransactionBackend + 'static
             )));
 
         if !direct_send_result && !store_and_forward_send_result {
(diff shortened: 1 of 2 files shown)


stringhandler (approved):
Looks good

PR #2505, opened 2020-12-28
Base node displays status every 30 seconds

stringhandler:
Base node now prints out a general status message every 30 seconds. This can also be achieved via the `status` command. 

Also included was a refactor of the command execution from `Parser` into its own class `CommandHandler` to help with the futures and `mut`s that are required for the `select!`  in `cli_loop`

### DIFF
--- a/applications/tari_base_node/src/command_handler.rs
+++ b/applications/tari_base_node/src/command_handler.rs
@@ -1,0 +1,1908 @@
+// Copyright 2020. The Tari Project
+//
+// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+// following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+// disclaimer.
+//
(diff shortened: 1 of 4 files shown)


delta1 (approved):
LGTM - is there a way for the “listening status” to have the sync/tip info too? Would be nice to have

stringhandler:
It shows the last block in the db which is pretty much the only data that's interesting in Listening status
`State: Listening, Tip: 18043 (Mon, 28 Dec 2020 16:17:09 +0000),`

delta1:
Oh I missed that, nice!

CjS77:
At some point the complexity of the command parser may benefit shifting to a purpose-written crate, e.g.

https://docs.rs/indicatif/0.15.0/indicatif/
and /or
https://github.com/phsym/shrust

But this is a step in the right direction so +1

CjS77 (approved):
LGTM

stringhandler:
Those look amazing. Definitely keen to try them out at some point

PR #2506, opened 2021-01-04
Fix to use last header instead of tip

stringhandler:
## Description

In a previous  commit, the `fetch_tip_header`  was  changed  to return the header at tip according to the `ChainMetadata`. `fetch_last_header`  should  be used when  trying to retrieve the  newest header that was added.


CjS77 (dismissed):
LGTM - but rustmft failure

PR #2509, opened 2021-01-04
Add TXO validation to console wallet

delta1:
## Description

- Console wallet now validates all Unspent, Spent, and Invalid TXOs once wallet is started
- `OutputManagerEventSender` and `OutputManagerEventReceiver` updated to wrap `OutputManagerEvent` in  `Arc<>` to match Transaction and Connectivity service event streams. 
- Console wallet refreshes balance correctly once TXOs are validated 
- Fixed encryption test in `tari_wallet_ffi` broken by #2494   

## Motivation and Context

An important feature that has already been implemented in the base node wallet 

## How Has This Been Tested?

- manually edited wallet sqlite outputs to alter their statuses, the console wallet then validated and updated to correct respective statuses 
- `cargo test -p tari_wallet`
- `cargo test -p tari_wallet_ffi` 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/init/mod.rs
+++ b/applications/tari_console_wallet/src/init/mod.rs
@@ -37,7 +37,10 @@ use tari_shutdown::ShutdownSignal;
 use tari_wallet::{
     base_node_service::config::BaseNodeServiceConfig,
     error::{WalletError, WalletStorageError},
-    output_manager_service::config::OutputManagerServiceConfig,
+    output_manager_service::{
+        config::OutputManagerServiceConfig,
+        protocols::txo_validation_protocol::{TxoValidationRetry, TxoValidationType},
+    },
     storage::sqlite_utilities::initialize_sqlite_database_backends,
     transaction_service::config::TransactionServiceConfig,
     wallet::WalletConfig,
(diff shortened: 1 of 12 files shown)


applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -340,5 +343,37 @@ pub async fn start_wallet(wallet: &mut WalletSqlite, base_node: &Peer) -> Result
         error!(target: LOG_TARGET, "Problem restarting transaction protocols: {}", e);
     }
 
+    debug!(target: LOG_TARGET, "Starting TXO validations.");
```
hansieodendaal, line 346: These 
- `validate_txos(TxoValidationType::Unspent`
- `validate_txos(TxoValidationType::Spent`
- `validate_txos(TxoValidationType::Invalid`

should be verified to be completed successfully and go before `Restart transaction protocols` are started, similar to the base node wallet logic
- `wallet.transaction_service.restart_transaction_protocols().await`
- `wallet.transaction_service.restart_broadcast_protocols().await`

We want to ensure transaction and broadcast protocols only use a fully validated TXO set, thus they should be delayed until after the prior validation. See `async fn monitor_validation_protocol(`

hansieodendaal (commented):
Hi @delta1, just some comments on the logic around TXO validation and restarting of transaction and broadcast protocols.

applications/tari_console_wallet/src/init/mod.rs
delta1: Thanks for the feedback! Will do.

hansieodendaal: Also, please implement the higher level initial sync logic in `async fn start_transaction_protocols_and_txo_validation(`, i.e. "Only start the transaction broadcast and TXO validation protocols once the ~~local~~ _connected base_ node is synced".

base_layer/wallet/tests/output_manager_service/service.rs
```diff
@@ -885,10 +885,15 @@ fn test_utxo_validation() {
         loop {
             futures::select! {
                 event = event_stream.select_next_some() => {
-                    match event.unwrap() {
-                        OutputManagerEvent::TxoValidationTimedOut(_) => {
-                            timeouts+=1;
-                         },
+                    match event {
```
philipr-za, line 888: So this doesn't need to change or be reverted but just a philosophical comment for the future. The removal of the `unwrap()` in the `tari_base_node::transaction_monitor` definitely needed to happen but in tests it is, in my opinion, preferable to leave an unwrap rather than gracefully deal with error like you have done here so that the test panics if there is a problem. In a test we prefer to see the panics explicitly.

applications/tari_console_wallet/src/init/mod.rs
philipr-za: I agree with Hansie on this, I think the validations and restarting of the *broadcast* protocols should be done in an asynchronous task. However, I think that the transaction negotiation protocols can be restarted without waiting for the validation to be completed. These protocols are between wallets and not between the base node and this wallet. We want wallet-to-wallet negotiation to occur even if the base node is unreachable. If there is a problem with the outputs in a given transaction being negotiated between wallets it will be picked up when they finish their negotiation and try broadcast.

philipr-za (commented):
Looking good, I have a couple comments and generally agree with Hansie's comment.

delta1:
closing for #2597

PR #2513, opened 2021-01-06
Application: mining node

dunnock:
## Description
This PR adds mining node application. Mining node has minimum dependency on tari libraries as to make it good example for portability (so far it is still dependent on tari_core BlockHeader and ProofOfWork math), further decoupling might be implemented in follow up PRs. To get the data necessary for mining node requires grpc connection to tari base node and console wallet.

Mining is performed via stages:
1. Validate if base node is bootstrapped
2. Retrieve block template from base node
3. Retrieve coinbase from console wallet
4. Retrieve new block from base node based on template enriched with coinbase output and kernel
5. Mine for block with matching difficulty
6. Submit mined block to base node and restart from 1

When started locally mining node will try to get settings from `[mining_node]` section, including `base_node_grpc_address` and `wallet_grpc_address`. If grpc addresses not set it will try to load from relevant sections from a global config, so when both wallet and node are runnign locally it can start with zero-configuration. By default if `num_mining_threads` setting is not set it will use number of logical CPU cores. It also supports network section overrides.

In addition PR also fixes ConfigLoader trait to use `network` setting (it was still using `use_network` which I think was changed some time ago). Also changes some usize numbers wrapping for current CpuMiner implementation to rust native.

Also there is a little improvement in initialization of applications, as all applications seems be doing same - initialize bootstrap, initialize logs and load global config moved this part to one library function for base node, wallet and mining node.

I've also started moving sha3 calculation into mining node, though left that not completed as it is being changed now.

## Motivation and Context
This application would give example of how to write a miner node for tari, which might spike more opensource miners implementation.

## How Has This Been Tested?
I have run mining node in local network with 2 nodes, one of nodes was mining too. Fixed tests for config changes for config loaders.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_utilities/src/initialization.rs
+++ b/applications/tari_app_utilities/src/initialization.rs
@@ -1,0 +1,29 @@
+use crate::utilities::ExitCodes;
+use config::Config;
+use structopt::StructOpt;
+use tari_common::{configuration::bootstrap::ApplicationType, ConfigBootstrap, GlobalConfig};
+
+pub const LOG_TARGET: &str = "tari::application";
+
+pub fn init_configuration(
+    application_type: ApplicationType,
+) -> Result<(ConfigBootstrap, GlobalConfig, Config), ExitCodes> {
+    // Parse and validate command-line arguments
+    let mut bootstrap = ConfigBootstrap::from_args();
+
+    // Check and initialize configuration files
+    bootstrap.init_dirs(application_type)?;
+
+    // Load and apply configuration file
+    let cfg = bootstrap.load_configuration()?;
+
+    // Initialise the logger
+    bootstrap.initialize_logging()?;
+
+    // Populate the configuration struct
+    let global_config = GlobalConfig::convert_from(cfg.clone()).map_err(|err| {
+        log::error!(target: LOG_TARGET, "The configuration file has an error. {}", err);
+        ExitCodes::ConfigError(format!("The configuration file has an error. {}", err))
+    })?;
+    Ok((bootstrap, global_config, cfg))
+}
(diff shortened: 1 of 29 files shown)


applications/tari_app_utilities/src/initialization.rs
```diff
@@ -0,0 +1,29 @@
+use crate::utilities::ExitCodes;
```
delta1, line 1: new files should have the copyright preamble I think

applications/tari_mining_node/src/config.rs
```diff
@@ -0,0 +1,78 @@
+//! Miner specific configuration
```
delta1, line 23: copyright

applications/tari_mining_node/src/difficulty.rs
```diff
@@ -0,0 +1,84 @@
+use sha3::{Digest, Sha3_256};
```
delta1, line 23: copyright

delta1 (changes requested):
Looking good @dunnock ! 

A few things, I know this is wip: 

- new files need the copyright/licence preamble at the top 
- tari_mining_node crate needs a readme with some instructions 
- maybe `use log::*` to bring log functions into scope instead of having to them all fully qualified 
- config options should be listed in the config samples as well (eg: windows.toml) 

I'm still trying to get things working enough to actually test the mining! But that's due to network and node syncing issues 

Personally I think that the crate could be named `tari_miner` - what do you think @mikethetike ?

dunnock:
Thank you for feedback @delta1 , added copyrights, README and config examples. Also imported `log::` stuff. Retained from renaming until Mike would confirm.

dunnock:
I'm also waiting for https://github.com/tari-project/tari/pull/2514 to land to accomodare grpc changes in this PR.

stringhandler (commented):
Looks good, will test locally

PR #2526, opened 2021-01-12
Add fee estimate method to output manager service and wallet_ffi

delta1:
## Description

Adds a `fee_estimate` method to the output manager service and also to the wallet_ffi 

## Motivation and Context

This will allow the mobile app to show accurate fees for the transaction 

## How Has This Been Tested?

- `cargo test -p tari_wallet`
- `cargo test -p tari_wallet_ffi`

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/handle.rs
+++ b/base_layer/wallet/src/output_manager_service/handle.rs
@@ -65,6 +65,7 @@ pub enum OutputManagerRequest {
     ApplyEncryption(Box<Aes256Gcm>),
     RemoveEncryption,
     GetPublicRewindKeys,
+    FeeEstimate((MicroTari, MicroTari, u64, u64)),
 }
 
 impl fmt::Display for OutputManagerRequest {
@@ -92,6 +93,7 @@ impl fmt::Display for OutputManagerRequest {
             OutputManagerRequest::RemoveEncryption => f.write_str("RemoveEncryption"),
             OutputManagerRequest::GetCoinbaseTransaction(_) => f.write_str("GetCoinbaseTransaction"),
             OutputManagerRequest::GetPublicRewindKeys => f.write_str("GetPublicRewindKeys"),
+            OutputManagerRequest::FeeEstimate(_) => f.write_str("FeeEstimate"),
         }
     }
 }
@@ -119,6 +121,7 @@ pub enum OutputManagerResponse {
     EncryptionApplied,
     EncryptionRemoved,
     PublicRewindKeys(Box<PublicRewindKeys>),
+    FeeEstimate(MicroTari),
 }
 
 pub type OutputManagerEventSender = broadcast::Sender<OutputManagerEvent>;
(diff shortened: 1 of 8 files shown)


base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -131,6 +134,7 @@ pub enum OutputManagerEvent {
     TxoValidationSuccess(u64),
     TxoValidationFailure(u64),
     TxoValidationAborted(u64),
+    FeeEstimate(MicroTari),
```
philipr-za: I don't think this is the appropriate way to return this value. The event bus is used to return feedback for long running async tasks. These are typically operations that require communications to external nodes.

This operation only requires a single DB read to selection the outputs so I think this should be returned directly by the Output Manager Service api call.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -236,6 +240,31 @@ impl OutputManagerHandle {
...
 
+    /// Get a fee estimate for an amount of MicroTari, at a specified fee per gram and given number of kernels and
+    /// outputs
+    pub async fn fee_estimate(
+        &mut self,
+        amount: MicroTari,
+        fee_per_gram: MicroTari,
+        num_kernels: u64,
+        num_outputs: u64,
+    ) -> Result<MicroTari, OutputManagerError>
```
philipr-za, line 250: So here I see it is returned by API call in the handle so then I am not sure why you need to add a variant to the event enum.

base_layer/wallet/tests/output_manager_service/service.rs
```diff
@@ -275,6 +275,51 @@ fn sending_transaction_and_confirmation<T: Clone + OutputManagerBackend + 'stati
...
+    // minimum fee
+    let fee_per_gram = MicroTari::from(1);
+    let fee = runtime
+        .block_on(oms.fee_estimate(MicroTari::from(100), fee_per_gram, 1, 1))
+        .unwrap();
+    assert_eq!(fee, MicroTari::from(100));
+
+    let fee_per_gram = MicroTari::from(25);
+    let fee = runtime
+        .block_on(oms.fee_estimate(MicroTari::from(100), fee_per_gram, 1, 1))
```
philipr-za: I think it's important to test this while varying the number of outputs in a transaction.

base_layer/wallet/src/output_manager_service/handle.rs
delta1: My mistake

base_layer/wallet/tests/output_manager_service/service.rs
delta1: will do

base_layer/wallet/src/output_manager_service/handle.rs
delta1: sorry this was mistakenly included

philipr-za (approved):
🎉

base_layer/wallet_ffi/wallet.h
```diff
@@ -461,6 +461,9 @@ unsigned long long wallet_get_pending_incoming_balance(struct TariWallet *wallet
 // Gets the outgoing balance from a TariWallet
 unsigned long long wallet_get_pending_outgoing_balance(struct TariWallet *wallet,int* error_out);
 
+// Get a fee estimate from a TariWallet for a given amount
+unsigned long long wallet_get_fee_estimate(struct TariWallet *wallet, unsigned long long amount, unsigned long long fee_per_gram, unsigned long long num_kernels, unsigned long long num_outputs, int* error_out);
```
StriderDM, line 465: @delta1 When this merges, this method still needs to be added to the wrappers for iOS and Android. Just mentioning this here.

delta1: thanks @StriderDM agreed

PR #2528, opened 2021-01-13
[Tari Mining Node] Add estimated time to block to logs

stringhandler:
## Description

Add estimated time to find a block to logs. This is not accurate since it is probabilistic, but will give the user an idea of how long it would take to solo mine a block at the current network hash rate.

Example:
```
2021-01-13 09:23:58.199382000 [tari_mining_node] DEBUG Miner 3 reported 0.22MH/s with total 1.34MH/s over 6 threads. Height: 20940. Target: 423872797, Estimated block in approx. 4m14s (+/- Ave. 315s)
```

## Motivation and Context

If I download the miner and start mining, I can get an idea of the performance of my machine vs the current network hash rate by comparing the target block time to the time it would take to mine a block. E.g. if it would take 20 mins to mine a SHA block, I know there are around 4 other miners at a similar hash rate.  

## How Has This Been Tested?

Tested by running locally

### DIFF
--- a/applications/tari_mining_node/src/main.rs
+++ b/applications/tari_mining_node/src/main.rs
@@ -39,6 +39,7 @@ mod utils;
 
 use errors::{err_block_header, err_empty, MinerError};
 use miner::Miner;
+use std::time::{Duration, Instant};
 
 /// Application entry point
 fn main() {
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
LGTM

PR #2530, opened 2021-01-13
Use parity-multiaddr github lib with backported build fix

sdbondi:
Use a forked version of parity-multiaddr that fixes a build failure
caused by a mistaken import of a serde re-export that has become
private in serde 1.0.119.

Ref https://github.com/tari-project/rust-libp2p/pull/1


sdbondi:
Closing this in favour of work being done here: https://github.com/delta1/tari/tree/quay3 @delta1

PR #2535, opened 2021-01-14
Remove incorrect error response in blockchain rewind

sdbondi:
## Description

`rewind_to_height` only rewind block bodies _if they exist_.
This means that calling `rewind_to_height(h)` where `h` is greater than the current
_block_ height is valid.

This fixes a continuous failure of header sync, causing nodes to get
stuck at a particular block height.

Also fixed 2 unrelated tari_core tests

## Motivation and Context

Reports of seed nodes being stuck at a height

## How Has This Been Tested?

Existing tests pass, tested syncing from scratch

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1582,12 +1582,12 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
     // Delete blocks
     let metadata = db.fetch_chain_metadata()?;
     let block_height = metadata.height_of_longest_chain();
-    let steps_back = block_height.checked_sub(height).ok_or_else(|| {
-        ChainStorageError::InvalidQuery(format!(
-            "Cannot rewind to height ({}) that is greater than the block height {}.",
-            height, block_height
-        ))
-    })?;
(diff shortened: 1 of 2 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1582,12 +1582,7 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
     // Delete blocks
     let metadata = db.fetch_chain_metadata()?;
     let block_height = metadata.height_of_longest_chain();
-    let steps_back = block_height.checked_sub(height).ok_or_else(|| {
-        ChainStorageError::InvalidQuery(format!(
-            "Cannot rewind to height ({}) that is greater than the block height {}.",
-            height, block_height
-        ))
-    })?;
+    let steps_back = block_height.saturating_sub(height);
```
stringhandler, line 1585: I think it needs to return here, otherwise it will set the chain metadata to the tip at line 1608, which may be higher than the current tip

sdbondi: Ah you're right - fixed

stringhandler (approved):

PR #2539, opened 2021-01-17
Boost mining speeds

dunnock:
## Description
This PR improves miner application:
- Improved early mining restart logic: when mine_on_tip_only option is set miner will exit if node is reporting that current height is already taken by another block (tip).
- Improved mining restart logic, to start new threads in parallel with closing down previous threads.
- Implement own sha3 hashing logic, which caches hash product before nonce, accounting for timestamp changes too. Gives boost on mining +50%.

## Motivation and Context
This should improve mining speeds as well as lower down wasted mining power.

## How Has This Been Tested?
Run locally with 2 disconnected nodes and a wallet.

## Types of changes

* [ ] New Tests
* [X] Documentation

## Checklist:

* [X] My change requires a change to the documentation.
* [X] I have updated the documentation accordingly.

### DIFF
--- a/common/config/presets/tari_sample.toml
+++ b/common/config/presets/tari_sample.toml
@@ -30,15 +30,3 @@ monerod_use_auth=false
 monerod_username=""
 monerod_password=""
 
-[mining_node]
-## Number of mining threads - will use number of logica CPU cores if not set
-#num_mining_threads=8
-
-## GRPC address of base node. Will use local setting from the [base_node] config section by default.
-#base_node_grpc_address = "127.0.0.1:18142"
-
-## GRPC address of console wallet. Will use local setting from the [base_node] config section by default.
-#wallet_grpc_address = "127.0.0.1:18143"
-
-## Will start mining only when base node has reported that it is bootstrapped
-#mine_on_tip_only=true
--- a/common/config/presets/windows.toml
+++ b/common/config/presets/windows.toml
@@ -389,14 +389,25 @@ monerod_password=""
 
 
 [mining_node]
-## Number of mining threads - will use number of logica CPU cores if not set
+# Number of mining threads
+# Default: number of logical CPU cores
 #num_mining_threads=8
 
-# GRPC address of base node. Will use local setting from the [base_node] config section by default.
+# GRPC address of base node
+# Default: value from `base_node.grpc_base_node_address`
 #base_node_grpc_address = "127.0.0.1:18142"
 
-# GRPC address of console wallet. Will use local setting from the [base_node] config section by default.
+# GRPC address of console wallet
+# Default: value from `base_node.grpc_console_wallet_address`
 #wallet_grpc_address = "127.0.0.1:18143"
 
-## Will start mining only when base node has reported that it is bootstrapped
+# Start mining only when base node is bootstrapped 
+# and current block height is on the tip of network
+# Default: true
 #mine_on_tip_only=true
+
+# Will check tip with node every N seconds and restart mining
+# if height already taken and option `mine_on_tip_only` is set 
+# to true
+# Default: 30 seconds
+#validate_tip_timeout_sec=30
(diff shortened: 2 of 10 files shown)


common/config/presets/tari_sample.toml
```diff
@@ -42,3 +42,7 @@ monerod_password=""
 
 ## Will start mining only when base node has reported that it is bootstrapped
 #mine_on_tip_only=true
+
```
hansieodendaal: This config file is intended to be a minimal set of options for runtime.
- Currently, none of the settings has any description, propose to remove the descriptions in this file.
- The minimal settings for the standalone miner to run should not be commented out.
- Maybe 8 threads are a bit too much for a default?

common/config/presets/windows.toml
```diff
@@ -400,3 +400,7 @@ monerod_password=""
 
 ## Will start mining only when base node has reported that it is bootstrapped
 #mine_on_tip_only=true
+
```
hansieodendaal, line 408: The minimal settings for the standalone miner to run should not be commented out.

common/config/presets/windows.toml
```diff
@@ -400,3 +400,7 @@ monerod_password=""
...
+## Will check tip with node every N seconds and restart mining if height already taken and
+## option `mine_on_tip_only` is set to true
+#validate_tip_timeout_sec=30
```
hansieodendaal, line 413: Please add all `[mining_node]` config setting to `tari_config_example.toml` as well, with the minimal settings not commented out.

hansieodendaal (commented):
Please see my comments about the config files. I know it is a bit confusing, and on our TODO list to fix/consolidate the config files. This is the current state:
- `tari_sample.toml` has minimal settings for all Tari executables, but is not really used in our official releases
- `windows.toml` is used to create the Windows installer, and apart from the path separators identical to `tari_config_example.toml`
- `tari_config_example.toml` is used to create the OSX and Ubuntu installation tarbals, and apart from the path separators identical to `windows.toml`

common/config/presets/windows.toml
dunnock: Actually node will start even when no configuration is specified, as it will use defaults.

common/config/presets/tari_sample.toml
dunnock: I see, thanks, will remove settings. Actually mining node will start with no configuration (e.g. zero-configuration), by default it will use  `base_node.ridcully.grpc_base_node_address` to connect, so I guess can remove whole section from here. Added them here as it is clear how to connect to node running on another machine. 

After your last comment I now see that I used wrong file, will move options to https://github.com/tari-project/tari/blob/development/common/config/tari_config_example.toml

Regarding mining threads - as this is standalone mining node I suppose by default what would users want is to mine using all the possible CPU power, so it will use that if no options specified (e.g. it looks up number of logical CPU cores on user machine). `num_mining_threads` option allows to override that, e.g. if someone wants to mine just on one thread etc..

common/config/presets/windows.toml
dunnock: Cool, that's the file I was missing, will add there thanks

common/config/presets/tari_sample.toml
dunnock: Applied requested changes, keeping this comment open as it has open question about mining threads

common/config/presets/windows.toml
dunnock: Done

dunnock:
> Please see my comments about the config files. I know it is a bit confusing, and on our TODO list to fix/consolidate the config files. This is the current state:
> 
> * `tari_sample.toml` has minimal settings for all Tari executables, but is not really used in our official releases
> * `windows.toml` is used to create the Windows installer, and apart from the path separators identical to `tari_config_example.toml`
> * `tari_config_example.toml` is used to create the OSX and Ubuntu installation tarbals, and apart from the path separators identical to `windows.toml`

Cool, thank you, this is clear now, I definitely put settings in a wrong file. Should be fixed now

stringhandler (approved):
Tested and working great!

PR #2540, opened 2021-01-17
Add wallet GRPC `transfer` method

sdbondi:
## Description

Add GRPC method called `transfer` that initiates many transactions to recipients.
Once all recipient addresses have been checked, a transaction is created per recipient entry.

## Motivation and Context

An initial implementation of a function that will be useful for pool payouts.

## How Has This Been Tested?

Not tested - integration tests in another PR

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -31,6 +31,30 @@ import "google/protobuf/timestamp.proto";
 service Wallet {
     // This returns a coinbase transaction
     rpc GetCoinbase (GetCoinbaseRequest) returns (GetCoinbaseResponse);
+    // Send Tari to a number of recipients
+    rpc Transfer (TransferRequest)  returns (TransferResponse);
+}
+
+message TransferRequest {
+    repeated PaymentRecipient recipients = 1;
+}
+
+message PaymentRecipient {
+    string address = 1;
+    uint64 amount = 2;
+    uint64 fee_per_gram = 3;
+    string message = 4;
+}
+
+message TransferResponse {
+  repeated TransferResult results = 1;
+}
+
+message TransferResult {
+    string address = 1;
+    uint64 transaction_id = 2;
+    bool is_success = 3;
+    string failure_message = 4;
 }
 
 message GetCoinbaseRequest {
--- a/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -1,6 +1,19 @@
-use tari_app_grpc::tari_rpc::{wallet_server, GetCoinbaseRequest, GetCoinbaseResponse};
-use tari_wallet::WalletSqlite;
+use futures::future;
+use log::*;
+use tari_app_grpc::tari_rpc::{
+    wallet_server,
+    GetCoinbaseRequest,
+    GetCoinbaseResponse,
+    TransferRequest,
+    TransferResponse,
+    TransferResult,
+};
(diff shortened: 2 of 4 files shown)


applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -29,7 +48,62 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            .enumerate()
+            .map(|(idx, dest)| -> Result<_, String> {
+                let pk = CommsPublicKey::from_hex(&dest.address)
+                    .map_err(|_| format!("Destination address at index {} is malformed", idx))?;
+                Ok((dest.address, pk, dest.amount, dest.fee_per_gram, dest.message))
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(Status::invalid_argument)?;
+
+        let transfers = recipients
```
stringhandler, line 68: Perhaps it should try reserve all the funds before hand so that the batch executes as a unit?

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -31,6 +31,36 @@ import "google/protobuf/timestamp.proto";
...
+}
+
+message TransferResult {
+    string address = 1;
+    uint64 transaction_id = 2;
+    enum Status {
+        STATUS_PENDING = 0;
+        STATUS_FAILED = 1;
+        STATUS_CANCELLED = 2;
+        STATUS_BROADCAST = 3;
```
stringhandler: I don't think this can ever be returned though, same for `STATUS_COMPLETE`

sdbondi: The "query a transaction status" (`get_transaction` or `get_transaction_status`)  method will return these statuses

sdbondi: But I could always add those statuses when those calls land in a PR

stringhandler: I think it's probably best to keep it concise, it may even be worth creating a new enum for each call and mapping. These protos can be used by any wallet and the implementers may want to know what to do for these enums

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
sdbondi: I believe the terminology used is to "encumber" the UTXO (@philipr-za ?) - there isn't a call for that in the transaction service at the moment, and I'm not completely sure of the implications. I'll look into it.

I mostly want to get an interface to call out so that we can iterate on these calls while integrating with pool software. However, I'm also ok with having a branch with all the required changes that only get merged once complete.

applications/tari_app_grpc/proto/wallet.proto
sdbondi: Done. I removed the `Status` enum in favour of `bool is_success` and `string failure_message` fields for this call.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
philipr-za: Yes I have been using the term "encumber" to reserve outputs for a given transaction. Currently the only way it happens is to initiate a transaction via the transaction service so some logic will need to be added to both services to reserve funds for a batch of transactions without starting the transactions.

stringhandler: Perhaps that is out of scope for this, and can be added as a TODO or issue

stringhandler (approved):
LGTM

PR #2542, opened 2021-01-19
[Merge Mining] Remove merge mining template on successful submission

StriderDM:
## Description
Prevents multiple solutions being submitted for the same block.

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -341,12 +341,15 @@ impl InnerService {
                             status,
                             details: "failed to submit block".to_string(),
                         }) {
-                        Ok(..) => debug!(
-                            target: LOG_TARGET,
-                            "Submitted block #{} to Tari node in {:.0?} (SubmitBlock)",
-                            height,
-                            start.elapsed()
-                        ),
+                        Ok(..) => {
+                            debug!(
+                                target: LOG_TARGET,
+                                "Submitted block #{} to Tari node in {:.0?} (SubmitBlock)",
+                                height,
+                                start.elapsed()
+                            );
+                            block_templates_clone.remove(&hash).await;
+                        },
                         _ => debug!(
                             target: LOG_TARGET,
                             "Problem submitting block #{} to Tari node, responded in  {:.0?} (SubmitBlock)",
@@ -360,7 +363,7 @@ impl InnerService {
             } else {
                 info!(
                     target: LOG_TARGET,
-                    "Block submitted but no matching block template was found"
+                    "Block submitted but no matching block template was found, possible duplicate submission"
                 );
             }
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3073,9 +3073,9 @@ dependencies = [
 
 [[package]]
 name = "randomx-rs"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f132cf96745b4803e0c86e2fa0c2b301ebf1eb608a705a2788d6c876062089b"
+checksum = "d5b1588934c19acc7218925816a7d6af57699c4e0d43c1c1ccda3463f619e53b"
 dependencies = [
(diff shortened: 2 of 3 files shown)


applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -341,12 +341,15 @@ impl InnerService {
...
-                            start.elapsed()
-                        ),
+                        Ok(..) => {
+                            debug!(
+                                target: LOG_TARGET,
+                                "Submitted block #{} to Tari node in {:.0?} (SubmitBlock)",
+                                height,
+                                start.elapsed()
+                            );
+                            block_templates_clone.remove(&hash).await;
```
hansieodendaal, line 351: This negates the need for the call to `block_templates_clone.remove_outdated().await;` further down completely, which I believe is actually the correct behaviour.

hansieodendaal (commented):
I believe this PR is treating a symptom and not the cause. A previous PR introduced `block_templates_clone.remove_outdated().await;` to let templates live longer and not be removed after every base node submission; this PR reverses that. It should be possible to submit any block that is valid, even if it is using the same block template again. I think a better cause of action would only be to print a warning message in the log file if multiple block submissions are made using the same block template. 

_**Note:** Please see ~~https://github.com/tari-project/xmrig/pull/2~~ https://github.com/tari-project/xmrig/pull/3 where the root cause of multiple block submissions using the same block template has been addressed. This phenomenon specifically started to happen due to `self-select` with `submit-to-origin` merge mining testing._

applications/tari_merge_mining_proxy/src/proxy.rs
StriderDM: This PR does not reverse that. This one is slightly different, it removes the template after a successful submit to prevent another submission for the same block which was accepted. Previously a template was removed whether the submit was successful or not (due to the difficulty being too low), which was a bug.

StriderDM: I think what we have ended up doing is guarded against the same edge case in two places @hansieodendaal

hansieodendaal: I do not agree that we should delete a template after successful submission, as it is treating a symptom and not a cause. 

As a side note, and I know this was only briefly discussed, this implementation would also not allow multiple XMRig miners connecting to the same merge mining proxy to submit competing solutions for the same block, as the first one to submit will remove the opportunity for the others to submit.

StriderDM: You do make a good point.

stringhandler: @hansieodendaal Why would you want to continue mining on a template that has already been accepted to the chain?

hansieodendaal: @mikethetike we should definitely not mine on the same template again, but preventing it to be submitted to the blockchain again does not prevent mining on it.

StriderDM:
@hansieodendaal I have gone back and thought about this more and still think this is the correct approach.

`block_templates_clone.remove_outdated().await;` is still relevant as it will remove templates that have expired for which solutions were not found in time or abandoned by the miner and will prevent memory continually growing in the proxy. This also prevents templates being removed prematurely due to an invalid solution as we have experienced before (difficulty too low).

This solution will only allow one solution per accepted block, the miner should then move on to the next block at the earliest possible time. If a miner continually mines on the same block template, if it has already achieved the network difficulty or greater, I think the miner should be either viewed as malfunctioning or malicious. Since in this scenario, the miner could at some point in time within the validity window, potentially find a solution to the block that is greater in difficulty than another block further on in the chain. I highly doubt we would want that.

So this is a safe-guard in addition to the solution present in tari-project/xmrig#3 

While we have differing opinions on this, I think in the end, both this PR AND tari-project/xmrig#3 are required.

hansieodendaal (approved):
LGTM

stringhandler (approved):

PR #2543, opened 2021-01-20
[Chore] Update core lib versions to v0.7.3

stringhandler:
## Description

Update the libraries that were on v0.2.x. This PR brings the libs and the binaries onto the same versions. libwallet (wallet_ffi) still has it's own version


delta1 (approved):
LGTM

PR #2550, opened 2021-01-22
Encrypt Node_id and Tor_id at rest

SWvheerden:
## Description
This PR adds in the NodeID and TorID to the wallet's database as encrypted fields 

## Motivation and Context
This allows us to protect the sensitive private keys for the Tari Comms and Tor comms as they are not stored in a plain text file at rest anymore. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/init/mod.rs
+++ b/applications/tari_console_wallet/src/init/mod.rs
@@ -25,23 +25,36 @@ use crate::{
     wallet_modes::{PeerConfig, WalletMode},
 };
 use log::*;
+use rand::rngs::OsRng;
 use rpassword::prompt_password_stdout;
 use std::{fs, str::FromStr, sync::Arc};
-use tari_app_utilities::{
-    identity_management,
-    utilities::{setup_wallet_transport_type, ExitCodes},
-};
+use tari_app_utilities::utilities::{setup_wallet_transport_type, ExitCodes};
 use tari_common::{ConfigBootstrap, GlobalConfig, Network};
-use tari_comms::{peer_manager::Peer, NodeIdentity};
+use tari_comms::{
+    peer_manager::{Peer, PeerFeatures},
+    NodeIdentity,
+};
 use tari_comms_dht::{DbConnectionUrl, DhtConfig};
-use tari_core::{consensus::Network as NetworkType, transactions::types::CryptoFactories};
-use tari_p2p::{initialization::CommsConfig, seed_peer::SeedPeer, DEFAULT_DNS_SEED_RESOLVER};
+use tari_core::{
+    consensus::Network as NetworkType,
+    transactions::types::{CryptoFactories, PrivateKey},
+};
+use tari_crypto::keys::SecretKey;
+use tari_p2p::{
+    initialization::CommsConfig,
+    seed_peer::SeedPeer,
+    transport::TransportType::Tor,
+    DEFAULT_DNS_SEED_RESOLVER,
+};
 use tari_shutdown::ShutdownSignal;
 use tari_wallet::{
     base_node_service::config::BaseNodeServiceConfig,
     error::{WalletError, WalletStorageError},
     output_manager_service::config::OutputManagerServiceConfig,
-    storage::sqlite_utilities::initialize_sqlite_database_backends,
+    storage::{
+        database::{DbKey, DbKeyValuePair, DbValue, WalletBackend, WriteOperation},
+        sqlite_utilities::initialize_sqlite_database_backends,
+    },
     transaction_service::config::TransactionServiceConfig,
     wallet::WalletConfig,
     Wallet,
@@ -88,7 +101,7 @@ fn prompt_password(prompt: &str) -> Result<String, ExitCodes> {
 
 pub async fn change_password(
     config: &GlobalConfig,
-    node_identity: Arc<NodeIdentity>,
+    node_identity: Option<Arc<NodeIdentity>>,
     arg_password: Option<String>,
     shutdown_signal: ShutdownSignal,
 ) -> Result<(), ExitCodes>
@@ -167,7 +180,7 @@ pub fn wallet_mode(bootstrap: ConfigBootstrap) -> WalletMode {
 /// Setup the app environment and state for use by the UI
 pub async fn init_wallet(
     config: &GlobalConfig,
-    node_identity: Arc<NodeIdentity>,
+    node_identity: Option<Arc<NodeIdentity>>,
     arg_password: Option<String>,
     shutdown_signal: ShutdownSignal,
 ) -> Result<WalletSqlite, ExitCodes>
@@ -206,21 +219,77 @@ pub async fn init_wallet(
             }
         },
     };
-
     let (wallet_backend, transaction_backend, output_manager_backend, contacts_backend) = backends;
 
     debug!(
         target: LOG_TARGET,
         "Databases Initialized. Wallet encrypted? {}.", wallet_encrypted
     );
-
+    // TODO remove after next TestNet
+    // If we know node_identity is not passed in anymore we dont have to check if we need to write it to db. This
+    // assumes that if its passed in we need to still save it the database.
+    if let Some(mut id_arc) = node_identity {
+        let v = (*Arc::make_mut(&mut id_arc)).clone();
+        wallet_backend
+            .write(WriteOperation::Insert(DbKeyValuePair::Identity(Box::new(v))))
+            .map_err(|e| ExitCodes::WalletError(format!("Error creating Wallet database backends. {}", e)))?;
+    }
+    let node_identity = match wallet_backend
+        .fetch(&DbKey::Identity)
+        .map_err(|e| ExitCodes::WalletError(format!("Error creating Wallet database backends. {}", e)))?
+    {
+        Some(DbValue::Identity(v)) => Arc::new(v),
+        _ => {
+            let private_key = PrivateKey::random(&mut OsRng);
+            let node_id = NodeIdentity::new(
+                private_key,
+                config.public_address.clone(),
+                PeerFeatures::COMMUNICATION_CLIENT,
+            )
+            .map_err(|e| ExitCodes::WalletError(format!("We were unable to construct a node identity. {}", e)))?;
+            wallet_backend
+                .write(WriteOperation::Insert(DbKeyValuePair::Identity(Box::new(
+                    node_id.clone(),
+                ))))
+                .map_err(|e| ExitCodes::WalletError(format!("Error creating Wallet database backends. {}", e)))?;
+            Arc::new(node_id)
+        },
+    };
     // TODO remove after next TestNet
     transaction_backend.migrate(node_identity.public_key().clone());
+    let transport_type = setup_wallet_transport_type(&config);
+    let transport_type = match transport_type {
+        Tor(mut tor_config) => {
+            tor_config.identity = match tor_config.identity {
+                Some(v) => {
+                    // This is temp code and should be removed after testnet
+                    wallet_backend
+                        .write(WriteOperation::Insert(DbKeyValuePair::TorId((*v).clone())))
+                        .map_err(|e| {
+                            ExitCodes::WalletError(format!("Error creating Wallet database backends. {}", e))
+                        })?;
+                    std::fs::remove_file(&config.console_wallet_tor_identity_file)
+                        .map_err(|e| ExitCodes::WalletError(format!("Could not delete identity file {}", e)))?;
+                    Some(v)
+                },
+                _ => {
+                    match wallet_backend.fetch(&DbKey::TorId).map_err(|e| {
+                        ExitCodes::WalletError(format!("Error creating Wallet database backends. {}", e))
+                    })? {
+                        Some(DbValue::TorId(v)) => Some(Box::new(v)),
+                        _ => None,
+                    }
+                },
+            };
+            Tor(tor_config)
+        },
+        _ => transport_type,
+    };
 
     let comms_config = CommsConfig {
         node_identity,
         user_agent: format!("tari/wallet/{}", env!("CARGO_PKG_VERSION")),
-        transport_type: setup_wallet_transport_type(&config),
+        transport_type,
         datastore_path: config.console_wallet_peer_db_path.clone(),
         peer_database_name: "peers".to_string(),
         max_concurrent_inbound_tasks: 100,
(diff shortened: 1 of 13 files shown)


applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -206,21 +222,83 @@ pub async fn init_wallet(
...
+            },
+            Err(e) => {
+                return Err(ExitCodes::WalletError(format!(
+                    "Error creating Wallet database backends. {}",
+                    e
+                )));
+            },
+        }
+    };
+    debug!(target: LOG_TARGET, "after testnet case");
```
stringhandler: Debugging code?

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -206,21 +222,83 @@ pub async fn init_wallet(
...
     };
-
     let (wallet_backend, transaction_backend, output_manager_backend, contacts_backend) = backends;
 
     debug!(
         target: LOG_TARGET,
         "Databases Initialized. Wallet encrypted? {}.", wallet_encrypted
-    );
-
+    ); // TODO remove after next TestNet
```
stringhandler: Perhaps some more detail here?

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -206,21 +222,83 @@ pub async fn init_wallet(
...
+            Err(e) => {
+                return Err(ExitCodes::WalletError(format!(
+                    "Error creating Wallet database backends. {}",
+                    e
+                )));
+            },
+        }
+    };
+    debug!(target: LOG_TARGET, "after testnet case");
+    let _ = wallet_backend.fetch(&DbKey::Identity);
```
stringhandler: Duplicate code

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -206,21 +222,83 @@ pub async fn init_wallet(
...
-
     let (wallet_backend, transaction_backend, output_manager_backend, contacts_backend) = backends;
 
     debug!(
         target: LOG_TARGET,
         "Databases Initialized. Wallet encrypted? {}.", wallet_encrypted
-    );
-
+    ); // TODO remove after next TestNet
+    if node_identity.is_some() {
```
stringhandler: clippy usually picks this up as `if let Some(x) = ... {`

base_layer/wallet/src/storage/database.rs
```diff
@@ -47,12 +52,20 @@ pub trait WalletBackend: Send + Sync + Clone {
...
     ClientKey(String),
 }
 
 pub enum DbValue {
     CommsSecretKey(CommsSecretKey),
     CommsPublicKey(CommsPublicKey),
+    CommsAddress(Multiaddr),
+    CommsFeatures(u64),
+    Identity(NodeIdentity),
+    TorID(TorIdentity),
```
stringhandler:
```suggestion
    TorId(TorIdentity),
```

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -204,31 +230,93 @@ impl WalletSqliteDatabase {
...
+                WalletSettingSql::new(DbKey::TorID.to_string(), ciphertext_integral_nonce.to_hex()).set(&conn)?;
+            },
+        }
+
+        Ok(())
+    }
+
+    fn get_tor_id(&self, conn: &SqliteConnection) -> Result<Option<DbValue>, WalletStorageError> {
+        let cipher = acquire_read_lock!(self.cipher);
+        if let Some(key_str) = WalletSettingSql::get(format!("{}", DbKey::TorID), &conn)? {
```
stringhandler: nit: `.to_string()`

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -344,6 +485,21 @@ impl WalletBackend for WalletSqliteDatabase {
             ckv.set(&conn)?;
         }
 
+        // remove tor id encryption if present
+        let key_str = WalletSettingSql::get(format!("{}", DbKey::TorID), &conn)?;
+        match key_str {
+            Some(v) => {
+                let decrypted_key_bytes = decrypt_bytes_integral_nonce(&cipher, from_hex(v.as_str())?)
+                    .map_err(|_| WalletStorageError::AeadError("Decryption Error".to_string()))?;
```
stringhandler: Swallowing an error here

stringhandler (commented):
Happy in general, but a couple of clippy warnings and style nits. There are also a number of places where you have swallowed errors

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: Changed the error type here to not swallow any errors anymore

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: Removed all format!("{}" and replaced with to_string()

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -206,21 +219,84 @@ pub async fn init_wallet(
...
     debug!(
         target: LOG_TARGET,
         "Databases Initialized. Wallet encrypted? {}.", wallet_encrypted
     );
-
+    // TODO remove after next TestNet
+    // If we know node_identity is not passed in anymore we dont have to check if we need to write it to db. This
+    // assumes that if its passed in we need to still save it the database.
+    if node_identity.is_some() {
+        match Arc::try_unwrap(node_identity.unwrap()) {
```
philipr-za: Pretty sure you can do away with all of this unwrapping stuff and just go:
```Rust
(*Arc::make_mut(&mut node_identity)).clone()
```
Assuming you use the suggested `if let Some(x)` above to get inside the option

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -204,33 +229,99 @@ impl WalletSqliteDatabase {
...
+        if let Some(key_str) = WalletSettingSql::get(DbKey::TorId.to_string(), &conn)? {
+            let tor_json = match cipher.as_ref() {
+                None => key_str,
+                Some(cipher) => {
+                    let decrypted_key_bytes = decrypt_bytes_integral_nonce(&cipher, from_hex(&key_str)?)
+                        .map_err(|_| WalletStorageError::AeadError("Decryption Error".to_string()))?;
+                    let tor_json = std::str::from_utf8(&decrypted_key_bytes)
+                        .map_err(|_| WalletStorageError::AeadError("Decryption Error".to_string()))?
+                        .to_string();
+                    // Encryption and decryption breaks the json string, so we fix it.
```
philipr-za: This feels weird, the encryption should return the bytes stream byte for byte?

applications/tari_console_wallet/src/init/mod.rs
SWvheerden: Did all clippy comments

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: Its because we are encoding a string
so this:
"{\"private_key\":{\"Ed25519V3\":\"aFn0Bod+Vvq8/obUEbVPVsa9Vdk9p44FOavDKzJVoU72dyJ65k9Zy6+NuWrwI1uI51ApzG5IjDUH2NjW9fKmoA==\"},\"service_id\":\"pqdggzs7wlxguwdyynojpez3a7tfzo2zacks2ppivyp4pnsu7lmkpzqd\",\"onion_port\":18101}"
Becomes:
"\"{\\"private_key\\":{\\"Ed25519V3\\":\\"aFn0Bod+Vvq8/obUEbVPVsa9Vdk9p44FOavDKzJVoU72dyJ65k9Zy6+NuWrwI1uI51ApzG5IjDUH2NjW9fKmoA==\\"},\\"service_id\\":\\"pqdggzs7wlxguwdyynojpez3a7tfzo2zacks2ppivyp4pnsu7lmkpzqd\\",\\"onion_port\\":18101}\""

applications/tari_console_wallet/src/init/mod.rs
SWvheerden: Clippy was happy with this one,  but I updated the code with (*Arc::make_mut(&mut node_identity)).clone()

base_layer/wallet/src/storage/sqlite_db.rs
stringhandler: Why not serialize it as bytes with serde?

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -204,33 +229,99 @@ impl WalletSqliteDatabase {
...
     }
+
+    fn get_comms_secret_key(&self, conn: &SqliteConnection) -> Result<Option<DbValue>, WalletStorageError> {
+        let cipher = acquire_read_lock!(self.cipher);
+        if let Some(key_str) = WalletSettingSql::get(DbKey::CommsSecretKey.to_string(), &conn)? {
+            let secret_key = match cipher.as_ref() {
+                None => CommsSecretKey::from_hex(key_str.as_str())?,
+                Some(cipher) => {
+                    let decrypted_key_bytes = decrypt_bytes_integral_nonce(&cipher, from_hex(key_str.as_str())?)
+                        .map_err(|_| WalletStorageError::AeadError("Decryption Error".to_string()))?;
```
stringhandler: Still swallowing this error

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -243,6 +334,27 @@ impl WalletBackend for WalletSqliteDatabase {
                     Some(DbValue::ClientValue(v.value))
                 },
             },
+            DbKey::CommsAddress => self.get_comms_address(&conn)?,
+            DbKey::Identity => {
+                let secret_key = match self.get_comms_secret_key(&conn)? {
+                    Some(DbValue::CommsSecretKey(v)) => v,
+                    _ => return Ok(None),
+                };
+                let address = match self.get_comms_address(&conn)? {
```
stringhandler, line 338: Why not rather return a valid `Option<CommsAddress>` here instead of wrapping it in a `DbValue`? Seems to add a lot of code in the checking stage

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -187,6 +192,26 @@ impl WalletSqliteDatabase {
         Ok(())
     }
 
+    fn set_node_id(&self, node_id: &NodeIdentity, conn: &SqliteConnection) -> Result<(), WalletStorageError> {
+        let cipher = acquire_read_lock!(self.cipher);
+        WalletSettingSql::new(DbKey::CommsPublicKey.to_string(), node_id.public_key().to_hex()).set(&conn)?;
+        WalletSettingSql::new(DbKey::CommsAddress.to_string(), node_id.public_address().to_string()).set(&conn)?;
+        WalletSettingSql::new(DbKey::CommsFeatures.to_string(), node_id.features().bits().to_string()).set(&conn)?;
```
stringhandler, line 199: Why are features saved to the DB? Are these configurable?

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2339,7 +2340,7 @@ pub unsafe extern "C" fn transport_tcp_create(
 pub unsafe extern "C" fn transport_tor_create(
     control_server_address: *const c_char,
     tor_cookie: *const ByteVector,
-    tor_identity: *const ByteVector,
+    _tor_identity: *const ByteVector,
```
stringhandler: Probably best to just remove it

base_layer/wallet/src/storage/sqlite_db.rs
philipr-za: This is part of this modular storage backend pattern that we conceived in the very beginning of the project to allow for multiple backends to be written that implement the `fetch(...)` and `write(...)` methods of the trait. But I agree I think the boilerplate this creates is not worth the benefit for a potential easier future integration of some other backend. Though removing it now is a cleanup refactor in itself so 🤷‍♂️

stringhandler: Fair, but, I was more referring to the fact that `get_comms_address` could be written to return the result in the correct type (i.e. `CommsAddress`), and only the instance that needed to return a `DbValue` (i.e. `fetch`) wrap it.

SWvheerden: This is the same schematics we use all over the code base, from the BaseNode to the wallet for db write transactions.  I only did it this way to be consistent with the rest of the code.

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: I removed these all as well.

base_layer/wallet_ffi/src/lib.rs
philipr-za: careful, this will break the Swift and JNI wrappers if we don't update those at the same time.

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: features is, 
The other two is purely convenance so that we dont have to recalc them. I added everything from the node_id.json file.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2951,6 +2970,15 @@ pub unsafe extern "C" fn wallet_create(
 
             match w {
                 Ok(mut w) => {
+                    // lets ensure the wallet tor_id is saved
+                    if let Some(hs) = w.comms.hidden_service() {
+                        if let Err(e) = runtime.block_on(w.db.set_tor_identity(hs.tor_identity().clone())) {
+                            warn!(
+                                target: LOG_TARGET,
+                                "Could not restart transaction negotiation protocols: {}", e
```
philipr-za: This message doesn't seem related to the error.

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: `fn get_comms_address` is a private function. 
You either have to unwrap the option here, and use above straight
OR
Use here straight, and wrap above

stringhandler: I disagree that it is like that in the code base, and lines 330 and 324 are both examples of doing it correctly, and `get_comms_address` is a new method created in this PR. I would say change line 337 to 
```
DbKey::CommsAddress => self.get_comms_address(&conn)?.map(|v| DbValue::CommsAddress(v)),
```
and `get_comms_address` to:
```
fn get_comms_address(&self, conn: &SqliteConnection) -> Result<Option<CommsAddress>, WalletStorageError> 
```

base_layer/wallet/src/storage/sqlite_db.rs
SWvheerden: I will try and do that when saving as encrypted

PR #2553, opened 2021-01-24
Add get_version, get_transaction_info call to wallet GRPC

sdbondi:
## Description

`getVersion` returns the version of the wallet. This is useful for
testing (is the wallet listening where I think it is? Is my GRPC client working?) and for client compatibility.

`get_transaction_info` bulk queries transactions given by their
transaction IDs returning basic information required to tell that a
transaction has succeeded or is failed/canceled.

## Motivation and Context

Initial implementation of calls needed for pool integration. These calls may change when a "transaction queue" is implemented.

## How Has This Been Tested?

Cucumber tests are TODO. Not explicitly tested.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -25,6 +25,8 @@ package tari.rpc;
 
 import "google/protobuf/timestamp.proto";
 
+/// An Empty placeholder for endpoints without request parameters
+message Empty {}
 
 // The BlockHeader contains all the metadata for the block, including proof of work, a link to the previous block
 // and the transaction kernels.
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -24,15 +24,18 @@ syntax = "proto3";
 package tari.rpc;
 
 import "types.proto";
-import "google/protobuf/wrappers.proto";
-import "google/protobuf/timestamp.proto";
 
 // The gRPC interface for interacting with the wallet.
 service Wallet {
+    rpc GetVersion (Empty) returns (GetVersionResponse);
     // This returns a coinbase transaction
     rpc GetCoinbase (GetCoinbaseRequest) returns (GetCoinbaseResponse);
     // Send Tari to a number of recipients
     rpc Transfer (TransferRequest)  returns (TransferResponse);
+}
+
+message GetVersionResponse {
+    string version = 1;
 }
 
 message TransferRequest {
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -23,16 +23,21 @@ syntax = "proto3";
 
 package tari.rpc;
 
+import "google/protobuf/timestamp.proto";
 import "types.proto";
 
 // The gRPC interface for interacting with the wallet.
 service Wallet {
-    rpc GetVersion (Empty) returns (GetVersionResponse);
+    rpc GetVersion (GetVersionRequest) returns (GetVersionResponse);
     // This returns a coinbase transaction
     rpc GetCoinbase (GetCoinbaseRequest) returns (GetCoinbaseResponse);
     // Send Tari to a number of recipients
     rpc Transfer (TransferRequest)  returns (TransferResponse);
+    // Returns the transaction details for the given transaction IDs
+    rpc GetTransactionInfo (GetTransactionInfoRequest) returns (GetTransactionInfoResponse);
 }
+
+message GetVersionRequest { }
 
 message GetVersionResponse {
     string version = 1;
@@ -58,6 +63,49 @@ message TransferResult {
     uint64 transaction_id = 2;
     bool is_success = 3;
     string failure_message = 4;
+}
+
+message GetTransactionInfoRequest {
+    repeated uint64 transaction_ids = 1;
+}
+
+message GetTransactionInfoResponse {
+    repeated TransactionInfo statuses = 1;
+}
+
+message TransactionInfo {
+    uint64 tx_id = 1;
+    bytes source_pk = 2;
+    bytes dest_pk = 3;
+    TransactionStatus status = 4;
+    TransactionDirection direction = 5;
+    uint64 amount = 6;
+    uint64 fee = 7;
+    bool is_cancelled = 8;
+    bytes excess_sig = 9;
+    google.protobuf.Timestamp timestamp = 10;
+    string message = 11;
+}
+
+enum TransactionDirection {
+    TRANSACTION_DIRECTION_UNKNOWN = 0;
+    TRANSACTION_DIRECTION_INBOUND = 1;
+    TRANSACTION_DIRECTION_OUTBOUND = 2;
+}
+
+enum TransactionStatus {
+    // This transaction has been completed between the parties but has not been broadcast to the base layer network.
+    TRANSACTION_STATUS_COMPLETED = 0;
+    // This transaction has been broadcast to the base layer network and is currently in one or more base node mempools.
+    TRANSACTION_STATUS_BROADCAST = 1;
+    // This transaction has been mined and included in a block.
+    TRANSACTION_STATUS_MINED = 2;
+    // This transaction was generated as part of importing a spendable UTXO
+    TRANSACTION_STATUS_IMPORTED = 3;
+    // This transaction is still being negotiated by the parties
+    TRANSACTION_STATUS_PENDING = 4;
+    // This is a created Coinbase Transaction
+    TRANSACTION_STATUS_COINBASE = 5;
 }
 
 message GetCoinbaseRequest {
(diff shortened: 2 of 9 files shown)


delta1 (dismissed):
LGTM - untested

applications/tari_app_grpc/proto/types.proto
```diff
@@ -25,6 +25,8 @@ package tari.rpc;
 
 import "google/protobuf/timestamp.proto";
 
+/// An Empty placeholder for endpoints without request parameters
+message Empty {}
```
stringhandler, line 29: I'm not a fan of having an `Empty` type. It kind of goes against the forward compatibility that `protobuf` is trying to achieve. It would be better to have a `GetVersionRequest` that is empty so that fields can be added to it in future. I'll admit that it is unlikely that `GetVersionRequest` ever needs members, but I think it's simpler for new devs if there is a clear pattern that we stick to

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -57,6 +62,29 @@ message TransferResult {
...
+    repeated uint64 transaction_ids = 1;
+}
+
+message GetTransactionInfoResponse {
+    repeated TransactionInfo statuses = 1;
+}
+
+message TransactionInfo {
+    uint64 tx_id = 1;
+    string address = 2;
```
stringhandler: Not `bytes` maybe?

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -57,6 +62,29 @@ message TransferResult {
...
+    enum Status {
+        STATUS_PENDING = 0;
+        STATUS_CANCELLED = 1;
+        STATUS_COMPLETE = 2;
+        STATUS_MINED = 3;
+    }
+    Status status = 3;
+    uint64 amount = 4;
+    uint64 fee = 5;
+    string transactionHash = 6;
```
stringhandler:
```suggestion
    bytes transaction_hash = 6;
```

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -57,6 +62,29 @@ message TransferResult {
...
+}
+
+message GetTransactionInfoResponse {
+    repeated TransactionInfo statuses = 1;
+}
+
+message TransactionInfo {
+    uint64 tx_id = 1;
+    string address = 2;
+    enum Status {
```
stringhandler: Perhaps move out of nested so it can be reused?

stringhandler (commented):
Similar to #2554, although it only returns a single tx

applications/tari_app_grpc/proto/types.proto
sdbondi: You can replace `Empty` with `GetVersionRequest` later. The protobuf wire-format doesn't care about the message "type" just about it's shape. Though philosophically, I agree. I'll remove the empty type here and in the other base node call where it is used.

applications/tari_app_grpc/proto/wallet.proto
sdbondi: A string was chosen because it makes storage easier (blob vs varchar, select query) and removes the need for a hex / base 58 dependency and conversions. In effect, the caller needs to know how to tari converts the byte (i.e bincode) representation to a string. If this field is a "canonical string representation of an address", we can treat this string a like a unique id without caring what it encodes.

applications/tari_app_grpc/proto/wallet.proto
sdbondi: Same argument as for address. If we want a user to be able to query a transaction hash within the pool db, then it simplifies things to have a canonical string representation.

applications/tari_app_grpc/proto/wallet.proto
stringhandler: I don't know. GRPC handles byte arrays really well and all of the other protos are defined as bytes. I would say we should keep it consistent as bytes, or change all places to use a canonical string representation. This change would also be needed to be applied to `tari_crypto`

sdbondi: Ok fair enough, I'll make the change

applications/tari_app_grpc/proto/wallet.proto
sdbondi: Will change

hansieodendaal:
Not getting the correct response here using BloomRPC.

Request:
```
{
  "transaction_ids": [
    17080252999620409121
  ]
}
```

Response:
```
{
  "error": "5 NOT_FOUND: Transaction ID at position 0 not found"
}
```

Also trying to test with a different interface.

**Edit:**
This input format works fine using quotes (thanks @mikethetike), getting the proper responses:
{
  "transaction_ids": [
    "17080252999620409121"
  ]
}

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+fn convert_wallet_transaction_into_transaction_info(
+    tx: models::WalletTransaction,
+    wallet_pk: &CommsPublicKey,
+) -> TransactionInfo
+{
+    use models::WalletTransaction::*;
+    match tx {
+        PendingInbound(tx) => TransactionInfo {
+            tx_id: tx.tx_id,
+            source_pk: tx.source_public_key.to_vec(),
```
hansieodendaal, line 176: I think this returned value should be in hex, so that a human user can see, verify and correlate easily with their console wallet.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            dest_pk: wallet_pk.to_vec(),
```
hansieodendaal, line 177: ditto hex

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            status: TransactionStatus::from(tx.status) as i32,
+            amount: tx.amount.into(),
+            is_cancelled: tx.cancelled,
+            direction: TransactionDirection::Inbound as i32,
+            fee: 0,
+            excess_sig: Default::default(),
```
hansieodendaal, line 183: ditto hex

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            timestamp: Some(naive_datetime_to_timestamp(tx.timestamp)),
```
hansieodendaal, line 184: I think the timestamp format should correspond to that used in the console wallet, for the benefit of human users.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            message: tx.message,
+        },
+        PendingOutbound(tx) => TransactionInfo {
```
hansieodendaal, line 187: ditto `PendingInbound`

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -107,4 +126,94 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            excess_sig: Default::default(),
+            timestamp: Some(naive_datetime_to_timestamp(tx.timestamp)),
+            message: tx.message,
+        },
+        Completed(tx) => TransactionInfo {
+            tx_id: tx.tx_id,
+            source_pk: tx.source_public_key.to_vec(),
+            dest_pk: tx.destination_public_key.to_vec(),
+            status: TransactionStatus::from(tx.status) as i32,
+            amount: tx.amount.into(),
```
hansieodendaal, line 205: ditto `PendingInbound`

hansieodendaal (commented):
Thank you for adding the additional fields Stan. My comments below revolve around the use case of a human user wanting to correlate one or more of their transactions.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
sdbondi: see here https://github.com/tari-project/tari/pull/2553#discussion_r563550766

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
sdbondi: protobuf is not a human readable format. `google.protobuf.Timestamp` is the standard protobuf type for timestamps.

stringhandler (approved):

PR #2554, opened 2021-01-25
Get transaction details via GRPC

hansieodendaal:
## Description
This change adds a method to the console wallet's GRPC server to query the status of a transaction. Most relevant transaction metadata are returned.

Also removed spam log messages in `base_layer/wallet/src/base_node_service/service.rs`.

## Motivation and Context
Extending the console wallet's GRPC interface is important in order to enhance its flexibility of use.

## How Has This Been Tested?
Tested on a running console wallet in Windows interfacing with a GRPC client.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -31,6 +31,8 @@ import "google/protobuf/timestamp.proto";
 service Wallet {
     // This returns a coinbase transaction
     rpc GetCoinbase (GetCoinbaseRequest) returns (GetCoinbaseResponse);
+    // This returns the transaction status
+    rpc GetTransactionDetails (GetTransactionDetailsRequest) returns (GetTransactionDetailsResponse);
     // Send Tari to a number of recipients
     rpc Transfer (TransferRequest)  returns (TransferResponse);
 }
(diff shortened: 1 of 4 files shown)


sdbondi:
To clarify #2553 is needed by pools to bulk query the status of an arbitrary number of transactions in order to maintain their internal payout state. Perhaps `get_bulk_transaction_info` is a better name. It has been proposed to build a system that allows transaction to be 'queued' by the wallet as UTXOs become available, `get_bulk_transaction_info` would include info on the status of these 'queued' transactions.

The call in the PR provides single transaction detail, and should reflect the data model used within the existing transaction service, perhaps `get_transaction` is a better name.

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -66,3 +68,23 @@ message GetCoinbaseRequest {
...
+    string txid = 1;
+    string source_pub_key = 2;
+    string dest_pub_key = 3;
+    string direction = 4;
+    string amount = 5;
+    string fee = 6;
+    string status = 7;
+    bool cancelled = 8;
+    string message = 9;
+    string time_stamp = 10;
```
stringhandler, line 89: Should be `timestamp` and use the existing Timestamp type

stringhandler (commented):
Similar to #2553 so should be merged into one. There are a number of values returned as `string` here when they should be byte arrays or integers

hansieodendaal:
> The call in the PR provides single transaction detail, and should reflect the data model used within the existing transaction service, perhaps `get_transaction` is a better name.

The information returned is similar to what is displayed in the console wallet, and specifically does not include the transaction itself. therefore `get_transaction` has not been deemed a proper name, either `get_transaction_status` because a bit more information is returned than just the status.

Console wallet `Transaction Details`:
```
TxID:                   12530268558120013370
Source Public Key:      829cb95ab5a9abacd29b6ede281450405e236f48e4aaf8cb5956a84498986915
Destination Public Key: 829cb95ab5a9abacd29b6ede281450405e236f48e4aaf8cb5956a84498986915
Direction:              Inbound
Amount:                 5481.089221 T
Fee:                    0 µT
Status:                 Mined
Message:                Coinbase Transaction for Block 23870
Timestamp:              2021-01-21 12:15:51
Excess:                 9d7b239634a3b0b01fca77753db5deb1069292fff8edac59c07003c6d02d230e
```

hansieodendaal:
As per discussion, this functionality implemented into #2553

PR #2561, opened 2021-01-26
Add WalletTransfer feature to cucumber tests

StriderDM:
## Description
Added WalletTransfer feature.
Added walletClient.js
Updated steps.js
Updated walletProcess.js
Updated package.json and package-lock.json

## Motivation and Context
Tests for wallet transfer grpc method.

## How Has This Been Tested?
```
$ ./node_modules/.bin/cucumber-js --name "As a wallet I want to submit multiple transfers"
```

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:
* [x] I'm merging against the `development` branch.
* [ ] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [] I have added tests to cover my changes.

PR #2562, opened 2021-01-26
Fix returned field name in getTransactionInfo call

sdbondi:
Returned field name in getTransactionInfo call should be `transactions`
instead of `statuses`


sdbondi:
Ref #2561

stringhandler (approved):

PR #2565, opened 2021-01-27
Refactor wallet bootup to not cause confusion

SWvheerden:
## Description

The wallet previously required a node_id file to run. This was only created if the wallet was run with --create_id. After which the node will ext. 

This behaviour is changed in that the node will create an id, if none is found and not exit. But if run the node with --create_id it would create a new id file, delete this and exit. This is now changed to exit before deletion of the file. 

A warning was also added to explain to the error message from the shared code between the wallet and the BN, which still requires the error message. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

* [x ] I'm merging against the `development` branch.

### DIFF
--- a/applications/tari_console_wallet/src/main.rs
+++ b/applications/tari_console_wallet/src/main.rs
@@ -61,6 +61,18 @@ fn main_inner() -> Result<(), ExitCodes> {
         Ok(v) => Some(v),
         _ => None,
     };
+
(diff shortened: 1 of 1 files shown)


stringhandler (approved):

hansieodendaal:
LGTM

PR #2567, opened 2021-01-27
Add ability to set maximum number of RandomX VMs

stringhandler:
## Description

Some low memory machines cannot allocate memory to store multiple RandomX VMs, resulting in a CreateError. This PR adds the ability to configure the number of VMs, which can be 1. Higher numbers will perform better when the seed switches often, which might only happen during an attack on the network, but 1 VM seems to have little performance impact.


delta1 (approved):
LGTM

sdbondi (approved):

PR #2583, opened 2021-01-29
Remove "add one to port" hack in console wallet

sdbondi:
The "add one to port" hack was used to ensure
that port numbers don't conflict between the base node and it's wallet .
Due to deficiencies in the current configuration framework assigning custom 
configuration to other applications requires a lot of boilerplate and mapping code.

This PR removes this hack from the console wallet code (only effects TCP
transport) as it creates confusion when implementing cucumber tests.


hansieodendaal (approved):
Tested, working perfectly! LGTM

stringhandler (approved):

PR #2590, opened 2021-01-29
Filter out mature UTXO's when choosing UTXO's for new transactions

SWvheerden:
## Description

This PR adds in a handle to the OMS to the BNS so that it can query the last updated chain_meta_state from its connected base_node. 
This allows the OMS to filter out UTXO's that are not yet mature when its selecting UTXO to send out. 

## Motivation and Context

Select not yet mature UTXO's will result in the Base_node rejecting the transaction.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/bootstrap/wallet.rs
+++ b/applications/tari_base_node/src/bootstrap/wallet.rs
@@ -49,12 +49,18 @@ use tari_p2p::{
 use tari_service_framework::{ServiceHandles, ServiceInitializerContext, StackBuilder};
 use tari_shutdown::ShutdownSignal;
 use tari_wallet::{
+    base_node_service::{config::BaseNodeServiceConfig, handle::BaseNodeServiceHandle, BaseNodeServiceInitializer},
     output_manager_service::{
         config::OutputManagerServiceConfig,
         storage::sqlite_db::OutputManagerSqliteDatabase,
         OutputManagerServiceInitializer,
     },
-    storage::{sqlite_utilities, sqlite_utilities::WalletDbConnection},
+    storage::{
+        database::WalletDatabase,
+        memory_db::WalletMemoryDatabase,
+        sqlite_utilities,
+        sqlite_utilities::WalletDbConnection,
+    },
     transaction_service::{
         config::TransactionServiceConfig,
         storage::sqlite_db::TransactionServiceSqliteDatabase,
@@ -117,7 +123,7 @@ impl WalletBootstrapper {
         let comms_config = self.create_comms_config();
         let transport_type = comms_config.transport_type.clone();
 
-        let base_node_peer = self.base_node_peer;
+        let base_node_peer = self.base_node_peer.clone();
 
         let mut handles = StackBuilder::new(self.interrupt_signal)
             .add_initializer(P2pInitializer::new(comms_config, publisher))
@@ -155,12 +161,21 @@ impl WalletBootstrapper {
                     broadcast_send_timeout: config.transaction_broadcast_send_timeout,
                     ..Default::default()
                 },
-                peer_message_subscriptions,
+                peer_message_subscriptions.clone(),
                 transaction_db,
                 self.node_identity.clone(),
                 self.factories,
             ))
+            .add_initializer(BaseNodeServiceInitializer::new(
+                BaseNodeServiceConfig::default(),
+                peer_message_subscriptions,
+                WalletDatabase::new(WalletMemoryDatabase::new()),
+            ))
             .build()
+            .await?;
+        let mut base_node_service_handle = handles.expect_handle::<BaseNodeServiceHandle>();
+        base_node_service_handle
+            .set_base_node_peer(self.base_node_peer.clone())
             .await?;
 
         let comms = handles
--- a/base_layer/wallet/src/base_node_service/handle.rs
+++ b/base_layer/wallet/src/base_node_service/handle.rs
@@ -74,6 +74,14 @@ impl BaseNodeServiceHandle {
         self.event_stream_sender.subscribe().fuse()
     }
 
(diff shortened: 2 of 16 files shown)


stringhandler (commented):
I disagree with the approach here, specifically around the splitting of UTXOSelectionStrategy into UTXOSortingStrategy and UTXOFilteringStrategy. 

The UTXOSelectionStrategy is about how to pick UTXOs, and should only ever pick valid ones. It should encompass the whole process of choosing a set of UTXOs for the amount. Sorting and filtering are not independently configurable. 

I would suggest leaving UTXOSelectionStrategy as is, and changing the code to make sure it can only select valid UTXOs

SWvheerden:
Perhaps some the choices in `UTXOFilteringStrategy` is redundant like the select not mature utxo's. 
If we remove `UTXOFilteringStrategy` which of the two remaining do we force? Do we stick to `TryOnlyMature`?

applications/tari_base_node/src/bootstrap/wallet.rs
```diff
@@ -162,13 +168,22 @@ impl WalletBootstrapper {
...
-                peer_message_subscriptions,
+                peer_message_subscriptions.clone(),
                 transaction_db,
                 self.node_identity.clone(),
                 self.factories,
             ))
+            .add_initializer(BaseNodeServiceInitializer::new(
+                BaseNodeServiceConfig::default(),
+                peer_message_subscriptions,
+                WalletDatabase::new(WalletMemoryDatabase::new()),
```
stringhandler, line 172: Is Memory correct here?

base_layer/wallet/src/storage/database.rs
```diff
@@ -76,6 +79,7 @@ pub enum DbKeyValuePair {
     ClientKeyValue(String, String),
     Identity(Box<NodeIdentity>),
     TorId(TorIdentity),
+    BNChainMeta(ChainMetadata),
```
stringhandler:
```suggestion
    BaseNodeChainMeta(ChainMetadata),
```

base_layer/wallet/src/storage/database.rs
```diff
@@ -236,6 +266,7 @@ impl Display for DbValue {
             DbValue::CommsAddress(_) => f.write_str(&"Comms Address".to_string()),
             DbValue::TorId(v) => f.write_str(&format!("Tor ID: {}", v)),
             DbValue::Identity(v) => f.write_str(&format!("Node Identity: {}", v)),
+            DbValue::BNChainMeta(v) => f.write_str(&format!("Last seen Chain meta data from Base node:{}", v)),
```
stringhandler:
```suggestion
            DbValue::BNChainMeta(v) => f.write_str(&format!("Last seen chain metadata from base node:{}", v)),
```

applications/tari_base_node/src/bootstrap/wallet.rs
SWvheerden: Yes, In the current BaseNode we dont run a full wallet, only some services. We dont currently boot up a new sql db for the private key, tor key etc. 

The normal console wallet etc does have a DB, so we can pass that in this service. This is more temp code until we remove the basenode wallet. But it also does not need persistant storage for the base_node info, as the wallet inside of the base_node will always be connected to the base_node itself.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -902,6 +906,27 @@ where
...
+        let trimmed = uo.len();
+        trace!(target: LOG_TARGET, "We found {} UTXOs to select from", trimmed);
+
+        // Now lets try to filter the UTXOs we cant yet spend.
+        let tip_height: u64 = self
+            .base_node_service
+            .get_connected_base_node_state()
+            .await?
+            .height_of_longest_chain();
+        uo = uo
```
stringhandler: nit: This is inefficient with a large number of utxos. It will be better to combine this filter with the next `iter`

base_layer/wallet/src/base_node_service/handle.rs
```diff
@@ -74,6 +74,14 @@ impl BaseNodeServiceHandle {
         self.event_stream_sender.subscribe().fuse()
     }
 
+    pub async fn get_connected_base_node_state(&mut self) -> Result<ChainMetadata, BaseNodeServiceError> {
+        match self.handle.call(BaseNodeServiceRequest::GetChainMetadata).await?? {
+            BaseNodeServiceResponse::ChainMetadata(Some(v)) => Ok(v),
+            BaseNodeServiceResponse::ChainMetadata(None) => Err(BaseNodeServiceError::NoChainMetadata),
```
philipr-za, line 80: Don't see why you need a whole new error type for this, just pass the Option out?

SWvheerden: I see it as more complete. For this to return an error. The wallet must have never talked to a basenode in its entire life or at least be conencted to one.

philipr-za: So this is a style nit but I see the fact that the value returned was a None communicates that information which is not actually an Error. An error means something when wrong with the call i.e. getting the connected base node state broke somehow like the db was broken. Getting that state was done without error and the state was None meaning we haven't spoken to a base node.

stringhandler:
Just need some clippy love

stringhandler (approved):

base_layer/wallet/src/base_node_service/handle.rs
stringhandler: Surprising how much this error that was never supposed to happen came up

SWvheerden: This points me to there is some other problem in code. Either by the way we handle linking of components or by us not properly linking them up. 
The current iteration I have seen is from a wallet that was/is connected to a node. Meaning something in the backend is not working 100%

Issue #2579, opened 2021-01-28
Compile error when try to compile tari_comms


jont4:
**Describe the bug**
When trying to compile tari_comms, tari_app_grpc, tari_core and maybe other related, I got the following error:

tari_comms v0.8.0 (/root/tari/comms)                                     Caused by: process didn't exit successfully: 
`/root/tari/target/release/build/tari_comms-72d5d76e61bf6005/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'status: exit code: 1 - Operation not permitted (os error 1)      ', common/src/protobuf_build.rs:28:13
stack backtrace:
0: rust_begin_unwind   at / rustc/576d27c5a6c80cd39ef57d7398831d8e177573cc/library/std/src/panicking.rs:475                                             1: std::panicking::begin_panic_fmt  at /rustc/576d27c5a6c80cd39ef57d7398831d8e177573cc/library/std/src/panicking.rs:429                                             2: tari_common::protobuf_build::fmt
3: tari_common::protobuf_build::ProtoCompiler::compile
4: build_script_build::main
5: core::ops::function::FnOnce::call_once

**To Reproduce**
1. Just compile: cargo build --release

**Smartphone (please complete the following information):**
 - Device: Lenovo Z5 Pro GT, Snapdragon 855
 - OS: Android 10 (BlissROM), running Ubuntu 20.04 on termux via chroot (configured via Andronix)

delta1:
Hey @jontaix, thanks for reporting! From what I can find, this seems to be a file permissions issue ([link](https://stackoverflow.com/questions/10937806/oserror-error-1-operation-not-permitted)) 

`profobuf_build.rs` is trying to a write a file and doesn't have permissions to do so, please can you check the permissions on all files and folders in your source code folder? 

You can use `chown` to set ownership of the files to your user/group, and `chmod` to enable write permissions. 

https://www.baeldung.com/linux/chown-chmod-permissions

sdbondi:
@jontaix During build, code is generated from protobuf files and written to the file system. It appears that your environment does not have write permissions.

jont4:
> Hey @jontaix, thanks for reporting! From what I can find, this seems to be a file permissions issue ([link](https://stackoverflow.com/questions/10937806/oserror-error-1-operation-not-permitted))
> 
> `profobuf_build.rs` is trying to a write a file and doesn't have permissions to do so, please can you check the permissions on all files and folders in your source code folder?
> 
[...]

PR #2591, opened 2021-02-01
Skip rustfmt by defaut in protobuf builds

sdbondi:
rustfmt is now skipped by default in protobuf builds.
It can be enabled by calling the `perform_rustfmt()` method.

Ref #2579


delta1 (approved):
LGTM

PR #2595, opened 2021-02-01
[MM_Proxy] Submit to origin daemon configuration

StriderDM:
## Description
This PR allows configuring the proxy to submit to the original monero daemon. 

## Motivation and Context
When using self-select mode in XMRig, submitting to monero will be done by the pool and is not necessary to do so from the proxy.

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -29,6 +29,7 @@ use bytes::BytesMut;
 use futures::{StreamExt, TryFutureExt};
 use hyper::{
     body::Bytes,
+    header::HeaderValue,
     http::{header, response::Parts},
     service::Service,
     Body,
@@ -37,6 +38,7 @@ use hyper::{
     Response,
     StatusCode,
     Uri,
+    Version,
 };
 use jsonrpc::error::StandardError;
 use reqwest::{ResponseBuilderExt, Url};
@@ -75,6 +77,7 @@ pub struct MergeMiningProxyConfig {
     pub grpc_base_node_address: SocketAddr,
     pub grpc_console_wallet_address: SocketAddr,
     pub proxy_host_address: SocketAddr,
+    pub proxy_submit_to_origin: bool,
     pub wait_for_initial_sync_at_startup: bool,
 }
 
@@ -89,6 +92,7 @@ impl From<GlobalConfig> for MergeMiningProxyConfig {
             grpc_base_node_address: config.grpc_base_node_address,
             grpc_console_wallet_address: config.grpc_console_wallet_address,
             proxy_host_address: config.proxy_host_address,
+            proxy_submit_to_origin: config.proxy_submit_to_origin,
             wait_for_initial_sync_at_startup: config.wait_for_initial_sync_at_startup,
         }
     }
@@ -114,6 +118,18 @@ impl MergeMiningProxyService {
     pub async fn check_connections<W: Write>(&self, w: &mut W) -> bool {
         let mut is_success = true;
         let inner = &self.inner;
+
+        if inner.config.proxy_submit_to_origin {
+            let _ = writeln!(
+                w,
+                "Solo mining configuration detected, configured to submit to Monero daemon."
+            );
+        } else {
+            let _ = writeln!(
+                w,
+                "Pooled mining configuration detected, configured to not submit to Monero daemon."
+            );
+        }
 
         let _ = writeln!(w, "Connections:");
 
@@ -606,7 +622,7 @@ impl InnerService {
         );
         let mut builder = self
             .http_client
-            .request(request.method().clone(), monerod_uri)
+            .request(request.method().clone(), monerod_uri.clone())
             .headers(request.headers().clone());
 
         if self.config.monerod_use_auth {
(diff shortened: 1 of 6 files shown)


applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -615,13 +619,27 @@ impl InnerService {
         }
 
         trace!(target: LOG_TARGET, "Sending request {:?}", request.body(),);
-        let resp = builder
+
+        let json_response;
+        if submit_block && !self.config.proxy_submit_to_origin {
+            // Assume it would be accepted
+            let id = json["id"].as_i64().unwrap_or_else(|| -1);
+            let accept_response = format!(
```
sdbondi: Rather use the json macro 

```suggestion
            let accept_response =  json::json! {
               id: id,
               jsonrpc: "2.0",
               result: {},
               status: "OK",
               untrusted: false
            };
```

StriderDM: Thanks, done.

common/src/configuration/global.rs
```diff
@@ -557,6 +558,9 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         .get_bool(&key)
         .map_err(|e| ConfigurationError::new(&key, &e.to_string()))?;
 
+    let key = config_string("merge_mining_proxy", &net_str, "proxy_submit_to_origin");
+    let proxy_submit_to_origin = cfg.get_bool(&key).unwrap_or_else(|_| true);
```
hansieodendaal, line 562: Please add the config setting with explanatory text to `tari_config_example.toml` and `windows.toml` as well, The settings can be on the default but commented out.

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -615,13 +619,29 @@ impl InnerService {
         }
 
         trace!(target: LOG_TARGET, "Sending request {:?}", request.body(),);
-        let resp = builder
+
+        let json_response;
+        if submit_block && !self.config.proxy_submit_to_origin {
```
hansieodendaal, line 636: I think we can probably print a once-off status message to the proxy console to advocate if blocks will not be submitted to `monerod`.

common/src/configuration/global.rs
```diff
@@ -614,6 +618,7 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         wallet_base_node_service_request_max_age,
         prevent_fee_gt_amount,
         proxy_host_address,
+        proxy_submit_to_origin,
```
hansieodendaal, line 621: I think this will be more aptly named `proxy_submit_to_monerod` as the user have no context about `origin`. The text can instruct the user to makes this false if doing pool merged mining with `self-select` and `submit-to-origin`.

hansieodendaal (commented):
Hi @StriderDM, now seeing error messages about `required field "blocktemplate_blob" not found` when doing merge mining with `self-select` and `submit-to-origin`, so I do not think this fix work 100% as intended.

image

Please also see comments below.

applications/tari_merge_mining_proxy/src/proxy.rs
StriderDM: Good idea.

StriderDM:
@hansieodendaal Odd. I'll investigate the error. There shouldn't be anything about blocktemplate_blob on submit though.

StriderDM:
@hansieodendaal  Managed to fix it.

Screen Shot 2021-02-02 at 12 09 51 AM

common/src/configuration/global.rs
StriderDM: Done

common/src/configuration/global.rs
StriderDM: Have added comments to the configuration file to explain it.

common/config/presets/tari_sample.toml
```diff
@@ -27,7 +27,7 @@ pruning_horizon = 100
 [merge_mining_proxy.stibbons]
 monerod_url = "http://18.133.55.120:38081"
 proxy_host_address = "127.0.0.1:7878"
+proxy_submit_to_origin = false
```
hansieodendaal: I think this should reflect the default, which is `true`, therefore it does not need to be in this minimal sample config file.

common/config/presets/windows.toml
```diff
@@ -373,6 +373,9 @@ monerod_url = "http://18.133.55.120:38081"
 # Address of the tari_merge_mining_proxy application
 proxy_host_address = "127.0.0.1:7878"
 
+# Whether the proxy is configured to submit to monerod (solo mining) or not (pool mining)
+proxy_submit_to_origin = false
```
hansieodendaal: I think this should reflect the default, which is `true`, and commented out. The text could be something like:
```
# In sole merged mining, the block solution is usually submitted to the Monero blockchain 
# (monerod) as well as to the Tari blockchain, then this setting should be "true". With pool 
# merged mining, there is no sense in submitting the solution to the Monero blockchain as the 
# pool does that, then this setting should be "false". (default = true).
#proxy_submit_to_origin = true
```

common/config/tari_config_example.toml
```diff
@@ -373,6 +373,9 @@ monerod_url = "http://18.133.55.120:38081"
 # Address of the tari_merge_mining_proxy application
 proxy_host_address = "127.0.0.1:7878"
 
+# Whether the proxy is configured to submit to monerod (solo mining) or not (pool mining)
+proxy_submit_to_origin = false
```
hansieodendaal: Please see comment above.

hansieodendaal (commented):
Hi @StriderDM, tested, working really well now!
Just the remaining comment about the once-off status message to the proxy console and suggestions for the config files below. 
Thanks.

hansieodendaal (approved):
LGTM

stringhandler (approved):

PR #2602, opened 2021-02-02
WIP Update constants while syncing

SWvheerden:
## Description

This makes the target_difficulties struct aware of changes in the consensus_constants. It will now reload those values into its sub structs to ensure that calculated values are correctly calculated.  

## Motivation and Context

Currently the sync takes the consensus constants value at sync startup, but this never changes even thou the values might have changed. This allows them to change. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/proof_of_work/lwma_diff.rs
+++ b/base_layer/core/src/proof_of_work/lwma_diff.rs
@@ -6,12 +6,16 @@
 // https://github.com/zawy12/difficulty-algorithms/issues/3#issuecomment-442129791
 // https://github.com/zcash/zcash/issues/4021
 
-use crate::proof_of_work::{
-    difficulty::{Difficulty, DifficultyAdjustment},
-    error::DifficultyAdjustmentError,
+use crate::{
+    consensus::ConsensusConstants,
+    proof_of_work::{
+        difficulty::{Difficulty, DifficultyAdjustment},
+        error::DifficultyAdjustmentError,
+        PowAlgorithm,
+    },
 };
 use log::*;
-use std::cmp;
+use std::{cmp, convert::TryFrom};
 use tari_crypto::tari_utilities::epoch_time::EpochTime;
 
 pub const LOG_TARGET: &str = "c::pow::lwma_diff";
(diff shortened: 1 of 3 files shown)


base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -143,6 +147,14 @@ impl LinearWeightedMovingAverage {
         }
         self.target_difficulties.push((timestamp, target_difficulty));
     }
+
+    pub fn update_consensus_constants(&mut self, constants: &ConsensusConstants, pow_algo: PowAlgorithm) {
+        self.block_window = usize::try_from(constants.get_difficulty_block_window())
+            .expect("difficulty block window exceeds usize::MAX");
+        self.target_time = constants.get_diff_target_block_interval(pow_algo);
+        self.initial_difficulty = constants.min_pow_difficulty(pow_algo);
```
stringhandler, line 155: This will have no effect

base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -143,6 +147,14 @@ impl LinearWeightedMovingAverage {
         }
         self.target_difficulties.push((timestamp, target_difficulty));
     }
+
+    pub fn update_consensus_constants(&mut self, constants: &ConsensusConstants, pow_algo: PowAlgorithm) {
+        self.block_window = usize::try_from(constants.get_difficulty_block_window())
```
stringhandler, line 152: This also will not behave as expected

base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -143,6 +147,14 @@ impl LinearWeightedMovingAverage {
         }
         self.target_difficulties.push((timestamp, target_difficulty));
     }
+
+    pub fn update_consensus_constants(&mut self, constants: &ConsensusConstants, pow_algo: PowAlgorithm) {
```
stringhandler, line 151: Exposing this method here seems like a wrong coupling. It suggests an order that methods in this class may be called in.

stringhandler (commented):
I don't really like this approach. The changes to `target_difficulty` and `target_difficulties` seem fine, but I think IMHO that the interface that `lwma_diff` would expose after this is not generic and leads to tight coupling with how it is caused. Some of the fields in the `lwma_diff` struct can't be changed after creation (e.g. `block_window` and `initial_difficulty`)

I haven't looked deeply into the surrounding code, but perhaps a better approach is to leave `lwma_diff` as is, and constrain the min/max `target_difficulty` before it is sent to  `lwma_diff::add_front` and `lwma::add_back`

PR #2604, opened 2021-02-02
[base-node] Move capping of target difficulty

stringhandler:
An alternative fix to #2603 and #2602 

The main theory here is that the caller of `TargetDifficultyWindow::calculate()` has control of the max and min values allowed, and thus can use the latest constants for that height. My assumption is that if this is done at every point where the target difficulty is calculated just before it is saved to the database, it will always be within the bounds of the consensus constants for that height

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -249,18 +249,31 @@ impl ConsensusConstants {
 
     pub fn stibbons() -> Vec<Self> {
         let mut algos = HashMap::new();
-        // seting sha3/monero to 40/60 split
+        // Previously these were incorrectly set to `target_time` of 20 and 30, so
+        // most blocks before 1400 hit the minimum difficulty of 60M and 60k
+        // algos.insert(PowAlgorithm::Sha3, PowAlgorithmConstants {
+        //     max_target_time: 1800,
+        //     min_difficulty: 60_000_000.into(),
+        //     max_difficulty: u64::MAX.into(),
+        //     target_time: 30,
+        // });
+        // algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+        //     max_target_time: 1200,
+        //     min_difficulty: 60_000.into(),
+        //     max_difficulty: u64::MAX.into(),
+        //     target_time: 20,
+        // });
         algos.insert(PowAlgorithm::Sha3, PowAlgorithmConstants {
             max_target_time: 1800,
             min_difficulty: 60_000_000.into(),
-            max_difficulty: u64::MAX.into(),
-            target_time: 30,
+            max_difficulty: 60_000_000.into(),
+            target_time: 300,
         });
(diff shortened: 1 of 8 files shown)


hansieodendaal:
In my opinion, I am somewhat uncomfortable that target difficulty is now calculated in three different places with possible three different outcomes, instead of in one place. It can be called multiple times, but it should never be possible to get a different answer for the same height. The code will need to be changed in three places if anything is changed in future.

stringhandler:
@hansieodendaal I disagree with this. The fact I changed the `calculate` method to take in a min and max means that it cannot be called without providing a suitable value, which prompts the developer to think about where these values should come from. Previously the developer would have had to provide these values on creation of the object, which would be stale by the time they call it

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -249,18 +249,31 @@ impl ConsensusConstants {
...
+        //     min_difficulty: 60_000.into(),
+        //     max_difficulty: u64::MAX.into(),
+        //     target_time: 20,
+        // });
         algos.insert(PowAlgorithm::Sha3, PowAlgorithmConstants {
             max_target_time: 1800,
             min_difficulty: 60_000_000.into(),
-            max_difficulty: u64::MAX.into(),
-            target_time: 30,
+            max_difficulty: 60_000_000.into(),
```
hansieodendaal, line 269: This does not seem to be right, why is the `max_difficulty = min_difficulty` now?

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -249,18 +249,31 @@ impl ConsensusConstants {
...
+            target_time: 300,
         });
         algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
             max_target_time: 1200,
             min_difficulty: 60_000.into(),
-            max_difficulty: u64::MAX.into(),
-            target_time: 20,
+            max_difficulty: 60_000.into(),
```
hansieodendaal, line 275: Same comment w.r.t. `max_difficulty = min_difficulty` as above.

base_layer/core/src/consensus/consensus_constants.rs
stringhandler: Comment in line 252 explains this, perhaps I can make it more clear?

stringhandler: Basically because the target time was so low, there are blocks where the difficulty was the `min_difficulty`

hansieodendaal: Looking at it again, I would say the comment is not conveying the reason that this condition, `max_difficulty = min_difficulty`, should be implemented, being that we want to ensure old blocks mined up to 1400 are still accepted when validated.

sdbondi (approved):
After a few tries and some banned nodes, node synched to the expected chain. This workaround gets my vote as the best way forward

PR #2605, opened 2021-02-02
Update consensus constants as header sync progresses

sdbondi:
## Description

Another proposed solution to the consensus constants issue (PR #2604, #2603 #2602).

As header sync progresses, a change in consensus constants at a given height results in
reinitialisation of header validator state. This ensures the state is
correct for an arbitrary set of consensus constants with effective
height in range of the current sync.

## Motivation and Context

Header sync does not account for changing consensus constants effective from any height. This PR rectifies that.

## How Has This Been Tested?

Tested on base node

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -370,7 +370,8 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                 header.pow_algo(),
             );
             let height = header.height;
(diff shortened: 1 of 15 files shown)


hansieodendaal:
This seems like a robust approach.

sdbondi:
This version does not account for the possibility of a consensus change breakpoint happening during the initial chain split header validation. Given that #2604 provides a work-around for the immediate problem on stibbons and the low chance that we'll have to add consensus change breakpoints for target difficulty on mainnet, I don't think it's worth-while adding code to support the above edge-case.

sdbondi:
Closing because the particular stibbons testnet issue has been adequately addressed by #2604

PR #2612, opened 2021-02-04
[ci] Optimize CI

stringhandler:
Optimize CI process:
1. Remove `cargo fmt` checks from other steps. It is currently already done by the clippy check
~~2. Tests are not run in release mode, hopefully this speeds up compile time~~
3. Pre-compile binaries for cucumber tests  
4. Formatted Cucumber scenarios


delta1:
Something about this is breaking compilation in CI it seems

.circleci/config.yml
```diff
@@ -52,7 +46,7 @@ commands:
             - v6-cargo-cache-{{arch}}-{{checksum "rust-version"}}-<<parameters.release>>-{{checksum "Cargo.lock"}}
       - run:
           name: Build
-          command: cargo build --all --all-features --jobs=3 <<#parameters.release>>--release<</parameters.release>>
```
delta1: maybe it's this jobs param?

PR #2614, opened 2021-02-05
 New Transaction Validation protocol for Wallet

philipr-za:
# Attention: This PR will not work until the base nodes are running this code

## Description

This PR adds a transaction validation process that will check the validity of its mined transactions when it is connected to a base node. It will ask the base node if it has Kernels for all of its Mined transactions (both valid and invalid). Based on the response transactions will be marked as valid or invalid and can be updated from Confirmed to Unconfirmed or vice-versa.  This is done so that the mined transaction state of the wallet is consistent with the base node it is connected to. When the wallet is pointed at a different base node the validation will be performed again.

The validation protocols uses a new RPC method on the base node which accepts batches of signatures for transaction kernel queries allowing us to batch out queries into fewer calls.

Main elements of the PR are:
- Added `transaction_batch_query(…)` to the BaseNodeWalletRpcService
- Implement Transaction Validation Protocol as an asynchronous worker task
- Update the transaction service to run the validation protocol before the Broadcast Protocols are restarted and also after the second base node address update going forward.
- The Console Wallet is updated to show if a transaction is marked as Invalid
- Console wallet updated to refresh its state on validation completion and on MinedUncofirmed events
- Removed the Transaction MemoryDB as it was becoming cumbersome to keep it in sync with the sqlite DB behaviour. A helper function was added to improve ergonomics of making a sqlite db for tests.

## How Has This Been Tested?
Tests are provided for the RPC server update, the validation protocol itself and for the integration of the protocol into the Transaction Service.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
```diff
@@ -451,6 +473,14 @@ where TBackend: TransactionBackend + 'static
             },
         };
 
+        if !response.is_synced {
+            info!(
+                target: LOG_TARGET,
+                "Base Node reports not being synced, submission will be retried."
+            );
+            return Ok(false);
+        }
```
hansieodendaal: We have a bit of a logic problem here. This does not account for the case where a transaction has been mined with the required number of confirmations, even though the base node may not be synced. It should still count as mined.

hansieodendaal: Maybe:
``` rust
        if !response.is_synced {
            if !(response.location == TxLocation::Mined && response.confirmations >= self.resources.config.num_confirmations_required as u64) {
                info!(
                    target: LOG_TARGET,
                    "Base Node reports not being synced, submission will be retried."
                );
                return Ok((false, Some(response.height)));
            }
        }
```

philipr-za: I agree, this is a safe assumption. Will update.

delta1:
Looking good will test

delta1 (approved):
LGTM, tested and working - I had some issues with RPC failing to connect, but my CPU was pegged I believe that was the issue.

hansieodendaal (commented):
_(Duplicate comment - see next.)_

hansieodendaal (approved):
LGTM

Ran system-level tests and integration tests (as part of PR #2640), which were updated there to accommodate and test the new `mined unconfirmed` and `mined confirmed` transaction statuses as well. All works well.

hansieodendaal:
**Note:** Please also merge https://github.com/tari-project/wallet-grpc-client/pull/8 from `tari-project/wallet-grpc-client` with this PR if not done already.

stringhandler (approved):

PR #2615, opened 2021-02-05
Cucumber test for GRPC method relating to block explorer

StriderDM:
## Description
The Block Explorer depends on the GetNetworkDifficulty method, adding cucumber test to prevent its' accidental removal or deprecation

## Motivation and Context
As above

## How Has This Been Tested?
Manually run the test

## Types of changes
* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/integration_tests/features/BlockExplorerGRPC.feature
+++ b/integration_tests/features/BlockExplorerGRPC.feature
@@ -1,0 +1,12 @@
+@block-explorer
+Feature: Block Explorer GRPC
+
+  @critical
+  Scenario: GetNetworkDifficulty
+    Given I have a seed node NODE
+    And I have wallet WALLET connected to all seed nodes
+    And I have a merge mining proxy PROXY connected to NODE and WALLET
+    When I merge mine 2 blocks via PROXY
+    Then all nodes are at height 2
+    When I request the difficulties of a node NODE
+    Then Difficulties are available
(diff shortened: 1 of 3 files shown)


integration_tests/features/BlockExplorerGRPC.feature
```diff
@@ -0,0 +1,12 @@
...
+
+  @critical
+  Scenario: GetNetworkDifficulty
+    Given I have a seed node NODE
+    And I have wallet WALLET connected to all seed nodes
+    And I have a merge mining proxy PROXY connected to NODE and WALLET
+    When I merge mine 2 blocks via PROXY
+    Then all nodes are at height 2
+    When I request the difficulties of a node NODE
+    Then Difficulties are available
```
stringhandler, line 12:
```suggestion
    Then the last network difficulty is 1
```

stringhandler (commented):
Happy, just a suggestion for a better name

integration_tests/features/BlockExplorerGRPC.feature
StriderDM: Hmm, while the values of only the first block is checked (reverse ordered so it at the end) as we know it will always be the same, it does also check that there are more than just that value in the array. 

Would you still like this changed?

stringhandler (approved):

PR #2616, opened 2021-02-05
submit_block returns OK if block was submitted for tari or monero

sdbondi:
## Description

Pool software expects for the submit block call to succeed if a block it
just mined was able to be submitted to the blockchain. Previously, if
the block was successfully submitted to the tari blockchain but not the
monero blockchain, the call would fail.

This PR changes the behaviour of submit_block to succeed if either a
tari or monero block succeeded.

## Motivation and Context

As in description.

## How Has This Been Tested?

Tested that the call succeeds on the pool-side and successfully mined blocks are noted by the pool.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/block_template_data.rs
+++ b/applications/tari_merge_mining_proxy/src/block_template_data.rs
@@ -66,46 +66,34 @@ impl BlockTemplateRepository {
             hex::encode(hash.as_ref())
         );
         let b = self.blocks.read().await;
-        match b.get(hash.as_ref()).cloned() {
-            Some(item) => Some(item.data),
-            None => None,
-        }
+        b.get(hash.as_ref()).map(|item| item.data.clone())
     }
 
-    pub async fn save<T: AsRef<[u8]>>(&mut self, hash: T, block_template: BlockTemplateData) {
+    pub async fn save(&mut self, hash: Vec<u8>, block_template: BlockTemplateData) {
         trace!(
             target: LOG_TARGET,
             "Saving blocktemplate with merge mining hash: {:?}",
-            hex::encode(hash.as_ref())
+            hex::encode(&hash)
         );
         let mut b = self.blocks.write().await;
         let repository_item = BlockTemplateRepositoryItem::new(block_template);
-        b.insert(Vec::from(hash.as_ref()), repository_item);
+        b.insert(hash, repository_item);
     }
 
     pub async fn remove_outdated(&mut self) {
         trace!(target: LOG_TARGET, "Removing outdated blocktemplates");
         let mut b = self.blocks.write().await;
         let threshold = Utc::now() - Duration::minutes(20);
-        for item in b.clone().iter() {
-            if item.1.datetime() < threshold {
-                trace!(
-                    target: LOG_TARGET,
-                    "Blocktemplate removed with merge mining hash {:?}",
-                    hex::encode(item.0)
-                );
-                b.remove(item.0);
-            }
-        }
+        *b = b.drain().filter(|(_, i)| i.datetime() >= threshold).collect();
     }
 
     pub async fn remove<T: AsRef<[u8]>>(&mut self, hash: T) {
-        let mut b = self.blocks.write().await;
         trace!(
             target: LOG_TARGET,
             "Blocktemplate removed with merge mining hash {:?}",
             hex::encode(hash.as_ref())
         );
+        let mut b = self.blocks.write().await;
         b.remove(hash.as_ref());
     }
 }
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -24,6 +24,7 @@ use crate::{
     block_template_data::{BlockTemplateDataBuilder, BlockTemplateRepository},
     error::MmProxyError,
     helpers,
+    json_rpc,
 };
 use bytes::BytesMut;
 use futures::{StreamExt, TryFutureExt};
@@ -40,6 +41,7 @@ use hyper::{
     Uri,
     Version,
 };
+use json::json;
 use jsonrpc::error::StandardError;
 use reqwest::{ResponseBuilderExt, Url};
 use serde_json as json;
@@ -55,6 +57,7 @@ use std::{
         Arc,
     },
     task::{Context, Poll},
+    time::Instant,
 };
 use tari_app_grpc::{tari_rpc as grpc, tari_rpc::GetCoinbaseRequest};
 use tari_common::{GlobalConfig, Network};
@@ -62,10 +65,11 @@ use tari_core::{
     blocks::{Block, NewBlockTemplate},
     proof_of_work::monero_rx,
 };
-use tokio::runtime::Handle;
 use tracing::{debug, error, info, instrument, trace, warn};
 
-pub const LOG_TARGET: &str = "tari_mm_proxy::proxy";
+const LOG_TARGET: &str = "tari_mm_proxy::proxy";
+/// The JSON object key name used for merge mining proxy response extensions
+pub(crate) const MMPROXY_EXTENSIONS_KEY_NAME: &str = "_merge_mining";
 
 #[derive(Debug, Clone)]
 pub struct MergeMiningProxyConfig {
@@ -225,7 +229,15 @@ impl Service<Request<Body>> for MergeMiningProxyService {
 
                     Ok(Response::builder()
                         .status(StatusCode::INTERNAL_SERVER_ERROR)
-                        .body(standard_rpc_error(StandardError::InternalError, None))
+                        .body(
+                            json::to_string(&json_rpc::standard_error_response(
+                                None,
+                                StandardError::InternalError,
+                                None,
+                            ))
+                            .expect("unexpected failure")
+                            .into(),
+                        )
                         .unwrap())
                 },
             }
@@ -285,7 +297,7 @@ impl InnerService {
             "Monero height = #{}, Tari base node height = #{}", json["height"], height
         );
 
-        json["height"] = json::json!(cmp::max(json["height"].as_i64().unwrap_or_default(), height as i64));
+        json["height"] = json!(cmp::max(json["height"].as_i64().unwrap_or_default(), height as i64));
 
         Ok(into_body(parts, json))
     }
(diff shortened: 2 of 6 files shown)


applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -317,75 +320,77 @@ impl InnerService {
...
+            header_mut.pow.as_mut().unwrap().pow_data = bincode::serialize(&monero_data)?;
+
+            let mut base_node_client = self.connect_grpc_client().await?;
+            let start = Instant::now();
+            match base_node_client.submit_block(block_data.tari_block).await {
+                Ok(_) => {
+                    json_resp["result"]["tari_status"] = json!("OK");
+
+                    // If monero returned an error, remove it
+                    if !json_resp["error"].is_null() {
```
StriderDM: Return `accept_response` here, it is in the format which xmrig expects (as well as other consumers of the MoneroD RPC interface). See line 644 on how to construct it. Was going to embed a message into the json as well to indicate that it was accepted to Tari which we could optionally display in XMRig (or other consumers).

StriderDM (changes requested):

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: The json rpc format is still adhered to (by monerod), we are just changing the contents of the monero response. Happy to change it to create a new response (getting rid of the `"error": null` field at least)

StriderDM: I haven't tested it but I think `"error": null` could potentially cause problems for some applications. 

Anything that checks for the "presence of" could result in undefined behaviour in the consumer. If it also doesn't necessarily check the null (as it is not necessarily expected here) and tries to access as normal, will result in a segfault.

While this is actually a problem from the side of the consumer application, it brings with the complaint of "But it works everywhere else except here" from a user's point of view.

Thoughts?

sdbondi: Yeah, I've removed `"error": null`.

I've just noticed "special code" for submit_block in the proxy_request_to_monerod function. We probably want to consolidate functionality to the handler for submit_block. Will update on this PR with what I've changed

StriderDM: The "special code" you're referring to is probably related to using XMRig in self-select mode.

applications/tari_merge_mining_proxy/src/block_template_data.rs
```diff
@@ -66,46 +66,34 @@ impl BlockTemplateRepository {
...
+    pub async fn save(&mut self, hash: Vec<u8>, block_template: BlockTemplateData) {
         trace!(
             target: LOG_TARGET,
             "Saving blocktemplate with merge mining hash: {:?}",
-            hex::encode(hash.as_ref())
+            hex::encode(&hash)
         );
         let mut b = self.blocks.write().await;
         let repository_item = BlockTemplateRepositoryItem::new(block_template);
-        b.insert(Vec::from(hash.as_ref()), repository_item);
```
sdbondi, line 83: Extra: Fix unnecessary clone

applications/tari_merge_mining_proxy/src/block_template_data.rs
```diff
@@ -66,46 +66,34 @@ impl BlockTemplateRepository {
...
+        b.insert(hash, repository_item);
     }
 
     pub async fn remove_outdated(&mut self) {
         trace!(target: LOG_TARGET, "Removing outdated blocktemplates");
         let mut b = self.blocks.write().await;
         let threshold = Utc::now() - Duration::minutes(20);
-        for item in b.clone().iter() {
```
sdbondi, line 90: Extra: Fix unnecessary clone of entire block hashmap

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: I've kept the `submit_block` and `proxy_submit_to_origin == false` special case as it is not within the scope of this PR to change anything there. 

In the case where a pool is using the Tari merge mining proxy, the behaviour is as follows:
Block is submitted to monero (if proxy_submit_to_origin == true)
Block is submitted to tari
if monero and/or tari succeeds, [return a success result](https://github.com/tari-project/tari/pull/2616/files#diff-c12374a32500513f6277a01e907665e87a10f680738b198f60bab0213f4e1294R373)
if monero and tari fail, return an error result
if proxy_submit_to_origin == false, and tari fails, then we [return an error](https://github.com/tari-project/tari/pull/2616/files#diff-c12374a32500513f6277a01e907665e87a10f680738b198f60bab0213f4e1294R398) (even though the response is success for monerod because of aforementioned `proxy_submit_to_origin == false`)

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -567,6 +598,8 @@ impl InnerService {
             "Difficulties: Tari ({}), Monero({}), Selected({})", tari_difficulty, monero_difficulty, mining_difficulty
         );
         monerod_resp["result"]["difficulty"] = mining_difficulty.into();
+        monerod_resp["result"]["difficulty_tari"] = tari_difficulty.into();
+        monerod_resp["result"]["difficulty_monero"] = monero_difficulty.into();
```
StriderDM: Reason for adding these to the response? They will not be used anywhere by monero related software.

Should we change the rpc interface, we will have to do it in a way that is generic and possibly submit changes to the monero codebase to account for them.

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -296,96 +305,118 @@ impl InnerService {
...
+
+            let mut base_node_client = self.connect_grpc_client().await?;
+            let start = Instant::now();
+            match base_node_client.submit_block(block_data.tari_block).await {
+                Ok(_) => {
+                    json_resp = json_rpc_success_response(
+                        json_resp["id"].as_i64(),
+                        json!({
+                            "status": json_resp["result"]["status"],
+                            "tari_status": "OK",
```
StriderDM: "merge_mining": "Ok"

Has to be generic enough to be used by any coin.

applications/tari_merge_mining_proxy/src/block_template_data.rs
StriderDM: Nice

applications/tari_merge_mining_proxy/src/block_template_data.rs
StriderDM: Nice

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -791,3 +807,45 @@ pub(super) async fn read_body_until_end(body: &mut Body) -> Result<BytesMut, MmP
...
+    err_data: Option<json::Value>,
+) -> json::Value
+{
+    json!({
+        "id":  req_id.unwrap_or(-1),
+        "jsonrpc": "2.0",
+        "error": {
+            "code": err_code,
+            "message": err_message,
+            "data": err_data,
```
StriderDM: See previous comment.

StriderDM (commented):
Little concerned we are adding a number of values to standardized methods documented here:
https://www.getmonero.org/resources/developer-guides/daemon-rpc.html

Changed here should be sufficient (excludes the clone fixes):
https://github.com/StriderDM/tari/commit/ca3b2d5b7f2369e8dba8f81d069cc6876a51cff7

I am assuming this is for debugging with the node-js pool?

Otherwise it looks good.

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: This is a standard JSON RPC error https://www.jsonrpc.org/specification#error_object

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: There is currently no way to tell which difficulty was achieved.

JSON is a pretty flexible format, adding fields cannot break any API consumer so these are non breaking changes. There is no need to change the monerod rpc. If a pool (or any mm proxy api consumer) supports it, it can use mm proxy extension data in calls, otherwise it will not be aware of the added fields.

My thoughts are that we can have a consistent key for MM proxy extensions.

Something like

```javascript
{
  "id": 123,
  ... etc
  "result": {
    "difficulty": ....,
    "_merge_mining": { // underscore because non-standard, non-stable
      "difficulty_tari": ....,
      "difficulty_monero": ..... 
   }
  }
}

File: applications/tari_merge_mining_proxy/src/proxy.rs (hunk 5 above)
Comment (sdbondi) [reply applications/tari_merge_mining_proxy/src/proxy.rs]: I think the goal here is to have an API that can be used by a supporting consumer that is using the MM proxy, not to create a generic solution to merge mining with Monero in general (although the learnings from this could be used to create that spec). 

Perhaps something like

```javascript
{
  "id": 123,
  ... etc
  "result": {
    "status": "OK", // <- this means the call worked, the block was accepted by monero/tari
    "_merge_mining": { // underscore because non-standard, non-stable 
      "statuses": {"xtr": "ACCEPTED", "xmr": "REJECTED"} 
   }
  }
}
```

applications/tari_merge_mining_proxy/src/proxy.rs
StriderDM: Through the logs we can tell which difficulty was targeted.

Is it definitely needed since we would be submitting to the proxy at least if the template difficulty was achieved?

Template difficulty through the proxy will always be `min(tari,monero)` in the current code.

I do agree with your suggestion of using underscores to denote non-standard fields, well thought out.

applications/tari_merge_mining_proxy/src/proxy.rs
StriderDM: With the exception of the tari specific grpc calls we make witin the proxy, the solution is pretty generic already.

I agree with your above suggestion, that way statuses can just be displayed off an iterator. Nice.

applications/tari_merge_mining_proxy/src/proxy.rs
StriderDM: True, though Monero does have errors mapped to codes and messages depending on what is wrong:
```
CORE_RPC_ERROR_CODE_WRONG_BLOCKBLOB
CORE_RPC_ERROR_CODE_WRONG_PARAM
CORE_RPC_ERROR_CODE_TOO_BIG_HEIGHT
CORE_RPC_ERROR_CODE_TOO_BIG_RESERVE_SIZE
CORE_RPC_ERROR_CODE_WRONG_WALLET_ADDRESS
CORE_RPC_ERROR_CODE_MINING_TO_SUBADDRESS
CORE_RPC_ERROR_CODE_INTERNAL_ERROR
CORE_RPC_ERROR_CODE_BLOCK_NOT_ACCEPTED
```

The definitions can be found in:
```
core_rpc_server_error_codes.h
```

And usage in:
```
core_rpc_server.cpp
```

Assuming configuration is correct, we will only get this error on failure:
```
CORE_RPC_ERROR_CODE_BLOCK_NOT_ACCEPTED
```

I'm not against using the above however, perhaps you want to underscore data?

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: In the pool context, we need to keep track of the payout accounting for monero and tari. So we need some way to tell which block was won. We could also add a custom extension call that gives us this info, but I feel that since get_block_template gives you the required mining difficulty, it may as well provide metadata for difficulties on the merge mined and parent chains. Open to thoughts on this though (however there is nothing stopping us from changing this later as the pool integration is ongoing and subject to change so I don't think there is too much value in attempting to get the "perfect api interface" in this PR).  

EDIT:
To keep this in line with other `_merge_mining` extensions
```js
{
  "_merge_mining": {
    "difficulties": {"xtr": ...., "xmr": .......}
  }
}
```

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: These helpers allow you to create any JSON RPC success or error response object. They have nothing to do with monerod or the submit_block call. I've added comments to these functions.

The submit_block call uses this function to create the "block not accepted" error response object [see here](https://github.com/tari-project/tari/pull/2616/files#diff-c12374a32500513f6277a01e907665e87a10f680738b198f60bab0213f4e1294R404)

The `data` field is part of the [JSON RPC spec](https://www.jsonrpc.org/specification#error_object), so it can stay.

applications/tari_merge_mining_proxy/src/proxy.rs
sdbondi: I've decided not to include any extension data in submit_block call since it probably isn't needed and it keeps things simple. Especially since the monerod api allows multiple blocks to be submitted in one call. 

The only change made to this call is to return a success result if Tari and/or Monero succeed.

StriderDM (dismissed):
LGTM

stringhandler (approved):

PR #2626, opened 2021-02-09
Add tx validator check for duplicate inputs and outputs

SWvheerden:
## Description

This adds in a check for checking if a received tx had duplicated inputs and outputs.

## Motivation and Context

A tx cannot have duplicated inputs and outputs. 

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -170,6 +170,66 @@ impl AggregateBody {
             .any(|input| self.outputs.iter().any(|o| o.is_equal_to(input)))
     }
 
+    /// This will check that there are no duplicated inputs. If the body is not sorted, this will sort the body
+    pub fn contains_duplicated_inputs(&self) -> bool {
+        // This can be expensive so we add in a sort check if its sorted
+        if self.sorted {
+            return self.contains_duplicated_inputs_sorted();
+        }
+        for i in 0..self.inputs().len() {
+            for j in (i + 1)..self.inputs().len() {
+                if self.inputs()[i] == self.inputs()[j] {
+                    return true;
+                }
+            }
+        }
+        false
+    }
+
+    /// This will check that there are no duplicated outputs. If the body is not sorted, this will sort the body
+    pub fn contains_duplicated_outputs(&self) -> bool {
+        // This can be expensive so we add in a sort check if its sorted
+        if self.sorted {
+            return self.contains_duplicated_outputs_sorted();
+        }
+        for i in 0..self.outputs().len() {
+            for j in (i + 1)..self.outputs().len() {
+                if self.outputs()[i] == self.outputs()[j] {
+                    return true;
+                }
+            }
+        }
+        false
+    }
+
+    /// This will check that there are no duplicated outputs, this will only work on a sorted transaction, will return
+    /// true if its not sorted. If the body is not sorted, this will sort the body
+    pub fn contains_duplicated_outputs_sorted(&self) -> bool {
+        if !self.sorted {
+            return true;
+        }
+        for i in 1..self.outputs().len() {
+            if self.outputs()[i] == self.outputs()[i - 1] {
+                return true;
+            }
+        }
+        false
+    }
+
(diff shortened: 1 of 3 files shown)


base_layer/core/src/transactions/transaction.rs
```diff
@@ -1095,6 +1095,27 @@ mod test {
         assert_eq!(tx3.body.kernels().len(), 2);
     }
 
+    #[test]
+    fn check_duplicate_inputs_outputs() {
+        let (mut tx, _, outputs) = create_tx(50000000.into(), 15.into(), 1, 2, 1, 2);
+        dbg!(&tx.body.check_duplicated_outputs());
+        println!("tx: {}", tx);
```
delta1: forgot these

base_layer/core/src/validation/transaction_validators.rs
```diff
@@ -125,6 +126,20 @@ fn verify_not_stxos<B: BlockchainBackend>(tx: &Transaction, db: &B) -> Result<()
     Ok(())
 }
 
+/// This function checks the at the tx contains no duplicated inputs or outputs.
+fn verify_no_duplicated_inputs_outputs(tx: &Transaction) -> Result<(), ValidationError> {
+    let mut mut_copy = tx.clone();
+    if !mut_copy.body.check_duplicated_inputs() {
```
delta1: Personal opinion: I think these should be named like `contains_duplicate_inputs` and swap around the returning of true and false. 

Then this line becomes `if mut_copy.body.contains_duplicate_inputs()` which I think is more intuitive.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -170,6 +170,28 @@ impl AggregateBody {
             .any(|input| self.outputs.iter().any(|o| o.is_equal_to(input)))
     }
 
+    /// This will check that there are no duplicated inputs. If the body is not sorted, this will sort the body
+    pub fn contains_duplicated_inputs(&mut self) -> bool {
```
stringhandler: This can be achieved without the need for `mut`. Later in the PR you convert a borrow into a `mut` which I'm surprised rust lets you do. I think a better approach would be to either do a n^2 search for duplicates or to make a reference array that is sorted `(n*log(n))`.

Alternatively, you could check if it's already sorted and do the linear check, and do the expensive n^2 check if it's not.

stringhandler (commented):
I agree with the validation, but I'm surprised you managed to get this to compile. The borrow checker should have denied the conversion of a reference into a mut. If we can change the validation step to only use references (i.e. not change the structure) I'd be happy to merge it.

Calling validation should not have any side effects

base_layer/core/src/transactions/aggregated_body.rs
delta1: Isn’t it making a mutable copy? The original remains

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -170,6 +170,66 @@ impl AggregateBody {
...
+            }
+        }
+        false
+    }
+
+    /// This will check that there are no duplicated outputs, this will only work on a sorted transaction, will return
+    /// true if its not sorted. If the body is not sorted, this will sort the body
+    pub fn contains_duplicated_outputs_sorted(&self) -> bool {
+        if !self.sorted {
+            return true;
```
stringhandler, line 209: This method should have been inlined and should definitely not be public. It's a bunch of extra code and now has a logic branch that makes no sense.

stringhandler (approved):
I'm merging this because the code is needed, but there is really no need to extra this logic out into it's own method as per my comments. I might do a follow up PR to fix this.

base_layer/core/src/transactions/aggregated_body.rs
sdbondi: Yeah, we should do something similar to what we do with the block body validation (i.e. transaction must be sorted on arrival, validate sorted and no dups in one loop)

PR #2631, opened 2021-02-10
[fix] Fix missing Content-Type header in some json responses

sdbondi:
Adds `Content-Type: application/json` to 'custom' proxy responses.
This is technically more correct and some clients do not interpret the
contents as JSON without it.


delta1:
Clippy loves you like a crazy stalker @sdbondi 😆

sdbondi:
😆  At least that one makes sense - fixed

delta1 (dismissed):
LGTM

PR #2632, opened 2021-02-10
Update TXO validation protocol to use RPC interface to base node

philipr-za:
# Attention:
1. This PR will only work when it is also deployed to the base nodes
2. This PR is based on https://github.com/tari-project/tari/pull/2614 as it needs part of that, so this is not as big as it looks and once that PR is in this can be rebased.

## Description
Currently the TXO validation in the Output Manager is done using the fire-and-forget messages used in the P2P comms stack. This PR updates this protocol to use an RPC interface for more deterministic interactions with the base node. 

The `fetch_matching_utxos` method is added to the Base Node’s Wallet RPC interface and the Output Manager’s TXO validation is updated to use it. This involves some refactor of the launching of the TXO validation protocol in the Output Manager Service to support the new style of protocol but the functionality is the same.

## How Has This Been Tested?
Tests are provided to test the new protocol for spent, invalid and unspent outputs and how the protocol behaves when switching base nodes.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:


delta1:
Looks like this needs a rebase

PR #2636, opened 2021-02-10
Add Nodejs block explorer

stringhandler:
## Description

Add a no-script, simple NodeJs block explorer to help debugging. Also update the mempool GetState request to return all transactions instead of just the excess, and expose it via GRPC.

image

## Motivation and Context

The current block explorer is pretty, but it doesn't provide much info on the contents of the blocks or the mempool. I was trying to fix the problem with horizon sync and wanted to get the UTXOs and kernels in a block, but there was no easy interface.


delta1 (dismissed):
Awesome great work

PR #2637, opened 2021-02-11
Update Transaction Receiver protocol to persist transaction earlier

philipr-za:
## Description
Currently when a wallet receives a transaction it attempts to send the reply before persisting the received transaction data to the database. Send the reply can take a while so in the iOS application where this process was done by the constrained background service the process was closing before the data was persisted and so it was not available when the user opened the app.

This PR makes a simple change where the data is persisted as soon as possible and only then the reply is attempted to be sent. If the process closes before the reply is sent it is not an issue. After a reply is sent a timestamp of when it was sent is written to the database. If the wallet is opened later it checks to see if the reply sent timestamp exists or was a long time ago and if it wasn’t sent recently it will send the reply again.

## How Has This Been Tested?
Existing Tests

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
@@ -155,20 +155,20 @@ where TBackend: TransactionBackend + 'static
                 Utc::now().naive_utc(),
             );
 
+            self.resources
+                .db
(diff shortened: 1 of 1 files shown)


delta1 (approved):
LGTM

sdbondi (approved):
Makes sense - LGTM

stringhandler (approved):

PR #2638, opened 2021-02-11
[common] Ban peers that flood messages

sdbondi:
## Description

Ban a peer if the number of messages received over a timespan
(`DhtConfig::flood_ban_timespan`) exceeds a threshold
(`DhtConfig::flood_ban_max_msg_count`).

The current config defaults are an average of 10mps over 100 seconds result in banning for 6 hours.
(`flood_ban_max_msg_count = 1000, flood_ban_timespan = 100s`)

**How it works**

A layer is added to the incoming pipeline that queues up metrics
(producer) for the metric collector (consumer).
The metric collector is a consumer that reads from a channel and updates
it's in-memory state. The collector is lock-free.

The Dht connectivity actor periodically (2 minutes) queries the
collector for peer's exceeding the configured rate and bans them.

## Motivation and Context

A peer should not be allowed to flood another peer without being 
forcefully disconnected for a period of time.

The metrics collector is written in a way that it could be expanded to collect 
more metrics which could be useful for heuristic-based peer banning, monitoring 
and system visibility. 

## How Has This Been Tested?

Some basic unit tests, tested on base node that is being flooded by other base nodes

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/config.rs
+++ b/comms/dht/src/config.rs
@@ -102,9 +102,18 @@ pub struct DhtConfig {
     pub network: Network,
     /// Network discovery config
     pub network_discovery: NetworkDiscoveryConfig,
-    pub short_ban_duration: Duration,
+    /// Length of time to ban a peer if the peer misbehaves at the DHT-level.
+    /// Default: 6 hrs
+    pub ban_duration: Duration,
     /// This allows the use of test addresses in the network.
     pub allow_test_addresses: bool,
+    /// The maximum number of messages over `flood_ban_timespan` to allow before banning the peer (for `ban_duration`)
+    /// Default: 1000 messages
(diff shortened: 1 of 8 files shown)


stringhandler (approved):
This potentially could end up banning the whole network in a bug, but since it's configurable I think we could recover from that situation manually if needed

sdbondi:
@mikethetike in the current environment my node had 0 base node connections after a few minutes. Either this will quench the source of the issue and some nodes will still be connected to each other (reorgs ensue) on a quiet network, or all nodes will not connect to any other node. The main reason to have this would be to prevent a node from starting a flood in the first place. Just noting here that this is not a fix for the flooding issue.

PR #2643, opened 2021-02-14
[common] Don't repropagate SAF messages if we have already propagated them

stringhandler:
## Description

Add a flag to determine if we have already tried to propagate this message

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/config.rs
+++ b/comms/dht/src/config.rs
@@ -24,7 +24,7 @@ use crate::{envelope::Network, network_discovery::NetworkDiscoveryConfig, storag
 use std::time::Duration;
 
 /// The default maximum number of messages that can be stored using the Store-and-forward middleware
-pub const SAF_MSG_STORAGE_CAPACITY: usize = 10_000;
+pub const SAF_MSG_STORAGE_CAPACITY: usize = 100_000;
 /// The default time-to-live duration used for storage of low priority messages by the Store-and-forward middleware
 pub const SAF_LOW_PRIORITY_MSG_STORAGE_TTL: Duration = Duration::from_secs(6 * 60 * 60); // 6 hours
 /// The default time-to-live duration used for storage of high priority messages by the Store-and-forward middleware
@@ -166,7 +166,7 @@ impl Default for DhtConfig {
             saf_auto_request: true,
             saf_max_message_size: 512 * 1024, // 500 KiB
             saf_minimum_request_period: SAF_HIGH_PRIORITY_MSG_STORAGE_TTL,
-            msg_hash_cache_capacity: 10_000,
+            msg_hash_cache_capacity: 100_000,
             msg_hash_cache_ttl: Duration::from_secs(5 * 60),
             database_url: DbConnectionUrl::Memory,
             discovery_request_timeout: Duration::from_secs(2 * 60),
--- a/comms/dht/src/store_forward/forward.rs
+++ b/comms/dht/src/store_forward/forward.rs
@@ -168,6 +168,7 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
             decryption_result,
             dht_header,
             is_saf_stored,
+            is_already_forwarded,
             ..
         } = message;
 
@@ -199,15 +200,28 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
         let mut send_params = SendMessageParams::new();
         match (dest_node_id, is_saf_stored) {
             (Some(node_id), Some(true)) => {
+                debug!(
+                    target: LOG_TARGET,
+                    "Forwarding SAF message directly to node: {}, Tag#{}", node_id, dht_header.message_tag
+                );
                 send_params.closest_connected(node_id.clone(), excluded_peers);
             },
             _ => {
+                debug!(
+                    target: LOG_TARGET,
+                    "Not storing this SAF message for {}, propagating it. Tag#{}",
+                    dht_header.destination,
+                    dht_header.message_tag
+                );
+
                 send_params.propagate(dht_header.destination.clone(), excluded_peers);
             },
         };
 
-        send_params.with_dht_header(dht_header.clone());
-        self.outbound_service.send_raw(send_params.finish(), body).await?;
+        if !is_already_forwarded {
+            send_params.with_dht_header(dht_header.clone());
+            self.outbound_service.send_raw(send_params.finish(), body).await?;
+        }
 
         Ok(())
     }
--- a/comms/dht/src/store_forward/service.rs
+++ b/comms/dht/src/store_forward/service.rs
@@ -90,7 +90,7 @@ impl FetchStoredMessageQuery {
 #[derive(Debug)]
 pub enum StoreAndForwardRequest {
     FetchMessages(FetchStoredMessageQuery, oneshot::Sender<SafResult<Vec<StoredMessage>>>),
-    InsertMessage(NewStoredMessage),
+    InsertMessage(NewStoredMessage, oneshot::Sender<SafResult<bool>>),
     RemoveMessages(Vec<i32>),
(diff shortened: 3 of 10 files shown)


sdbondi:
I have my doubts that this code causes the flood as described here. The distance metric is essentially a big (104-bit) random number (XOR metric), so it is impossible given 3 different node IDs that the == case would ever come into it (`given  distinct A, B, C  A^B != A^C != B^C`). We are seeing an encrypted propagated message with the body probably changing on each hop/rebroadcast with a different nonce on arrival at dest. We can see this because we are observing unique body hashes in the stored_messages db table. stored_messages has a unique constraint on the `body_hash` field, and essentially provides another check for duplicates (insofar as, if the number of rows increase, we have many unique encrypted messages). This means that the dedup layer will let messages through. We need to figure out why messages are changing slightly on each hop or why one message in results in many unique messages out. 

Some observations:
- All messages are encrypted for some other node, the node cannot decrypt them and therefore stores and propagates them
- The stored_messages table grows and has many unique body hashes. This should mean that there are unique messages being sent i.e. the domain layer is sending them - however it only reaches the domain layer if it can be decrypted. Obviously there could also be a DHT level bug, but it should be easier to rule out a domain level bug (is the wallet receiving many txns, what is it doing with those messages?).
- My intuition on the cause without digging into it, is the receipt of many messages via SAF to the destination node (i.e. can be decrypted) results one/many unique messages being sent out (some nonce is changed) which results in messages being stored in SAF, which results in more messages arriving at the destination. Something like that, but it's difficult to know without digging into the flood of logs.

stringhandler:
@sdbondi I've updated this, I think that the problem is actually more prevalent when the message is stored. I have updated this to only try to send directly to the target node id, and then also only send it if it's the first time it is stored. 

Tests with memory net return working results with all SAF's working with less messages overall (average 60 per node)

comms/dht/src/store_forward/forward.rs
```diff
@@ -168,6 +168,7 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
             decryption_result,
             dht_header,
             is_saf_stored,
+            is_already_forwarded,
             ..
         } = message;
```
sdbondi, line 174: Suggest adding the check here and exiting early if `is_already_forwarded == true`

comms/dht/src/store_forward/forward.rs
```diff
@@ -199,15 +200,27 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
         let mut send_params = SendMessageParams::new();
         match (dest_node_id, is_saf_stored) {
             (Some(node_id), Some(true)) => {
-                send_params.closest_connected(node_id.clone(), excluded_peers);
+                debug!(
+                    target: LOG_TARGET,
+                    "Forwarding SAF message directly to node: {}, Tag#{}", node_id, dht_header.message_tag
+                );
+                send_params.direct_node_id(node_id.clone());
```
sdbondi: Little concerned we will drop messages on the floor if we fail to get in direct contact with the dest node. If we're connected to the peer then closest_connected will send it to the dest, otherwise peers that are closer will receive+store the message (remember that it is possible for us to consider the peer as a neighbour, but the peer doesnt consider us a neighbour). 

I rather suggest that we add a check for `BroadcastStrategy::Closest` if we're already connected to the dest peer only send to that peer, otherwise send to closer.

comms/dht/src/store_forward/service.rs
```diff
@@ -258,20 +259,35 @@ impl StoreAndForwardService {
...
-                            .map(|p| format!("public key '{}'", p))
-                            .or_else(|| node_id.map(|n| format!("node id '{}'", n)))
-                            .unwrap_or_else(|| "<Anonymous>".to_string())
-                    ),
+                    Ok(existed) => {
+                        if !existed {
+                            info!(
+                                target: LOG_TARGET,
+                                "Stored message for {}",
+                                public_key
```
sdbondi: nit: 
could DRY a bit. Also think debug level is more appropriate.

```rust
let pubkey_str = public_key
                   .map(|p| format!("public key '{}'", p))
                   .or_else(|| node_id.map(|n| format!("node id '{}'", n)))
                   .unwrap_or_else(|| "<Anonymous>".to_string());
if existed {
  debug!(...)
} else {
  debug!(...)
}
```

comms/dht/src/store_forward/forward.rs
sdbondi: Also worth noting that this is much more likely to succeed in memorynet than in a testnet/mainnet network.

comms/dht/src/config.rs
```diff
@@ -152,7 +152,7 @@ impl Default for DhtConfig {
             saf_auto_request: true,
             saf_max_message_size: 512 * 1024, // 500 KiB
             saf_minimum_request_period: SAF_HIGH_PRIORITY_MSG_STORAGE_TTL,
-            msg_hash_cache_capacity: 10_000,
+            msg_hash_cache_capacity: 100_000,
```
sdbondi, line 169: I don't foresee any issues with this

comms/dht/src/config.rs
```diff
@@ -24,7 +24,7 @@ use crate::{envelope::Network, network_discovery::NetworkDiscoveryConfig, storag
 use std::time::Duration;
 
 /// The default maximum number of messages that can be stored using the Store-and-forward middleware
-pub const SAF_MSG_STORAGE_CAPACITY: usize = 10_000;
+pub const SAF_MSG_STORAGE_CAPACITY: usize = 100_000;
```
sdbondi, line 27: Might want to revert this, between 10_000 and 50_000 seems like a more sensible default esp as message perf begins to deteriorate at 65k

sdbondi (commented):
Great stuff, completely agree with using the SAF table as an additional dedup. Some comments.

comms/dht/src/store_forward/forward.rs
stringhandler: I hear you, but I see it as this: 
1. If we are storing the message, the direct node must be able to reach us, otherwise there is no need to store it.
2. If we are storing the message, why would we ask someone who is better to also store it. It's not our message at the end of the day, and therefore not our responsibility to ensure we get it to the destination. We are already storing it on the wallet's behalf or we have forwarded it and I think that's the end of our responsibility.

The fact that it is possible for a peer to not consider us a neighbour is a problem, but that is a different problem. I think if anything, we should reduce the criteria of us storing the message

comms/dht/src/store_forward/forward.rs
sdbondi: Or rather just below the destination_matches_source check. There is no point in building send params that arent used in the end and the code is a little easier to follow (if is already forwarded look no further)

stringhandler:
Removed any controversial bits of this and have only kept the `already_forwarded` flag

sdbondi (approved):
Ship it!

comms/dht/src/store_forward/forward.rs
sdbondi: 1. Not sure I follow the reasoning here. Of course the wallet is _able_ to reach the node (us in this example?) but it may not because it may have other nodes that are closer to it once it came online.
2. Ideally, I agree in those responsibilities - but in order for that network rule to have the level of deliverability we need the storing node to be highly available (always online), perfect network knowledge (which is probably like attempting to have perfect mempool synchronisation), a static network (no nodes coming/going which could effect neighborhoods) -  the further propagation of the message ensures some level of message redundancy which increases the chance that the message will be received via SAF. 

I think that these (fairly small) changes will make the biggest positive impact:
- This PR
- not storing messages for peers that are already connected
- ensuring that the 'send closer' broadcast strat will send directly to the peer if it is already connected (like propagate strat does)

PR #2645, opened 2021-02-15
Update TariScript OpCodes

CjS77:
Updates the TariScript Opcode definitions and examples to match v3 of
the TariScript proposal.

v3 requiers a pubkey to remain on the stack, so all the scripts have
been updated to reflect this. 1 is now defined as True ans 0 as False,
inline with more common conventions.


SWvheerden:
first scan through looks good. Will address any issues in a followup review/PR 
LGTM

PR #2646, opened 2021-02-15
Fix transaction propagation 

SWvheerden:
Set bootstrap default false

## Description

This fixes the transaction propagation and the bootstrap flag.
Fixes transaction builder in cucumber. 
Adds in a cucumber test to test transaction propagation.
Adds cucumber test for double spends. 
Adds cucumber test for mempool clearing
Adds in a GRPC call to search the mempool for a transaction excess signature. 

## Motivation and Context

Transactions are sometimes sent and they dont propagate through he network. The Bootstrap flag is set to false everytime the node enters sync. When the bootstrap flag is set, the node will ignore all blocks and transactions sent to it. It should only ignore these during initial sync and not every sync. 

## How Has This Been Tested?

Integration tests

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

* [x I ran `cargo-fmt --all` before pushing.

### DIFF
--- a/applications/tari_app_grpc/proto/base_node.proto
+++ b/applications/tari_app_grpc/proto/base_node.proto
@@ -49,7 +49,7 @@ service BaseNode {
     // Get network difficulties
     rpc GetNetworkDifficulty(HeightRequest) returns (stream NetworkDifficultyResponse);
     // Get the block template
-    rpc GetNewBlockTemplate(PowAlgo) returns (NewBlockTemplateResponse);
+    rpc GetNewBlockTemplate(NewBlockTemplateRequest) returns (NewBlockTemplateResponse);
     // Construct a new block from a provided template
     rpc GetNewBlock(NewBlockTemplate) returns (GetNewBlockResult);
     // Submit a new block for propagation
@@ -64,10 +64,10 @@ service BaseNode {
     rpc SearchKernels(SearchKernelsRequest) returns (stream HistoricalBlock);
     // Fetch any utxos that exist in the main chain
     rpc FetchMatchingUtxos(FetchMatchingUtxosRequest) returns (stream FetchMatchingUtxosResponse);
-
     // get all peers from the base node
     rpc GetPeers(GetPeersRequest) returns (stream GetPeersResponse);
     rpc GetMempoolTransactions(GetMempoolTransactionsRequest) returns (stream GetMempoolTransactionsResponse);
+    rpc TransactionState(TransactionStateRequest) returns (TransactionStateResponse);
 }
 
 message SubmitBlockResponse {
@@ -84,6 +84,13 @@ message NewBlockTemplateResponse {
     NewBlockTemplate new_block_template = 1;
     bool initial_sync_achieved = 3;
     MinerData miner_data = 4;
+}
+
+/// return type of NewBlockTemplateRequest
+message NewBlockTemplateRequest{
+    PowAlgo algo = 1;
+    //This field should be moved to optional once optional keyword is standard
+    uint64 max_weight = 2;
 }
 
 // Network difficulty response
@@ -288,4 +295,19 @@ message GetMempoolTransactionsRequest {
 
 message GetMempoolTransactionsResponse {
     Transaction transaction = 1;
+}
+
+message TransactionStateRequest {
+    Signature excess_sig  = 1;
+}
+
+message TransactionStateResponse {
+    TransactionLocation result =1;
+}
+
+enum TransactionLocation {
+    UNKNOWN = 0;
+    MEMPOOL = 1;
+    MINED = 2;
+    NOT_STORED = 3;
 }
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -29,6 +29,7 @@ use std::{
     convert::{TryFrom, TryInto},
     sync::Arc,
 };
+
 use tari_app_grpc::{
     tari_rpc,
     tari_rpc::{CalcType, Sorting},
@@ -47,9 +48,9 @@ use tari_core::{
     crypto::tari_utilities::hex::Hex,
     mempool::{service::LocalMempoolService, TxStorageResponse},
     proof_of_work::PowAlgorithm,
-    transactions::transaction::Transaction,
+    transactions::{transaction::Transaction, types::Signature},
 };
-use tari_crypto::tari_utilities::Hashable;
+use tari_crypto::tari_utilities::{message_format::MessageFormat, Hashable};
 use tokio::{runtime, sync::mpsc};
 use tonic::{Request, Response, Status};
 
@@ -398,24 +399,30 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
 
     async fn get_new_block_template(
         &self,
-        request: Request<tari_rpc::PowAlgo>,
+        request: Request<tari_rpc::NewBlockTemplateRequest>,
     ) -> Result<Response<tari_rpc::NewBlockTemplateResponse>, Status>
     {
         let request = request.into_inner();
         debug!(target: LOG_TARGET, "Incoming GRPC request for get new block template");
-        let algo: PowAlgorithm = (request.pow_algo as u64)
+        trace!(target: LOG_TARGET, "Request {:?}", request);
+        let algo: PowAlgorithm = ((request.algo)
+            .ok_or_else(|| Status::invalid_argument("No valid pow algo selected".to_string()))?
+            .pow_algo as u64)
             .try_into()
             .map_err(|_| Status::invalid_argument("No valid pow algo selected".to_string()))?;
         let mut handler = self.node_service.clone();
 
-        let new_template = handler.get_new_block_template(algo).await.map_err(|e| {
-            warn!(
-                target: LOG_TARGET,
-                "Could not get new block template: {}",
-                e.to_string()
-            );
-            Status::internal(e.to_string())
-        })?;
+        let new_template = handler
+            .get_new_block_template(algo, request.max_weight)
+            .await
+            .map_err(|e| {
+                warn!(
+                    target: LOG_TARGET,
+                    "Could not get new block template: {}",
+                    e.to_string()
+                );
+                Status::internal(e.to_string())
+            })?;
 
         let status_watch = self.state_machine_handle.get_status_info_watch();
         let pow = algo as i32;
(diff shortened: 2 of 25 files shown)


applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -264,3 +264,19 @@ message GetMempoolTransactionsRequest {
...
+message TransactionStateRequest {
+    Transaction transaction  = 1;
+}
+
+message TransactionStateResponse {
+    TransactionStateResult result =1;
+}
+
+enum TransactionStateResult {
+    UNKOWN = 0;
```
delta1: spelling: `UNKNOWN`

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -264,3 +264,19 @@ message GetMempoolTransactionsRequest {
...
+
+message TransactionStateRequest {
+    Transaction transaction  = 1;
+}
+
+message TransactionStateResponse {
+    TransactionStateResult result =1;
+}
+
+enum TransactionStateResult {
```
philipr-za: I would call this TransactionLocation, Reorg is not a state?

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
```diff
@@ -216,7 +216,7 @@ pub struct StatusInfo {
 impl StatusInfo {
     pub fn new() -> Self {
         Self {
-            bootstrapped: false,
+            bootstrapped: true,
```
philipr-za: This doesn't make sense. The whole point of Bootstrapped is to indicate that the node has not synced even once since startup. That is why when you first start the node it is in the StartUp state and Bootstrapped = false. It should only flip to True once it has synced at least once.

philipr-za: This is the code that should flip it to true inside the Listening State:
```Rust
if sync_mode.is_lagging() {
                        return StateEvent::FallenBehind(sync_mode);
                    } else {
                        if !shared.bootstrapped_sync && sync_mode == SyncStatus::UpToDate {
                            shared.bootstrapped_sync = true;
                            debug!(
                                target: LOG_TARGET,
                                "Initial sync achieved, bootstrap done: {}", sync_mode
                            );
                        }
                        self.is_synced = true;
                        shared.set_state_info(StateInfo::Listening(ListeningInfo::new(true)));
                    }
```

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
```diff
@@ -74,7 +74,7 @@ impl BlockSync {
             ));
 
             let _ = status_event_sender.broadcast(StatusInfo {
-                bootstrapped: false,
+                bootstrapped: true,
```
philipr-za: This should not be false (same for in the HeaderSync state) but just setting this to true nullifies the whole purpose of the Bootstrapped flag. The current value should be drawn from the Shared state value `shared.bootstrapped_sync`

philipr-za (commented):
This PR breaks the Bootstrapped flag

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
SWvheerden: I think this comes down to, should header sync change this value?

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
SWvheerden: I Think this might have slipped through. I have been changing values left and right to get it to work correctly

base_layer/core/src/base_node/state_machine_service/states/block_sync.rs
philipr-za: nope, only the first time you get to Listening should change it

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -80,6 +80,13 @@ message NewBlockTemplateResponse{
     MinerData miner_data = 4;
 }
 
+/// return type of GetNewBlockTemplate
+message NewBlockTemplateRequest{
+    PowAlgo algo = 1;
+    //This field should be moved to optional once optional keyword is standard
```
stringhandler, line 92: Comment is not needed since all fields are optional in protobuf

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -80,6 +80,13 @@ message NewBlockTemplateResponse{
     MinerData miner_data = 4;
 }
 
+/// return type of GetNewBlockTemplate
```
stringhandler: Comment looks like a bad copy pasta

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -264,3 +271,19 @@ message GetMempoolTransactionsRequest {
 message GetMempoolTransactionsResponse {
     Transaction transaction = 1;
 }
+
+message TransactionStateRequest {
+    Transaction transaction  = 1;
```
stringhandler: Seems heavy to send a whole transaction in. Perhaps just the excess or a hash of a kernel?

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -264,3 +271,19 @@ message GetMempoolTransactionsRequest {
...
+}
+
+message TransactionStateResponse {
+    TransactionLocation result =1;
+}
+
+enum TransactionLocation {
+    UNKNOWN = 0;
+    UNCONFIRMED = 1;
```
stringhandler: Unconfirmed should return 'Mempool' and REORG should probably map to 'NotStored' or 'Mined'. The caller does not care about this internal details. @philipr-za what do you think?

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -529,6 +535,54 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+            })?;
+        let response = match res {
+            TxStorageResponse::UnconfirmedPool => tari_rpc::TransactionStateResponse {
+                result: tari_rpc::TransactionLocation::Unconfirmed.into(),
+            },
+            TxStorageResponse::ReorgPool => tari_rpc::TransactionStateResponse {
+                result: tari_rpc::TransactionLocation::Reorg.into(),
+            },
+            TxStorageResponse::NotStoredAlreadySpent => tari_rpc::TransactionStateResponse {
+                result: tari_rpc::TransactionLocation::Mined.into(),
```
stringhandler: I don't think this mapping is correct

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -529,6 +535,54 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+            .map_err(|e| {
+                error!(target: LOG_TARGET, "Error submitting query:{}", e);
+                Status::internal(e.to_string())
+            })?;
+        let response = match res {
+            TxStorageResponse::UnconfirmedPool => tari_rpc::TransactionStateResponse {
+                result: tari_rpc::TransactionLocation::Unconfirmed.into(),
+            },
+            TxStorageResponse::ReorgPool => tari_rpc::TransactionStateResponse {
+                result: tari_rpc::TransactionLocation::Reorg.into(),
```
stringhandler: I think this is more likely that the tx is Mined at this point. I'm sure there is logic similar to this in the base node already, perhaps on the RPC side

base_layer/core/src/base_node/comms_interface/comms_request.rs
```diff
@@ -53,7 +53,7 @@ pub enum NodeCommsRequest {
     FetchBlocksWithKernels(Vec<Signature>),
     FetchBlocksWithStxos(Vec<Commitment>),
     FetchBlocksWithUtxos(Vec<Commitment>),
-    GetNewBlockTemplate(PowAlgorithm),
+    GetNewBlockTemplate(PowAlgorithm, u64),
```
stringhandler: This should be changed to a struct with named params

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -132,11 +136,28 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static
...
+                Some(ref v) => {
+                    let status_watch = v.get_status_info_watch();
+                    if (*status_watch.borrow()).bootstrapped {
+                        break;
+                    }
+                    debug!(
+                        target: LOG_TARGET,
+                        "Mempool sync still on hold, waiting for bootstrap to finish",
+                    );
+                    tokio::time::delay_for(Duration::from_secs(60)).await;
```
stringhandler: This would be more efficient as a `select!` macro with a timeout

integration_tests/features/Mempool.feature
```diff
@@ -0,0 +1,83 @@
+@propagation
+Feature: Mempool
+
+ 
+  Scenario: Transactions are propagated through a network
+    Given I have 10 seed nodes
+    And I have a base node SENDER connected to all seed nodes
+    And I have 10 base nodes connected to all seed nodes
+    When I wait 100 seconds
```
stringhandler, line 10: These waits need justifications in comments so that they can be improved upon in future. Describe why the wait is 100 seconds so that a future dev can know where to start optimizing

stringhandler (commented):
I like the tests and the concept, but the transaction state should ask both the mempool and the chain for where a transaction is. There are lots of other style nits and a mapping error. 

I'm sure this exact task of asking where a transaction kernel is has been done in the base node RPC, but I might be wrong. The logic should not differ between these implementations

applications/tari_app_grpc/proto/base_node.proto
SWvheerden: Fields for messages are optional, but not primitives:
This was released 5 days ago:
https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
SWvheerden: Its what the mempool sends back if a transaction is already spent. If its spent, it has to be mined.

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
SWvheerden: The wallets have proper mapping for this. But the BaseNode's not. All the test just assume if the mempool says accepted, its in the mempool and or mined. .

applications/tari_app_grpc/proto/base_node.proto
philipr-za: This is the route taken in the Base Node RPC service exposed to the wallet because we want to keep the idea of the Reorg pool as an internal idea. So I would think that this should also be the case in the gRPC which is an external interface. I understand that in the test you want to check how the system reacts to Double Spends vs Orphans but that can be done in an internal Rust test while the Cucumber test should be an Integration test that tests the system as if you are interacting with it from the outside as it will be in the normal operation. 

In the Cucumber test you can create a double spend and see that it is rejected correctly.

integration_tests/features/Mempool.feature
```diff
@@ -0,0 +1,83 @@
...
+    When I mine a block on SENDER with coinbase CB1
+    When I mine 2 blocks on SENDER
+    When I create a transaction TX1 spending CB1 to UTX1
+    When I submit transaction TX1 to SENDER
+    Then SENDER has TX1 in its mempool inside of UNCONFIRMED pool
+    When I wait 100 seconds
+    Then TX1 is in the UNCONFIRMED pool of all nodes
```
philipr-za: It is new stuff Hansie and I have started to use in the Cucumber steps but look at the `I wait` type steps in https://github.com/tari-project/tari/pull/2681 and https://github.com/tari-project/tari/pull/2686 to see a better way to do this with the time constraint so you don't have to wait a 100 second each time but wait until the condition has been met OR the max time has passed.

stringhandler (dismissed):

Issue #2620, opened 2021-02-05
 API extensions for Tari merge mining proxy


sdbondi:
## [get_block_header_by_hash](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_block_header_by_hash)

**Purpose of call in nodejs-pool**

Called to check the block still exists (i.e. has not been reorged out) and that other details (such as achieved difficulty, block reward) match.

**Current Proxy Behaviour**
No transformations, proxies "as is"

**Proposed behaviour**
Add a `params.auxChainId` field to the RPC request that contains the ID of the chain to query (this is typically the genesis block hash of the chain). To maintain backward compatibility, if this field is omitted, monerod is called as usual. Otherwise the supported chain (currently the proxy only supports a single chain) with the corresponding genesis hash is queried. 

Request: 

```json
"params": {
   "auxChain": "xxxxxxxx",
   (...)
}
```

## [get_block_template](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_block_template)
*Example*

**Aux chain fields**

- _aux.chains[0].id* - string; The name of the coin being included (xtr)
- _aux.chains[0].difficulty* - unsigned int; Difficulty of next  block  for  this chain
- _aux.chains[0].height* - unsigned int; Height this block will be if added to the chain
- _aux.chains[0].mining_hash -  string; Hash of aux block (provided  by chain)
- _aux.chains[0].miner_reward - unsigned_int

```json
"result" : {
  "_aux": {
    "chains": [{
       "id": "xtr",
       "difficulty": 724633,
       "height": 11234,
       "mining_hash": "ab65c22c92619ce2e2037757df5a3cff51b3cab06ea9d3a4c76e49e491e8b557",
       "miner_reward" : 5678123123
    }]
}
```

## [submit_block](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#submit_block)

Submit the mined block to the blockchain. Due to the way the merge mining proxy works, this can be at either the monero and/or tari difficulty.

**Current Proxy Behaviour**
If the block does new achieve the monero difficulty, the call fails (however the block is still submitted to tari).

**Proposed behaviour**
Call should succeed if the block is successfully submitted to monero and/or tari. If monero and tari fail, the call must fail as before.

** Aux chain data **

```json
"result" : {
  "_aux": {
    "chains": [{
       "id": "xtr",
[...]

PR #2654, opened 2021-02-17
[merge-mining] Implement get_block_header_by_hash and aux chain extensions

sdbondi:
## Description

- Add get_header_by_hash grpc call and associated base node backend calls
- Because of the way the base node service requests are tied to the base node 
  comms message types, GetHeaderByHash added to comms requests/responses
- Call get_header_by_hash from merge mining proxy
- Replace `_merge_mining` key with aux chain data in get_block_template response (see #2620)

## Motivation and Context

Support for merge mined pools


stringhandler (approved):
Looks good

PR #2478, opened 2020-12-05
[wallet] reuse generated coinbase tx when mining

delta1:
## Description

Changes the `generate_coinbase_transaction` function to check for a previously generated coinbase tx and to reuse that if it exists, instead of always generating a new one (and leaving many cancelled coinbase txs when mining).

## Motivation and Context

Before this change, a new coinbase transaction was generated each time it was requested, even if it was for the same height and amount. Now if the height or amount have not changed, the same transaction will be reused instead of being cancelled. 

## How Has This Been Tested?

`cargo test coinbase` and manually during merge mining with xmrig, mm proxy and console wallet 

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #2457, opened 2020-11-26
[wallet] Fix bug where cancelled output is spent

delta1:
## Description

During merge mining, sometimes a number of cancelled outputs are added to the wallet before the mined coinbase utxo.
When spending this valid utxo, the _cancelled_ output was marked as encumbered and then spent, instead of the valid utxo.
This led to the bug where after "spending" a utxo, it would still exist in the db as unspent. It would then be selected for transactions, but reported as spent and be unspendable.
Fundamentally the issue was that `OutputSql::find_by_commitment` was not excluding cancelled outputs.

If testing this on a wallet that already exhibits the issue, the sqlite db will have to be cleared of matching outputs that exist in both spent and unspent status. 

## Motivation and Context

Fixes #2443

## How Has This Been Tested?

cargo test, and manually checked in console wallet

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #2655, opened 2021-02-17
[tests] Remove cancelled output excluded from spending test

delta1:
## Description

Removes the `test_cancelled_outputs_excluded_sqlite_db` test that is no longer applicable.

## Motivation and Context

This test was broken due to changes from #2644, where we now enforce unique commitments in the outputs table.

Previously, it tested that cancelled tx outputs with the same commitment would not be spent instead of the valid output (issue #2457).

This is now solved in 2 ways:
- coinbase outputs are reused, instead of being regenerated with the same commitment (#2478)
- unique commitments are enforced in the outputs table (#2644)

Given that our invariants have changed, this test is no longer required.

## How Has This Been Tested?

`cargo test` 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


philipr-za (approved):
Agreed, not needed anymore

PR #2665, opened 2021-02-18
[base-node] Don't send duplicate kernels and utxos in sync mode

stringhandler:
## Description

Previously when requesting kernels and utxos (e.g. 2-1000), if the end MMR node requested was in the middle of a block, the rest of the block would be transmitted resulting in duplicate MMR nodes being sent. This caused the error: 
MMR did not match for kernels, b776677e03d29f9debfcec364286e9c94a5b8b98593826168cfc382c2c68ddf0 != 2a834b68f70a355034c11bba8f1ace763dbbdf0535540a33cd8f96848bb33f29

Also added the weight and number of blocks in the transaction to the mempool stats displayed in the `status()` command of the base node

### DIFF
--- a/applications/tari_base_node/src/command_handler.rs
+++ b/applications/tari_base_node/src/command_handler.rs
@@ -163,12 +163,15 @@ impl CommandHandler {
             let banned_peers = banned_peers(&peer_manager).await.unwrap();
             let conns = connectivity.get_active_connections().await.unwrap();
             println!(
-                "{}: State: {}, Tip: {} ({}), Mempool: {} tx, Mining (H/R): {} ({} MH/s), Connections: {}, Banned: {}",
+                "{}: State: {}, Tip: {} ({}), Mempool: {}tx ({}g, +/- {}blks), Mining: {} ({} MH/s), Connections: {}, \
+                 Banned: {}",
                 Utc::now().format("%H:%M"),
                 state.state_info.short_desc(),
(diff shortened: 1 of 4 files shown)


delta1:
Awesome looks good, testing

delta1:
This is working a lot better. I still get this kernel error: 

```
WARN  Synchronizing horizon state has failed. MMR did not match for Kernel at height 427. 2a834b68f70a355034c11bba8f1ace763dbbdf0535540a33cd8f96848bb33f29 did not equal b776677e03d29f9debfcec364286e9c94a5b8b98593826168cfc382c2c68ddf0
``` 

but the sync progresses on to the utxo sync, and then during the final validation, gets stuck in this loop of `sync utxos > validate final state > UTXO set did not balance with the expected emission at height 3839 > repeat` 

```
WARN  Synchronizing horizon state has failed. Final state validation failed: Final state validation failed: The UTXO set did not balance with the expected emission at height 3839
```

stringhandler:
@delta1 this is a fix for the RPC server that is sending the nodes, so you need to make sure you are running the code there.

Unfortunately block 1806 doesn't pass the balance validation, I'm not sure why, but I don't think that should prevent this from being merged

delta1:
Yep I was running the same code on the RPC server node.

delta1 (approved):
LGTM

PR #2670, opened 2021-02-18
[wallet] Remove tx/s from make-it-rain

stringhandler:
Remove transactions per second from this call in favour of a single amount of transactions to be sent.

The existing code could not guarantee that the tx/s would be reached and was buggy. When calling it with 10 tx/s for 10 seconds, it would always use the 2 minute max timing.  It's a much simpler interface to use a full set of transactions and call them in one go

### DIFF
--- a/applications/tari_console_wallet/src/automation/commands.rs
+++ b/applications/tari_console_wallet/src/automation/commands.rs
@@ -106,37 +106,32 @@ pub async fn make_it_rain(
 {
     use ParsedArgument::*;
 
-    let txps = match args[0].clone() {
-        Float(r) => Ok(r),
+    let txs = match args[0].clone() {
+        Int(r) => Ok(r),
         _ => Err(CommandError::Argument),
     }?;
 
-    let duration = match args[1].clone() {
-        Int(s) => Ok(s),
-        _ => Err(CommandError::Argument),
-    }?;
-
-    let start_amount = match args[2].clone() {
+    let start_amount = match args[1].clone() {
         Amount(mtari) => Ok(mtari),
         _ => Err(CommandError::Argument),
     }?;
 
-    let inc_amount = match args[3].clone() {
+    let inc_amount = match args[2].clone() {
         Amount(mtari) => Ok(mtari),
         _ => Err(CommandError::Argument),
     }?;
 
-    let start_time = match args[4].clone() {
+    let start_time = match args[3].clone() {
         Date(dt) => Ok(dt as DateTime<Utc>),
         _ => Err(CommandError::Argument),
     }?;
 
-    let public_key = match args[5].clone() {
+    let public_key = match args[4].clone() {
         PublicKey(pk) => Ok(pk),
         _ => Err(CommandError::Argument),
     }?;
 
-    let message = match args[6].clone() {
+    let message = match args[5].clone() {
         Text(m) => Ok(m),
         _ => Err(CommandError::Argument),
     }?;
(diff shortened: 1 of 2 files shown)


delta1:
Looks good, I think the instructional text just needs to be updated to reflect the change 

```
const MAKE_IT_RAIN_USAGE: &str = "\nmake-it-rain [Txs/s] [duration (s)] [start amount (uT)] [increment (uT)/Tx] \
                                  [\"start time (UTC)\" / 'now' for immediate start] [public key or emoji id to send \
                                  to] [message]\n";
```

hansieodendaal:
Hi there, I believe the symptoms you are describing is indicative of another problem. The transaction submission rate was implemented to:
- enable planning a stress test to submit transactions spread out equally over time, e.g. try to submit 36,000 transactions from 6 sending wallets in one hour.
- not over-stress the system as one could easily pump 10,000 transactions into the system in a really short time. Tests showed that 25 txn/s was a safe sending rate as going above that sometimes resulted in crashing the Tokio environment on the sending side (_issue never solved as this was considered test code only_).

Regarding the 120s maximum delay (old lines 171 - 174); that would only come into play if the transactions were really submitted very fast and you had to slow down the system. The only thing this code is doing is to not let transactions go faster than what the intention is. Thus:

| Cumulative target time | Cumulative actual time | Delay for |
| ---------------------- | ---------------------- | --------- |
| 20 s                   | 10 s                   | 10 s      |
| 10 s                   | 20 s                   | 0 s       |
| 200 s                  | 10 s                   | 120 s     |
| 10 s                   | 200 s                  | 0 s       |

One may up the 25 txn/s default rate limit to see if we still get the Tokio crush, perhaps with a config setting, and as you suggested, perhaps measuring and reporting on the actual achieved submission rate.

delta1:
I think tx/s should be an optional argument only, and have a sensible default

hansieodendaal:
> I think tx/s should be an optional argument only, and have a sensible default

An optional command-line argument could work, but I do not think we should release code that crashes out of the box, hence a config setting to limit the maximum rate.

SWvheerden:
At the moment the current code only sends a tx every 2 minutes. This PR at least makes it useable

delta1:
> At the moment the current code only sends a tx every 2 minutes. This PR at least makes it useable

That's very strange, and definitely a bug. What command are you using? Just for info.

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -159,19 +154,9 @@ pub async fn make_it_rain(
...
-    let started_at = Utc::now();
-
-    for i in 0..num_txs {
-        // Manage Tx rate
-        let actual_ms = (Utc::now() - started_at).num_milliseconds() as u64;
-        let target_ms = (i as f64 / (txps / 1000.0)) as u64;
-        if target_ms - actual_ms > 0 {
-            // Maximum delay between Txs set to 120 s
-            delay_for(Duration::from_millis((target_ms - actual_ms).min(120_000u64))).await;
-        }
```
hansieodendaal, line 174: @delta1 and @SWvheerden, I think this is the problem right here. In the original implementation for the base node wallet, this calculation was specifically done with `i64` and not `u64` to prevent wrap around.

hansieodendaal: The original implementation worked perfectly.

delta1: Definitely the cause of the delay, but subtracting u64s don't wrap around, it panics with: 

`thread 'main' panicked at 'attempt to subtract with overflow', applications/tari_console_wallet/src/automation/commands.rs:263:41` 

After changing the minus to a saturating sub, I ran a test successfully at 10 tx/sec for 5 seconds. But SW was running for 50 seconds and said that it immediately starting delaying by 2 minutes. 

Regardless, the logic needs a re-work if it's just about managing the maximum rate.

PR #2674, opened 2021-02-19
Add delay_lagging option to config 

SWvheerden:
## Description

Add delay_lagging option to delay the time a node will go from listing to lagging.
Added cucumber tests for this. 

## Motivation and Context

Adding this option to test block propagation. Default set to 6 blocks. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/bootstrap/base_node.rs
+++ b/applications/tari_base_node/src/bootstrap/base_node.rs
@@ -157,6 +157,7 @@ where B: BlockchainBackend + 'static
                     pruning_horizon: config.pruning_horizon,
                     orphan_db_clean_out_threshold: config.orphan_db_clean_out_threshold,
                     max_randomx_vms: config.max_randomx_vms,
+                    blocks_behind_before_considered_lagging: self.config.blocks_behind_before_considered_lagging,
                     ..Default::default()
                 },
                 self.rules,
--- a/common/src/configuration/global.rs
+++ b/common/src/configuration/global.rs
@@ -116,6 +116,7 @@ pub struct GlobalConfig {
     pub max_randomx_vms: usize,
     pub console_wallet_notify_file: Option<PathBuf>,
     pub auto_ping_interval: u64,
+    pub blocks_behind_before_considered_lagging: u64,
 }
 
 impl GlobalConfig {
@@ -403,6 +404,10 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         Err(e) => return Err(ConfigurationError::new(&key, &e.to_string())),
     };
 
+    // blocks_behind_before_considered_lagging when a node should switch over from listening to lagging
+    let key = config_string("base_node", &net_str, "blocks_behind_before_considered_lagging");
+    let blocks_behind_before_considered_lagging = optional(cfg.get_int(&key))?.unwrap_or(6) as u64;
+
     // set wallet_db_file
     let key = "wallet.wallet_db_file".to_string();
     let wallet_db_file = cfg
@@ -652,6 +657,7 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         max_randomx_vms,
         console_wallet_notify_file,
         auto_ping_interval,
+        blocks_behind_before_considered_lagging,
     })
 }
 
--- a/integration_tests/features/Propagation.feature
+++ b/integration_tests/features/Propagation.feature
@@ -50,3 +50,34 @@ Feature: Block Propagation
     Then I receive an error containing 'Orphan block'
     And all nodes are at height 1
 
+  Scenario: Nodes should never switch to block sync but keep insync via propagation
+    Given I have a base node MINER connected to all seed nodes
+    And I have a lagging delayed node LAG1 connected to node MINER with blocks_behind_before_considered_lagging 10000
+    Given I have a lagging delayed node LAG2 connected to node MINER with blocks_behind_before_considered_lagging 10000
+    #wait for node to so start and get into listing mode.
+    When I wait 100 seconds
+    When I mine 5 blocks on MINER
+    Then node MINER is at height 5
+    Then all nodes are at height 5
+    When I mine 15 blocks on MINER
+    Then all nodes are at height 20
+
+    Scenario: Node should lag when for while before syncing
+    Given I have a base node MINER connected to all seed nodes
+    And I have a lagging delayed node LAG1 connected to node MINER with blocks_behind_before_considered_lagging 6
+    When I mine 1 blocks on MINER
+    When I wait 100 seconds
+    When I stop LAG1
+    When I wait 10 seconds
+    When I mine 5 blocks on MINER
+    When I wait 100 seconds
+    When I start LAG1
+    #wait for node to so start and get into listing mode.
+    When I wait 100 seconds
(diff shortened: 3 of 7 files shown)


common/src/configuration/global.rs
```diff
@@ -403,6 +404,10 @@ fn convert_node_config(network: Network, cfg: Config) -> Result<GlobalConfig, Co
         Err(e) => return Err(ConfigurationError::new(&key, &e.to_string())),
     };
 
+    // delay_lagging when a node should switch over from listing to lagging
```
stringhandler:
```suggestion
    // delay_lagging when a node should switch over from listening to lagging
```

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
     Then I receive an error containing 'Orphan block'
     And all nodes are at height 1
 
+  Scenario: Nodes should never sync but keep insync
```
stringhandler: Huh?

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
...
+    And I have a lagging delayed node LAG1 connected to node MINER with delay_lagging 10000
+    Given I have a lagging delayed node LAG2 connected to node MINER with delay_lagging 10000
+    When I wait 100 seconds
+    When I mine 5 blocks on MINER
+    Then node MINER is at height 5
+    Then all nodes are at height 5
+    When I mine 15 blocks on MINER
+    Then all nodes are at height 20
+
+    Scenario: Node should lag
```
stringhandler:
```suggestion
    Scenario: Node should lag when.....
```

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
     Then I receive an error containing 'Orphan block'
     And all nodes are at height 1
 
+  Scenario: Nodes should never sync but keep insync
+    Given I have 2 seed nodes
+    And I have 4 base nodes connected to all seed nodes
```
stringhandler: The seed nodes and these 4 base nodes don't seem to take part in this test. Can they be removed/reduced?

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
...
+    And I have a base node MINER connected to all seed nodes
+    And I have a lagging delayed node LAG1 connected to node MINER with delay_lagging 10000
+    Given I have a lagging delayed node LAG2 connected to node MINER with delay_lagging 10000
+    When I wait 100 seconds
```
stringhandler, line 58: Need a comment for the reason of this delay

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
...
+    When I mine 5 blocks on MINER
+    Then node MINER is at height 5
+    Then all nodes are at height 5
+    When I mine 15 blocks on MINER
+    Then all nodes are at height 20
+
+    Scenario: Node should lag
+    Given I have 2 seed nodes
+    And I have 4 base nodes connected to all seed nodes
```
stringhandler: Same comment for these nodes

integration_tests/features/Propagation.feature
```diff
@@ -50,3 +50,35 @@ Feature: Block Propagation
...
+    And I have a base node MINER connected to all seed nodes
+    And I have a lagging delayed node LAG1 connected to node MINER with delay_lagging 6
+    When I mine 1 blocks on MINER
+    When I wait 100 seconds
```
stringhandler, line 69: Need a reason for every `wait`, e.g. when can this wait be removed

stringhandler (commented):
Happy with the approach. I think the config setting could have a better name, perhaps `blocks_behind_before_considered_lagging`

integration_tests/features/Propagation.feature
delta1: "Nodes should not switch to block sync state, but should instead stay synced through block propagation"

integration_tests/features/Propagation.feature
SWvheerden: They can, it was just to provide a small network.

integration_tests/features/Propagation.feature
SWvheerden: Left comment. If its not there the node misses all the messages

integration_tests/features/support/steps.js
```diff
@@ -54,6 +56,14 @@ Given(/I have a pruned node (.*) connected to node (.*)/, {timeout: 20*1000}, as
     await sleep(1000);
 });
 
+Given(/I have a lagging delayed node (.*) connected to node (.*) with blocks_behind_before_considered_lagging (\d+)/, {timeout: 20*1000}, async function (name, node, delay) {
+    const miner =  new BaseNodeProcess(name, { blocks_behind_before_considered_lagging: delay});
```
hansieodendaal, line 61: Looking good, just one comment.

Maybe my bad, but I do not understand how this can have any effect if you have not added the option to [`function mapEnvs(options) {` in `config.js`](https://github.com/SWvheerden/tari/blob/sw_Add_delay_lagging/integration_tests/helpers/config.js#L7) (or see [main repo](https://github.com/tari-project/tari/blob/development/integration_tests/helpers/config.js#L7)) ?

applications/tari_base_node/src/bootstrap/base_node.rs
```diff
@@ -157,6 +157,7 @@ where B: BlockchainBackend + 'static
                     pruning_horizon: config.pruning_horizon,
                     orphan_db_clean_out_threshold: config.orphan_db_clean_out_threshold,
                     max_randomx_vms: config.max_randomx_vms,
+                    blocks_behind_before_considered_lagging: self.config.blocks_behind_before_considered_lagging,
```
hansieodendaal, line 160: **NIT:**

 Suggest add debug log message after the `StackBuilder` instruction:
```
debug!(target: LOG_TARGET, "Blocks behind before considered lagging set to {}", self.config.blocks_behind_before_considered_lagging);
```

SWvheerden: It logs info, every-time it receives a metadata and its behind.

hansieodendaal (approved):
Tested after change to set optional `TARI_BASE_NODE__LOCALNET__BLOCKS_BEHIND_BEFORE_CONSIDERED_LAGGING` environment variable in `config.js`. Rust implementation works fine, cucumber works fine.

**NIT**
The one feature could be simplified though, by removing some of the waits. This worked on my system:
```
    Scenario: Node should lag when for while before syncing
    Given I have a base node MINER connected to all seed nodes
    And I have a lagging delayed node LAG1 connected to node MINER with blocks_behind_before_considered_lagging 7
    When I mine 1 blocks on MINER
    When I stop LAG1
    When I mine 5 blocks on MINER
    When I start LAG1
        #wait for node to so start and get into listing mode.
    When I wait 100 seconds
    Then node MINER is at height 6
        #node was shutdown, so it never received the propagation messages
    Then node LAG1 is at height 1
    When I mine 2 blocks on MINER
    Then all nodes are at height 8

```

PR #2677, opened 2021-02-19
[base-node] Fix UTXO sum for horizon

stringhandler:
## Description

Fix the spent utxo sum so that it balances.

## Motivation and Context

Pruned nodes trying to sync with a horizon higher than block 1806 were failing due to the incorrect emission. 

## How Has This Been Tested?

Tested locally

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
@@ -30,7 +30,14 @@ use crate::{
         sync::rpc,
     },
     blocks::BlockHeader,
-    chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend, ChainStorageError, HorizonData, MmrTree},
+    chain_storage::{
+        async_db::AsyncBlockchainDb,
+        BlockchainBackend,
+        ChainStorageError,
+        HorizonData,
+        MmrTree,
+        PrunedOutput,
+    },
     proto::base_node::{SyncKernelsRequest, SyncUtxosRequest, SyncUtxosResponse},
     transactions::{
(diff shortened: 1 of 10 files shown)


delta1:
I'm still getting 

```
WARN  Error during sync:Final state validation failed: Final state validation failed: The UTXO set did not balance with the expected emission at height 6626
```

stringhandler:
Interesting. I will try it again

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -411,14 +422,54 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
...
+                curr_header.height(), curr_header.header.kernel_mmr_size, prev_kernel_mmr, curr_header.header.kernel_mmr_size -1);
+            let kernels = self.db().fetch_kernels_by_mmr_position(prev_kernel_mmr, curr_header.header.kernel_mmr_size -1 ).await?;
+
+            debug!(target: LOG_TARGET,
+                   "Number of kernels returned: {}", kernels.len());
+            for k in kernels {
+                pruned_kernel_sum = &k.excess + &pruned_kernel_sum;
+
+            }
+            prev_kernel_mmr = curr_header.header.kernel_mmr_size;
```
SWvheerden: We dont have to do this for the kernels, we already save the kernel sum per block in the `block_accumulated_data` which is already saved in this loop as `accum_data `

We can replace this for loop with:
`pruned_kernel_sum = accum_data.kernel_sum() + &pruned_kernel_sum;`

SWvheerden (dismissed):
Looks good, we can improve the code by not replacing the kernel_sum in the accumulated data. But need to count the utxos'. Pr looks good.

SWvheerden (commented):
LGTM

PR #2679, opened 2021-02-22
[wallet] Change select_utxos to not fail without chain metadata

delta1:
## Description

Changes `select_utxos` function to not fail when no chain metadata is available. In that case, unmatured utxos might be selected and the transaction would fail, but I suggest that's a better option than failing outright to even give a fee estimate etc. If no metadata is available, and no strategy was specified for the utxo selection, then `MaturityThenSmallest` strategy will be used. 

todo:
- new tests 

## Motivation and Context

Numerous issues with utxo selection 

## How Has This Been Tested?

- Existing tests
- tx sent between console wallets

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/base_node_service/service.rs
+++ b/base_layer/wallet/src/base_node_service/service.rs
@@ -320,7 +320,10 @@ where
                     let metadata: ChainMetadata = chain_metadata
                         .try_into()
                         .map_err(BaseNodeServiceError::InvalidBaseNodeResponse)?;
-                    self.db.set_chain_meta(metadata.clone()).await?;
+
+                    // store chain metadata in the wallet db
+                    self.db.set_chain_metadata(metadata.clone()).await?;
+
                     let state = BaseNodeState {
                         is_synced: Some(message.is_synced),
                         chain_metadata: Some(metadata),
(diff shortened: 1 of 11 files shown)


base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -389,8 +392,12 @@ where
                 Ok(BaseNodeServiceResponse::BaseNodePeerSet)
             },
             BaseNodeServiceRequest::GetChainMetadata => match self.state.chain_metadata.clone() {
-                Some(v) => Ok(BaseNodeServiceResponse::ChainMetadata(Some(v))),
-                None => Ok(BaseNodeServiceResponse::ChainMetadata(self.db.get_chain_meta().await?)),
+                Some(metadata) => Ok(BaseNodeServiceResponse::ChainMetadata(Some(metadata))),
+                None => {
+                    // if we don't have live state, check if we've previously stored state in the wallet db
+                    let metadata = self.db.get_chain_metadata().await?;
```
philipr-za, line 398: Should maybe cater for the case that neither of these cases are true e.g. in a completely fresh wallet before the first chain metadata response is received.

delta1: yeah so `get_chain_metadata` now returns an Option so that would be `None`

stringhandler (approved):

PR #2681, opened 2021-02-23
Add coinbase handling ability via GRPC with cucumber tests for reorg scenario

hansieodendaal:
## Description
Added GRPC interfaces for `get balance` and `get all completed transactions`. Added new cucumber integration tests to verify handling of coinbase transactions after a reorg. Fixed bug in coinbase monitoring protocol regarding the update of statues in the sqlite db.

~~**Note:** Changes to default logging files will be reverted before finalizing this PR; those are needed to do proper analysis for cucumber tests.~~

## Motivation and Context
Problem: Mined coinbases that are reorged out continues to show "mined unconfirmed" instead of being cancelled in the console wallet.

## How Has This Been Tested?
Cucumber integration test `Scenario: Wallets monitoring coinbase after a reorg`.

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -4,6 +4,9 @@ use tari_app_grpc::{
     conversions::naive_datetime_to_timestamp,
     tari_rpc::{
         wallet_server,
+        GetAllCompletedTransactionsRequest,
+        GetBalanceRequest,
+        GetBalanceResponse,
         GetCoinbaseRequest,
         GetCoinbaseResponse,
         GetIdentityRequest,
@@ -23,6 +26,7 @@ use tari_app_grpc::{
 use tari_comms::types::CommsPublicKey;
 use tari_core::tari_utilities::{hex::Hex, ByteArray};
 use tari_wallet::{
+    output_manager_service::handle::OutputManagerHandle,
     transaction_service::{handle::TransactionServiceHandle, storage::models},
     WalletSqlite,
 };
@@ -42,6 +46,10 @@ impl WalletGrpcServer {
     fn get_transaction_service(&self) -> TransactionServiceHandle {
         self.wallet.transaction_service.clone()
     }
+
+    fn get_output_manager_service(&self) -> OutputManagerHandle {
+        self.wallet.output_manager_service.clone()
+    }
 }
 
 #[tonic::async_trait]
@@ -60,6 +68,20 @@ impl wallet_server::Wallet for WalletGrpcServer {
             public_key: identity.public_key().to_string().as_bytes().to_vec(),
             public_address: identity.public_address().to_string(),
             node_id: identity.node_id().to_string().as_bytes().to_vec(),
+        }))
+    }
+
+    async fn get_balance(&self, _request: Request<GetBalanceRequest>) -> Result<Response<GetBalanceResponse>, Status> {
+        let mut output_service = self.get_output_manager_service();
+        let balance;
+        match output_service.get_balance().await {
+            Ok(b) => balance = b,
+            Err(e) => return Err(Status::not_found(format!("GetBalance error! {}", e))),
+        }
+        Ok(Response::new(GetBalanceResponse {
+            available_balance: balance.available_balance.0,
+            pending_incoming_balance: balance.pending_incoming_balance.0,
+            pending_outgoing_balance: balance.pending_outgoing_balance.0,
         }))
     }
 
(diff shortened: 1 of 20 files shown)


applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -175,6 +197,53 @@ impl wallet_server::Wallet for WalletGrpcServer {
 
         Ok(Response::new(GetTransactionInfoResponse { transactions }))
     }
+
+    async fn get_all_completed_transactions(
+        &self,
+        _request: Request<GetAllCompletedTransactionsRequest>,
```
stringhandler, line 203: This should either return a stream or use some paging of some kind. It's not hard to add

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -196,6 +265,7 @@ fn convert_wallet_transaction_into_transaction_info(
             excess_sig: Default::default(),
             timestamp: Some(naive_datetime_to_timestamp(tx.timestamp)),
             message: tx.message,
+            valid: true,
```
stringhandler, line 268: `valid` as a term is a bit vague. Under what conditions is `valid == false`?

integration_tests/features/CoinbaseReorg.feature
```diff
@@ -0,0 +1,78 @@
+@coinbase_reorg
+Feature: Wallet Routing Mechanism
```
stringhandler: The Feature name should match the file name. I would suggest WalletMonitoring as a file and feature name

integration_tests/features/CoinbaseReorg.feature
```diff
@@ -0,0 +1,78 @@
...
+    Then wallet WALLET_B1 detects all transactions are at least Broadcast
+        #
+        # Connect Chain 1 and 2
+        #
+        # TODO: This wait is needed to stop next base node task from continuing
+    When I wait 1 seconds
+    And I have a base node NODE_C connected to all seed nodes
+        # Wait for the reorg to filter through
+    When I wait 30 seconds
+    Then all nodes are at height 10
```
stringhandler: I think rather mine a significantly stronger chain on B, so that it always reorgs from A->B. It should make it easier to test specific values and statuses on A and B in the following steps. We can't guarantee 15 blocks will be stronger than 10 at the moment, but we might be able to in the future. Until then, if it occasionally fails because it reorgs to A instead, that isn't a train smash

stringhandler (commented):
Looks like it's on the right track, some suggestions

integration_tests/features/CoinbaseReorg.feature
hansieodendaal: I do not know if I agree with this approach, though. The current test metrics do not care which chain wins, just that the invalid coinbase transactions will be discarded when the tip moves ahead. The outcome is always predictable, and when the test fails something is definitely wrong.

integration_tests/features/CoinbaseReorg.feature
hansieodendaal: Ooops... copy paste error. Will update.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
hansieodendaal: This exactly mimics the behaviour of 
   `impl From<OutboundTransaction> for CompletedTransaction {`
 and 
   `impl From<InboundTransaction> for CompletedTransaction {`.


 The meaning of `valid` in `InboundTransaction` and `OutboundTransaction` is actually not defined and therefore always `true`; it only gets to mean something if the transaction progressed from `PendingInbound` or `PendingOutbound` to `Completed`.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
hansieodendaal: Proposed as future change (i.e. next PR)

stringhandler (approved):
Looks good. Will test locally and merge

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -175,6 +197,53 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+                message: tx.1.message,
+                valid: tx.1.valid,
+            })
+            .collect();
+
+        Ok(Response::new(GetTransactionInfoResponse { transactions }))
+    }
 }
 
 fn convert_wallet_transaction_into_transaction_info(
```
philipr-za, line 249: This is beyond the scope of this PR but this function duplicates the functionality of the following two implementations:
https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/transaction_service/storage/models.rs#L312
https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/transaction_service/storage/models.rs#L334

Seems unnecessary to maintain this in two places.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1664,6 +1664,16 @@ where
                     target: LOG_TARGET,
                     "Coinbase Transaction monitoring Protocol for TxId: {} completed successfully", id
                 );
+                if let Ok(tx) = self.resources.db.get_completed_transaction(id).await {
```
philipr-za, line 1667: So I understand that this helped diagnose a problem but you are adding in a whole DB access here for a trace log. This is something I feel is a wasted use of resources. Add them in if you have to debug an issue but leaving this here is a waste for the rest of the life of this code.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
hansieodendaal: Agreed - maybe for a future PR

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Hi @philipr-za will remove in a next PR

PR #2684, opened 2021-02-23
DNM [merge-mining] Submit_block AUX data includes tari block hash

sdbondi:
Calls to submit_block return the tari hash if the block is accepted by
the Tari blockchain.

- [fix] Return the correct response when tari's get_header_by_hash GRPC call
  returns "not found"
- [feat] get_block_template includes tari block height in aux chain data
- [fix] get_block_header_by_hash should not forward the request to tari
  base node if block was found by monerod.
- [fix] Add correct `Host` http header on proxied requests. Some public monero
  daemons require this.
- [chore] Minor request logging improvements


stringhandler (dismissed):
Looks good, I suspect there may be a cargo fmt error though. Will wait for CI

StriderDM:
Please rebase against https://github.com/tari-project/tari/pull/2699 after it merges.

PR #2686, opened 2021-02-23
Implement "stress test" cucumber test

philipr-za:
## Description
This PR adds the components required to run a stress test Cucumber test and implement an example stress test cucumber test.

The PR adds a `coin_split` command to the wallet gRPC interface and so that a cucumber test can coin split and then adds a number of cucumber utility functions to allow for sending mass transactions and monitoring for those transaction to be complete while also mining.

The purpose of this test is to provide the framework in which one could conduct a variety of different stress tests in a variety of configurations. The example stress test passes on a MacBook after some time but comments are left in to show an example of the number of transactions that would cause the test to start failing. 

## How Has This Been Tested?
This is a test

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/integration_tests/features/StressTest.feature
+++ b/integration_tests/features/StressTest.feature
@@ -1,0 +1,39 @@
+Feature: Stress Test
+
+@long-running
+Scenario: Stress Test
+    Given I have a seed node NODE1
+    And I have a seed node NODE2
+    And I have 1 base nodes connected to all seed nodes
+    And I have wallet WALLET_A connected to seed node NODE1
+    And I have wallet WALLET_B connected to seed node NODE2
+    And I have a merge mining proxy PROXY connected to NODE1 and WALLET_A
+    # We need to ensure the coinbase lock heights are reached; mine enough blocks
+    # The following line is how you could mine directly on the node
+    When I merge mine 8 blocks via PROXY
+
+    Then all nodes are at height 8
+    When I wait for wallet WALLET_A to have at least 15100000000 tari
+    # A coin split can produce at most 499 outputs
+    And I coin split tari in wallet WALLET_A to produce 499 UTXOs of 7000 uT each with fee_per_gram 20 uT
+    And I coin split tari in wallet WALLET_A to produce 499 UTXOs of 7000 uT each with fee_per_gram 20 uT
+
+    # Make sure enough blocks are mined for the coin split transaction to be confirmed
+    When I merge mine 6 blocks via PROXY
+
+    Then all nodes are at height 14
+    Then wallet WALLET_A detects all transactions as Mined_Confirmed
+    When I wait 1 seconds
+    When I send 400 transactions of 1111 uT each from wallet WALLET_A to wallet WALLET_B at fee_per_gram 20
+    # The following line really taxes SEED NODE A
+    # When I send 1400 transactions of 1111 uT each from wallet WALLET_A to wallet WALLET_B at fee_per_gram 20
+
+    # Mine enough blocks for the first block of transactions to be confirmed.
+    When I merge mine 4 blocks via PROXY
+    Then all nodes are at height 18
+    # Now wait until all transactions are detected as confirmed in WALLET_A, continue to mine blocks if transactions
+    # are not found to be confirmed as sometimes the previous mining occurs faster than transactions are submitted
+    # to the mempool
+    Then while merge mining via NODE1 all transactions in wallet WALLET_A are found to be Mined_Confirmed
+    # Then wallet WALLET_B detects all transactions as Mined_Confirmed
+    Then while mining via NODE1 all transactions in wallet WALLET_B are found to be Mined_Confirmed
--- a/integration_tests/features/support/steps.js
+++ b/integration_tests/features/support/steps.js
@@ -224,6 +224,14 @@ When(/I mine (\d+) blocks on (.*)/, {timeout: 600*1000}, async function (numBloc
     }
 });
 
+When(/I mine (\d+) blocks using wallet (.*) on (.*)/, {timeout: 600*1000}, async function (numBlocks, walletName,nodeName) {
+    let nodeClient = this.getClient(nodeName);
+    let walletClient =this.getWallet(walletName).getClient()
+    for(let i=0;i<numBlocks;i++) {
+        await nodeClient.mineBlock(walletClient);
+    }
+});
+
 When(/I merge mine (.*) blocks via (.*)/, {timeout: 600*1000}, async function (numBlocks, mmProxy) {
     for(let i=0;i<numBlocks;i++) {
         await this.mergeMineBlock(mmProxy);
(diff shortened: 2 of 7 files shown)


hansieodendaal:
Quick feedback. Running this timed out close to the end. Maybe increase function timeout here to `1200*1000`:
```
Then('all nodes are at height {int}', {timeout: 120*1000},async function (height) {
    await this.forEachClientAsync(async (client, name) => {
        await waitFor(async() => client.getTipHeight(), height, 115*1000);
        const currTip = await client.getTipHeight();
        console.log(`Node ${name} is at tip: ${currTip} (should be ${height})`);
        expect(currTip).to.equal(height);
    })
});
```

Console:
```
..Node NODE1 is at tip: 18 (should be 18)
waiting for process... 500 50
waiting for process... 500 50
waiting for process... 500 100
waiting for process... 500 100
waiting for process... 500 150
waiting for process... 500 150
waiting for process... 500 200
waiting for process... 500 200
Node BaseNode0 is at tip: 14 (should be 18)
FNode NODE2 is at tip: 14 (should be 18)
--Stopping nodes
```

integration_tests/features/support/steps.js
```diff
@@ -675,6 +675,32 @@ Then(/wallet (.*) detects all transactions as Mined_Confirmed/, {timeout: 1200*1
     }
 });
 
+Then(/while mining via (.*) all transactions in wallet (.*) are found to be Mined_Confirmed/, {timeout: 1200*1000}, async function (mmProxy, walletName) {
```
hansieodendaal:
```suggestion
Then(/while mining via (.*) all transactions in wallet (.*) are found to be Mined_Confirmed/, {timeout: 3800*1000}, async function (mmProxy, walletName) {
```

hansieodendaal: My test timed out with the lesser value. This issue is more pronounced with more transactions. Upping the value as suggested worked fine for this 400 odd test.

integration_tests/features/StressTest.feature
```diff
@@ -0,0 +1,35 @@
...
+    When I wait for wallet WALLET_A to have at least 15100000000 tari
+    # A coin split can produce at most 499 outputs
+    And I coin split tari in wallet WALLET_A to produce 499 UTXOs of 7000 uT each with fee_per_gram 20 uT
+    And I coin split tari in wallet WALLET_A to produce 499 UTXOs of 7000 uT each with fee_per_gram 20 uT
+    # Make sure enough blocks are mined for the coin split transaction to be confirmed
+    When I merge mine 6 blocks via PROXY
+    Then all nodes are at height 14
+    Then wallet WALLET_A detects all transactions as Mined_Confirmed
+    When I wait 1 seconds
+    When I send 400 transactions of 1111 uT each from wallet WALLET_A to wallet WALLET_B at fee_per_gram 20
```
hansieodendaal, line 27: **NIT:** I would add a transaction broadcast confirmation step here before continuing with the test as the next merge mining steps can be really quick:
```suggestion
    When I send 400 transactions of 1111 uT each from wallet WALLET_A to wallet WALLET_B at fee_per_gram 20
    Then all wallets detect all transactions are at least Broadcast
```

philipr-za: This issue is handled by the fact that while I am waiting for transaction to be confirmed I am also mining. I felt that is a much more realistic situation when you consider how the real network would pump transactions in the system.

integration_tests/features/support/steps.js
```diff
@@ -766,4 +792,51 @@ Then('Difficulties are available', function () {
...
+            "message": "Cucumber coinsplit",
+            "lockheight": 0,
+        }
+    );
+    console.log("Coin split to produce", splitNum, "values of", splitValue, "uT completed with TxId: ", result);
+    this.addTransaction(walletInfo["public_key"], result["tx_id"]);
+    this.lastResult = result;
+});
+
+When(/I send (.*) transactions of (.*) uT each from wallet (.*) to wallet (.*) at fee_per_gram (.*)/, {timeout: 25*5*1000}, async function (numTransactions, amount, sourceWallet, dest, fee) {
```
hansieodendaal, line 859:
```suggestion
When(/I send (.*) transactions of (.*) uT each from wallet (.*) to wallet (.*) at fee_per_gram (.*)/, {timeout: 1200*1000}, async function (numTransactions, amount, sourceWallet, dest, fee) {
```

philipr-za:
> Quick feedback. Running this timed out close to the end. Maybe increase function timeout here to `1200*1000`:
> 
> ```
> Then('all nodes are at height {int}', {timeout: 120*1000},async function (height) {
>     await this.forEachClientAsync(async (client, name) => {
>         await waitFor(async() => client.getTipHeight(), height, 115*1000);
>         const currTip = await client.getTipHeight();
>         console.log(`Node ${name} is at tip: ${currTip} (should be ${height})`);
>         expect(currTip).to.equal(height);
>     })
> });
> ```
> 
> Console:
> 
> ```
> ..Node NODE1 is at tip: 18 (should be 18)
> waiting for process... 500 50
> waiting for process... 500 50
> waiting for process... 500 100
> waiting for process... 500 100
> waiting for process... 500 150
> waiting for process... 500 150
> waiting for process... 500 200
> waiting for process... 500 200
> Node BaseNode0 is at tip: 14 (should be 18)
> FNode NODE2 is at tip: 14 (should be 18)
> --Stopping nodes
> ```

2 minutes seems plenty for this process. I don't think this case needs an increased timeout but of course this is a stress test so if the added traffic is making the base node unable to provide data for mining that is an issue to be fundamentally addressed.

integration_tests/features/StressTest.feature
hansieodendaal: **NIT:** That is true, but from a test resiliency viewpoint, this step may be necessary, perhaps with a `# TODO: Remove this detect broadcast step when the system is more resilient` comment?  In my current test I have been waiting 3800 s for that step to complete for the `When I send 1400 transactions` test case and it did not finish.

hansieodendaal:
> > Quick feedback. Running this timed out close to the end. Maybe increase function timeout here to `1200*1000`:
> > ```
> > Then('all nodes are at height {int}', {timeout: 120*1000},async function (height) {
> >     await this.forEachClientAsync(async (client, name) => {
> >         await waitFor(async() => client.getTipHeight(), height, 115*1000);
> >         const currTip = await client.getTipHeight();
> >         console.log(`Node ${name} is at tip: ${currTip} (should be ${height})`);
> >         expect(currTip).to.equal(height);
> >     })
> > });
> > ```
> > 
> > 
> > Console:
> > ```
> > ..Node NODE1 is at tip: 18 (should be 18)
> > waiting for process... 500 50
> > waiting for process... 500 50
> > waiting for process... 500 100
> > waiting for process... 500 100
> > waiting for process... 500 150
> > waiting for process... 500 150
> > waiting for process... 500 200
> > waiting for process... 500 200
> > Node BaseNode0 is at tip: 14 (should be 18)
> > FNode NODE2 is at tip: 14 (should be 18)
> > --Stopping nodes
> > ```
> 
> 2 minutes seems plenty for this process. I don't think this case needs an increased timeout but of course this is a stress test so if the added traffic is making the base node unable to provide data for mining that is an issue to be fundamentally addressed.

**NIT:** Not hard and fast on this really, just providing some test feedback, to be used or not.
Maybe a slightly different philosophy here. Give the test ample opportunity to finish but provide performance metrics as a summary. Then most would be able to run the test. Also, `Wallet Routing Mechanism` shows an example where a table of values can be provided so this could be done stepwise more difficult.

integration_tests/features/StressTest.feature
philipr-za: I would argue that the purpose of this test is not for it to "finish" successfully, it's to be able to produce a stress test that is close to a real world configuration of the network and reproduce the issues we see when we do a stress test on the main network to as to debug how it breaks.

As such I wouldn't want to add steps that take this further away from how the real network behaves in order to achieve a passing test.

integration_tests/features/support/steps.js
philipr-za: Honestly I feel that these timeouts should actually be short because these components will be used by other tests where we want to enforce a short timeout. So lengthening them for the stress test I don't think is correct. If you need them long which using the stress test to find an issue then you can update them.

hansieodendaal:
Will be able to test again when the merge mining fix #2699 is merged in.

PR #2690, opened 2021-02-24
Use rusttls for dnssec for peer seeds

sdbondi:
Use `dns-over-rustls` feature on `trust-dns-client` crate. This should
remove the `openssl-sys` dependency from many crates that depend on the
`tari-p2p` crate.

Note: openssl-sys is still a dependency of hyper and reqwest (merge mining proxy) as well as `git2` build dependency
          but I expect that removing the openssl dep from tari-p2p will remove it from release builds of the base node and wallet   

UPDATE: peer seed resolver works on local base node test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3270,6 +3270,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ring"
+version = "0.16.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+dependencies = [
+ "cc",
+ "libc",
+ "once_cell",
+ "spin",
+ "untrusted",
+ "web-sys",
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "rmp"
 version = "0.8.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


StriderDM:
Not sure it will be as simple as this, see:
https://crates.io/crates/native-tls
https://crates.io/crates/trust-dns-rustls

In one way or another openssl-sys ends up as a dependency.

The mobile build will need to be updated as well as the mobile apps to remove openssl as a dependency. However it seems that for mobile, openssl will always be used.

sdbondi:
Yeah not 100% sure, difficult to trace all the dependencies. 

AFAICS the dns-over-rustls feature does not use native-tls (there is a dns-over-native-tls feature for that)
and trust-dns-rustls only has openssl as a dev dependency. So I expect that we at least won't require openssl for release builds.

Since we are just changing the peer seed resolver (no other component is effected) and that works on my local base node test, I figure it's worth a try.

StriderDM:
Been running through some of the docs for rustls and it's dependencies.

We will definitely need to check the mobile apps with this, see here:
https://github.com/briansmith/ring#online-automated-testing

For the following platforms:
```
aarch64-linux-android
armv7-linux-androideabi
aarch64-apple-ios
```

hansieodendaal:
Tested this on Windows 10 after removing Open SSL from the path; runtime works perfectly without it.

**Note:** Tested (1) base node, (2) console wallet, (3) merge mining proxy and (4) mining node.

PR #2692, opened 2021-02-24
[merge-mining] Add Tari aux chain data to getlastblockheader

sdbondi:
## Description

Merge mining proxy now includes tari block header data as aux chain data
    on getlastblockheader call.

## Motivation and Context

Pool uses this to know if a new block template should be requested.

## How Has This Been Tested?

Tested on nodejs-pool fork

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,9 +11,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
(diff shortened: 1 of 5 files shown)


StriderDM:
Please rebase against https://github.com/tari-project/tari/pull/2699 after it merges.

PR #2694, opened 2021-02-24
[base-node] Fix a possible error when state is not reinitialized

stringhandler:
Difficult to verify, but added in a line that reinitializes the validator state in the header sync. My theory is that it could be in an invalid state from a previous branch, so best to reset it just before we're adding headers to it again

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4472,7 +4472,7 @@ dependencies = [
 
 [[package]]
 name = "tari_wallet_ffi"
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
I think this is worth a try, if the bug disappears we'll know where to look

SWvheerden:
This is a good possibility that this is a bug. 
Recommend we merge it in

PR #2710, opened 2021-03-01
[tests] Add cucumber reports and tests clean up

stringhandler:
## Description

Add `cucumber-pretty` and the ability to generate an HTML report when running cucumber. Also shifted around some of the `@critical` and `@long-running` tags


SWvheerden (approved):
LGTM

PR #2715, opened 2021-03-01
Remove wallet from base node

hansieodendaal:
## Description
Removed the internal base node wallet from the base node application, as it is now deprecated.

_**Note:** Dependent on #2709; must be merged after that one._

## Motivation and Context
See above

## How Has This Been Tested?
- Unit tests
- System-level tests:
  - base node sync from scratch
  - base node, stand-alone wallet, mining node

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

* [X] My change requires a change to the documentation.
* [X] I have updated the documentation accordingly.


delta1:
needs a cargo fmt

delta1:
it's awesome to see these big red PRs, minus 2k lines! 👏

applications/tari_base_node/src/command_handler.rs
```diff
@@ -1243,625 +930,21 @@ impl CommandHandler {
...
-            let code = QrCode::new(qr_link).unwrap();
-            let image = code
-                .render::<unicode::Dense1x2>()
-                .dark_color(unicode::Dense1x2::Dark)
-                .light_color(unicode::Dense1x2::Light)
-                .build();
-            println!("{}", image);
-            println!();
-        }
         println!("======== Base Node ==========");
```
SWvheerden: We can remove this list as well. There is only going to be the basenode in the whoami

applications/tari_base_node/src/parser.rs
```diff
@@ -103,12 +88,6 @@ pub struct Parser {
     command_handler: Arc<CommandHandler>,
 }
 
-const MAKE_IT_RAIN_USAGE: &str = "\nmake-it-rain [Txs/s] [duration (s)] [start amount (uT)] [increment (uT)/Tx] \
-                                  [\"start time (UTC)\" / 'now' for immediate start] [public key or emoji id to send \
-                                  to] [message]\n";
```
SWvheerden, line 108: We need to add this the console_wallet. If we remove this, we lose this piece of documentation. The console wallet only has the command listed.

SWvheerden: We should probably add one for coin-split as well. We are deleting its only documentation as well.

common/src/configuration/global.rs
```diff
@@ -77,9 +77,7 @@ pub struct GlobalConfig {
     pub base_node_tor_identity_file: PathBuf,
     pub wallet_db_file: PathBuf,
     pub console_wallet_db_file: PathBuf,
-    pub wallet_identity_file: PathBuf,
     pub console_wallet_identity_file: PathBuf,
-    pub wallet_tor_identity_file: PathBuf,
     pub console_wallet_tor_identity_file: PathBuf,
```
SWvheerden, line 81: We should probably look at removing these as well. The wallet is not supposed to use them anymore. Its saves this data inside of the database.

SWvheerden (commented):
Looks good, we should just ensure we don't lose all the current documentation. I left comments on those. We can also remove some of the other console_wallet id files They should not be used.

applications/tari_base_node/src/parser.rs
delta1: Both of those are already in the console wallet.

common/src/configuration/global.rs
hansieodendaal: Agreed, they should be removed, but I think its for another PR, has to do with the console wallet.

applications/tari_base_node/src/parser.rs
hansieodendaal: Agreed; another PR.

PR #2719, opened 2021-03-02
Update Cucumber stress test to be a Scenario Outline

philipr-za:
## Description
This PR updates the cucumber stress test into a Scenario Outline that test sending 10, 100, 1000 and 10000 transactions. Some small fixes were required to get the stress test to work including plumbing in the `flood_ban_max_msg_count` config setting so that the wallet and base nodes don’t ban each other for message flooding during a cucumber test.

Another problem encountered was that when you send 10_000 transactions when the broadcast monitoring timeout is low, which is desired in other cucumber tests, the CPU load when we read the last transactions is very high so this PR include the ability to create a “stress-test” wallet that has a long broadcast monitoring timeout than a standard wallet.

Currently the stress test can send all 10_000 transactions but at that stage the base nodes ban each other when they are unable to sync headers due to RPC concurrency limits so the mined blocks no longer propagate.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


hansieodendaal:
With a similar test setup, I could run 10,000 transactions successfully with:
- `And I have stress-test wallet ... with broadcast monitoring timeout 60`
- RPC: `maximum_concurrent_sessions: Some(10000)` instead of `maximum_concurrent_sessions: Some(100)`
- `flood_ban_max_msg_count: 50000` instead of `flood_ban_max_msg_count: 1000`
- `flood_ban_timespan: Duration::from_secs(60)` instead of `flood_ban_timespan: Duration::from_secs(100)`

philipr-za:
> With a similar test setup, I could run 10,000 transactions successfully with:
> 
> * `And I have stress-test wallet ... with broadcast monitoring timeout 60`
> * RPC: `maximum_concurrent_sessions: Some(10000)` instead of `maximum_concurrent_sessions: Some(100)`
> * `flood_ban_max_msg_count: 50000` instead of `flood_ban_max_msg_count: 1000`
> * `flood_ban_timespan: Duration::from_secs(60)` instead of `flood_ban_timespan: Duration::from_secs(100)`

Thanks for the numbers. I found that simply increasing the RPC `maximum_concurrent_sessions` to 10000 and the `flood_ban_max_msg_count` to 100000 was enough for a 10000 message stress test to pass.

I believe we need to discuss this further because I don't think just upping the limit is the correct way to handle this but at least the cucumber test updates in this PR fully pass now.

philipr-za:
So I found a bug where the Transaction Broadcast protocol was not closing its RPC session after making it call. This means that it held the RPC session open for its full retry timeout without using it. By closing the session between tries I was able to get the 10_000 transaction stress test to pass with 1000 `max_concurrent_sessions` and the increased `flood_ban_max_msg_count`

integration_tests/features/support/steps.js
```diff
@@ -546,14 +585,18 @@ Then(/all wallets detect all transactions are at least Completed/, {timeout: 120
...
+            console.log("\nNo transactions for " + walletName + "!")
+            expect(false).to.equal(true);
+        }
+        console.log("\nDetecting", txIds.length, "transactions as at least Completed: ", walletName, txIds)
         for (i = 0; i < txIds.length; i++) {
             // Get details
-            console.log("\n" + wallet.name + ": Waiting for transaction " + txIds[i] + " to register at least Completed in the wallet ...");
-            await waitFor(async() => wallet.getClient().isTransactionAtLeastCompleted(txIds[i]), true, 1100*1000, 5*1000, 5);
+            console.log("(" + (i + 1) + "/" + txIds.length + ") - " + wallet.name + ": Waiting for TxId:" + txIds[i] + " to register at least Completed in the wallet ...");
+                                                                                                                           await waitFor(async() => wallet.getClient().isTransactionAtLeastCompleted(txIds[i]), true, 1100*1000, 5*1000, 5);
```
hansieodendaal: This got pushed out to the right somehow, must have been me with the PR I submitted to your branch:
```suggestion
            await waitFor(async() => wallet.getClient().isTransactionAtLeastCompleted(txIds[i]), true, 1100*1000, 5*1000, 5);
```

common/src/configuration/global.rs
```diff
@@ -117,6 +117,7 @@ pub struct GlobalConfig {
     pub console_wallet_notify_file: Option<PathBuf>,
     pub auto_ping_interval: u64,
     pub blocks_behind_before_considered_lagging: u64,
+    pub flood_ban_max_msg_count: usize,
```
hansieodendaal, line 120: I believe this option should be added to the config toml files, `tari_config_example.toml` and `windows.toml`, as well.

integration_tests/features/StressTest.feature
```diff
@@ -1,39 +1,72 @@
...
+        When I wait for wallet WALLET_A to have at least 5100000000 tari
+
+        Then I coin split tari in wallet WALLET_A to produce <NumTransactions> UTXOs of 5000 uT each with fee_per_gram 20 uT
+        When I merge mine 3 blocks via PROXY
+        When I merge mine <NumCoinsplitsNeeded> blocks via PROXY
+        Then all nodes are at current tip height
+        Then wallet WALLET_A detects all transactions as Mined_Confirmed
+        When I send <NumTransactions> transactions of 1111 uT each from wallet WALLET_A to wallet WALLET_B at fee_per_gram 20
+        # Mine enough blocks for the first block of transactions to be confirmed.
+        When I merge mine 4 blocks via PROXY
```
hansieodendaal, line 26: In my testing for the 10,000 transaction scenario, the `Then while merge mining via PROXY...` step timed out every time when it wants to detect the 1st transaction in the list as mined. Investigating the logs, only 13 consecutive full blocks were mined, where theoretically 16 are needed. The specific transaction in question was still in the mempool. Thus, mine more blocks before starting the detection.
```suggestion
        When I merge mine <NumCoinsplitsNeeded> blocks via PROXY
```

philipr-za: The statement after this mines a block every time it times out so it should catch up. I feel that is more representative of reality compared to just mining a huge number of blocks in one go.

hansieodendaal: Sadly, in my tests, it did not catch up, but timed out, hence my comment. 

To pause between mining blocks is not a bad idea at all, which could be done in a separate function, it is just that in this case not enough blocks have been mined before the detection function times out.

philipr-za: Then something is broken because this statement:
`Then while merge mining via PROXY all transactions in wallet WALLET_A are found to be Mined_Confirmed` should mine a block every 5 seconds.

hansieodendaal: But let us wait and see if it passes in `ci/circleci: run-integration-tests`; if so all good then it must be something in my desktop environment.

PR #2721, opened 2021-03-02
[chore] Force update generic-array dep in Cargo.lock

delta1:
Update `generic-array` dep in Cargo.lock due to security advisory. 

Temporary fix until tari-crypto can be fully updated (see #2714). 

With thanks to @mbrubeck 🙏


StriderDM (approved):
LGTM.

PR #2723, opened 2021-03-03
[network] Add explicit rejection reason to RPC handshake

sdbondi:
## Description

Client RPC handshake can now be explicitly rejected by the server
for the following reasons:

- The given protocol version is not supported
- The server has reached it's session limit
- The protocol is not supported
- Some other unspecified reason

Additionally, add an explicit error for server responses that are larger
than the permitted max frame size.

This protocol change is backward compatible.

## Motivation and Context

Previously if the session concurrency limit is reached for a peer, the peer would close the client's connection immediately.
As with other request/response based protocols (e.g. HTTP), if a server is not available the correct response should be returned. This PR adds these responses. Helpfully, this allows the client to act on a given rejection reason appropriately (e.g. try another node if the first node is too busy)

## How Has This Been Tested?

New unit tests for handshake and _smoke_ test for the rejection interaction.
Tested backward compatibility by syncing a base node on current stibbons network, as well as sending txns via the console wallet to an old wallet.   

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -108,7 +108,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                     debug!(target: LOG_TARGET, "{}", err);
                 },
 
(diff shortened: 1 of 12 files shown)


philipr-za (dismissed):
LGTM

PR #2725, opened 2021-03-03
[chore] Replace divide with checked_div

stringhandler:
Remove plain /  and replace with `checked_div` to avoid divide by zero errors


delta1:
seems to be a compilation error

PR #2726, opened 2021-03-03
[tests] Add custom log path for cucumber tests

stringhandler:
- Added custom log file for cucumber tests. A custom file can also be specified by the world parameters.
- reduced CI of cucumber to only run critical and not long-running
- updated readme
- Removed unused files


hansieodendaal:
Tested this, works well for specifying the base node log4rs file, we just need this for the proxy and console wallet as well.

PR #2728, opened 2021-03-04
Write up of wallet to wallet negotiation with TariScript

philipr-za:
## Description
This PR contains the detailed writeup of the negotiation process that to wallets need to engage in with the new TariScript proposal

This was put into a new mdbook in the repo until `/docs/` to view the document do the following:

In a terminal navigate to /docs
`cargo install mdbook`
`mdbook serve`
In a browser navigate to `localhost:3000`

Also contains a few typo fixes for the RFC

## How Has This Been Tested?
Just docs

## Types of changes

* [ ] New Tests
* [x] Documentation

## Checklist:


docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+4. Script \\( \alpha_a \\) whose hash \\( \sigma_a = H(\alpha_a) \\),
+5. Script input \\( \theta_a \\) for script \\( \alpha_a \\),
+6. Script key \\( k_{sa} \\) whose public value (\\( K_{sa} = k_{sa} \cdot G \\)) is the key left over when script \\( \alpha_a \\) is executed with input \\( \theta_a \\)   
+7. Height \\( h_a \\) where \\( C_a \\) was mined.
+
+Alice then chooses the following values:
+1. Amount \\( v_b \\) being sent to Bob
+2. Fee value \\( f \\). The amounts need to balance out so \\( v_a = v_b + f \\)    
+3. A random nonce \\( r_a \\) whose public value is \\( R_a = r_a \cdot G \\)
+4. An UTXO offset key for Bob \\( k_{Ob} \\) whose public value is \\( K_{Ob} = k_{Ob} \cdot G \\)
```
SWvheerden:
```suggestion
4. An UTXO offset key \\( k_{Ob} \\)  for Bob's UTXO \\( C_b \\)  whose public value is \\( K_{Ob} = k_{Ob} \cdot G \\)
```
In an Alice send to Bob use case, the offset key is chosen by Alice, and only Alice knows the offset key \\( k_{Ob} \\) 

EDIT, I see it is correct below, but the text here above confused me.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+
+Alice then sends the following values to Bob:
+
+| Symbol  | Description                                      |
+|---------|--------------------------------------------------                                   |
+| \\( R_a \\)       | Alice's Public Nonce                                                      |
+| \\( v_b \\)       | Amount being sent to Bob                                                  |
+| \\( f \\)         | fee                                                                       |
+| \\( m \\)         | Transaction metadata (currently just lockheight)                          |
+| \\( \sigma_b \\)  | Hash of \\( C_b \\)'s spending script,  \\( \sigma_b = H(\alpha_b) \\)    |
```
SWvheerden: Does this not need to be the full script. Although only the hash is required at this time, to spend it in the future, he will need to know the script.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+| \\( K_{Ob} \\)    | Public UTXO script offset                                                 |
+| message           | A unicode string                                                          |
+
+### Bob Replies
+Once Bob receives these values he will choose the following:
+1. Spending key \\( k_b \\) of \\( C_b \\)
+2. A random nonce \\( r_b \\) whose public value is \\( R_b = r_b \cdot G \\)
+3. His script key \\( k_s \\) whose public value \\( K_s = k_s \cdot G \\) must be the final value on the stack after script \\( alpha_b \\) executes.
```
SWvheerden, line 65: What script are we spending. This is important at this point in time. Alice sends Bob the script hash, so the script needs to be finalized at this point in time. If Bob only now chooses his script key, then the script cannot contain that. 

For example, if the script contains a `CheckSig` or the like, then Alice needed to know the public key \\(K_s\\). But if the script ONLY contains a `NO_OP` then this will work, but then Bob also does not have to choose his \\(k_s\\) now.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+    height : u64,
+    /// Script private key, k_s
+    script_private_key: PrivateKey,
+}
+```
+**NOTE:** The height in the previous struct will need to be updated AFTER the transaction negotiation is complete once the Receiver has detected the UTXO has been mined.
+
+
+## Questions
+1. Does the receiver need to get the script itself earlier than finding it in the Finalized transaction?
```
SWvheerden, line 215: This depends on what you are doing and what the script is. 
In a normal Alice spends to Bob with Bob's script key in the script, I would argue that Bob chooses the script.  I would argue that in most cases the receiver would like to or should create the script.  Only in multi-party cases or one-sided transactions should the sender choose the script.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+2. Do parties agree of the Output Features of the received output or does Bob just choose it? Currently it's assumed default
```
SWvheerden, line 218: I think we can leave this as assumed default as in most cases this does not matter.  But the mere fact that it's in the tx and signed by the script and in the total script offset at least the person who signs the script_offset needs to agree on it. If both parties sign it, they both need to agree to it. In most cases, this will only be the sender.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+3. Do you need the script offset public key in both the Transaction Input and the Transaction Output?
```
SWvheerden, line 219: This is a very good question. As per current in the RFC it's mentioned that it is in both. But technically it can be dropped out of the input as the offset key is only used in the offset. After it's mined, it's never used again. I think we left it in because it's part of the "utxo" object. But like the Rangeproof this can in all safety be dropped. It's probably a good optimization.

SWvheerden (commented):
Looks good, the script thou complicates the tx negotiation a lot.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: That's what I thought. I would like to keep the process described in this document as generic as possible so it should work for any script. So that does mean we need to add in one extra round of communication where Bob sends script data to Alice. When implementing standard MW Tx we can assume the NO_OP script and skip that step.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: Agreed, answered this point in the questions below. I believe then Alice doesn't send the script hash but rather Bob sends her the script in an extra round of communication.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: Ja, this was taken direct from the RFC example where the script is assumed NO_OP. I agree with this and the comments below address it further.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: Will update to be clearer.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: Ja, I think in the implementation we can assume default for now because only Coinbases currently use a non-default value but will leave it in this Doc description for a time when that is not the case.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: Cool, that is what I thought.

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: @mikethetike What is your take on how we should handle exchanging the script before hand. In certain use cases we can just assume the script (Like a standard transaction) but if we want to keep the wallet implementation generic regarding the script we will need to include this extra round of communication. 

When we get to the implementation should we use the assumed NO_OP script for standard Tx's and then implement a second negotiation protocol for generic scripts?

philipr-za: Actually thinking about this more the receiver actually doesn't even need to know the details of Bob's script to construct any part of the transaction because Bob is the one that commits to it in the beta key of his Rangeproof. So I think we can do it this way:
1. Alice send the initial sending data including the UTXO script offset K_O
2. Bob replies with his UTXO and the script he is applying to his UTXO
3. Alice can confirm the script Bob gives her is correct by checking the Rangeproof and if she agrees with the nature of the script then she completes the Tx

I guess the only different case would be a fancy script that requires Alice to give Bob some info to construct it, I suppose that can be part of the initial message?

@SWvheerden ^

SWvheerden: I agree with this. All Bob requires to complete the UTXO is the public script offset key. Alice can give him this. 
But from a sender point, in 99% of cases do we care what the script is? We only care about the script if it's some fancy script where there is extra stuff in it and only then if there is a chance we can be the spender. But most scripts will only be of concern for the receiver.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+3. His script key \\( k_s \\) whose public value \\( K_s = k_s \cdot G \\) must be the final value on the stack after script \\( alpha_b \\) executes.
+4. Output features \\( F_b \\) for \\( C_b \\) if any.
+
+Bob produces the rangeproof \\(RP_b) for the modified commitment
+
+\\( \hat{C}_b = (k_b + \beta_b) \cdot G + v_b \cdot H \\) 
+
+where 
+
+\\( \beta_b = H(\sigma_b || F_b || K_{Ob}) \\).
```
stringhandler: What is K_0b in this case?

philipr-za: The Public UTXO Offset key given to Bob by Alice

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+
+Bob then produces his partial signature of the transaction as per vanilla MW:
+
+\\( s_b = r_b + ek_b \\)
```
stringhandler, line 82: Should this not be `e*beta_b`?

philipr-za: Nope, beta is used for the modified Rangeproof key

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+| Transaction Kernel |       |
+|--------------------|-------|
+| Public Excess      | \\( X_s + K_b \\)    |
+| Signature          | \\( s, R \\)         |
+| Fee                | f                    |
+| Metadata           | m                    |
+
+Alice will now send this finalized transaction back to Bob so he has a copy of it and the script and both parties can now submit this transaction to the blockchain.
+
+### Bob after transaction broadcast
```
philipr-za: I just realized that Alice will also have to do this for her Change output.

docs/src/wallet_to_wallet_with_tariscript.md
```diff
@@ -0,0 +1,205 @@
...
+
+Alice then produces a script signature (\\( s_{sa}, R_{sa} \\)) for spending \\( C_a \\) by signing \\( H( R_{sa} || \alpha_a || \theta_a || h) \\) with her private script key \\( k_{sa} \\).
+
+Alice can now calculate the script offset, \\( \gamma \\) for this transaction which will include the script of the input \\( C_a \\) and the two outputs \\( C_b \\) and \\( C_c \\).
+
+\\( \gamma =  k_{sa} - k_{Ob}U_b - k_{Oc}U_c \\)
+
+### Final transaction
+
+Alice can now construct the final transaction:
```
stringhandler, line 115: I think it's worth adding the `offset` and `script_offset` at the transaction level to this table

docs/src/wallet_to_wallet_with_tariscript.md
stringhandler: I may be missing something, but can the kernel signature still be validated with this signature? Won't the excess contain `beta_b - offset`?

philipr-za: You don't actually put the modified commitment into the UTXO, it is constructed for the rangeproof calculation and the verifier constructs it when they verify. The normal commitment is what is in the UTXO and what contributes to the excess.

stringhandler: nvm, I see

docs/src/wallet_to_wallet_with_tariscript.md
philipr-za: So after some offline discussion we decided that Alice will construct the script and send the full script to Bob in round 1

stringhandler (approved):
Nice. I think there may be changes in future, but they can be iterated in further prs

PR #2732, opened 2021-03-05
Expiry of SAF messages

StriderDM:
## Description
This PR will add an expiry timestamp to SAF messages.

When trying to forward or store these messages, should the timestamp be less than the current time, the message will be discarded.

Should there not be a expiry timestamp, the message will be treated as normal for backwards compatibility.

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/dht/src/config.rs
+++ b/comms/dht/src/config.rs
@@ -42,6 +42,10 @@ pub struct DhtConfig {
     /// Send to this many peers when using the propagate strategy
     /// Default: 4
     pub propagation_factor: usize,
+    /// The amount of seconds added to the current time (Utc) which will then be used to check if the message has
+    /// expired or not when processing the message
+    /// Default: 10800
+    pub saf_msg_validity: Duration,
     /// The maximum number of messages that can be stored using the Store-and-forward middleware.
     /// Default: 100,000
     pub saf_msg_storage_capacity: usize,
@@ -172,6 +176,7 @@ impl Default for DhtConfig {
             flood_ban_max_msg_count: 1000,
             flood_ban_timespan: Duration::from_secs(100),
             offline_peer_cooldown: Duration::from_secs(24 * 60 * 60),
+            saf_msg_validity: Duration::from_secs(10800),
         }
     }
 }
--- a/comms/dht/src/envelope.rs
+++ b/comms/dht/src/envelope.rs
@@ -24,6 +24,7 @@ use bitflags::bitflags;
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 use std::{
+    cmp,
     convert::{TryFrom, TryInto},
     fmt,
     fmt::Display,
@@ -34,6 +35,33 @@ use thiserror::Error;
 
 // Re-export applicable protos
 pub use crate::proto::envelope::{dht_header::Destination, DhtEnvelope, DhtHeader, DhtMessageType, Network};
+use chrono::{DateTime, NaiveDateTime, Utc};
+use prost_types::Timestamp;
+use tari_utilities::epoch_time::EpochTime;
+
+/// Utility function that converts a `chrono::DateTime<Utc>` to a `prost::Timestamp`
+pub(crate) fn datetime_to_timestamp(datetime: DateTime<Utc>) -> Timestamp {
+    Timestamp {
+        seconds: datetime.timestamp(),
+        nanos: datetime.timestamp_subsec_nanos().try_into().unwrap_or(std::i32::MAX),
+    }
+}
+
+/// Utility function that converts a `prost::Timestamp` to a `chrono::DateTime<Utc>`
+pub(crate) fn timestamp_to_datetime(timestamp: Timestamp) -> DateTime<Utc> {
+    let naive = NaiveDateTime::from_timestamp(timestamp.seconds, cmp::max(0, timestamp.nanos) as u32);
+    DateTime::from_utc(naive, Utc)
+}
+
+/// Utility function that converts a `chrono::DateTime` to a `EpochTime`
+pub(crate) fn datetime_to_epochtime(datetime: DateTime<Utc>) -> EpochTime {
+    EpochTime::from(datetime)
+}
+
+/// Utility function that converts a `EpochTime` to a `chrono::DateTime`
+pub(crate) fn epochtime_to_datetime(datetime: EpochTime) -> DateTime<Utc> {
+    DateTime::from(datetime)
+}
 
 #[derive(Debug, Error)]
 pub enum DhtMessageError {
@@ -110,6 +138,7 @@ pub struct DhtMessageHeader {
     pub network: Network,
     pub flags: DhtMessageFlags,
     pub message_tag: MessageTag,
+    pub expires: Option<EpochTime>,
 }
 
 impl DhtMessageHeader {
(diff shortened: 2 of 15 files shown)


comms/dht/src/envelope.rs
```diff
@@ -110,6 +135,7 @@ pub struct DhtMessageHeader {
     pub network: Network,
     pub flags: DhtMessageFlags,
     pub message_tag: MessageTag,
+    pub expires: Option<DateTime<Utc>>,
```
stringhandler: This should probably just be an Epoch, same as we use in the block header

comms/dht/src/config.rs
```diff
@@ -42,6 +42,10 @@ pub struct DhtConfig {
     /// Send to this many peers when using the propagate strategy
     /// Default: 4
     pub propagation_factor: usize,
+    /// The amount of seconds added to the current time (Utc) which will then be used to check if the message has
+    /// expired or not when processing the message
+    /// Default: 10800
+    pub saf_msg_validity: i64,
```
sdbondi: This config item should be a `std::time::Duration`

comms/dht/src/outbound/broadcast.rs
```diff
@@ -306,6 +318,9 @@ where S: Service<DhtOutboundMessage, Response = (), Error = PipelineError>
                     }
                 }
 
+                // Todo, move config value in here for expires timestamp on messages
+                let expires = Utc::now() + Duration::seconds(self.message_validity_window);
```
sdbondi: Convert `message_validity_window` (now a `std::time::Duration`) to a `chrono::Duration`

```suggestion
                let expires = Utc::now() + chrono::Duration::seconds(self.message_validity_window.as_secs());
```

For code clarity, I suggest removing the `use chrono::Duration` import, as that may be confusing in future with the canonical `std::time::Duration` and always using the unambiguous `chrono::Duration`

comms/dht/src/envelope.rs
```diff
@@ -71,6 +74,28 @@ bitflags! {
     }
 }
 
+/// Utility function that converts a `Option<chrono::DateTime<Utc>>` to a `Option<prost::Timestamp>`
+pub(crate) fn datetime_to_timestamp(datetime: Option<DateTime<Utc>>) -> Option<Timestamp> {
```
sdbondi: These utility functions can already be found in `crate::store_forward::message::` - perhaps move them into `crate::envelope` and update `crate::store_forward::message::` accordingly.

Also these utility functions should not deal with `Option` (the clue is `None` always yields a `None`) - this is what `Option::map` is for.

e.g.
```rust
/// Utility function that converts a `chrono::DateTime<Utc>` to a `prost::Timestamp`
pub(crate) fn datetime_to_timestamp(datetime: DateTime<Utc>) -> Timestamp {
    Timestamp {
        seconds: datetime.timestamp(),
        nanos: datetime.timestamp_subsec_nanos().try_into().unwrap_or(std::i32::MAX),
    }
}

let foo = Some(Utc::now());
let converted = foo.map(datetime_to_timestamp);
assert_eq!(converted, Some(.....));

let foo = None;
let converted = foo.map(datetime_to_timestamp);
assert_eq!(converted, None);

```

sdbondi (commented):
Looks good - just minor code clarity comments 

Noting that the expiry field is currently malleable and needs to be committed to, however that would be a breaking change and needs to be coordinated in another PR.

comms/dht/src/envelope.rs
StriderDM: Done.

comms/dht/src/config.rs
StriderDM: Done.

comms/dht/src/outbound/broadcast.rs
StriderDM: Done.

comms/dht/src/envelope.rs
StriderDM: Have moved them. Thanks.

comms/dht/src/config.rs
```diff
@@ -42,6 +42,10 @@ pub struct DhtConfig {
     /// Send to this many peers when using the propagate strategy
     /// Default: 4
     pub propagation_factor: usize,
+    /// The amount of seconds added to the current time (Utc) which will then be used to check if the message has
+    /// expired or not when processing the message
+    /// Default: 10800
+    pub saf_msg_validity: chrono::Duration,
```
sdbondi: Don't use chrono::Duration for public interfaces. Rather use std::time::Duration and convert it further down. Without looking deeply into it or following the flow, we should be able to avoid using chrono (directly) at all for this feature, `std::time::Duration` for the validity setting and `Epoch::now()` to convert to `protobuf Timestamp`

comms/dht/src/store_forward/forward.rs
```diff
@@ -189,6 +190,19 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
             return Ok(());
         }
 
+        if let Some(expires) = &dht_header.expires {
```
SWvheerden, line 193: What do we do if it does not have an expires on it?
Currently, it seems that if you want your saf messages to live longer, which is something a wallet will want to do, you can just leave this out and the messages wont expire? As they will be forwarded constantly?

Should we not add a expire timestamp here before we handle it further?

SWvheerden (commented):
Looks good, one comment about improving expiry thou. 
For an honest actor, this works great, but I think it's still exploitable by a dishonest actor.

comms/dht/src/store_forward/forward.rs
StriderDM: Currently treat it as if it had no expiry, for backwards compatibility.

comms/dht/src/config.rs
StriderDM: Think this is sorted now. Sorry had some confusion between two of your previous comments.

comms/dht/src/store_forward/forward.rs
StriderDM: There will be a separate PR to add the field to the signature, though this will be a breaking change as previously mentioned in the above comments.

comms/dht/src/store_forward/service.rs
```diff
@@ -437,7 +438,17 @@ impl StoreAndForwardService {
             Anonymous => db.find_anonymous_messages(query.since, limit).await?,
         };
 
-        Ok(messages)
+        // Reduce the overall amount of messages returned to the window, should be close enough since this is also
+        // checked when storing or propagating the messages.
+        let validity_window = chrono::Duration::from_std(self.config.saf_msg_validity).unwrap();
```
stringhandler: This is the only part of the PR I disagree with. The storing node already has rules around what it stores. Also, this gives two meanings to `saf_msg_validity`: 1. The expiry on sent messages and 2. The max value of expiry of stored messages. This is confusing about what value is being used.

I would suggest remove this change to the file completely as it's already handled by existing code

stringhandler (changes requested):
Happy in general, but I think the below should be reverted to prevent future confusion

comms/dht/src/store_forward/service.rs
StriderDM: This has been removed.

PR #2734, opened 2021-03-08
Enable wallet to validate upon reorg

hansieodendaal:
## Description
- This PR added the ability for a wallet to redo TXO and transaction validation protocols based on reorg information received from the base node when it receives new tip info data. Whenever a reorg is detected during the addition of blocks to the database backend, reorg information is updated in the blockchain database. The tip info struct has been updated to include the reorg information, and this is then provided to the wallet upon its chain metadata update cycle. In addition, two logical bugs were found and corrected in (a) the transaction validation protocol that should also validate broadcast transactions and (b) the broadcast monitoring protocol that should stop when transactions are not valid anymore.
- As an aside, fixed RPC connections that were being kept open when they should have been closed in the coinbase monitoring and transaction validation protocols.

_**Note:** The `ReorgInfo` struct contains two fields that are not being used at the moment; this is intentional and left for a follow-up optimization PR where the wallet can decide based on its own contents and chain height if TXO and transaction validation must be redone or not. The characteristics of hybrid mining with two or more algorithms are such that blockchain reorgs are numerous, especially 1 or 2 block deep reorgs. The current re-validation implementation will be triggered for all those reorgs._

## Motivation and Context
Wallets are not aware of transactions potentially being invalidated or validated if a base node they are connected to undergoes a reorg.

## How Has This Been Tested?
- Expanded cucumber test `Wallets monitoring coinbase and normal transactions after a reorg` to also verify transaction validity after a reorg.
- Verified in trace logs that the broadcast monitoring protocol stops after transactions become invalid.
- All unit test passed.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


stringhandler:
My concern here is that the ChainMetadata is sent to all peers to advertise height. Adding in the reorg data for one use case (a connected wallet needs to know about it) means that there is a lot more data travelling the network for non-related use cases. 

Let me think on it a bit, but ideally you would only want this data to be on the network for the network calls that need it

hansieodendaal:
> Let me think on it a bit, but ideally you would only want this data to be on the network for the network calls that need it

Sure. There are pros and cons to both ways.

delta1:
> ideally you would only want this data to be on the network for the network calls that need it

@hansieodendaal may I suggest that this branch is rebased against #2753 - that way the `ReorgInfo` could be added to `TipInfoResponse` instead of `ChainMetadata`. I believe TipInfoResponse is only used for the console wallet and merge mining proxy.  @mikethetike thoughts?

stringhandler:
Agree

hansieodendaal:
Will do

hansieodendaal:
@delta1 and @mikethetike, changes applied as requested.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -830,6 +839,39 @@ where B: BlockchainBackend
             block,
         )?;
 
+        // Update reorg info
+        if let BlockAddResult::ChainReorg(_, reorg_chain) = block_add_result.clone() {
+            let tip_height = reorg_chain
+                .first()
+                .expect("Cannot extract meta data from the reorged chain")
```
stringhandler: Possible `panic`, better to replace with an `ok_or` or similar?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -830,6 +839,39 @@ where B: BlockchainBackend
...
+                .block
+                .header
+                .height -
+                1;
+            let num_blocks_reorged = reorg_chain.len() as u64;
+            let last_reorg_best_block = Some(
+                reorg_chain
+                    .first()
+                    .expect("Cannot extract meta data from the reorged chain")
```
stringhandler: Same

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -830,6 +839,39 @@ where B: BlockchainBackend
...
+                num_blocks_reorged,
+                tip_height,
+            };
+            trace!(
+                target: LOG_TARGET,
+                "Updating reorg info in the database - after restore ({})",
+                reorg_info
+            );
+            let mut txn = DbTransaction::new();
+            txn.set_reorg_info(reorg_info);
```
stringhandler: This should be done inside of the same LMDB transaction as `add_block` otherwise could lead to a mismatch of data. The `mut db` prevents this currently, but the two steps are not distinct and should really be done together.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1493,6 +1535,21 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
     );
     db.write(txn)?;
 
+    // Update reorg info
+    let reorg_info = ReorgInfo {
```
stringhandler, line 1504: This seems to be a duplicate of the `add_block` logic. Are you sure this is not done twice?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1493,6 +1535,21 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+        tip_height: last_header.height,
+    };
+    trace!(
+        target: LOG_TARGET,
+        "Updating reorg info in the database - after rewind ({})",
+        reorg_info
+    );
+    let mut txn = DbTransaction::new();
+    txn.set_reorg_info(reorg_info);
+    db.write(txn)?;
```
stringhandler, line 1516: Should be done in the previous transaction

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1726,6 +1805,24 @@ fn restore_reorged_chain<T: BlockchainBackend>(
         insert_block(&mut txn, block)?;
     }
     db.write(txn)?;
+
+    // Update reorg info
+    if last_reorg_best_block.is_some() {
```
stringhandler: This is already done in `rewind_to_height`, is this duplicate code?

base_layer/common_types/src/chain_metadata.rs
```diff
@@ -142,6 +142,47 @@ impl Display for ChainMetadata {
...
+    /// The tip hash of the last reorg chain
+    pub last_reorg_best_block: Option<HashOutput>,
+    /// The number of blocks contained in the last reorg
+    pub num_blocks_reorged: u64,
+    /// The tip height the chain was at after the last reorg
+    pub tip_height: u64,
+}
+
+impl ReorgInfo {
+    pub fn new() -> Self {
```
stringhandler, line 157: I'd prefer to not have a Default for this struct. The struct should only exist if there was a reorg. Callers can use an `Option<ReorgInfo>` and use `None` for no reorg info. The `tip_height` of `0` is a valid value and cannot be used for detecting `None`

stringhandler: The `last_reorg_best_block` can then be changed to `HashOutput` and used without checking every time

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -223,137 +213,116 @@ where
...
+                task::spawn(async move {
+                    info!(
                         target: LOG_TARGET,
-                        "Received a base node response currently unaccounted for: {:?}",
-                        message
+                        "Chain reorg detected, restarting transaction validation"
                     );
-                },
-            }
+                    if let Err(e) = transaction_service
```
stringhandler, line 281: Should these not be propagated with `?`

base_layer/wallet/src/transaction_service/protocols/transaction_coinbase_monitoring_protocol.rs
```diff
@@ -389,6 +389,7 @@ where TBackend: TransactionBackend + 'static
                             "Coinbase transaction (TxId: {}) not mined yet, still waiting.", self.tx_id,
                         );
                         // Wait out the remainder of the delay before proceeding with next loop
+                        drop(client);
```
stringhandler, line 392: What is the reason for dropping here?

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -328,6 +328,7 @@ where TBackend: TransactionBackend + 'static
                                             );
                                             e
                                         });
+                                    drop(client);
```
stringhandler, line 332: What is the reason for dropping here

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -348,6 +349,7 @@ where TBackend: TransactionBackend + 'static
                                     );
                                     e
                                 });
+                                drop(client);
```
stringhandler, line 355: What is the reason for dropping here

integration_tests/helpers/util.js
```diff
@@ -153,7 +153,7 @@ function consoleLogBalance(balance) {
     );
 }
 
-function consoleLogCoinbaseDetails(txnDetails) {
+function consoleLogTransactionDetailsFormat2(txnDetails) {
```
stringhandler: Let's just use one formatter. I as a developer would not know in which case to use one or the other

integration_tests/features/support/steps.js
```diff
@@ -965,24 +1008,103 @@ Then(/the number of coinbase transactions for wallet (.*) and wallet (.*) are (.
...
+    let validB = transactionsB[0]["valid"];
+    for (let i = 0; i<transactionsB.length; i++) {
+        if (transactionsB[0]["valid"] != transactionsB[i]["valid"]) {
+            expect("\n" + walletNameB + "'s coinbase transactions do not have a consistent validity status").to.equal("")
+        };
+    }
+    expect(validA).to.equal(!validB)
+});
+
+Then(/all normal transactions for wallet (.*) and wallet (.*) have consistent but opposing validity/, {timeout: 20*1000}, async function (
```
stringhandler: Can't these two methods be combined with a regular expression rather? There is a lot of code being duplicated here

integration_tests/features/support/steps.js
```diff
@@ -965,24 +1008,103 @@ Then(/the number of coinbase transactions for wallet (.*) and wallet (.*) are (.
...
+    if ((transactions === undefined)) {
+        expect("\nNo transactions found!").to.equal("")
+    }
+    let valid = transactions[0]["valid"];
+    for (let i = 0; i<transactions.length; i++) {
+        expect(transactions[i]["valid"]).to.equal(true)
+    }
+});
+
+Then(/all normal transactions for wallet (.*) are valid/, {timeout: 20*1000}, async function (
```
stringhandler: Same for this duplicate code

integration_tests/features/support/steps.js
```diff
@@ -287,6 +287,21 @@ When(/I mine (\d+) blocks on (.*)/, {timeout: 600*1000}, async function (numBloc
     this.tipHeight += parseInt(numBlocks);
 });
 
+When(/I sync mine (\d+) blocks on (.*)/, {timeout: 1200*1000}, async function (numBlocks, name) {
```
stringhandler: What does `sync mine` mean?

stringhandler (commented):
I appreciate there is a lot of work in this PR, but there are a lot of cases where you are preventing the rust compiler from helping here. Any time you implement something out of the ordinary like calling `drop` or `expect`, or not propagating an error you should have a comment justifying the reason that you done so. When you do something like that, a code reviewer will immediately have questions, so you want to answer them in the comment. Obviously it is better to not use them at all, but in some cases I understand that you need to.

The approach you've used here requires the base node to keep track of the reorg and there are a few problems that can arise from this:
1. Due to a bug (current or future), this is falls out of sync and the data is junk
2. There is no incentive for base nodes to keep this data. They could also lie
3. We need to cover all scenarios in tests to make sure that this data never gets out of sync

As an alternative, what do you think about having the wallet store the header hash of the tip that it last saw, and then sending it to the base node to find out if it's still in the best chain? The advantage of this is that there are fewer logic paths in the base node implementation and no chance of ReorgInfo getting out of sync. The methods for retrieving data and finding chain splits already exist in the base node, so require less testing

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: There are different pathways/edge cases catered for here, one of which are when the entire database a base node has assembled are reorged away.

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Not duplicate code, different pathways/edge cases catered for here.

base_layer/wallet/src/transaction_service/protocols/transaction_coinbase_monitoring_protocol.rs
hansieodendaal: Will add a comment; this is fixing a bug where the RPC connections are not dropped but stay open.

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
hansieodendaal: As per above, will add a comment.

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
hansieodendaal: As per above, will add a comment.

integration_tests/features/support/steps.js
hansieodendaal: sync mine = mine some blocks while waiting for all other nodes to keep in sync

The problem here is these two regular expressions look similar to cucumber:
`When(/I mine (\d+) blocks on (.*) /`
       vs.
`When(/I mine (\d+) blocks on (.*) while waiting for all other nodes to keep in sync/`

Maybe this will work:
`When(/I keep all nodes in sync and mine (\d+) blocks on (.*)/`

integration_tests/features/support/steps.js
hansieodendaal: Yip, will combine

integration_tests/features/support/steps.js
hansieodendaal: Ditto

integration_tests/helpers/util.js
hansieodendaal: Yip

philipr-za:
Just some thoughts on the global comments:

> 2. There is no incentive for base nodes to keep this data. They could also lie

This is true for many elements of the base node. They could lie in the chainsplit request used in the alternative below. The incentive, I think, is this approach is less work for the BN
 
> As an alternative, what do you think about having the wallet store the header hash of the tip that it last saw, and then sending it to the base node to find out if it's still in the best chain? The advantage of this is that there are fewer logic paths in the base node implementation and no chance of ReorgInfo getting out of sync. The methods for retrieving data and finding chain splits already exist in the base node, so require less testing

This approach can work but its significantly more work for the BN to do this. Instead of just recording when reorgs happen , which is a very low overhead as its done once for every reorg event, every time an individual wallet makes a request to this base node is has to retrieve a fair amount of data and do the processing to find a chain split. As opposed to just grabbing the reorg info from the db and serving it.

base_layer/wallet/src/base_node_service/service.rs
hansieodendaal: My opinion is that it is good enough to read in the log that there was an error here, as re-validation is important but not critical, and not to stop the rest of the code from executing, which is much more important/critical.

PR #2735, opened 2021-03-08
[base-node] Add messages in last 60s to status

stringhandler:
## Description

Add a counter of the total messages received by this node in the last 60 seconds to the status log.
Also updated the status ping to be faster (every 2s) for the first minutes, and then slows down to every 30 seconds after five minutes.
Also switched `Vec` for `VecDeque` to save an order(n) operation when the buffer is at capacity.

## Motivation and Context

The main reason is to gain insight as to how busy the network is. The reason for speeding up the status ping on startup is that a user will probably be watching this closely as it starts. Ideally I would have liked to change it to run more frequently in different states of the Base Node (e.g. quickly during Sync, slow during listening), but this is a simple implementation that serves the same purpose.

### DIFF
--- a/comms/dht/src/connectivity/metrics.rs
+++ b/comms/dht/src/connectivity/metrics.rs
@@ -28,7 +28,7 @@ use futures::{
 };
 use log::*;
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, VecDeque},
     future::Future,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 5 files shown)


sdbondi (approved):
~~Looks good!~~ See below

comms/dht/src/connectivity/metrics.rs
```diff
@@ -188,9 +210,9 @@ impl<T> TimeSeries<T> {
             "TimeSeries::values exceeded capacity"
         );
         if self.values.len() == self.values.capacity() {
```
sdbondi: I just remembered that a `VecDeque`'s capacity cannot be relied upon. [playground example](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=9097195b9830da4f16f666250363f064)

I suggest keeping a separate `capacity: usize` field in `TimeSeries`

sdbondi (changes requested):
Just remembered a difference between `VecDeque` and `Vec` that will cause issues

SWvheerden:
Apart from the comment from @sdbondi , it looks good

sdbondi (approved):
LGTM

PR #2738, opened 2021-03-09
Allow prune mode nodes to rewind past pruning horizon

SWvheerden:
## Description

We need pruned nodes to be able to handle a reorg past their pruning horizon. This pr fixes prune mode to allow to happen as well as add a cucumber test proving this is possible. 

## How Has This Been Tested?

This was tested as follows:
* Cucumber test. 
* Synced a node up to tip height of 19616 with a pruning horizon of 1000

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -313,8 +313,8 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self
     }
 
-    pub fn update_kernel_sum(&mut self, header_hash: HashOutput, kernel_sum: Commitment) -> &mut Self {
-        self.transaction.update_kernel_sum(header_hash, kernel_sum);
+    pub fn update_deleted_with_diff(&mut self, header_hash: HashOutput, deleted: Bitmap) -> &mut Self {
```
stringhandler, line 311: Rather rename the `update_deteled` method

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1471,11 +1472,15 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
         last_block_height,
         last_block_height - steps_back
     );
+    let (last_header, header_accumulated_data) = db.fetch_header_and_accumulated_data(height)?;
+
+    if steps_back > metadata.pruning_horizon() && metadata.pruning_horizon() > 0 {
+        debug!(target: LOG_TARGET, "WARNING, reorg past pruning horizon");
```
stringhandler: Why not make this a `warn`

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -332,15 +332,40 @@ impl LMDBDatabase {
                     block_accum_data.deleted = DeletedBitmap { deleted };
                     self.update_block_accumulated_data(&write_txn, height, &block_accum_data)?;
                 },
+                UpdateDeletedBlockAccumulatedDataWithDiff {
```
stringhandler, line 321: The `UpdateDeletedBlockAccumulateData` enum variant is most likely not used anymore and should be deleted

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -332,15 +332,40 @@ impl LMDBDatabase {
...
+                    let height = self.fetch_height_from_hash(&write_txn, &header_hash).or_not_found(
+                        "BlockHash",
+                        "hash",
+                        header_hash.to_hex(),
+                    )?;
+                    let prev_block_accum_data = if height > 0 {
+                        self.fetch_block_accumulated_data(&write_txn, height - 1)?
+                            .unwrap_or_else(BlockAccumulatedData::default)
+                    } else {
+                        BlockAccumulatedData::default()
```
stringhandler: nit: Probably better to throw an Error (Attempt to update genesis block)

stringhandler (commented):
Nice so far, but I don't think this entirely solves the problem. 

The cucumber tests a reorg, but you really should be testing this situation:
1. The reorg happens at point R on ChainR, with chains ChainRA and ChainRB
2. On ChainRA, there is a spend of an output in ChainR, but in ChainRB that output is not spent.
3. When reorging to ChainRB, the output is no longer spent.

I am not sure that your PR solves this adequately, but I could be wrong. As far as I see it, the output will be pruned in ChainRB and unrecoverable

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1472,23 +1475,55 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
         last_block_height,
         last_block_height - steps_back
     );
-    let (last_header, header_accumulated_data) = db.fetch_header_and_accumulated_data(height)?;
 
+    let mut prune_past_horizon = false;
     if steps_back > metadata.pruning_horizon() && metadata.pruning_horizon() > 0 {
-        debug!(target: LOG_TARGET, "WARNING, reorg past pruning horizon");
+        info!(target: LOG_TARGET, "WARNING, reorg past pruning horizon");
```
stringhandler: Why not make this a `warn!`

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1472,23 +1475,55 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+        info!(target: LOG_TARGET, "WARNING, rewinding back to 0");
```
stringhandler: suggestion: Combine this with the previous line

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1471,20 +1475,55 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+        debug!(target: LOG_TARGET, "Deleting block {}", last_block_height - h,);
         let block = fetch_block(db, last_block_height - h)?;
-        if block.is_pruned() {
-            unimplemented!("Rewinding past the pruning horizon not implemented.");
-        }
         let block = Arc::new(block.clone().try_into_chain_block()?);
         txn.delete_block(block.block.hash());
         txn.delete_header(last_block_height - h);
-        txn.insert_chained_orphan(block.clone());
+        if !prune_past_horizon {
```
stringhandler, line 1497: None of these blocks can be pruned, otherwise they would error at `try_into_chain_block()`, so no need for this check

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1471,20 +1475,55 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+        for h in 0..(last_block_height - steps_back) {
+            debug!(
+                target: LOG_TARGET,
+                "Deleting blocks and utxos {}",
+                last_block_height - h - steps_back,
+            );
+            let block = fetch_block(db, last_block_height - h - steps_back)?;
+            let block = Arc::new(block.clone().into_pruned_chain_block());
+            txn.delete_block(block.block.hash());
+            removed_blocks.push(block);
```
stringhandler: This is potentially a LOT of blocks in memory

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
     Then node B is at tip BTip1
 
-
-
+@reorg
+  Scenario: Pruned mode reorg past horizon
+    Given I have a base node NODE1 connected to all seed nodes    
+    When I mine a block on NODE1 with coinbase CB1
+    Given I have a base node NODE2 connected to node NODE1
+    When I mine but don't submit a block BLOCK2 on NODE2
```
stringhandler, line 38: nit: this can be rolled into one step with some clever javascript

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    And I stop NODE2
+    When I mine 3 blocks on NODE1
+    When I create a transaction TX1 spending CB1 to UTX1
+    When I submit transaction TX1 to NODE1
+    Then NODE1 has TX1 in MEMPOOL state
+    When I mine 6 blocks on NODE1
+    Given I have a pruned node PNODE2 connected to node NODE1
+    Then node PNODE2 is at height 10
+    When I start NODE2
+    When I submit block BLOCK2 to NODE2
```
stringhandler, line 66: same here

stringhandler: When I submit blocks BLOCK2,BLOCK3,BLOCK4 to NODE2

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    When I mine but don't submit a block BLOCK18 on NODE2
+    When I mine but don't submit a block BLOCK19 on NODE2
+    When I mine but don't submit a block BLOCK20 on NODE2   
+    And I stop NODE2
+    When I mine 3 blocks on NODE1
+    When I create a transaction TX1 spending CB1 to UTX1
+    When I submit transaction TX1 to NODE1
+    Then NODE1 has TX1 in MEMPOOL state
+    When I mine 6 blocks on NODE1
+    Given I have a pruned node PNODE2 connected to node NODE1
```
stringhandler: Might be more clear to explicitly specify what the pruning horizon is here so that it can be verified without opening the js code

stringhandler: it also prevents it from breaking if the default is changed for another test

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    When I submit block BLOCK12 to NODE2
+    When I submit block BLOCK13 to NODE2
+    When I submit block BLOCK14 to NODE2
+    When I submit block BLOCK15 to NODE2
+    When I submit block BLOCK16 to NODE2
+    When I submit block BLOCK17 to NODE2
+    When I submit block BLOCK18 to NODE2
+    When I submit block BLOCK19 to NODE2
+    When I submit block BLOCK20 to NODE2
+    Then node NODE2 is at height 20
```
stringhandler: Then all nodes are at height 20

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    When I mine but don't submit a block BLOCK9 on NODE2
+    When I mine but don't submit a block BLOCK10 on NODE2
+    When I mine but don't submit a block BLOCK11 on NODE2
+    When I mine but don't submit a block BLOCK12 on NODE2
+    When I mine but don't submit a block BLOCK13 on NODE2
+    When I mine but don't submit a block BLOCK14 on NODE2
+    When I mine but don't submit a block BLOCK15 on NODE2
+    When I mine but don't submit a block BLOCK16 on NODE2
+    When I mine but don't submit a block BLOCK17 on NODE2
+    When I mine but don't submit a block BLOCK18 on NODE2
```
stringhandler, line 54: Aren't all of these blocks mined at height 2? Did this test pass?

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    When I mine but don't submit a block BLOCK19 on NODE2
+    When I mine but don't submit a block BLOCK20 on NODE2   
+    And I stop NODE2
+    When I mine 3 blocks on NODE1
```
stringhandler, line 58: I don't know if this is necessary

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    When I submit block BLOCK17 to NODE2
+    When I submit block BLOCK18 to NODE2
+    When I submit block BLOCK19 to NODE2
+    When I submit block BLOCK20 to NODE2
+    Then node NODE2 is at height 20
+    Then node NODE1 is at height 20
+    Then node PNODE2 is at height 20
+    # Because TX1 should have been re_orged out we should be able to spend CB1 again
+    When I create a transaction TX2 spending CB1 to UTX2
+    When I submit transaction TX2 to NODE1
```
stringhandler:
```suggestion
    When I submit transaction TX2 to PNODE1
```

integration_tests/features/Reorgs.feature
```diff
@@ -30,5 +30,62 @@ Feature: Reorgs
...
+    Then NODE1 has TX2 in MEMPOOL state
```
stringhandler: NODE1 is the wrong node to test here, you should be testing that the pruned node accepts it

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: Yes, these blocks can't be pruned but it does not help if save block 8,9,10 in the orphan DB, but we delete 5,6,7 for example. 

This is here if we know we are rewinding behind the pruning horizon. We know we are destroying our "big block" so we need to recreate it anyway. So no need in saving these.  Archival nodes save these to maybe swing back to them if the chain fails syncing or the like.

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: Yip, I am in two minds on this, keep it like this, or just don't give these to the mempool. Or we trigger a mempool wipe or the like.

integration_tests/features/Reorgs.feature
SWvheerden: No, what happens here, is NODE2 accepts these blocks and it moves forward in the chain.  This just stops the block from propagating.

integration_tests/features/Reorgs.feature
SWvheerden: It's here to ensure `CB1` is past lock height

integration_tests/features/Reorgs.feature
SWvheerden: NODE1 and PNODE1 should have the same chain, butt swapping it over.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1382,14 +1393,25 @@ impl BlockchainBackend for LMDBDatabase {
...
+                } else {
+                    self.fetch_block_accumulated_data(&txn, height - 1)
+                        .or_not_found("BlockAccumulatedData", "height", height.to_string())?
+                        .deleted()
+                        .clone()
+                };
+                // integrity check, the diff should always be smaller than the total
+                let total_size = block_accum_data.cardinality();
+                let diff_bitmap = block_accum_data.xor(&prev_block_accum_data);
+                assert!(diff_bitmap.cardinality() <= total_size);
```
stringhandler: This is not the place for an assert
```suggestion
               
```

stringhandler (dismissed):
Happy, will test locally

stringhandler (dismissed):

PR #2707, opened 2021-02-28
[wallet] Allow transactions to send to self

sdbondi:
## Description

Allow wallet to create and broadcast transactions to self.

- Add call to transaction service that creates and broadcasts the transaction
- Add TransactionCompletedImmediately event and hooked that up to console wallet UI
- Minor optimisations (unnecessary clones, early loop exit)

## Motivation and Context

Currently, transactions sent to self stay in pending forever as p2p negotiation cannot occur.
This PR detects and handles this case by instantly creating and broadcasting the transaction.

## How Has This Been Tested?

1 Functional and 1 Integration test. Tested manually in console wallet.

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [ ] New Tests
* [ ] Documentation

## Checklist:

* [ ] I'm merging against the `development` branch.
* [ ] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [ ] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #2741, opened 2021-03-09
[wallet] Fix duplicate coinbase key generation, derive from height

sdbondi:
Use block height as coinbase key index instead of next_key. This bug
was introduced in PR #2707.

This causes the key_index for the coinbase key manager to start from 0 every wallet start
leading to duplicate coinbase spend keys. This would cause mined blocks to be rejected (duplicate kernel)
if miners had previously won blocks with coinbases generated with this bug.


philipr-za (dismissed):

stringhandler (approved):
Seems ok. Untested

PR #2743, opened 2021-03-09
[MM_Proxy] Fix logic flaw in proxy_request_to_monerod

StriderDM:
## Description
Reverts logic.

Only the submit_block request needs special handling depending on the proxy_submit_to_origin configuration option.

Additionally also fixes the following issue with trying to parse the body of GET requests, detailed below:
```
2021-03-09 16:19:57.311444000 [tari_mm_proxy::proxy] DEBUG [monerod] request: GET http://18.132.124.81:18081/getheight 
2021-03-09 16:19:57.311534000 [tari_mm_proxy::proxy] ERROR Error handling request: JSON error: EOF while parsing a value at line 1 column 0 
```

## Motivation and Context
Logic and bug fix

## How Has This Been Tested?
XMRig and Merge Mining Proxy configured for:
Solo mining with submit to origin enabled (tari & monero test).
Solo mining with submit to origin disabled (tari only test).
Self-select with submit to origin disabled (tari & monero test, only valid configuration).

Against both Monero stagenet and mainnet.

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -22,7 +22,7 @@
 
 use crate::{
     block_template_data::{BlockTemplateDataBuilder, BlockTemplateRepository},
-    common::{json_rpc, merge_mining, monero_rpc::CoreRpcErrorCode, proxy},
+    common::{json_rpc, merge_mining, monero_rpc::CoreRpcErrorCode, proxy, proxy::convert_json_to_hyper_json_response},
     error::MmProxyError,
 };
 use bytes::Bytes;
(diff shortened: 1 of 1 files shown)


sdbondi (approved):

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -775,53 +775,55 @@ impl InnerService {
             monerod_uri,
         );
 
-        let json_response = if self.config.proxy_submit_to_origin {
+        let mut submit_block = false;
+        let body: Bytes = request.body().clone();
+        let json = json::from_slice::<json::Value>(&body[..]).unwrap_or_default();
```
sdbondi, line 780: The reason I put the json parsing only when self.config.proxy_submit_to_origin == true is because that is the only case it is needed i.e. you dont need to parse the json request an extra time when self.config.proxy_submit_to_origin is true, that and to avoid having the `mut submit_block: bool`. The perf hit is negligible (unless the sender sends massive payoads), so I'm indifferent about this change. I understand why you squashed the json error (although ideally you should only parse json for POST requests) but not completely sure why you changed the rest.

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -775,53 +775,55 @@ impl InnerService {
             monerod_uri,
         );
 
-        let json_response = if self.config.proxy_submit_to_origin {
+        let mut submit_block = false;
```
sdbondi, line 778: Mutable boolean flags that a reader needs to keep track of has code smell

sdbondi (commented):
I'm more than happy to merge as is

PR #2749, opened 2021-03-12
Prevent overflow on node_index

CjS77:
An arithmetic overflow (multiply overflow) issue can occur if leaf_index > MAX/2


CjS77:
Tthinking about it, it's probably better to leave that function as is, make it private, and put guards on things that call it.

PR #2751, opened 2021-03-14
[basenode] Minor fix for list-headers

sdbondi:
- Fixes a bug that returns many headers while header sync is in progress
- Fixes minor bug with `list-headers` that made it impossible to list the genesis block using `list-header [from-tip]`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4516,7 +4516,7 @@ dependencies = [
 
 [[package]]
 name = "tari_wallet_ffi"
-version = "0.16.23"
+version = "0.16.24"
 dependencies = [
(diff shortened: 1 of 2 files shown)


hansieodendaal (dismissed):
LGTM

**NIT**
Edge case while syncing headers; if you input `list-headers 2` for example, it lists all of the headers instead of only the two requested.

sdbondi:
@hansieodendaal Ah nice one, l've noticed that too - pushed a fix

hansieodendaal (approved):
LGTM!

stringhandler (approved):

PR #2759, opened 2021-03-16
[base-node] Fix handling of large chain reorgs in header sync

sdbondi:
## Description

Changes header sync to continue downloading headers until a higher
accumulated PoW than the current tip is achieved.

**Determine chain split hash**

image

**Download more headers until higher PoW than the current tip**
image

**Reorg chain (if necessary) once higher accumulated PoW found**  

image

**Continue downloading and validating headers until the complete, adding 1000 at a time to the end of the chain**

image

**Header sync complete**

image

## Motivation and Context

Previously, in > 1000 height reorgs, headers at split + 1000 are
compared which is insufficient to determine if the remote chain is
stronger than the local one.

## How Has This Been Tested?

New cucumber test ` Full block sync with (large|small) reorg`
New basic unit tests for header validator

Following test cases on stibbons base node:
- sync from scratch
- simple sync, node behind ~200 blocks
- split from genesis (mined from 0, then after 27 blocks, connected to main stibbons network)
- node that had previously had the erroneous weaker target difficult issue (resulting in bans) 
  - large reorg with split ~3500 blocks deep, ~9500 blocks behind  
- resume interrupted sync 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:


base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -105,23 +105,21 @@ impl<B: BlockchainBackend + 'static> BlockHeaderSyncValidator<B> {
             timestamps,
             target_difficulties,
             previous_accum,
+            // One large allocation is usually better even if it is not always used.
+            valid_headers: Vec::with_capacity(1000),
```
SWvheerden, line 109: I disagree here. I think this is too large. 
Headers can be quite large, I think we should start with a more reasonable number that is just above the median. Perhaps around 200 for now, mainnet should be way less

sdbondi: Yeah this wasn't particularly thought out (as in I didn't try to size up the block header struct to get an exact memory cost - I will now and report) although it is true that many small allocations is significantly slower than one large one. Remember this will allocate the "shallow" size of the struct (i.e. a vec on the heap is not part of what is allocated in the `valid_headers` vec) 

Because there is really (easy/reliable) no way to know for sure what is the optimal value - for initial sync it should be 1000 (the interval at which headers are "flushed") - if we're doing a large reorg it could grow to 1000s before being committed - if a node is slightly behind this is wasteful. 

I'll report on the exact cost of preallocation.

sdbondi: The "shallow size" of the block header struct is 208 bytes (`std::mem::size_of::<BlockHeader>()`) So 208kb is allocated as a single block (this excludes any nested Vecs that are part of the BlockHeader struct which are allocated separately)

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -116,13 +122,16 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                     debug!(target: LOG_TARGET, "Block header validation failed: {}", err);
                     self.ban_peer_long(node_id, err.into()).await?;
                 },
+                Err(err @ BlockHeaderSyncError::InvalidBlockHeight(_, _)) => {
```
delta1: TIL about pattern binding

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -320,15 +333,19 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
     /// Check for a chain split with the given peer, validate and add header state. State will be rewound if a higher
     /// proof of work is achieved. If, after this step, the headers are already in sync, true is returned, otherwise a
     /// header sync should proceed to download the remaining headers.
-    async fn check_chain_split(
```
delta1, line 323: The comment above needs to be updated I think

integration_tests/features/support/steps.js
```diff
@@ -162,6 +160,23 @@ Then(/node (.*) is at height (\d+)/, {timeout: 120*1000}, async function (name,
...
+    let tipHash = null;
+    await this.forEachClientAsync(async (client, name) => {
+        await waitFor(async() => client.getTipHeight(), height, 115*1000);
+        const currTip = await client.getTipHeader();
+        expect(currTip.height).to.equal(height);
+        if (!tipHash) {
+            tipHash = currTip.hash.toString('hex');
+            console.log(`Node ${name} is at tip: ${tipHash}`);
+        } else {
+            let curTipHash = currTip.hash.toString('hex');
```
delta1: should probably be `let currTipHash` - I got `ReferenceError: currTipHash is not defined` here

base_layer/core/src/base_node/sync/header_sync/validator.rs
delta1: I don't see a problem with 1000 and agree that fewer allocations are better. If we are arguing performance issues then we need some profiling otherwise we are just speculating.

delta1 (changes requested):
Looking good, I'm still trying to get the large reorg integration test done successfully locally - I had to increase the timeout and make that JS change to curTipHash

base_layer/core/src/base_node/sync/header_sync/validator.rs
sdbondi: I don't think that it's worth putting engineering effort into optimising a potentially wasteful 208kb allocation while in sync mode (the base node only syncs every, say, few minutes at the very worst). We may want to add a temporary LMDB database for this purpose, eliminating the issue entirely but I didn't want to include changes to the blockchain backend when an in-memory variable likely suffices.

integration_tests/features/support/steps.js
sdbondi: Ah thanks - dammit nodejs... 😆

base_layer/core/src/base_node/sync/header_sync/validator.rs
delta1: Agreed.

sdbondi:
@delta1 Fixed the typo and disabled timeouts for the mining tasks in favour of small timeouts for each block mined, making it proportional to the number of blocks requested

delta1:
Nice the integration test passed first time for me - looking good

delta1 (dismissed):
LGTM - tested

PR #2760, opened 2021-03-17
[chore] Removed some unused dependencies

sdbondi:
Removed unused dependencies from basenode, console wallet and core


delta1:
Nice how did you pick these up? `cargo udeps`?

sdbondi:
@delta1 That would have been a good way to do it 🤣  I just spotted some that probably weren't used and then did a scoped search in my editor to see if they were imported.

PR #2761, opened 2021-03-17
[wallet] Differentiate "funds pending" case in UTXO selection

delta1:
## Description

Adds a new error case to the output manager service. When selecting UTXOs and not enough spendable funds are available, check if the pending incoming balance would cover the shortfall. If so, return the new `FundsPending` error instead of `NotEnoughFunds`. This way, the caller can differentiate between the two cases.

## Motivation and Context

Currently when selecting UTXOs for a fee estimate, "not enough funds" is returned even when the pending incoming balance makes it appear that the funds are available.

## How Has This Been Tested?

New assertions added in UTXO selection tests. Existing tests pass. 

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet_ffi/src/error.rs
+++ b/base_layer/wallet_ffi/src/error.rs
@@ -106,6 +106,10 @@ impl From<WalletError> for LibWalletError {
                 code: 101,
                 message: format!("{:?}", w),
             },
+            WalletError::OutputManagerError(OutputManagerError::FundsPending) => Self {
+                code: 115,
+                message: format!("{:?}", w),
+            },
             WalletError::OutputManagerError(OutputManagerError::IncompleteTransaction(_)) => Self {
                 code: 102,
                 message: format!("{:?}", w),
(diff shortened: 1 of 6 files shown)


SWvheerden (approved):
LGTM

base_layer/wallet_ffi/src/error.rs
```diff
@@ -106,6 +106,10 @@ impl From<WalletError> for LibWalletError {
                 code: 101,
                 message: format!("{:?}", w),
             },
+            WalletError::OutputManagerError(OutputManagerError::FundsPending) => Self {
+                code: 115,
```
stringhandler, line 110: This is going to set someone's OCD off

stringhandler (approved):

base_layer/wallet_ffi/src/error.rs
delta1: lol I figured better to keep it next to NotEnoughFunds 🤷‍♂️

PR #2771, opened 2021-03-19
Fix drop RPC client bug in transaction monitoing protocols, improve cucumber

hansieodendaal:
## Motivation and Context
- Fixed RPC connections being kept open; closing RPC connections in the coinbase monitoring and transaction validation protocols.
- Fixed transaction validation protocol to also validate broadcast transactions.
- Fixed broadcast monitoring protocol to stop when transactions are not valid any more.
- Simplified some cucumber test steps, also with more robust measurements.
- Fixed some failing cucumber test scripts

## How Has This Been Tested?
- RPC connections bugfix tested in PR #2734
- Cucumber tests

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
@@ -328,6 +328,9 @@ where TBackend: TransactionBackend + 'static
                                             );
                                             e
                                         });
+                                    // Note: Dropping the client here reduces the number of concurrent RPC connections
+                                    drop(client);
+                                    // Wait out the remainder of the delay before proceeding with next loop
                                     delay.await;
                                     retries += 1;
                                     batches = self.get_transaction_batches().await.map_err(|e| TransactionServiceProtocolError::new(self.id, e))?;
@@ -348,6 +351,9 @@ where TBackend: TransactionBackend + 'static
                                     );
                                     e
                                 });
+                                // Note: Dropping the client here reduces the number of concurrent RPC connections
+                                drop(client);
+                                // Wait out the remainder of the delay before proceeding with next loop
                                 delay.await;
                                 batches.push(batch);
                                 retries += 1;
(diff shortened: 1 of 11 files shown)


StriderDM (dismissed):
LGTM.

Might want to consider removing the commented code in the cucumber tests? Or is that there for future debugging?

hansieodendaal:
> LGTM.
> 
> Might want to consider removing the commented code in the cucumber tests? Or is that there for future debugging?

Hi David, yes, I specifically left it there to make it easier to debug in future, as I do not know of a #Debug directive for Node.js.

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -562,7 +568,9 @@ where TBackend: TransactionBackend + 'static
             .await?
             .values()
             .filter(|tx| {
-                tx.status == TransactionStatus::MinedUnconfirmed || tx.status == TransactionStatus::MinedConfirmed
+                tx.status == TransactionStatus::Broadcast ||
```
philipr-za, line 571: Why did you add the Broadcast here? I don't think you can validate a Broadcast transaction because it is not in the chain? If it is in the Broadcast state it will be checked when a broadcast protocol is started for it.

hansieodendaal: This is to enable the wallet to update its transaction status after a reorg. It may have broadcasted a transaction using a UTXO that has been reorged out, same as the other two included there. In that case the broadcast monitoring protocol must be started so the transaction can be invalidated.

SWvheerden (commented):
It looks like this branch has some other changes in it?
Some of these changes are in PR2775 as well?

hansieodendaal:
> It looks like this branch has some other changes in it?
> Some of these changes are in PR2775 as well?

The only duplicate changes are fixing errors in cucumber scripts I could find so they do not fail in Circle CI

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
philipr-za: But this will invalidate normal transactions that have been broadcast because they are not in the chain yet. You cannot validate a transaction that is not mined.

hansieodendaal:
Thanks for the comments, let us close this one.

Issue #2253, opened 2020-09-22
Syncing after big reorg broken, dependent on size of orphan pool


hansieodendaal:
**Describe the bug**

The reorg strategy is making use of the orphan pool to perform a reorg. Syncing to the stronger chain after a reorg is not possible if the size of the orphan pool (set with `orphan_storage_capacity`) is not large enough to accommodate the size of the chain split. In normal operation, if the orphan pool is filled to capacity, oldest blocks will be discarded from the orphan pool and newer blocks will be added. The orphan pool is used to search for a block with a connection to the chain split, and if that block is not found, the split state is maintained forever.

**To Reproduce**

Steps to reproduce the behaviour:
A chain split that is bigger than a node's orphan pool is needed, where the node's own chain remains the better chain for more blocks than the size of its orphan pool, after which the contesting chain should become the better chain.

**Expected behaviour**

Increasing `orphan_storage_capacity` to a large enough value before the chain spilt starts occurring will solve this problem, but choosing a small value for `orphan_storage_capacity` should not break the blockchain. Syncing after a reorg should always be possible.

**Screenshots**

n/a

**Desktop (please complete the following information):**

Windows, MacOS

**Smartphone (please complete the following information):**

n/a

**Additional context**

Related to #2244 and #2247.

hansieodendaal:
PR #2775 proved that issue can be closed.

PR #2775, opened 2021-03-23
Add massive reorg cucumber test

hansieodendaal:
## Description
- Added two massive reorg tests greater than the  `orphan_storage_capacity`
  - _Massive reorg simple case_
  - _Massive multiple reorg_
- Fixed a bug where all orphans are cleared out every time  block sync occurs, instead of only trimming it to the specified `orphan_storage_capacity`
- Fixed some failing cucumber test scripts

**Note:** Both tests fail for blockchain length of 500+ blocks if only one node is used to connect to the fork chain; other than what is described in the log file, the cause has not been found yet:

``` rust
2021-03-23 06:23:22.354303100 [c::cs::lmdb_db::lmdb_db] ERROR Failed to apply DB transaction: You tried to execute an invalid Database operation: There is a different header stored at height 1 already. New header (99b94cfaf67093c69bb514a6f99f80dba6d175ad97a2613c3aad6b7aae1159b0), current header: (d88e1c1be6581314a32e7bcc5b901699f8de757b395d1daaefefd12371a18a29)
2021-03-23 06:23:22.354467500 [c::bn::header_sync] DEBUG Failed to synchronize headers from peer `2333e6514823aa73efae2c81ae`: Chain storage error: You tried to execute an invalid Database operation: There is a different header stored at height 1 already. New header (99b94cfaf67093c69bb514a6f99f80dba6d175ad97a2613c3aad6b7aae1159b0), current header: (d88e1c1be6581314a32e7bcc5b901699f8de757b395d1daaefefd12371a18a29)
2021-03-23 06:23:22.354497500 [c::bn::header_sync] WARN  Banned sync peer because Failed to synchronize headers from peer: Chain storage error: You tried to execute an invalid Database operation: There is a different header stored at height 1 already. New header (99b94cfaf67093c69bb514a6f99f80dba6d175ad97a2613c3aad6b7aae1159b0), current header: (d88e1c1be6581314a32e7bcc5b901699f8de757b395d1daaefefd12371a18a29)
```

## Motivation and Context
This test was created to #2253

## How Has This Been Tested?
Cucumber test

## Types of changes

* [X] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/integration_tests/features/Reorgs.feature
+++ b/integration_tests/features/Reorgs.feature
@@ -34,9 +34,9 @@ Feature: Reorgs
     And I mine a block on B at height 4 with an invalid MMR
     Then node B is at tip BTip1
 
-@critical @reorg
+@critical @reorg @ignore
   Scenario: Pruned mode reorg past horizon
-    Given I have a base node NODE1 connected to all seed nodes    
+    Given I have a base node NODE1 connected to all seed nodes
     When I mine a block on NODE1 with coinbase CB1
     Given I have a base node NODE2 connected to node NODE1
     When I mine but don't submit a block BLOCK2 on NODE2
@@ -57,7 +57,7 @@ Feature: Reorgs
     When I mine but don't submit a block BLOCK17 on NODE2
     When I mine but don't submit a block BLOCK18 on NODE2
     When I mine but don't submit a block BLOCK19 on NODE2
-    When I mine but don't submit a block BLOCK20 on NODE2   
+    When I mine but don't submit a block BLOCK20 on NODE2
     And I stop NODE2
     When I mine 3 blocks on NODE1
     When I create a transaction TX1 spending CB1 to UTX1
--- a/integration_tests/features/Reorgs.feature
+++ b/integration_tests/features/Reorgs.feature
@@ -91,3 +91,147 @@ Feature: Reorgs
     When I create a transaction TX2 spending CB1 to UTX2
     When I submit transaction TX2 to PNODE1
     Then PNODE1 has TX2 in MEMPOOL state
+
+Scenario Outline: Massive multiple reorg
+        #
+        # Chain 1a:
+        #   Mine X1 blocks (orphan_storage_capacity default set to 10)
+        #
+    Given I have a seed node SEED_A1
+        # Add multiple base nodes to ensure more robust comms
+    And I have a base node NODE_A1 connected to seed SEED_A1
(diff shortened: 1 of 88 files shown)


integration_tests/log4rs_integration_tests.yml
```diff
@@ -102,68 +102,109 @@ appenders:
...
+    appenders:
+      - core
+    additive: false
+  tari::base_node::grpc:
+    level: trace
+    appenders:
+      - core
     additive: false
-  # Route log events sent to the "wallet" logger to the "base_layer" appender
+  # base_layer (mainly wallet)
```
philipr-za: Feels weird to have the wallet be Base Layer but not core? If you are going to the trouble to relabel these various appenders then perhaps we should rename base_layer to wallet?

hansieodendaal: Sorry, my mistake, this file should not have been added to the commit as it is what I sometimes use for testing; will revert to the previous version.

philipr-za (dismissed):

integration_tests/features/Reorgs.feature
```diff
@@ -91,3 +91,147 @@ Feature: Reorgs
...
+    And I connect node NODE_A1 to node NODE_A3 and wait 1 seconds
+    And I connect node NODE_A2 to node NODE_A4 and wait 1 seconds
+    And I connect node SEED_A1 to node SEED_A2 and wait <SYNC_TIME> seconds
+    Then node SEED_A1 is at the same height as node SEED_A2
+    When I mine 10 blocks on SEED_A1
+    Then node SEED_A2 is at the same height as node SEED_A1
+    Then node NODE_A1 is at the same height as node SEED_A1
+    Then node NODE_A2 is at the same height as node SEED_A1
+    Then node NODE_A3 is at the same height as node SEED_A1
+    Then node NODE_A4 is at the same height as node SEED_A1
```
SWvheerden, line 130: Would it not be better to replace this with the function: then all nodes are at the same height as X

SWvheerden (dismissed):
LGTM

integration_tests/features/Reorgs.feature
hansieodendaal: That is an approach I considered, but it (the _all nodes_ part) would not work for chain 2 though if someone changes the number of blocks that are mined, so to keep with the pattern, I added the explicit steps.

PR #2792, opened 2021-03-23
[Mining Node] Added additional command line arguments

StriderDM:
## Description
Added max-blocks, max-difficulty, min-difficulty for use in integration tests
Add constrainedMiningNode to integration tests

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:


integration_tests/helpers/constrainedMiningNodeProcess.js
```diff
@@ -0,0 +1,101 @@
...
+const {expect} = require('chai');
+const {createEnv} = require("./config");
+
+let outputProcess;
+
+class ConstrainedMiningNodeProcess {
+    constructor(name, baseNodeAddress, walletAddress ) {
+        this.name = name;
+        this.maxBlocks = 1;
+        this.minDiff = 10000;
```
stringhandler: 0

integration_tests/helpers/constrainedMiningNodeProcess.js
```diff
@@ -0,0 +1,101 @@
...
+        this.maxDiff = 100000;
+        this.nodeAddress = baseNodeAddress.split(":")[0];
+        this.nodeGrpcPort = baseNodeAddress.split(":")[1];
+        this.walletAddress = walletAddress.split(":")[0];
+        this.walletGrpcPort = walletAddress.split(":")[1];
+    }
+
+    async init(maxBlocks, minDiff, maxDiff) {
+        this.maxBlocks = maxBlocks;
```
stringhandler: Add a default so that these don't have to be specified

```suggestion
        this.maxBlocks = maxBlocks || this.maxBlocks;
```

integration_tests/helpers/constrainedMiningNodeProcess.js
```diff
@@ -0,0 +1,101 @@
...
+                    resolve(ps);
+                }
+            });
+
+            expect(ps.error).to.be.an('undefined');
+            this.ps = ps;
+        });
+    }
+
+    async startNew() {
```
stringhandler: This method requires that init has been called before it. Perhaps you should merge that code in here, or add a call to init here. 

The other nodes require a separate `init` step and that's why they are separate. This process does not require the additional step, so you can maybe put the min/max args directly in here

applications/tari_mining_node/src/main.rs
```diff
@@ -113,6 +122,7 @@ async fn mining_cycle(
     node_conn: &mut BaseNodeClient<Channel>,
     wallet_conn: &mut WalletClient<Channel>,
     config: &MinerConfig,
+    bootstrap: &ConfigBootstrap,
```
stringhandler, line 125: I don't like the reuse of ConfigBootstrap here. It's much better to create a new struct that is specific to this app only

integration_tests/helpers/constrainedMiningNodeProcess.js
StriderDM: The step which determines the number of blocks to be mined is decoupled from the creation of the miner, hence the separate step.

Will add defaults, though think this will still be needed in the tests.

PR #2794, opened 2021-03-23
Implement TariScript base node validation and default scripts in transactions

philipr-za:
## Description
This PR includes the work of @SWvheerden, @hansieodendaal and myself to implement the foundations of TariScript support in the Tari base node and wallets.

The goal of this PR is to add all the data required for TariScripts while building transactions and also to validate the TariScripts by base nodes. 

On the wallet side of things this PR updates the wallet Sql database to store all the elements that the wallet needs to remember to spend an output with a script and updates the Transaction Negotiation Protocols to work with TariScripts. The updates are able to support general TariScripts but in the Wallet the current assumption is that all outputs will have the “Default” Nop script for now.

This PR also adds the fields to the headers required for script validation and the base node validation has been updated to also check the script elements for individual transactions and the accumulated offset for constructed blocks and the blockchain. The PR also contains a new genesis block for the Weatherwax test net which will be the testnet we first run TariScript on. The genesis block is 4000 lines long so this PR is not as huge as it looks.

One last feature that is included in this PR as it was easy to add in during the Transaction Protocol refactor is that the Coin Split function in the Wallet now produces outputs that are rewindable and thus can be recovered by a wallet recovery.

## How Has This Been Tested?
The existing unit and functional tests in the base node and wallet have been updated to take the TariSCript additions into account and should test the transaction negotiation and validation and base node validation.

The coin split rewindable output update was also tested by an updated test in the Output Manager Service.

## Types of changes

* [x] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/db_transaction.rs
+++ b/base_layer/core/src/chain_storage/db_transaction.rs
@@ -120,9 +120,17 @@ impl DbTransaction {
     }
 
     /// Adds a UTXO into the current transaction and update the TXO MMR.
-    pub fn insert_utxo(&mut self, utxo: TransactionOutput, header_hash: HashOutput, mmr_leaf_index: u32) -> &mut Self {
+    pub fn insert_utxo(
+        &mut self,
+        utxo: TransactionOutput,
+        header_hash: HashOutput,
+        header_height: u64,
+        mmr_leaf_index: u32,
+    ) -> &mut Self
+    {
         self.operations.push(WriteOperation::InsertOutput {
             header_hash,
+            header_height,
             output: Box::new(utxo),
             mmr_position: mmr_leaf_index,
         });
@@ -134,11 +142,13 @@ impl DbTransaction {
         output_hash: HashOutput,
         proof_hash: HashOutput,
         header_hash: HashOutput,
+        header_height: u64,
         mmr_leaf_index: u32,
     ) -> &mut Self
     {
         self.operations.push(WriteOperation::InsertPrunedOutput {
             header_hash,
+            header_height,
             output_hash,
             proof_hash,
             mmr_position: mmr_leaf_index,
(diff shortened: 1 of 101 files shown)


base_layer/core/src/chain_storage/db_transaction.rs
```diff
@@ -120,9 +120,17 @@ impl DbTransaction {
     }
 
     /// Adds a UTXO into the current transaction and update the TXO MMR.
-    pub fn insert_utxo(&mut self, utxo: TransactionOutput, header_hash: HashOutput, mmr_leaf_index: u32) -> &mut Self {
+    pub fn insert_utxo(
+        &mut self,
+        utxo: TransactionOutput,
+        header_hash: HashOutput,
+        header_height: u64,
```
stringhandler, line 127: The header height here can be inferred on retrieval. There's no real need to store it

stringhandler (commented):
There are a lot of changes here, so it's difficult to review all of them. One thing that stands out is adding the height to the transaction output in the database. The header hash is already stored in the key of the output so it doesn't really make sense to store the height as well. You could also replace the hash with the height if you really needed to, but it's extra data that takes space for no real use case. When you are retrieving an output, even by itself, you would need to know the header it is in, which means you most likely already know the height

base_layer/core/src/validation/block_validators.rs
```diff
@@ -149,6 +144,60 @@ impl<B: BlockchainBackend> PostOrphanBodyValidation<B> for BodyOnlyValidator {
     }
 }
 
+/// This is a test only validator as it skips the mined height rule of a input.
```
stringhandler, line 147: if this is a test only validator, it should be in a` #[cfg(test)]` block and in it's own file. Otherwise it may get used

PR #2816, opened 2021-03-30
Improve wallet recovery resiliency

hansieodendaal:
## Description
- Resiliency improvements:
  - Added multiple retries connecting to a peer seed node instead of only trying once.
  - Added reconnect logic for every RPC call if the RPC connection drops.
- Improved logging.
- Wallet recovery time improvements:
  - Efficiency improvements with chain metadata RPC calls; a significant overhead was measured if done for every block of UTXOs to stream. This is not done for every block of UTXOs to stream anymore, just once at the start and then close to the end.
  - Efficiency improvements when streaming UTXOs; a significant overhead was measured in fetching start and end headers at the heights in question for every block of UTXOs to stream. Block size was increased 10 fold to reduce the amount of calls fetching headers.

**Issues for a next PR:**
- The recovered amount after recovery is reduced when connecting to a new base node and performing UTXO validation; many recovered UTXOs are invalidated straight away.
- This new issue seen, but not often: `[wallet::recovery] WARN  Error communicating to RPC client (get_header_by_height): Error performing wallet recovery: 'Failed to decode message: failed to decode Protobuf message: BlockHeader.version: invalid wire type: LengthDelimited (expected Varint)'`
- RPC response times seem generally slow; 2.4s average to fetch both the start and end headers for the new block of UTXOs to stream.

## Motivation and Context
Wallet recovery was not resilient in the face of disconnecting base nodes and RPC errors and also slower than it could have been.

## How Has This Been Tested?
Many test runs performing wallet recovery.

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/recovery.rs
+++ b/applications/tari_console_wallet/src/recovery.rs
@@ -25,7 +25,7 @@ use futures::{FutureExt, StreamExt};
 use log::*;
 use rustyline::Editor;
 use tari_app_utilities::utilities::ExitCodes;
-use tari_comms::peer_manager::Peer;
+use tari_comms::types::CommsPublicKey;
 use tari_core::transactions::types::PrivateKey;
 use tari_crypto::tari_utilities::hex::Hex;
 use tari_key_manager::mnemonic::to_secretkey;
@@ -62,8 +62,15 @@ pub fn prompt_private_key_from_seed_words() -> Result<PrivateKey, ExitCodes> {
 /// Recovers wallet funds by connecting to a given base node peer, downloading the transaction outputs stored in the
 /// blockchain, and attempting to rewind them. Any outputs that are successfully rewound are then imported into the
 /// wallet.
-pub async fn wallet_recovery(wallet: &mut WalletSqlite, base_node: &Peer) -> Result<(), ExitCodes> {
-    let mut recovery_task = WalletRecoveryTask::new(wallet.clone(), base_node.public_key.clone());
+pub async fn wallet_recovery(
+    wallet: &mut WalletSqlite,
+    peer_seed_public_keys: Vec<CommsPublicKey>,
+) -> Result<(), ExitCodes>
+{
+    println!("\nPress Ctrl-C to stop the recovery process\n");
+    let mut recovery_task = WalletRecoveryTask::new(wallet.clone(), peer_seed_public_keys);
+    recovery_task.set_connection_retries_limit(250);
+    recovery_task.set_print_to_console(true);
 
     let mut event_stream = recovery_task
         .get_event_receiver()
@@ -81,7 +88,10 @@ pub async fn wallet_recovery(wallet: &mut WalletSqlite, base_node: &Peer) -> Res
                     },
                     WalletRecoveryEvent::Progress(current, total) => {
                         let percentage_progress = ((current as f32) * 100f32 / (total as f32)).round() as u32;
-                        println!("{}: Recovery process {}% complete.", Local::now(), percentage_progress);
+                        println!(
+                            "{}: Recovery process {}% complete ({} of {} blocks).",
+                            Local::now(), percentage_progress, current, total
+                        );
                     },
                     WalletRecoveryEvent::Completed(num_utxos, total_amount) => {
                         println!("Recovered {} outputs with a value of {}", num_utxos, total_amount);
@@ -89,7 +99,9 @@ pub async fn wallet_recovery(wallet: &mut WalletSqlite, base_node: &Peer) -> Res
                 }
             },
             recovery_result = recovery_join_handle => {
-               return recovery_result.map_err(|e| ExitCodes::RecoveryError(format!("{}", e)))?.map_err(|e| ExitCodes::RecoveryError(format!("{}", e)));
+                return recovery_result.map_err(|e| ExitCodes::RecoveryError(format!("{}", e)))?
+                    .map_err(|e| ExitCodes::RecoveryError(format!("{}", e))
+                );
             }
         }
     }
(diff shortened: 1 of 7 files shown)


applications/tari_console_wallet/src/recovery.rs
```diff
@@ -62,8 +62,15 @@ pub fn prompt_private_key_from_seed_words() -> Result<PrivateKey, ExitCodes> {
...
 /// wallet.
-pub async fn wallet_recovery(wallet: &mut WalletSqlite, base_node: &Peer) -> Result<(), ExitCodes> {
-    let mut recovery_task = WalletRecoveryTask::new(wallet.clone(), base_node.public_key.clone());
+pub async fn wallet_recovery(
+    wallet: &mut WalletSqlite,
+    peer_seed_public_keys: Vec<CommsPublicKey>,
+) -> Result<(), ExitCodes>
+{
+    println!("\nPress Ctrl-C to stop the recovery process\n");
+    let mut recovery_task = WalletRecoveryTask::new(wallet.clone(), peer_seed_public_keys.clone());
```
sdbondi:
```suggestion
    let mut recovery_task = WalletRecoveryTask::new(wallet.clone(), peer_seed_public_keys);
```

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -221,6 +221,9 @@ where T: WalletBackend + 'static
 
         let tip_info = client.get_tip_info().await?;
 
+        // Note: RPC connections must be dropped explicitly otherwise they stay open
+        drop(client);
```
sdbondi, line 225: This is not true - if you look at how the `drop` function is implemented you will understand why

base_layer/wallet/src/tasks/wallet_recovery.rs
```diff
@@ -26,49 +26,328 @@ use crate::{
...
+                );
+                if select_new_base_node && self.print_to_console {
+                    println!(
+                        "Trying to connect to {} (retries {} of {})",
+                        new_base_node_public_key.clone(),
+                        self.connection_retries,
+                        self.connection_retries_limit
+                    );
+                }
+                let base_node_node_id = NodeId::from_key(&new_base_node_public_key.clone()).map_err(|e| e)?;
```
sdbondi: Use the infallible `from_public_key` method 
```suggestion
                let base_node_node_id = NodeId::from_public_key(&new_base_node_public_key);
```

sdbondi (commented):
I think there is unnecessary complexity that comes from attempting to retry per RPC call. The flow I'd probably be looking for is 

```
loop {
    let connection = get_next_peer_connection()
    let client = connection.connect_rpc()
    let result = do_work_from_where_we_left_off(client);
   if result.is_an_error() {
      continue; 
   } 
   if result.is_a_success() {
     break;
   }
}
```

But I think we can merge this as is. Extra logging is great!

applications/tari_console_wallet/src/main.rs
```diff
@@ -107,6 +108,11 @@ fn main_inner() -> Result<(), ExitCodes> {
     // get base node/s
     let base_node_config = runtime.block_on(get_base_node_peer_config(&config, &mut wallet))?;
     let base_node = base_node_config.get_base_node_peer()?;
+    let peer_seed_public_keys: Vec<CommsPublicKey> = base_node_config
+        .peer_seeds
+        .iter()
+        .map(|f| f.public_key.clone())
+        .collect();
```
delta1: These nodes are already available in the `base_node_config`, which is a `PeerConfig` already passed in to `start_wallet`. So you can use this logic to get the public keys using that where you need it 

image

hansieodendaal: Good observation! Saving one level of functions here then.

base_layer/wallet/src/base_node_service/service.rs
hansieodendaal: Changing the comment to: `// Note: Dropping the client here reduces the number of concurrent RPC connections`. 
Happy? With stress testing in cucumber this made a definite improvement.

sdbondi: I ran the cucumber tests, however the `Stress` tests are not currently working. 

Given:
1. There is nothing special about `drop(T)`
2. The client is dropped implicitly (exactly the same as `drop(T)`) at any return point of the current function

Why and how are the claims true? 

I'm concerned about spreading misinformation/magic around how the client works, or future developers thinking that an explicit drop is something they must do

```rust
fn my_fn() {
  let client = NewClient
  client.do_something();
  // Reduces the number of concurrent connections, or the connection stays open (leaks) unless we explicitly drop 
drop(client);
}
```

The comments are clearly untrue.

I suggest either removing the drop or `// Drop the client as early as possible` (without giving a reason why), or ignore this comment if you feel strongly about it.

sdbondi (approved):
Did a recovery and it worked - LGTM

base_layer/wallet/src/base_node_service/service.rs
hansieodendaal: Seems worthwhile to get to the bottom of this, as #2771 does a similar thing. 

I am repeating a pattern that was introduced into the broadcast monitoring protocols by @philipr-za in #2719 when cucumber stress tests were developed and system-level transaction performance improved, and subsequent to that, @mikethetike requested clarification about the drop statement in #2734. 

What I recall when reviewing #2719:
- I was able to complete the `10,000` sized stress test when upping the RPC concurrency limit to `10,000` from `100`.
- Philip was able to complete the same test when upping the RPC concurrency limit to `1,000` from `100` and adding `drop(client);` to the broadcast monitoring protocol.
- The reason why `drop(client);` improved system-level transaction performance was not clear as the `client` should be destroyed when it goes out of scope.

Seeing that we will have visibility into concurrent RPC connections when #2830 is merged, maybe we can remove the `drop(client);` altogether from the transaction and base node service protocols?

PR #2818, opened 2021-03-31
get_transaction_info no longer errors if any transaction is not found

sdbondi:
Changes get_transaction_info to mark an item as found/not_found using a
boolean flag, rather than returning an error.


philipr-za (approved):
LGTM

PR #2820, opened 2021-03-31
[chore] Use prettier for integration tests

sdbondi:
Standard formatting (almost exactly like `rust-fmt`) for js


delta1 (commented):
Nice desperately needed. We also need the check done on commit and on CI but that can be additional. Maybe an update to the readme to indicate that `npm run fmt` should be run before commit and/or prettier integration used in IDE on save?

integration_tests/.prettierrc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+trailingComma: "all"
+tabWidth: 4
+semi: true
+singleQuote: true
+printWidth: 150
```
delta1: I would much rather use the default options === nothing to bikeshed

integration_tests/features/support/world.js
```diff
@@ -1,8 +1,8 @@
 const { setWorldConstructor, After, BeforeAll } = require("cucumber");
 
-const BaseNodeProcess = require('../../helpers/baseNodeProcess');
-const MergeMiningProxyProcess = require('../../helpers/mergeMiningProxyProcess');
-const WalletProcess = require('../../helpers/walletProcess');
+const BaseNodeProcess = require("../../helpers/baseNodeProcess");
+const MergeMiningProxyProcess = require("../../helpers/mergeMiningProxyProcess");
+const WalletProcess = require("../../helpers/walletProcess");
```
delta1: sigh javascript, if `singleQuote` is true then why are these double quotes?

integration_tests/package.json
```diff
@@ -16,7 +18,8 @@
     "cucumber-pretty": "^6.0.0",
     "grpc": "^1.24.3",
     "grpc-promise": "^1.4.0",
-    "mocha": "^8.1.3"
+    "mocha": "^8.1.3",
```
delta1: `mocha` can be removed now afaik it's unused

sdbondi:
Good ideas - done

integration_tests/.prettierrc.yaml
sdbondi: Haha yeah I agree

delta1:
> Good ideas - done

sorry about the timing of these comments

integration_tests/features/support/world.js
sdbondi: I want to blame js (really) but I think this one was me

delta1: I don't know, I ran the check and fmt locally and it stayed the same lol

sdbondi: Shakes head at JS... typical

PR #2829, opened 2021-04-05
Add database concurrency check

stringhandler:
## Description

Add database concurrency check to make sure that another thread hasn't altered the database since we had read the data. 
Specifically we check that we can:
- only delete the latest header
- only insert a header at the tip
- only insert a block body if the header has been inserted
- only delete a header if the block body has been deleted 

## Motivation and Context

If you mine and submit a block to a node while it is syncing headers, it would replace the block and header that was there already, and then continue inserting data. This can lead to an incorrect chain of `header`->`previous header`, which I believe leads to the target difficulty being incorrect

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1184,10 +1184,11 @@ fn insert_block(txn: &mut DbTransaction, block: Arc<ChainBlock>) -> Result<(), C
         txn.insert_monero_seed_height(&monero_seed, block.block.header.height);
     }
 
-    // Update metadata
+    let height = block.height();
     let accumulated_difficulty = block.accumulated_data.total_accumulated_difficulty;
-    txn.set_best_block(block.block.header.height, block_hash, accumulated_difficulty)
-        .insert_block(block);
+    txn.insert_header(block.block.header.clone(), block.accumulated_data.clone())
+        .insert_block_body(block)
+        .set_best_block(height, block_hash, accumulated_difficulty);
 
     Ok(())
 }
@@ -1459,7 +1460,7 @@ fn rewind_to_height<T: BlockchainBackend>(
             ))
         })?;
 
-    debug!(
+    info!(
         target: LOG_TARGET,
         "Rewinding headers from height {} to {}",
         last_header_height,
@@ -1480,7 +1481,7 @@ fn rewind_to_height<T: BlockchainBackend>(
     }
 
     let mut removed_blocks = Vec::with_capacity(steps_back as usize);
-    debug!(
+    info!(
         target: LOG_TARGET,
         "Rewinding blocks from height {} to {}",
         last_block_height,
@@ -1500,7 +1501,7 @@ fn rewind_to_height<T: BlockchainBackend>(
     let (last_header, header_accumulated_data) = db.fetch_header_and_accumulated_data(height)?;
 
     for h in 0..steps_back {
-        debug!(target: LOG_TARGET, "Deleting block {}", last_block_height - h,);
+        info!(target: LOG_TARGET, "Deleting block {}", last_block_height - h,);
         let block = fetch_block(db, last_block_height - h)?;
         let block = Arc::new(block.clone().try_into_chain_block()?);
         txn.delete_block(block.block.hash());
@@ -1735,7 +1736,7 @@ fn reorganize_chain<T: BlockchainBackend>(
     for block in chain {
         let mut txn = DbTransaction::new();
         let block_hash_hex = block.accumulated_data.hash.to_hex();
-        txn.delete(DbKey::OrphanBlock(block.accumulated_data.hash.clone()));
+        txn.delete_orphan(block.accumulated_data.hash.clone());
         if let Err(e) = block_validator.validate_body_for_valid_orphan(&block, backend) {
             warn!(
                 target: LOG_TARGET,
@@ -1789,7 +1790,7 @@ fn restore_reorged_chain<T: BlockchainBackend>(
     // Add removed blocks in the reverse order that they were removed
     // See: https://github.com/tari-project/tari/issues/2182
     for block in previous_chain.into_iter().rev() {
-        txn.delete(DbKey::OrphanBlock(block.accumulated_data.hash.clone()));
+        txn.delete_orphan(block.accumulated_data.hash.clone());
         insert_block(&mut txn, block)?;
     }
     db.write(txn)?;
@@ -1923,7 +1924,7 @@ fn find_orphan_descendant_tips_of<T: BlockchainBackend>(
 // Discard the the orphan block from the orphan pool that corresponds to the provided block hash.
 fn remove_orphan<T: BlockchainBackend>(db: &mut T, hash: HashOutput) -> Result<(), ChainStorageError> {
     let mut txn = DbTransaction::new();
-    txn.delete(DbKey::OrphanBlock(hash));
+    txn.delete_orphan(hash);
     db.write(txn)
 }
 
--- a/base_layer/core/src/chain_storage/db_transaction.rs
+++ b/base_layer/core/src/chain_storage/db_transaction.rs
@@ -70,19 +70,14 @@ impl DbTransaction {
         DbTransaction::default()
     }
 
-    /// A general insert request. There are convenience functions for specific delete queries.
-    pub fn delete(&mut self, delete: DbKey) -> &mut Self {
-        self.operations.push(WriteOperation::Delete(delete));
-        self
-    }
-
     pub fn delete_orphan(&mut self, hash: HashOutput) -> &mut Self {
-        self.delete(DbKey::OrphanBlock(hash))
+        self.operations.push(WriteOperation::DeleteOrphan(hash));
+        self
     }
 
     /// Delete a block header at the given height
     pub fn delete_header(&mut self, height: u64) -> &mut Self {
-        self.operations.push(WriteOperation::Delete(DbKey::BlockHeader(height)));
+        self.operations.push(WriteOperation::DeleteHeader(height));
         self
     }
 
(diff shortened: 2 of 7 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1500,7 +1500,7 @@ fn rewind_to_height<T: BlockchainBackend>(
     let (last_header, header_accumulated_data) = db.fetch_header_and_accumulated_data(height)?;
 
     for h in 0..steps_back {
-        debug!(target: LOG_TARGET, "Deleting block {}", last_block_height - h,);
+        info!(target: LOG_TARGET, "Deleting block {}", last_block_height - h,);
```
SWvheerden, line 1504: should this not be debug? We already notified via info that we will be deleting these headers. We don't want to spam info?

base_layer/core/src/chain_storage/db_transaction.rs
```diff
@@ -289,7 +284,8 @@ pub enum WriteOperation {
         proof_hash: HashOutput,
         mmr_position: u32,
     },
-    Delete(DbKey),
+    DeleteHeader(u64),
```
SWvheerden, line 287: I know this might be coming from the old way of deleting.
But is this not a dangerous delete instruction? 

Deleting header at height 10, we don't if height 10 is still the correct header? If the header changed we might delete the wrong header as delete header 10 is still a valid header instruction. 
```suggestion
    DeleteHeader(HashOutput),
```
Should be "safe" and checkable

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -172,18 +172,10 @@ impl LMDBDatabase {
             trace!(target: LOG_TARGET, "[apply_db_transaction] WriteOperation: {}", op);
             match op {
                 InsertOrphanBlock(block) => self.insert_orphan_block(&write_txn, &block)?,
-                Delete(delete) => self.op_delete(&write_txn, delete)?,
                 InsertHeader { header } => {
-                    let height = header.header.height;
-                    if !self.insert_header(&write_txn, &header.header, &header.accumulated_data)? {
```
SWvheerden, line 178: Is there a reason we removed this check?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -667,70 +704,76 @@ impl LMDBDatabase {
...
     }
 
-    fn op_delete(&mut self, txn: &WriteTransaction<'_>, key: DbKey) -> Result<(), ChainStorageError> {
-        match key {
-            DbKey::BlockHeader(k) => {
-                let val: Option<BlockHeader> = lmdb_get(txn, &self.headers_db, &k)?;
-                if let Some(v) = val {
-                    let hash = v.hash();
-                    // Check that there are no utxos or kernels linked to this.
+    fn delete_header(&mut self, txn: &WriteTransaction<'_>, height: u64) -> Result<(), ChainStorageError> {
```
SWvheerden, line 709:
```suggestion
    fn delete_header(&mut self, txn: &WriteTransaction<'_>, hash: Vec<u8>) -> Result<(), ChainStorageError> {
```

This will make sure we delete the exact header we want, and not the header at the exact height

SWvheerden (commented):
Looks good so far, I think we can improve this a bit more

base_layer/core/src/chain_storage/blockchain_database.rs
stringhandler: I think this is an important enough even that the user would want to know about

base_layer/core/src/chain_storage/db_transaction.rs
stringhandler: Yeah, I looked at it and would prefer to use the hash, but the reorgs don't read out the headers at the moment. This would be a slightly more involved implementation and would involve a read at each step. I think there may be a way of doing it, but I'd have to look deeper into it. 

Out of interest, I originally put in the hash and went a fair way down the line implementing before reverting it. I may change it to use a hash in a followup pr

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
stringhandler: Yes, the check has been moved into the method and will return the error there. The method now returns `Result<(),...>` instead of `Result<bool, ...>` and errors when an existing header exists

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -629,22 +628,60 @@ impl LMDBDatabase {
         txn: &WriteTransaction<'_>,
         header: &BlockHeader,
         accum_data: &BlockHeaderAccumulatedData,
-    ) -> Result<bool, ChainStorageError>
+    ) -> Result<(), ChainStorageError>
```
sdbondi, line 629: Comment above needs to be updated/removed

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -629,22 +628,60 @@ impl LMDBDatabase {
...
                 return Err(ChainStorageError::InvalidOperation(format!(
                     "There is a different header stored at height {} already. New header ({}), current header: ({})",
                     header.height,
                     hash.to_hex(),
                     accum_data.hash.to_hex()
                 )));
             }
-            return Ok(false);
+            return Err(ChainStorageError::InvalidOperation(format!(
+                "This header aready has been created. Header: {}",
```
sdbondi: nit: Something like this may be better
```suggestion
                "The header at height {} already exists. Existing header hash: {}",
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -667,70 +704,76 @@ impl LMDBDatabase {
...
+        // sufficient
+
+        let hash = header.hash();
+        // Check that there are no utxos or kernels linked to this.
+
+        if !lmdb_fetch_keys_starting_with::<TransactionKernelRowData>(hash.to_hex().as_str(), &txn, &self.kernels_db)?
+            .is_empty()
+        {
+            return Err(ChainStorageError::InvalidOperation(
+                "Cannot delete header because there are kernels linked to it".to_string(),
```
sdbondi: nit: more descriptive error messages

```suggestion
                format!("Cannot delete header {} ({}) because there are kernels linked to it", header.height, hash.to_hex())
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -667,70 +704,76 @@ impl LMDBDatabase {
...
+            ));
+        }
+        if !lmdb_fetch_keys_starting_with::<TransactionOutputRowData>(hash.to_hex().as_str(), &txn, &self.utxos_db)?
+            .is_empty()
+        {
+            return Err(ChainStorageError::InvalidOperation(
+                "Cannot delete header because there are utxos linked to it".to_string(),
```
sdbondi: same nit suggestion as above

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -667,70 +704,76 @@ impl LMDBDatabase {
...
+        lmdb_delete(&txn, &self.headers_db, &height)?;
+        lmdb_delete(&txn, &self.header_accumulated_data_db, &height)?;
+        lmdb_delete(&txn, &self.kernel_mmr_size_index, &header.kernel_mmr_size.to_be_bytes())?;
+        lmdb_delete(&txn, &self.output_mmr_size_index, &header.output_mmr_size.to_be_bytes())?;
+
+        Ok(())
+    }
+
+    fn delete_orphan(&mut self, txn: &WriteTransaction<'_>, hash: HashOutput) -> Result<(), ChainStorageError> {
+        if let Some(orphan) = lmdb_get::<_, Block>(&txn, &self.orphans_db, hash.as_slice())? {
```
sdbondi, line 763: Any reason we are less strict about orphans? Perhaps we should error when trying to delete an orphan that does not exist.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -751,6 +794,21 @@ impl LMDBDatabase {
             body.to_counts_string()
         );
 
+        // Check that the database has not been changed by another thread
+        // 1. The header we are inserting for matches the header at that height
+        if let Some(current_header_at_height) = lmdb_get::<_, BlockHeader>(txn, &self.headers_db, &header.height)? {
```
sdbondi: It should also be invalid to insert a block body if the header does not exist

sdbondi (commented):
Agree that these checks belong in the "model layer" and will make any data consistency bugs much more obvious. Few clippy issues to fix. Also echo what @SWvheerden mentioned about using unambiguous hashes instead of heights, for a future PR.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
stringhandler: Added, thanks

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
stringhandler: I started writing the check but then thought about it. There is another process that cleans up orphans and it would be difficult to synchronise these in code. The intention here is to prevent mangling of data by different threads resulting in an inconsistent state. I don't think that can happen in this case

comms/dht/src/config.rs
```diff
@@ -161,7 +161,7 @@ impl Default for DhtConfig {
             saf_auto_request: true,
             saf_max_message_size: 512 * 1024,
             saf_minimum_request_period: Duration::from_secs(3 * 24 * 60 * 60), // 3 days
-            msg_hash_cache_capacity: 100_000,
+            msg_hash_cache_capacity: 1,
```
sdbondi: Think this should be reverted in this PR

sdbondi (approved):
LGTM

PR #2837, opened 2021-04-09
Clarify one sided payment rules

SWvheerden:
## Description

RFC 201, for tari script, laid out a way to do one-sided payments. This RFC clarifies the calculation of the blinding factor k_b to ensure there is nothing left over to think about by explicitly typing out how to calculate it using Diffie-hellman. 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -498,6 +498,7 @@ $$
 $$
 
 Next Alice next uses Bulletproof rewinding to encrypt the value \\( v_b \\) into the the Bulletproof for the commitment \\( C_b \\). For this she uses (\\( k_{rewind} =  Hash(k_{b}) \\) as the rewind_key and (\\( k_{blinding} =  Hash(Hash(k_{b})) \\) as the blinding key.
+*Note, deriving the keys here should be secure, but should be confirmed before mainnet.
 
 Alice knows the script-redeeming private key \\( k_{Sa}\\) for the transaction input.
 
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -482,10 +482,22 @@ Once again, transaction fees are ignored to simplify the illustration.
 Alice owns \\( C_a \\) and provides the required script to spend the UTXO as was described in the previous cases.
 
 Alice needs a public key from Bob, \\( K_{Sb} \\) to complete the one-sided transaction. This key can be obtained
-out-of-band, and will typically be Bob's wallet public key on the Tari network.
+out-of-band, and might typically be Bob's wallet public key on the Tari network.
+
+Bob requires the value \\( v_b \\) and blinding factor \\( k_b \\) to claim his payment, but he needs to be able to claim it without asking Alice for them.
+
+This information can be obtained by using Diffie-Hellman and Bulletproof rewinding. If the blinding factor \\( k_b \\) was calculated with Diffie-Hellman using the offset public keypair, (\\( k_{Ob} \\),\\( K_{Ob} \\)) as sender keypair and
+the keypair, (\\( k_{Sb} \\),\\( K_{Sb} \\)) as the receiver keypair, the blinding factor \\( k_b \\) can be securely calculated without communication.
 
 Alice uses Bob's public key to create a shared secret, \\( k_b \\) for the output commitment, \\( C_b \\), using
 Diffie-Hellman key exchange.
+
+Alice calculates \\( k_b \\) as
+$$
+    k_b = k_{Ob} * {K_Sb}
+$$
+
+Next Alice next uses Bulletproof rewinding to encrypt the value \\( v_b \\) into the the Bulletproof for the commitment \\( C_b \\). For this she uses (\\( k_{rewind} =  Hash(k_{b}) \\) as the rewind_key and (\\( k_{blinding} =  Hash(Hash(k_{b})) \\) as the blinding key.
 
 Alice knows the script-redeeming private key \\( k_{Sa}\\) for the transaction input.
 
@@ -495,7 +507,6 @@ $$
     \so = k_{Sa} - k_{Ob} \cdot \HU_b
 $$
 
-
 For the script hash, she provides the hash of a script that locks the output to Bob's public key, `PushPubkey(K_Sb)`.
 This script will only resolve successfully if the spender can provide a valid signature as input that demonstrates proof
 of knowledge of \\( k_{Sb} \\) which only Bob knows.
@@ -503,12 +514,20 @@ of knowledge of \\( k_{Sb} \\) which only Bob knows.
 Any base node can now verify that the transaction is complete, verify the signature on the script, and verify the script
 offset.
 
-For Bob to claim his commitment, \\( C_b \\) he requires the blinding factor, \\( k_b \\), and his private key for the script.
-
-In this case, the script hash is analogous to an address in Bitcoin or Monero. Bob's wallet can scan the blockchain
+For Bob to claim his commitment he will scan the blockchain for a known script hash because he knowns that the script will be `PushPubkey(K_Sb)` he can scan for that hash. In this case, the script hash is analogous to an address in Bitcoin or Monero. Bob's wallet can scan the blockchain
 looking for hashes that he would know how to resolve. For all outputs that he discovers this way, Bob would need to know
-who the sender is so that he can derive the shared secret.  This information can be obtained without needing to communicate
-with the sender, if she uses the offset public key, \\( K_{Ob} \\) as the key in deriving the Diffie-Hellman exchange.
+who the sender is so that he can derive the shared secret.
+
+When Bob's wallet spots a known hash he requires he requires the blinding factor, \\( k_b \\) and the value \\( v_b \\). First he uses Diffie-Hellman to calculate \\( k_b \\). 
+
+Bob calculates \\( k_b \\) as
+$$
+    k_b = K_{Ob} * {k_Sb}
+$$
+
+Next Bob's wallet calculates \\( k_{rewind} \\), using \\( k_{rewind} =  Hash(k_{b})\\) and (\\( k_{blinding} =  Hash(Hash(k_{b})) \\), using those to rewind the Bulletproof to get the value \\( v_b \\). 
+
+Because Bob's wallet already knowns \\( k_Sb \\), he now knows all the values required to spend the commitment \\( C_b \\)
 
 For Bob's part, when he discovers one-sided payments to himself, he should spend them to new outputs using a traditional
 transaction to thwart any potential horizon attacks in the future.


RFC/src/RFC-0201_TariScript.md
```diff
@@ -507,8 +506,10 @@ For Bob to claim his commitment, \\( C_b \\) he requires the blinding factor, \\
 
 In this case, the script hash is analogous to an address in Bitcoin or Monero. Bob's wallet can scan the blockchain
 looking for hashes that he would know how to resolve. For all outputs that he discovers this way, Bob would need to know
-who the sender is so that he can derive the shared secret.  This information can be obtained without needing to communicate
-with the sender, if she uses the offset public key, \\( K_{Ob} \\) as the key in deriving the Diffie-Hellman exchange.
+who the sender is so that he can derive the shared secret.  
+
+This information can be obtained without needing to communicate with the sender. If the blinding factor \\( k_b \\) was calculated with diffie-hellman using the offset public keypair, (\\( k_{Ob} \\),\\( K_{Ob} \\)) as sender keypair and
+the keypair, (\\( k_{Sb} \\),\\( K_{Sb} \\)) as the receiver keypair, the blinding factor \\( k_b \\) can be securely calculated without communication between the 2 parties.
```
sdbondi: Not super necessary, but might be clearer to be 100% specific. 
i.e.
The sender calculates k_b = k_Ob * K_Sb, and the receiver calculates k_b = K_Ob * k_Sb

sdbondi (dismissed):
LGTM

RFC/src/RFC-0201_TariScript.md
hansieodendaal: Agreed

stringhandler: How does the receiver know the value though?

hansieodendaal: The value must be sent out of band.

stringhandler: if the value is sent out of band, why not send the whole commitment out of band? Isn't the purpose here to not do out of band? 
"This information can be obtained with need to communicate with the sender"

stringhandler (changes requested):
This doesn't seem to solve the problem of not communicating with the sender

hansieodendaal (commented):
Clarification of the Diffie-Hellman exchange is great.

**Discussion:**
Maybe we should add a short discussion about how the one-sided payment receiver can expect payments, but not exactly how much it would be, without communicating out of band with the sender. If the value of the payment is not known, the commitment cannot be opened without huge amounts of computing power.

One option is to know who will pay you and to provide them with a short identification string for one-sided payments, for example, an account number, that would be embedded as the proof message in the Bulletproof range proof. 
- Sender(s) knows
  - Receiver public wallet key
  - Tari one-sided payment script variants that would be accepted by the receiver
  - Unique sender proof message, e.g. an account number
- Receiver knows
  - Sender(s) public rewind key(s)
  - Sender(s) public wallet key(s)
  - Unique sender proof message

The receiver can then scan the blockchain for payments and open the commitments using the following method:
- Scan all UTXOs for all acceptable payment script hashes; there would only be a few at most, and if the script resolves when inserting the receiver public key, it is for the receiver.
- With those identified UTXOs, do Bulletproof rewinding with the range of registered sender public rewind keys. If the proof message matches, the rewound value is more than likely the correct value of the commitment.
- Test opening of the commitment:
  - Use the Diffie-Hellman exchange with corresponding public wallet key to derive the commitment's spending key and the value obtained from Bulletproof rewinding
  - If the commitment can be opened, it comes from the identified sender

stringhandler:
@hansieodendaal seems ok, but if the sender rewind keys are public, can't anyone just read all the values? Why not just include the value in metadata? Or maybe encrypt the value and then add it to the metadata?

hansieodendaal:
The sender's public rewind key is not common knowledge; it has to be shared out of band, and will only be shared with specific parties. The sender can also at any time change the rewind key or use more than one.

Although additional information can be added as metadata, we get up to 23 bytes worth of data for free that can be embedded in the Bulletproof rewind proof message. 

As you suggested, adding the value as encrypted metadata would also be feasible because the opening of the commitment can be tested with the decrypted value and all registered one-sided payment sender public keys. If the correct value was encrypted, the commitment can be opened.

SWvheerden:
Would using the blinding factor of the utxo or a derived key of that for the rewind key not solve the problem?

sdbondi:
I'm not very familiar with the RP rewinding maths but would it not be possible to add the value to the range proof using the Diffie-Hellman key `k_b`  or perhaps `Hash(k_b)` as the rewind blinding key? As part of scanning you would need to derive the DH key for each UTXO and attempt to rewind. 

This makes the problem of obtaining the value the same problem as "how do we efficiently/effectively scan for new one-sided payments?".

sdbondi:
@SWvheerden Beat me to it 😆

hansieodendaal:
> Would using the blinding factor of the utxo or a derived key of that for the rewind key not solve the problem?

This is a very good idea yes (@SWvheerden and @sdbondi) [using `Hash(k_b)`], as the sender will not want to share their own private rewind key (`pvt_rewind_key`). In this way, the sender could still use their own private blinding key (`pvt_blinding_key`) and keep it secret.
(See [Bulletproof rewinding](https://github.com/dalek-cryptography/bulletproofs/issues/335) for context)

SWvheerden:
> > ould using the blinding factor of the utxo or a derived key of that for the rewind key not solve the problem?

RFC/src/RFC-0201_TariScript.md
```diff
@@ -524,6 +524,10 @@ $$
 For Bob's part, when he discovers one-sided payments to himself, he should spend them to new outputs using a traditional
 transaction to thwart any potential horizon attacks in the future.
 
+One other problem with one sided communication is the communication of the value of the transaction being sent, and the wallet recovery of one sided payments. We can solve both of these by using (\\( k_{Ob} \\).
+Because this is a secret key known by Alice and Bob and no one else. We hash this key as a new key, (\\( k_{rewind} =  Hash(k_{Ob}) \\). We know have a rewinding secret key for use in bulletproof rewinding. And now
```
hansieodendaal: We ~~know~~ _now_ have ...

RFC/src/RFC-0201_TariScript.md
```diff
@@ -524,6 +524,10 @@ $$
...
+we can encrypt the value (\\( v \\) inside of the bulletproof rewinding, allowing Bob to unlock his payment without revealing any information publicly.
```
hansieodendaal, line 529: ~~we can encrypt the value (\\( v \\) inside of the bulletproof rewinding~~
we use the standard Bulletproof rewinding scheme to encrypt the value (\\( v \\) inside the Bulletproof range proof,

hansieodendaal:
```suggestion
we use the standard Bulletproof rewinding scheme to encrypt the value (\( v \) inside the Bulletproof range proof,, allowing Bob to unlock his payment without revealing any information publicly. (See [Bulletproof rewinding](https://github.com/dalek-cryptography/bulletproofs/issues/335) for context)).
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -524,6 +524,10 @@ $$
...
+
```
hansieodendaal: I would add something like this:

-----------

In summary, receiving party scanning the blockchain for one-sided payments therefore involves:
- Scan all UTXOs trying to unlock the payment script hashes; if the script resolves when inserting the receiver public key, it is for the receiver.
- With those identified UTXOs, do Bulletproof rewinding with the range of registered sender public rewind keys ((\( k_{rewind}G \) as per above). If the proof message matches, the rewound value is more than likely the correct value of the commitment.
- Test opening of the commitment using the value obtained from rewinding and the spending key from the Diffie-Hellman exchange; if the commitment can be opened, it is intended for the receiver.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -507,12 +506,28 @@ For Bob to claim his commitment, \\( C_b \\) he requires the blinding factor, \\
...
+Bob calculates \\( k_b \\) as
+$$
+    k_b = K_{Ob} * {k_Sb}
+$$
 
 For Bob's part, when he discovers one-sided payments to himself, he should spend them to new outputs using a traditional
 transaction to thwart any potential horizon attacks in the future.
 
+One other problem with one sided communication is the communication of the value of the transaction being sent, and the wallet recovery of one sided payments. We can solve both of these by using (\\( k_{Ob} \\).
```
stringhandler:
```suggestion
One other problem with one sided communication is the communication of the value of the transaction being sent, and the wallet recovery of one sided payments. We can solve both of these by using (\\( k_b \\).
```

RFC/src/RFC-0201_TariScript.md
stringhandler:
```suggestion
Because this is a secret key known by Alice and Bob and no one else. We hash this key as a new key, (\\( k_{rewind} =  Hash(k_b) \\). We know have a rewinding secret key for use in bulletproof rewinding. And now
```

RFC/src/RFC-0201_TariScript.md
stringhandler: Should be "We now use the standard...". Don't need to start a sentence with "And"

RFC/src/RFC-0201_TariScript.md
stringhandler: Don't agree with point 2. Point 2 should be rewound using the hash of the blinding factor

RFC/src/RFC-0201_TariScript.md
SWvheerden: We should be using k_{Ob}, the offset pubkey. k_{b} is the blinding factor.

RFC/src/RFC-0201_TariScript.md
SWvheerden: Do we have a link explaining it in our own repo, not Dalek's?

RFC/src/RFC-0201_TariScript.md
stringhandler: Bob does not know `k_{Ob}` though?

SWvheerden: Yes, He only knows `K_{Ob}` and `k_S`

stringhandler: The wording is ambiguous, but the next line is correct, so will let it pass

RFC/src/RFC-0201_TariScript.md
```diff
@@ -482,11 +482,23 @@ Once again, transaction fees are ignored to simplify the illustration.
 Alice owns \\( C_a \\) and provides the required script to spend the UTXO as was described in the previous cases.
 
 Alice needs a public key from Bob, \\( K_{Sb} \\) to complete the one-sided transaction. This key can be obtained
-out-of-band, and will typically be Bob's wallet public key on the Tari network.
+out-of-band, and might typically be Bob's wallet public key on the Tari network.
+
+Bob requires the value \\( v_b \\) and blinding factor \\( k_b \\) to claim his payment, but he needs to be able to claim it without asking Alice for them.
+
+This information can be obtained by using diffie-hellman and bulletproof rewinding. If the blinding factor \\( k_b \\) was calculated with diffie-hellman using the offset public keypair, (\\( k_{Ob} \\),\\( K_{Ob} \\)) as sender keypair and
```
stringhandler:
```suggestion
This information can be obtained by using Diffie-Hellman and bulletproof rewinding. If the blinding factor \\( k_b \\) was calculated with diffie-hellman using the offset public keypair, (\\( k_{Ob} \\),\\( K_{Ob} \\)) as sender keypair and
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -482,11 +482,23 @@ Once again, transaction fees are ignored to simplify the illustration.
...
 
 Alice uses Bob's public key to create a shared secret, \\( k_b \\) for the output commitment, \\( C_b \\), using
 Diffie-Hellman key exchange.
 
+Alice calculates \\( k_b \\) as
+$$
+    k_b = k_{Ob} * {K_Sb}
+$$
+
+Next alice next uses bulletproof rewinding to encrypt the value \\( v_b \\) into the the bulletproof for the commitment \\( C_b \\). For this she uses (\\( k_{rewind} =  Hash(k_{Ob}) \\) as the encryption key.
```
stringhandler:
```suggestion
Next Alice next uses bulletproof rewinding to encrypt the value \\( v_b \\) into the the bulletproof for the commitment \\( C_b \\). For this she uses (\\( k_{rewind} =  Hash(k_{Ob}) \\) as the encryption key.
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -495,20 +507,27 @@ $$
...
+who the sender is so that he can derive the shared secret.
+
+When Bob's wallet spots a known hash he requires he requires the blinding factor, \\( k_b \\) and the value \\( v_b \\). First he uses Diffie-Hellman to calculate \\( k_b \\). 
+
+Bob calculates \\( k_b \\) as
+$$
+    k_b = K_{Ob} * {k_Sb}
+$$
+
+Next Bob's wallet calculates \\( k_{rewind} \\), using \\( k_{rewind} =  Hash(k_{Ob})\\) and uses that to rewind the bulletproof to get the value \\( v_b \\).
```
stringhandler: Where does Bob find out `k_{0}b` in order to hash it?

stringhandler (commented):
Seems like this refactor has reverted to some previous errors

RFC/src/RFC-0201_TariScript.md
hansieodendaal: **Edit: ** _Strike that_ ~~We equate the private rewind key to the hash of the blinding factor. Bulletproof rewinding is done with the public rewind key, not the private rewind key.~~

Please see the explanation below.

RFC/src/RFC-0201_TariScript.md
hansieodendaal: We can always update it later when we have documented it ourselves.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -495,20 +507,27 @@ $$
...
+who the sender is so that he can derive the shared secret.
+
+When Bob's wallet spots a known hash he requires he requires the blinding factor, \\( k_b \\) and the value \\( v_b \\). First he uses Diffie-Hellman to calculate \\( k_b \\). 
+
+Bob calculates \\( k_b \\) as
+$$
+    k_b = K_{Ob} * {k_Sb}
+$$
+
+Next Bob's wallet calculates \\( k_{rewind} \\), using \\( k_{rewind} =  Hash(k_{b})\\) and uses that to rewind the Bulletproof to get the value \\( v_b \\).
```
hansieodendaal: **Edit:** _Strike that_ ~~We equate the private rewind key to the hash of the blinding factor. Bulletproof rewinding is done with the public rewind key, not the private rewind key.~~

hansieodendaal: We actually need to convey two values, not only one. To be clear on the previous comment, Bulletproof rewinding is done with the two public rewind keys, not the private rewind keys.

From [Tari fork of Bulletproofs]( https://github.com/tari-project/bulletproofs/blob/a18a3d2c91cea4e5f999446a566a4b0af7c9e187/tests/range_proof.rs#L154)
``` rust
    // The two rewind keys can be shared with a trusted 3rd party, whom will be able to extract
    // the value of the commitment as well as the extra data
    let pub_rewind_key_1 =
        RistrettoPoint::from(&pvt_rewind_key * &RISTRETTO_BASEPOINT_TABLE).compress();
    let pub_rewind_key_2 =
        RistrettoPoint::from(&pvt_blinding_key * &RISTRETTO_BASEPOINT_TABLE).compress();
    // The rewind nonces are derived from the public rewind keys and the Pedersen commitment
    let rewind_nonce_1 = get_rewind_nonce_from_pub_key(&pub_rewind_key_1, &committed_value);
    let rewind_nonce_2 = get_rewind_nonce_from_pub_key(&pub_rewind_key_2, &committed_value);
    // Get secret nonces from the private keys - this emulates the owner of a wallet
    let blinding_nonce_1 = get_secret_nonce_from_pvt_key(&pvt_rewind_key, &committed_value);
    let blinding_nonce_2 = get_secret_nonce_from_pvt_key(&pvt_blinding_key, &committed_value);

    // Get Value Test 1 - provide correct rewind nonces
    let mut rewind_transcript = Transcript::new(b"Bulletproof-Rewind Test");
    assert_eq!(
        proof.rewind_single_get_value_only(
            &bp_gens,
            &mut rewind_transcript,
            &committed_value,
            64,
            &rewind_nonce_1,
            &rewind_nonce_2,
        ),
        Ok((confidential_value, proof_message))
    );
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -482,11 +482,23 @@ Once again, transaction fees are ignored to simplify the illustration.
...
 
 Alice uses Bob's public key to create a shared secret, \\( k_b \\) for the output commitment, \\( C_b \\), using
 Diffie-Hellman key exchange.
 
+Alice calculates \\( k_b \\) as
+$$
+    k_b = k_{Ob} * {K_Sb}
+$$
+
+Next Alice next uses Bulletproof rewinding to encrypt the value \\( v_b \\) into the the Bulletproof for the commitment \\( C_b \\). For this she uses (\\( k_{rewind} =  Hash(k_{b}) \\) as the encryption key.
```
hansieodendaal: We need two private keys here: `pvt_rewind_key` and `pvt_blinding_key`. This could be something like:
```
k_{rewind} =  Hash(k_{b})
k_{blinding} =  Hash(Hash(k_{b}) | k_{b}))
```
Please see the Bulletproofs code snippet for the full explanation.

hansieodendaal (approved):
LGTM

PR #2841, opened 2021-04-12
[Distribution] OSX and Ubuntu Scripts

StriderDM:
## Description
Amalgamated old and new distribution scripts for osx and ubuntu.

## Motivation and Context
Scripts needed to be updated and ones no longer needed cleaned up

## How Has This Been Tested?
From the repository root:
Archive:
```
sh ${PWD}/scripts/build_dists_archive.sh ${tag}
```
OSX Package (after running archive):
```
sh ${PWD}/applications/tari_base_node/osx-pkg/build_pkg.sh ${version} ${destDir}
```

## Types of changes
* [ ] New Tests
* [ ] Documentation

## Checklist:


scripts/build_dists_zip.sh
```diff
@@ -199,53 +203,77 @@ echo "Hash file $hashFile"
...
-  "scripts/install_tor.sh"
-)
-
-for COPY_FILE in "${COPY_FILES[@]}"; do
-  cp -vr "$COPY_FILE" "$distDir/dist/"
-done
-
-pushd $distDir/dist
+# One click miner
+cp -f -P "${PWD}/applications/tari_base_node/osx/start_all" "${distDir}/start_all"
```
leet4tari: Will the Ubuntu Archive be including the OSX files? Are the OSX files and Ubuntu the same? I have not checked.

StriderDM: There is a `if [ "$osname" == "osx" ]; then` a bit further down, missed these for some reason, will fix. Thanks.

applications/tari_base_node/osx-pkg/build_pkg.sh
```diff
@@ -39,50 +39,74 @@ else
...
-
-pkgbuildResult=$(pkgbuild --root $destDir/pkgRoot \
+# One click miner
+cp "${destDir}/start_all" "${destDir}/pkgRoot/usr/local/bin/"
+cp "${destDir}/runtime/start_all.sh" "${destDir}/pkgRoot/usr/local/bin/runtime/"
+
+# Base Node
+cp "${destDir}/start_tari_base_node" "${destDir}/pkgRoot/usr/local/bin/"
+cp "${destDir}/start_tor" "${destDir}/pkgRoot/usr/local/bin/"
+cp "${destDir}/runtime/start_tari_base_node.sh" "${destDir}/pkgRoot/usr/local/bin/runtime/"
```
leet4tari: Is this FHS compliant? Are we willing to pollute install file system? Will this run without modifying the user path environment?

scripts/build_dists_zip.sh
```diff
@@ -199,53 +203,77 @@ echo "Hash file $hashFile"
...
+cp -f -P "${PWD}/applications/tari_merge_mining_proxy/osx/start_tari_merge_mining_proxy" "${distDir}/start_tari_merge_mining_proxy"
+cp -f -P "${PWD}/applications/tari_merge_mining_proxy/osx/start_xmrig" "${distDir}/start_xmrig"
+cp -f "${PWD}/applications/tari_merge_mining_proxy/osx/runtime/start_tari_merge_mining_proxy.sh" "${distDir}/runtime/start_tari_merge_mining_proxy.sh"
+cp -f "${PWD}/applications/tari_merge_mining_proxy/osx/runtime/start_xmrig.sh" "${distDir}/runtime/start_xmrig.sh"
+cp -f "${PWD}/target/release/tari_merge_mining_proxy" "${distDir}/runtime/tari_merge_mining_proxy"
+
+# 3rd party install
+if [ "$osname" == "osx" ]; then
+cp -f "${PWD}/buildtools/install_xmrig.sh" "${distDir}/runtime/install_xmrig.sh"
+cp -f "${PWD}/buildtools/get_xmrig_osx.ps1" "${distDir}/runtime/get_xmrig_osx.ps1"
```
leet4tari: Are we using somebody else powershell script? I am not sure how well this would go over with Nix users that have bash already installed?

StriderDM: The powershell scripts are our own.

Suggestions?

applications/tari_base_node/osx-pkg/build_pkg.sh
StriderDM: A while back all the files for the different operating systems were aligned to use the same directory layout.

Suggestions?

scripts/build_dists_zip.sh
leet4tari: Everything else is using bash for the Nix environments, I would suggest sticking with that would be already installed and in use.

applications/tari_base_node/osx-pkg/build_pkg.sh
leet4tari: For archives, keeping the same folder layout is great, but for install packages I would suggest following the operating systems packaging suggests or running the risk of installs files/folders into poor locations. ie ```/usr/local/bin/runtime``` is not good. If you wish to install files and folders in a single directory, maybe ```/opt/tari_base_node```, but you will need admin access to write into ```/opt```

scripts/build_dists_zip.sh
hansieodendaal: If you can replace `install_xmrig.sh` and `get_xmrig_osx.ps1` using bash only that would be great; the problem to solve here is to get and install the latest version of `xmrig`, which the `ps1` script does.

hansieodendaal:
On Ubuntu, I could not get any of the scripts to run using the soft links in the root folder, except `start_tor` (`open: invalid option -- 'a'`).
```
$ ./start_tari_base_node 

Starting Base Node

Creating new /tmp/tari_base_node-ubuntu-18.04-x64-unversion--gitclean/config/base_node_id.json
Installing new logfile configuration at /tmp/tari_base_node-ubuntu-18.04-x64-unversion--gitclean/config/log4rs_base_node.yml
Initializing logging according to "/tmp/tari_base_node-ubuntu-18.04-x64-unversion--gitclean/config/log4rs_base_node.yml"
Using existing /tmp/tari_base_node-ubuntu-18.04-x64-unversion--gitclean/config/log4rs_base_node.yml

Spawning Base Node into new terminal..
open: invalid option -- 'a'
Usage: open [OPTIONS] -- command

This utility help you to start a program on a new virtual terminal (VT).

Options:
  -c, --console=NUM   use the given VT number;
  -e, --exec          execute the command, without forking;
  -f, --force         force opening a VT without checking;
  -l, --login         make the command a login shell;
  -u, --user          figure out the owner of the current VT;
  -s, --switch        switch to the new VT;
  -w, --wait          wait for command to complete;
  -v, --verbose       print a message for each action;
  -V, --version       print program version and exit;
  -h, --help          output a brief help message.
```

StriderDM:
@hansieodendaal Those are osx scripts. Did you update your copy of the branch to origin after the branch was forced pushed? Building in full Ubuntu or WSL?

scripts/build_dists_zip.sh
StriderDM: Used a combination of curl and jq in place of powershell in the bash script.

hansieodendaal:
> @hansieodendaal Those are osx scripts. Did you update your copy of the branch to origin after the branch was forced pushed? Building in full Ubuntu or WSL?

This was done on full Ubuntu, not WSL, as the latter does not support opening of terminals form the command line.
The previous `buildtools/create_ubuntu_install_zip.sh` was removed, so I ran `scripts/build_dists_zip.sh`, extracted the zip and used that to execute the Tari links.

I updated and now it works better.

buildtools/windows_inno_installer.iss
```diff
@@ -56,7 +56,7 @@ AppPublisher={#MyAppPublisher}
 AppPublisherURL={#MyAppURL}
 AppSupportURL={#MyAppSuppURL}
 AppUpdatesURL={#MyAppURL}
-DefaultDirName={userdocs}\..\.tari-testnet
+DefaultDirName={userdocs}\..
```
hansieodendaal: If the default install folder name needs to be changed to something else, only a single line needs to be changed, for example, change
```
DefaultDirName={userdocs}\..\.tari-testnet
```
to
```
DefaultDirName={userdocs}\..\tari
```
or to
```
DefaultDirName={userdocs}\..\tari-testnet
```
All other changes are unnecessary.

applications/tari_base_node/osx-pkg/build_pkg.sh
StriderDM: Path has been updated to `/usr/local/tari` now.

hansieodendaal (changes requested):
Hi, I have reviewed the Ubuntu install and changes to the Windows installer. In general Ubuntu tarball creation and runtime work well, but we still have some niggling issues.

1. We need a method to test if Tor is already running in Ubuntu, as an additional terminal trying to start Tor will be opened every time the base node or console wallet is started, even Tor is running already.
1. The Windows installer works perfectly well and already follows the established pattern. 
    - With the Windows installer, a user can select their preferred installation location, which can be anywhere they are permitted to write. The way that you have implemented this change will result in a subfolder `\tari` to always be created in the installation location chosen by the user, so if a user chooses `%USERPROFILE%\tari` as the installation folder, the result would be `%USERPROFILE%\tari\tari`, which will be irritating. 
    - You have also changed the location where the Tari icons will be made available; the temp folder should be in the user root not in `Documents`. The exact location where the Tari shortcuts expect the Tari icons to be is in `%USERPROFILE%\temp\tari_icons`, not in `%USERPROFILE%\Documents\temp\tari_icons`. A temp folder in `Documents` can easily be deleted then the icons will be missing from the shortcuts.
    - Please see the suggested change to the Windows installer script to change the default path.
    - My proposal for the default installation location would be `%USERPROFILE%\tari-testnet`, not `%USERPROFILE%\tari`

Thanks

buildtools/windows_inno_installer.iss
StriderDM: Fair point, will revert.

leet4tari: Just a note, ```.file``` or ```.folder``` in Unix is a hidden file or folder and is meant to be like a hidden settings folder. I think the Windows folder name is just following from the Nix developement. If we keeping with standard folder structure, we should use the same for Windows/OSX/Linux or else we need a doc that shows the mapping for each supported OS?

buildtools/install_xmrig.sh
```diff
@@ -3,34 +3,67 @@
...
+then
+    found_brew=$(which brew)
+    if [ -z "${found_brew}" ]
+    then
+         ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    fi
+fi
+
+# Install Curl
+found_curl=$(which jq)
```
leet4tari: Is this not meant to be ```found_curl=$(which curl)```

StriderDM: Yep, thanks.

scripts/build_dists_archive.sh
```diff
@@ -33,7 +34,8 @@ if [ -f "$envFile" ]; then
 fi
 
 if [ -f "Cargo.toml" ]; then
-  if [ "$1" == "clean" ]; then
+  if [ "$1" == "clean" ] || [ "$(uname)" == "Darwin" ]; then
```
leet4tari: The OR part is going to break this script, as it's going to shift the command line options, if it's OSX or if the first options is ```clean```. Unless you rework this totally, I would suggest removing the OR Darwin bit. This should be up for the build to know.

scripts/build_dists_archive.sh
```diff
@@ -50,8 +52,9 @@ if [ "$(uname)" == "Darwin" ]; then
   #system_profiler SPSoftwareDataType
   #sw_vers
   osversion=$(sw_vers -productVersion | cut -d '.' -f 1,2)
-
+  export OPENSSL_STATIC=1
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+  sDist="ubuntu"
```
leet4tari: Doubling Envs - osname is already set to osx/ubuntu?

scripts/build_dists_archive.sh
```diff
@@ -102,10 +105,14 @@ elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
 
 elif [ "$(expr substr $(uname -s) 1 10)" == "MINGW32_NT" ]; then
   # Do something under 32 bits Windows NT platform
-  echo "32 bits Windows NT platform"
+ echo "32 bits Windows NT platform"
+ echo "Unsupported, use installer"
```
leet4tari: I would suggest a URL explain why it's not support - ie no 32bit support.

scripts/build_dists_archive.sh
```diff
@@ -102,10 +105,14 @@ elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
...
+ exit 3
 elif [ "$(expr substr $(uname -s) 1 10)" == "MINGW64_NT" ]; then
   # Do something under 64 bits Windows NT platform
   echo "64 bits Windows NT platform"
+  echo "Unsupported, use installer"
```
leet4tari: Maybe a link to the Win64 installer?

scripts/build_dists_archive.sh
StriderDM: I haven't found a way yet to set env variables using cargo for the build itself without going further into nightly rust. 

For mac we run into the problem with OpenSSL not being signed if it is dynamically linked, so we need to ensure that this is statically linked here.

This is meant to catch the case of human error when building.

Assume clean always by default and remove the envar?

scripts/build_dists_archive.sh
StriderDM: Nope. osname is not strictly fixed to just those two values in the script.

leet4tari: You only supporting two build options and that is where it is set, else it's going to break ... ie RedHat builds are going to break. I think it's wasteful and more to develop with ... You defaulting to OSX and the setting to Ubuntu in the Linux part, which is what ```$osname``` does, right above it? Your patch, just adding suggestion.

scripts/build_dists_archive.sh
leet4tari: This script is only going to be run by Distro Builders, so I think that forcing clean build for OSX builder every time is wasteful. This only needs to be the case on first build after switching to ```statically linked```? These builds in GHA would always be clean builds, but if you testing locally, doubling your build time seems wrong.  
  
If you wish to always do clean builds, remove the options as an override. I don't have an issue with that, but having this script break for OSX users, because of the use of ```shift``` in the if block, that is just going to be poor.

scripts/build_dists_archive.sh
```diff
@@ -199,53 +206,70 @@ echo "Hash file $hashFile"
...
+
+echo "$(cat ${distDir}/${hashFile})" | shasum -a ${shaSumVal} --check
+
+zip -r "${distDir}/${archiveFile}" *
+echo ${distDir}/${archiveFile}
+shasum -a ${shaSumVal} "${archiveFile}" >> "${distDir}/${archiveFile}.sha${shaSumVal}sum"
+
+echo "$(cat ${distDir}/${archiveFile}.sha${shaSumVal}sum)" | shasum -a ${shaSumVal} --check
 popd
-echo "Delete $distDir"
```
leet4tari: I would leave this in, as a reminder to clean out the temp folder that is created. Thou most Nix OSs do a temp clean up on every reboot. If you not worried about space or debugging, then just remove the temp folder with the ```rm -fr $distDir```

scripts/build_dists_archive.sh
leet4tari: Not a idea to check if the ```osname``` env is set to either of the supported options and exit with a warning if not?

scripts/build_dists_archive.sh
hansieodendaal: I think anything other than OSX or Ubuntu should just say "Platform not supported with this build script."

scripts/build_dists_archive.sh
hansieodendaal: See above

scripts/build_dists_archive.sh
StriderDM: Stripped out the architectures which were not ubuntu or osx.

scripts/build_dists_archive.sh
StriderDM: Removed the shift and clean argument.

scripts/build_dists_archive.sh
StriderDM: Re-added

StriderDM:
@hansieodendaal Pretty sure I have addressed the points in your last comment with the latest commit.

scripts/build_dists_archive.sh
```diff
@@ -0,0 +1,245 @@
...
+
+envFile="$sPath/.env"
+if [ -f "$envFile" ]; then
+  echo "Overriding Environment with $envFile file for settings ..."
+  source "$envFile"
+fi
+
+if [ -f "Cargo.toml" ]; then
+    echo "Cleaning Cargo ... "
+    cargo clean
```
hansieodendaal, line 36: **NIT:** I do not think this (`cargo clean`) should be the default, or it should be able to bypass it with a command-line argument.

buildtools/windows_inno_installer.iss
```diff
@@ -56,7 +56,7 @@ AppPublisher={#MyAppPublisher}
 AppPublisherURL={#MyAppURL}
 AppSupportURL={#MyAppSuppURL}
 AppUpdatesURL={#MyAppURL}
-DefaultDirName={userdocs}\..\.tari-testnet
+DefaultDirName={userdocs}\..\tari
```
hansieodendaal, line 59: **NIT:** I think `tari-testnet` is more suited at this time than `tari`

hansieodendaal (dismissed):
LGTM
- Could not finalize Ubuntu testing due to network connectivity issues
- OSX and Ubuntu startup scripts for `start_all` may need some delays between starting the various Tari executables, as sometimes the miner will start before the base node and/or console wallet, with a resulting GRPC connection error message and no start.
- Couple of NITs noted, nothing serious

PR #2846, opened 2021-04-14
[comms] Use grace period on client side handshake

sdbondi:
## Description

Makes use of the request/response grace period in the handshake on the client-side.
The handshake is no different from any other request response and should
not be treated any differently i.e. by allowing extra time for latency on the client-side.

Also increased client-side handshake timeout (the client-side should wait longer than the server-side given it has a full RTT) 

## Motivation and Context

Log observations: the client would connect, the server would accept (20 seconds later), the client would timeout (15s) before the handshake response was received. It is unclear why messages can take that long to be received, but it may be because of the current peer sync protocol using up yamux credits. In any case, the client-side handshake should include a grace period.  

## How Has This Been Tested?

Existing tests pass

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/comms/src/protocol/rpc/client.rs
+++ b/comms/src/protocol/rpc/client.rs
@@ -213,6 +213,11 @@ impl RpcClientConfig {
     pub fn timeout_with_grace_period(&self) -> Option<Duration> {
         self.deadline.map(|d| d + self.deadline_grace_period)
     }
+
+    /// Returns the handshake timeout
+    pub fn handshake_timeout(&self) -> Duration {
+        self.handshake_timeout
+    }
 }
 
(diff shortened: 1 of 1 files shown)


stringhandler (commented):
Disagree with this approach. I don't think it's worth making handshake timeout a calculated value, instead just increase the handshake timeout to a higher value

sdbondi:
Ah yeah, I agree with that actually since the well-defined grace period only comes into play when there is an RPC deadline  - the main improvement here is to have the client with a greater timeout than the server

sdbondi:
@mikethetike Updated

stringhandler (approved):

PR #2847, opened 2021-04-14
[wallet] Leverage streaming for UTXO recovery

sdbondi:
## Description

Implements a streaming protocol for sync_utxo method. This differs from
the previous protocol in the following ways:
- a single request is made that streams the entire utxo set
- it streams many smaller messages rather than a "batch"
- it optionally does not send pruned/deleted UTXOs (not necessary for wallet recovery)
- it optionally does not send the deleted bitmaps (also not for wallet
  recovery)

The recovery client side was changed in these ways:
- implements the new protocol
- current UTXO MMR index is recorded as part of the receovery resume
  state
- previous UTXO mmr index is used to resume recovery
- retry recovery for each peer a given amount of times before giving up
- print_to_console removed in favour of leveraging recovery events

**Compatibility Notes**
- Wallets including this change will not be able to recover using base nodes that do not support the new protocol
- Current pruned sync is unaffected (Previous sync protocol kept)

## Motivation and Context

Recovery times greatly reduced. 
Last test took 7 minutes (including a failure and resume due to minor reorg)

## How Has This Been Tested?

Manually tested:
- recovering a wallet 
- resuming after wallet shutdown/startup
- resuming after base node shutdown/startup
- recovery after base node error due to minor reorg 

## Types of changes

* [ ] New Tests
* [ ] Documentation

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/recovery.rs
+++ b/applications/tari_console_wallet/src/recovery.rs
@@ -27,7 +27,6 @@ use rustyline::Editor;
 use tari_app_utilities::utilities::ExitCodes;
 use tari_comms::types::CommsPublicKey;
 use tari_core::transactions::types::PrivateKey;
-use tari_crypto::tari_utilities::hex::Hex;
 use tari_key_manager::mnemonic::to_secretkey;
 use tari_wallet::{
     tasks::wallet_recovery::{WalletRecoveryEvent, WalletRecoveryTask},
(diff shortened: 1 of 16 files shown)


stringhandler (commented):
Good to hear that speed is improved and I like the approach in general. However, I don't like the `sync_utxo2` name. We haven't released mainnet yet, so no need to keep versions around for backwards compatiblity. It's also confusing for a new dev to understand the differences. 

I think pruned mode could also gain some improvements here, why not change the horizon sync to use the same method?

sdbondi:
Reasons are as follows:
1. the names of the methods can be changed at any time without breaking changes to the wire protocol (happy to change it to whatever you want)
2. changing horizon sync is not trivial (a day or two of work) and is not in scope of this PR - it can easily be self contained in a subsequent PR as this PR is already quite big - at that point we remove the old method and rename the new one

Let me know what you would like me to do

stringhandler (approved):
Not really happy about the name, but hoping that it gets updated and merged with the horizon sync PR. Otherwise looks good

sdbondi:
@mikethetike thanks, yeah definitely it's a temporary name. Will be named sync_utxos in subsequent PR

hansieodendaal:
Really nice solution @sdbondi !

hansieodendaal:
Tested it to complete wallet recoverey just under 50 min

PR #2851, opened 2021-04-14
[Wallet] Fix wallet service API loop blocking

philipr-za:
## Description
It was observed that during a stress test which fires a series of send_transaction calls in the wallet that there was a significant delay between some calls up to 30 seconds long during which no other Transaction Service events were firing in the services main select! loop. This turned out to be because the `send_transaction `operation involved a call to the Base Node Service (BNS) to check the latest tip height and this service was blocking its select! Loop when it made RPC calls to the base node. If these RPC calls took long the BNS service would not respond to API calls until The RPC request was complete. This blocked the select loops of the Output Manager Service and Transaction Service that relied on this call being fast. This also impacted the `get_balance` operation from the Output Manager Service which relies on a check of the last known tip height.

To fix this the BNS service was refactored so that the GetChainMetadata monitoring occurs in a separate asynchronous task that updates the services state when it gets an answer via RPC. The main select! loop was thus freed up to only respond to API requests that provide either the last updated chain metadata OR if no contact has been made with a base node during this run it will get the last persisted data from the database.

Another improvement that was made for the make-it-rain operation in the console wallet was to prevent the console wallet from performing the UTXO and Transaction validation operations when being run in `script` or `command` mode. The validation operations are long running tasks and should only be run if the console wallet is being run in a long running TUI mode.

@kukabi 

## How Has This Been Tested?
Existing tests 

## Checklist:

### DIFF
--- a/base_layer/wallet/src/base_node_service/service.rs
+++ b/base_layer/wallet/src/base_node_service/service.rs
@@ -25,24 +25,21 @@ use super::{
     error::BaseNodeServiceError,
     handle::{BaseNodeEvent, BaseNodeEventSender, BaseNodeServiceRequest, BaseNodeServiceResponse},
 };
-use crate::storage::database::WalletDatabase;
+use crate::storage::database::{WalletBackend, WalletDatabase};
+use chrono::{NaiveDateTime, Utc};
+use futures::{future::Fuse, pin_mut, select, FutureExt, StreamExt};
+use log::*;
 use std::{
     convert::TryFrom,
-    sync::Arc,
-    time::{Duration, Instant},
+    sync::{Arc, RwLock},
+    time::Duration,
 };
-
-use chrono::{NaiveDateTime, Utc};
-use futures::{pin_mut, StreamExt};
-use log::*;
 use tari_common_types::chain_metadata::ChainMetadata;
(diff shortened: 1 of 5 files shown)


base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -300,3 +206,212 @@ where T: WalletBackend + 'static
         }
     }
 }
+
+async fn monitor_chain_metadata_task<T: WalletBackend + 'static>(
```
sdbondi, line 210: I would rather have wrapped this in a new struct, you would then only have the to `match` out `Err(..)` and `wait_or_shutdown().await` in one place and use `?` elsewhere.

philipr-za: I considered that but also was reusing the existing code as much as possible and you will see that there are two levels of behaviour when an error occurs. Some places result in creating a new client and some just wait the timeout and reuse the current client. So even when using the structure you propose there will still be complexity in how you handle the error thrown by the `?` and now you have split that handling logic into two places which is arguably less readable in a different way 🤷‍♂️

SWvheerden (approved):
LGTM

sdbondi (approved):
Not had a chance to test but LGTM - looks like someone (shakes head at @sdbondi) introduced 🖇️ .

PR #2852, opened 2021-04-15
(WIP-complete unit tests) [tari-script] One sided payment protocol

hansieodendaal:
## Description
Added one-sided-to-other payment transaction to the transaction sending protocols. Consecutive PRs must be extended to include wiring into the user interfaces.

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/handle.rs
+++ b/base_layer/wallet/src/output_manager_service/handle.rs
@@ -42,6 +42,7 @@ use tari_core::transactions::{
     ReceiverTransactionProtocol,
     SenderTransactionProtocol,
 };
+use tari_crypto::{ristretto::RistrettoPublicKey, script::TariScript};
 use tari_service_framework::reply_channel::SenderService;
 use tokio::sync::broadcast;
 use tower::Service;
@@ -54,8 +55,8 @@ pub enum OutputManagerRequest {
     GetCoinbaseTransaction((u64, MicroTari, MicroTari, u64)),
     ConfirmPendingTransaction(u64),
     ConfirmTransaction((u64, Vec<TransactionInput>, Vec<TransactionOutput>)),
-    PrepareToSendTransaction((MicroTari, MicroTari, Option<u64>, String)),
-    CreatePayToSelfTransaction((MicroTari, MicroTari, Option<u64>, String)),
+    PrepareToSendTransaction((MicroTari, MicroTari, Option<u64>, String, TariScript)),
+    CreateOneSidedTransaction((MicroTari, MicroTari, Option<u64>, String, Option<RistrettoPublicKey>)),
     CancelTransaction(u64),
     TimeoutTransactions(Duration),
     GetPendingTransactions,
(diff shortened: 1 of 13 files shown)


base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -54,8 +55,8 @@ pub enum OutputManagerRequest {
     GetCoinbaseTransaction((u64, MicroTari, MicroTari, u64)),
     ConfirmPendingTransaction(u64),
     ConfirmTransaction((u64, Vec<TransactionInput>, Vec<TransactionOutput>)),
-    PrepareToSendTransaction((MicroTari, MicroTari, Option<u64>, String)),
-    CreatePayToSelfTransaction((MicroTari, MicroTari, Option<u64>, String)),
+    PrepareToSendTransaction((MicroTari, MicroTari, Option<u64>, String, TariScript)),
+    CreateOneSidedTransaction((MicroTari, MicroTari, Option<u64>, String, Option<RistrettoPublicKey>)),
```
philipr-za, line 59: So as discussed I think this special case shouldn't be handled in the OMS and you just make the `PrepareToSendTransaction` to accept a script as you have above

base_layer/wallet/src/transaction_service/handle.rs
```diff
@@ -52,7 +52,7 @@ pub enum TransactionServiceRequest {
     GetCompletedTransaction(TxId),
     GetAnyTransaction(TxId),
     SetBaseNodePublicKey(CommsPublicKey),
-    SendTransaction((CommsPublicKey, MicroTari, MicroTari, String)),
+    SendTransaction((CommsPublicKey, MicroTari, MicroTari, String, bool)),
```
philipr-za, line 55: I disagree with this. Adding a boolean argument is almost always bad because its not readable. Just look at this line: With no other context what does that last argument do? You can't know. The other arguments at least tell you something about their nature. Furthermore look at all the changes you had to make to TestNet utils to support this.

I would just add a new descriptive member to this enum and make calls for one side payments a new API call.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -642,9 +661,11 @@ where
             return Ok(tx_id);
         }
 
+        // TODO: Complete wiring of sending transactions with other tari script feature upstream
```
philipr-za, line 664: The reason I wouldn't reuse this function is you are going to have two completely different behaviours in this function now. After you have called prepare_transaction_to_send you will handle this tx differently to this standard tx. Reusing this function means you are going to make this function huge with lots of sub branches. 

The pay to self above is just a special case of a traditional MW transaction where one-sided is a completely different beast.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -3839,13 +3839,16 @@ pub unsafe extern "C" fn wallet_send_transaction(
         CString::new("").unwrap().to_str().unwrap().to_owned()
     };
 
+    // TODO: Complete wiring of one-sided-tx-to-other feature upstream
+    let one_sided_to_other = false;
```
philipr-za, line 3843: I feel like you shouldn't need to touch this crate yet.

hansieodendaal:
Thanks @philipr-za, going the different route now

PR #2857, opened 2021-04-18
[base-node] Optimise pruned UTXO sync streaming protocol

sdbondi:
## Description

Streams a pruned UTXO set to a client. Each block boundary (given by the
header) is punctuated by a deleted MMR difference bitmap.

Replaces the previous sync_rpc method, which has been removed and so,
this PR is not backward-compatible.

## Motivation and Context

Continues from changes in https://github.com/tari-project/tari/pull/2847. 
There are a few other performance optimisations still to be made for horizon sync. 

## How Has This Been Tested?

Manually tested. Sync took 15m.

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
@@ -31,16 +31,23 @@ use crate::{
     },
     blocks::BlockHeader,
     chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend, ChainStorageError, MmrTree, PrunedOutput},
-    proto::base_node::{SyncKernelsRequest, SyncUtxosRequest, SyncUtxosResponse},
+    proto::base_node::{
+        sync_utxo as proto_sync_utxo,
+        sync_utxos_response::UtxoOrDeleted,
+        SyncKernelsRequest,
+        SyncUtxo,
+        SyncUtxosRequest,
+        SyncUtxosResponse,
+    },
     transactions::{
         transaction::{TransactionKernel, TransactionOutput},
-        types::{HashDigest, HashOutput, RangeProofService},
+        types::{HashDigest, RangeProofService},
     },
 };
 use croaring::Bitmap;
 use futures::StreamExt;
 use log::*;
-use std::convert::TryInto;
+use std::convert::{TryFrom, TryInto};
 use tari_comms::PeerConnection;
 use tari_crypto::{
     commitment::HomomorphicCommitment,
@@ -90,7 +97,9 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
             }
         })?;
 
-        match self.begin_sync(&header).await {
+        let mut client = self.sync_peer.connect_rpc::<rpc::BaseNodeSyncRpcClient>().await?;
+
+        match self.begin_sync(&mut client, &header).await {
             Ok(_) => match self.finalize_horizon_sync().await {
                 Ok(_) => Ok(()),
                 Err(err) => {
@@ -105,15 +114,25 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
         }
     }
 
-    async fn begin_sync(&mut self, to_header: &BlockHeader) -> Result<(), HorizonSyncError> {
+    async fn begin_sync(
+        &mut self,
+        client: &mut rpc::BaseNodeSyncRpcClient,
+        to_header: &BlockHeader,
+    ) -> Result<(), HorizonSyncError>
+    {
         debug!(target: LOG_TARGET, "Synchronizing kernels");
-        self.synchronize_kernels(to_header).await?;
+        self.synchronize_kernels(client, to_header).await?;
         debug!(target: LOG_TARGET, "Synchronizing outputs");
-        self.synchronize_outputs(to_header).await?;
+        self.synchronize_outputs(client, to_header).await?;
         Ok(())
     }
 
-    async fn synchronize_kernels(&mut self, to_header: &BlockHeader) -> Result<(), HorizonSyncError> {
+    async fn synchronize_kernels(
+        &mut self,
+        client: &mut rpc::BaseNodeSyncRpcClient,
+        to_header: &BlockHeader,
+    ) -> Result<(), HorizonSyncError>
+    {
         let local_num_kernels = self.db().fetch_mmr_size(MmrTree::Kernel).await?;
 
         let remote_num_kernels = to_header.kernel_mmr_size;
@@ -138,12 +157,6 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
             remote_num_kernels - local_num_kernels,
         );
 
-        self.sync_kernel_nodes(local_num_kernels, remote_num_kernels, to_header.hash())
-            .await
-    }
-
-    async fn sync_kernel_nodes(&mut self, start: u64, end: u64, end_hash: HashOutput) -> Result<(), HorizonSyncError> {
-        let mut client = self.sync_peer.connect_rpc::<rpc::BaseNodeSyncRpcClient>().await?;
         let latency = client.get_last_request_latency().await?;
         debug!(
             target: LOG_TARGET,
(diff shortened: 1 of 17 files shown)


base_layer/core/src/base_node/sync/rpc/service.rs
```diff
@@ -409,218 +409,184 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
...
-            req.include_deleted_bitmaps
-        );
+                    let start = cmp::max(self.request.start, prev_header.output_mmr_size);
+                    let end = current_header.output_mmr_size - 1;
 
-        if req.start > end_header.output_mmr_size {
-            return Err(RpcStatus::bad_request("start index is greater than end index"));
-        }
+                    if tx.is_closed() {
+                        debug!(target: LOG_TARGET, "Exiting sync_utxos early because client has gone",);
```
philipr-za, line 508: But where does a client go? Did it go to the big server farm in the sky?

philipr-za (dismissed):
LGTM

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -181,14 +198,12 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
             if mmr_position == current_header.header.kernel_mmr_size - 1 {
                 debug!(
                     target: LOG_TARGET,
-                    "Checking header {}, added {} kernels",
+                    "Block #{}, added {} kernels",
```
SWvheerden: I think the header is more clear what's actually happening here.  We do download them in batches from blocks, but we only down the header and the kernel.

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -287,71 +305,111 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
...
+                            current_header.header.output_mmr_size, mmr_position
+                        )));
+                    }
 
-                if mmr_position == current_header.header.output_mmr_size - 1 {
-                    trace!(
+                    debug!(
                         target: LOG_TARGET,
-                        "Checking header {}, added {} utxos, added {} txos)",
+                        "UTXO: {} (Block #{}), added {} utxos, added {} txos",
```
SWvheerden:
```suggestion
                        "UTXO: {} (Header #{}), added {} utxos, added {} txos",
```

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -287,71 +305,111 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
...
+                .utxo_or_deleted
+                .ok_or_else(|| HorizonSyncError::IncorrectResponse("Peer sent no transaction output data".into()))?;
+
+            match txo {
+                UtxoOrDeleted::Utxo(SyncUtxo {
+                    utxo: Some(proto_sync_utxo::Utxo::Output(output)),
+                }) => {
+                    trace!(
+                        target: LOG_TARGET,
+                        "UTXO {} received from sync peer for block #{}",
```
SWvheerden:
```suggestion
                        "UTXO {} received from sync peer for header #{}",
```

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -240,7 +255,12 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
         Ok(())
     }
 
-    async fn synchronize_outputs(&mut self, to_header: &BlockHeader) -> Result<(), HorizonSyncError> {
+    async fn synchronize_outputs(
```
SWvheerden, line 258: This function looks much more complex than before. Is this passing with `cognitive-complexity-threshold = 15`

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1436,96 +1437,107 @@ impl BlockchainBackend for LMDBDatabase {
...
-                0
-            } else {
-                let header: BlockHeader =
-                    lmdb_get(&txn, &self.headers_db, &(start_height - 1))?.expect("Header should exist");
-                debug!(target: LOG_TARGET, "Previous header:{}", header);
-                header.output_mmr_size
-            };
+        let start_height = lmdb_first_after(&txn, &self.output_mmr_size_index, &(start + 1).to_be_bytes())?
+            .ok_or_else(|| {
+                ChainStorageError::InvalidQuery(format!(
```
SWvheerden, line 1445: Is there not error to handle this conversion?
I see prev it just had a "?"

SWvheerden (commented):
Looks good. 
I think the `fn synchronize_outputs(` is a bit more complex. 
I also don't like the use of the word `block` in the logs for the pruned mode sync. This is technically incorrect and I think confusing as we never download blocks. I flag some but not all uses

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
sdbondi: I think that's fair

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: No. Previously, if it was None it would return `  Ok((vec![], vec![]))`

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
sdbondi: The function was previously split ~~in two~~ just for the sake of making it somewhat shorter without encapsulating anything (it was called `sync_output_nodes` but may as well be called `synchronize_outputs_inner`). Nor did it improving readability, reduce code duplication or reduce cognitive load. I did not want to pass the client from the outer to the inner nor establish a new client (I'm sure you'd agree, passing additional bytes over the wire in order to avoid adding a single function parameter is an anti-pattern). I agree that some of the code within the function should be better encapsulated however it is somewhat simpler than what it was.

stringhandler: The previous method was split into 2 because it was closer to the method before that. Happy to have them merged into one

stringhandler (commented):
Happy with the approach, but I am concerned about the single DeletedBitmap and can't see how it maps to the MMR hash in the header in all cases.

sdbondi:
@mikethetike The deleted _difference_ bitmap is sent at the end fo each block boundary (as given by output mmr size).
so : UTXO_1 , UTXO_2, ... , deleted diff indexes (signals that we should check the MMR/end of the block) 

 The only difference to what is sent from before is that instead of many difference bitmaps `Vec<Bitmap>` I send one union bitmap containing all the _newly_ spent UTXO indexes, these are equivalent in the data they contain. On the client side, we take the previous blocks bitmap and union the differences and check the MMR. So in effect, nothing has changed from before except it's a little smaller over the wire and only 1 bitmap needs to be serialized.

EDIT:
The way fetch_utxos_by_mmr_position works is very confusing, but it will always give you the difference in deleted MMR index differences between `h - 1` and `h` where `h` is the header heights where start, end index overlap - even if start does not start at the beginning of a block. I don't think we really need this mapping now that we have the output_mmr_size in the headers - outputs should simply be indexed by the integer key mmr_position. Fetching the deleted difference could probably be another call, riffing but something like  `fetch_spent_txo_diff_bitmap_at(height: u64)`

EDIT2:
I think that the biggest perf optimisation after this will be syncing kernels and UTXOs concurrently followed by building an accumulated commitment while syncing for use in the finialise step

stringhandler (approved):
Ok, happy

PR #2858, opened 2021-04-19
Add one-sided payments transaction protocol for tari script

hansieodendaal:
## Description
- Added one-sided payment transaction to other to the transaction sending protocols as another tari script transaction variant. Consecutive PRs must extend the wiring into the user interfaces.
- Added one-sided unit tests.
- Generalized the normal `prepare_transaction_to_send` interface in the output manager service to accept a variable tari script

## Motivation and Context
Expanding tari script

## How Has This Been Tested?
Developed two unit tests

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/transaction_protocol/sender.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/sender.rs
@@ -258,6 +258,23 @@ impl SenderTransactionProtocol {
         }
     }
 
+    /// This function will return the script offset private keys for a single recipient
+    pub fn get_recipient_script_offset_private_key(&self, recipient_index: usize) -> Result<PrivateKey, TPE> {
+        match &self.state {
+            SenderState::Initializing(info) |
+            SenderState::Finalizing(info) |
+            SenderState::SingleRoundMessageReady(info) |
+            SenderState::CollectingSingleSignature(info) => Ok({
+                info.recipient_script_offset_private_keys
+                    .get(recipient_index)
+                    .ok_or_else(|| TPE::ScriptOffsetPrivateKeyNotFound)?
+                    .clone()
+            }),
+            SenderState::FinalizedTransaction(_) => Err(TPE::InvalidStateError),
+            SenderState::Failed(_) => Err(TPE::InvalidStateError),
+        }
+    }
+
     /// This function will return the value of the fee of this transaction
     pub fn get_fee_amount(&self) -> Result<MicroTari, TPE> {
         match &self.state {
@@ -265,7 +282,9 @@ impl SenderTransactionProtocol {
             SenderState::Finalizing(info) |
             SenderState::SingleRoundMessageReady(info) |
             SenderState::CollectingSingleSignature(info) => Ok(info.metadata.fee),
-            SenderState::FinalizedTransaction(_) => Err(TPE::InvalidStateError),
+            SenderState::FinalizedTransaction(info) => {
+                Ok(info.body.kernels().first().ok_or_else(|| TPE::InvalidStateError)?.fee)
+            },
             SenderState::Failed(_) => Err(TPE::InvalidStateError),
         }
     }
--- a/base_layer/wallet/src/transaction_service/error.rs
+++ b/base_layer/wallet/src/transaction_service/error.rs
@@ -40,6 +40,8 @@ use tokio::sync::broadcast::RecvError;
 pub enum TransactionServiceError {
     #[error("Transaction protocol is not in the correct state for this operation")]
     InvalidStateError,
+    #[error("One-sided transaction error: `{0}`")]
+    OneSidedTransactionError(String),
     #[error("Transaction Protocol Error: `{0}`")]
     TransactionProtocolError(#[from] TransactionProtocolError),
     #[error("The message being processed is not recognized by the Transaction Manager")]
@@ -137,6 +139,8 @@ pub enum TransactionServiceError {
     ProtobufConversionError(String),
     #[error("Maximum Attempts Exceeded")]
     MaximumAttemptsExceeded,
+    #[error("Byte array error")]
+    ByteArrayError(#[from] tari_crypto::tari_utilities::ByteArrayError),
 }
 
 #[derive(Debug, Error)]
--- a/base_layer/wallet/src/transaction_service/handle.rs
+++ b/base_layer/wallet/src/transaction_service/handle.rs
@@ -52,10 +52,11 @@ pub enum TransactionServiceRequest {
     GetCompletedTransaction(TxId),
     GetAnyTransaction(TxId),
     SetBaseNodePublicKey(CommsPublicKey),
-    SendTransaction((CommsPublicKey, MicroTari, MicroTari, String)),
+    SendTransaction(CommsPublicKey, MicroTari, MicroTari, String),
+    SendOneSidedTransaction(CommsPublicKey, MicroTari, MicroTari, String),
     CancelTransaction(TxId),
     ImportUtxo(MicroTari, CommsPublicKey, String),
-    SubmitTransaction((TxId, Transaction, MicroTari, MicroTari, String)),
+    SubmitCoinSplitTransaction(TxId, Transaction, MicroTari, MicroTari, String),
     SetLowPowerMode,
     SetNormalPowerMode,
     ApplyEncryption(Box<Aes256Gcm>),
@@ -89,12 +90,15 @@ impl fmt::Display for TransactionServiceRequest {
             Self::GetCancelledCompletedTransactions => f.write_str("GetCancelledCompletedTransactions"),
             Self::GetCompletedTransaction(t) => f.write_str(&format!("GetCompletedTransaction({})", t)),
             Self::SetBaseNodePublicKey(k) => f.write_str(&format!("SetBaseNodePublicKey ({})", k)),
-            Self::SendTransaction((k, v, _, msg)) => {
-                f.write_str(&format!("SendTransaction (to {}, {}, {})", k, v, msg))
+            Self::SendTransaction(k, v, _, msg) => f.write_str(&format!("SendTransaction (to {}, {}, {})", k, v, msg)),
+            Self::SendOneSidedTransaction(k, v, _, msg) => {
+                f.write_str(&format!("SendOneSidedTransaction (to {}, {}, {})", k, v, msg))
             },
             Self::CancelTransaction(t) => f.write_str(&format!("CancelTransaction ({})", t)),
             Self::ImportUtxo(v, k, msg) => f.write_str(&format!("ImportUtxo (from {}, {}, {})", k, v, msg)),
-            Self::SubmitTransaction((id, _, _, _, _)) => f.write_str(&format!("SubmitTransaction ({})", id)),
+            Self::SubmitCoinSplitTransaction(tx_id, _, _, _, _) => {
+                f.write_str(&format!("SubmitTransaction ({})", tx_id))
+            },
             Self::SetLowPowerMode => f.write_str("SetLowPowerMode "),
             Self::SetNormalPowerMode => f.write_str("SetNormalPowerMode"),
             Self::ApplyEncryption(_) => f.write_str("ApplyEncryption"),
@@ -221,12 +225,35 @@ impl TransactionServiceHandle {
     {
         match self
             .handle
-            .call(TransactionServiceRequest::SendTransaction((
+            .call(TransactionServiceRequest::SendTransaction(
                 dest_pubkey,
                 amount,
                 fee_per_gram,
                 message,
-            )))
+            ))
+            .await??
+        {
+            TransactionServiceResponse::TransactionSent(tx_id) => Ok(tx_id),
+            _ => Err(TransactionServiceError::UnexpectedApiResponse),
+        }
+    }
+
+    pub async fn send_one_sided_transaction(
+        &mut self,
+        dest_pubkey: CommsPublicKey,
+        amount: MicroTari,
+        fee_per_gram: MicroTari,
+        message: String,
+    ) -> Result<TxId, TransactionServiceError>
+    {
+        match self
+            .handle
+            .call(TransactionServiceRequest::SendOneSidedTransaction(
+                dest_pubkey,
+                amount,
+                fee_per_gram,
+                message,
+            ))
             .await??
         {
             TransactionServiceResponse::TransactionSent(tx_id) => Ok(tx_id),
@@ -391,7 +418,7 @@ impl TransactionServiceHandle {
 
     pub async fn submit_transaction(
         &mut self,
-        tx_id: u64,
+        tx_id: TxId,
         tx: Transaction,
         fee: MicroTari,
         amount: MicroTari,
@@ -400,9 +427,9 @@ impl TransactionServiceHandle {
     {
         match self
             .handle
-            .call(TransactionServiceRequest::SubmitTransaction((
+            .call(TransactionServiceRequest::SubmitCoinSplitTransaction(
                 tx_id, tx, fee, amount, message,
-            )))
+            ))
             .await??
         {
             TransactionServiceResponse::TransactionSubmitted => Ok(()),
--- a/base_layer/wallet/src/transaction_service/service.rs
+++ b/base_layer/wallet/src/transaction_service/service.rs
@@ -35,7 +35,7 @@ use crate::{
         },
         storage::{
             database::{TransactionBackend, TransactionDatabase},
-            models::{CompletedTransaction, TransactionDirection, TransactionStatus},
+            models::{CompletedTransaction, InboundTransaction, TransactionDirection, TransactionStatus},
         },
         tasks::{
             send_finalized_transaction::send_finalized_transaction_message,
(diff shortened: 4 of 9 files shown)


base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -258,14 +258,45 @@ impl SenderTransactionProtocol {
         }
     }
 
+    /// This function will return the script offset private keys for multiple recipients
+    pub fn get_recipient_script_offset_private_keys(&self) -> Result<Vec<PrivateKey>, TPE> {
```
philipr-za: I think you can make do with a single method for this one and the one below that is more useful than either.

I would rather use a method with the following signature: `pub fn get_recipient_script_offset_private_key(&self, recipient_index: usize)`. If you are dealing with a single recipient you know to use index 0, for more then you ask as appropriate. Then there is only one method to do this rather than 2 with lots of repetition.

base_layer/wallet/src/transaction_service/error.rs
```diff
@@ -40,6 +40,8 @@ use tokio::sync::broadcast::RecvError;
 pub enum TransactionServiceError {
     #[error("Transaction protocol is not in the correct state for this operation")]
     InvalidStateError,
+    #[error("One-sided tx error: `{0}`")]
```
philipr-za: Personally I prefer to use the full word transaction.

philipr-za (commented):
Looks good, just a couple small comments.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Makes sense, thanks

base_layer/wallet/src/transaction_service/error.rs
hansieodendaal: Better yes

philipr-za (approved):
LGTM

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+            JoinHandle<Result<u64, TransactionServiceProtocolError>>,
+        >,
+    ) -> Result<TxId, TransactionServiceError>
+    {
+        if self.node_identity.public_key() == &dest_pubkey {
+            warn!(target: LOG_TARGET, "One-sided spend-to-self transactions not supported");
+            return Err(TransactionServiceError::OneSidedTransactionError(
+                "One-sided spend-to-self transactions not supported".to_string(),
+            ));
+        }
```
SWvheerden, line 718: Is there a reason we are blocking this?
I think it could be useful for just updating the script and or changing the blinding factor etc.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+        // Prepare sender part of the transaction
+
+        let mut stp = self
+            .output_manager_service
+            .prepare_transaction_to_send(
+                amount,
+                fee_per_gram,
+                None,
+                message.clone(),
+                script!(PushPubKey(Box::new(dest_pubkey.clone()))),
```
SWvheerden, line 729: This script is not correct. 
It will verify for any 0 input. 

If we want p2pkh we need: 
`script!(Dup HashBlake256 PushHash(dest_pubkey.hash()) EqualVerify);`

if we want p2pk
`script!(Dup PushPubKey(dest_pubkey) EqualVerify);`

base_layer/wallet/tests/transaction_service/service.rs
```diff
@@ -663,6 +664,186 @@ fn single_transaction_to_self() {
...
+    let mut runtime = create_runtime();
+
+    let factories = CryptoFactories::default();
+    // Alice's parameters
+    let alice_node_identity = Arc::new(
+        NodeIdentity::random(&mut OsRng, get_next_memory_address(), PeerFeatures::COMMUNICATION_NODE).unwrap(),
+    );
+
+    // Bob's parameters
+    let bob_node_identity = Arc::new(
```
SWvheerden, line 678: Copy paste?
We have Bob here, but bob is never used?

SWvheerden (changes requested):
The script wont work as implemented here. 

I think we need a test with an Alice Wallet sending, and a bob wallet claiming.  We need to verify that we can claim the one-sided transaction with a valid input only.

base_layer/wallet/tests/transaction_service/service.rs
hansieodendaal: It is used in line 689 and 728; Alice sending to Bob.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: The reason it is blocked is because we can send a normal transaction to self. This will also achieve changing the blinding factor.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: I could easily change this, but first please confirm that [RFC-0202](https://rfc.tari.com/RFC-0202_TariScriptOpcodes.html#one-sided-transactions) is wrong? It states:
_One-sided transactions lock the input to a predetermined public key provided by the recipient; essentially the same method that Bitcoin uses. The simplest form of this is to simply post the new owner's public key as the script: `PushPubkey(P_B)`_

philipr-za: The simple script works as only the owner can produce the script signature. Do we really need to also have them confirm that they know the destination public key in the P2PKH example? p2pk adds no value as far as I can tell.

base_layer/wallet/tests/transaction_service/service.rs
SWvheerden: 689 is a log message
But I see, 728 is sending to his pubkey. Missed that one. 

We should try and let Bob spend this tx. Then we know its valid. 
We only currently know its constructed. Not if its correctly constructed.

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: You are both correct, I stand corrected. `PushPubkey(P_B)` is in all essence the same as p2pk. 

Anyone can provide a valid input to this script as it's an empty stack. But only the true recipient can sign the script offset as p_B is required for that and only the owner of P_B should know that.

SWvheerden (approved):

base_layer/wallet/tests/transaction_service/service.rs
philipr-za: That is for the next sprint.

hansieodendaal: Agreed; ROI will not be worth it without the receiver bits implemented.

base_layer/wallet/src/transaction_service/handle.rs
```diff
@@ -53,9 +53,10 @@ pub enum TransactionServiceRequest {
     GetAnyTransaction(TxId),
     SetBaseNodePublicKey(CommsPublicKey),
     SendTransaction((CommsPublicKey, MicroTari, MicroTari, String)),
+    SendOneSidedTransaction((CommsPublicKey, MicroTari, MicroTari, String)),
```
stringhandler: nit: I see this is the same as SendTransaction, but in future, it would be better to use field names instead of a tuple.

The extra wrapping `()` are also not needed

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+            )
+            .await?;
+        let tx_id = stp.get_tx_id()?;
+        if !stp.is_single_round_message_ready() {
```
stringhandler: Under what conditions can this if statement ever be true?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+            return Err(TransactionServiceError::OneSidedTransactionError(
+                "Sender protocol single round message not ready".to_string(),
+            ));
+        }
+
+        let _ = stp
+            .build_single_round_message()
+            .map_err(|e| TransactionServiceProtocolError::new(tx_id, TransactionServiceError::from(e)))?;
+        if !stp.is_collecting_single_signature() {
```
stringhandler: Under what conditions is this check not met?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+            return Err(TransactionServiceError::OneSidedTransactionError(
+                "Sender protocol single round signature not being collected".to_string(),
+            ));
+        }
+        self.output_manager_service
+            .confirm_pending_transaction(tx_id)
+            .await
+            .map_err(|e| TransactionServiceProtocolError::new(tx_id, TransactionServiceError::from(e)))?;
```
stringhandler:
```suggestion
            .map_err(|e| TransactionServiceProtocolError::new(tx_id, e.into())?;
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+
+        // Prepare receiver part of the transaction
+
+        // Diffie-Hellman shared secret `k_Ob * K_Sb = K_Ob * k_Sb` results in a public key, which is converted to
+        // bytes to enable conversion into a private key to be used as the spending key
+        // TODO: Add a standardized Diffie-Hellman method to the tari_crypto library, then come back and use it here
```
stringhandler: Is there not one?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+            &self.resources.factories,
+        );
+
+        let recipient_reply = rtp.get_signed_data()?.clone();
+
+        // Start finalizing
+
+        stp.add_single_recipient_info(recipient_reply, &self.resources.factories.range_proof)
+            .map_err(|e| TransactionServiceProtocolError::new(tx_id, TransactionServiceError::from(e)))?;
+        if !stp.is_finalizing() {
```
stringhandler: Again, this check is probably not needed

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+
+        stp.finalize(KernelFeatures::empty(), &self.resources.factories)
+            .map_err(|e| {
+                error!(
+                    target: LOG_TARGET,
+                    "Transaction (TxId: {}) could not be finalized. Failure error: {:?}", tx_id, e,
+                );
+                TransactionServiceProtocolError::new(tx_id, TransactionServiceError::from(e))
+            })?;
+        if !stp.is_finalized() {
```
stringhandler: Check again probably not needed

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+                "Sender protocol single round tx not being finalized".to_string(),
+            ));
+        }
+
+        // Finalize
+
+        stp.finalize(KernelFeatures::empty(), &self.resources.factories)
+            .map_err(|e| {
+                error!(
+                    target: LOG_TARGET,
```
stringhandler, line 780: Is there a reason that for the extra log here?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -673,6 +694,143 @@ where
...
+                TransactionServiceProtocolError::new(tx_id, TransactionServiceError::from(e))
+            })?;
+        if !stp.is_finalized() {
+            return Err(TransactionServiceError::OneSidedTransactionError(
+                "Sender protocol single round tx not finalized".to_string(),
+            ));
+        }
+        trace!(target: LOG_TARGET, "Finalized one-side transaction TxId: {}", tx_id);
+
+        let _ = self
```
stringhandler, line 789: What's the reason for swallowing this error?

stringhandler (commented):
Happy with the approach. I don't think the `if` checks that I noted are necessary, but please see comments and answer them individually if they are

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Not currently no

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Because we do not use the series of transaction protocols over the wire directly here, certain checks are needed to make sure the sender transaction protocol has the correct state. 

This test is specifically inserted to let a future maintainer know that the sender transaction protocol should be in a specific state. This is a cheap operation and worth much more than a comment. If something is changed in the transaction protocols that influence the state, this test must still pass.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Please see previous comment

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Please see previous comment

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Please see previous comment

base_layer/wallet/src/transaction_service/handle.rs
```diff
@@ -53,9 +53,10 @@ pub enum TransactionServiceRequest {
...
     CancelTransaction(TxId),
     ImportUtxo(MicroTari, CommsPublicKey, String),
-    SubmitTransaction((TxId, Transaction, MicroTari, MicroTari, String)),
+    SubmitTransaction(CompletedTransaction),
```
stringhandler: Is this not leaking the CompletedTransaction (which is in the storage mod) to the public? I think it is exposing it here more than it should be

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Yes, it is to help with system-level debugging.

base_layer/wallet/src/transaction_service/handle.rs
stringhandler: Or perhaps CompletedTransaction should be moved out of the storage mod. That would be outside the scope of this pr

base_layer/wallet/src/transaction_service/service.rs
stringhandler: Is there another thread that can access this data somehow? Surely the object is only available to be edited by this code in the method? These extra checks take up space and clutter the actual code that is necessary

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: This event being sent is important, but not critical to the protocol being successful. It could be that the receiver dropped a handle for some reason.

_From the Tokio documentation:_
A return value of Ok does not mean that the sent value will be observed by all or any of the active Receiver handles. Receiver handles may be dropped before receiving the sent message. A return value of Err does not mean that future calls to send will fail.

base_layer/wallet/src/transaction_service/service.rs
stringhandler: Not this one? https://github.com/tari-project/tari-crypto/blob/b9d434b19c48993d47da3a37b22548d6a3bab265/src/ristretto/ristretto_keys.rs#L247

```
impl DiffieHellmanSharedSecret for RistrettoPublicKey {
```

base_layer/wallet/src/transaction_service/service.rs
stringhandler: but won't it get logged by a caller of this method? Adding in extra error logging clouds the important code

base_layer/wallet/src/transaction_service/handle.rs
hansieodendaal: I did not think so no, as this interface is only used for coin spilt. I will change it.

philipr-za: We use the CompletedTransaction as a public model all over (sent in events, returned from accessors etc) so I suppose it should come out of storage if you think that  is not the place for it. Though while I agree the internals of storage is private it does feel like the right place to define data structures?

base_layer/wallet/src/transaction_service/service.rs
stringhandler: Anything like this that is out of the ordinary should have a comment stating why it can be swallowed, so that new devs don't copy this code thinking it's the norm

base_layer/wallet/src/transaction_service/handle.rs
stringhandler: Yeah, potentially a subjective choice. I would say that Models should be outside of storage since they can be used without persistence. But sounds like that is a discussion outside of the scope of this PR

base_layer/wallet/src/transaction_service/handle.rs
hansieodendaal: - `SendOneSidedTransaction` is specifically made into a separate interface from `SendTransaction`, otherwise, a differentiating parameter must be added to the interface to distinguish the two variants, with widespread impact. I believe this to be a cleaner solution.
- Tuple removed from the new and old interfaces

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: But, you are correct, the methods are deterministic, so if it advances beyond the error check, the test is not needed.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Will try it out

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: In my experience add the log where it is needed

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Agreed, added a proper comment

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Ok, that is halfway there. Used it, but we still need a Diffie-Hellman that will return a private key.

stringhandler (approved):

PR #2864, opened 2021-04-20
refactor: additional DB audit of methods

SWvheerden:
## Description

This provides a audit removing and increasing security over the following db methods from the WriteOperation enum:

```rust
InsertChainOrphanBlock(Arc<ChainBlock>),
InsertInput {
 header_hash: HashOutput,
input: Box<TransactionInput>,
mmr_position: u32,
},

InsertKernel {
 header_hash: HashOutput,
kernel: Box<TransactionKernel>,
mmr_position: u32,
},

InsertOutput {
 header_hash: HashOutput,
output: Box<TransactionOutput>,
mmr_position: u32,
},

DeleteHeader(u64),

DeleteOrphanChainTip(HashOutput),

InsertOrphanChainTip(HashOutput),

SetBestBlock {
 height: u64,
hash: HashOutput,
accumulated_difficulty: u128,
},

SetPruningHorizonConfig(u64),

SetPrunedHeight {
 height: u64,
kernel_sum: Commitment,
utxo_sum: Commitment,
},
```

## How Has This Been Tested?

This synced to tip, and passed all unit tests

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
@@ -538,6 +538,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                 .await?
                 .into_bitmap(),
         );
+        let expected_prev_best_block = self.shared.db.get_chain_metadata().await?.best_block().clone();
         for h in 0..=header.height() {
             let curr_header = self.db().fetch_chain_header(h).await?;
 
@@ -622,6 +623,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                 header.height(),
                 header.hash().clone(),
                 header.accumulated_data().total_accumulated_difficulty,
+                expected_prev_best_block,
             )
             .set_pruned_height(header.height(), pruned_kernel_sum, pruned_utxo_sum)
             .commit()
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -226,6 +226,7 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
                     block.height(),
                     header_hash,
                     block.accumulated_data().total_accumulated_difficulty,
+                    block.header().prev_hash.clone(),
                 )
                 .commit()
                 .await?;
@@ -249,17 +250,6 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
         }
 
         if let Some(block) = current_block {
-            // Update metadata to last tip header
-            let header = &block.header();
-            let height = header.height;
-            let best_block = header.hash();
-            let accumulated_difficulty = block.accumulated_data().total_accumulated_difficulty;
-            self.db
-                .write_transaction()
-                .set_best_block(height, best_block.to_vec(), accumulated_difficulty)
-                .commit()
-                .await?;
-
             self.hooks.call_on_complete_hooks(block);
         }
 
--- a/base_layer/core/src/chain_storage/db_transaction.rs
+++ b/base_layer/core/src/chain_storage/db_transaction.rs
@@ -223,11 +223,18 @@ impl DbTransaction {
         self
     }
 
-    pub fn set_best_block(&mut self, height: u64, hash: HashOutput, accumulated_difficulty: u128) -> &mut Self {
+    pub fn set_best_block(
+        &mut self,
+        height: u64,
+        hash: HashOutput,
+        accumulated_difficulty: u128,
+        expected_prev_best_block: HashOutput,
+    ) -> &mut Self {
         self.operations.push(WriteOperation::SetBestBlock {
             height,
             hash,
             accumulated_difficulty,
+            expected_prev_best_block,
         });
         self
     }
@@ -323,6 +330,7 @@ pub enum WriteOperation {
         height: u64,
         hash: HashOutput,
         accumulated_difficulty: u128,
+        expected_prev_best_block: HashOutput,
     },
     SetPruningHorizonConfig(u64),
     SetPrunedHeight {
@@ -421,6 +429,7 @@ impl fmt::Display for WriteOperation {
                 height,
                 hash,
                 accumulated_difficulty,
+                expected_prev_best_block: _,
             } => write!(
                 f,
                 "Update best block to height:{} ({}) with difficulty: {}",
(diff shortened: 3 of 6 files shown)


base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/error.rs
```diff
@@ -66,4 +66,6 @@ pub enum HorizonSyncError {
     ConversionError(String),
     #[error("MerkleMountainRangeError: {0}")]
     MerkleMountainRangeError(#[from] MerkleMountainRangeError),
+    #[error("The header height changed during validation")]
+    HeaderHeightChanged,
```
stringhandler: Suggestion: A concurrency error might be more general. The problem with the naming here is that it sounds like the height field of a single header has changed. Perhaps also `ChainHeightChanged` or `TipChanged`
```suggestion
     ConcurrencyError,
```

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -524,6 +525,15 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
             target: LOG_TARGET,
             "Horizon state validation succeeded! Committing horizon state."
         );
+        if syncing_header_hash !=
```
stringhandler: This check should be moved to inside the `write_transaction`. There is still a chance that another thread can change the database in between this check and the `write_transaction` starting

stringhandler (commented):
I like the removal of code. The check you have added to horizon_state_synchronization is not sufficiently thread-safe though.

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -238,22 +239,6 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
...
-            let height = header.height;
-            let best_block = header.hash();
-            let accumulated_difficulty = block.accumulated_data.total_accumulated_difficulty;
-            self.db
-                .write_transaction()
-                .set_best_block(height, best_block.to_vec(), accumulated_difficulty)
-                .commit()
-                .await?;
-
-            self.hooks.call_on_complete_hooks(block);
```
sdbondi, line 265: `call_on_complete_hooks` is used by block sync base node state to notify the mempool that the chain has changed. This needs to be called after the loop with the current tip block.

SWvheerden: yeah but it calls it twice? or am I reading the code wrong?
Its called in line 227

sdbondi: 227 is different, that is the "progress" hook  
```rust
  self.hooks
                .call_on_progress_block_hooks(block.clone(), tip_height, &[peer.clone()]);
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -261,7 +267,30 @@ impl LMDBDatabase {
...
+                        if from != prev {
+                            return Err(ChainStorageError::InvalidOperation(format!(
+                                "There was a change in best_block, the best block is suppose to be: ({}), but it \
+                                 currently is: ({})",
+                                from.to_hex(),
+                                prev.to_hex(),
+                            )));
+                        };
+                    }
+                    let header_height: Option<u64> = lmdb_get(&write_txn, &self.block_hashes_db, hash.as_slice())?;
```
sdbondi: Use `lmdb_exists` (skips deserialisation of the value)
`if !lmdb_exists(&write_txn, &self.block_hashes_db, hash.as_slice())? {`

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -261,7 +267,30 @@ impl LMDBDatabase {
                     height,
                     hash,
                     accumulated_difficulty,
+                    from,
                 } => {
+                    // for security we check that the best block does exist, and we check the previous value
+                    // we dont want to check this if the prev block has never been set, this means a empty hash of 32
+                    // bytes.
+                    if from != vec![0; 32] {
```
sdbondi: I think we should check (before update):
1. the new best block header exists (does the header hash exist i.e. `lmdb_exists`)
2. the new block previous hash is set to the current best block hash (from metadata)
3. The height is one more than the current best block height (from metadata)

These checks are pretty cheap and assert that the new best block follows from the previous best block and the best block actually exists

sdbondi: We then do not rely on `from` being set correctly - we can remove `from`

sdbondi: Ok 2 and 3 are too restrictive given what is needed for pruned mode - 1. is already included below

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -261,7 +267,29 @@ impl LMDBDatabase {
...
```
sdbondi: 1. AFAICS you want to ignore the genesis block - so can use `if height > 0`?
2. `from` is a bit fuzzy, perhaps `expected_prev_block`?

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -242,16 +244,6 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
...
-            // Update metadata to last tip header
-            let header = &block.block.header;
-            let height = header.height;
-            let best_block = header.hash();
-            let accumulated_difficulty = block.accumulated_data.total_accumulated_difficulty;
-            self.db
-                .write_transaction()
-                .set_best_block(height, best_block.to_vec(), accumulated_difficulty)
-                .commit()
-                .await?;
```
sdbondi:
```suggestion
```

base_layer/core/src/chain_storage/db_transaction.rs
```diff
@@ -211,11 +202,19 @@ impl DbTransaction {
         self
     }
 
-    pub fn set_best_block(&mut self, height: u64, hash: HashOutput, accumulated_difficulty: u128) -> &mut Self {
+    pub fn set_best_block(
+        &mut self,
+        height: u64,
+        hash: HashOutput,
+        accumulated_difficulty: u128,
+        from: HashOutput,
```
sdbondi: I still think `from` does not give enough meaning, maybe `expected_prev_best_block` (I know it's long, but the reason for the parameter is clear) - then comment on this function

```rust
/// Sets the best block metadata. 
/// This operation will fail if the header hash was not previously added and if the `expected_prev_best_block` hash 
/// does not match the current best block. This ensures that a data race results in an error and is rolled back.
```

stringhandler (approved):

PR #2865, opened 2021-04-20
[Wallet] Update wallet gRPC server to send one-sided payments

philipr-za:
## Description
This PR adds an enum to the PaymentRecipient Protocol struct that is used with the Transfer method to allow the user to specify if a payment is made using the traditional Mimblewimble method or as a One-sided payment.

## How Has This Been Tested?
This was tested manually as the Cucumber tests seem broken in the TariScript branch currently.

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -69,6 +69,11 @@ message PaymentRecipient {
     uint64 amount = 2;
     uint64 fee_per_gram = 3;
     string message = 4;
+    enum PaymentType {
+        STANDARD_MIMBLEWIMBLE = 0;
+        ONE_SIDED = 1;
+    }
+    PaymentType payment_type = 5;
 }
 
 message TransferResponse {
--- a/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -3,6 +3,7 @@ use log::*;
 use tari_app_grpc::{
     conversions::naive_datetime_to_timestamp,
     tari_rpc::{
+        payment_recipient::PaymentType,
         wallet_server,
         CoinSplitRequest,
         CoinSplitResponse,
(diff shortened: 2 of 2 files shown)


hansieodendaal (approved):
LGTM

applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -69,6 +69,11 @@ message PaymentRecipient {
     uint64 amount = 2;
     uint64 fee_per_gram = 3;
     string message = 4;
+    enum PaymentType {
+        STANDARD_MIMBLEWIMBLE = 0;
```
sdbondi, line 73: Oh nice - seems like prost doesn't require enums to be named `PAYMENT_TYPE_ONE_SIDED` anymore

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -123,29 +124,49 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+            } else if payment_type == PaymentType::OneSided as i32 {
+                one_sided_transfers.push(async move {
+                    (
+                        address,
+                        transaction_service
+                            .send_one_sided_transaction(pk, amount.into(), fee_per_gram.into(), message)
+                            .await,
+                    )
+                });
+            }
```
sdbondi, line 161: I would probably have just combined the calls in one async block (simpler, no need to allocate any vecs)
```rust
 let transfers = recipients
            .into_iter()
            .map(|(address, pk, amount, fee_per_gram, message, payment_type)| {
                let mut transaction_service = self.get_transaction_service();
                async move {

   if payment_type == PaymentType::StandardMimblewimble as i32 {
                
                    (
                        address,
                        transaction_service
                            .send_transaction(pk, amount.into(), fee_per_gram.into(), message)
                            .await,
                    )
                
            } else if payment_type == PaymentType::OneSided as i32 {
                
                    (
                        address,
                        transaction_service
                            .send_one_sided_transaction(pk, amount.into(), fee_per_gram.into(), message)
                            .await,
                    )
                }
});
```

sdbondi (commented):
Looks good - one nit

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
philipr-za: I tried that first but compiler was adamant that the futures were of different types. Even though the two methods return the same types, something about the generator.

applications/tari_app_grpc/proto/wallet.proto
philipr-za: Oh didn't realize that would be a possible issue but ja seemed to generate just fine.

SWvheerden (approved):
LGTM

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
sdbondi: Are you sure? We arent creating two futures in each of the if-else branches we're creating one future as before that returns the same thing - the `if` is inside the one future

PR #2872, opened 2021-04-23
One-sided payment UTXO scanning 

SWvheerden:
## Description

This PR adds in the ability to scan one-sided payments from the blockchain. 
This refactored the recovery task to a new utxo scanner service. This service is used by the recovery tasks and the wallet to scan the blockchain. 
This PR also fixes the sent one-sided payments to include the correct rewinding factors.

## Motivation and Context

Wallets needs to be able to recover one-sided payments sent to miners for mining.

## Checklist:


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1264,6 +1295,50 @@ where TBackend: OutputManagerBackend + 'static
...
+
+    /// Attempt to scan and then rewind all of the given transaction outputs into unblinded outputs based on known
+    /// pubkeys
+    async fn scan_outputs(
+        &mut self,
+        outputs: Vec<TransactionOutput>,
+        height: u64,
+    ) -> Result<Vec<UnblindedOutput>, OutputManagerError>
+    {
+        let hashes: Vec<KnownScript> = self.resources.db.get_all_known_scripts().await?;
```
```suggestion
        let known_scripts: Vec<KnownScript> = self.resources.db.get_all_known_scripts().await?;
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1264,6 +1295,50 @@ where TBackend: OutputManagerBackend + 'static
...
     }
 
+    async fn add_known_script(&mut self, known_script: KnownScript) -> Result<(), OutputManagerError> {
+        debug!(target: LOG_TARGET, "Adding new script to output manager service");
+        Ok(self.resources.db.add_known_script(known_script).await?)
+    }
+
+    /// Attempt to scan and then rewind all of the given transaction outputs into unblinded outputs based on known
+    /// pubkeys
+    async fn scan_outputs(
```
stringhandler: This method is very specific to one-sided payments and should probably be named as such. E.g. `scan_outputs_for_one_sided_payments`

base_layer/wallet/src/tasks/utxo_scanning.rs
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    async fn get_chain_tip_header(
+        &self,
+        client: &mut BaseNodeSyncRpcClient,
+        peer: &NodeId,
+    ) -> Result<BlockHeader, WalletError>
+    {
+        let chain_metadata = client.get_chain_metadata().await.map_err(to_wallet_scanning_error)?;
+        if chain_metadata.effective_pruned_height > 0 {
```
stringhandler: Is this important? Why can't a pruned node be used

base_layer/wallet/src/tasks/utxo_scanning.rs
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    async fn get_start_utxo_mmr_pos(&self) -> Result<u64, WalletError> {
+        match self.mode {
+            UtxoScannerMode::Recovery => {
+                let previous_sync_height = self
+                    .get_metadata::<u64>(ScanningMetadataKey::RecoveryHeight)
+                    .await
+                    .ok()
+                    .flatten();
```
stringhandler: What's the reason for this `ok().flatten()` pattern? What are you trying to achieve?

base_layer/wallet/src/tasks/utxo_scanning.rs
```diff
@@ -0,0 +1,651 @@
...
+
+                if previous_sync_height.is_none() || previous_utxo_index.is_none() {
+                    // Set a value in here so that if the recovery fails on the genesis block the client will know a
+                    // recover was started. Important on Console wallet that otherwise makes this decision based on the
+                    // presence of the data file
+                    self.set_metadata(ScanningMetadataKey::RecoveryHeight, 0u64).await?;
+                    self.set_metadata(ScanningMetadataKey::RecoveryUtxoIndex, 0u64).await?;
+                }
+
+                Ok(previous_utxo_index.unwrap_or(0u64))
```
stringhandler: Why do you get the height and then not used it?

base_layer/wallet/src/tasks/utxo_scanning.rs
```diff
@@ -0,0 +1,651 @@
...
+        self.publish_event(UtxoScannerEvent::ConnectedToBaseNode(
+            peer.clone(),
+            latency.unwrap_or_default(),
+        ));
+
+        let timer = Instant::now();
+        let mut total_scanned = 0u64;
+        loop {
+            let start_index = self.get_start_utxo_mmr_pos().await?;
+            let tip_header = self.get_chain_tip_header(&mut client, &peer).await?;
```
stringhandler: Don't rely on the tip, use the last hash and height to know whether the base node has reorged away from our last point
```suggestion
            let (start_index, start_header_hash, start_header_height) = self.get_last_sync_point().await?;
            let header = self.get_chain_header(&mut client, &peer, start_header_height).await?;
            if header.hash() != start_header_hash {
             // chain has reorg-ed, move last_start_point back and check from there
             ....
             
```

base_layer/wallet/src/tasks/utxo_scanning.rs
stringhandler: See above comment, this method is probably not needed

common/config/presets/windows.toml
```diff
@@ -100,6 +100,8 @@ console_wallet_db_file = "wallet\\console-wallet.dat"
 # This is the timeout period that will be used to monitor TXO queries to the base node (default = 60). Larger values
 # are needed for wallets with many (>1000) TXOs to be validated.
 base_node_query_timeout = 120
+scan_for_utxo_interval = 43200
```
stringhandler, line 104: Maybe a comment to say how long this is

base_layer/wallet/src/tasks/utxo_scanning.rs
```diff
@@ -0,0 +1,651 @@
...
+        num_retries: usize,
+        retry_limit: usize,
+        error: String,
+    },
+    ScanningRoundFailed {
+        num_retries: usize,
+        retry_limit: usize,
+    },
+    /// Progress of the recovery process (current_block, current_chain_height)
+    Progress(u64, u64),
```
stringhandler: Why not use the struct syntax like the above variants?

stringhandler (commented):
Looks ok so far. I would recommend renaming KnownScript and related methods to be specific about one-sided payment scripts. The use case and logic is very specific to this, and KnownScript sounds like it is intended to be generic to accommodate future implementations, which it is very unlikely to do

base_layer/wallet/src/tasks/utxo_scanning.rs
SWvheerden: This is part of the original code. I think @sdbondi  should be able to answer this, he wrote this originally.

SWvheerden:
I am very much open to name suggestions, they are important. 
I like `KnownScript` as it portraits what's happening in the backend, the service is scanning UTXO's for script hashes it knowns how to claim. Although we are assuming it's for one-sided payment scripts.

base_layer/wallet/src/tasks/utxo_scanning.rs
SWvheerden: So `get_metadata ` returns an `Option<T>`, this calls `get_client_key_from_str` which returns an `Option<T>`, so this function in respect returns `Option<Option<T>>`

base_layer/wallet/src/tasks/utxo_scanning.rs
SWvheerden: Changed this to get the fork header, if any and work from there

PR #2874, opened 2021-04-26
[base-node] Refactor chain storage and rule out some target diff bugs

sdbondi:
## Description

- Adds `AchievedTargetDifficulty` that cannot be constructed unless achieved
  difficulty > target difficulty
- Header validator returns `AchievedTargetDifficulty`
- Refactor BlockHeaderAccumulatedDataBuilder to prevent possible bugs
  and to use `AchievedTargetDifficulty`
- Removes public fields from Chain(Block|Header)
- Chain(Block|Header) guarantee that the accumulated data hash matches
  (small cost)
- Remove in-memory cache of ChainMetadata
  - lmdb already is in-memory
  - block sync performance was similar
  - a stale value has carries a high risk of db corruption
- Rollback + error if duplicate orphans inserted (insert_orphan and
  insert_chained_orphan)
- `chain-storage` DeleteOrphan deletes accumulated orphan data
- Adds more reorg unit tests to blockchain database
- ~~Allow serde to (de)serialize reference counted fields.~~ (no longer needed)
  (https://serde.rs/feature-flags.html#-features-rc - the behaviour
  defined here is expected/correct)

The majority of the changed LOC are a result of making fields private
and other structural changes that can prevent bugs.

## Motivation and Context

Prevent future bugs by making it impossible (without changing the relevant public interfaces) to make mistakes that were previously easy to do.

## How Has This Been Tested?

Added a number of blockchain database unit tests. Manually tested on pruned/archival nodes.

## Checklist:

### DIFF
--- a/applications/tari_base_node/src/command_handler.rs
+++ b/applications/tari_base_node/src/command_handler.rs
@@ -769,7 +769,7 @@ impl CommandHandler {
                     println!("No headers found");
                     return;
                 },
-                Ok(h) => h.into_iter().map(|ch| ch.header).rev().collect::<Vec<_>>(),
+                Ok(h) => h.into_iter().map(|ch| ch.into_header()).rev().collect::<Vec<_>>(),
                 Err(err) => {
                     println!("Failed to retrieve headers: {:?}", err);
                     warn!(target: LOG_TARGET, "Error communicating with base node: {}", err,);
@@ -888,11 +888,11 @@ impl CommandHandler {
                     },
                 };
                 height -= 1;
-                if block.block().header.timestamp.as_u64() > period_ticker_end {
+                if block.header().timestamp.as_u64() > period_ticker_end {
                     print!("\x1B[{}D\x1B[K", (height + 1).to_string().chars().count());
                     continue;
                 };
-                while block.block().header.timestamp.as_u64() < period_ticker_start {
+                while block.header().timestamp.as_u64() < period_ticker_start {
                     results.push((
                         period_tx_count,
                         period_hash,
@@ -910,10 +910,10 @@ impl CommandHandler {
                 }
                 period_tx_count += block.block().body.kernels().len() - 1;
                 period_block_count += 1;
-                let st = if prev_block.block().header.timestamp.as_u64() >= block.block().header.timestamp.as_u64() {
+                let st = if prev_block.header().timestamp.as_u64() >= block.header().timestamp.as_u64() {
                     1.0
                 } else {
-                    (block.block().header.timestamp.as_u64() - prev_block.block().header.timestamp.as_u64()) as f64
+                    (block.header().timestamp.as_u64() - prev_block.header().timestamp.as_u64()) as f64
                 };
                 let diff = block.accumulated_data.target_difficulty.as_u64();
                 period_difficulty += diff;
@@ -970,31 +970,29 @@ impl CommandHandler {
                 let header = try_or_print!(db.fetch_chain_header(height).await);
 
                 // Optionally, filter out pow algos
-                if pow_algo.map(|algo| header.header.pow_algo() != algo).unwrap_or(false) {
+                if pow_algo.map(|algo| header.header().pow_algo() != algo).unwrap_or(false) {
                     continue;
                 }
 
                 let target_diff = try_or_print!(db.fetch_target_difficulties(prev_header.hash().clone()).await);
+                let pow_algo = header.header().pow_algo();
 
-                let min = consensus_rules
-                    .consensus_constants(height)
-                    .min_pow_difficulty(header.header.pow_algo());
-                let max = consensus_rules
-                    .consensus_constants(height)
-                    .max_pow_difficulty(header.header.pow_algo());
+                let min = consensus_rules.consensus_constants(height).min_pow_difficulty(pow_algo);
+                let max = consensus_rules.consensus_constants(height).max_pow_difficulty(pow_algo);
 
-                let calculated_target_difficulty = target_diff.get(header.header.pow_algo()).calculate(min, max);
-                let existing_target_difficulty = header.accumulated_data.target_difficulty;
-                let achieved = header.accumulated_data.achieved_difficulty;
-                let solve_time = header.header.timestamp.as_u64() as i64 - prev_header.header.timestamp.as_u64() as i64;
+                let calculated_target_difficulty = target_diff.get(pow_algo).calculate(min, max);
+                let existing_target_difficulty = header.accumulated_data().target_difficulty;
+                let achieved = header.accumulated_data().achieved_difficulty;
+                let solve_time =
+                    header.header().timestamp.as_u64() as i64 - prev_header.header().timestamp.as_u64() as i64;
                 let normalized_solve_time = cmp::min(
                     cmp::max(solve_time, 1) as u64,
                     consensus_rules
                         .consensus_constants(height)
-                        .get_difficulty_max_block_interval(header.header.pow_algo()),
+                        .get_difficulty_max_block_interval(pow_algo),
                 );
-                let acc_sha3 = header.accumulated_data.accumulated_blake_difficulty;
-                let acc_monero = header.accumulated_data.accumulated_monero_difficulty;
+                let acc_sha3 = header.accumulated_data().accumulated_blake_difficulty;
+                let acc_monero = header.accumulated_data().accumulated_monero_difficulty;
 
                 writeln!(
                     output,
@@ -1005,9 +1003,9 @@ impl CommandHandler {
                     calculated_target_difficulty.as_u64(),
                     solve_time,
                     normalized_solve_time,
-                    header.header.pow_algo(),
-                    chrono::DateTime::from(header.header.timestamp),
-                    target_diff.get(header.header.pow_algo()).len(),
+                    pow_algo,
+                    chrono::DateTime::from(header.header().timestamp),
+                    target_diff.get(pow_algo).len(),
                     acc_monero.as_u64(),
                     acc_sha3.as_u64(),
                 )
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -402,11 +402,11 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
             .map_err(BlockHeaderSyncError::ReceivedInvalidHeader)?;
         let num_new_headers = headers.len();
 
-        // We can trust that the header associated with this hash exists because block_hashes is data this node
-        // supplied. usize conversion overflow has already been checked above
-        let chain_split_hash = block_hashes[fork_hash_index as usize].clone();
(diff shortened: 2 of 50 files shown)


applications/tari_base_node/src/command_handler.rs
```diff
@@ -970,31 +970,29 @@ impl CommandHandler {
...
                 let normalized_solve_time = cmp::min(
                     cmp::max(solve_time, 1) as u64,
                     consensus_rules
                         .consensus_constants(height)
-                        .get_difficulty_max_block_interval(header.header.pow_algo()),
+                        .get_difficulty_max_block_interval(pow_algo),
                 );
-                let acc_sha3 = header.accumulated_data.accumulated_blake_difficulty;
-                let acc_monero = header.accumulated_data.accumulated_monero_difficulty;
+                let acc_sha3 = header.accumulated_data().accumulated_blake_difficulty;
```
SWvheerden, line 994: `accumulated_blake_difficulty`
We should really put in a todo to refactor this next reset.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -402,11 +402,11 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
             .map_err(BlockHeaderSyncError::ReceivedInvalidHeader)?;
         let num_new_headers = headers.len();
 
-        // We can trust that the header associated with this hash exists because block_hashes is data this node
+        // NOTE: We can trust that the header associated with this hash exists because block_hashes is data this node
```
SWvheerden: This sentence does not read right

base_layer/core/src/blocks/block_header.rs
```diff
@@ -134,12 +134,11 @@ impl BlockHeader {
     }
 
     /// Create a new block header using relevant data from the previous block. The height is incremented by one, the
-    /// previous block hash is set, and the timestamp is set to the current time and the proof of work is partially
-    /// initialized, although the `accumulated_difficulty_<algo>` stats are updated using the previous block's proof
-    /// of work information.
-    pub fn from_previous(prev: &BlockHeader) -> Result<BlockHeader, BlockHeaderValidationError> {
+    /// previous block hash is set, the timestamp is set to the current time, and the kernel/output mmr sizes are set to
+    /// the previous block. All other fields, including proof of work are set to defaults.
+    pub fn from_previous(prev: &BlockHeader) -> BlockHeader {
```
SWvheerden, line 139: Is there a reason we don't calculate the accumulated pow data here?
We have the previous block's accumulated pow data and its achieved pow data.

base_layer/core/src/chain_storage/accumulated_data.rs
```diff
@@ -182,84 +187,75 @@ impl<'de> Visitor<'de> for DeletedBitmapVisitor {
...
+        let achieved_target = self.current_achieved_target.ok_or_else(|| {
+            ChainStorageError::InvalidOperation("Current achieved difficulty not provided".to_string())
+        })?;
+
+        let (monero_diff, blake_diff) = match achieved_target.pow_algo() {
+            PowAlgorithm::Monero => (
+                previous_accum.accumulated_monero_difficulty + achieved_target.achieved(),
+                previous_accum.accumulated_blake_difficulty,
+            ),
+            PowAlgorithm::Blake => unimplemented!(),
```
SWvheerden, line 245: why do we still have this? Is this used?

base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -315,8 +313,8 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self
     }
 
-    pub fn insert_header(&mut self, header: BlockHeader, accum_data: BlockHeaderAccumulatedData) -> &mut Self {
-        self.transaction.insert_header(header, accum_data);
+    pub fn insert_header(&mut self, chain_header: ChainHeader) -> &mut Self {
```
SWvheerden:
```suggestion
    pub fn insert_chain_header(&mut self, chain_header: ChainHeader) -> &mut Self {
```
This will be more clear what's going on

SWvheerden (commented):
Looks good so far. Have not finished reviewing this thou.

applications/tari_base_node/src/command_handler.rs
sdbondi: Yeah agree - I'll add a "tagged comment" #testnetreset on the field. I suspect we could rename it now without breaking any serialised data, but havent tried.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: Could be better - updated it, basically it's saying that unwrap is safe because we know/control what's in block_hashes

base_layer/core/src/blocks/block_header.rs
sdbondi: We don't have the achieved here AFAICS, that happens after mining and this function is used to supply the block to mine. Though could be a good idea for Chain header maybe. Will leave that for another PR

base_layer/core/src/chain_storage/accumulated_data.rs
sdbondi: Because removing it breaks the db - it should have been removed last reset but wasn't - it's already marked with the #testnetreset tag for next time :)

base_layer/core/src/chain_storage/async_db.rs
sdbondi: Had to be renamed in a few places - but agree it's a little clearer, changed

base_layer/core/src/chain_storage/accumulated_data.rs
```diff
@@ -182,84 +187,81 @@ impl<'de> Visitor<'de> for DeletedBitmapVisitor {
...
-            .accumulated_monero_difficulty
-            .ok_or_else(|| ChainStorageError::InvalidOperation("difficulty not provided".to_string()))?;
+        let previous_accum = self.previous_accum;
+        let hash = self
+            .hash
+            .ok_or_else(|| ChainStorageError::InvalidOperation("hash not provided".to_string()))?;
+
+        if hash == self.previous_accum.hash {
+            return Err(ChainStorageError::InvalidOperation(
+                "Hash was set to the same hash that is contained in previous accumualted data".to_string(),
```
stringhandler: nit: spelling

base_layer/core/src/chain_storage/accumulated_data.rs
```diff
@@ -324,27 +327,112 @@ impl Display for ChainHeader {
...
+
+    #[inline]
+    pub fn accumulated_data(&self) -> &BlockHeaderAccumulatedData {
+        &self.accumulated_data
+    }
+
+    pub fn to_chain_header(&self) -> ChainHeader {
+        // NOTE: Panic is impossible, a ChainBlock cannot be constructed if inconsistencies between the header and
+        // accum data exist
+        ChainHeader::try_construct(self.block.header.clone(), self.accumulated_data.clone()).unwrap()
```
stringhandler, line 440: This may be impossible now, but it may not always be. The implementation of try_construct may change in future

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -217,15 +217,18 @@ where B: BlockchainBackend
     pub fn create_chain_header_if_valid(
```
stringhandler: This method seems to only be used in tests and should be deleted or moved to a test mod

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1756,105 +1732,109 @@ fn insert_orphan_and_find_new_tips<T: BlockchainBackend>(
...
-                res.extend(find_orphan_descendant_tips_of(
-                    db,
-                    &child.header,
-                    &accum_data,
-                    validator,
-                    txn,
-                )?);
+
+                let chain_header = ChainHeader::try_construct(child.header, accum_data)
+                    .expect("program correctness: accumulated data did not match the given block hash");
```
stringhandler: I think this should be propagated. There is still a small chance this will happen and it will be easier to find in the logs than a panic

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -473,32 +473,52 @@ impl LMDBDatabase {
...
+    ) -> Result<(), ChainStorageError>
+    {
+        if !lmdb_exists(txn, &self.orphans_db, header_hash.as_slice())? {
+            return Err(ChainStorageError::InvalidOperation(format!(
+                "set_accumulated_data_for_orphan: orphan {} does not exist",
+                header_hash.to_hex()
+            )));
+        }
+
+        lmdb_replace(
```
stringhandler:
```suggestion
        lmdb_insert(
```

stringhandler (dismissed):
Looks good

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Ok will do, though it may (I'll see what I can do) require duplicate code because it's used in unit and rust integration tests (public interface, cfg(test) wont work)

base_layer/core/src/chain_storage/accumulated_data.rs
sdbondi: I think that we should guarantee that with unit tests actually, I will add that test.

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: As it stands now, it is impossible because the child hash is set to the accumulated data hash (i.e. you would have to make a programming error/go out of your way to trigger this panic). That being said, if that happens the logs should make it pretty clear. Will change it to an error .

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -473,32 +473,53 @@ impl LMDBDatabase {
...
         Ok(())
     }
 
-    fn insert_orphan_block(&mut self, txn: &WriteTransaction<'_>, block: &Block) -> Result<(), ChainStorageError> {
+    fn insert_orphan_block(&self, txn: &WriteTransaction<'_>, block: &Block) -> Result<(), ChainStorageError> {
         let k = block.hash();
-        if lmdb_exists(txn, &self.orphans_db, k.as_slice())? {
-            return Ok(());
-        }
-
```
SWvheerden, line 492: we did we remove this check?

SWvheerden (dismissed):
LGTM

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: Yeah it now errors if the orphan has already been inserted and checks-for-existence were added where necessary (i.e. add_block)

PR #2878, opened 2021-04-28
Emission math audit

sdbondi:
## Description

Check for overflows with integer math in emission schedule and adds
tests.

Shift right on u64 will overflow if shifting 64 or more bits. The shift
is ultimately controlled by hard-coded consensus constants. This PR adds
a panic if the emission decay is set too high (i.e. >=64). In release, an
shift right overflow "wraps around", meaning that in the event that
a hard-coded value is set incorrectly/maliciously, it could cause an inflation bug.
This PR adds a check that will panic if the emission  decay will cause an overflow.

EDIT: Interestingly/weirdly rust has two behaviours in release for overflows depending on whether the shifted value is static/constant `123u8 >> 256 == 123u8` (wraps around, info preserved) or dynamic  `let mut n = 123u8; for _ in 0..256*3 {  n >>= 1; }; assert_eq!(n, 0);` (bits shifted into 0, original information lost)

## Motivation and Context

Consensus critical code check + cleanup of old emission code

https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=4be9af5dbe4db22518d3c763a4c60e2b

## How Has This Been Tested?

New unit tests

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4502,7 +4502,7 @@ dependencies = [
 
 [[package]]
 name = "tari_wallet_ffi"
-version = "0.16.27"
+version = "0.16.28"
 dependencies = [
  "chrono",
  "env_logger 0.7.1",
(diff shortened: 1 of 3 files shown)


SWvheerden (dismissed):
LGTM

CjS77 (dismissed):
Excellent. Thanks for this. Been on my radar to do for a long time.

stringhandler (approved):

PR #2879, opened 2021-04-29
Fix potential overflow and improve performance of n_leaves

sdbondi:
## Description

- Fix overflow when using usize::MAX as input to n_leaves
- Remove Vec allocation from n_leaves calculation
- Remove unused type conversion impl, if used could panic

Details:

- found by @pventuzelo
- found using manual code audit/review & fuzzing
- related code (n_leaves): https://github.com/tari-project/tari/blob/95499a7aea664407efc68d44aec0f97b6044ba86/base_layer/mmr/src/common.rs#L184-L186
- can also be triggered using leaf_index: https://github.com/tari-project/tari/blob/95499a7aea664407efc68d44aec0f97b6044ba86/base_layer/mmr/src/common.rs#L40-L42

## Motivation and Context

Fix potential integer overflow

## How Has This Been Tested?

Added overflow case to existing unit test

## Checklist:

### DIFF
--- a/base_layer/mmr/src/common.rs
+++ b/base_layer/mmr/src/common.rs
@@ -37,8 +37,11 @@ pub fn node_index(leaf_index: usize) -> usize {
 }
 
 /// Returns the leaf index derived from the MMR node index.
-pub fn leaf_index(node_index: usize) -> u32 {
-    n_leaves(node_index) as u32
+pub fn leaf_index(node_index: u32) -> u32 {
+    let n = checked_n_leaves(node_index as usize)
+        .expect("checked_n_leaves can only overflow for `usize::MAX` and that is not possible");
+    // Conversion is safe because n < node_index
+    n as u32
 }
 
 /// Is this position a leaf in the MMR?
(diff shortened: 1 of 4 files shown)


base_layer/mmr/src/common.rs
```diff
@@ -147,47 +150,42 @@ pub fn peak_map_height(mut pos: usize) -> (usize, usize) {
...
+/// Example: on input 5 returns (2 + 1 + 1) as mmr state before adding 5 was
 ///    2
 ///   / \
 ///  0   1   3   4
-pub fn peak_sizes_height(size: usize) -> (Vec<usize>, usize) {
+/// None is returned if the number of leaves exceeds the maximum value of a usize
+pub fn checked_n_leaves(size: usize) -> Option<usize> {
     if size == 0 {
-        return (vec![], 0);
+        return Some(0);
```
CjS77, line 172: I we know that None is only returned for MAX, add it as a special case here, and you don't have to do the bounds checks below

base_layer/mmr/src/common.rs
```diff
@@ -147,47 +150,42 @@ pub fn peak_map_height(mut pos: usize) -> (usize, usize) {
...
     }
     let mut peak_size = ALL_ONES >> size.leading_zeros();
-    let mut sizes = vec![];
+    let mut nleaves = 0usize;
     let mut size_left = size;
     while peak_size != 0 {
         if size_left >= peak_size {
-            sizes.push(peak_size);
+            nleaves += peak_size.checked_add(1)? / 2;
```
CjS77: If you really want to push performance, `n/2` ==> `n >> 1`

CjS77 (dismissed):
LGTM - Some small suggestions

base_layer/mmr/src/common.rs
sdbondi: I 100% do want to push performance 🏎️  - thanks

base_layer/mmr/src/common.rs
sdbondi: Good one - thanks

stringhandler (approved):

PR #2880, opened 2021-04-30
Fix Cucumber tests in `TariScript` branch

philipr-za:
## Description

This PR is not nearly as big as it looks. Most the lines are due to a cargo.lock and package-lock.json files.

This PR aims to fix the broken cucumber tests in the TariScript branch. The PR contains the following:

- Implement the TariScript logic for producing coin bases and transactions in the cucumber support javascript. The updates add a no_op script to UTXO’s.
- Removed the `height` from TariScript signatures and validation
- Update tari_crypto’s WASM interface to allow for clients to do private key arithmetic
- Update the grpc client used by the cucumber tests to support the new fields in the transaction protobuf
- Fixed a bug that causes panics when displaying an TransactionOutput with a malformed range proof
- Remove a number of duplicate Cucumber step file entries from a bad merge.

This PR requires the following updates to supporting repos:
- https://github.com/tari-project/tari-crypto/pull/41 adds requires WASM methods to the `tari-crypto` crate. This update will also require pushing an update to NPM. This was used locally by using `npm link` to use a locally built version of `tari-crypto`.
- https://github.com/tari-project/wallet-grpc-client/pull/15 updates the wallet grpc client used by the cucumber tests.

## How Has This Been Tested?
The cucumber tests themselves.

## Checklist:


applications/tari_app_grpc/src/conversions/aggregate_body.rs
```diff
@@ -51,9 +51,9 @@ impl TryFrom<grpc::AggregateBody> for AggregateBody {
     type Error = String;
 
     fn try_from(body: grpc::AggregateBody) -> Result<Self, Self::Error> {
-        let inputs = try_convert_all(body.inputs)?;
-        let outputs = try_convert_all(body.outputs)?;
-        let kernels = try_convert_all(body.kernels)?;
+        let inputs = try_convert_all(body.inputs).map_err(|err: String| format!("inputs {}", err.to_string()))?;
+        let outputs = try_convert_all(body.outputs).map_err(|err: String| format!("outputs {}", err.to_string()))?;
+        let kernels = try_convert_all(body.kernels).map_err(|err: String| format!("kernels {}", err.to_string()))?;
```
sdbondi, line 56: nit:  no need for the additional string conversion 
```suggestion
        let inputs = try_convert_all(body.inputs).map_err(|err| format!("inputs {}", err))?;
        let outputs = try_convert_all(body.outputs).map_err(|err| format!("outputs {}", err))?;
        let kernels = try_convert_all(body.kernels).map_err(|err| format!("kernels {}", err))?;
```

applications/tari_app_grpc/src/conversions/new_block_template.rs
```diff
@@ -71,10 +71,10 @@ impl TryFrom<grpc::NewBlockTemplate> for NewBlockTemplate {
...
     fn try_from(block: grpc::NewBlockTemplate) -> Result<Self, Self::Error> {
         let header = block.header.clone().ok_or_else(|| "No header provided".to_string())?;
-        let total_kernel_offset =
-            BlindingFactor::from_bytes(&header.total_kernel_offset).map_err(|err| err.to_string())?;
-        let total_script_offset =
-            BlindingFactor::from_bytes(&header.total_script_offset).map_err(|err| err.to_string())?;
+        let total_kernel_offset = BlindingFactor::from_bytes(&header.total_kernel_offset)
+            .map_err(|err| format!("total_kernel_offset {}", err.to_string()))?;
+        let total_script_offset = BlindingFactor::from_bytes(&header.total_script_offset)
+            .map_err(|err| format!("total_script_offset {}", err.to_string()))?;
```
sdbondi, line 77: nit: Same here, no need to allocate an additional string on the heap when using format!

integration_tests/helpers/transactionBuilder.js
```diff
@@ -28,15 +28,82 @@ class TransactionBuilder {
     return Buffer.from(final).toString("hex");
   }
 
+  buildScriptChallenge(publicNonce, script, input_data, height) {
+    var KEY = null; // optional key
+    var OUTPUT_LENGTH = 32; // bytes
+    var context = blake2bInit(OUTPUT_LENGTH, KEY);
```
sdbondi, line 34: nit: `var` [should never really be used in modern JS](https://javascript.info/var) - `const` is similar to `let` in rust (a little less restrictive, state can still change internal to the value), and `let` is similar to `let mut` in rust.
```suggestion
    const KEY = null; // optional key
    const OUTPUT_LENGTH = 32; // bytes
    const context = blake2bInit(OUTPUT_LENGTH, KEY); 
```

integration_tests/helpers/transactionBuilder.js
```diff
@@ -28,15 +28,82 @@ class TransactionBuilder {
...
+    blake2bUpdate(context, buff);
+    blake2bUpdate(context, script);
+    blake2bUpdate(context, input_data);
+    // blake2bUpdate(context, height);
+    let final = blake2bFinal(context);
+    return Buffer.from(final).toString("hex");
+  }
+
+  hashOutput(features, commitment, script_hash, script_offset_public_key) {
+    var KEY = null; // optional key
```
sdbondi, line 45: nit `s/var/const`

sdbondi (approved):
Looks good - few nits

PR #2884, opened 2021-05-04
Remove "end of time" panic in emission

sdbondi:
## Description

Removed `Option::expect` when total supply overflows `u64`. In this
case, the `EmissionRate` iterator considers itself "finished" and so
returns None as per the Iterator contract.

Removes the `Emission` trait, although nothing incorrect about using a
trait in this case, it did not provide much of an adantage (decoupling)
over using the concrete type. It also hid the Iterator impl, that should
have been used in `get_tokens_in_circulation`

## Motivation and Context

Remove a panic which is not impossible. The (working) policy for panics is never to panic if a panic is possible (no matter how unlikely) in normal program operation.  

## How Has This Been Tested?

Existing emission tests updated. Synced base node from scratch

## Checklist:


sdbondi:
Closing in favour of PR to tari-script branch

PR #2886, opened 2021-05-04
Remove "end of time" panic in emission

sdbondi:
## Description

Removed `Option::expect` when total supply overflows `u64`. In this
case, the `EmissionRate` iterator considers itself "finished" and so
returns None as per the Iterator contract.

Removes the `Emission` trait, although nothing incorrect about using a
trait in this case, it did not provide much of an advantage (decoupling)
over using the concrete type. It also hid the Iterator impl, that should
have been used in `get_tokens_in_circulation`

Merging into tari-script branch

## Motivation and Context

Remove a panic which is not impossible. The (working) policy for panics is never to panic if a panic is possible (no matter how unlikely) in normal program operation.  

## How Has This Been Tested?

Existing emission tests updated. ~~Pruned synced base node from scratch~~ currently not possible on current tari-script chain

## Checklist:

### DIFF
--- a/base_layer/core/src/consensus/emission.rs
+++ b/base_layer/core/src/consensus/emission.rs
@@ -21,12 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::transactions::tari_amount::MicroTari;
-use num::pow;
-
-pub trait Emission {
-    fn block_reward(&self, height: u64) -> MicroTari;
-    fn supply_at_block(&self, height: u64) -> MicroTari;
-}
+use std::cmp;
 
 /// The Tari emission schedule. The emission schedule determines how much Tari is mined as a block reward at every
 /// block.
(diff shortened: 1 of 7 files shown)


philipr-za (approved):

base_layer/core/src/consensus/emission.rs
```diff
@@ -59,8 +54,17 @@ impl EmissionSchedule {
...
+    ///
+    /// ## Panics
+    ///
+    /// The shift right operation will overflow if shifting more than 63 bits. `new` will panic if any of the decay
+    /// values are greater than or equal to 64.
     pub fn new(initial: MicroTari, decay: &'static [u64], tail: MicroTari) -> EmissionSchedule {
+        assert!(
+            decay.iter().all(|i| *i < 64),
+            "Decay value would overflow. All `decay` values must be less than 64"
+        );
```
SWvheerden, line 67: Won't this only work in debug/test code and not production?

Perhaps the function should be swapped to return a result rather?

SWvheerden (commented):
Looks good

base_layer/core/src/consensus/emission.rs
sdbondi: Too late this has already been merged into dev 🤣  (just appears in this diff because this is merging to tari-script)

Assert will panic in release too, however since the values for decay are hard coded (importantly, NOT dynamic) consensus values so i think it's ok to do this for the sake of keeping the function more ergonimic/creating an error just for the "programmer hard coded the wrong values" case

PR #2887, opened 2021-05-04
Reset tari script gen block

SWvheerden:
## Description

We removed the height from the signature check for the time being from the transactions. This caused some block to fail validation if you try to sync. 
This PR will reset the gen block so that we can start a new chain on tari script. 

## Checklist:


philipr-za (approved):
LGTM

hansieodendaal (approved):
LGTM - have a new chain running on this

PR #2888, opened 2021-05-04
Consolidate config files into one

hansieodendaal:
## Description
- Consolidated all the three configuration files that were in the code base into one file, usable by all operating systems.
- Removed some deprecated entries in the config file.
- Updated the installation files/installers to use the new config file.

## Motivation and Context
Having three files where configuration options were managed created lots of confusion, and also resulted in many testnet wallets not connecting to the seed nodes due to having the wrong information.

## How Has This Been Tested?
Tested in Windows by running the exe files by itself.
Created a Windows installer and tested all shortcuts are working after the installation.

## Checklist:


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #2894, opened 2021-05-04
[Wallet] Derive Comms Secret Key from the Master Seed key

philipr-za:
## Description
This PR implements the derivation of the Comms private key in the wallet from the Master Seed Key that is used as the seed for the key managers used for the UTXO’s on the wall.

Previously the Comms Secret Key was generated by the client applications and it was stored in the Wallet Database. This key was also used to determine if the database was encrypted and if the provided passphrase is correct or not by storing the public key in the database and seeing if the stored private key corresponds to the permitted public key after decryption.

This PR removes the Comms Secret and Public keys from the wallet database and replaces them with the Master Seed Key which becomes the single secret key from which the Comms Secret Key, UTXO Spending Keys and Rewind keys are derived. The Comms Secret key is now derived when the wallet is started based on the persisted Master Secret Key. On the first run the wallet will generate the Master Secret Key and store it. The Output Manager Service used to generate and store this key but now it will accept it from the Wallet and use it to initialise its Key Managers. 

The Console Wallet and Wallet FFI is updated to account for this change.

## Motivation and Context
This means that a wallet can be fully recovered using the seed words, including its network identity.

## How Has This Been Tested?
Tests updated and an OMS test added to check what happens if a Master Secret Key does not correspond to the previous ones used to initialize the OMS Key Managers

## Checklist:


base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -533,6 +403,107 @@ impl WalletBackend for WalletSqliteDatabase {
...
+                        target: LOG_TARGET,
+                        "Cipher is provided but Master Secret Key is not encrypted"
+                    );
+                    return Err(WalletStorageError::InvalidEncryptionCipher);
+                }
+                sk
+            },
+            Err(_) => {
+                // This means the secret key was encrypted. Try decrypt
+                if let Some(cipher_inner) = cipher.clone() {
```
sdbondi, line 443: nit: 
```suggestion
                if let Some(cipher_inner) = cipher.as_ref() {
```

or pass in a ref `cipher: Option<&Aes256Gcm>` if you really want to save clones

base_layer/wallet/src/storage/sqlite_db.rs
```diff
@@ -533,6 +403,107 @@ impl WalletBackend for WalletSqliteDatabase {
...
+            if public_key != db_master_public_key {
+                if cipher.is_some() {
+                    error!(
+                        target: LOG_TARGET,
+                        "Cipher is provided but does not decrypt the stored Master Secret Key that produces the \
+                         stored Comms Public Key."
+                    );
+                    return Err(WalletStorageError::InvalidEncryptionCipher);
+                } else {
+                    // If the db is not encypted then update the stored public key to keep it in sync.
```
sdbondi:
```suggestion
                    // If the db is not encrypted then update the stored public key to keep it in sync.
```

sdbondi (approved):
Looks good - not tested

SWvheerden (approved):
Looks good

philipr-za:
Second commit contains a number of test fixes and some updates based on feedback from @kukabi about how the clients managed the Comms Secret Key. Because this is now completely handled in the backed we can remove the `comms_config_set_secret_key` function. It was also confirmed that the backend is fully responsible for managing the TorID so the old code that provided some migration support for the client to manage the TorID in the past has been removed.

StriderDM (approved):
+1

PR #2896, opened 2021-05-06
Fix wallet recovery

SWvheerden:
## Description

Wallet recovery atm in time can only recover non coinbase outputs. This PR fixes the coinbase UTXO importing. This will allow the wallet to recover coinbase UTXOs.

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -69,6 +69,7 @@ use tari_core::{
 use tari_crypto::{
     keys::{PublicKey as PublicKeyTrait, SecretKey as SecretKeyTrait},
     range_proof::REWIND_USER_MESSAGE_LENGTH,
+    tari_utilities::hex::Hex,
 };
 use tari_key_manager::{
     key_manager::KeyManager,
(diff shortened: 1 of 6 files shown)


hansieodendaal (approved):
LGTM

*Nit:* Maybe we need a unit test that also test importing UTXOs with non-default features?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1132,9 +1133,24 @@ where TBackend: OutputManagerBackend + 'static
...
                     .ok()
+                    .map(|v| (v, output.features))
+            })
+            .map(|(output, features)| {
+                UnblindedOutput::new(output.committed_value, output.blinding_factor, Some(features))
             })
-            .map(|output| UnblindedOutput::new(output.committed_value, output.blinding_factor, None))
             .collect();
+        for output in &rewound_outputs {
+            trace!(
```
stringhandler, line 1143: This could be `info` or `debug` even

stringhandler (approved):

delta1 (changes requested):
There's a test for `import_utxo` in `base_layer/wallet/tests/wallet/mod.rs` which should be updated to cover this change.

SWvheerden:
That test, would not catch the problem. I am updating PR  #2872  to include these changes as they have changed

PR #2897, opened 2021-05-06
Replace custom ChaCha20 cipher with RustCrypto ChaCha20 implementation

philipr-za:
## Description
Fuzzing revealed that the custom ChaCha20 implementation in `tari_utilities` was not correct so this PR migrates the DHT encryption to use the audited RustCrypto ChaCha20 encryption. 

Had to bump the toolchain version for this update so updated to the most current version of nightly. The only update needed was adding a new feature flag #![feature(min_type_alias_impl_trait)] which is a new flag for the same issue we already use (i.e. #![feature(type_alias_impl_trait)])

The new nightly toolchain contained an update to the linting used by rustfmt so I have included a second commit that is just new formatting to make this easier to review.

## How Has This Been Tested?
Tests updated

## Checklist:


stringhandler (approved):

comms/dht/src/crypt.rs
```diff
@@ -20,29 +20,55 @@
...
+use std::mem::size_of;
 
 pub fn generate_ecdh_secret<PK>(secret_key: &PK::K, public_key: &PK) -> PK
 where PK: PublicKey + DiffieHellmanSharedSecret<PK = PK> {
     PK::shared_secret(secret_key, public_key)
 }
 
-pub fn decrypt(cipher_key: &CommsPublicKey, cipher_text: &[u8]) -> Result<Vec<u8>, CipherError> {
-    ChaCha20::open_with_integral_nonce(&cipher_text.to_vec(), cipher_key.as_bytes())
+pub fn decrypt(cipher_key: &CommsPublicKey, cipher_text: &[u8]) -> Result<Vec<u8>, DhtOutboundError> {
```
CjS77, line 43: Is decypting with a pubkey correct? How is the en- / decrpytion used in the comms? If it's all with pubkeys, then the encryption is likely meaningless.

comms/dht/src/crypt.rs
```diff
@@ -63,10 +89,10 @@ mod test {
     fn decrypt_fn() {
         let key = CommsPublicKey::default();
         let cipher_text =
-            from_hex("7ecafb4c0a88325c984517fca1c529b3083e9976290a50c43ff90b2ccb361aeaabfaf680e744b96fc3649a447b")
+            from_hex("24bf9e698e14938e93c09e432274af7c143f8fb831f344f244ef02ca78a07ddc28b46fec536a0ca5c04737a604")
```
CjS77, line 99: This is odd, I thought only edge cases were broken on the handrolled chacha. If so, then this should still work with the same input and this suggests that the API has changed significantly, meaning other parts of comms could break.

comms/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 #![recursion_limit = "512"]
 // Allow `type Future = impl Future`
 #![feature(type_alias_impl_trait)]
+#![feature(min_type_alias_impl_trait)]
```
CjS77, line 20: This feature [may not be stabilised for a while](https://github.com/rust-lang/rust/issues/63063#issuecomment-795682305) so you might be hampering the effort to move Tari to stable Rust.

CjS77 (commented):
A quick overview. I have a few comments and questions.

comms/dht/src/crypt.rs
philipr-za: Its a Diffie-Hellman key

comms/src/lib.rs
philipr-za: It is the same issue the feature above which is critical to large portions of the code, they share the same tracking Issue in the Rust codebase.

comms/dht/src/crypt.rs
stringhandler: The type here is confusing. Perhaps just a different type?

CjS77: DH keys are typically secret keys. Yeah, it's just confusing.

philipr-za: Would need to update the way its implemented in the `tari_crypto` crate. 
https://github.com/tari-project/tari-crypto/blob/main/src/keys.rs#L78

CjS77: Yes, it's incorrect in tari_crypto. @hansieodendaal picked that up a few weeks ago and was going to push a PR changing it.

comms/dht/src/crypt.rs
philipr-za: So looked further into this and I cannot actually get the RustCrypto output to agree with the custom implementation or with the test vectors defined in https://datatracker.ietf.org/doc/html/rfc8439#section-2.4.1. However, I checked the output against the Python Cryptodome library and the RustCrypto and Cryptodome implementations agree 🤷‍♂️

stringhandler (approved):

philipr-za:
Updating the rust toolchain in this PR resulted in the rustfmt and clippy being updated so most of the PR are those updates. Just for clarity the only file that was actually updated in terms of logic is noted below.

comms/dht/src/crypt.rs
```diff
@@ -20,29 +20,62 @@
...
 use tari_crypto::{
     keys::{DiffieHellmanSharedSecret, PublicKey},
-    tari_utilities::{
-        ciphers::{
-            chacha20::ChaCha20,
-            cipher::{Cipher, CipherError},
-        },
-        ByteArray,
-    },
+    tari_utilities::ByteArray,
```
philipr-za, line 35: This is the file that was modified in this PR.

PR #2898, opened 2021-05-06
Re-implemented methods for block explorer

StriderDM:
## Description
Re-implemented `fetch_block_with_kernel`
Re-implemented `fetch_block_with_utxo`

## Motivation and Context
Panic occurs when block explorer attempts to call `fetch_block_with_kernel` method.

## How Has This Been Tested?
cargo test --all

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -63,7 +63,7 @@ use std::{
     time::Instant,
 };
 use tari_common_types::{chain_metadata::ChainMetadata, types::BlockHash};
-use tari_crypto::tari_utilities::{hex::Hex, Hashable};
+use tari_crypto::tari_utilities::{hex::Hex, ByteArray, Hashable};
 use tari_mmr::{MerkleMountainRange, MutableMmr};
 use uint::static_assertions::_core::ops::RangeBounds;
 
@@ -198,7 +198,7 @@ where B: BlockchainBackend
                 Ok(_) => info!(target: LOG_TARGET, "Orphan database cleaned out at startup.",),
                 Err(e) => warn!(
                     target: LOG_TARGET,
-                    "Orphan database could not be cleaned out at startup: ({}).", e
+                    "Orphan database could not be cleaned out at startup: ({:?}).", e
                 ),
             }
         }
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -319,7 +319,7 @@ where T: BlockchainBackend + 'static
                         target: LOG_TARGET,
                         "A peer has requested a block with hash {}", hash_hex
                     );
-                    match self.blockchain_db.fetch_block_with_stxo(hash).await {
+                    match self.blockchain_db.fetch_block_with_transaction_output(hash).await {
```
SWvheerden: Is not reproducing the same results here, is this intentional? 
STXO != TXO,  TXO = STXO || UTXO

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -343,7 +343,7 @@ where T: BlockchainBackend + 'static
                         target: LOG_TARGET,
                         "A peer has requested a block with hash {}", hash_hex,
                     );
-                    match self.blockchain_db.fetch_block_with_utxo(hash).await {
+                    match self.blockchain_db.fetch_block_with_transaction_output(hash).await {
```
SWvheerden: Is not reproducing the same results here, is this intentional? 
UTXO != TXO,  TXO = STXO || UTXO

SWvheerden (changes requested):
Replacing these methods change their meaning and outcome. 
We need to either change the fn names is this is the intended design or we need to properly implement these. 

STXO = Spend Tansactions Output. 
UTXO = Unspent Transactional Output

If you only want one blockchain_db method that's fine, but upstream where you request the STXO, you need to load up the header and get the spent bitmap to see if its unspent or not. 


If you want to combine `fetch_block_with_utxo` and `fetch_block_with_stxo` you need to return potentially 2 blocks. If a TXO was spent, it has a created-in block, and a spent-in block.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1212,114 +1209,42 @@ fn fetch_blocks<T: BlockchainBackend>(
...
+            field: "Excess sig".to_string(),
+            value: excess_sig.get_signature().to_hex(),
+        }),
+    }
 }
 
-fn fetch_block_with_stxo<T: BlockchainBackend>(
-    _db: &T,
-    _commitment: Commitment,
+fn fetch_block_with_transaction_output<T: BlockchainBackend>(
```
SWvheerden: This function only fetches the block the UTXO was created in, not spent

StriderDM: Thanks for clarifying, will fix.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1212,114 +1212,74 @@ fn fetch_blocks<T: BlockchainBackend>(
...
+            value: tip.to_hex(),
+        })?;
+    match db.fetch_output(&commitment.to_vec()) {
+        Ok(output) => match output {
+            Some((_output, leaf)) => {
+                // check if output is spent as of latest block
+                if !block_data.deleted().contains(leaf) {
+                    return Ok(None);
+                }
+                fetch_block_by_hash(db, tip)
```
SWvheerden: This function will not do what it's supposed to do. 
What this does is return the tip block, if the utxo was spent.

You need to do the following:

1.  **Get the UTXO and leaf index** If this exists, continue, else exit
2.  **Get the tip and see if it's deleted** If it is, continue, else exit
3.  **Loopback through the block accumulated data and check when it was deleted** The last block with that information in it, will be the block the TXO was spent in

I don't think we have an exact lookup for this currently so we will have to search for this.

StriderDM: Thanks @SWvheerden.

As per out chat I have reverted this method back to unimplemented for the time being since there doesn't seem to be an instance of it being used.

I think you're right in that we don't have an exact lookup for this currently.

SWvheerden (approved):
LGTM

stringhandler (commented):
Merging this because there's back and forth with no real progress. fetch_block_stxo will still fail as it's unimplemented, but is not used anywhere

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1212,87 +1212,42 @@ fn fetch_blocks<T: BlockchainBackend>(
...
-    // }
-    // Err(ChainStorageError::ValueNotFound {
-    //     entity: "Utxo".to_string(),
-    //     field: "Commitment".to_string(),
-    //     value: commitment.to_hex(),
-    // })
+    match db.fetch_output(&commitment.to_vec()) {
+        Ok(output) => match output {
+            Some((_output, leaf)) => {
+                let header = db.fetch_header_containing_utxo_mmr(leaf as u64)?;
```
stringhandler, line 1240: If you look at the implementation, an output is stored using the header hash. That could have also been used to retrieve the hash

PR #2899, opened 2021-05-06
Fix flaky tests on tari_script

SWvheerden:
**transaction_service::transaction_protocols::tx_broadcast_protocol_submit_success**

- Upped the timeouts
- Added sleep time after each update tx to ensure db has enough time to write changes, before being asked form them.

**output_manager_service::service::coin_split_with_change_sqlite_db**

- Scanned for both values of the coinsplit as the check only cares about the rewind, and not the value

## Checklist:

### DIFF
--- a/base_layer/wallet/tests/transaction_service/transaction_protocols.rs
+++ b/base_layer/wallet/tests/transaction_service/transaction_protocols.rs
@@ -64,7 +64,7 @@ use tari_wallet::{
     transaction_service::{
         config::TransactionServiceConfig,
         error::TransactionServiceError,
-        handle::{TransactionEvent, TransactionEventSender},
+        handle::{TransactionEvent, TransactionEventReceiver, TransactionEventSender},
         protocols::{
             transaction_broadcast_protocol::TransactionBroadcastProtocol,
             transaction_validation_protocol::TransactionValidationProtocol,
@@ -100,6 +100,7 @@ pub async fn setup(
     broadcast::Sender<Duration>,
     Shutdown,
     TempDir,
+    TransactionEventReceiver,
 ) {
     let client_node_identity = build_node_identity(PeerFeatures::COMMUNICATION_NODE);
     let server_node_identity = build_node_identity(PeerFeatures::COMMUNICATION_NODE);
@@ -143,7 +144,8 @@ pub async fn setup(
     let outbound_mock_state = mock_outbound_service.get_state();
     task::spawn(mock_outbound_service.run());
 
-    let (ts_event_publisher, _): (TransactionEventSender, _) = broadcast::channel(200);
+    let (ts_event_publisher, ts_event_receiver): (TransactionEventSender, TransactionEventReceiver) =
+        broadcast::channel(200);
 
     let shutdown = Shutdown::new();
 
@@ -175,6 +177,7 @@ pub async fn setup(
         timeout_update_publisher,
         shutdown,
         temp_dir,
+        ts_event_receiver,
     )
 }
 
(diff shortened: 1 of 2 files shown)


base_layer/wallet/tests/transaction_service/transaction_protocols.rs
```diff
@@ -283,9 +283,10 @@ async fn tx_broadcast_protocol_submit_success() {
 
     // Wait for 2 queries
     let _ = rpc_service_state
-        .wait_pop_submit_transaction_calls(4, Duration::from_secs(5))
+        .wait_pop_submit_transaction_calls(4, Duration::from_secs(6))
         .await
         .unwrap();
+    tokio::time::delay_for(Duration::from_secs(1)).await;
```
philipr-za: I don't think these delays after the  `wait_pop_submit_transaction_calls` do anything useful, the `wait_pop_submit_transaction_calls` are already waiting?

SWvheerden: So what happens is that call to read the database gets in before the call to write the database. 
The Tx manager already has the info, but it has not yet saved it to the db. That's why this small sleep solves this. When it fails you can see its because the read tx got its read lock before the write tx could get it.

philipr-za: I don't see how this solves the problem you are stating, the two statements you are putting the delay between only change the response returned to the transaction protocol when it makes a query, it has nothing to do with a database? The transaction protocol is running in its own task so this delay has no effect on it. The thing this is delaying between is changing the response the RPC mock gives to the transaction protocol when it makes its RPC request. By adding this delay you actually make it that the transaction protocol will probably put in a second request before the response type changes (the protocol is set to poll every 1 second).

SWvheerden: I think this time_delay does not help and might be redundant. The next ones should give time for the DB to write its changes. 

But after this discussion, it's better to move to an event-based fix. I have added an event listen here to wait for the db write.

base_layer/wallet/tests/transaction_service/transaction_protocols.rs
```diff
@@ -349,7 +352,17 @@ async fn tx_broadcast_protocol_submit_success() {
         .wait_pop_transaction_query_calls(1, Duration::from_secs(5))
         .await
         .unwrap();
-
+    // lets wait for the transaction service event to notify us of a confirmed tx
+    // We need to do this to ensure that the wallet db has been updated to "Mined"
+    while let Some(v) = transaction_event_receiver.next().await {
+        let event = v.unwrap();
```
philipr-za, line 357: You could just use pattern matching here instead of this unwrap.

base_layer/wallet/tests/transaction_service/transaction_protocols.rs
```diff
@@ -349,7 +352,17 @@ async fn tx_broadcast_protocol_submit_success() {
         .wait_pop_transaction_query_calls(1, Duration::from_secs(5))
         .await
         .unwrap();
-
+    // lets wait for the transaction service event to notify us of a confirmed tx
+    // We need to do this to ensure that the wallet db has been updated to "Mined"
+    while let Some(v) = transaction_event_receiver.next().await {
```
philipr-za, line 356: This should be fine though I am a little wary of a potentially infinite loop in a test. If something breaks then this test might never complete so I tend to add a timeout in case but it shouldn't happen so 🤷‍♂️

philipr-za (approved):

PR #2900, opened 2021-05-07
Fix cucumber tests and add eslint

stringhandler:
## Description

- Fix cucumber tests
- Add ESLint to integration tests
- Add ESLint to CI run

## Motivation and Context

To fix the tests.
Added ESLint to add standards for javascript

## Checklist:

* [] I have squashed my commits into a single commit.


delta1:
Nice definitely needed

.circleci/config.yml
```diff
@@ -92,6 +92,9 @@ commands:
      - run:
          name: Check formatting
          command: cd integration_tests && npm run check-fmt
+     - run:
+         name: Check eslint
+         command: cd integration_tests && npx eslint .
```
delta1: any reason to not add eslint to package.json with a script there?

integration_tests/cucumber.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  default: "",
-  critical: "--format @cucumber/pretty-formatter --tags @critical",
-};
+  default: '',
+  critical: '--format @cucumber/pretty-formatter --tags @critical'
+}
```
delta1: I think there's a conflict between prettier and eslint here (single quotes vs double quotes, and semi-colons)

delta1 (commented):
I think we just need to resolve differences between prettier and eslint

.circleci/config.yml
stringhandler: lol, because my js is not great... that is the reason

PR #2901, opened 2021-05-07
Update all Import UTXOs functions to correctly carry options

SWvheerden:
## Description

Import utxo fn use defaults down the line, this will cause any UTXO that's inserted into the wallet to be invalidated due to mismatched scripts, features, etc. 

Requires #2872 

## Motivation and Context

There is a lot of use defaults for the import utxo functions, I removed all of these and pointed them to the caller to let the caller decide if it should be default or not. 

## How Has This Been Tested?

Added a unit test to confirm that import utxo is importing the correct utxo. 

## Checklist:

### DIFF
--- a/base_layer/wallet/src/transaction_service/storage/database.rs
+++ b/base_layer/wallet/src/transaction_service/storage/database.rs
@@ -604,6 +604,7 @@ where T: TransactionBackend + 'static
         source_public_key: CommsPublicKey,
         comms_public_key: CommsPublicKey,
         message: String,
+        maturity: Option<u64>,
     ) -> Result<(), TransactionStorageError> {
         let transaction = CompletedTransaction::new(
             tx_id,
@@ -622,7 +623,7 @@ where T: TransactionBackend + 'static
             message,
             Utc::now().naive_utc(),
             TransactionDirection::Inbound,
-            None,
+            maturity,
         );
 
         let db_clone = self.db.clone();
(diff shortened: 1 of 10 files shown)


stringhandler:
Seems like there is much more to this PR than the description mentions. Can you please describe in detail what it does and what the motivation for it is

SWvheerden:
> Seems like there is much more to this PR than the description mentions. Can you please describe in detail what it does and what the motivation for it is

It needed a rebase, this PR is now much smaller

base_layer/wallet/src/transaction_service/storage/database.rs
```diff
@@ -604,7 +604,9 @@ where T: TransactionBackend + 'static
         source_public_key: CommsPublicKey,
         comms_public_key: CommsPublicKey,
         message: String,
-    ) -> Result<(), TransactionStorageError> {
+        coinbase_block_height: Option<u64>,
```
stringhandler: Why the different name here? It is lock_height in the others. "coinbase_block_height" is not a commonly used term, is it the height of the coinbase, or is it the time it matures?

base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
```diff
@@ -675,7 +675,12 @@ where TBackend: WalletBackend + 'static
         let tx_id = self
             .resources
             .transaction_service
-            .import_utxo(unblinded_output.value, source_public_key.clone(), message)
+            .import_utxo(
+                unblinded_output.value,
+                source_public_key.clone(),
+                message,
+                Some(unblinded_output.features.maturity),
```
stringhandler, line 678: Here it is called `maturity`. It's much less confusing if you use the same term throughout

stringhandler (commented):
The mix of names/terms in this PR are going to make it confusing. Please choose one (I'd suggest maturity) and use it throughout

stringhandler (approved):

PR #2911, opened 2021-05-11
Remove broken or flaky tests from CI

stringhandler:
## Description

Remove cucumber tests that fail often from CI so that we can identify tests that break in the future. Also fixed a compile error

## Motivation and Context

Currently, the integration tests step fails, making it hard to see if the current PR had broken something

## How Has This Been Tested?

run-tests.sh

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,10 @@ commands:
           command: cargo build --release --bin tari_merge_mining_proxy
       - run:
           name: Run cucumber scenarios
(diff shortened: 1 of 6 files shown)


delta1:
looks good, `As a wallet I want to submit transfers to myself` also fails often on CI but passes locally

delta1:
Almost there 

image

philipr-za (approved):
Green!

PR #2914, opened 2021-05-12
[MMProxy] Connection Limiting

StriderDM:
## Description
Limit number of idle connections for reqwest::Client
Used cheap clones of original connections for grpc based connections.

## Motivation and Context
Appears to solve "too many open files" problem

## How Has This Been Tested?
Tested with multiple xmrig clients mining against the merge mining proxy over a few hours. Current `ulimit -n` value is 2560. 

## Checklist:

### DIFF
--- a/applications/tari_merge_mining_proxy/src/proxy.rs
+++ b/applications/tari_merge_mining_proxy/src/proxy.rs
@@ -26,7 +26,6 @@ use crate::{
     error::MmProxyError,
 };
 use bytes::Bytes;
-use futures::TryFutureExt;
 use hyper::{service::Service, Body, Method, Request, Response, StatusCode, Uri};
 use json::json;
 use jsonrpc::error::StandardError;
@@ -37,7 +36,6 @@ use std::{
     cmp::min,
     convert::TryFrom,
     future::Future,
-    io::Write,
     net::SocketAddr,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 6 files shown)


applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -98,12 +98,17 @@ pub struct MergeMiningProxyService {
 }
 
 impl MergeMiningProxyService {
-    pub fn new(config: MergeMiningProxyConfig, block_templates: BlockTemplateRepository) -> Self {
+    pub fn new(
+        config: MergeMiningProxyConfig,
+        http_client: reqwest::Client,
```
delta1, line 101: looks like this change needs to be propagated to the tests too

StriderDM: Thanks @delta1

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -393,7 +400,7 @@ impl InnerService {
                     }
                 },
             }
-
+            drop(base_node_client);
```
stringhandler: is this necessary? Doesn't it drop at line 405?

stringhandler (commented):
Untested. Looks good, but the drops must either be removed or have a comment saying why they are necessary

stringhandler (approved):
Seems to be a lot of deleted code, and I didn't test it locally, but should be quick to discover if there is a bug here

sdbondi:
@StriderDM Any reason why the check_connections function was removed? It's pretty useful to quickly validate that everything is configured and hooked up correctly esp with the pool. cc. @mikethetike 

EDIT: Ah I see the behaviour is to crash/exit if the connections cannot be made.  I seem to remember there were reasons raised why it shouldn't work like that, personally, I'm fine with either :)

StriderDM:
@sdbondi We're trying to work around the bug where file descriptors are being leaked by grpc in linux. 

Initially tried an explicit drop for the connections but then this was reworked to reusing connections.

Behaviour is that instead of opening new connections, existing connections are reused, otherwise it should exit.

Open to other ideas on this?

sdbondi:
@StriderDM Nope that's all good 👍🏻  thanks

PR #2951, opened 2021-05-17
Improve error messages in tari applications

SWvheerden:
## Description

This PR updates some of the error messages in tari utilities to not display wallet. 
It also prints out a pretty error message when the base_node cannot connect to a base_node

## Motivation and Context

This piece of code is shared between all tari applications and the error messages here should not display wallet, base_node, etc. 
All of the application-specific names were changed to application, except for wallet_error which should only originate inside of a wallet. 

For example currently the base_node displays: `The wallet exited because an unknown error occurred. Check the logs for details.` if tor is not corrected. 
After this PR it should display:
`The application exited because an unknown error occurred. Check the logs for details. if tor is not corrected.`

The base_node should not print out a generic error  when it cannot connect to a tor port. This creates a pretty error to print out that it cant connect to tor as well as give a potential fix for it.

After this PR the base_node now prints out:
```Unable to connect to the Tor control port.
Please check that you have the Tor proxy running and that access to the Tor control port is turned on.
If you are unsure of what to do, use the following command to start the Tor proxy:
tor --allow-missing-torrc --ignore-missing-torrc --clientonly 1 --socksport 9050 --controlport 127.0.0.1:9051 --log "notice stdout" --clientuseipv6 1
Tor connection is offline
```

## How Has This Been Tested?

Run to ensure that the error message prints out compelelty. 

## Checklist:

### DIFF
--- a/applications/tari_app_utilities/src/utilities.rs
+++ b/applications/tari_app_utilities/src/utilities.rs
@@ -52,15 +52,15 @@ pub const LOG_TARGET: &str = "tari::application";
 pub enum ExitCodes {
     #[error("There is an error in the wallet configuration: {0}")]
     ConfigError(String),
-    #[error("The wallet exited because an unknown error occurred. Check the logs for details.")]
+    #[error("The application exited because an unknown error occurred. Check the logs for details.")]
     UnknownError,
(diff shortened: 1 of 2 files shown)


philipr-za (commented):
LGTM, though your cargo fmt is outdated.

PR #2952, opened 2021-05-17
Fix pruning in pruning mode

SWvheerden:
## Description

When running a prune mode, the node would display this error in the logs:
`Failed to apply DB transaction: The requested BlockHeader was not found via mmr_position`

## Motivation and Context

We need prune nodes to be made to prune spent UTXOs. We save the MMR position as a BigEndian u64, but when we load them, we ask for them by BigEndian u32. These are not equal. The u64 should be converted to u64 before being read.  

## How Has This Been Tested?

Run the node and let it prune on stibbons dev. 

## Checklist:


sdbondi (approved):
Nice LGTM

PR #2955, opened 2021-05-19
Implement cucumber tests for one-sided recovery and scanning

philipr-za:
## Description
This PR implements cucumber tests for recovering one-sided payments, scanning for them under normal wallet operation and also adds to the existing standard recovery test to spend the output after it has been recovered which revealed a few bugs. The PR also contains a number of bug fixes to make these tests possible.

- Console wallet will not enter TUI mode after recovery if its started in daemon mode, this crashes cucumber
- Persist “Known one-sided scripts” when a Comms Private/Public key is derived to be scanned for
- Fix the UTXO scanner when started as a service.
- Add Cucumber helpers to support the tests.

## How Has This Been Tested?
Cucumber tests provided

## Checklist:

### DIFF
--- a/integration_tests/features/WalletRecovery.feature
+++ b/integration_tests/features/WalletRecovery.feature
@@ -1,15 +1,50 @@
 @wallet-recovery
 Feature: Wallet Recovery
 
-    @critical @broken
-    Scenario: Wallet recovery with connected base node staying online
-        Given I have a seed node NODE
-        And I have 1 base nodes connected to all seed nodes
-        And I have wallet WALLET_A connected to all seed nodes
-        And I have a merge mining proxy PROXY connected to NODE and WALLET_A with default config
-        When I merge mine 10 blocks via PROXY
-        When I mine 10 blocks on NODE
-        When I wait for wallet WALLET_A to have at least 55000000000 uT
-        Then all nodes are at height 20
-        When I recover wallet WALLET_A into wallet WALLET_B connected to all seed nodes
-        Then wallet WALLET_A and wallet WALLET_B have the same balance
+@critical @broken
+Scenario: Wallet recovery with connected base node staying online
+    Given I have a seed node NODE
+    And I have 1 base nodes connected to all seed nodes
+    And I have wallet WALLET_A connected to all seed nodes
+    And I have a merge mining proxy PROXY connected to NODE and WALLET_A with default config
+    When I merge mine 10 blocks via PROXY
+    When I mine 5 blocks on NODE
+    When I wait for wallet WALLET_A to have at least 55000000000 uT
+    Then all nodes are at height 15
+    When I recover wallet WALLET_A into wallet WALLET_B connected to all seed nodes
+    Then wallet WALLET_A and wallet WALLET_B have the same balance
+    And I have wallet WALLET_C connected to all seed nodes
+    And I send 100000 uT from wallet WALLET_B to wallet WALLET_C at fee 100
+    Then wallet WALLET_B detects all transactions are at least Broadcast
+    When I mine 5 blocks on NODE
+    Then all nodes are at height 20
+    Then I wait for wallet WALLET_C to have at least 100000 uT
+
+@critical
+Scenario: Recover one-sided payments
+    Given I have a seed node NODE
+    And I have 1 base nodes connected to all seed nodes
+    And I have wallet WALLET_A connected to all seed nodes
+    And I have a merge mining proxy PROXY connected to NODE and WALLET_A with default config
+    When I merge mine 10 blocks via PROXY
+    Then all nodes are at height 10
+    And I have wallet WALLET_B connected to all seed nodes
+    And I stop wallet WALLET_B
+    # Send 2 one-sided payments to WALLET_B so it can spend them in two cases
+    Then I send a one-sided transaction of 1000000 uT from WALLET_A to WALLET_B at fee 100
+    Then I send a one-sided transaction of 1000000 uT from WALLET_A to WALLET_B at fee 100
+    When I merge mine 5 blocks via PROXY
+    Then all nodes are at height 15
+    When I recover wallet WALLET_B into wallet WALLET_C connected to all seed nodes
+    Then I wait for wallet WALLET_C to have at least 2000000 uT
+    # Send one of the recovered outputs back to Wallet A as a one-sided transactions
+    Then I send a one-sided transaction of 900000 uT from WALLET_C to WALLET_A at fee 100
+    When I merge mine 5 blocks via PROXY
+    Then all nodes are at height 20
+    Then I wait for wallet WALLET_C to have less than 1100000 uT
+    # Send the remaining recovered UTXO to self in standard MW transaction
+    Then I send 1000000 uT from wallet WALLET_C to wallet WALLET_C at fee 100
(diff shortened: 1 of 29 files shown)


integration_tests/features/WalletRecovery.feature
```diff
@@ -1,15 +1,49 @@
...
+    When I mine 5 blocks on NODE
+    When I wait for wallet WALLET_A to have at least 55000000000 uT
+    Then all nodes are at height 15
+    When I recover wallet WALLET_A into wallet WALLET_B connected to all seed nodes
+    Then wallet WALLET_A and wallet WALLET_B have the same balance
+    And I have wallet WALLET_C connected to all seed nodes
+    And I send 100000 uT from wallet WALLET_B to wallet WALLET_C at fee 100
+    When I mine 5 blocks on NODE
+    Then all nodes are at height 20
+    Then I wait for wallet WALLET_C to have at least 100000 uT
```
hansieodendaal: Nice addition

integration_tests/features/WalletRecovery.feature
```diff
@@ -1,15 +1,49 @@
...
+
+@critical
+Scenario: Scan for one-sided payments
```
hansieodendaal: **NIT**: Maybe the name should change here as the test is more about recovery and not so much scanning. With the latter, I would expect WALLET_B to be started up again in its normal mode and being issued a command to scan for one-sided transactions.
Maybe something like `Recover one-sided payments not received yet` ?

hansieodendaal:
Wow, lots of effort went into this

integration_tests/features/WalletRecovery.feature
philipr-za: Sure, can change the name. In the end recovery is the same process of scanning and this is in the Recovery Feature but 🤷‍♂️

hansieodendaal: Just a **NIT** though

hansieodendaal: Let me mark it as such

delta1:
Looks great, just a lint warning (how were you able to commit?! `npm install` should install husky with a pre-commit hook) 

```
1440:9  error  'identityInfo' is assigned a value but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

philipr-za:
Rebased and passing locally with `npm test -- --tags "@critical and not @long-running and not @broken"` but still requires a manual `npm link tari_crypto` with the latest version of `tari_crypto`

hansieodendaal:
> Rebased and passing locally with `npm test -- --tags "@critical and not @long-running and not @broken"` but still requires a manual `npm link tari_crypto` with the latest version of `tari_crypto`

Yes, I omitted to change `"tari_crypto": "^0.8.0"` to `"tari_crypto": "^0.9.0"` in `package.json`, which does the trick. Fixed in consecutive PRs.

integration_tests/features/WalletRecovery.feature
```diff
@@ -1,15 +1,49 @@
...
+    And I have a merge mining proxy PROXY connected to NODE and WALLET_A with default config
+    When I merge mine 10 blocks via PROXY
+    When I mine 5 blocks on NODE
+    When I wait for wallet WALLET_A to have at least 55000000000 uT
+    Then all nodes are at height 15
+    When I recover wallet WALLET_A into wallet WALLET_B connected to all seed nodes
+    Then wallet WALLET_A and wallet WALLET_B have the same balance
+    And I have wallet WALLET_C connected to all seed nodes
+    And I send 100000 uT from wallet WALLET_B to wallet WALLET_C at fee 100
+    When I mine 5 blocks on NODE
```
hansieodendaal, line 19: We need a wait of a second or so here **before this step** (or some other method like mine one block then check heights) otherwise mining starts before the transaction is finalized, especially on a faster system. This test failed on one system and passed on the other due to the confirmations only going to count of 2; 2 blocks were mined and the next block template issued before the transaction was ready and validated in the mempool.

philipr-za: hmm, perhaps I will rather put in a statement to wait for the transaction to be broadcast. I prefer that to adding in a finite wait which might not work on all setups.

philipr-za: added `Then wallet WALLET_B detects all transactions are at least Broadcast`

integration_tests/features/support/steps.js
```diff
@@ -996,21 +1001,42 @@ When(
...
+      async () => walletClient.isBalanceAtLeast(amount),
+      true,
+      240 * 1000,
+      5 * 1000,
+      5
+    );
+    if (!walletClient.isBalanceAtLeast(amount)) {
+      console.log("Balance not adequate!");
     }
+    consoleLogBalance(await walletClient.getBalance());
```
hansieodendaal:
```suggestion
    consoleLogBalance(await walletClient.getBalance());
    if (!(await walletClient.isBalanceAtLeast(amount))) {
      console.log("Balance not adequate!");
    }
    expect(await walletClient.isBalanceAtLeast(amount)).to.equal(true);
```

hansieodendaal: We need an assert here

integration_tests/features/support/steps.js
```diff
@@ -996,21 +1001,42 @@ When(
...
+      async () => walletClient.isBalanceLessThan(amount),
+      true,
+      700 * 1000,
+      5 * 1000,
+      5
+    );
+    if (!walletClient.isBalanceLessThan(amount)) {
+      console.log("Balance has not dropped below specified amount!");
+    }
+    consoleLogBalance(await walletClient.getBalance());
```
hansieodendaal:
```suggestion
    consoleLogBalance(await walletClient.getBalance());
    if (!(await walletClient.isBalanceLessThan(amount))) {
      console.log("Balance has not dropped below specified amount!");
    }
    expect(await walletClient.isBalanceLessThan(amount)).to.equal(true);
```

hansieodendaal: We need an assert here

hansieodendaal (commented):
I only reviewed the cucumber tests and have some additional comments.

Comment already addressed:
"_Mining can start before the transaction is finalized, especially on a faster system. The `Wallet recovery with connected base node staying online` test failed on one system and passed on the other due to the confirmations only going to count of 2; 2 blocks were mined and the next block template issued before the transaction was ready and validated in the mempool._"

integration_tests/features/WalletRecovery.feature
hansieodendaal: Yip, that did the trick.

hansieodendaal:
> > Rebased and passing locally with `npm test -- --tags "@critical and not @long-running and not @broken"` but still requires a manual `npm link tari_crypto` with the latest version of `tari_crypto`
> 
> Yes, I omitted to change `"tari_crypto": "^0.8.0"` to `"tari_crypto": "^0.9.0"` in `package.json`, which does the trick. Fixed in consecutive PRs.

If you change  `"tari_crypto": "^0.8.0"` to `"tari_crypto": "^0.9.0"` in `package.json` the `run-integration-tests` step should be able to pass.

stringhandler (approved):

PR #2956, opened 2021-05-19
Fix birthday attack vulnerability in tari script offset

SWvheerden:
## Description

This updates the script offset inside of tari script to secure against birthday attacks by removing the hash of the UTXO and only secure it by signatures. 

## Motivation and Context

The current tari script features a `script offset`  `γ=∑k_s - ∑k_o*Hash(UTXO)` but its possible to reduce this to `-Hash(x) = Hash(Y) + Hash(Z)` and this is susceptible to the birthday attack using Wagner's algorithm. 

By making the script offset only `γ=∑k_s - ∑k_o` we remove the attack vector. But this only provided that the key provided was correct. On the output, we need to sign with `k_o` all the metadata attached to an output. 

`k_s` should be secure most of the time by the fact that the script should prove or validate the key, but with scripts that have more than one valid input such as a NOP script, it is possible to swap out the `k_s` key with a different key. This also introduces a signature where the sender signs the public `K_s` with the commitment to prove the owner is the sender. 

## How Has This Been Tested?

Docs only

## Checklist:

This was Co-authored by @hansieodendaal

### DIFF
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -448,9 +448,7 @@ validation steps.
 For this use case we have Alice who sends Bob some Tari.
 Bob's wallet is  online and is able to countersign the transaction.
 
-Alice creates a new transaction spending \\( C\_a \\) to a new output containing the commitment \\( C\_b \\) (ignoring fees for now). 
-Because Alice is creating the transaction,  she has a final say over the script \\( \script_b \\), similar to a 
-[bitcoin transaction]. If Bob provides a script she has to approve it or she can choose something akin to a `NOP` script.
+Alice creates a new transaction spending \\( C\_a \\) to a new output containing the commitment \\( C\_b \\) (ignoring fees for now).
 
 To spend \\( C\_a \\), she provides:
 
@@ -468,9 +466,14 @@ $$
  \tag{10}
 $$
 
+
 Alice sends the usual first round data to Bob. Bob can then completes his side of the transaction and returns his commitment 
 \\(C\_b\\) along with its rangeproof, the partial signature and public blinding factor back to Alice. All this happens as per the 
-[standard Mimblewimble protocol]
+[standard Mimblewimble protocol]. 
+
+Because Alice is creating the transaction, she has a final say over the script \\( \script_b \\), similar to a 
+[bitcoin transaction], but Bob can also opt to send Alice a script \\(\script\_b\\) of his choosing. If Bob did not send a script, 
+she chooses something akin to a `NOP` script for the script \\(\script\_b\\).
 
 Alice calculates the sender signature \\( s_{Mb} \\) with: 
 
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -308,16 +308,16 @@ script input, _public script key_ and the commitment:
 
 $$
 \begin{aligned}
- s_{Si} = (s_{S1i}, s_{S2i}, R_{Si} )
+ s_{Si} = (a_{Si}, b_{Si}, R_{Si} )
  \end{aligned}
  \tag{3}
 $$
 Where
 $$
 \begin{aligned}
-R_{Si} &= r_{S1i} \cdot H + r_{S2i} \cdot G \\\\
-s_{S1i}  &= r_{S1i} +  e(v_{i}) \\\\
-s_{S2i} &= r_{S2i} +  e(k_{Si}+k_i) \\\\
+R_{Si} &= r_{Si_a} \cdot H + r_{Si_b} \cdot G \\\\
+a_{Si}  &= r_{Si_a} +  e(v_{i}) \\\\
+b_{Si} &= r_{Si_b} +  e(k_{Si}+k_i) \\\\
 e &= \hash{ R_{Si} \cat \alpha_i \cat \input_i \cat K_{Si} \cat C_i} \\\\
 \end{aligned}
 \tag{4}
@@ -326,7 +326,7 @@ $$
 This is verified by the following:
  $$
 \begin{aligned}
-s_{S1i} \cdot H + s_{S2i} \cdot G = R_{Si} + (C_i+K_{Si})e
+a_{Si} \cdot H + b_{Si} \cdot G = R_{Si} + (C_i+K_{Si})e
  \end{aligned}
  \tag{5}
 $$
@@ -456,7 +456,7 @@ To spend \\( C\_a \\), she provides
 
 * An input matching \\( C\_a \\), some output already on the blockchain.
 * The script input, \\( \input_a \\).
-* A valid script signature, \\( (s_{S1a}, s_{S2a}, R_{Sa}) \\) as per (3),(4) proving that she owns the commitment 
+* A valid script signature, \\( (a_{Sa}, b_{Sa}, R_{Sa}) \\) as per (3),(4) proving that she owns the commitment 
   \\( C\_a \\), knows the private key, \\( k_{Sa} \\), corresponding to \\( K_{Sa} \\), the public key left on the stack 
   after executing \\( \script_a \\) with \\( \input_a \\).
 * An _offset public key_, \\( k_{Ob} \\).
@@ -505,7 +505,7 @@ Base nodes validate the transaction as follows:
 * The _script signature_ on Alice's input is valid by checking:
   $$
 \begin{aligned}
-    s_{S1a} \cdot H + s_{S2a} \cdot G = R_{Sa} + (C_a + K_{Sa})* \hash{ R_{Sa} \cat \alpha_a \cat \input_a \cat K_{Sa} \cat C_a}
+    a_{Sa} \cdot H + b_{Sa} \cdot G = R_{Sa} + (C_a + K_{Sa})* \hash{ R_{Sa} \cat \alpha_a \cat \input_a \cat K_{Sa} \cat C_a}
   \end{aligned}
  \tag{13}
   $$
@@ -621,7 +621,7 @@ To summarise, the information required for one-sided transactions is as follows:
 | script            | \\( \alpha_a \\)                      | Public                                                                                        |
 | script input      | \\( \input_a \\)                      | Public                                                                                        |
 | height            | \\( h_a \\)                           | Public                                                                                        |
-| script signature  | \\( s_{S1a},s_{S2a}, R_{Sa} \\)       | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
+| script signature  | \\( a_{Sa},b_{Sa}, R_{Sa} \\)       | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
 | offset public key | \\( K_{Oa} \\)                        | Not used in this transaction                                                                  |
 
 | Transaction output | Symbols                               | Knowledge                                                              |
@@ -741,7 +741,7 @@ To summarise, the information required for creating a multiparty UTXO is as foll
 | script                      | \\( \alpha_a \\)                      | Public                                                                                        |
 | script input                | \\( \input_a \\)                      | Public                                                                                        |
 | height                      | \\( h_a \\)                           | Public                                                                                        |
-| script signature            | \\( (s_{S1a},s_{S2a}, R_{Sa}) \\)     | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
+| script signature            | \\( (a_{Sa},b_{Sa}, R_{Sa}) \\)     | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
 | offset&nbsp;public&nbsp;key | \\( K_{Oa} \\)                        | Not used in this transaction                                                                  |
 
 <br>
@@ -764,7 +764,7 @@ When spending the multi-party input:
 | script                      | \\( \alpha_s \\)                        | Public                                                                                                                                                             |
 | script input                | \\( \input_s \\)                        | Public                                                                                                                                                             |
 | height                      | \\( h_a \\)                             | Public                                                                                                                                                             |
-| script&nbsp;signature       | \\( (s_{S1s} ,s_{S2s} , R_{Ss}) \\)     | Alice knows \\( (k_{SsA},\\, r_{SsA}) \\), Bob knows \\( (k_{SsB},\\, r_{SsB}) \\). Both parties know \\( (k_{s},\\, v_{s}) \\). Neither party knows \\( k_{Ss}\\) |
+| script&nbsp;signature       | \\( (a_{Ss} ,b_{Ss} , R_{Ss}) \\)     | Alice knows \\( (k_{SsA},\\, r_{SsA}) \\), Bob knows \\( (k_{SsB},\\, r_{SsB}) \\). Both parties know \\( (k_{s},\\, v_{s}) \\). Neither party knows \\( k_{Ss}\\) |
 | offset&nbsp;public&nbsp;key | \\( K_{Os} \\)                          | As above, Alice and Bob each know part of the offset key                                                                                                           |
 
 
@@ -860,7 +860,7 @@ refer to a UTXO _receiver_ and _script_ respectively.
 | \\( \so_t \\)             | The script offset for transaction _t_, as \\( \so_t = \sum_j{ k_{Sjt}} - \sum_i{k_{Oit}}\\)                                                                                                                                           |
 | \\( C_i \\)               | A Pedersen commitment to a value \\( v_i \\), as \\( C_i = k_i \cdot{G} + v_i \cdot H \\)                                                                                                                                             |
 | \\( \input_i \\)          | The serialised input for script \\( \script_i \\)                                                                                                                                                                                     |
-| \\( s_{Si} \\)            | A script signature for output \\( i \\), as \\( s_{Si} = (s_{S1i}, s_{S2i}, R_{Si} ) = (r_{S1i} +  e(v_{i})), (r_{S2i} + e(k_{Si}+k_i)) \\; \text{where} \\; e = \hash{ R_{Si} \cat \script_i \cat \input_i \cat K_{Si} \cat C_i} \\) |
+| \\( s_{Si} \\)            | A script signature for output \\( i \\), as \\( s_{Si} = (a_{Si}, b_{Si}, R_{Si} ) = (r_{Si_a} +  e(v_{i})), (r_{Si_b} + e(k_{Si}+k_i)) \\; \text{where} \\; e = \hash{ R_{Si} \cat \script_i \cat \input_i \cat K_{Si} \cat C_i} \\) |
 | \\( s_{Mi} \\)            | A sender signature for output \\( i \\), as \\( s_{Mi} = r_{Mi} + k_{Oi}\hash{ \script_i \cat F_i \cat R_{Mi}  } \\)                                                                                                                  |
 
 ## Extensions
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -233,7 +233,7 @@ pub struct TransactionOutput {
     /// The script offset pubkey, K_O
     script_offset_public_key: PublicKey
     /// UTXO signature with the script offset private key, k_O
-    sender_signature : Signature
+    sender_metadata_signature : Signature
 }
 ```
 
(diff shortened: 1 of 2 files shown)


RFC/src/RFC-0201_TariScript.md
```diff
@@ -287,65 +285,69 @@ The updated input definition is:
...
 
 $$
-  s_{Si} = r_{Si} + \hash{ R_{Si} \cat \alpha_i \cat \input_i \cat h_i} k_{Si}
-$$ 
-
-### Consensus changes
-
-The Mimblewimble balance for blocks and transactions stays the same.
+\begin{aligned}
+ s_{Si} = (s_{S1i}, s_{S2i}, R_{Si} )
```
philipr-za, line 306: Nit: This is a readability nit, for the two commitment components of the CommSig I think perhaps use the u and v  notation from the paper as it is more readable than S1 and S2. There are already a lot of S's and i's in the notation and u and v will stand out better. 
```suggestion
 s_{Si} = (u_{Si}, v_{Si}, R_{Si} )
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -287,65 +285,69 @@ The updated input definition is:
...
+ \end{aligned}
+ \tag{3}
+$$
+Where
+$$
+\begin{aligned}
+R_{Si} &= r_{S1i} \cdot H + r_{S2i} \cdot G \\\\
```
philipr-za, line 318:
```suggestion
R_{Si} &= r_{Si_u} \cdot H + r_{Si_v} \cdot G \\\\
```

hansieodendaal (commented):
This PR seems to address all concerns and proposal raised by [DavidBurkett's review](https://gist.github.com/DavidBurkett/b2ace786c6298179607b1337c3657a78), although not directly in all cases.

The addition of the `sender_signature` (2) in the transaction output commits to all its metadata once mined, and therefore it cannot be changed by anyone upon spending when applicable values are reused in this scheme.

The addition of the commitment signature (3) (`ComSig`), and combining the signing of its challenge with (a) the commitment value as well as (b) the aggregated script private key and commitment blinding factor (4), introduces a really tight committing to all the metadata of the input, which is also easy to verify (5). 

As shown with **### Script Offset security**, this `ComSig` also prevents any malleability with the components of the transaction script offset (6), (7) and aggregated block script offset (8), (9) (`gamma_total`), in which the final security of Tari Script lies.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
 
 $$
-  s_{Si} = r_{Si} + \hash{ R_{Si} \cat \alpha_i \cat \input_i \cat h_i} k_{Si}
-$$ 
-
-### Consensus changes
-
-The Mimblewimble balance for blocks and transactions stays the same.
+\begin{aligned}
+ s_{Si} = (s_{S1i}, s_{S2i}, R_{Si} )
```
philipr-za: Nit: This is a readability nit, for the two commitment components of the CommSig I think perhaps use new letters as it is more readable than S1 and S2. There are already a lot of S's and i's in the notation and distinct symbols will stand out better.  Seeing as `v` is used for value and `k` is used for key maybe `a `and `b`
```suggestion
  s_{Si} = (a_{Si}, b_{Si}, R_{Si} )
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+ \end{aligned}
+ \tag{3}
+$$
+Where
+$$
+\begin{aligned}
+R_{Si} &= r_{S1i} \cdot H + r_{S2i} \cdot G \\\\
```
philipr-za:
```suggestion
R_{Si} &= r_{Si_a} \cdot H + r_{Si_b} \cdot G \\\\
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+s_{S1i}  &= r_{S1i} +  e(v_{i}) \\\\
```
philipr-za:
```suggestion
a_{Si}  &= r_{Si_a} +  e(v_{i}) \\\\
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+s_{S2i} &= r_{S2i} +  e(k_{Si}+k_i) \\\\
```
philipr-za:
```suggestion
b_{Si} &= r_{Si_b} +  e(k_{Si}+k_i) \\\\
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+\end{aligned}
+\tag{4}
+$$
 
-The range proof commits all the values comprising the transaction output. So now instead of verifying the range proof
-using the standard output commitment, \\( C_i \\), we use the modified commitment,
+This is verified by the following:
+ $$
+\begin{aligned}
+s_{S1i} \cdot H + s_{S2i} \cdot G = R_{Si} + (C_i+K_{Si})e
```
philipr-za:
```suggestion
a_{Si} \cdot H + b_{Si} \cdot G = R_{Si} + (C_i+K_{Si})e
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -233,8 +230,10 @@ pub struct TransactionOutput {
     proof: RangeProof,
     /// The serialised script
     script: Vec<u8>,
-    /// The offset pubkey, K_O
+    /// The script offset pubkey, K_O
     script_offset_public_key: PublicKey
+    /// UTXO signature with the script offset private key, k_O
+    sender_signature : Signature
```
philipr-za: Perhaps this could be called the `sender_metadata_signature` to make it clear from the field name what it signs?

RFC/src/RFC-0201_TariScript.md
```diff
@@ -394,84 +448,90 @@ validation steps.
...
+\begin{aligned}
+  s_{Mb} = r_{mb} + k_{Ob} \hash{ \script_b \cat F_b \cat R_{Mb} }
+   \end{aligned}
+ \tag{11}
+$$
 
-Alice sends Bob this proto-UTXO, along with her signature nonce, as per the standard Mimblewimble protocol. However, she
-also provides Bob with \\( K_{Ob} \\), the script offset public key.
+Alice sends her excess, along with her signature nonce, as per the standard Mimblewimble protocol. However, she also 
+provides Bob with \\( \so \\), the _script offset_, as wel as a meta_signature \\(s_{Mb}\\) for his output \\(C_b\\).
```
philipr-za:
```suggestion
provides Bob with \\( \so \\), the _script offset_, as well as a meta_signature \\(s_{Mb}\\) for his output \\(C_b\\).
```

philipr-za: I think she has to send bob the Public Script Offset, why does she need to send Gamma?

philipr-za: Honestly I wonder if there is much value in sending the metadata signature to bob at all? He doesn't include it in any of his calculations and after he has replied she could change it to whatever she wants. He will obviously be able to pick up the changes when he receives the completed transaction but at this stage of the dance what is he really verifying that can't be changed by alice at the next stage?

philipr-za: I suppose you mean for him to include it in the Output but can't Alice do that in the next step?

RFC/src/RFC-0201_TariScript.md
```diff
@@ -394,84 +448,90 @@ validation steps.
...
-  \\( K_{Sa} \\).
-* The script offset is verified by checking that the balance
+* The sender signature \\(s_{Ma}\\) on Bob's output,
+* The input _script_ must execute successfully using the provided input data; and the script result must be a valid 
+  public key,
+* The _script signature_ on Alice's input is valid by checking:
   $$
-    \so \cdot{G} = K_{Sa} - \HU_b K_{Ob}
+\begin{aligned}
+    s_{S1a} \cdot H + s_{S2a} \cdot G = R_{Sa} + (C_a + K_{Sa})* \hash{ R_{Sa} \cat \alpha_a \cat \input_a \cat K_{Sa} \cat C_a}
```
philipr-za:
```suggestion
    a_{Sa} \cdot H + b_{Sa} \cdot G = R_{Sa} + (C_a + K_{Sa})* \hash{ R_{Sa} \cat \alpha_a \cat \input_a \cat K_{Sa} \cat C_a}
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -484,145 +544,178 @@ Alice owns \\( C_a \\) and provides the required script to spend the UTXO as was
...
+|-------------------|---------------------------------------|-----------------------------------------------------------------------------------------------|
+| commitment        | \\( C_a = k_a \cdot G + v \cdot H \\) | Alice knows the blinding factor and value                                                     |
+| features          | \\( F_a \\)                           | Public                                                                                        |
+| script            | \\( \alpha_a \\)                      | Public                                                                                        |
+| script input      | \\( \input_a \\)                      | Public                                                                                        |
+| height            | \\( h_a \\)                           | Public                                                                                        |
+| script signature  | \\( s_{S1a},s_{S2a}, R_{Sa} \\)       | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
```
philipr-za:
```suggestion
| script signature  | \\( a_{Sa},b_{Sa}, R_{Sa} \\)       | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -484,145 +544,178 @@ Alice owns \\( C_a \\) and provides the required script to spend the UTXO as was
...
+|--------------------|---------------------------------------|------------------------------------------------------------------------|
+| commitment         | \\( C_b = k_b \cdot G + v \cdot H \\) | Alice and Bob know the blinding factor and value                       |
 | features           | \\( F_b \\)                           | Public                                                                 |
-| script hash        | \\( \scripthash_b \\)                 | Script is effectively public. Only Bob knows the correct script input. |
+| script             | \\( \script_b \\)                     | Script is public. Only Bob knows the correct script input.             |
 | range proof        |                                       | Alice and Bob know opening parameters                                  |
 | offset public key  | \\( K_{Ob} \\)                        | Alice knows \\( k_{Ob} \\)                                             |
+| sender signature   | \\( s_{Mb}, R_{Mb} \\)                | Alice knows \\( k_{Ob} \\)  and the metadata)                          |
```
philipr-za:
```suggestion
| sender metadata signature   | \\( s_{Mb}, R_{Mb} \\)                | Alice knows \\( k_{Ob} \\)  and the metadata                          |
```

philipr-za: I would add the contents of the challenge here

RFC/src/RFC-0201_TariScript.md
philipr-za: I would add the contents of the challenge here

philipr-za (commented):
Looks good, I spent some time looking at the CommSig formulation and I also believe it is secure. I have some notation nits and a couple questions about the transaction dance.

I didn't review the HTLC section so if you make the notation changes to the previous sections just carry them through to that section too.

RFC/src/RFC-0201_TariScript.md
SWvheerden: Yeah, the whole idea here is to complete the dance with as few as possible steps. But this is all implementation details as they need to ensure both are happy with all aspects of the transaction.

RFC/src/RFC-0201_TariScript.md
SWvheerden: The challenge contents are all public information, these tables are for private information.

RFC/src/RFC-0201_TariScript.md
SWvheerden: The challenge contents are all public information, these tables are for private information.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
 
 $$
-  s_{Si} = r_{Si} + \hash{ R_{Si} \cat \alpha_i \cat \input_i \cat h_i} k_{Si}
-$$ 
-
-### Consensus changes
-
-The Mimblewimble balance for blocks and transactions stays the same.
+\begin{aligned}
+ s_{Si} = (a_{Si}, b_{Si}, R_{Si} )
```
stringhandler, line 312: There are a lot of variables in the subscript of this equation, can this not be simplified and also use `u` and `v` from the ComSig paper?

```suggestion
 s_i = (u_i, v_i, R_i )
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+\end{aligned}
+\tag{4}
+$$
 
-The range proof commits all the values comprising the transaction output. So now instead of verifying the range proof
-using the standard output commitment, \\( C_i \\), we use the modified commitment,
+This is verified by the following:
+ $$
+\begin{aligned}
+a_{Si} \cdot H + b_{Si} \cdot G = R_{Si} + (C_i+K_{Si})e
```
stringhandler, line 330: I may be wrong, but has `K_S` been mentioned before this point in the RFC?

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+$$
 
-$$ \hat{C_i} = v_i\cdot H  + \bigl( k_i + \rpc_i \bigr) \cdot G  $$
+This signature ensures that only the owner can provide the input data to the TransactionInput and that the transaction's 
+_script offset_ \\(\so\\) cannot be indirectly changed by changing the script private key \\(k\_{Si}\\).
 
-We can then verify the range proof. If the range proof is valid, we know that the value v, is positive and that none of
-the values have been changed.
+This stops any attack that attempts to change the information inside of the challenge \\(e\\) where more than one 
+possible input is valid such as with a `NOP` script. If an attacker can change the \\(K\_{Si}\\) keys of the input then
```
stringhandler: This paragraph is out of place. It assumes you have read or know about the attack you are mentioning. Perhaps it would be better to write out an example of the attack and how it's not possible without knowing the key

RFC/src/RFC-0201_TariScript.md
stringhandler: i.e. where does K_S get calculated

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
+ $$
+\begin{aligned}
+a_{Si} \cdot H + b_{Si} \cdot G = R_{Si} + (C_i+K_{Si})e
+ \end{aligned}
+ \tag{5}
+$$
 
-$$ \hat{C_i} = v_i\cdot H  + \bigl( k_i + \rpc_i \bigr) \cdot G  $$
+This signature ensures that only the owner can provide the input data to the TransactionInput and that the transaction's 
+_script offset_ \\(\so\\) cannot be indirectly changed by changing the script private key \\(k\_{Si}\\).
```
stringhandler:
```suggestion
_script offset_ \\(\so\\) (see below) cannot be indirectly changed by changing the script private key \\(k\_{Si}\\).
```

RFC/src/RFC-0201_TariScript.md
```diff
@@ -370,19 +410,33 @@ pub struct BlockHeader {
...
+transaction or the block, as that will invalidate the _script offset_ balance equation, either (7) or (9) depending on 
+whether the scope is a transaction or block. It is important to know that this also stops cut&#8209;through so that we 
+can verify all spent UTXO scripts. Because the _private script key_ and _private script offset key_ is not publicly 
+known, its impossible to create a new _script offset_.
+
+Certain scripts may allow more than one valid set of _input data_. Users might be led to believe that this will allow a 
+third party to change the _script keypair_ \\((k\_{Si}\\),\\(K\_{Si})\\), but this is not possible (see 
+[Script Offset security](#script-offset-security)). If equation (7) or (9) balances then we know that every included 
+input and output in the transaction or block has its correct _public script key_ and _public script offset key_, which 
+in turn means we can trust and know that the supplied signatures (2),(3) are valid for all contained metadata.
```
stringhandler:
```suggestion
in turn means we can know that the supplied signatures (2),(3) are valid for all contained metadata.
```

stringhandler: all contained metadata? What exactly do you mean here?

stringhandler: I don't know what this sentence means. Surely you still need to check the signatures

RFC/src/RFC-0201_TariScript.md
```diff
@@ -394,84 +448,90 @@ validation steps.
...
+Alice creates a new transaction spending \\( C\_a \\) to a new output \\( C\_b \\) (ignoring fees for now).
+Because Alice is spending the transaction she chooses the _script_ \\( \script_b \\), she can either ask Bob for one, or 
+choose something akin to a `NOP` script.
 
-To spend \\( C_a \\), she provides
+To spend \\( C\_a \\), she provides
 
-* A script, \\( \alpha_a \\) such that the script hash, \\( \scripthash_a = \hash{ \alpha_a }\\) matches the blockchain
-  record for the UTXO containing \\( C_a \\).
+* An input matching \\( C\_a \\), some output already on the blockchain.
```
stringhandler: Isn't this a duplicate of the line above?

RFC/src/RFC-0201_TariScript.md
stringhandler: I think it should be made more clear here. There seems like there is no need for Alice to chose a script before hand, or to send the signature before Bob chooses an output.  Seems like she could just send the signature at the same time she sends the signature for the kernel

RFC/src/RFC-0201_TariScript.md
```diff
@@ -394,84 +448,90 @@ validation steps.
...
 
-Bob returns the UTXO and partial signature along with his nonce, \\( R_b \\), back to Alice.
-
-Alice will then complete the transaction by calculating the script offset, \\( \so\\):
-$$
-\so  = k_{Sa} - k_{Ob} \HU_b
-$$
-
-She can then construct and broadcast the transaction to the network.
+Bob can then submit the completed transaction.
```
stringhandler: Is this correct? Feels like a deviation from MW https://tlu.tarilabs.com/protocols/mimblewimble-1/MainReport.html

RFC/src/RFC-0201_TariScript.md
```diff
@@ -394,84 +448,90 @@ validation steps.
...
+  signature. This check does not change nor do the other validation rules, such as confirming that all inputs are in
   the UTXO set etc.
-* The range proof of Bob's output is validated with \\( \hat{C}_b \\) rather than \\( C_b \\),
-* The script signature on Alice's input is validated against the script, input and mining height,
-* The script hash must match the hash of the provided input script,
-* The input script must execute successfully using the provided input data; and the script result must be a valid public key,
-  \\( K_{Sa} \\).
-* The script offset is verified by checking that the balance
+* The sender signature \\(s_{Ma}\\) on Bob's output,
+* The input _script_ must execute successfully using the provided input data; and the script result must be a valid
```
stringhandler: why the italics here

RFC/src/RFC-0201_TariScript.md
```diff
@@ -360,7 +374,33 @@ pub struct Transaction {
...
+\so_{total} \cdot G = \sum_j\mathrm{K_{Sj}} - \sum_i\mathrm{K_{Oi}} \\; \text{for each input}, j,\\, \text{and each output}, i
+ \end{aligned}
+ \tag{9}
+$$
+
+As can be seen all information required to verify (8) is contained in a block's inputs and outputs. One important 
+distinction to make is that the Coinbase UTXO does not count towards the _script offset_. This is because the Coinbase 
+UTXO already has special rules accompanying it and it has no input, thus we cannot generate a _script offset_ 
+\\( \so \\) for a Coinbase transaction.
+The coinbase can allow any _script_ \\(\script_i\\) and  _script offset public key_ \\( K\_{Oi} \\) as long as it does
```
stringhandler: why not make the script offset pub key and signature empty for the coinbases? Seems like a waste to check this signature

RFC/src/RFC-0201_TariScript.md
```diff
@@ -278,74 +278,88 @@ pub struct TransactionInput {
...
-\\( \so\\) is calculated and verified as part of every block and transaction validation. This is calculated as follows:
+For every transaction an accompanying script offset \\( \so \\) needs to be provided. This is there to prove that every 
+public script key \\( K\_{Sj} \\) and every _public script offset key_ \\( K\_{Oi} \\) supplied with the UTXOs are the 
+correct ones. The sender will know and provide _script offset private keys_ \\(k_{Oi} \\) and script private keys 
+\\(k_{Si} \\); these are combined to create the _script offset_ \\( \so \\), which is calculated as follows:
 
 $$
-\so = \sum_i\mathrm{k_{Si}} - \sum_j(\mathrm{k_{Oj}  \HU_j}) \\; \text{for each input}, i,\\, \text{and each output}, j
+\begin{aligned}
+\so = \sum_j\mathrm{k_{Sj}} - \sum_i\mathrm{k_{Oi}} \\; \text{for each input}, j,\\, \text{and each output}, i
```
stringhandler, line 347: I think that this now allows cut through, you could remove a cut-through transaction and it's gamma from the block and publish the new gamma

stringhandler (commented):
I feel like there could be a more clear name for the script offset keys. 

The `script_offset_public_key` in each output is a key generated by the sender and the signature is a preventing malleability of the output features and script. Perhaps the field should be called `output_witness_sig` and `output_witness_public_key`

RFC/src/RFC-0201_TariScript.md
SWvheerden: I am specifically avoiding v_i, as it's the value of the blinding factor of commitment I.

RFC/src/RFC-0201_TariScript.md
hansieodendaal: Signatures (2) & (3) are checked independently from script offset verification (7) and (9), and looked at in isolation those could verify correctly but still be fake. When doing verification in (7) and (9) you know that the signatures and the message/metadata signed by the private keys can be trusted.

RFC/src/RFC-0201_TariScript.md
hansieodendaal: Agreed, the description must be cleared up a bit to follow the basic protocol.

RFC/src/RFC-0201_TariScript.md
SWvheerden: Clarification added to` ## Protocol modifications` section explicitly stating where this comes from. It was stated but not as: this is K_S

RFC/src/RFC-0201_TariScript.md
SWvheerden: Coinbase script and script_offset_keys are not checked when creating the coinbase output. But this allows a user to create a custom script for his coinbase that still needs validating on spending. This allows for example the miner to lock up the coinbase for children based on certain criteria. The meta_signature should still be singed to prevent malleability on the metadata such as the script.

RFC/src/RFC-0201_TariScript.md
```diff
@@ -153,56 +150,57 @@ chain synchronisation in pruned mode.
...
+* We must be able to prove that the UTXO originator provides the script and no-one else.
 * The scripts and their redeeming inputs must be stored on the block chain. In particular, the input data must not be
   malleable.
 
 The next section discusses the specific proposals for achieving these requirements.
 
 ## Protocol modifications
 
-At a high level, Tari script works as follows:
+Please refer to [Notation](#notation), which provides important pre-knowledge for the remainder of the report.
```
mikethetike, line 182: Just add the notation section here I think

RFC/src/RFC-0201_TariScript.md
```diff
@@ -370,19 +404,37 @@ pub struct BlockHeader {
     pub total_script_offset: BlindingFactor,
 }
 ```
-One important distinction to make is that the coinbase utxo does not count towards the _script offset_. This is because the coinbase UTXO already has special rules accompanying it and it has no input. Thus we cannot generate a \\( \so \\) for a coinbase transaction.
-The coinbase can allow any script and \\( k_{O} \\) as long as the range proof is validly constructed for \\( \hat{C_i} \\) and it does not break any of the rules in [RFC 120](RFC-0120_Consensus.md).
+
+This notion of the script offset \\(\so\\) means that the no third party can remove any input or output from a 
+transaction or the block, as that will invalidate the script offset balance equation, either (7) or (9) depending on 
+whether the scope is a transaction or block. It is important to know that this also stops 
+[cut&#8209;through](#cut-through) so that we can verify all spent UTXO scripts. Because the private script key and private
```
stringhandler, line 413: A party that knows `k_o` and `k_s` from an output being spent and the input spending it could still perform cut through. Not sure if this makes a difference, but probably worth mentioning

RFC/src/RFC-0201_TariScript.md
stringhandler: But pub key used to sign the script is also malleable, so the attacker would just change that as well

RFC/src/RFC-0201_TariScript.md
SWvheerden: While this is true, only the sender knows `k_s` and `k_o` so only he can do this.

stringhandler (approved):
I think we can merge this now, and any further comments can be created as PRs to this RFC

PR #2960, opened 2021-05-20
Improve LMWA stability

SWvheerden:
## Description

This PR improves the LMWA by swapping it over to only use integer math. 
It Also fixes a potential overflow bug. 

## Motivation and Context

The LWMA currently uses floating arithmetic to calculate and then ceil the final value to get rid of any rounding issues. This could still potentially cause issues, so this is now replaced with integer math. The unit tests have all been updated with the new values as the values somewhat change.

The LWMA also had a potential overflow in it, if the difficulty reaches 1/90 with the current values of u64 math, it will overflow as it averages the difficulties of the past 90 blocks. This changes the average calculation to use u128 so that it cannot overflow. 

## How Has This Been Tested?

All unit tests still pas. 
Synced up to height 3000 on tari_script using the new integer version, while the chain was calculated using floating-point version. 

## Checklist:

### DIFF
--- a/base_layer/core/src/proof_of_work/lwma_diff.rs
+++ b/base_layer/core/src/proof_of_work/lwma_diff.rs
@@ -20,7 +20,7 @@ pub const LOG_TARGET: &str = "c::pow::lwma_diff";
 pub struct LinearWeightedMovingAverage {
     target_difficulties: VecDeque<(EpochTime, Difficulty)>,
     block_window: usize,
-    target_time: u64,
+    target_time: u128,
     max_block_time: u64,
 }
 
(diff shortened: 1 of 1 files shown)


base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -47,9 +47,9 @@ impl LinearWeightedMovingAverage {
             .target_difficulties
             .iter()
             .skip(1)
-            .fold(0u64, |difficulty, (_, d)| difficulty + d.as_u64());
+            .fold(0u128, |difficulty, (_, d)| difficulty + d.as_u64() as u128);
 
-        let ave_difficulty = difficulty as f64 / n as f64;
+        let ave_difficulty = (difficulty / n as u128) as u64;
```
stringhandler: Perhaps add a comment stating why you are doing this. 

Also, maybe just set `n` as a `u128` when you declare it

stringhandler: You should also add a unit test to prove that this doesn't overflow

stringhandler (commented):
Nice. Needs a test to prove the overflow can't occur though

base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -29,27 +29,29 @@ impl LinearWeightedMovingAverage {
...
+        let mut weighted_times: u128 = 0;
         let difficulty = self
             .target_difficulties
             .iter()
             .skip(1)
-            .fold(0u64, |difficulty, (_, d)| difficulty + d.as_u64());
+            .fold(0u128, |difficulty, (_, d)| difficulty + d.as_u64() as u128);
 
-        let ave_difficulty = difficulty as f64 / n as f64;
+        let ave_difficulty = difficulty / n as u128;
```
delta1: Isn't `n` already cast to `u128` on line 45?

SWvheerden: yip should have removed this as soon as I declared n as u128

stringhandler:
Nice, please squash commits and I'll merge

PR #2967, opened 2021-05-24
[base-node] Create new orphan chain tip on chain reorg

delta1:
## Description

Updates the `handle_possible_reorg` function to insert a new orphan chain tip from the removed blocks, and a test to check this logic. Some minor typos and comments updated.

## Motivation and Context

On reorgs where blocks were removed, the new orphan chain tip was not added to the database. 

## How Has This Been Tested?

New unit test that was failing before the fix.

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1265,6 +1265,8 @@ fn check_for_valid_height<T: BlockchainBackend>(db: &T, height: u64) -> Result<(
     Ok((tip_height, height < pruned_height))
 }
 
+/// Removes blocks from the db from current tip to specified height.
+/// Returns the blocks removed, ordered from tip to height.
 fn rewind_to_height<T: BlockchainBackend>(
     db: &mut T,
     mut height: u64,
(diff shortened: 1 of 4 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1512,9 +1514,17 @@ fn handle_possible_reorg<T: BlockchainBackend>(
 
     let num_added_blocks = reorg_chain.len();
     let removed_blocks = reorganize_chain(db, block_validator, fork_height, &reorg_chain)?;
-
     let num_removed_blocks = removed_blocks.len();
 
+    if let Some(block) = removed_blocks.first() {
```
stringhandler: I think this makes more sense just before the last line of `reorganize_chain`. That method is already managing when the db transactions are written, so it's easier to find, and also easier think of `reorganize_chain` as a single DB operation

stringhandler (commented):
Just one change, then should be fine to go in

base_layer/core/src/chain_storage/blockchain_database.rs
delta1: Sounds good, done

stringhandler (approved):

PR #2973, opened 2021-05-28
Increment wallet key manager index during recovery

philipr-za:
## Description
Currently, when recovery is performed the key manager index is not incremented as outputs are found. This means that when new transactions are performed the key manager produces previously used spending keys for the new outputs, reusing private keys. We also did not apply the correct script private key to the recovered output. This wasn’t a problem because we are using `Nop` scripts but for any script that requires this key to be correct we would have had a problem.

This PR adds the logic that when an output is rewound during recovery that we find the key manager index for the recovered spending key, update the key manager index to one more than the highest index found during recovery and also use this index to apply the correct script private key to the recovered output.

This PR also contains a but of neatness refactoring in the Output Manager Service.

## How Has This Been Tested?
Existing tests

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/handle.rs
+++ b/base_layer/wallet/src/output_manager_service/handle.rs
@@ -23,9 +23,9 @@
 use crate::{
     output_manager_service::{
         error::OutputManagerError,
-        protocols::txo_validation_protocol::TxoValidationType,
         service::Balance,
         storage::{database::PendingTransactionOutputs, models::KnownOneSidedPaymentScript},
+        tasks::TxoValidationType,
         TxId,
     },
     types::ValidationRetryStrategy,
@@ -71,7 +71,7 @@ pub enum OutputManagerRequest {
     RemoveEncryption,
     GetPublicRewindKeys,
     FeeEstimate((MicroTari, MicroTari, u64, u64)),
-    RewindOutputs(Vec<TransactionOutput>, u64),
+    ScanForRecoverableOutputs(Vec<TransactionOutput>, u64),
     ScanOutputs(Vec<TransactionOutput>, u64),
     UpdateMinedHeight(u64, u64),
     AddKnownOneSidedPaymentScript(KnownOneSidedPaymentScript),
@@ -103,7 +103,7 @@ impl fmt::Display for OutputManagerRequest {
             GetCoinbaseTransaction(_) => write!(f, "GetCoinbaseTransaction"),
             GetPublicRewindKeys => write!(f, "GetPublicRewindKeys"),
             FeeEstimate(_) => write!(f, "FeeEstimate"),
-            RewindOutputs(_, _) => write!(f, "RewindAndImportOutputs"),
+            ScanForRecoverableOutputs(_, _) => write!(f, "ScanForRecoverableOutputs"),
             ScanOutputs(_, _) => write!(f, "ScanRewindAndImportOutputs"),
             UpdateMinedHeight(_, _) => write!(f, "UpdateMinedHeight"),
             AddKnownOneSidedPaymentScript(_) => write!(f, "AddKnownOneSidedPaymentScript"),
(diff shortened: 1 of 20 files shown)


delta1 (changes requested):
Looks great! The update to `package.json` means that you'll have to commit `package-lock.json` as part of this change, `npm ci` installs directly from the lockfile which is out of date now

philipr-za:
> Looks great! The update to `package.json` means that you'll have to commit `package-lock.json` as part of this change, `npm ci` installs directly from the lockfile which is out of date now

Done

delta1 (approved):
LGTM

base_layer/wallet/src/output_manager_service/master_key_manager.rs
```diff
@@ -0,0 +1,222 @@
...
+        let mut skm = self.coinbase_script_key_manager.lock().await;
+        let script_key = skm.next_key()?;
+        Ok((spending_key.k, script_key.k))
+    }
+
+    /// Return the Seed words for the current Master Key set in the Key Manager
+    pub async fn get_seed_words(&self) -> Result<Vec<String>, OutputManagerError> {
+        Ok(from_secret_key(
+            self.utxo_key_manager.lock().await.master_key(),
+            &MnemonicLanguage::English,
```
SWvheerden: nit: Should this not rather be a config option for `MasterKeyManager`. 
This seems like a smell, hard-coded option within a function. 
I would rather say let `MasterKeyManager` have a field language and hard code it in there for now.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -0,0 +1,153 @@
...
+        Self {
+            master_key_manager,
+            factories,
+            db,
+        }
+    }
+
+    /// Attempt to rewind all of the given transaction outputs into unblinded outputs. If they can be rewound then add
+    /// them to the database and increment the key manager index
+    pub async fn scan_and_recover_outputs(
```
SWvheerden, line 70:
```suggestion
    pub async fn scan_recover_and_add_outputs(
```

This function now scans, recovers, and adds the outputs it finds. 
I am not completely for this as this function now does more than one thing, but we should at least update the name to reflect that.

SWvheerden (commented):
Looks good, 2 nits thou.

base_layer/wallet/src/output_manager_service/master_key_manager.rs
philipr-za: Fair enough, gave this function a argument to specify the language and added it to the Output Manager Config and its a default now.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
philipr-za: I will argue that the verb `recover` is the process of adding it to the database. The method does to two things, `scans` which is going through and trying to rewind an output and `recover` which is adding it to the db and updating the keymanager. I thought about splitting the second part out but decided that I couldn't think of a use case for every allowing these to happen in isolation and that the interface should only allow them to be done together.

delta1: I agree with @philipr-za , "recover" here already means "recover for spending" which means add them to your wallet.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -103,7 +103,7 @@ impl fmt::Display for OutputManagerRequest {
             GetCoinbaseTransaction(_) => write!(f, "GetCoinbaseTransaction"),
             GetPublicRewindKeys => write!(f, "GetPublicRewindKeys"),
             FeeEstimate(_) => write!(f, "FeeEstimate"),
-            RewindOutputs(_, _) => write!(f, "RewindAndImportOutputs"),
+            ScanForRecoverableOutputs(_, _) => write!(f, "ScanAndImportRecoverableOutputs"),
```
delta1: just noticed these don't match, not sure it's an issue though

philipr-za: fixed

PR #2979, opened 2021-05-31
Fix cucumber error introduced by full script in transaction output

philipr-za:
This PR updates the transaction builder in the Cucumber helpers to calculate the Beta hash and the Transaction Output hash correctly now that the whole script is included in the output.

## How Has This Been Tested?
PR updates tests

## Checklist:

### DIFF
--- a/integration_tests/helpers/transactionBuilder.js
+++ b/integration_tests/helpers/transactionBuilder.js
@@ -64,7 +64,7 @@ class TransactionBuilder {
   }
 
   addInput(input) {
-    let nop_script_bytes = Buffer.from([0x73]);
+    let nopScriptBytes = Buffer.from([0x73]);
     let scriptPublicKey = tari_crypto.pubkey_from_secret(
       input.scriptPrivateKey.toString("hex")
     );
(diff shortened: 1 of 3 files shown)


integration_tests/helpers/transactionBuilder.js
```diff
@@ -125,9 +125,10 @@ class TransactionBuilder {
     let scriptOffsetPublicKey = tari_crypto.pubkey_from_secret(
       scriptOffsetPrivateKey.toString("hex")
     );
+    let nop_script_bytes = Buffer.from([0x73]);
 
     let beta = calculateBeta(
-      nopScriptHash,
+      nop_script_bytes,
```
delta1: I think we prefer `camelCase` for JS, and if lint doesn't complain then it should! Maybe I'm just bikeshedding, but everything else is camelCase (I think).

philipr-za: fair enough

delta1: personally I prefer snake_case but 🤷‍♂️

stringhandler (approved):

PR #2982, opened 2021-06-01
Fix flaky test when syncing small nodes.

Cifko:
## Description
The test was flaky, because it compared 5 additional blocks on one node vs 7 additional block on another node and expected the longer chain to "win". But that's not always the case. Sometime the shorter chain had higher accumulated difficulty.
The fix set the maximum difficulty to 1 for the shorter chain. So when the nodes sync it's always the longer chain with higher accumulated difficulty.


SWvheerden (approved):
LGTM

PR #2983, opened 2021-06-01
Update console wallet when one sided payment is received

stringhandler:
## Description

- Publish event on transaction import
- Update console wallet when event fires
- Updated genesis block because it can't sync with current network
- Reduce one sided payment scanning time default to 1 minute

## How Has This Been Tested?

Manual testing

## Checklist:

### DIFF
--- a/common/config/presets/tari_config_example.toml
+++ b/common/config/presets/tari_config_example.toml
@@ -119,6 +119,8 @@ base_node_query_timeout = 120
 # (options: "DirectOnly", "StoreAndForwardOnly", DirectAndStoreAndForward". default: "DirectAndStoreAndForward").
 #transaction_routing_mechanism = "DirectAndStoreAndForward"
 
+scan_for_utxo_interval=60
+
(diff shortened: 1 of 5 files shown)


common/config/presets/tari_config_example.toml
```diff
@@ -119,6 +119,8 @@ base_node_query_timeout = 120
 # (options: "DirectOnly", "StoreAndForwardOnly", DirectAndStoreAndForward". default: "DirectAndStoreAndForward").
 #transaction_routing_mechanism = "DirectAndStoreAndForward"
 
+scan_for_utxo_interval=60
```
hansieodendaal, line 122: In my opinion, this is much too frequent. I think a really aggressive setting would be 10 minutes.

hansieodendaal:
I propose we also change the wire mode on this one

common/config/presets/tari_config_example.toml
stringhandler: My view is to use this short duration to help debug performance and other problems, and we can increase it later

PR #2987, opened 2021-06-03
Improvement: Remove beta range-proof calculation 

SWvheerden:
## Description

This PR removes the beta calculation for the RangeProof. 

## Motivation and Context

In RFC 201, TariScript, the beta RangeProof calculation was used to prevent malleability. This is now removed as per the new updated RFC. This is to be replaced in a future PR with a signature signed by the sender. 

## How Has This Been Tested?

All critical cucumber test still pass
All running Unit test pass except:
`chain_storage::blockchain_database::test::test_handle_possible_reorg_target_difficulty_is_correct_case_1`
`chain_storage::blockchain_database::test::test_handle_possible_reorg_target_difficulty_is_correct_case_2`
which currently fail on tari_script branch.  

## Checklist:

* [ ] I'm merging against the `development` branch.
* [x I have squashed my commits into a single commit.

PR #2988, opened 2021-06-04
Revert #2987 to bring back beta in the range proof

hansieodendaal:
## Description
- This PR reverts removing of beta in the range proof as was suggested in the [Tari Script review (https://gist.github.com/DavidBurkett/b2ace786c6298179607b1337c3657a78).
- @mikethetike picked up that the receiver now has no "formal" say in what the spending script attached to a UTXO should be, thus this suggestion to bring back beta. Reading this change in conjunction with the new sender signature as proposed in #2956, we now have: _"Alice and Bob agree about the script; Alice commits to it in the sender signature and Bob commits to it when he calculates the range proof."_

## Motivation and Context
See bove.

## How Has This Been Tested?
Reverted back to commit `7172a02`

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -72,6 +72,7 @@ use tari_core::{
     },
 };
 use tari_crypto::{
+    common::Blake256,
     inputs,
     keys::{DiffieHellmanSharedSecret, PublicKey as PublicKeyTrait, SecretKey},
     range_proof::REWIND_USER_MESSAGE_LENGTH,
@@ -1255,13 +1256,21 @@ where TBackend: OutputManagerBackend + 'static
                     .map(|v| (v, output.features, output.script, output.script_offset_public_key))
             })
             .map(|(output, features, script, script_offset_public_key)| {
+                let beta_hash = Blake256::new()
+                    .chain(script.as_bytes())
+                    .chain(features.to_bytes())
+                    .chain(script_offset_public_key.as_bytes())
+                    .result()
+                    .to_vec();
+                let beta = PrivateKey::from_bytes(beta_hash.as_slice())
+                    .expect("Should be able to construct a private key from a hash");
                 // TODO actually increment the keymanager with found private keys so that we don't reuse private keys
                 // once recovery is complete and use the correct script private key
                 UnblindedOutput::new(
                     output.committed_value,
-                    output.blinding_factor.clone(),
+                    output.blinding_factor.clone() - beta,
                     Some(features),
-                    script,
+                    script!(Nop),
                     inputs!(PublicKey::from_secret_key(&output.blinding_factor)),
                     height,
                     output.blinding_factor,
@@ -1327,11 +1336,19 @@ where TBackend: OutputManagerBackend + 'static
                 let rewound =
                     output.full_rewind_range_proof(&self.resources.factories.range_proof, &rewind_key, &blinding_key);
 
+                let beta_hash = Blake256::new()
+                    .chain(output.script.as_bytes())
+                    .chain(output.features.to_bytes())
+                    .chain(output.script_offset_public_key.as_bytes())
+                    .result()
+                    .to_vec();
+                let beta = PrivateKey::from_bytes(beta_hash.as_slice())?;
+
                 if rewound.is_ok() {
                     let rewound_output = rewound.unwrap();
                     rewound_outputs.push(UnblindedOutput::new(
                         rewound_output.committed_value,
-                        rewound_output.blinding_factor.clone(),
+                        rewound_output.blinding_factor.clone() - beta,
                         Some(output.features),
                         known_one_sided_payment_scripts[i].script.clone(),
                         known_one_sided_payment_scripts[i].input.clone(),
--- a/base_layer/core/src/transactions/coinbase_builder.rs
+++ b/base_layer/core/src/transactions/coinbase_builder.rs
@@ -223,11 +223,14 @@ mod test {
             CoinbaseBuilder,
         },
     };
+    use digest::Digest;
     use rand::rngs::OsRng;
     use tari_crypto::{
         commitment::HomomorphicCommitmentFactory,
+        hash::blake2::Blake256,
         keys::SecretKey as SecretKeyTrait,
         script::{ExecutionStack, TariScript},
+        tari_utilities::ByteArray,
     };
 
     fn get_builder() -> (CoinbaseBuilder, ConsensusManager, CryptoFactories) {
@@ -336,7 +339,7 @@ mod test {
             .with_nonce(p.nonce.clone())
             .with_spend_key(p.spend_key.clone())
             .with_rewind_data(rewind_data);
-        let (tx, _) = builder
+        let (tx, unblinded_output) = builder
             .build(rules.consensus_constants(42), rules.emission_schedule())
             .unwrap();
         let block_reward = rules.emission_schedule().block_reward(42) + 145 * uT;
@@ -344,9 +347,16 @@ mod test {
         let rewind_result = tx.body.outputs()[0]
             .full_rewind_range_proof(&factories.range_proof, &rewind_key, &rewind_blinding_key)
             .unwrap();
+        let beta_hash = Blake256::new()
+            .chain(unblinded_output.script.as_bytes())
+            .chain(unblinded_output.features.to_bytes())
+            .chain(unblinded_output.script_offset_public_key.as_bytes())
+            .result()
+            .to_vec();
+        let beta = PrivateKey::from_bytes(beta_hash.as_slice()).unwrap();
         assert_eq!(rewind_result.committed_value, block_reward);
         assert_eq!(&rewind_result.proof_message, proof_message);
-        assert_eq!(rewind_result.blinding_factor, p.spend_key);
+        assert_eq!(rewind_result.blinding_factor, p.spend_key + beta);
     }
 
     #[test]
--- a/base_layer/core/src/transactions/helpers.rs
+++ b/base_layer/core/src/transactions/helpers.rs
@@ -37,6 +37,7 @@ use crate::transactions::{
     types::{Commitment, CommitmentFactory, CryptoFactories, PrivateKey, PublicKey, Signature},
     SenderTransactionProtocol,
 };
+use digest::Digest;
 use num::pow;
 use rand::rngs::OsRng;
 use std::sync::Arc;
@@ -48,6 +49,7 @@ use tari_crypto::{
     range_proof::RangeProofService,
     script,
     script::TariScript,
+    tari_utilities::ByteArray,
 };
 
 /// Create a random transaction input for the given amount and maturity period. The input ,its unblinded
@@ -405,7 +407,18 @@ pub fn create_utxo(
     let features = features.unwrap_or_default();
     let commitment = factories.commitment.commit_value(&keys.k, value.into());
     let offset_pub_key = PublicKey::from_secret_key(&keys.k);
-    let proof = factories.range_proof.construct_proof(&keys.k, value.into()).unwrap();
+    // let construct beta for the utxo
+    let beta_hash = Blake256::new()
+        .chain(script.as_bytes())
+        .chain(features.to_bytes())
+        .chain(offset_pub_key.as_bytes())
+        .result()
+        .to_vec();
+    let beta = PrivateKey::from_bytes(beta_hash.as_slice()).unwrap();
+    let proof = factories
+        .range_proof
+        .construct_proof(&(&keys.k + &beta), value.into())
+        .unwrap();
 
     let utxo = TransactionOutput::new(features, commitment, proof.into(), script.clone(), offset_pub_key);
     (utxo, keys.k, offset_private_key.k)
--- a/base_layer/core/src/transactions/transaction.rs
+++ b/base_layer/core/src/transactions/transaction.rs
@@ -52,7 +52,8 @@ use std::{
     ops::Add,
 };
 use tari_crypto::{
-    commitment::HomomorphicCommitmentFactory,
+    commitment::{HomomorphicCommitment, HomomorphicCommitmentFactory},
+    hash::blake2::Blake256,
     keys::{PublicKey as PublicKeyTrait, SecretKey},
     range_proof::{
         FullRewindResult as CryptoFullRewindResult,
@@ -265,6 +266,14 @@ impl UnblindedOutput {
     }
 
     pub fn as_transaction_output(&self, factories: &CryptoFactories) -> Result<TransactionOutput, TransactionError> {
+        let beta_hash = Blake256::new()
+            .chain(self.script.as_bytes())
+            .chain(self.features.to_bytes())
+            .chain(self.script_offset_public_key.as_bytes())
+            .result()
+            .to_vec();
+        let beta = PrivateKey::from_bytes(beta_hash.as_slice())
+            .map_err(|e| TransactionError::ConversionError(e.to_string()))?;
         let commitment = factories.commitment.commit(&self.spending_key, &self.value.into());
         let output = TransactionOutput {
             features: self.features.clone(),
@@ -272,7 +281,7 @@ impl UnblindedOutput {
             proof: RangeProof::from_bytes(
                 &factories
                     .range_proof
-                    .construct_proof(&self.spending_key, self.value.into())?,
+                    .construct_proof(&(self.spending_key.clone() + beta), self.value.into())?,
             )
             .map_err(|_| TransactionError::RangeProofError(RangeProofError::ProofConstructionError))?,
             script: self.script.clone(),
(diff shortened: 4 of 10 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1255,13 +1256,21 @@ where TBackend: OutputManagerBackend + 'static
                     .map(|v| (v, output.features, output.script, output.script_offset_public_key))
             })
             .map(|(output, features, script, script_offset_public_key)| {
+                let beta_hash = Blake256::new()
+                    .chain(script.as_bytes())
+                    .chain(features.to_bytes())
+                    .chain(script_offset_public_key.as_bytes())
```
stringhandler, line 1262: Not sure if this is still necessary to keep in

stringhandler (approved):
utACK 

Looks good

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: Yip, maybe not.

hansieodendaal:
~~But.... I still do not quite understand how a third party will validate that the correct beta is part of the Bulletproof, or that the published commitment is not actually inflating the currency because they are verifying the proof for something else.~~
```
C     = -vH +          kG           .... [published]
C_hat =  vH + (k + beta)G           .... [verified]
```

Ok.. got it. the verifier can calculate beta and add the two commitments
```
C_hat = C + (0H + betaG)
      = (vH + kG) + (0H + betaG)
      = vH + (k + beta)G
```

SWvheerden (commented):
I am totally against this. 

This gives us a more complicated proof for zero security gain, 
This Is based no the idea that Alice cannot change the script after she and Bob creates a tx. But this does nothing to prevent it as Alice can just create a new tx double-spending the previous one.

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: But this will enhance the contract between Alice and Bob

stringhandler:
To be clear, the attack that exists here (before PR):

After Bob has returned the transaction, including his script and kernel signature to Alice, Alice can change the script on Bob's output to any value, create a new ScriptOffset key, signature and gamma and publish that transaction to the blockchain.

I believe the solution presented in this PR is to prevent that because Alice cannot change the script and still create a valid rangeproof without knowing the blinding factor in the commitment.

@SWvheerden I don't understand the double spend method you are referring to, maybe you can elaborate on it?

SWvheerden:
> e attack that exists here (before PR):

Because Alice owns the commitment. She can do what she wants. We can try to stop Alice from changing the transaction we can, but we cannot stop her from creating a new transaction. So if Alice and Bob create a tx, and Bob locks the script in the UTXO, Bob still does not have the final tx. Alice has to provide him with a final signature.  

But even after all of this. If Alice does not want the tx to go through as it is, she can create a double-spend tx respending her original UTXO to a new UTXO and the other TX will not be mined and Bob still has no funds. To me, this is not an attack. It is almost like me telling you I will pay you some money but I never pay you. 

The only guarantee Bob has to get funds from Alice is if the tx has been mined. Before that Bob has no guarantee and no money. If Alice wants to modify the tx to put some script on it that Bob cannot spend, it's Akin to Alice create a double-spend tx, and Bob still has no money. All Alice did was cost herself some fee money. 

This is also the reason RFC states in the multi-party section that Bob also waits for a mined tx before he starts the whole agreement. Alice can do more than one thing to stop him from spending the tx. But if Bob can verify from a mined tx that the UTXO has the correct script, he knows they can proceed. As with this whole thing, Alice wants something from Bob, she is not giving home money for free. If she "gives" it to him in some way he cannot spend it, then Bob does not do what Alice wants.

philipr-za:
> To be clear, the attack that exists here (before PR):
> 
> After Bob has returned the transaction, including his script and kernel signature to Alice, Alice can change the script on Bob's output to any value, create a new ScriptOffset key, signature and gamma and publish that transaction to the blockchain.

I do not believe this is really an attack.

The fact is that the funds at risk belong to the Sender and the Receiver has nothing to lose until they have the completed transaction in their hands. Up to that point they must just consider that they have received nothing as yet. Even without TariScript if the Receiver replies to a transaction negotiation they have zero guarantee that the Sender will complete the final transaction and give it to them. They can do nothing with the Receiver half of the transaction until then.

So when it comes to the script that binds the output they are about to receive does it matter that it’s malleable by the Sender till the end? Sure the parties could agree to a script that will bind the Receiver output first, then do the dance and then the Sender changes the script before handing the completed transaction back to the Receiver but so what? That is the same thing as the Sender burning those funds to spending them elsewhere. The Receiver has not lost anything. Adding Beta does nothing to stop a non-cooperative Sender from pretending to transact with you and then just do something else.

hansieodendaal:
Consider the Tari blog post if we write this up:

_"Alice and Bob agree about the script. Alice creates the script and then commits to it in the sender signature and Bob has to wait until the transaction is mined to confirm if Alice honoured the conditions."_

vs.

_"Alice and Bob agree about the script; Alice commits to it in the sender signature and Bob commits to it when he calculates the range proof."_

philipr-za:
> Consider the Tari blog post if we write this up:
> 
> _"Alice and Bob agree about the script. Alice creates the script and then commits to it in the sender signature and Bob has to wait until the transaction is mined to confirm if Alice honoured the conditions."_
> 
> vs.
> 
> _"Alice and Bob agree about the script; Alice commits to it in the sender signature and Bob commits to it when he calculates the range proof."_

I think it's a dangerous mentality to have any view except the first example. Until that transaction is on the blockchain to a certain depth is the *only* time that Bob can *know* the conditions have been honoured. Any assumption about that fact before that point is fallible to countless games.

stringhandler:
I've drawn up a table here to highlight the different approaches. I am not necessarily in favour of `beta` being added to the rangeproof, but I think there is a problem here when the script and features are malleable. There may be other options, like Bob signing the script and features with a comsig or other sig that can't be forged.

But let's compare this approach to standard mimblewimble with no scripts. The "double spend" referenced by @SWvheerden is the same as Alice just aborting or rejecting the transaction, so I've only included that scenario.

**Standard MW** refers to MW in the original spec and also Tari's pre-script implementation
**MW with malleable script and features**  refers to the approach @SWvheerden and @philipr-za are proposing of not implementing some kind of malleablity prevention
**MW with nonmalleable script and features** refers to either this PR's approach of using the rangeproof to prevent malleablity or to use another method to prevent it.

 | Standard MW | MW with malleable Tari Script and features | MW with nonmalleable Tari Script and features | 
 | --- | --- | --- |
 | **1. Alice sends kernel to Bob** |||
 | a. Bob can reject tx | a. Bob can reject | a. Bob can reject |
 | **2. Bob creates utxo and sends to Alice** | **2. Bob creates UTXO and script and sends to Alice** ||
 | a. Alice can reject or abort tx | a. Alice can reject or abort tx | a. Alice can reject or abort tx |
 | . | b. Alice can continue and change script and features | . |
 | **3. Alice finishes signing and publishes to blockchain** |||
 
I would argue from this table that non-malleable scripts and features is closer to the original mimblewimble approach and that if we want to allow malleable scripts and features, __there should be a justification why we are now allowing that__.

---

Someone familiar with Tari's code base will note that the output features are already malleable in the standard mw implementation, but I believe the general consensus is that this is incorrect and should be fixed. In fact the [RFC 0201](https://github.com/tari-project/tari/blob/development/RFC/src/RFC-0201_TariScript.md) states: 
> Note: Currently, the output features are actually malleable. TariScript fixes this.

If we go with the middle approach, at the very least, this line should be removed from the RFC.

This statement from the RFC would also have to be corrected:
> As long as those conditions are suitably committed to, are not malleable throughout the existence of the UTXO, and one can prove that the script came from the UTXO owner, then these conditions are not that different to the requirement of having range proofs attached to UTXOs, which require that the value of Tari commitments is non-negative.

philipr-za:
My argument is really that I think your model laid out above is missing a final critical step that is needed with or without the inclusion of beta which is added to solve the malleability of the script that is possible by the Sender after step 2. No matter how steps 1 thru 3 are organized the Receiver MUST only consider the output as theirs after they have:
1. Found it mined on the blockchain to a certain depth.
2. Confirmed that it is still formed as they expect it to be.

A receiver must not make any assumptions before this point because, beyond malleability of the script, there are other possible games a malicious sender could play. So with or without the inclusion of Beta we must still adhere to this strict philosophy for every protocol between two parties.

While I agree it adds some rigidity during the transaction dance between two parties if we strictly apply the rule that a Receiver only considers an output theirs when it meets the two condition I lay out above then does Beta add any actual security?

> > Note: Currently, the output features are actually malleable. TariScript fixes this.
> 
> If we go with the middle approach, at the very least, this line should be removed from the RFC.

The malleability in this case is not referring to malleability during the negotiation of the transaction but the fact that output features can be changed by a miner or the Receiver AFTER the transaction has been completed by the Sender and put in the mempool. Again, it's the Sender's funds that are being spent at this point so they are the final arbiter of the transaction. We need to protect against the transaction being changed by a party OTHER than the sender until it is mined.

> This statement from the RFC would also have to be corrected:
> 
> > As long as those conditions are suitably committed to, are not malleable throughout the existence of the UTXO, and one can prove that the script came from the UTXO owner, then these conditions are not that different to the requirement of having range proofs attached to UTXOs, which require that the value of Tari commitments is non-negative.

I would only add an extra detail to this statement which is `the existence of the UTXO *in the blockchain*`. Even in this statement the UTXO needs to come from the OWNER which is the Sender in a transaction, not the receiver. Only after the transaction is in the immutable ledger does this fact change.

stringhandler:
@philipr-za I agree that a receiver should only see their funds as received when they are spendable, meaning it must be mined and all of the conditions in the script must be met. Wallets should be written in a way to display this correctly to the user. 

My problem is that if the script is malleable by anyone other than Bob, he has no say in it. In fact, he should always assume that Alice will chose a NO-OP script, and not send any script to Alice. He can only be certain of the script when he is also the sender, so if he wanted to include a script on a utxo, he would have to spend it to himself. 

This has a few implications:
1. There will be more transactions on the blockchain to achieve the same result.
2. Transactions are more identifiable. e.g. transactions with scripts are most likely self spends, no ops are most likely different people
3. Interactions are harder to audit for vulnerabilities. For example, atomic swaps are now much more complicated to think about because we need to consider the case where Alice changes the script at the last minute.

Lastly, I think the onus is on the parties wanting to add malleability to scripts and features to justify the divergence from MW. I have not yet seen an argument that provides any kind of benefit for Bob not having the ability to lock this down

philipr-za:
I think a difference of opinion we have is that Bob should have any "say" in the nature of the script that Alice attaches to the output she is giving to him. Alice is the owner of the input and they are the final arbiter on how that input becomes a new output on the blockchain. If Alice does this is a way that Bob doesn't expect then they have not transferred anything to Bob. I guess Alice could try assert to a third party "but I included a spending key that Bob generated so it's his now" but a third party won't be able to verify that without asking Bob if Alice is telling the truth and Bob can just say "Nope". So in that case Alice has just tainted one of her outputs that Bob may or may not spend under certain conditions but she hasn't adhered to the protocol so Bob just doesn't accept it and has lost nothing in the exchange.

Attaching a script to the output is something Alice wants to do, not Bob. Bob would way prefer a conditionless Nop script than any kind of condition on how the output is given to him. I think this is the other perspective that I feel we are getting backwards in this discussion.

I mean lock it down with Beta, it adds some more rigidity to the exchange between Alice and Bob but I am trying to make sure that we don't consider this commitment during the dance as a reason to change our outlook that Bob only owns an output once its on the blockchain. 

I think it's also important for us to come to a consensus on the power dynamic between a Sender and Receiver during the various stages of a transaction because it is going to colour our approach to these designs in the future.

CjS77:
I tend to agree with @SWvheerden and @philipr-za here, but playing Devil's advocate for a moment:

Can we find a class of instances where Bob _actually_ has something to lose by Alice changing the script so that:
* Bob is worse off than if the script was never published (i.e. Alice witholding)
* Bob is worse off than if the original script was used (the base case)

This is not strictly the same as a pure double spend as mentioned above because Bob might be providing information that Alice can use to her advantage.

A contrived example might be that Bob reveals some data if the tx is published, but Alice modifies the script so that the info is revealed but does not pay Bob.

My hunch is that even in such cases, there are precautions that Bob can take to prevent the worse-off instances (such as using Atomic swaps) rather than try and do a multiparty commitment to the script.

This is obviously not a proof in either direction, but rather an attempt to frame the debate a little more clearly.

philipr-za:
I agree that if the Receiver has something to lose in the transaction that can be linked to the script on an output then this is indeed an issue. The current formulation of the MW negotiation doesn't allow for the Receiver to add an input to the transaction, it is not a trivial thing to add. In a multi-party transaction there will be many rounds of commitment and verification by all the parties before the final signing.

But anyway, the only thing this PR costs us is additional complexity in the cryptography. If this extra complexity is worth it for more rigidity in the initial exchange then that is fine. However, what it doesn't do is change the threat model of the Receiver at all. With or without this in place I feel it's extremely important for the Receiver to view the Sender in absolute suspicion until the output is immutable on the blockchain. The way the Receiver looks for outputs on the blockchain is by requesting their hash from a base node and if the script is modified that hash is no longer the same. So it still feels that even our current code in the wallet and base node would catch this script modification with literally no further work.

hansieodendaal:
> ...... But anyway, the only thing this PR costs us is additional complexity in the cryptography. If this extra complexity is worth it for more rigidity in the initial exchange then that is fine. However, what it doesn't do is change the threat model of the Receiver at all. With or without this in place I feel it's extremely important for the Receiver to view the Sender in absolute suspicion until the output is immutable on the blockchain. 

Agreed

> ...... The way the Receiver looks for outputs on the blockchain is by requesting their hash from a base node and if the script is modified that hash is no longer the same. So it still feels that even our current code in the wallet and base node would catch this script modification with literally no further work.

I think this is a flaw as well in our wallet code; identification of a UTXO should be based on the value commitment alone. This is how wallet recovery works in any case; Bulletproof rewinding only looks at the ability to open the value commitment inside a UTXO, not at its hash.

hansieodendaal:
> I tend to agree with @SWvheerden and @philipr-za here, but playing Devil's advocate for a moment:
> 
> Can we find a class of instances where Bob _actually_ has something to lose by Alice changing the script so that:
> 
> * Bob is worse off than if the script was never published (i.e. Alice witholding)
> * Bob is worse off than if the original script was used (the base case)

There is a number of cases where leaving the script malleable could hurt Bob or both Alice and Bob.

**Who can change the script?**
- Alice change the script after Bob sees it and the transaction gets mined: 
  - Bob's wallet will recognize the transaction as mined, have the output in his wallet, but with a foreign script.
  - Bob has no way to prove he did not agree with the fully signed transaction although he provided a partially signed Mimblewimble signature, value commitment and range proof.
- A 3rd party intercepts the script after Bob sees it and changes it before sending it to Alice and the transaction gets mined: 
  - All of the above.
  - Alice was not aware the script was changed when she finalized the transaction.

**Examples** _(There will definitely be more)_
- Time-locked script centuries into the future: 
  - Bob could be dependent on those funds now.
  - Someone else wants to hurt Bob; his funds will not be spendable.
  - Someone else wants to hurt Alice; she will be out of pocket and Bob will refuse to release the goods she paid for until he can spend the funds.
- `PushPubkey(3_rd_party_pub_key)`
  - The 3rd party could hold Bob's funds ransom and exhort payments from him.

stringhandler:
Let's move the discussion to #2997 which discusses the problem in a general sense, not only the solution of adding beta back

philipr-za:
Look, the only reason I am still arguing this point is because I think these arguments show a difference in views on the roles and power dynamics, which influence the game theory, of the two participants during a transaction negotiation in a decentralized system. I feel it's important for us to at least come to a general consensus on the threat model that the Sender and Receiver need to work under as they interact.

> There is a number of cases where leaving the script malleable could hurt Bob or both Alice and Bob.
> 
> **Who can change the script?**
> 
> * Alice change the script after Bob sees it and the transaction gets mined:
>   
>   * Bob's wallet will recognize the transaction as mined, have the output in his wallet, but with a foreign script.

As our code is written *right now* this is not possible because Bob checks to see that the output is what he expects before he complete the transaction and adds it to the database as spendable. When we validates the output is mined he requests it from the base node using its hash and if the script has changed it will not be returned. This is already built into the system because Bob does not trust Alice.

>   * Bob has no way to prove he did not agree with the fully signed transaction although he provided a partially signed Mimblewimble signature, value commitment and range proof.

This point is a weird one in a decentralized system. Who does he need to prove it to? Why is the onus on him to prove anything? The only other party in this interaction is Alice. If she changes the script and then somehow tries to assert to a 3rd party that "Bob agreed to this" how will the third party verify that? They will have to ask Bob and Bob knows the truth.

> * A 3rd party intercepts the script after Bob sees it and changes it before sending it to Alice and the transaction gets mined:

Alice sets the script not Bob. Again, it is Alice that wants to apply the script conditions to the output, not Bob. Bob would always prefer an output with no conditions so who in the interaction wants to apply the script, it is Alice. So when does a third party tell Alice what the script must be?

Now if the script requires some chosen input from Bob I guess you could argue we could use Beta as the commitment to the parameters that Bob chooses in the script but I honestly think we would still have to add in extra communication rounds to a transaction negotiation of that nature anyway using more traditional signatures.

As I said previously the only cost of this is added complexity. If we feel it's worth it then that is fine. I still contend that Bob will still have to treat the outputs with the same level of suspicion with or without this and that by working on a strict philosophy of "the output can only be trusted once I find it immutably in the blockchain" catches these cases 🤷‍♂️

stringhandler:
Closing this PR in favour of continuing the discussion on #2997

PR #2992, opened 2021-06-07
[wallet] Ensure recovery will not overwrite existing wallet

delta1:
## Description

- wallet given a recovery key should not overwrite an existing wallet master key
- console wallet sending message box was squished 

## Motivation and Context

Before running recovery the console wallet already checks that it won't overwrite an existing wallet, this check is now included in the wallet library itself. 

## How Has This Been Tested?

new unit test, and tested manually

## Checklist:


stringhandler (approved):
utACK

PR #2993, opened 2021-06-08
Modify gamma calculation for Tari script

hansieodendaal:
## Description
- Modified gamma calculation for Tari script according to the latest RFC-0201 proposals.
- Fixed private keys re-use issue in `transactions/helpers.rs fn create_tx`.
- Clarified use of test params in `transactions/helpers.rs fn spend_utxos`.
- Ensured `consensus_validation_large_tx` will fail if the transaction internal consistency is not ok (gamma is calculated here as well).

## Motivation and Context
See above.

## How Has This Been Tested?
Cucumber & unit tests.

## Checklist:


stringhandler (approved):
utACK

philipr-za (approved):

philipr-za:
reviewed but untested

PR #2995, opened 2021-06-08
Fix reorg difficulty unit tests

philipr-za:
## Description
The changes in PR #2960 changed the LWMA difficulty calculation to be integer based. This changed the nature of the calculation for small values. The two broken tests affected by this change were:
`test_handle_possible_reorg_target_difficulty_is_correct_case_1`
`test_handle_possible_reorg_target_difficulty_is_correct_case_2`

These tests relied on assumptions about the difficulty calculated by LWMA based on the block times specified in the tests. Just updating the test to use much higher difficulties would have meant the change to integer math would not have had such a breaking effect but the way the unit tests have to mine for new hashes based on the difficulties meant that the difficulties needed to be kept as low as possible.

This PR tweaks the block times in these tests in order to produce varying difficulties but keeping them low enough for the unit test environment.

## How Has This Been Tested?
Fixes to tests

## Checklist:


delta1 (approved):
LGTM, tested

stringhandler (approved):
utACK

PR #2991, opened 2021-06-07
RFC 0132 Monero Merge Mining

sdbondi:
## Description

RFC to describe the current protocol for monero merge mining.
[Rendered](https://demo.hedgedoc.org/s/y1dCJecba)

Note: there are fairly minor but consensus breaking differences in the transaction proof 
and other structs that will be PRd on the tari script branch.  

## Motivation and Context

RFC describing merge mining protocol were missing. Part of investigations into using a minimal merkle proof.

## How Has This Been Tested?

N/A

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I have squashed my commits into a single commit.

PR #2996, opened 2021-06-09
Mininal merkle proof for monero pow data

sdbondi:
## Description

- Adds a port of Monero's `tree_branch` that returns a merkle proof
- Extensive unit tests for create_merkle_proof
- Uses a merkle proof of the transaction hash set instead of the full
  transaction hash set.
- Adds more extensive unit tests for tree_hash
- tree_hash now has at worst one allocation
- RandomX seed is represented in a fixed size array instead fo a
  unbounded String
- Use monero's consensus de/encoding for MoneroPowData
- Updated monero-rs to 0.13.0
- Removed monero-rs dep from merge mining proxy in favour of encapsulating it in the monero_rx module

This is a consensus breaking change - the genesis timestamp has been updated.

## Motivation and Context

Proof size for monero merge mining is many bytes larger than necessary.   
This is an implementation of RFC 132 (still pending review) https://github.com/tari-project/tari/pull/2991

Proof size coinbase-only transaction reduced from 457 to 280 bytes (~39% decrease), larger gains will be 
seen for monero blocks with many transactions. 

Further unimplemented size optimisations are possible:
- The Merkle proof is not needed at all for coinbase only transactions because the hash of the coinbase is the equivalent proof
- The depth field (4-bytes) is not strictly needed because it is always equal to the number of hashes included in the proof, except in the coinbase-only case  
- 3 bytes can be saved in the FixedByteArray, since the randomx key is a maximum of 60 bytes
- 2 bytes for the transaction count, which is only necessary at mining time (i.e. can be thrown away after producing the blockhashing_blob )

## How Has This Been Tested?

A number of additional unit tests using mainnet monero data to verify the correctness of the merkle tree Rust port. 
Existing tests pass.
Successfully mined blocks on hard fork  

image

## Checklist:


stringhandler (commented):
Nice! Looks good so far, but I haven't tested it. Will test, but clippy and FFI compile still needs to be fixed

base_layer/core/src/proof_of_work/monero_rx/mod.rs
```diff
@@ -741,20 +519,26 @@ mod test {
...
-            total_kernel_offset: RistrettoSecretKey::from(0),
-            total_script_offset: RistrettoSecretKey::from(0),
+            total_kernel_offset: Default::default(),
+            total_script_offset: Default::default(),
             nonce: 0,
             pow: ProofOfWork::default(),
         };
-        let monero_data = MoneroData::default();
-        let serialized = bincode::serialize(&monero_data).unwrap();
+        let monero_data = MoneroPowData {
```
SWvheerden: Is there not a default for this anymore?

SWvheerden (approved):
LGTM

base_layer/core/src/proof_of_work/monero_rx/mod.rs
sdbondi: I'm 50-50 on this, but I didn't see it fitting well - what is a default Transaction? Does a default Merkle Proof make any sense? It was used in one place, so I went with rather not having the concept of Default MoneroPowData (which I can't clearly see as having any meaning - 0/empty initialisation doesn't have a good meaning at the moment - is empty MerkleProof a MerkleProof? )

sdbondi:
Updating the merge mining proxy code which broke because of the monero-rs lib update

hansieodendaal (commented):
Looks really good, well done. (Tested.)

I have one philosophical question: 
Porting Monero code should be more efficient running in Rust (I assume), so how will we keep our port in sync with related changes to Monero? I suspect this is a really small real-world problem.

sdbondi:
@hansieodendaal Since it is consensus critical, I don't see the proof construction changing at all (at least in a way that changes the merkle root). This code would be better off in a shared crate or a library like monero-rs though. I'm AFK for the next week, but after that can perhaps work on a PR for that.

hansieodendaal:
> @hansieodendaal Since it is consensus critical, I don't see the proof construction changing at all (at least in a way that changes the merkle root). This code would be better off in a shared crate or a library like monero-rs though. I'm AFK for the next week, but after that can perhaps work on a PR for that.

Yeah, I also think it should not be a problem though. I PR like you suggest is something that is maybe _desired_ or _could do_.

PR #3004, opened 2021-06-10
Fix `Unique Constraint` bug when requesting a coinbase output at same height

philipr-za:
## Description
If a miner requested a Coinbase Transaction at the same height as one that was requested previously and that output had the exact same commitment the Output Manager database would return a Unique Constraint error due to the existing output in the database. This would happen rarely when a miner is experiencing many reorgs while mining.

To fix this issue logic is added to the Output Manager service to clear any existing coinbase outputs with the same commitment at the same block height as the one that was just requested before adding it to the database. I decided this presents no danger of losing funds because the newly added coinbase will have the exact same commitment as the one being removed.

## How Has This Been Tested?
Unit test provided

## Checklist:


stringhandler (approved):
utACK

Agree that it should be fine to clear it

PR #3008, opened 2021-06-10
Review of RFC-0100

philipr-za:
## Description
This PR contains a small update to RFC-0100 to reflect the current state of the project

## Checklist:

### DIFF
--- a/RFC/src/RFC-0100_BaseLayer.md
+++ b/RFC/src/RFC-0100_BaseLayer.md
@@ -67,9 +67,10 @@ The Tari Base Layer network comprises the following major pieces of software:
   Tari base layer and cryptocurrency. The base nodes validate and transmit transactions and blocks, and maintain
   consensus about the longest valid proof-of-work blockchain.
 * Mining software. Mining nodes perform proof-of-work to secure the base layer and compete to submit the
-  next valid block into the Tari blockchain. Tari is merge-mined with Monero. The Tari source provides two alternatives
-  for Tari miners:
-  * A standalone miner
+  next valid block into the Tari blockchain. Tari uses two Proof of Work (PoW) algorithms, the first is merge-mined with Monero and a second native SHA3 PoW. 
+  The Tari source provides three alternatives for Tari miners:
+  * A standalone miner for SHA3 mining
+  * A merge-mining proxy to be used with XMRig to merge mine Tari with Monero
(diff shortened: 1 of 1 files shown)


RFC/src/RFC-0100_BaseLayer.md
```diff
@@ -67,9 +67,10 @@ The Tari Base Layer network comprises the following major pieces of software:
...
   consensus about the longest valid proof-of-work blockchain.
 * Mining software. Mining nodes perform proof-of-work to secure the base layer and compete to submit the
-  next valid block into the Tari blockchain. Tari is merge-mined with Monero. The Tari source provides two alternatives
-  for Tari miners:
-  * A standalone miner
+  next valid block into the Tari blockchain. Tari uses two Proof of Work (PoW) algorithms, the first is merge-mined with Monero and a second native SHA3 PoW. 
+  The Tari source provides three alternatives for Tari miners:
+  * A standalone miner for SHA3 mining
+  * A merge-mining proxy to be used with XMRig to merge mine Tari with Monero
   * A stratum-compatible pool miner.
```
SWvheerden, line 74: Is this even still on the cards?
@StriderDM  you might know?

SWvheerden (approved):
Looks good

RFC/src/RFC-0100_BaseLayer.md
StriderDM: @SWvheerden As far as I know it is, though it would be two alternatives. First and third bullet would exist within the same application, just as an additional mode for supported pool software. Point 3 is also still under development, so these docs may need to change again in the near future.

StriderDM: Think this can go in as-is for now though :)

delta1 (approved):

RFC/src/RFC-0100_BaseLayer.md
philipr-za: Ja, I didn't know exactly what form these would take so just added what I knew was there and left the stratum miner for the future because that would exist in some form at some point.

StriderDM: We also need to mention monero-pool and nodejs-pool for merge mining as well at some point.

CjS77 (approved):
LGTM

PR #3009, opened 2021-06-10
[base-node] Check minimum number of headers for calc-timing

delta1:
## Description

- Checks for the number of blocks requested when calling `calc-timing`
- Start renaming `calc-timing` to more descriptive `block-timing`, will be done in 2 stages since block explorer must be updated as well.  

## Motivation and Context

`calc-timing 1` would previously return the value for u64::MAX as min block time (which is the initial value in the calculation). It's impossible to know how long 1 block took to mine without the timestamp of the previous block.

```
>> calc-timing 1
Timing for blocks #2975 - #2975
Max block time: 0
Min block time: 18446744073709551615
Avg block time: 0
```

## How Has This Been Tested?

- New rust tests added. 
- Command tested manually in base node. 
- GRPC call for `CalcTiming` tested with bloomRPC to ensure it's still working.

## Checklist:


stringhandler (approved):
utACK.

Looks good.

PR #3012, opened 2021-06-11
Fix double spend test.

Cifko:
The problem was that TransactionInput has now new fields, including signature which is never equal because of the random nonce.

## Description
Change the PartialEq trait for TransactionInput to only compare commitment (UTXO).

### DIFF
--- a/base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
+++ b/base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
@@ -168,8 +168,10 @@ impl UnconfirmedPool {
     fn find_duplicate_input(array_of_tx: &[Arc<Transaction>], tx: &Arc<Transaction>) -> bool {
         for transaction in array_of_tx {
             for input in transaction.body.inputs() {
-                if tx.body.inputs().contains(input) {
-                    return true;
+                for tx_input in tx.body.inputs() {
+                    if tx_input.commitment == input.commitment {
+                        return true;
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


delta1:
Is this for the `Mempool.feature` "Double spend" integration test?

delta1 (approved):
LGTM, tested the "Double spend" integration test which is now working 👍 

@philipr-za can you give a 👀 please?

stringhandler (commented):
I think this is definitely problem, so nice find, but I don't agree with the solution you have proposed.

I think this is better solved by fixing the code  in the mempool to look at the commitment instead of whether inputs are equal. Two inputs are not strictly equal if they are spending the same commitment, because there may be two inputs spending the same output, but with entirely different script inputs. 
For example, consider [the timelocked example in rfc](https://rfc.tari.com/RFC-0202_TariScriptOpcodes.html#time-locked-contract). In this case there may be an input from Bob and an input from Alice that take different script inputs and signatures to spend the input.

SWvheerden (approved):
While this might I would caution against this way as we are now saying that 2 `TransactionInputs` are equal if their commitments are the same. 

For example, if we have 2 `TransactionInput` with the same commitment, and change the one's feature flags, we would expect that `TransactionInput_1 =! TransactionInput_2`. 
But with this PR `TransactionInput_1 == TransactionInput_2`

If this is the problem for the double-spend, perhaps we should create another method in TransactionInput to only do this comparison or manually do the comparison in the mempool where this occurs.

delta1:
Cool, that's the feedback I was looking for thanks @mikethetike and @SWvheerden.

delta1:
Sorry wrong button

Cifko:
Ok, I put back the derive trait of PartialEq for TransactionInput. And I changed how mempool find duplicates. The test is ok. Can you please confirm if this is the way to go? Thanks.

stringhandler (approved):
utACK

Nice

delta1 (approved):
utACK

PR #3013, opened 2021-06-14
Add RFC-0173 - Versioning to the RFCs

philipr-za:
## Description
This PR contains a new RFC with a proposal of how we will handle versioning of the wire format, P2P messaging and consensus rules on the Tari network.

## How Has This Been Tested?
Just docs

## Checklist:

### DIFF
--- a/RFC/src/RFC-0173_Versioning.md
+++ b/RFC/src/RFC-0173_Versioning.md
@@ -1,0 +1,133 @@
+# RFC-0173/Versioning
+
+## Versioning
+
+![status: draft](theme/images/status-draft.svg)
+
+**Maintainer(s)**: [Philip Robinson](https://github.com/philipr-za)
+
+# Licence
+
+[The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright 2021 The Tari Development Community
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
+
+1. Redistributions of this document must retain the above copyright notice, this list of conditions and the following
+   disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+   derived from this software without specific prior written permission.
+
+THIS DOCUMENT IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS", AND ANY EXPRESS OR IMPLIED WARRANTIES,
+INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+## Language
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", 
+"NOT RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in 
+[BCP 14](https://tools.ietf.org/html/bcp14) (covering RFC2119 and RFC8174) when, and only when, they appear in all capitals, as 
+shown here.
+
+## Disclaimer
+
+This document and its content are intended for information purposes only and may be subject to change or update
+without notice.
+
+This document may include preliminary concepts that may or may not be in the process of being developed by the Tari
+community. The release of this document is intended solely for review and discussion by the community of the
+technological merits of the potential system outlined herein.
+
+## Goals
+
+The aim of this Request for Comment (RFC) is to describe the various types of versioning that nodes on the Tari network 
+will use during interaction with other nodes.
+
+## Related Requests for Comment
+
+- [RFC-0710: Tari Communication Network and Network Communication Protocol](RFC-0170_NetworkCommunicationProtocol.md)
+- [RFC-0171: MessageSerialization](RFC-0171_MessageSerialisation.md)
+
+## Description
+
+In a decentralized system the set of nodes on the network will run a variety of software versions as time goes on. Some 
+of these versions will be compatible and others not. For example, if a crucial consensus change is added during a hard 
+fork event. Furthermore, there will be multiple networks running Tari code, i.e. Mainnet vs Testnet. Versioning refers 
+to the strategies we will use for nodes to determine if they can communicate.
+
+Tari will contain three different versioning schemes:
+1. WireMode is the first byte a peer sends when connecting to another peer, used to identify the network and/or protocol bytes that follow 
+2. P2P message versions that will accompany every P2P message,
+3. Consensus rules versions that will be exchanged on connection and are included in each block header.
+
+### WireMode byte
+In the Bitcoin P2P protocol messages are preceded by 4 
+[magic values](https://en.bitcoin.it/wiki/Protocol_documentation#Common_structures) or bytes. These values are used to 
+delimit when a new message starts in a byte stream and also are used to indicate which of the Bitcoin networks the node 
+is speaking on, such as TestNet or Mainnet.
+
+Tari message packets are encapsulated using the Noise protocol so we do not need the delimiting functionality of these 
+bytes but Tari will include a single WireMode byte at the beginning of every connection session. This byte will indicate 
+which network a node is communicating on, so that if the counterparty is on a different network it can reject this 
+connection cheaply without having to perform any further operations, like completing the Noise protocol handshake.
+
+The following is a proposed mapping of the WireMode byte. Space is left between Mainnet, Stagenet and Testnet bytes for
+future use. 
+```rust,ignore
+   #[repr(u8)]
+   enum Network {
+      Mainnet = 0x00,
+      Stagenet1 = 0x51,
+      Testnet1 = 0xa1,
+      Testnet2 = 0xa2
+   }
+```
+### P2P message version
+Peer to Peer messages on the Tari network are encapsulated into message envelopes. The body of message envelopes are 
+defined, serialized and deserialized using Protobuf. These messages will only be updated by adding new fields to the 
+Protobuf definitions, never removing fields. This is done in order to preserve backwards compatibility where newer nodes 
+can still communicate with older nodes. 
+
+The P2P messaging protocol will see many changes in its lifetime. Some will be minor changes that are fully backwards 
+compatible and some changes will be breaking where older nodes will not be able to communicate with newer nodes. In 
+order to document these two classes of changes each P2P message header will contain a `version` field that will use
+a two-part semantic versioning scheme with the format of `major.minor` integer versions. The `minor` version will be 
+incremented whenever there is any change. The `major` version be incremented when there is a breaking change made to 
+the P2P protocol. Each integer can be stored separately.
+
+### Consensus version
+The final aspect of the Tari system that will be versioned are the Consensus rules. These rules will change as the 
+network matures. Changes to consensus rules can be achieved using either a Soft fork or Hard fork. Soft forks are where 
+new consensus rules are added that older nodes will see as valid (thus backwards compatible) but newer nodes will reject 
+blocks from older nodes that are not aware of the new consensus rules. A hard fork means that the new consensus rules 
+are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new 
+transactions and blocks.
+
+The consensus version will be used by a node to determine if it can interact with another node successfully or not. A 
+list of fork versions will be maintained within the code. When a connection is started with a new node the two nodes 
+will exchange `Version` messages detailing the consensus version they are each running and the blockheight at which they
+are currently operating. Both nodes will need to reply with a `Version Acknowledge` message to confirm that they are 
+compatible with the counterparty's version. It is possible for a newer node to downgrade its protocol to speak to an 
+older node so this must be decided during this handshake process. Only once the acknowledgments have been exchanged can 
+further messages be exchanged by the parties. This is the method currently employed on the 
+[Bitcoin network](https://developer.bitcoin.org/devguide/p2p_network.html#connecting-to-peers)
+
+For example, if we have two nodes, Node A and Node B, where Node A is ahead of Node B in version and block height. 
+During the handshake Node B will not recognize Node A's version but should wait for Node A to reject or confirm the
+connection because Node A could potentially downgrade their version to match Node B's. Node A will speak to Node B if
+and only if Node A recognizes Node B's version and Node B's block height is in the correct range for its advertised 
+version according to Node A's fork version list.
+
+Tari Block Headers contain a `version` field which will be used to indicate the version of consensus rules that are 
+used in the construction and validation of this block. Consensus rules versions will only consist of breaking changes
+and as such will be represented with a single incremented integer. This coupled with the internal list of fork versions,
+that includes the height at which they came into effect, will be used to validate whether the consensus rules specified 
+in the header are valid for that block's height.
--- a/RFC/src/SUMMARY.md
+++ b/RFC/src/SUMMARY.md
@@ -15,6 +15,7 @@
     - [RFC-0170: Network Communication Protocol](RFC-0170_NetworkCommunicationProtocol.md)
       - [RFC-0171: Message Serialisation](RFC-0171_MessageSerialisation.md)
       - [RFC-0172: Peer to Peer Messaging Protocol](RFC-0172_PeerToPeerMessagingProtocol.md)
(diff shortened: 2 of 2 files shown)


RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+This document and its content are intended for information purposes only and may be subject to change or update
+without notice.
+
+This document may include preliminary concepts that may or may not be in the process of being developed by the Tari
+community. The release of this document is intended solely for review and discussion by the community of the
+technological merits of the potential system outlined herein.
+
+## Goals
+
+The aim of this Request for Comment (RFC) is to describe the various types of versioning that nodes on the Tari network will use while interaction with other nodes.
```
delta1:
```suggestion
The aim of this Request for Comment (RFC) is to describe the various types of versioning that nodes on the Tari network will use during interaction with other nodes.
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+1. A WireMode byte that precede every low level network packet,
+2. P2P message versions that will accompany every P2P message,
+3. Consensus rules versions that will be exchanged on connection and are included in each block header.
+
+### WireMode byte
+In the Bitcoin P2P protocol messages are preceded by 4 [magic values](https://en.bitcoin.it/wiki/Protocol_documentation#Common_structures) or bytes. These values are used to 
+delimit when a new message starts in a byte stream and also are used to indicate which of the Bitcoin networks the node is speaking on, such as TestNet or Mainnet.
+
+Tari message packets are encapsulated using the Noise protocol so we do not need the delimiting functionality of these bytes but Tari will include a single WireMode byte at the beginning
+of every connection session. This byte will indicate which network a node is communicating on, so that if the counter-party is on a different network it can reject this connection cheaply without
```
delta1:
```suggestion
of every connection session. This byte will indicate which network a node is communicating on, so that if the counterparty is on a different network it can reject this connection cheaply without
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+is not true. As the P2P messaging protocol evolves fields might fall out of use. In order to keep track of which fields are currently being used, each P2P message header contains a `version` field. 
+As major changes are made to the messaging formats this version will increase to allow clients to know if they are able to properly deserialize these messages.
+
+### Consensus version
+The final aspect of the Tari system that will be versioned are the Consensus rules. These rules will change as the network matures. Changes to consensus rules can be achieved using either a Soft fork or Hard fork. 
+Soft forks are where new consensus rules are added that older nodes will see as valid (thus backwards compatible) but newer nodes will reject blocks from older nodes that are not aware of the new consensus rules. 
+A hard fork mean that the new consensus rules are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new transactions and blocks.
```
delta1:
```suggestion
A hard fork means that the new consensus rules are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new transactions and blocks.
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+
+The consensus version will be used by a node to determine if it can interact with another node successfully or not. A list of fork versions will be maintained within the code. When a connection is 
+started with a new node the two nodes will exchange `Version` messages detailing the consensus version they are each running. Both nodes will need to reply with a `Version Acknowledge` message to confirm that they are
+compatible with the counter-party's version. Only once the acknowledgments have been exchanged can further messages be exchanged by the parties. This is the method currently employed on
```
delta1:
```suggestion
compatible with the counterparty's version. Only once the acknowledgments have been exchanged can further messages be exchanged by the parties. This is the method currently employed on 
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+A hard fork mean that the new consensus rules are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new transactions and blocks.
+
+The consensus version will be used by a node to determine if it can interact with another node successfully or not. A list of fork versions will be maintained within the code. When a connection is 
+started with a new node the two nodes will exchange `Version` messages detailing the consensus version they are each running. Both nodes will need to reply with a `Version Acknowledge` message to confirm that they are
+compatible with the counter-party's version. Only once the acknowledgments have been exchanged can further messages be exchanged by the parties. This is the method currently employed on 
+the [Bitcoin network](https://developer.bitcoin.org/devguide/p2p_network.html#connecting-to-peers)
+
+Tari Block Header's contain a `version` field which will be used to indicate the version of consensus rules that are used in the construction and validation of this block. This coupled with the internal list of
```
delta1:
```suggestion
Tari Block Headers contain a `version` field which will be used to indicate the version of consensus rules that are used in the construction and validation of this block. This coupled with the internal list of
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+fork versions, that will include the height at which they came into force, will be used to validate whether the consensus rules specified in the header are valid for block's height.
```
delta1:
```suggestion
fork versions, that include the height at which they came into effect, will be used to validate whether the consensus rules specified in the header are valid for that block's height.
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
+# RFC-0171/MessageSerialization
```
delta1:
```suggestion
# RFC-0173/Versioning
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Message Serialization
```
delta1:
```suggestion
## Versioning
```

delta1 (changes requested):
Nice one, just a few improvements!

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+delimit when a new message starts in a byte stream and also are used to indicate which of the Bitcoin networks the node is speaking on, such as TestNet or Mainnet.
+
+Tari message packets are encapsulated using the Noise protocol so we do not need the delimiting functionality of these bytes but Tari will include a single WireMode byte at the beginning
+of every connection session. This byte will indicate which network a node is communicating on, so that if the counter-party is on a different network it can reject this connection cheaply without
+having to perform any further operations, like completing the Noise protocol handshake.
+
+### P2P message version
+Peer to Peer messages on the Tari network are encapsulated into message envelopes. The body of message envelopes are defined, serialized and deserialized using Protobuf. These messages will only be updated by adding 
+new fields to the Protobuf definitions, never removing fields. This is done in order to preserve backwards compatibility where newer nodes can still communicate with older nodes, but the reverse
```
SWvheerden: I may be wrong here, but as far as I have it you can delete Protobuf fields. You just should not reuse the field numbers.

SWvheerden (dismissed):
Looks good

RFC/src/RFC-0173_Versioning.md
philipr-za: Ja, I think you are right.

philipr-za: Though by removing the field you make Protobuf not backwards compatible when deserializing into a generated Rust struct. Should these messages keep the old fields to be able to speak to older nodes?

delta1: Yeah as I understand it you want to keep them for backwards compatibility?

delta1 (dismissed):
LGTM

RFC/src/RFC-0173_Versioning.md
CjS77: Here's a blog on the topic:
https://www.beautifulcode.co/blog/88-backward-and-forward-compatibility-protobuf-versioning-serialization

We could use explicit version numbers; or handle missing fields in code, which would be a strong indicator of difference versions (but not a guarantee of course). 

In the former case, do you propose integer versioning, or semver-type? If you make a small change to a message that is 100% backwards compatible (wherein the new field will have the default value for the type) and doesn't require any special handling, do you inc the version? Strictly speaking, one should, right? But it feels like overkill.

On the otherhand semver-type versioning feels like a lot of overhead here.

Maybe the compromise is major.minor, where minor is inc with _any_ change, and major with any breaking change.
Then if peer A receives a msg from B where A.version.major > B/major.version, it could (possibly) drop down to the older version if it can, or esle reject the message.
Alternatively, if B get a message from A, it must reject.

The minor version indicator is simply there to mark that the messages aren't 100% identical, but doesn't result in message rejection.

My feeling is that trying to manage versions by dealing with missing info as in the blog post is fraught with unforeseen circumstances and is best avoided.

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,94 @@
...
+Soft forks are where new consensus rules are added that older nodes will see as valid (thus backwards compatible) but newer nodes will reject blocks from older nodes that are not aware of the new consensus rules. 
+A hard fork means that the new consensus rules are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new transactions and blocks.
+
+The consensus version will be used by a node to determine if it can interact with another node successfully or not. A list of fork versions will be maintained within the code. When a connection is 
+started with a new node the two nodes will exchange `Version` messages detailing the consensus version they are each running. Both nodes will need to reply with a `Version Acknowledge` message to confirm that they are
+compatible with the counterparty's version. Only once the acknowledgments have been exchanged can further messages be exchanged by the parties. This is the method currently employed on 
+the [Bitcoin network](https://developer.bitcoin.org/devguide/p2p_network.html#connecting-to-peers)
```
CjS77: Here it makes sense to inc version numbers by full integers, and not maintain a minor version.

I suggest you add a paragraph that makes the handling explicit,

Node A will talk to Node B  IFF A recognizes the version AND B.blockheight is in the range where that version was active, according to A's fork version list. This means that B shouldn't reject A immediately; it should wait for A to reject him (since A might be able to downgrade) before killing the connection.

RFC/src/RFC-0173_Versioning.md
stringhandler: The [Language Guide[(https://developers.google.com/protocol-buffers/docs/proto3#updating) recommends renaming them to "OBSOLETE_" instead of deleting.

I like the idea of having a version number in the message header. Since missing fields are rendered as defaults in protobuf 3, nodes can reject newer versions, and handle missing fields in older versions. I think handling defaults in the protobuf conversions is tricky even at the best of times, so care will need to be taken in these parts of code. I also think the changes on the p2p messages will be small enough that semver is not needed and there is a good chance that the minor versions will not be used

RFC/src/RFC-0173_Versioning.md
stringhandler: Agree

RFC/src/RFC-0173_Versioning.md
philipr-za: Wrote up a two-part semantic versioning proposal for the P2P messages

RFC/src/RFC-0173_Versioning.md
philipr-za: Added a paragraph with explicit example and explicitly stated a single incrementing integer

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Description
+
+In a decentralized system the set of nodes on the network will run a variety of software versions as time goes on. Some 
+of these versions will be compatible and others not. For example, if a crucial consensus change is added during a hard 
+fork event. Furthermore, there will be multiple networks running Tari code, i.e. Mainnet vs Testnet. Versioning refers 
+to the strategies we will use for nodes to determine if they can communicate.
+
+Tari will contain three different versioning schemes:
+1. A WireMode byte that precede every low level network packet,
```
sdbondi: This description of WireMode is not correct.

```suggestion
1. WireMode is the first byte a peer sends when connecting to another peer, used to identify the network and/or protocol bytes that follow 
```

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,123 @@
...
+### WireMode byte
+In the Bitcoin P2P protocol messages are preceded by 4 
+[magic values](https://en.bitcoin.it/wiki/Protocol_documentation#Common_structures) or bytes. These values are used to 
+delimit when a new message starts in a byte stream and also are used to indicate which of the Bitcoin networks the node 
+is speaking on, such as TestNet or Mainnet.
+
+Tari message packets are encapsulated using the Noise protocol so we do not need the delimiting functionality of these 
+bytes but Tari will include a single WireMode byte at the beginning of every connection session. This byte will indicate 
+which network a node is communicating on, so that if the counterparty is on a different network it can reject this 
+connection cheaply without having to perform any further operations, like completing the Noise protocol handshake.
```
sdbondi, line 80: Ideally, we should define the new WireMode mapping.

Currently these are hardcoded values in comms denoting a "mode", all of which are valid (in the case of liveness, the byte is always recognised even if it is disabled). 

Essentially we need to move the mapping to the domain layer and allow it to pass the valid byte to comms.   

I suggest the existing `Network` enum is expanded to contain that mapping. 
e.g.
```rust
#[repr(u8)]
enum Network {
   Mainnet = 0x00,
   Testnet1 = 0x01,
   Testnet2 = 0x02,
}

```

It is probably worthwhile to define that mapping here

sdbondi (commented):
Looks good, one correction and one possible addition

RFC/src/RFC-0173_Versioning.md
philipr-za: Sounds good, which Testnet's should be included? Stibbons and Weatherwax?

philipr-za: @mikethetike ^

stringhandler: I think we should give some space to version testnets and stagenets from mainnets.

Perhaps:
```rust
#[repr(u8)]
enum Network {
   Mainnet = 0x00,
   Stagenet1 = 0x51,
   Testnet1 = 0xa1,
    Testnet2 = 0xa2
}
```

stringhandler: Stibbons and Weatherwax are fine as examples

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,122 @@
...
+### P2P message version
+Peer to Peer messages on the Tari network are encapsulated into message envelopes. The body of message envelopes are 
+defined, serialized and deserialized using Protobuf. These messages will only be updated by adding new fields to the 
+Protobuf definitions, never removing fields. This is done in order to preserve backwards compatibility where newer nodes 
+can still communicate with older nodes. 
+
+The P2P messaging protocol will see many changes in its lifetime. Some will be minor changes that are fully backwards 
+compatible and some changes will be breaking where older nodes will not be able to communicate with newer nodes. In 
+order to document these two classes of changes each P2P message header will contain a `version` field that will use
+a two-part semantic versioning scheme with the format of `major.minor` integer versions. The `minor` version will be
```
sdbondi, line 102: Not sure about the use of a string here. There are minor overheads in dealing with a string:
- an extra byte for length vs 1/2 varint fields
- possible for large strings to be sent
- parsing / splitting of strings

I suggest using 2 varint fields `major` and `minor`

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,122 @@
...
+compatible and some changes will be breaking where older nodes will not be able to communicate with newer nodes. In 
+order to document these two classes of changes each P2P message header will contain a `version` field that will use
+a two-part semantic versioning scheme with the format of `major.minor` integer versions. The `minor` version will be 
+incremented whenever there is any change. The `major` version be incremented when there is a breaking change made to 
+the P2P protocol.
+
+### Consensus version
+The final aspect of the Tari system that will be versioned are the Consensus rules. These rules will change as the 
+network matures. Changes to consensus rules can be achieved using either a Soft fork or Hard fork. Soft forks are where 
+new consensus rules are added that older nodes will see as valid (thus backwards compatible) but newer nodes will reject
```
sdbondi, line 109: newer nodes won't always reject blocks?

RFC/src/RFC-0173_Versioning.md
philipr-za: I pictured it being stored as two bytes. I wasn't trying to specify the data types used to store it in the system.

philipr-za: I mention that we should store the two values a separate ints.

RFC/src/RFC-0173_Versioning.md
```diff
@@ -0,0 +1,133 @@
...
+The final aspect of the Tari system that will be versioned are the Consensus rules. These rules will change as the 
+network matures. Changes to consensus rules can be achieved using either a Soft fork or Hard fork. Soft forks are where 
+new consensus rules are added that older nodes will see as valid (thus backwards compatible) but newer nodes will reject 
+blocks from older nodes that are not aware of the new consensus rules. A hard fork means that the new consensus rules 
+are not backwards compatible and so only nodes that know about the new rules will be able to produce and validate new 
+transactions and blocks.
+
+The consensus version will be used by a node to determine if it can interact with another node successfully or not. A 
+list of fork versions will be maintained within the code. When a connection is started with a new node the two nodes 
+will exchange `Version` messages detailing the consensus version they are each running and the blockheight at which they
```
sdbondi, line 116: The Tari protocol already has provision for this - we should probably describe that here.

sdbondi (commented):
Looks good - a similar protocol from bitcoins already exists and should be described (see https://www.tari.com/lessons/06_tari_comms_i.html).

Perhaps we should have these sections:

`Network` - single byte that enumerates the target network (mainnet, testnet etc)
`Protocol version` - each protocol (identity (similar to bitcoins version protocol), messaging, each RPC service etc) negotiates a version before starting (usually in a half round trip)
`Consensus` - version on the block header

RFC/src/RFC-0173_Versioning.md
philipr-za: Sure, but the implication is that if they are not aware of the new rules that when they break them the blocks will be rejected. However old nodes will never reject blocks from new nodes because the fork is backwards compatible.

sdbondi: Ok fair enough, perhaps I would say (up to you if you want to change it - otherwise just resolve), in the second part of the sentence, that newer nodes may perform new validations and reject incoming blocks.

stringhandler (approved):

PR #3016, opened 2021-06-17
Feat: Change script_signature type to ComSig

SWvheerden:
## Description

This swaps the script signature in the input to a type ComSig from a normal Schnorr signature. 

## Motivation and Context

RFC 201 details that the script signature should be of type Commitment signature.

## How Has This Been Tested?

Manually tested transaction sending. 
All unit and integration tests pass. 

## Checklist:


integration_tests/helpers/util.js
```diff
@@ -68,10 +69,10 @@ function dec2hex(n) {
...
 
 function toLittleEndianInner(n) {
-  const hexar = dec2hex(n);
-  return hexar
-    .map((h) => (h < 16 ? "0" : "") + h.toString(16))
-    .concat(Array(4 - hexar.length).fill("00"));
+  let hexar = dec2hex(n);
+  hexar =  hexar.map((h) => (h < 16 ? "0" : "") + h.toString(16));
+  if (hexar.length < 4) {  return   hexar.concat(Array(4 - hexar.length).fill("00"));}
+  else {return hexar;}
```
delta1: You need to run `npm run fmt` and `npm run lint` from the integration_tests directory and fix up any issues

applications/tari_app_grpc/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2018"
 tari_common_types = { version = "^0.8", path = "../../base_layer/common_types"}
 tari_core = {  path = "../../base_layer/core"}
 tari_wallet = {  path = "../../base_layer/wallet"}
-tari_crypto = { git = "ssh://git@github.com/tari-project/tari-crypto.git", branch = "main" }
+tari_crypto = { git = "https://github.com/SWvheerden/tari-crypto.git", rev = "3e8817" } #switch back to official after merge #switch back to official after merge
```
delta1: Looks like you're using an old commit on your branch here? just use `branch = "sw_comsig"` 

image

stringhandler (approved):
utACK

Happy, let's see what happens with CI

stringhandler:
Ok I see I need to publish a new tari_crypto wasm

base_layer/core/src/transactions/transaction.rs
```diff
@@ -244,18 +245,27 @@ impl UnblindedOutput {
...
+            .chain(PublicKey::from_secret_key(&self.script_private_key).as_bytes())
+            .chain(commitment.as_bytes())
             .result()
             .to_vec();
-
-        let script_signature = Signature::sign(self.script_private_key.clone(), script_nonce, &e)
-            .map_err(|_| TransactionError::InvalidSignatureError("Generating script signature".to_string()))?;
+        let script_signature = ComSignature::sign(
+            self.value.into(),
+            self.script_private_key.clone() + self.spending_key.clone(),
```
DavidBurkett, line 262: Are `script_private_key` and `spending_key` always generated by the same individual? If not, we've got to be very careful we're not introducing any opportunities here for rogue key attacks.

SWvheerden: If we are sticking with MW rules, yes. This is specifically to force the user who generated both to be the same individual, as there are attacks that someone not being the spender be able to change the data here. 

For example:
Alice spends her UTXO with a NOP script on it, which allows any input on it. Basically mimicking vanilla MW behavior. Bon can then take Alice's input and change the input data and thus change the `script_private_key`. If Bob cab changes this, he can the spending conditions on his utxo.  This is discussed a bit more here in detail: https://github.com/tari-project/tari/blob/tari-script/RFC/src/RFC-0201_TariScript.md#script-offset-security

stringhandler (approved):
Looks good. untested. 

Please try to not delete package.json in future

PR #3018, opened 2021-06-18
Fix block_event_and_reorg_event_handling test.

Cifko:
## Description
This PR fix 3 issues that were failing:
1) The coinbase_lock_height is 2, so we had to add one empty block to be able to process the UTXO.
2) The TXs were discarded because their input was invalid. Solution was to move it after the processing of the block1.
3) Removal of assert_eq, there is proper one async_assert_eventually.


Cifko:
Is there a better way to add empty block?

base_layer/core/tests/mempool.rs
```diff
@@ -754,7 +754,7 @@ fn service_request_timeout() {
 #[ignore = "Flaky test that needs to be fixed"]
```
delta1: If the test is passing reliably now then you can remove this `#[ignore]` attribute so it can run on CI 👍

Cifko: done. thanks.

base_layer/core/tests/mempool.rs
```diff
@@ -751,10 +751,9 @@ fn service_request_timeout() {
 }
 
 #[test]
-#[ignore = "Flaky test that needs to be fixed"]
 #[allow(clippy::identity_op)]
 fn block_event_and_reorg_event_handling() {
-    // #flaky, this test seems to fail after submiting block B2A to bob.
+    // #flaky, this test seems to fail after submitting block B2A to bob.
```
delta1: Sorry I forgot to ask to remove this comment as well!

delta1 (dismissed):
Tested locally working, nice one.

base_layer/core/tests/mempool.rs
Cifko: My bad, my spellchecker just noticed that mistake, next time I will read the comments more carefully.

delta1: No worries 🙃

delta1 (approved):

Issue #2997, opened 2021-06-09
The sender can change the script of a UTXO before submitting to the blockchain


stringhandler:
## Problem:
In the negotiation of the transaction between Alice and Bob, when Bob sends the UTXO to Alice, Alice can change the script before signing the UTXO with the script offset key.

## History:
This is a continuation of a discussion on PR #2988.
In the initial version of Tari script, the script and other parts of the UTXO was included in the rangeproof as `beta`. Due to another vulnerability noted in https://gist.github.com/DavidBurkett/b2ace786c6298179607b1337c3657a78, a signature by Alice was added to the UTXO, and `beta` was removed. After it was removed, it is now possible for Alice to change the script before she signs it.

## Possible solutions:
1. Add beta back as suggested by PR #2988
2. Add a new signature, possibly a comsig by Bob that signs the script before sending it to Alice
3. Other suggestions?

stringhandler:
I may be misrepresenting @SWvheerden and @philipr-za on this, but their view in the discussion on #2988 is that this is not an issue, and we can allow it.

Some counter points to that:
1. If the protocol/RFC does not strictly forbid Alice from changing the script, it should be viewed that it is an allowed interaction for her to change it. It should not be ambiguous whether this is allowed or disallowed, because wallets should know whether to accept this transaction or not. 
2. In standard MW, Bob already can decide whether to receive the funds in a single UTXO or multiple UTXOs, if the fees would allow it (perhaps the mempool is empty), and Alice only has the power to continue with the transaction or abort. She has no power to alter, remove or combine the UTXOs Bob has chosen. If we allow Alice to change that, the power dynamic has shifted. I would suggest it makes more sense to leave the power of crafting the UTXO in it's entirety to Bob, and leaving Alice with the power of continuing or aborting
3. The original MW and tari script implementations before `beta` was removed had gone through rounds of review and this was not possible, so allowing Alice to change the script now feels like a regression
4. Since Alice has the power to change the script, Bob should never suggest a script at all. This could lead to the situation that the only times scripts are used is in one-sided payments and self spends. This makes interactions that could be simplified by scripts, much more complicated and involve more steps.

[...]

PR #3020, opened 2021-06-18
Add sender signature to transaction output

hansieodendaal:
## Description
- Added sender signature to the transaction output and unblinded output.
- Added sender signature validation to the transaction consistency check.
- Added unit test to verify sender signature.
- Added sender signature to cucumber (transaction builder).
- Simplified test code making use of unblinded outputs.
- Updated genesis block timestamp due to breaking changes.

**TODO for Future PR** - Update Wallet FFI _import UTXO_ interface (`pub unsafe extern "C" fn wallet_import_utxo`) to coincide with mobile wallet update.

## Motivation and Context
Updated to be in line with the latest RFC-0201

## How Has This Been Tested?
- Unit testing
- Cucumber integration testing
- Running base nodes, wallets & miners

## Checklist:

### DIFF
--- a/base_layer/core/src/transactions/transaction.rs
+++ b/base_layer/core/src/transactions/transaction.rs
@@ -34,6 +34,7 @@ use crate::transactions::{
         CommitmentFactory,
         CryptoFactories,
         HashDigest,
+        MessageHash,
         PrivateKey,
         PublicKey,
         RangeProof,
@@ -200,7 +201,7 @@ pub enum TransactionError {
 
 /// An unblinded output is one where the value and spending key (blinding factor) are known. This can be used to
 /// build both inputs and outputs (every input comes from an output)
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct UnblindedOutput {
     pub value: MicroTari,
     pub spending_key: BlindingFactor,
@@ -210,6 +211,7 @@ pub struct UnblindedOutput {
     pub height: u64,
     pub script_private_key: PrivateKey,
     pub script_offset_public_key: PublicKey,
+    pub sender_metadata_signature: Signature,
 }
 
 impl UnblindedOutput {
@@ -224,6 +226,7 @@ impl UnblindedOutput {
         height: u64,
         script_private_key: PrivateKey,
         script_offset_public_key: PublicKey,
+        sender_metadata_signature: Signature,
     ) -> UnblindedOutput {
         UnblindedOutput {
             value,
@@ -234,6 +237,7 @@ impl UnblindedOutput {
             height,
             script_private_key,
             script_offset_public_key,
+            sender_metadata_signature,
         }
     }
 
@@ -277,6 +281,7 @@ impl UnblindedOutput {
             .map_err(|_| TransactionError::RangeProofError(RangeProofError::ProofConstructionError))?,
             script: self.script.clone(),
             script_offset_public_key: self.script_offset_public_key.clone(),
+            sender_metadata_signature: self.sender_metadata_signature.clone(),
         };
         // A range proof can be constructed for an invalid value so we should confirm that the proof can be verified.
         if !output.verify_range_proof(&factories.range_proof)? {
@@ -311,6 +316,7 @@ impl UnblindedOutput {
             proof,
             script: self.script.clone(),
             script_offset_public_key: self.script_offset_public_key.clone(),
+            sender_metadata_signature: self.sender_metadata_signature.clone(),
         };
         // A range proof can be constructed for an invalid value so we should confirm that the proof can be verified.
         if !output.verify_range_proof(&factories.range_proof)? {
@@ -506,6 +512,8 @@ pub struct TransactionOutput {
     pub script: TariScript,
     /// Tari script offset pubkey, K_O
     pub script_offset_public_key: PublicKey,
+    /// UTXO signature with the script offset private key, k_O
+    pub sender_metadata_signature: Signature,
 }
 
 /// An output for a transaction, includes a range proof and Tari script metadata
@@ -517,6 +525,7 @@ impl TransactionOutput {
         proof: RangeProof,
         script: TariScript,
         script_offset_public_key: PublicKey,
+        sender_metadata_signature: Signature,
     ) -> TransactionOutput {
         TransactionOutput {
             features,
@@ -524,6 +533,7 @@ impl TransactionOutput {
             proof,
             script,
             script_offset_public_key,
+            sender_metadata_signature,
         }
     }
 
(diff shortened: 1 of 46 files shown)


base_layer/core/src/transactions/transaction.rs
```diff
@@ -200,7 +201,7 @@ pub enum TransactionError {
 
 /// An unblinded output is one where the value and spending key (blinding factor) are known. This can be used to
 /// build both inputs and outputs (every input comes from an output)
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
stringhandler, line 204: This is potentially a problem as it can leak a key, when we are serializing outputs?

base_layer/core/src/transactions/transaction.rs
```diff
@@ -582,6 +611,38 @@ impl TransactionOutput {
     pub fn is_coinbase(&self) -> bool {
         self.features.flags.contains(OutputFlags::COINBASE_OUTPUT)
     }
+
+    /// Convenience function that calculates the challenge for the sender Schnorr signature
+    pub fn build_sender_challenge(
+        script: &TariScript,
+        features: &OutputFeatures,
+        script_offset_public_key: &PublicKey,
+        puplic_nonce: &PublicKey,
```
stringhandler, line 620:
```suggestion
        public_nonce: &PublicKey,
```

base_layer/core/src/transactions/transaction.rs
```diff
@@ -582,6 +611,38 @@ impl TransactionOutput {
...
+        Challenge::new()
+            .chain(script.as_bytes())
+            .chain(features.to_bytes())
+            .chain(script_offset_public_key.as_bytes())
+            .chain(puplic_nonce.as_bytes())
+            .result()
+            .to_vec()
+    }
+
+    /// Create sender signature fore the output meta data
```
stringhandler, line 631:
```suggestion
    /// Create sender signature for the output metadata
```

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -108,22 +108,25 @@ impl SingleReceiverTransactionProtocol {
...
             features,
             commitment,
             RangeProof::from_bytes(&proof)
                 .map_err(|_| TPE::RangeProofError(RangeProofError::ProofConstructionError))?,
             sender_info.script.clone(),
             sender_info.script_offset_public_key.clone(),
-        ))
+            sender_info.sender_metadata_signature.clone(),
+        );
+        let _ = output.verify_sender_signature()?;
```
stringhandler, line 120: Doesn't this method return a unit and therefore can be ignored? `()`
There was another instance of this somewhere else as well. Perhaps an artefact of debugging?
```suggestion
        output.verify_sender_signature()?;
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -883,11 +883,18 @@ struct NewOutputSql {
...
 
 impl NewOutputSql {
-    pub fn new(output: DbUnblindedOutput, status: OutputStatus, tx_id: Option<TxId>) -> Self {
-        Self {
+    pub fn new(
+        output: DbUnblindedOutput,
+        status: OutputStatus,
+        tx_id: Option<TxId>,
+    ) -> Result<Self, OutputManagerStorageError> {
+        let sender_metadata_signature = serde_json::to_string(&output.unblinded_output.sender_metadata_signature)
```
stringhandler, line 895: I see why you did this, but it's probably better to store this as two fields (`sender_metadata_signature_nonce`, `sender_metadata_signature_sig`) in the database and save on the serde hit here

base_layer/wallet/src/wallet.rs
```diff
@@ -309,6 +309,9 @@ where
         source_public_key: &CommsPublicKey,
         features: OutputFeatures,
         message: String,
+        sender_signature: Signature,
```
stringhandler, line 312:
```suggestion
        sender_metadata_signature: Signature,
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4535,10 +4537,14 @@ pub unsafe extern "C" fn wallet_import_utxo(
...
         ExecutionStack::default(),
         &(*source_public_key).clone(),
-        // WARNING, this might be a problem in the future when importing anything else than a default features UTXO,
-        // the FFI function signature should be updated to take this in.
         OutputFeatures::default(),
         message_string,
+        // TODO: Add the actual sender signature here.
+        Signature::default(),
+        // TODO:Add the actual script private key here.
+        &Default::default(),
```
stringhandler, line 4545: This might mean the mobile wallet will be unable to spend this, but can be addressed later

stringhandler (commented):
This is a very large PR, so I think we may need to clean it up and make changes in a follow up PR. 

In general I am happy with the PR and I'm glad it passed the CI first time.
In concept I am not aligned with everyone on this. I don't like the idea of the sender choosing the receiver's script, as I've noted in #2997. In this PR, the sender now also chooses the receiver's output features.  As I said before though, I think let's merge this PR with the way you have done it and then we can add some changes later if needed. After all, there isn't yet a use case in the system where the receiver might want to suggest their own. 

One thing to note though is that you have not checked whether the receiver changed the script or the features in `add_single_recipient_info` of the sender protocol (or whenever the sender checks the recipient data). This means that the sender will only find out when the base node rejects the message, when it has the ability to fail faster.

stringhandler (approved):

base_layer/core/src/transactions/transaction.rs
hansieodendaal: This was done because `pub unblinded_change_output: Option<UnblindedOutput>,`  was added to `struct RawTransactionInfo` to minimize code duplication in other places.  
`struct RawTransactionInfo` already contains a lot of secret information and is dependent on the `Serialize, Deserialize` traits.

base_layer/wallet/src/wallet.rs
hansieodendaal: > One thing to note though is that you have not checked whether the receiver changed the script or the features in `add_single_recipient_info` of the sender protocol (or whenever the sender checks the recipient data). This means that the sender will only find out when the base node rejects the message, when it has the ability to fail faster.

This is done as part of `pub fn validate_internal_consistency(` when a transaction is constructed, together with 

```
        self.verify_kernel_signatures()?;
        self.validate_kernel_sum(total_offset, &factories.commitment)?;

        self.validate_range_proofs(&factories.range_proof)?;
        self.validate_sender_signatures()?;
        self.validate_script_offset(script_offset_g)
```

hansieodendaal:
> One thing to note though is that you have not checked whether the receiver changed the script or the features in `add_single_recipient_info` of the sender protocol (or whenever the sender checks the recipient data). This means that the sender will only find out when the base node rejects the message, when it has the ability to fail faster.

This is done as part of `pub fn validate_internal_consistency(` when a transaction is constructed, together with 

```
        self.verify_kernel_signatures()?;
        self.validate_kernel_sum(total_offset, &factories.commitment)?;

        self.validate_range_proofs(&factories.range_proof)?;
        self.validate_sender_signatures()?;
        self.validate_script_offset(script_offset_g)
```

PR #3025, opened 2021-06-22
[Docs] Update rfc to add witness_mr to header

SWvheerden:
## Description

This updates RFC 120 to update the header fields and to add an input_mr field. 

## Motivation and Context

There a few out-of-date changes to the header that is not reflected on this RFC, this was fixed. 

This PR adds in a new field, the input_mr. This field ensures that there is no malleability of the script fields and inputs. This is a current vulnerability in TariScript. 

## How Has This Been Tested?

Docs

## Checklist:

### DIFF
--- a/RFC/src/RFC-0120_Consensus.md
+++ b/RFC/src/RFC-0120_Consensus.md
@@ -86,9 +86,13 @@ Every [block header] MUST contain the following fields:
 * prev_hash;
 * timestamp;
 * output_mr;
+* output_mmr_size;
+* witness_mr;
 * range_proof_mr;
 * kernel_mr;
+* kernel_mmr_size;
 * total_kernel_offset;
+* script_kernel_offset;
 * nonce;
 * pow.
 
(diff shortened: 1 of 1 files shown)


RFC/src/RFC-0120_Consensus.md
```diff
@@ -140,13 +144,40 @@ The timestamp MUST confirm to the following:
...
 
+#### Output_mmr_size
+
+This is the total size of the leaves in the output merkle mountain range.
+
+The Output_mmr_size MUST confirm to the following:
+
+* Represented as a single unsigned 64-bit integer.
+
+#### Witness_mr
```
stringhandler: I think this should just be called "Input_mr" and should only include a hash of the input data, and should include all of the fields of the input, so that a proof can be constructed.
There is no need to include the output data because it is already covered by the output_mr and the range proof mr. 

It also should be mentioned that it is a simple merkle tree for this block only and does not need to include previous block's data like the output and kernel mmr's, since it is only used to verify the data in this block has not changed

stringhandler (approved):

RFC/src/RFC-0120_Consensus.md
```diff
@@ -140,18 +144,44 @@ The timestamp MUST confirm to the following:
...
 * The hashing function used MUST be blake2b with a 256 bit digest.
 
-#### Range_proof_mr
+#### Output_mmr_size
 
-This is the merkle root of the range proofs.
+This is the total size of the leaves in the output merkle mountain range.
 
-The range_proof_mr MUST confirm to the following:
+The Output_mmr_size MUST confirm to the following:
```
stringhandler, line 163: Should this not be conform? This error also appears in other places in this document 

```suggestion
The Output_mmr_size MUST conform to the following:
```

stringhandler (approved):
Approved, just a few spelling errors

PR #3026, opened 2021-06-22
Fix small issues related to #3020

hansieodendaal:
## Description
- As per reviewer comments

## Motivation and Context
See above.

## How Has This Been Tested?
Unit tests, cucumber tests

## Checklist:


base_layer/wallet/migrations/2021-06-22-143855_sender_meta_signature/up.sql
```diff
@@ -18,7 +18,8 @@ CREATE TABLE outputs (
     height INTEGER NOT NULL,
     script_private_key BLOB NOT NULL,
     script_offset_public_key BLOB NOT NULL,
-    sender_metadata_signature TEXT NOT NULL,
+    sender_metadata_signature_key BLOB NOT NULL,
```
philipr-za, line 21: Nit: Personally, I don't see the benefit of splitting this field into two. I noted the comment on #3020 but the fact is that Serde is doing the exact same thing as you are going to do manually turning these values into byte arrays. I suppose its a little bit less space in the db because Serde will encode it as a hex string but at the cost of code and db complexity.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -58,6 +58,8 @@ use tari_crypto::{
 
 /// This struct contains all the information that a transaction initiator (the sender) will manage throughout the
 /// Transaction construction process.
+// TODO: Investigate necessity to use the 'Serialize' and 'Deserialize' traits here; this could potentially leak
```
philipr-za, line 61: Could you explain this comment? We have to be able to serialize this info when we store incomplete transactions in the database.

base_layer/wallet/migrations/2021-06-22-143855_sender_meta_signature/up.sql
hansieodendaal: Agreed, I would also opt for less code and db complexity.

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: This comment came about because a variable of type `UnblindedOutput` was added to this struct and it, in turn, imposed the necessity to have 'Serialize' and 'Deserialize' traits on `UnblindedOutput`, which is possibly the bigger problem.

PR #3027, opened 2021-06-22
Remove `height` field from outputs

philipr-za:
## Description
In the discussion around [RFC-0201](https://rfc.tari.com/RFC-0201_TariScript.html) it was decided that the `height` field is not actually required to prevent replay attacks with the introduction of TariScript. We rely on the enforcement of unique commitments in the mempool and blockchain to prevent this. As such this PR removes the height field from the TransactionInput and UnblindedOutput structs.

## How Has This Been Tested?
Existing tests

## Checklist:

### DIFF
--- a/RFC/src/RFC-0201_TariScript.md
+++ b/RFC/src/RFC-0201_TariScript.md
@@ -622,7 +622,6 @@ To summarise, the information required for one-sided transactions is as follows:
 | features          | \\( F_a \\)                           | Public                                                                                        |
 | script            | \\( \alpha_a \\)                      | Public                                                                                        |
 | script input      | \\( \input_a \\)                      | Public                                                                                        |
-| height            | \\( h_a \\)                           | Public                                                                                        |
 | script signature  | \\( a_{Sa},b_{Sa}, R_{Sa} \\)         | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
 | sender offset public key | \\( K_{Oa} \\)                        | Not used in this transaction                                                                  |
 
@@ -742,7 +741,6 @@ To summarise, the information required for creating a multiparty UTXO is as foll
 | features                    | \\( F_a \\)                           | Public                                                                                        |
 | script                      | \\( \alpha_a \\)                      | Public                                                                                        |
 | script input                | \\( \input_a \\)                      | Public                                                                                        |
-| height                      | \\( h_a \\)                           | Public                                                                                        |
 | script signature            | \\( (a_{Sa},b_{Sa}, R_{Sa}) \\)     | Alice knows \\( k_{Sa},\\, r_{Sa} \\) and \\( k_{a},\\, v_{a} \\) of the commitment \\(C_a\\) |
 | sender offset&nbsp;public&nbsp;key | \\( K_{Oa} \\)                        | Not used in this transaction                                                                  |
 
@@ -765,7 +763,6 @@ When spending the multi-party input:
 | features                    | \\( F_s \\)                             | Public                                                                                                                                                             |
 | script                      | \\( \alpha_s \\)                        | Public                                                                                                                                                             |
 | script input                | \\( \input_s \\)                        | Public                                                                                                                                                             |
-| height                      | \\( h_a \\)                             | Public                                                                                                                                                             |
 | script&nbsp;signature       | \\( (a_{Ss} ,b_{Ss} , R_{Ss}) \\)     | Alice knows \\( (k_{SsA},\\, r_{SsA}) \\), Bob knows \\( (k_{SsB},\\, r_{SsB}) \\). Both parties know \\( (k_{s},\\, v_{s}) \\). Neither party knows \\( k_{Ss}\\) |
 | sender offset&nbsp;public&nbsp;key | \\( K_{Os} \\)                          | As above, Alice and Bob each know part of the sender offset key                                                                                                           |
 
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -166,8 +166,6 @@ message TransactionInput {
     bytes script = 4;
     // The script input data, if any
     bytes input_data = 5;
-    // The block height that the UTXO was mined
-    uint64 height = 6;
     // A signature with k_s, signing the script, input data, and mined height
     ComSignature script_signature = 7;
     // The offset public key, K_O
@@ -318,8 +316,6 @@ message UnblindedOutput {
     bytes script = 4;
     // Tari script input data for spending
     bytes input_data = 5;
-    // The block height that the UTXO was mined
-    uint64 height = 6;
     // Tari script private key
     bytes script_private_key = 7;
     // Tari script offset pubkey, K_O
--- a/applications/tari_app_grpc/src/conversions/transaction_input.rs
+++ b/applications/tari_app_grpc/src/conversions/transaction_input.rs
@@ -59,7 +59,6 @@ impl TryFrom<grpc::TransactionInput> for TransactionInput {
             commitment,
             script,
             input_data,
-            height: input.height,
             script_signature,
             script_offset_public_key,
         })
@@ -78,7 +77,6 @@ impl From<TransactionInput> for grpc::TransactionInput {
             hash,
             script: input.script.as_bytes(),
             input_data: input.input_data.as_bytes(),
-            height: input.height,
             script_signature: Some(grpc::ComSignature {
                 public_nonce_commitment: Vec::from(input.script_signature.public_nonce().as_bytes()),
                 signature_u: Vec::from(input.script_signature.u().as_bytes()),
--- a/applications/tari_app_grpc/src/conversions/unblinded_output.rs
+++ b/applications/tari_app_grpc/src/conversions/unblinded_output.rs
@@ -43,7 +43,6 @@ impl From<UnblindedOutput> for grpc::UnblindedOutput {
             }),
             script: output.script.as_bytes(),
             input_data: output.input_data.as_bytes(),
-            height: output.height,
             script_private_key: output.script_private_key.as_bytes().to_vec(),
             script_offset_public_key: output.script_offset_public_key.as_bytes().to_vec(),
             sender_metadata_signature: Some(grpc::Signature {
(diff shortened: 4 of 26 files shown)


philipr-za:
When this PR goes in then we will need to merge https://github.com/tari-project/wallet-grpc-client/pull/20 to keep the Cucumber tests working.

base_layer/wallet/migrations/2021-06-21-141831_remove_height_from_output/down.sql
```diff
@@ -0,0 +1,26 @@
+PRAGMA foreign_keys=off;
+ALTER TABLE outputs RENAME TO outputs_old;
```
hansieodendaal: **NIT:** Just wondering if it is worth it to alter the old table and not just drop it?

hansieodendaal (approved):
LGTM, one small nit.

**Note:** The failing cucumber test (_"Mempool clearing out invalid transactions after a reorg"_) in CircleCI in this latest commit failed due to an unrelated network disconnect issue. It tested fine on my system when I tried it out.

base_layer/wallet/migrations/2021-06-21-141831_remove_height_from_output/down.sql
philipr-za: This is really the better way to do migrations I think. It copies any data from the old table into the new one and then drops the old table. The main reason we didn't do it this way in `2021-03-23-082938_update-outputs-for-tari-script` was because in that migration we were adding fields for which we couldn't create sensible default values in the Sql itself. 

Secondly the drop.sql is not ever actually going to be used but I still felt it worthwhile attempting to respect the migration contract.

stringhandler:
Copied from IRC #tari-dev on Libera:
```
8:31 AM <dburkett> https://github.com/tari-project/tari/pull/3027
8:31 AM <dburkett> So my understanding from this PR is that all output commitments must be retained in an indexed cache indefinitely. And they can never be reused. Is that correct?
10:04 AM <stringhandler> yeah, although nodes running the pruning horizon will not be able to validate it
10:05 AM <stringhandler> the pruning horizon at this point is configurable per node
10:06 AM <stringhandler> archival nodes must keep an index of the commitments though
10:09 AM → delta1 joined (~delta1@105.30.29.194)
10:10 AM <delta1> hey hey
10:23 AM <stringhandler> @dburkett I've invited you to the repo, the spam settings are super high at the moment
10:36 AM <stringhandler> afaik to replay the commitment on a pruned mode that had removed it, you would have to somehow find a different kernel excess
10:46 AM <dburkett> @stringhandler It'd be fairly trivial to force a chain split. Just create a transaction with the same output twice (doesn't even have to be a replay).
10:47 AM <dburkett> Do this 3 weeks after you've spent the output. Then nodes with pruning horizon more than 3 weeks ago won't accept it. Nodes with horizon less than that will.
10:48 AM <stringhandler> yeah
10:48 AM <dburkett> I can comment in the PR. Thanks for the access.
10:48 AM <stringhandler> perfect
10:48 AM <stringhandler> yeah, it's probably best to discuss it there
10:48 AM <stringhandler> Idk how permanent the IRC landscape is right now

```

DavidBurkett:
In the initial audit report I did for RFC-0201, I suggested a "windowed" approach for preventing replay attacks. You could just indicate a minimum and maximum block height in either the output or kernel (or just an expected block height and a pre-defined window), and then make sure there are no duplicates in that window. Limit the size of the window to something like 2 weeks, so you aren't required to index the whole history.

SWvheerden (changes requested):
`macro_rules! txn_schema {
    (from: $input:expr, to: $outputs:expr, fee: $fee:expr, lock: $lock:expr, mined_height: $mined_height:expr, features: $features:expr) => {{
        $crate::transactions::helpers::TransactionSchema {
            from: $input.clone(),`

This still has the mined_height in it and used to add in the output. This needs updating as well.

philipr-za:
> `macro_rules! txn_schema { (from: $input:expr, to: $outputs:expr, fee: $fee:expr, lock: $lock:expr, mined_height: $mined_height:expr, features: $features:expr) => {{ $crate::transactions::helpers::TransactionSchema { from: $input.clone(),`
> 
> This still has the mined_height in it and used to add in the output. This needs updating as well.

Updated

stringhandler:
I had incorrectly thought there was a unique index on commitment, but there is not. There is a unique constraint on both kernel excess and kernel excess signature, which both a pruned node and archival node will agree on. Our constraint is across the entirety of the kernel set, but since the users can change the offset, I don't think collisions will be a problem for honest participants. So our window is the entire kernel set. We don't  reuse kernels like Beam does, so I don't see the need for a window.

Unique excesses prevents the replay attack for interactive payments, but doesn't prevent for one-sided payment replay attacks where the sender knows the blinding factor, since the sender can re-sign the kernel with a new excess. I believe there are still some cases where replay attack may exist, but adding a height does not solve those, and I will rather explain it in an rfc or other document.

tl;dr: I think this can go in

stringhandler (approved):

DavidBurkett:
> So our window is the entire kernel set. We don't reuse kernels like Beam does, so I don't see the need for a window.

The window is for efficiency purposes, and likely necessary if you want to scale. Without a window, you have to keep the whole, perpetually-growing kernel set indexed to check uniqueness. The window significantly reduces the burden caused by this index, by limiting it to a tiny subset of the kernels (those within the short window timeframe).

> Unique excesses prevents the replay attack for interactive payments, but doesn't prevent for one-sided payment replay attacks where the sender knows the blinding factor, since the sender can re-sign the kernel with a new excess.

Help me understand the workflow for the one-sided replay attacks. How would they resign the kernel with a new excess while replaying the same outputs?

stringhandler:
Ok I see. Yeah, makes sense. We can add a window on it, I wonder what kind of size makes sense. For computer based pruned mode and archival nodes, the underlying database would usually be able to manage it fairly efficiently up to a very large size of kernels, but for possible mobile phone nodes that may be much smaller. Any ideas on size?

I think it's worth a larger write up for the replay attacks because there's not just one mitigation. 
A very short version though (My understanding):
1. Using normal MW transactions, Alice sends funds to Bob, who puts it in CommitmentA, who sends it to Charlie. At a later stage Alice (or someone else) sends Bob the same amount of funds, and for whatever reason Bob puts it in CommitmentA. This is possible because there is no unique constraint on Commitments. Charlie sees this and tries to put the same transaction he previously had on the blockchain, using the same excess and kernel signatures that were used previously. This should not be possible, because the excess and signature must be unique. Charlie can also not change the excess by subtracting more offset, or adding in outputs, because the kernel signature that Bob provided signs the public keys in the excess and Charlie would have to know the private key to regenerate Bob's signature. I think the fact that the public keys of the excess are included in the signature challenge shuts down most attacks where Charlie could add or change outputs.
2.   While Charlie does not know the blinding factor in (1), he cannot recreate the signature. This is not the case however with one-sided payment or other schemes where Charlie could get the blinding factor for Bob's commitment from Alice. A quick reminder: in one-sided payments, Alice creates the BF for the commitment from a DH of Bob's public receive address and the sender_script_offset_public_key. Alice does not know the correct script input to spend Bob's output and can also not generate a signature for the input. So in order to perform an attack, Alice and Charlie (working together, used interchangeably from now onwards) would have to generate a transaction where they can keep the script and signature, but change the other parts of the transaction.

> Note, at this point it's probably not really in the category as the previous replay attack. Bob's commitment is part of the challenge for the signature, so the commitment must be the same. IDK if the commitment was always part of the challenge though

  2.1. Attack 1: Alice tries to craft a transaction using the replayed commitment, input script and input signature with new outputs and kernels. She can generate the kernels and outputs fairly easily, but would not be able to generate the transaction level Gamma (is this the script_offset? Not sure of the name of this) without knowing Bob's private key.
2.2. Attack 2: Is it then possible to somehow perform this attack by using the same outputs? I think this is mitigated again by the kernel excess uniqueness.
Is there potentially some other permutation? Maybe?

DavidBurkett:
> Any ideas on size?

Not sure exactly. We want to target a range where we can be confident that a transaction is included in a block. This is probably somewhere on the order of a week, up to maybe 1 month or 2? Beyond that, transactions would be considered expired.

> I think it's worth a larger write up for the replay attacks because there's not just one mitigation.

I actually started doing just that in response to your earlier comments. You can follow along here: https://gist.github.com/DavidBurkett/af3c6c566cfb312c4c214e49c638baeb

> 2.2. Attack 2: Is it then possible to somehow perform this attack by using the same outputs? I think this is mitigated again by the kernel excess uniqueness.
> Is there potentially some other permutation? Maybe?

There might be. Let me think through this a little.

stringhandler:
I was thinking more like a year. I'd still like to believe that monthly salaries will be paid over crypto, so a month would be too short.

>I actually started doing just that in response to your earlier comments. You can follow along here:

Awesome. Will have a look

DavidBurkett:
> I'd still like to believe that monthly salaries will be paid over crypto, so a month would be too short.

You would expect it to be normal that a transaction for your monthly salary could take more than 1 month to confirm? I would probably not stay with that employer for long.

stringhandler:
I was thinking that if the kernel excess only had to be unique within a month, then you could perform this attack every month. Maybe I'm misunderstanding

DavidBurkett:
> I was thinking that if the kernel excess only had to be unique within a month, then you could perform this attack every month.

No, this completely eliminates the attack (for interactive txs). Our solution to replay attacks is to ensure kernels can't be reused by someone who doesn't have knowledge of the excess. While a unique index over the whole kernel set would work, the window limits the time range within which the kernel is valid, so we only have to check uniqueness within that window. Once the chain is past the kernel's max height (outside of the window), you can't replay that kernel at all. The only way to use the same kernel commitment beyond that is if you actually know the kernel excess, and therefore update the kernel height and resign.

stringhandler:
I think I'm on a different page. Let me read the window approach again

DavidBurkett:
You might be overthinking it. The windowed approach is just this:

1. Add a height range that is signed by both parties using the kernel excess
2. Nodes ensure the kernel is unique within that time frame.

Thanks to the uniqueness check, the only way to replay the transaction is to do it outside of the time frame specified in the original kernel (i.e. once it expired). But that requires changing the time frame the kernel is valid for (since it's now expired), and in order to do that, you need to know the kernel excess, which no single party knows.

stringhandler:
Agree that that would solve it for interactive

PR #3030, opened 2021-06-23
Network separation and protocol versioning 

sdbondi:
## Description

- Changes `WireMode::Comms` variant to contain a single byte.
- Allows the `WireMode::Comms` byte to be set from domain layer (tari_p2p).
- [partially network breaking] Removes redundant network specifier from messaging
  protocol.
- Removes DHT Network enum.
- Updates Network configuration enum to be representable as a byte.
- Network configuration enum is now the only Network specifier.
- Changes consensus Network enum to a struct (NetworkConsensus) that wraps
  the common Network enum.
- [network breaking] Use byte representation of multiaddr in identity
  message, rather than string. This reduces the bytes exchanged when
  establishing a new peer connection.
- [network breaking] Shave a few bytes off of identity and messaging protocol IDs to match the format of other protocols
- [network breaking] Add major and minor versions to messaging protocol
- [network breaking] Add major version check when establishing a
  connection
- [network breaking] Remove NodeId from PeerIdentityMsg as this can be derived from the
  public key obtained from the noise handshake

## Motivation and Context

This PR prevents networks from "cross-talk" at the peer connection protocol level.
In addition, previously there were a number (3) of Network enums causing maintenance issues, this PR uses a single enum to enumerate the types of networks. This point accounts for most of the LOC in this PR.

As part of the above, the DHT network enum was removed and so removed from the DHT-level message header, this is a network breaking change.
This PR also slightly reduces the number of bytes exchanged when establishing connections. This is a network breaking change that has been on my list for ages.

## How Has This Been Tested?

Added a few unit tests.
Existing tests pass. 
Basic test on private network. Nodes could only communicate with each other.

## Checklist:


delta1:
Looking good, just watching CI

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2628,6 +2626,7 @@ pub unsafe extern "C" fn comms_config_create(
             match ni {
                 Ok(ni) => {
                     let config = TariCommsConfig {
+                        network: Network::Weatherwax,
```
SWvheerden, line 2629: Should this not rather be an option, rather than a hardcoded value

SWvheerden (commented):
LGTTM, just one small nit thou

base_layer/wallet_ffi/src/lib.rs
sdbondi: I agree, however this is how it was previously, changing it is non-trivial so I think this is out of scope of this PR.

common/src/configuration/network.rs
```diff
@@ -0,0 +1,85 @@
...
+    fmt,
+    fmt::{Display, Formatter},
+    str::FromStr,
+};
+
+/// Represents the available Tari p2p networks. Only nodes with matching byte values will be able to connect, so these
+/// should never be changed once released.
+#[repr(u8)]
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+pub enum Network {
```
philipr-za, line 34: These bytes don't have the spacing between mainnet and testnets we discussed in the RFC.

comms/dht/src/envelope.rs
```diff
@@ -128,14 +128,14 @@ impl DhtMessageType {
 /// It is preferable to not to expose the generated prost structs publicly.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct DhtMessageHeader {
-    pub version: u32,
+    pub major: u32,
```
philipr-za, line 131: Nit discussion: at these fields maybe a bit big for the purpose? u16 should be plenty? There are just going to be a lot of headers on the network so might be worthwhile to save a few bytes here.

sdbondi: **TLDR** went with the dead simple approach because the saving is probably not worth any additional code.

Let's compare two approaches:

Note:
-  All ints, regardless of their proto file definition, are encoded using varints. e.g a protobuf field as `uint64` with a value < 128 takes 1 byte when encoded as varint.
- There is no uint16 protobuf integer type (I would have chosen this, but I suspect it makes no practical difference)
- Chose u32 in the non-generated header struct to avoid a fallible `u32 -> u16` conversion
- 0 values are free (omitted field is the same as a 0 value)
- Byte cost is over-the-wire

Given we want to represent major = 1, minor =2 - v1.2 for short

**Approach 1**

`x: u16` with `major = msb(x) -> u8` and `minor = lsb(x) -> u8`

Byte cost:  [field number 1 byte] + [2 bytes for the u16 number with value > 127] 
                  Or as binary [0000 0001] [1000 0010]
Total: 3 bytes

Code cost: Simple conversion code required to extract 2 u8s from a u16 and vice versa.

Advantage: Potentially 1 byte saving

** Approach 2 **

2 fields, major and minor

Byte cost: [`major` field number 1 byte] + [value (< 127) 1 byte] + [`minor` field number 1 byte] + [value 1 byte]
Total: 4 bytes

Code cost: None

Advantage: minor version can be bigger than 255, possible that we increment it with each release.

common/src/configuration/network.rs
sdbondi: Thanks, I'll update that

comms/dht/src/envelope.rs
philipr-za: Ja, if protobuf can't just do u16 out the box then def not worth it. I agree that u8 on minor version is not enough.

common/src/configuration/network.rs
sdbondi: Updated, though not sure exactly if this is what @mikethetike had in mind

stringhandler: it's fine.

comms/dht/src/envelope.rs
stringhandler: Since it's VarInt in protobuf, it's kinda moot. Not sure how it's encoded in the SAF db, but that can be update to varint as well and is not strict for consensus

stringhandler (approved):

PR #3031, opened 2021-06-23
Update connectivity manager defaults

philipr-za:
## Description
This PR tweaks the `offline_peer_cooldown` and `max_failures_mark_offline` default values to better cope with intermittent disconnections.

`offline_peer_cooldown` is reduced from 24 hours to 2 hours so that the peer will become eligible for a reconnection in less time.

`max_failures_mark_offline` is changed from 1 to 2 to allow for one more round of connection retries before giving up on the node until the above cool down expires.

## How Has This Been Tested?
Existing tests

## Checklist:


sdbondi (approved):
LGTM

PR #3032, opened 2021-06-23
Add filtering of abandoned coinbase txs to console wallet

philipr-za:
## Description
This PR adds the ability to filter abandoned coinbase transactions in the completed transaction list of the console wallet. The filter is on by default and can be toggled with the A key. Transactions that have the Coinbase status and are also cancelled are filtered out and when displayed have the status of `Abandoned`.

A TransactionFilter bitfield has been added to the AppState which currently only includes the AbandonedCoinbase flag but will allow for more filters to be added and combined. The filtering is done when the UI draw requests the vector of references to the current list of transactions. If no filter is applied the full vector of references is returned, otherwise a iterator filtered version is returned applying the conditions of the filter.

## How Has This Been Tested?
Manually tested, this is a console wallet UI change

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/app_state.rs
+++ b/applications/tari_console_wallet/src/ui/state/app_state.rs
@@ -34,6 +34,7 @@ use crate::{
     },
     wallet_modes::PeerConfig,
 };
+use bitflags::bitflags;
 use futures::{stream::Fuse, StreamExt};
 use log::*;
 use qrcode::{render::unicode, QrCode};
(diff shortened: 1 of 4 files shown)


applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -275,25 +275,29 @@ impl AppState {
...
 
     pub fn get_completed_tx(&self, index: usize) -> Option<&CompletedTransaction> {
-        if index < self.cached_data.completed_txs.len() {
-            Some(&self.cached_data.completed_txs[index])
+        let filtered_completed_txs = self.get_completed_txs();
+        if index < filtered_completed_txs.len() {
+            Some(filtered_completed_txs[index])
         } else {
             None
         }
```
sdbondi, line 303:
```suggestion
        self.get_completed_txs().get(index)
```

sdbondi (approved):
Lightly tested and LGTM

PR #3034, opened 2021-06-24
[common] Start removing `min_type_alias_impl_trait` feature and migrate to `async_trait` 

delta1:
## Description

Starts the migration process for Rust stable by removing some of the feature attributes for `min_type_alias_impl_trait` and replacing them with `async_trait` where possible. 

The `comms` and `comms/dht` crates are much more difficult to do, leaving them for now. 

## Motivation and Context

Progress on the road to Rust stable #3035 

## How Has This Been Tested?

 
Existing tests

## Checklist:


base_layer/service_framework/src/tower/service_ext.rs
```diff
@@ -117,10 +117,9 @@ mod test {
 
         impl Service<u32> for ReadyLater {
             type Error = ();
+            type Future = Pin<Box<dyn Future<Output = Result<u32, ()>>>>;
             type Response = u32;
 
-            type Future = impl Future<Output = Result<Self::Response, Self::Error>>;
```
sdbondi, line 122: For this one we would rather want to static type

```rust
type Future = future::Ready<Result<Self::Response, Self::Error>>;
```

and return `future::ok(...)`

sdbondi (approved):
Nice LGTM - just one nit (that is very slightly optimal)

base_layer/service_framework/src/tower/service_ext.rs
delta1: Nice thanks I'll try that!

sdbondi: Just noticed that this is in a test - so really doesnt matter at all - but thanks for entertaining my nit 😄

PR #3041, opened 2021-06-29
Add input_mr and witness_mr to header

sdbondi:
## Description

As per updated RFC 120 (ref #3025):

- Add `input_mr` to header, commits to all inputs in a block
- Add input merkle root validation check
- Commit to all transaction input fields
- Add `TransactionInput::output_hash` that produces the hash of the output 
   that was spent in the input 
- Remove range proof MR in favour of witness MR
- Witness MR commits to all rangeproofs and sender metadata
- Add witness_hash function that produces `Hash(RangeProof || sender_metadata_signature)`
- Updated weatherwax genesis timestamp because this change is a hard
  fork
- Updated header format in proto messages (network breaking)
- Commit to `input_mr` in sha3 and monero mining hashes
- Add some missing fields (`output_mmr_size`, `kernel_mmr_size` and `total_script_offset`) 
  to sha3 miner hash (w/ minor code clean up)

## Motivation and Context

As per updated RFC 120 (ref #3025):

## How Has This Been Tested?

Existing tests pass.

## Checklist:


base_layer/core/src/proof_of_work/sha3_pow.rs
```diff
@@ -43,7 +43,7 @@ pub fn sha3_hash(header: &BlockHeader) -> Vec<u8> {
         .chain(header.prev_hash.as_bytes())
         .chain(header.timestamp.as_u64().to_le_bytes())
         .chain(header.output_mr.as_bytes())
-        .chain(header.range_proof_mr.as_bytes())
+        .chain(header.witness_mr.as_bytes())
         .chain(header.kernel_mr.as_bytes())
```
SWvheerden, line 49: This needs the input_mr as well. If this is not part of the pow, it will be mutable.

integration_tests/helpers/baseNodeClient.js
```diff
@@ -362,7 +362,7 @@ class BaseNodeClient {
     const timestamp = parseInt(header.timestamp.seconds);
     hash.update(toLittleEndian(timestamp, 64));
     hash.update(header.output_mr);
-    hash.update(header.range_proof_mr);
+    hash.update(header.witness_mr);
```
SWvheerden, line 366: Needs the input_mr as well here

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -925,25 +927,28 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(db: &T, block: &Block) -> Resul
...
     }
 
     for output in body.outputs().iter() {
         output_mmr.push(output.hash())?;
-        proof_mmr.push(output.proof().hash())?;
+        witness_mmr.push(output.witness_hash())?;
     }
 
     for input in body.inputs().iter() {
+        let input_hash = input.hash();
```
SWvheerden: This does not technically include the entire input. We are only hashing the following fields here:  
```
features
commitment
script
```
We need all the fields. 

My proposal would be to rename the function `input.hash()` to `input.txo_hash()` or something along those lines. This hash would then match the hash of output.hash(). 
We then make `input.hash()` the hash of the entire `TransactionInput`

base_layer/core/src/transactions/transaction.rs
```diff
@@ -1172,6 +1178,17 @@ mod test {
         script::ExecutionStack,
     };
 
+    #[test]
+    fn input_and_output_hash_match() {
```
SWvheerden, line 1212: See comment above

SWvheerden (commented):
Looks good, two things. 
We need to lock down the input_data etc which is not currently in input.hash(). 
And we need the input_mr field in the pow as well.

base_layer/core/src/proof_of_work/sha3_pow.rs
sdbondi: Thanks done

integration_tests/helpers/baseNodeClient.js
sdbondi: Thanks done

hansieodendaal (commented):
LGTM from what I can see, but see above comments from @SWvheerden. 

Tested with base node, wallet, mining node and merge mining proxy. All running fine, building the blockchain.

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Ok that makes sense - perhaps `input.output_hash()` and we assert that it matches the input hash

base_layer/core/src/transactions/transaction.rs
sdbondi: Update test to use `output_hash()`

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Changes made

hansieodendaal (approved):
LGTM

stringhandler (approved):

PR #3043, opened 2021-06-30
[Feat]Allow zero-confirmation transactions

SWvheerden:
## Description

- Changed the mempool to search for "missing' inputs inside of itself to allow adding zero-confirmation transactions. 
- Changed the mempool selection to search the mempool for a chain of transactions when adding a zero-confirmation transaction that requires certain outputs to be mined.
- Added a utxo index to the mempool so that it known which transaction has which output.
- Changed the block validator to accept zero-confirmation transactions.
- Changed the MMR route calculation to search for inputs inside of it self as well as the database.
- Fixes cucumber transaction builder to use littleEndian as starting point for the math, and allowing wrap around.

## Motivation and Context

This PR allows the mempool and base node to accept zero-confirmation transactions and process them. 
## How Has This Been Tested?
Created a unit tests testing zero-confirmation transactions and cucumber integration tests testing zero-confirmation tests

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -940,13 +940,19 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(db: &T, block: &Block) -> Resul
     }
 
     for input in body.inputs().iter() {
-        let index = db
-            .fetch_mmr_leaf_index(MmrTree::Utxo, &input.output_hash())?
-            .ok_or_else(|| ChainStorageError::ValueNotFound {
-                entity: "UTXO".to_string(),
-                field: "hash".to_string(),
-                value: input.output_hash().to_hex(),
-            })?;
+        // Search the DB for the output leaf index so that it can be marked as spent/deleted.
+        // If the output hash is not found, check the current output_mmr. This allows zero-conf transactions
+        let index =
+            match db.fetch_mmr_leaf_index(MmrTree::Utxo, &input.output_hash())? {
+                Some(index) => index,
+                None => output_mmr.find_leaf_index(&input.output_hash())?.ok_or_else(|| {
+                    ChainStorageError::ValueNotFound {
+                        entity: "UTXO".to_string(),
+                        field: "hash".to_string(),
+                        value: input.output_hash().to_hex(),
+                    }
+                })?,
+            };
         input_mmr.push(input.hash())?;
 
         if !output_mmr.delete(index) {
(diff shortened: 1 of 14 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -940,13 +940,18 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(db: &T, block: &Block) -> Resul
     }
 
     for input in body.inputs().iter() {
-        let index = db
-            .fetch_mmr_leaf_index(MmrTree::Utxo, &input.output_hash())?
-            .ok_or_else(|| ChainStorageError::ValueNotFound {
-                entity: "UTXO".to_string(),
-                field: "hash".to_string(),
+        // db search should be much faster and should be the most widely used case,
+        // there for we search first in the db, then in the new mmr
```
sdbondi: nit: Rather explain the logic than guessing about perf

```suggestion
        // Search the DB for the output leaf index so that it can be marked as spent/deleted.
        // If the output hash is not found, check the current output_mmr. This allows zero-conf transactions
```

integration_tests/helpers/transactionBuilder.js
```diff
@@ -266,7 +290,7 @@ class TransactionBuilder {
       privateNonce,
       challenge
     );
-
+    console.debug(excess.commitment);
```
sdbondi: Remove?

base_layer/core/tests/mempool.rs
```diff
@@ -333,6 +333,306 @@ fn test_retrieve() {
...
+        to: vec![21 * T, 11 * T, 11 * T, 16 * T]
+    )];
+    // "Mine" Block 1
+    generate_new_block(&mut store, &mut blocks, &mut outputs, txs, &consensus_manager).unwrap();
+    mempool.process_published_block(blocks[1].to_arc_block()).unwrap();
+
+    // This transaction graph will be created, containing 3 levels of zero-conf transactions, inheriting dependent
+    // outputs from multiple parents
+    //
+    // tx01   tx02   tx03   tx04    Basis transactions using mined inputs (lowest fees, increases left to right)
```
sdbondi, line 352: It would be awesome if we could set up transaction tests easily by using a syntax e.g.

```
TX1: A -> B
TX2: B -> C
TX3: C -> D
```

```
TX1: A -> B
TX2: B -> C
TX3: A -> C
```

```
TX1: A -> [B,C,F]
TX2: B -> D
TX3: [C, F] -> E
```

 Looking at the current transaction builders, this seems possible but maybe a future job

sdbondi (commented):
Looks good - Took a close look at the code and once I grokked it, it made sense. Tests are nice and neat and happy that they cover the most important cases.

Tested mining and spending to self (not zero confs of course). Had to fix a minor compile error though. Once that is fixed, conflicts are fixed and clippy is happy I think this can go in.

base_layer/core/tests/mempool.rs
SWvheerden: This would be a nice future to do. Currently, we only have one test, testing this. If and when we extend this to more tests we could look into the time to build something fancier.

PR #3045, opened 2021-07-01
WIP Exit when synced

Cifko:
## Description
This is a draft PR.
The flag is only working half way. It will stop the node from any processing. But it doesn't quit the app.


StriderDM:
You have the right idea here. What you need to do is trigger the `interrupt_signal` to get to the `UserQuit` state and quit the app instead of setting it directly.

Cifko:
> You have the right idea here. What you need to do is trigger the `interrupt_signal` to get to the `UserQuit` state and quit the app instead of setting it directly.

Yes, but the mutable Shutdown is passed to the user input thread. I can either disable the user input (just the quit part) or we can somehow change how the Shutdown can be triggered.

delta1:
Just an idea, maybe you could just use this with `--daemon` mode and then there's no user input thread to worry about?

Cifko:
It appears that this PR will be reverted. The script for the test will be done without this flag. I'm glad, because passing the shutdown all the way down to the sync would pollute the code a lot. It would be nice to have a singleton for that shutdown, but I down know if that's even possible in rust. Anyway, doing the test without this flag is IMHO much better.

PR #3055, opened 2021-07-05
[WIP] Update the Ubuntu archive build for distributions with GHA

leet4tari:
Have an process on Tag release that is automated that will build and upload to S3 as an archive that is used by Ubuntu users.

## Description
Add noBuild option for ```build_dists_tarball.sh``` to be used by GHA for build an Ubuntu archive distribution. Also fix the GHA caching resource for reduced build time by adding a key on OS for builds. 

## Motivation and Context
Trying to use the code base inside the GHA build process, so that builds on GH and any other service will be the same as a local build without the need for more tools.

## How Has This Been Tested?
Built with GHA for Ubuntu, OSX and Windows, but with caveats:
 - OSX does not build in GHA due to issue with typenum v1.12.0
 - Windows 2016 has a cmake build issue (might not be worth fixing)
 - We can't have sub-folders for checksums - runtime folder usage is broken
 - Can't use scripts in their current state
 - Documentation needs updating to reflect changes

## Checklist:

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -28,7 +28,7 @@ jobs:
 #        os: [windows-2016]
 #        os: [ubuntu-latest, macos-latest, windows-latest, self-hosted]
 #        os: [ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0, windows-2016, windows-2019]
-        os: [ubuntu-18.04, macos-10.15, self-hosted]
+        os: [ubuntu-18.04, macos-10.15, windows-2019, self-hosted]
         # https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
 #        features: ["safe", "avx2"]
 #        features: ["safe"]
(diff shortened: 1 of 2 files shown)


delta1:
> OSX does not build in GHA due to issue with typenum v1.12.0

Use target_cpu = `x86-64` for macos

> Windows 2016 has a cmake build issue (might not be worth fixing)

Switch to windows-2019

.github/workflows/base_node_binaries.yml
```diff
@@ -24,20 +24,19 @@ jobs:
...
-#        features: ["safe"]
 #        target_cpu: ["x86-64", "broadwell", "skylake"]
         target_cpu: ["haswell"]
 #        target_release: ["release", "debug"]
-        exclude:
-          - os: "macos-10.15"
-            target_cpu: "haswell"
+#        exclude:
+#          - target_cpu: "x86-64"
+#            features: "avx2"
```
SWvheerden, line 39: There is a lot of lines that are commented out. These will just cause confusion later on. Can we remove them?

leet4tari: I would prefer not, as there are reminders or options I believe we would use later. Anyone is welcome to remove them as they want too, thou I am not sure how comments can cause confusion, unless the comments are wrong or are not comments?

PR #3058, opened 2021-07-06
Chore: Fix cucumber

SWvheerden:
## Description

We removed Blake from the proto files and updated the proto files in #3047 , but we never updated the cucumber tests calling the GRPC requests. This updates them to fix them again. 

## Checklist:


sdbondi:
We should probably put those in a "enum".

```javascript
const PowAlgo = { monero: 0, sha3: 1 };
```

PR #3061, opened 2021-07-06
Automatic scripts for testing syncing and wallet recovery

philipr-za:
## Description
This PR adds two NodeJS scripts that will use the tooling from the Cucumber tests to compile and run a Base Node  and a Wallet to test syncing and recovery times.

The `automatic_sync_test.js` script will compile the base node and run it while it syncs. It will monitor stdout until the node has synced at which point it will calculate the stats of the sync and write them to stdout and shutdown the base node. You can pass `pruned` as a command line argument to have the node run with a pruning horizon of 1000.

The `automatic_recovery_test.js` will compile and run a wallet and have it perform recovery from a seed phrase. It will monitor stdout until this process is complete and then print the statistics to stdout before shutting down. You can pass a command switch of `—sw` or `—seed-words` followed by a quote delimited string of seed words to use that seed phrase rather than the default.

## How Has This Been Tested?
The changes to the cucumber test tooling will be tested by existing tests.

## Checklist:

### DIFF
--- a/applications/utils/automatic_recovery_test.js
+++ b/applications/utils/automatic_recovery_test.js
@@ -1,0 +1,85 @@
+const WalletProcess = require("../../integration_tests/helpers/walletProcess");
+const WalletClient = require("../../integration_tests/helpers/walletClient");
+
+const fs = require("fs");
+const yargs = require("yargs");
+
+async function main() {
+  const argv = yargs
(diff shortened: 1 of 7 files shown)


sdbondi (dismissed):
Nice LGTM - tested running the script and syncing.

PR #3063, opened 2021-07-06
Fix db update error

hansieodendaal:
## Description
The transaction receive protocol should finish gracefully if the sender's metadata signature cannot be updated in the wallet database.

## Motivation and Context
- Unit test `test_restarting_transaction_protocols` failed
- The transaction receive protocol exited with an error if the sender's metadata signature could not be updated in the wallet database instead of running to completion.

## How Has This Been Tested?
Unit test

## Checklist:


stringhandler (approved):
This file is so complex, it's hard to know if this is ok or not. The error is logged at least

PR #3072, opened 2021-07-08
Update `tari-common` crate feature flags to exclude git2 from lib_wallet build

philipr-za:
## Description
This PR adds a feature flag to the Tari-common crate so that the git2 dependency can be excluded from he building of Core. This dependency contains a C dependency (libgit2) that breaks the LibWallet build.

## How Has This Been Tested?
Manually tested

## Checklist:


StriderDM (approved):
Nice @philipr-za. 

Approving so long, will start checking the docker builds for github actions with this and the updated nightly version.

PR #3081, opened 2021-07-12
Update parsing of `num_mining_threads` config field

philipr-za:
## Description
The `num_mining_threads` config option in the sample config file was ignore. This was due to the parsing logic looking for it as `base_node.NETWORK.num_mining_threads` where it was under the `mining_node` section without a network identifier. Thus only the `weatherwax` specific default was ever used.

This PR updates the config bootstrap to parse this field from under the `mining_node` section now.

## How Has This Been Tested?
Manually tested

## Checklist:


sdbondi (approved):
Tested and works :+1:

PR #3083, opened 2021-07-12
Update RFC-0140

philipr-za:
## Description
This PR updates RFC-0140 to better reflect the current implementation.

## How Has This Been Tested?
Only documentation updates.

## Checklist:

### DIFF
--- a/RFC/src/Glossary.md
+++ b/RFC/src/Glossary.md
@@ -86,6 +86,13 @@ kernels and headers up to that block height from the genesis block.
 A strategy for propagating messages amongst nodes in a peer-to-peer network. Example implementations of
 `BroadcastStrategy` include the Gossip protocol and flood fill.
 
+## Chain Reorganization
+[reorg]: #chain-reorg "After a chain split occurs nodes on the poorer chain must resync to the best chain"
+
+A chain reorganization occurs after a chain split occurs on the network, which commonly occurs due to network latency
+and connectivity issues. When a chain split occurs one chain will have the higher accumulated proof-of-work, this chain
+is considered the best chain. Nodes on the poorer chain will need to rewind and resync their chains to best chain. In 
+this process transaction in the mempool could become orphaned or invalid.
 
 ## Checkpoint
 [checkpoint]: #checkpoint "A summary of the state of a Digital Asset that is recorded on the base layer"
@@ -289,7 +296,7 @@ can be processed or attempts to spend [UTXO]s with time-locks.
 
 
 ## Pruning horizon
-[pruninghorizon]: #pruning-horizon	"Block height at which pruning will commence"
+[pruning horizon]: #pruning-horizon	"Block height at which pruning will commence"
 
 This is a local setting for each node to help reduce syncing time and bandwidth. This is the number of blocks from the 
 chain tip beyond which a chain will be pruned.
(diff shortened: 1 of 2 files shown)


RFC/src/Glossary.md
```diff
@@ -86,6 +86,13 @@ kernels and headers up to that block height from the genesis block.
 A strategy for propagating messages amongst nodes in a peer-to-peer network. Example implementations of
 `BroadcastStrategy` include the Gossip protocol and flood fill.
 
+## Chain Reorganization
+[reorg]: #chain-reorg "After a chain split occurs nodes on the poorer chain must resync to the best chain"
+
+A chain reorganization occurs after a chain split occurs on the network, which commonly occurs due to network latency
+and connectivity issues. When a chain split occurs one chain will have the higher accumulated proof-of-work, this chain
+is considered the best chain. Nodes on the poorer chain will need to resync their chains to best chain. In this process
```
SWvheerden:
```suggestion
is considered the best chain. Nodes on the poorer chain will need to rewind and resync their chains to best chain. In this process
```

RFC/src/RFC-0140_Syncing_and_seeding.md
```diff
@@ -60,83 +60,93 @@ The aim of this Request for Comment (RFC) is to describe the syncing, seeding an
...
-divided into two [SynchronizationStrategy]s: complete sync and sync. Complete sync means that the node communicates with 
-an archive node to get the complete history of every single block from genesis block. Sync involves the node getting 
-every block from its [pruning horizon] to [current head], as well as every block header 
-from genesis block. 
+divided into two [SynchronizationStrategy]s: complete sync and horizon sync. Complete sync means that the node 
+communicates with an archive node to get the complete history of every single block from genesis block. Horizon Sync 
+involves the node getting every block from its [pruning horizon] to [current head], as well as every block header 
+up to the genesis block. 
+
+To determine if the node needs to synchronise the node will monitor the broadcast chain_metadata provided by its neighbours.
```
SWvheerden, line 68:
```suggestion
To determine if the node needs to synchronise, the node will monitor the broadcasted chain_metadata provided by it's neighbours.
```

RFC/src/RFC-0140_Syncing_and_seeding.md
```diff
@@ -60,83 +60,93 @@ The aim of this Request for Comment (RFC) is to describe the syncing, seeding an
...
+1. Set [SynchronizationState] to `header_sync`.
+2. Sync all missing headers from the genesis block to the current chain tip. The initial header sync allows the node to
+   confirm that the syncing peer does indeed have a fully intact chain from which to sync that adheres to this nodes
+   consensus rules and has a valid proof-of-work that is higher than any competing chains.
+3. Set [SynchronizationState] to `horizon_sync`.
+4. Download all kernels from the current network tip back to this node's [pruning horizon].
+5. Validate kernel MMR root against headers.
+6. Download all [utxo]'s from the current network tip back to this node's [pruning horizon].
+7. Validate outputs and [utxo] MMR.
+8. Once all kernels and [utxo]s have been downloaded from the network tip back to this node's [pruning horizon] set
```
SWvheerden: Step 8  should be validate balance by comparing the utxo set to the total accumulated balance.
Step 9 should be do block sync to get to the listing state. 

horizon sync, only syncs to the horizon before handing over to normal block sync, it is also the only piece of code currently to check the total chain emmision

SWvheerden (dismissed):
Looks good, one or 2 small nits

RFC/src/RFC-0140_Syncing_and_seeding.md
philipr-za: nope

RFC/src/Glossary.md
philipr-za: updated

RFC/src/RFC-0140_Syncing_and_seeding.md
philipr-za: Ok updated, but just to confirm you are saying that only the horizon sync checks the total chain emission. Why isn't that done with normal block sync?

SWvheerden: Its done indirectly. 
So normal block propagation only checks that that single block has the correct emission for its height. 
It checks that:
`outputs - inputs = emission of that height. `
So it only checks that the block has increased the total emission by the emission value of that height. If you do this for every block, you can assume that the total is also correct. 

Horizon sync does not check every block, so it needs to check that:
`sum(all_ouputs)  = total accumulated emission at that height.`
It has no blocks to check that the block emission was correct.

stringhandler (approved):

PR #3084, opened 2021-07-12
Add one-sided transactions to `make-it-rain`, improve its async sending

hansieodendaal:
## Description
This PR focused on adding one-sided transactions to `make-it-rain` and to improve its async sending ability:
- Added one-sided transactions to the `make-it-rain` command.
- Made sending transactions for `make-it-rain` asynchronous,
- Updated the wallet README.md,
- Added test cases for `make-it-rain` command parser.
- Added time measurements for processing transaction service events to assist in finding processing bottlenecks, _**this can be removed later on**_.
- Improved logic around the database call to `get_balance` when selecting UTXOs for sending.

## Motivation and Context
- Before this PR executing the `make-it-rain` command was really slow.  Transactions were also not submitted at the required rate but held back by the slow transaction sending process.
- It was unclear where time was spent sending transactions, and if transactions were sent in parallel.
- After parallelizing the `make-it-rain` command, it was apparent that transaction sending became progressively slower as transactions were sent, which was also dependent on the UTXO set size in the wallet database to select from.
- Transaction sending becoming progressively slower was found to be caused by a database call to `get_balance` that was unnecessary 99% of the time. The balance query includes UTXOs encumbered to be sent and received, and with the growing set of those UTXOs as the test progressed, the query grew in complexity. The database call to `get_balance`'s primary purpose in the transaction sending context is to add UTXOs encumbered to be received to the available balance if the latter was insufficient when selecting UTXOs for sending. The solution was to not call `get_balance` if the available balance was sufficient.

With testing `make-it-rain` _on this branch_:
- After making `make-it-rain` asynchronous, _before fixing_ the call to `get_balance` :
  - Using a "big" sized wallet database (~8,000 available UTXOs):
    - 2,500 (`one-sided`, `negotiated`) transactions between one sender wallet and one receiver wallet in **one** `make-it-rain` task (not including mining):
      - Submitted all transactions at an actual rate of 5 transactions/s in parallel tasks.
      - Finalizing send commands at ~7.6 transactions/minute, irrespective of using `one-sided` or `negotiated` transactions.
  - Using a "medium" sized wallet database (~3,000  available UTXOs):
    - 1,000 (negotiated`) transactions between one sender wallet and one receiver wallet in **five** `make-it-rain` tasks (not including mining):
      - Submitted all transactions at an actual rate of 5 x 5 transactions/s in parallel tasks.
      - All 1,000 transactions' initial sending were concluded in 744s, thus ~80.6 transactions/minute
  - Using a "big" sized wallet database (~7,000 available UTXOs):
    - 2,500 (negotiated`) transactions between one sender wallet and one receiver wallet in **five** `make-it-rain` tasks (not including mining):
      - Submitted all transactions at an actual rate of 5 x 5 transactions/s in parallel tasks.
      - Finalizing send commands at ~9.3 transactions/minute, irrespective of using `one-sided` or `negotiated` transactions.
- After making `make-it-rain` asynchronous, _after fixing_ the call to `get_balance` :
  - Using a "big" sized wallet database (~9,500 available UTXOs):
    - 2,500 (negotiated`) transactions between one sender wallet and one receiver wallet in **five** `make-it-rain` tasks (not including mining):
      - Submitted all transactions at an actual rate of 5 x 5 transactions/s in parallel tasks.
      - Finalizing send commands at ~67.4 transactions/minute.

image

Intuitively, database requests for spendable UTXOs should become quicker as the list becomes smaller; this is now evident in the graph above.

## How Has This Been Tested?
- Unit test added
- System level testing

## Checklist:

### DIFF
--- a/applications/tari_console_wallet/README.md
+++ b/applications/tari_console_wallet/README.md
@@ -47,18 +47,37 @@ Monitoring 1 sent transactions to Broadcast stage...
 Done! All transactions monitored to Broadcast stage.
 ```
 
-- **make-it-rain**
+- **send-one-sided**
 
-Make it rain! Send many transactions to a public key or emoji id.
+Send an amount of Tari to a public key or emoji id in a one-sided transaction.
 
-`tari_console_wallet --command "make-it-rain <tx/sec> <duration> <amount> <increment> <start time or now> <pubkey> <optional message>"`
+`tari_console_wallet --command send-one-sided <amount> <pubkey> <optional message>"`
 
 example:
 
 ```
-$ tari_console_wallet --command "make-it-rain 1 10 8000 100 now c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 makin it rain yo"
+$ tari_console_wallet --command "send-one-sided 1T c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 coffee"
 
-1. make-it-rain 1 10 8000 µT 100 µT 2021-03-26 10:03:30.459157 UTC c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 makin it rain yo
+1. send-tari 1.000000 T c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 coffee
+
+Monitoring 1 sent transactions to Broadcast stage...
+Done! All transactions monitored to Broadcast stage.
+```
+
+- **make-it-rain**
+
+Make it rain! Send many transactions to a public key or emoji id.
+
+`tari_console_wallet --command "make-it-rain <tx/sec> <duration> <amount> <increment> <start time or now> <pubkey> <transaction type> <optional message>"`
+
+`<type>` can be `negotiated` or `one_sided`
+
+example:
+
+```
+$ tari_console_wallet --command "make-it-rain 1 10 8000 100 now c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 negotiated makin it rain yo"
+
+1. make-it-rain 1 10 8000 µT 100 µT 2021-03-26 10:03:30.459157 UTC c69fbe5f05a304eaec65d5f234a6aa258a90b8bb5b9ceffea779653667ef2108 negotiated makin it rain yo
 
 Monitoring 10 sent transactions to Broadcast stage...
 Done! All transactions monitored to Broadcast stage.
--- a/applications/tari_console_wallet/src/automation/command_parser.rs
+++ b/applications/tari_console_wallet/src/automation/command_parser.rs
@@ -81,6 +81,7 @@ pub enum ParsedArgument {
     OutputToCSVFile(String),
     CSVFileName(String),
     Address(Multiaddr),
+    Negotiated(bool),
 }
 
 impl Display for ParsedArgument {
@@ -96,6 +97,7 @@ impl Display for ParsedArgument {
             OutputToCSVFile(v) => write!(f, "{}", v.to_string()),
             CSVFileName(v) => write!(f, "{}", v.to_string()),
             Address(v) => write!(f, "{}", v.to_string()),
+            Negotiated(v) => write!(f, "{}", v.to_string()),
         }
     }
 }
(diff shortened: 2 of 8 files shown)


base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -281,8 +281,10 @@ where
             futures::select! {
                 //Incoming request
                 request_context = request_stream.select_next_some() => {
-                    trace!(target: LOG_TARGET, "Handling Service API Request");
+                    let start = Instant::now();
```
philipr-za, line 285: I am not a fan of adding all this code to log timings that could be found via log analysis, if you really think its worth leaving in then at least reduce it to a single timing from the top of the loop to the bottom rather than so many individual timing blocks in an already cumbersome select! statement.

philipr-za (dismissed):
LGTM, we must just remember to remove the timing code from the transaction service select loop at some point

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Yeah, we specifically needed to look at the timings within the `futures::select! {` here.

applications/tari_console_wallet/src/automation/command_parser.rs
```diff
@@ -226,6 +230,22 @@ fn parse_make_it_rain(mut args: SplitWhitespace) -> Result<Vec<ParsedArgument>,
     let pubkey = parse_emoji_id_or_public_key(pubkey).ok_or(ParseError::PublicKey)?;
     parsed_args.push(ParsedArgument::PublicKey(pubkey));
 
+    // transaction type
+    let txn_type = args
+        .next()
+        .ok_or_else(|| ParseError::Empty("transaction type".to_string()))?;
+    let negotiated = if txn_type == "negotiated" {
```
SWvheerden: Why not just use a match here rather than if else if  else

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -284,60 +283,149 @@ pub async fn make_it_rain(
...
+            num_txs,
+            transaction_type,
+            message,
+            Utc::now()
+        );
+    });
 
-    Ok(tx_ids)
+    // Return an empty vector here because we will not wait for actual transaction sending to finish
+    Ok(Vec::new())
```
SWvheerden: Rather then always returning an empty vec, why not just remove it?

SWvheerden (dismissed):
Looks good to me, I left 2 minor nits. 
I am also not a big fan of the timing code. It clutters the code and makes it hard to follow and read whats going on in the select.  From the look of it, it looks like debug code that got leftover.  If we want to trace what happened, just looking at the logging timestamp should be sufficient for most debugging purposes looking at log files after an event.

applications/tari_console_wallet/README.md
```diff
@@ -47,18 +47,37 @@ Monitoring 1 sent transactions to Broadcast stage...
...
+```
+
 - **make-it-rain**
 
 Make it rain! Send many transactions to a public key or emoji id.
 
-`tari_console_wallet --command "make-it-rain <tx/sec> <duration> <amount> <increment> <start time or now> <pubkey> <optional message>"`
+`tari_console_wallet --command "make-it-rain <tx/sec> <duration> <amount> <increment> <start time or now> <pubkey> <transaction type> <optional message>"`
+
+`<type>` can be `negotiated` or `one_sided`
```
stringhandler, line 73: I prefer the term "interactive" to "negotiated" and since it's generally referred as such in discussions/issues

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -284,60 +283,148 @@ pub async fn make_it_rain(
...
-        let tx_service = wallet_transaction_service.clone();
-        let tx_id = handle
-            .spawn(send_tari(tx_service, send_args))
-            .await
-            .map_err(CommandError::Join)??;
-
-        debug!(target: LOG_TARGET, "make-it-rain tx_id: {}", tx_id);
-        tx_ids.push(tx_id);
-    }
+        let transaction_type = if negotiated { "negotiated" } else { "one-sided" };
```
stringhandler, line 325: nit: This is a really long method, it would be worth your while splitting it up and simplifying it

stringhandler (approved):

PR #3087, opened 2021-07-13
Fallback to default flags if rxcache initialization fails

sdbondi:
## Description

Fallback to default flags if randomx cache allocation fails.
This should allow platforms with limited resources to validate blocks.
Similar to how xmrig falls back:
https://github.com/xmrig/xmrig/blob/02b2b87bb685ab83b132267aa3c2de0766f16b8b/src/crypto/rx/RxCache.cpp#L88
Adds some basic unit tests

## Motivation and Context

A user reported "Unable to allocate randomx cache". Since this is used for block validation (not mining), falling
back to the slowest but lightest randomx profile is acceptable.

## How Has This Been Tested?

Basic unit tests. Tested on weatherwax base node

## Checklist:

### DIFF
--- a/base_layer/core/src/proof_of_work/randomx_factory.rs
+++ b/base_layer/core/src/proof_of_work/randomx_factory.rs
@@ -9,38 +9,38 @@ use std::{
 
 const LOG_TARGET: &str = "c::pow::randomx_factory";
 
-#[derive(Default)]
-pub struct RandomXConfig {
-    pub use_large_pages: bool,
-}
-
-impl From<&RandomXConfig> for RandomXFlag {
-    fn from(source: &RandomXConfig) -> Self {
-        let mut result = RandomXFlag::get_recommended_flags();
-        if source.use_large_pages {
-            result |= RandomXFlag::FLAG_LARGE_PAGES
-        }
-        result
-    }
-}
-
 #[derive(Clone)]
 pub struct RandomXVMInstance {
     // Note: If the cache and dataset drops, the vm will be wonky, so have to store all
     // three for now
     instance: Arc<Mutex<(RandomXVM, RandomXCache, RandomXDataset)>>,
+    flags: RandomXFlag,
 }
 
 impl RandomXVMInstance {
     // Note: Can maybe even get more gains by creating a new VM and sharing the dataset and cache
-    pub fn new(key: &[u8]) -> Result<Self, RandomXError> {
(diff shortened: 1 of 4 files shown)


base_layer/core/src/proof_of_work/randomx_factory.rs
```diff
@@ -9,13 +9,13 @@ use std::{
...
 pub struct RandomXConfig {
     pub use_large_pages: bool,
 }
 
-impl From<&RandomXConfig> for RandomXFlag {
-    fn from(source: &RandomXConfig) -> Self {
+impl From<RandomXConfig> for RandomXFlag {
+    fn from(source: RandomXConfig) -> Self {
         let mut result = RandomXFlag::get_recommended_flags();
         if source.use_large_pages {
```
StriderDM: I think `source.use_large_pages` here should be removed. Useful for mining on supported machines, not for validation.

hansieodendaal: Yip, it seems that if `RandomXConfig` is not exposed in the config file and can only either be recommended or default (as implemented), then we do not need that struct at all,

base_layer/core/src/proof_of_work/randomx_factory.rs
```diff
@@ -118,12 +131,40 @@ impl RandomXFactoryInner {
...
+
+        let key = b"another-key";
+        let vm = factory.create(&key[..]).unwrap();
+        assert_ne!(vm.calculate_hash(&preimage[..]).unwrap(), hash1);
+    }
+
+    #[test]
+    fn large_page_fallback() {
+        // This only tests the fallback branch on platforms that do not support large pages (e.g. MacOS)
+        let factory = RandomXFactory::new(RandomXConfig { use_large_pages: true }, 1);
```
StriderDM: We should remove `use_large_pages`, it's an easy error to make to set this to true where it should be false.

RandomXFlag::get_recommended_flags() is the recommended way to use the library (perhaps fall back to RandomXFlag::FLAG_DEFAULT if this fails for any reason).

base_layer/core/src/proof_of_work/randomx_factory.rs
sdbondi: Inclined to agree, removed

SWvheerden (approved):
LGTM

PR #3088, opened 2021-07-13
[WIP][Tests] Add better prune mode tests

SWvheerden:
## Description

This fixes the horizon sync tests. 
This adds in a test to scan the log files for errors

## Motivation and Context

Currently as in this PR, the scan for errors, only scans for the "ERROR" word, but this will lead to many false negatives that will cause the tests to fail. 

## Checklist:


SWvheerden:
close for pr #3122

PR #3092, opened 2021-07-13
Fix when ESC is pressed while adding contact.

Cifko:
## Description
When ESC was pressed while adding contacts, the UI was broken. Still expecting to add the contacts, but didn't show anything. This PR fixes that. ESC properly ends adding contact and hides the contact list.


SWvheerden (approved):
LGTM tested and works as advertised

PR #3093, opened 2021-07-13
Clear the console after seeing the seed words.

Cifko:
## Description
Clearing the console after seeing the seed words.
The whole console is cleared, including history. The question is "are we okay with this approach?"
Need to be tested on different system (this was tested on windows only).


SWvheerden (approved):
LGTM, test and works just as advertised

stringhandler (approved):

PR #3096, opened 2021-07-14
Add contact lookup in wallet.

Cifko:
## Description
Show alias if present in contact list. If it's not there return pub key.
The pub key is still shown in the transactions details. That should stay there.

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/app_state.rs
+++ b/applications/tari_console_wallet/src/ui/state/app_state.rs
@@ -50,7 +50,7 @@ use tari_core::transactions::{
     tari_amount::{uT, MicroTari},
     types::PublicKey,
 };
-use tari_crypto::tari_utilities::hex::Hex;
+use tari_crypto::{ristretto::RistrettoPublicKey, tari_utilities::hex::Hex};
 use tari_shutdown::ShutdownSignal;
 use tari_wallet::{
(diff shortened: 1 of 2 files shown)


applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -152,6 +152,24 @@ impl AppState {
         Ok(())
     }
 
+    // Return alias or pub key if the contact is not in the list.
+    pub fn get_alias(&self, pub_key: &RistrettoPublicKey) -> String {
+        let pub_key_hex = format!("{}", pub_key);
+        // TODO: We can uncomment this to indicated unknown origin, otherwise there is our pub key.
```
delta1, line 158: Self-spends are possible, so imported utxos are not the only case where this public key is our own.

Cifko: So something like "This wallet"? Or we can omit that completely, and who wants it can add him/herself as a contact.

stringhandler: Yeah, I would prefer "Self" or "Me" to "Unknown"

stringhandler (commented):
Happy, just see the reply to the "unknown" comment

stringhandler:
I'm going to merge it as is though, since the code in question is commented

PR #3102, opened 2021-07-16
Installer option layout

CjS77:
The purpose here is to ultimately have a universal installer executable that runs on all platforms, and allows the user to interactively set up their Tari system including:

* base node
* wallet
* miner(s)
* (TBA) validator nodes

The installer will re-implement and encompass the functionality provided by the many scripts we currently have for each platform.

The intent is that users can also re-run the installer without destroying existing data (e.g. they want to add mining) and it gets handled elegantly.

The two main installation routes are:
* Compile form source
* Run in dockerised containers
 
We could also continue to support the "run from binary" option, though I'm not sure that it's necessary.

Basic Roadmap
* [ ] Describe extremely high level installer options - IN PROGRESS
* [ ] Base node docker route
* [ ] Wallet docker route
* [ ] Miner docker route
...


stringhandler (dismissed):
Looking great so far

CjS77:
Cheery picked in another PR

PR #3107, opened 2021-07-19
Add support for choosing Termion backend in console_wallet

philipr-za:
## Description
This PR adds support for using the Termion terminal backend in the console_wallet. This is chosen via a command line switch `console_wallet —termion`. If the switch is not present the default cross term backend will be used.

## Motivation and Context
@CjS77 was having some trouble with the tui in a Docker instance so this is to provide an alternative backend that might work better in Docker. Untested in Docker at this point.

## How Has This Been Tested?
manually

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2689,6 +2689,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "numtoa"
+version = "0.1.0"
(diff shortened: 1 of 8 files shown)


philipr-za:
Closing this for now, looks like crossterm wasn't the issue in Docker.

PR #3109, opened 2021-07-20
fix: accumulated block data bitmap now contains current stxo indexes

sdbondi:
## Description

Greatly reduces the amount of data stored per block by only storing the
spent indexes at a specific height instead of the entire spent bitmap.

The block chain database now stores a single full deleted bitmap at the
tip.

Breaking changes:

- [x] Blockchain Database (database must be reynced)
- [ ] Blockchain Consensus (hard/soft fork)
- [ ] P2P Network (some protocols are no longer compatible)
- [ ] GRPC (GRPC clients need to be updated)

## Motivation and Context

Bytes of storage per block reduced. At height 8033, the savings are 
130mb on the new database. As the STXO set increases, the size of 
the deleted bitmap typically increases (depends on compressibility of the data
in the bitmap), previously resulting in many megabytes of data per block.
This is no longer the case.

As a result of only having the tip bitmap, we can only calculate the output_mr for the next block 
from the tip. In practice this is all that is needed, however some unit tests needed to be updated.

## How Has This Been Tested?

- Some unit tests updated.
- Base node archival sync
- Currently not able to test pruned sync since it is broken (#3082)
- Mined monero and sha3 blocks 

## Checklist:

BREAKING CHANGE: Blockchain Database (database must be resynced)

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -83,7 +83,7 @@ use fs2::FileExt;
 use lmdb_zero::{ConstTransaction, Database, Environment, ReadTransaction, WriteTransaction};
 use log::*;
 use serde::{Deserialize, Serialize};
-use std::{convert::TryFrom, fmt, fs, fs::File, path::Path, sync::Arc, time::Instant};
+use std::{convert::TryFrom, fmt, fs, fs::File, ops::Deref, path::Path, sync::Arc, time::Instant};
 use tari_common_types::{
     chain_metadata::ChainMetadata,
     types::{BlockHash, BLOCK_HASH_LENGTH},
@@ -263,6 +263,11 @@ impl LMDBDatabase {
                 },
                 UpdateDeletedBlockAccumulatedDataWithDiff { header_hash, deleted } => {
                     self.update_deleted_block_accumulated_data_with_diff(&write_txn, header_hash, deleted)?;
+                },
+                UpdateDeletedBitmap { deleted } => {
+                    let mut bitmap = self.load_deleted_bitmap_model(&write_txn)?;
+                    bitmap.merge(&deleted)?;
+                    bitmap.finish()?;
                 },
                 PruneOutputsAndUpdateHorizon {
                     output_positions,
@@ -484,7 +489,7 @@ impl LMDBDatabase {
         k: MetadataKey,
         v: MetadataValue,
     ) -> Result<(), ChainStorageError> {
-        lmdb_replace(txn, &self.metadata_db, &(k as u32), &v)?;
+        lmdb_replace(txn, &self.metadata_db, &k.as_u32(), &v)?;
         Ok(())
     }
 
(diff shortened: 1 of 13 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1991,6 +2035,88 @@ impl fmt::Display for MetadataValue {
...
+        Self {
+            txn,
+            db,
+            cache: None,
+            is_dirty: false,
+        }
+    }
+
+    /// Returns a reference to the `DeletedBitmap`, loading the bitmap if required.
+    pub fn get(&mut self) -> Result<&DeletedBitmap, ChainStorageError> {
```
stringhandler: Strange that this method requires a mut, perhaps a better name or comment explaining why the mut is necessary

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1991,6 +2035,88 @@ impl fmt::Display for MetadataValue {
...
+                write!(f, "Deleted Bitmap ({} indexes)", deleted.bitmap().cardinality())
+            },
+        }
+    }
+}
+
+/// A struct that wraps a LMDB transaction and provides an interface to valid operations that can be performed
+/// on the current deleted bitmap state of the blockchain.
+/// A deleted bitmap contains the MMR leaf indexes of spent UTXOs.
+struct BlockchainDeletedBitmap<'a, T> {
```
stringhandler: nit: The name on this should perhaps reflect that it is meant to be used in an operation and not as a model struct that is used in other layers.

Perhaps `DeletedBitmapUpdater`, `DeletedBitmapOperation` or `DeletedBitmapEditor`, `..Mutator` idk

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -264,6 +264,9 @@ impl LMDBDatabase {
                 UpdateDeletedBlockAccumulatedDataWithDiff { header_hash, deleted } => {
                     self.update_deleted_block_accumulated_data_with_diff(&write_txn, header_hash, deleted)?;
                 },
+                UpdateDeletedBitmap { deleted } => {
```
stringhandler, line 285: I don't like exposing this operation on its own. The deleted bitmap should only ever be updated as part of another operation, so that the database is alway consistent. Where is this operation called on it's own?

sdbondi: Agree with that, but trying to avoid refactoring existing code too much. It's used in pruned mode where there are a number of operations that must be atomic to not break the database and could potentially be brought together into one operation.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: I've changed the code to not require get to be mutable

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: The intention here (although minimally executed) is for this struct to act as a model that contains the data and performs well-defined operations on the data. I was playing with this idea for how we should structure entities within the db in general. I think the cache/lazy load idea was flawed though, so I've removed that in favour of loading the model to instantiate it

sdbondi: I named this `DeletedBitmapModel` but open to suggestions (including removing the whole idea and using lmdb_xxx directly as we've been doing)

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
stringhandler: Let's add a comment there to discourage it's use

stringhandler (approved):

PR #3111, opened 2021-07-20
test: fix prune mode test

SWvheerden:
## Description

Currently, all prune mode cucumber tests do not test prune nodes, but rather runs as archival nodes. This fixes the prune mode test to correctly run with prune mode nodes.
## Checklist:


StriderDM (approved):
utACK

SWvheerden:
This scan for errors in the log file is way too sensitive and makes all cucumber tests flaky. Closing this for PR #3122

PR #3121, opened 2021-07-20
fix: console wallet now recognises wallet.network comms settings

sdbondi:
## Description

- fix: setup comms transport using application name prefix (`base_node`/ `wallet`)
- feat!: default transport is now tor
- feat: network can be configured by one of (in order) `common.network`,
  `{application}.network` and `base_node.network` (for backward-compatibility)
- chore: cleanup transport setup code in base node and common utilities
- chore: update preset config

BREAKING CHANGE: if you relied on default transport configuration (TCP
transport) previously, you will now need to set `transport = "tcp"`

## Motivation and Context

Some comms configuration on the console wallet was not possible or was configured 
using the base_node key. This makes docker configuration difficult.   

## How Has This Been Tested?

Tested by running console wallet and base node. Tested console wallet with tor_forward_address set.

## Checklist:

### DIFF
--- a/common/src/configuration/utils.rs
+++ b/common/src/configuration/utils.rs
@@ -1,4 +1,10 @@
-use crate::{dir_utils::default_subdir, ConfigBootstrap, ConfigError, LOG_TARGET};
+use crate::{
+    configuration::bootstrap::ApplicationType,
+    dir_utils::default_subdir,
+    ConfigBootstrap,
+    ConfigError,
+    LOG_TARGET,
+};
 use config::Config;
 use log::{debug, info};
 use multiaddr::{Multiaddr, Protocol};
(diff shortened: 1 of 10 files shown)


common/src/configuration/utils.rs
```diff
@@ -281,45 +287,64 @@ fn set_mining_node_defaults(cfg: &mut Config) {
...
-        .unwrap();
+fn set_transport_defaults(cfg: &mut Config) -> Result<(), config::ConfigError> {
+    let apps = &[ApplicationType::BaseNode, ApplicationType::ConsoleWallet];
+    for app in apps {
+        let app = app.as_config_str();
+        // Mainnet
+        cfg.set_default(&format!("{}.mainnet.transport", app), "tor")?;
+        cfg.set_default(
+            &format!("{}.mainnet.tcp_listener_address", app),
+            "/ip4/0.0.0.0/tcp/18089",
```
philipr-za: Shouldn't the wallet and base node have different listening ports when using TCP?

sdbondi: Ah good catch, I'll fix that

stringhandler (dismissed):

sdbondi:
Fixed a broken doc test

stringhandler (dismissed):

CjS77 (dismissed):
Nice

sdbondi:
Set `common.network` default instead of `base_node.network`, doctest picked this up

stringhandler (approved):

PR #3122, opened 2021-07-20
test: fix prune mode test

SWvheerden:
add pruned mode pruning check

Currently, all prune mode cucumber tests do not test prune nodes, but rather runs as archival nodes. This fixes the prune mode test to correctly run with prune mode nodes.
This also refactors the chain_meta to not send out the config setting pruning horizon, but rather send out the actual effective pruned height. 
## How Has This Been Tested?

Tests that prune node test actual pass after the fix, but fail before the sync. 
Added a test that ensures prune nodes perform pruning.

## Checklist:

### DIFF
--- a/applications/tari_app_grpc/proto/base_node.proto
+++ b/applications/tari_app_grpc/proto/base_node.proto
@@ -241,9 +241,8 @@ message MetaData {
     uint64 height_of_longest_chain = 1;
     // The block hash of the current tip of the longest valid chain, or `None` for an empty chain
     bytes best_block = 2;
-    // The number of blocks back from the tip that this database tracks. A value of 0 indicates that all blocks are
-    // tracked (i.e. the database is in full archival mode).
-    uint64 pruning_horizon = 4;
+    // This is the min height this node can provide complete blocks for. A 0 here means this node is archival and can provide complete blocks for every height.
+    uint64 pruned_height = 6;
     // The current geometric mean of the pow of the chain tip, or `None` if there is no chain
     bytes accumulated_difficulty = 5;
 }
--- a/base_layer/core/src/base_node/proto/chain_metadata.rs
+++ b/base_layer/core/src/base_node/proto/chain_metadata.rs
@@ -40,14 +40,20 @@ impl TryFrom<proto::ChainMetadata> for ChainMetadata {
         let mut acc_diff = [0; ACC_DIFFICULTY_ARRAY_LEN];
         acc_diff.copy_from_slice(&metadata.accumulated_difficulty[0..ACC_DIFFICULTY_ARRAY_LEN]);
         let accumulated_difficulty = u128::from_be_bytes(acc_diff);
+        let height_of_longest_chain = metadata
+            .height_of_longest_chain
+            .ok_or_else(|| "Height of longest chain is missing".to_string())?;
+        let pruning_horizon = if metadata.pruned_height == 0 {
+            metadata.pruned_height
+        } else {
+            height_of_longest_chain.saturating_sub(metadata.pruned_height)
+        };
 
         Ok(ChainMetadata::new(
-            metadata
-                .height_of_longest_chain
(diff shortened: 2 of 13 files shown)


base_layer/core/src/base_node/proto/chain_metadata.rs
```diff
@@ -40,13 +40,19 @@ impl TryFrom<proto::ChainMetadata> for ChainMetadata {
         let mut acc_diff = [0; ACC_DIFFICULTY_ARRAY_LEN];
         acc_diff.copy_from_slice(&metadata.accumulated_difficulty[0..ACC_DIFFICULTY_ARRAY_LEN]);
         let accumulated_difficulty = u128::from_be_bytes(acc_diff);
+        let height_of_longest_chain = metadata
+            .height_of_longest_chain
+            .ok_or_else(|| "Height of longest chain is missing".to_string())?;
+        let pruning_horizon = if metadata.effective_pruned_height == 0 {
+            metadata.effective_pruned_height
+        } else {
+            height_of_longest_chain - metadata.effective_pruned_height
```
sdbondi: This is untrusted data, so use `saturating_sub` here (`Sub` impl will panic on underflow)

applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -235,9 +235,8 @@ message MetaData {
     uint64 height_of_longest_chain = 1;
     // The block hash of the current tip of the longest valid chain, or `None` for an empty chain
     bytes best_block = 2;
-    // The number of blocks back from the tip that this database tracks. A value of 0 indicates that all blocks are
-    // tracked (i.e. the database is in full archival mode).
-    uint64 pruning_horizon = 4;
+    // This is the min height this node can provide complete blocks for. A 0 here means this node is archival and can provide complete blocks for every height.
+    uint64 effective_pruned_height = 6;
```
sdbondi: This was renamed to `pruned_height`, so better to keep it consistent.

integration_tests/features/Propagation.feature
```diff
@@ -90,4 +90,18 @@ Feature: Block Propagation
...
+    Given I have a pruned node PNODE1 connected to node SENDER with pruning horizon set to 5
+    When I mine a block on SENDER with coinbase CB1
+    When I mine 2 blocks on SENDER
+    When I create a transaction TX1 spending CB1 to UTX1
+    When I submit transaction TX1 to SENDER
+    When I mine 1 blocks on SENDER
+    Then TX1 is in the MINED of all nodes
+    When I mine 17 blocks on SENDER
+    Then all nodes are on the same chain at height 20
+    Then node PNODE1 has its effective pruned height at 15
```
sdbondi:
```suggestion
    Then node PNODE1 has a pruned height of 15
```

sdbondi (commented):
Looks good - left some comments

applications/tari_app_grpc/proto/base_node.proto
SWvheerden: Do you want to rename the RPC call to just pruned_height as well? I named this to match the RPC proto file name

stringhandler: I prefer `pruned_height` as well. Fun fact, I suggested the name `effective_pruned_height` to Yuko a long time ago and I always regretted the long name

integration_tests/helpers/baseNodeClient.js
```diff
@@ -83,6 +83,13 @@ class BaseNodeClient {
       .then((tip) => parseInt(tip.metadata.height_of_longest_chain));
   }
 
+  getEffectiveHeight() {
```
stringhandler: This name is misleading, should be `getPrunedHeight`

stringhandler (dismissed):

stringhandler (approved):

sdbondi (approved):
utACK

PR #3127, opened 2021-07-22
feat(comms): tcp-only p2p protocol listener

sdbondi:
## Description

- Allow an optional "auxilary" TCP listener to be configured on the base
  node, in addition to the usual transport.
- Allow the SOCKS transport to have a set of addresses that "bypass" the
  SOCKS proxy and protocol, and instead directly connect using TCP.
- Configuration for base node and wallet are hooked up for these.
- Remove `ConnectionManagerEvent::Listen` and `ConnectionManagerEvent::ListenFailed`
  events. These are replaced by a simpler 'wait_until_listening' call in
  the connection manager handle that waits for both listeners.
- Add `[app].[network].auxilary_tcp_listener_address` config item
- Add `[app].[network].tor_proxy_bypass_addresses` config item

## Motivation and Context

Allow wallet and base node to communicate directly

image

wallet:
```toml
 # set local address in the UI, or as a seed here:
peer_seeds = ["f27a6ae0321b471d1f65988ded21c48dc57a71edebb936837db1302a91ddc122::/dns4/my-basenode/tcp/9998"]
# Because the tor transport still (correctly) uses the tor network for DNS/IP addresses, we have to explicitly list which addresses to bypass. 
tor_proxy_bypass_addresses = ["/dns4/my-basenode/tcp/9998"]
```

base node:
```toml
auxilary_tcp_listener_address = "/ip4/127.0.0.1/tcp/9998"
```

The config will need to be updated when #3121 is merged

## How Has This Been Tested?

- New connection manager unit test
- Tested using configured console wallet and base node

## Checklist:

### DIFF
--- a/applications/tari_app_utilities/src/utilities.rs
+++ b/applications/tari_app_utilities/src/utilities.rs
@@ -172,6 +172,7 @@ pub fn create_transport_type(config: &GlobalConfig) -> TransportType {
             tor_socks_config: tor_socks_address.map(|proxy_address| SocksConfig {
                 proxy_address,
                 authentication: tor_socks_auth.map(convert_socks_authentication).unwrap_or_default(),
+                proxy_bypass_addresses: vec![],
             }),
         },
         CommsTransport::TorHiddenService {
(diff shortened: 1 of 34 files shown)


stringhandler (dismissed):
utACK

stringhandler (approved):

PR #3128, opened 2021-07-22
feat(wallet): add extra feedback to recovery monitoring callback in Wallet FFI

philipr-za:
## Description
This PR adds an extra stage of callback feedback to the recovery process in the Wallet FFI. The new stage shows when connection’s are being attempted and not just when they are made.

Edit: Added even more detail to the callback, this changes the FFI interface to support extra argument in the callback

## How Has This Been Tested?
For testing in the mobile client

## Checklist:


sdbondi (dismissed):
utACK

stringhandler (approved):

PR #3129, opened 2021-07-22
fix: replace usage of RangeProof MR with Witness MR

SWvheerden:
## Description

There are still some names references a rangeproof Mr, this does not exist anymore and there is only a witness Mr. 

## Checklist:


base_layer/core/src/validation/error.rs
```diff
@@ -59,8 +59,8 @@ pub enum ValidationError {
     InvalidOutputMr,
     #[error("Invalid kernel merkle root")]
     InvalidKernelMr,
-    #[error("Invalid range proof merkle root")]
-    InvalidRangeProofMr,
+    #[error("Invalid witness merkle root")]
+    InvalidWitnessMr,
```
sdbondi: This variant isnt used (assuming everything compiles) and can be removed. Did a quick search and same for `InvalidKernelMr` and `InvalidOutputMr` can be removed

sdbondi: ...and `WeakerAccumulatedDifficulty`, they are all handled in `BlockError(..)` and `PowError(...)`

stringhandler (dismissed):
Approved, but agree that they can be removed

stringhandler (approved):

sdbondi (approved):

PR #3130, opened 2021-07-22
feat: add persistent dedup cache for message hashes

hansieodendaal:
## Description
- Added a persistent dedup message hash cache for messages that will be trimmed to capacity every 5 minutes:
  - dedup cache size configurable for users;
  - trimming period configurable.
- Added a unit test to test the dedup cache operation.

## Motivation and Context
A recent stress test highlighted duplicate stages of transaction messages still being sent hours after a transaction has been mined, evoking a flood of SAF messages each time. A wallet also had no way of knowing which transactional message had been processed already. The dedup cache was changed from _time-to-live_ to being _persistent_.

## How Has This Been Tested?
- System-level testing under stress test conditions.
- New unit test.

## Checklist:


comms/dht/src/config.rs
```diff
@@ -155,8 +155,8 @@ impl Default for DhtConfig {
             saf_auto_request: true,
             saf_max_message_size: 512 * 1024,
             saf_minimum_request_period: Duration::from_secs(3 * 24 * 60 * 60), // 3 days
-            msg_hash_cache_capacity: 100_000,
-            msg_hash_cache_ttl: Duration::from_secs(5 * 60),
+            dedup_cache_capacity: 2_500,
+            dedup_cache_trim_period: Duration::from_secs(5 * 60),
```
sdbondi: nit: use interval in the name to indicate that it's an interval between repeating
```suggestion
            dedup_cache_trim_interval: Duration::from_secs(5 * 60),
```

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+        debug!(
+            target: LOG_TARGET,
+            "Message dedup cache capacity initialized at {}", capacity,
+        );
+        Self { connection, capacity }
+    }
+
+    /// Inserts and returns Ok(true) if the item already existed and Ok(false) if it didn't
+    pub async fn insert_body_hash_if_unique(&self, body_hash: Vec<u8>) -> Result<bool, StorageError> {
+        let body_hash_string = DedupCacheDatabase::body_hash_string(body_hash.clone());
```
sdbondi: This function takes a hash, so won't need to hash it again

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+        hex::to_hex(&Challenge::new().chain(body_hash).finalize())
+    }
+
+    fn formatted_naive_date_time() -> Result<NaiveDateTime, StorageError> {
+        NaiveDateTime::parse_from_str(
+            Utc::now().naive_utc().format("%Y-%m-%d %H:%M:%S").to_string().as_str(),
+            "%Y-%m-%d %H:%M:%S",
+        )
+        .map_err(|e| StorageError::ParseError(e.to_string()))
+    }
```
sdbondi: ?
```suggestion
   Utc::now().naive_utc()
```

comms/dht/src/storage/error.rs
```diff
@@ -30,6 +30,10 @@ pub enum StorageError {
     InvalidUnicodePath,
     #[error("ConnectionError: {0}")]
     ConnectionError(#[from] diesel::ConnectionError),
+    #[error("UniqueViolation")]
+    UniqueViolation(String),
+    #[error("ParseError")]
+    ParseError(String),
```
sdbondi: Not needed (formatted_naive_date_time is not needed)

sdbondi (commented):
Nice - there's one issue that I think should be resolved, and one minor code issue

comms/dht/src/actor.rs
```diff
@@ -282,14 +295,23 @@ impl DhtActor {
...
+                        Ok(already_exists) => {
+                            let _ = reply_tx.send(already_exists).map_err(|_| DhtActorError::ReplyCanceled);
+                        },
+                        Err(err) => {
+                            warn!(
+                                target: LOG_TARGET,
+                                "Unable to update message dedup cache because {:?}", err
+                            );
+                            let _ = reply_tx.send(false).map_err(|_| DhtActorError::ReplyCanceled);
+                            // Err(DhtActorError::StorageError(err))
```
sdbondi:
```suggestion

```

comms/dht/src/dedup/dedup_cache.rs
hansieodendaal: This function is to make it easier for humans to read the database entries; database field `stored_at` has a default `DEFAULT CURRENT_TIMESTAMP` and its format is automatically entered as `2021-07-23 04:01:14`. When the `last_hit_at` field is populated with `Utc::now().naive_utc()` it prints something like `2021-07-23 04:01:14.235873992`, which makes it difficult to compare visually.

comms/dht/src/storage/error.rs
hansieodendaal: Please see the previous comment.

SWvheerden (dismissed):
LGTM

comms/dht/src/dedup/dedup_cache.rs
stringhandler: I don't buy this reasoning. I would just store the data as is. SQL has the capability to display the data differently via SELECT statements for the poor human users who are struggling

common/config/presets/tari_config_example.toml
```diff
@@ -53,6 +53,9 @@ network = "weatherwax"
 #buffer_rate_limit_base_node = 20
 # - Rate limit for the base node wallet (min value = 5, default value = 20).
 #buffer_rate_limit_base_node_wallet = 20
+# The message deduplication persistent cache size - messages with these hashes in the cache will only be processed once.
+# The cache will also be trimmed down to size periodically (min value = 100, default value = 2500).
```
stringhandler: Why not allow a min value of 0 for testing?

common/config/presets/tari_config_example.toml
```diff
@@ -136,7 +139,8 @@ base_node_query_timeout = 120
 # (options: "DirectOnly", "StoreAndForwardOnly", DirectAndStoreAndForward". default: "DirectAndStoreAndForward").
 #transaction_routing_mechanism = "DirectAndStoreAndForward"
 
-scan_for_utxo_interval=60
+# UTXO scanning service interval (default = 12 hours, i.e. 60 * 60 * 12 seconds)
```
stringhandler, line 142: Did this sneak in?

comms/dht/migrations/2021-07-21-0846252_add_dedup_cache/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE dedup_cache (
```
stringhandler: Should we not add the sender address in here for inspection purposes?

comms/dht/src/actor.rs
```diff
@@ -236,6 +241,8 @@ impl DhtActor {
 
         let mut pending_jobs = FuturesUnordered::new();
 
+        let mut cleanup_ticker = time::interval(self.config.dedup_cache_trim_interval).fuse();
```
stringhandler:
```suggestion
        let mut dedup_cache_trim_ticker = time::interval(self.config.dedup_cache_trim_interval).fuse();
```

comms/dht/src/builder.rs
```diff
@@ -89,13 +89,13 @@ impl DhtBuilder {
...
     }
 
-    pub fn with_signature_cache_ttl(mut self, ttl: Duration) -> Self {
-        self.config.msg_hash_cache_ttl = ttl;
+    pub fn with_signature_dedup_cache_trim_interval(mut self, trim_interval: Duration) -> Self {
+        self.config.dedup_cache_trim_interval = trim_interval;
         self
     }
 
     pub fn with_signature_cache_capacity(mut self, capacity: usize) -> Self {
```
stringhandler: This method has a misleading name, where does 'signature' come from

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+use std::cmp::max;
+use tari_crypto::tari_utilities::ByteArray;
+use tari_utilities::hex;
+
+const LOG_TARGET: &str = "comms::dht::dedup_cache";
+
+//#[derive(Clone, Debug, Queryable, Insertable, PartialEq)]
+#[derive(Clone, Debug, Queryable, Identifiable)]
+#[table_name = "dedup_cache"]
+struct DedupCacheSql {
```
stringhandler: Please split these structs into individual files

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+        self.connection
+            .with_connection_async(move |conn| {
+                let insert_result = diesel::insert_into(dedup_cache::table)
+                    .values(UpdateDedupCacheSql {
+                        body_hash: Some(body_hash.clone()),
+                        number_of_hits: Some(1),
+                        last_hit_at: Some(DedupCacheDatabase::formatted_naive_date_time()?),
+                    })
+                    .execute(conn)
+                    .map_err(Into::into);
```
stringhandler:
```suggestion
                    ?;
```

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+                        DatabaseErrorKind::UniqueViolation => Err(StorageError::UniqueViolation(body_hash)),
+                        _ => Err(diesel::result::Error::DatabaseError(kind, e_info).into()),
+                    },
+                    Err(e) => Err(e.into()),
+                }
+            })
+            .await
+    }
+
+    async fn update_number_of_hits(&self, body_hash: String) -> Result<usize, StorageError> {
```
stringhandler: nit: this method does not take in a parameter, so increment is more accurate than update

```suggestion
    async fn increment_number_of_hits(&self, body_hash: String) -> Result<usize, StorageError> {
```

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+                let mut num_removed = 0;
+                let msg_count = dedup_cache::table
+                    .select(dsl::count(dedup_cache::id))
+                    .first::<i64>(conn)? as usize;
+                // Hysteresis added to minimize database impact
+                if msg_count > capacity {
+                    let remove_count = msg_count - capacity;
+                    let message_ids: Vec<i32> = dedup_cache::table
+                        .select(dedup_cache::id)
+                        .order_by(dedup_cache::stored_at.asc())
```
stringhandler: To match the previous TtlCache behaviour, you should actually sort by `last_hit_at` instead

comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+                        .limit(remove_count as i64)
+                        .get_results(conn)?;
+                    num_removed = diesel::delete(dedup_cache::table)
```
stringhandler, line 91: You should combine the above select into this delete so that it is atomic (ideal), or otherwise wrap both in a transaction and change the select to lock the table for deletion (less ideal)

stringhandler (commented):
Lots of comments, but I really do like this PR. Nice work

common/config/presets/tari_config_example.toml
hansieodendaal: That setting did not have a comment, and the spacing was not consistent, so it was fixed. No functional change.

comms/dht/src/dedup/dedup_cache.rs
hansieodendaal: This converts `QueryResult<usize>` into `Result<usize, StorageError>`; the suggested change results in `usize`.
The `UniqueViolation` error is used as a feature to determine if the entry is already present.

comms/dht/src/dedup/dedup_cache.rs
sdbondi: Good catch

comms/dht/src/dedup/dedup_cache.rs
sdbondi: Everything within `with_connection_async` is protected with an async Mutex. Can't remember now why I didn't implement SQL transactions (may have been time pressure), but `with_connection_async` guarantees you are the only one using the single connection (we don't use a connection pool, because it was never perf critical and caused issues on mobile). I think (but you'd need to check) that the DeleteStatement doesnt support limit, which causes an issue - otherwise, yeah they should ideally be in one query.

comms/dht/src/dedup/dedup_cache.rs
hansieodendaal: The TTL principle will remove the oldest entries first, similar to sorting on `stored_at`. Using `last_hit`, a 24-hour old entry with a recent hit will remain in the cache.

comms/dht/src/dedup/dedup_cache.rs
sdbondi: `SELECT  STRFTIME('%Y-%m-%d %H:%M:%S', last_hit_at) as last_hit_ts FROM dedup_cache ORDER BY last_hit_at` would do it.

comms/dht/src/storage/error.rs
sdbondi: Even if we keep the function, the error you return can never happen (it can never fail to parse because the input comes from a guaranteed well-formed string).

comms/dht/migrations/2021-07-21-0846252_add_dedup_cache/up.sql
hansieodendaal: Good idea

common/config/presets/tari_config_example.toml
stringhandler: my bad, sorry

comms/dht/src/dedup/dedup_cache.rs
stringhandler: I think the `?` operator is the same as `map_err(Into::into)`.

comms/dht/src/dedup/dedup_cache.rs
stringhandler: In the old TTL cache, when you insert a row, you give it a duration, when that expires, it is removed. When you re-insert an existing value, you insert the same duration, basically extending it's life, so if the duration was 24 hours and it was at 23 hours and a dup message came it, it would live another 24 hours.

```
                let already_exists = self
                    .msg_hash_cache
                    .insert(hash, (), self.config.msg_hash_cache_ttl)
 ```

File: comms/dht/src/dedup/dedup_cache.rs (hunk 3 above)
Comment (stringhandler) [reply comms/dht/src/dedup/dedup_cache.rs]: Happy to merge once this change is applied

Comment (hansieodendaal) [reply comms/dht/src/dedup/dedup_cache.rs]: Happy to remove `fn  formatted_naive_date_time `

File: comms/dht/src/storage/error.rs (hunk 4 above)
Comment (hansieodendaal) [reply comms/dht/src/storage/error.rs]: Removed `ParseError(String)` along with `fn formatted_naive_date_time `

File: comms/dht/src/dedup/dedup_cache.rs (hunk 14 above)
Comment (hansieodendaal) [reply comms/dht/src/dedup/dedup_cache.rs]: I concede about `order_by(dedup_cache::stored_at.asc())` being different than the original TTL.

My proposal here was actually that the oldest messages are always removed from the cache, not that a very old message with a recent hit gets an extended lifetime in the cache. This way part of a conversation can be replayed after the message's original stay in the cache expired if it was legitimate.

Changing to sort by `last_hit_at` as proposed.

File: comms/dht/src/dedup/dedup_cache.rs (hunk 12 above)
Comment (hansieodendaal) [reply comms/dht/src/dedup/dedup_cache.rs]: It  results in `usize` instead of `Result<usize, StorageError>`

File: comms/dht/src/actor.rs [hunk 16]
```diff
@@ -707,14 +736,114 @@ mod test {
...
+            assert!(!is_dup);
+        }
+        // Try to re-insert all; everything should be marked as duplicates due to cleanup ticker not active yet
+        for key in &signatures {
+            let is_dup = actor
+                .msg_hash_dedup_cache
+                .insert_body_hash_if_unique(key.clone(), CommsPublicKey::default())
+                .await
+                .unwrap();
+            // let is_dup = requester.insert_message_hash(key.clone()).await.unwrap();
```
sdbondi: elsewhere too
```suggestion

```

comms/dht/src/dedup/mod.rs
```diff
@@ -79,7 +85,10 @@ where
                 message.tag,
                 message.dht_header.message_tag
             );
-            if dht_requester.insert_message_hash(hash).await? {
+            if dht_requester
+                .insert_message_hash(hash, message.source_peer.public_key.clone())
```
sdbondi, line 87: Just a note, the source_peer is the source of this message. That is, the peer that it was directly received from perhaps via propagation, but not necessarily the peer that created the original message. You would expect this stored value for the same message to be different across peers. The origin_peer is the authenticated origin (the peer that created the message), but is only available to the peer that can decrypt the message. So I'm not too convinced how useful this field will end up being. Just FYI though, nothing against doing it anyway.

comms/dht/src/dedup/dedup_cache.rs
sdbondi: You can remove the `map_err(Into::into)`. The conversion is redundant here because it is already a `diesel::result::Error`, which is the type the match expects.  `Err(e) => Err(e.into()),` converts the error into a `StorageError` so that it can be returned.

comms/dht/src/dedup/mod.rs
hansieodendaal: Noted

hansieodendaal:
This latest commit:
- changes db insert/update operations to be atomic, as a race condition was present between the cache cleanup operation, that ran in a separate thread, to the db insert/update operation. The latter acquired an exclusive lock to the db three times for a single update operation; this caused db entries to be deleted in between where it was determined that an entry is a duplicate and where its stats were updated.
- changes the dedup cache size for base nodes and console wallets to be 25,000 as determined with testing. If the network is busy, between 2,000 and 9,500 new messages will be added to the cache every 5 minutes. The results below were obtained with a dedup cache size of 2,500 for testing purposes to enforce a quick cleanup cycle.

Race condition (during stress testing):
``` rust
2021-07-27 21:44:44.999526100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6006, capacity 2500, removed 3506
2021-07-27 21:44:45.039803500 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:44:45.083666500 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:44:45.539709800 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:44:45.590940200 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:46.558936200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6868, capacity 2500, removed 4368
2021-07-27 21:49:47.044757300 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:47.340537900 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:47.497784400 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:47.568265400 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:48.078426700 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:48.446095100 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:49:48.834408000 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:45.755165300 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6707, capacity 2500, removed 4207
2021-07-27 21:54:45.756107000 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:45.793008600 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:45.922960300 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:45.923114000 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:45.962302800 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.106246100 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.143228200 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.143793900 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.228639500 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.441927700 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.511398900 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
2021-07-27 21:54:46.540093600 [comms::dht::actor] WARN  Unable to update message dedup cache because ResultError(NotFound)
```

Race condition fixed with atomic insert/update (during stress testing):
``` rust
2021-07-28 07:38:19.546905000 [comms::dht::actor] DEBUG Message dedup cache will be trimmed to capacity every 300s
2021-07-28 07:38:19.546923000 [comms::dht::dedup_cache] DEBUG Message dedup cache capacity initialized at 2500
2021-07-28 07:38:19.642241700 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2529, capacity 2500, removed 29
2021-07-28 07:43:19.571672700 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2702, capacity 2500, removed 202
2021-07-28 07:48:19.633503700 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2725, capacity 2500, removed 225
2021-07-28 07:53:19.572425000 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2990, capacity 2500, removed 490
2021-07-28 07:58:19.569563200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2799, capacity 2500, removed 299
2021-07-28 08:03:19.565731100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2835, capacity 2500, removed 335
2021-07-28 08:08:19.572490500 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2907, capacity 2500, removed 407
2021-07-28 08:13:19.573576600 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 3148, capacity 2500, removed 648
2021-07-28 08:18:19.829696000 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 3688, capacity 2500, removed 1188
2021-07-28 08:23:20.128356000 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 4155, capacity 2500, removed 1655
2021-07-28 08:28:20.441050100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 4217, capacity 2500, removed 1717
2021-07-28 08:33:19.864351900 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 4752, capacity 2500, removed 2252
2021-07-28 08:38:20.316773500 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 4998, capacity 2500, removed 2498
2021-07-28 08:43:20.047123800 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 5206, capacity 2500, removed 2706
2021-07-28 08:48:19.983793200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 5086, capacity 2500, removed 2586
2021-07-28 08:53:19.880013800 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 5190, capacity 2500, removed 2690
2021-07-28 08:58:19.891490100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6431, capacity 2500, removed 3931
2021-07-28 09:03:20.065235800 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 5831, capacity 2500, removed 3331
2021-07-28 09:08:20.141300600 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6032, capacity 2500, removed 3532
2021-07-28 09:13:19.873829200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6804, capacity 2500, removed 4304
2021-07-28 09:18:20.044942500 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 7152, capacity 2500, removed 4652
2021-07-28 09:23:20.038902500 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6551, capacity 2500, removed 4051
2021-07-28 09:28:19.974987800 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6627, capacity 2500, removed 4127
2021-07-28 09:33:19.921464100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 7330, capacity 2500, removed 4830
2021-07-28 09:38:19.826990700 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 7240, capacity 2500, removed 4740
2021-07-28 09:43:19.760811400 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 6969, capacity 2500, removed 4469
2021-07-28 09:48:19.666592200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 7185, capacity 2500, removed 4685
2021-07-28 09:53:19.771869900 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 10304, capacity 2500, removed 7804
2021-07-28 09:58:20.000872200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 12048, capacity 2500, removed 9548
2021-07-28 10:03:20.095486100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 11191, capacity 2500, removed 8691
2021-07-28 10:08:19.613259300 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 9137, capacity 2500, removed 6637
2021-07-28 10:13:19.573098300 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2824, capacity 2500, removed 324
2021-07-28 10:18:19.602043100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2866, capacity 2500, removed 366
2021-07-28 10:23:19.575922000 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2816, capacity 2500, removed 316
2021-07-28 10:28:19.580211400 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2944, capacity 2500, removed 444
2021-07-28 10:33:19.573359100 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 3237, capacity 2500, removed 737
2021-07-28 10:38:19.578616600 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 3150, capacity 2500, removed 650
2021-07-28 10:43:19.575920200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 3125, capacity 2500, removed 625
2021-07-28 10:48:19.571850000 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2823, capacity 2500, removed 323
2021-07-28 10:53:19.579487200 [comms::dht::dedup_cache] DEBUG Message dedup cache: count 2769, capacity 2500, removed 269
```

stringhandler (approved):

sdbondi (approved):

PR #3134, opened 2021-07-25
chore: rename daemon mode to non-interactive mode

sdbondi:
## Description

Simple rename from `--daemon` to `--non-interactive`.

## Motivation and Context

Daemon was a misnomer and did not start a background process.

## How Has This Been Tested?

Console wallet starts without UI when `--non-interactive` is passed to it.

## Checklist:


hansieodendaal (commented):
Hi, some comments, maybe just NITS, otherwise looking good:
- Still picking up `daemon mode` and `--daemon-mode` in `supervisor.md`
- Still picking up `daemon` in `miningNodeProcess.js` and `walletProcess.js`
- Should `WalletMode::RecoveryDaemon` also be changed?

stringhandler (approved):

PR #3135, opened 2021-07-26
test: cucumber saf test

Cifko:
## Description
Integration test for SAF.

This test has some long sleeps, because we need to timeout the connections.
Once we solve the issue we can lower the sleeps (I've added TODO in the code).

### DIFF
--- a/integration_tests/features/WalletRoutingMechanism.feature
+++ b/integration_tests/features/WalletRoutingMechanism.feature
@@ -4,40 +4,59 @@
 @wallet-routing_mechanism @wallet
 Feature: Wallet Routing Mechanism
 
-@flaky
-Scenario Outline: Wallets transacting via specified routing mechanism only
-    Given I have a seed node NODE
-    And I have <NumBaseNodes> base nodes connected to all seed nodes
-    And I have non-default wallet WALLET_A connected to all seed nodes using <Mechanism>
-    And I have mining node MINER connected to base node NODE and wallet WALLET_A
-    And I have <NumWallets> non-default wallets connected to all seed nodes using <Mechanism>
+    @flaky
+    Scenario Outline: Wallets transacting via specified routing mechanism only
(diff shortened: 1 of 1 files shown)


Cifko:
This test is very flaky. I tried it manually as well.

hansieodendaal (commented):
Hi, just some initial comments:
- You may want to add this test to the `WalletRoutingMechanism.feature` instead.
- We already have the test called `Wallets transacting via specified routing mechanism only`. Although this test sends transactions when the other wallet is offline, I am not sure we test anything in addition to what has been tested already in the test mentioned.

integration_tests/features/WalletRoutingMechanism.feature
```diff
@@ -1,43 +1,60 @@
...
+    Scenario: Store and forward TX
+        Given I have a seed node SEED
+        And I have a base node BASE connected to seed SEED
+        And I have wallet SENDER connected to base node BASE
+        And I have wallet RECEIVER connected to base node BASE
+        And I stop wallet RECEIVER
+        And I have mining node MINE connected to base node BASE and wallet SENDER
+        And mining node MINE mines 5 blocks
+        Then I wait for wallet SENDER to have at least 1000000 uT
+        And I send 1000000 uT from wallet SENDER to wallet RECEIVER at fee 100
```
philipr-za, line 54: I am working on a very similar test and ran into an issue when I spotted your WIP PR.

So firstly I found that this line would fail quite often because the way the `send_tari` helper function is written it uses the gRPC client to get the Receiver's pubkey but if you stopped the receiver it can't do that.

Cifko: I've changed the send_tari function (forgot to push, as I was still working on it). Now it's working, but it's the longest test we have :).

stringhandler (approved):

PR #3137, opened 2021-07-26
fix: update transaction and block validator to use full deleted map

sdbondi:
## Description

Prevent double spends from being added to the mempool. This bug
was introduced by the blockchain deleted bitmap storage changes
in #3109. A new block template contained already spent transactions
which would then fail a double spend check in `calculate_mmr_roots`.

## Motivation and Context

Fixes:
```
"Chain storage error: You tried to execute an invalid Database operation: Could not delete index 141535 from the output MMR (length is 94741)" 
```

This error incorrectly attributes the failed deletion to an index that is larger than the total length. 
There are two corrections in this PR:
1. `output_mmr.len() != output_mmr.get_leaf_count()` - length is the number of remaining  unspent txos (`leaf_count - cardinality(deleted_bitmap)`).
2. The delete also can fail if the index is already added. The correct error is emitted in this case.

Expect to see many occurrences of ` WARN  Validation failed due to already spent output` in logs until double spends are cleared

## How Has This Been Tested?

Archival Sync current chain 
Mining non-empty blocks.
Cucumber tests pass (in 
progress)

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -35,6 +35,7 @@ use crate::{
         BlockchainBackend,
         ChainBlock,
         ChainHeader,
+        DeletedBitmap,
         HistoricalBlock,
         HorizonData,
         MmrTree,
@@ -58,6 +59,7 @@ use std::{
     cmp,
     cmp::Ordering,
     collections::VecDeque,
+    convert::TryFrom,
     mem,
     ops::Bound,
     sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
@@ -876,6 +878,12 @@ where B: BlockchainBackend
         let db = self.db_read_access()?;
         db.fetch_horizon_data()
     }
+
+    /// Returns the full deleted bitmap at the current blockchain tip
+    pub fn fetch_deleted_bitmap(&self) -> Result<DeletedBitmap, ChainStorageError> {
+        let db = self.db_read_access()?;
+        db.fetch_deleted_bitmap()
+    }
 }
 
 fn unexpected_result<T>(req: DbKey, res: DbValue) -> Result<T, ChainStorageError> {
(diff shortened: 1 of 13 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -944,24 +952,42 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(db: &T, block: &Block) -> Resul
...
+                            entity: "UTXO".to_string(),
+                            field: "hash".to_string(),
+                            value: output_hash.to_hex(),
+                        })?;
+                debug!(
+                    target: LOG_TARGET,
+                    "0-conf spend for UTXO index {} ({})",
+                    index,
+                    output_hash.to_hex()
+                );
```
SWvheerden: I would make this a trace log, or if it's debug, add some more info. 
If you run your node with debug logs, it's just suddenly going to log this line with zero context.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -158,13 +158,11 @@ fn check_sorting_and_duplicates(body: &AggregateBody) -> Result<(), ValidationEr
 
 /// This function checks that all inputs in the blocks are valid UTXO's to be spend
 fn check_inputs_are_utxos<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), ValidationError> {
-    let data = db
-        .fetch_block_accumulated_data(&block.header.prev_hash)?
-        .ok_or(ValidationError::PreviousHashNotFound)?;
+    let deleted = db.fetch_deleted_bitmap()?;
```
SWvheerden: This is not equivalent. 

this in the old code would have been:
`db.fetch_block_accumulated_data(tip_height).deleted`

I am not sure if we do run this for heights where `block.header.height -1` != `tip_height`

SWvheerden (dismissed):
Looks good, 2 comments

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Yeah agree, can add some context to that message

base_layer/core/src/validation/block_validators.rs
sdbondi: I think we should add a check to this validator (`validate_body_for_valid_orphan`) that the block being validated is next from the tip.

sdbondi: I've added that invariant/rule to the validator. The current usage of this validator is to only ever add (atomically) the next block from the tip.

SWvheerden (dismissed):
LGTM

base_layer/core/src/validation/block_validators.rs
```diff
@@ -125,6 +125,19 @@ impl<B: BlockchainBackend> PostOrphanBodyValidation<B> for BodyOnlyValidator {
     /// 1. Are all inputs and outputs not in the STXO set?
     /// 1. Are the block header MMR roots valid?
     fn validate_body_for_valid_orphan(&self, block: &ChainBlock, backend: &B) -> Result<(), ValidationError> {
+        let metadata = backend.fetch_chain_metadata()?;
```
stringhandler: I don't think this is the correct place to do this. This validator would prevent any kind of reorg if this is the case. I would say move this code to the GetBlockTemplate in the inbound handler

stringhandler (changes requested):

stringhandler (dismissed):
Yeah, much happier with this

stringhandler (approved):

SWvheerden (approved):

base_layer/core/src/validation/block_validators.rs
sdbondi: Changes discussed via DM

hansieodendaal (approved):
LGTM - ran a 3000 total transaction stress test from 3 sending wallets to 1 receiving wallet, with an independent XMrig miner running. Mempool accumulated and cleared as expected as blocks were mined.

hansieodendaal (approved):
LGTM - ran a 3000 total transaction stress test from 3 sending wallets to 1 receiving wallet, with an independent XMrig miner running. Mempool accumulated and cleared as expected as blocks were mined.

sdbondi:
@hansieodendaal Thanks for the tests

PR #3138, opened 2021-07-27
fix: fix prune mode sync

SWvheerden:
## Description

This fixes prune mode sync to work again after the bitmap change. The bitmap saved in the accumulated block data is not the completed bitmap anymore but rather the diff for that block. This means when a bitmap is required to know what the spent state of the utxo set is, we need to build a new bitmap rather than ask the accumulated block data like we used to do. 

This PR lets the blockchain_database send a completed bitmap from the tip out, that can be used to reconstruct a bitmap at the desired height. 

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
@@ -47,7 +47,10 @@ use crate::{
 use croaring::Bitmap;
 use futures::StreamExt;
 use log::*;
-use std::convert::{TryFrom, TryInto};
+use std::{
+    convert::{TryFrom, TryInto},
+    sync::Arc,
+};
 use tari_comms::PeerConnection;
 use tari_crypto::{
     commitment::HomomorphicCommitment,
@@ -206,9 +209,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                 let kernel_pruned_set = block_data.dissolve().0;
                 let mut kernel_mmr = MerkleMountainRange::<HashDigest, _>::new(kernel_pruned_set);
 
-                // let mut kernel_sum = HomomorphicCommitment::default();
                 for kernel in kernels.drain(..) {
-                    // kernel_sum = &kernel.excess + &kernel_sum;
                     kernel_mmr.push(kernel.hash())?;
                 }
 
@@ -531,6 +532,12 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
 
         let mut prev_mmr = 0;
         let mut prev_kernel_mmr = 0;
+        let bitmap = Arc::new(
+            self.db()
+                .fetch_complete_deleted_bitmap_at(header.hash().clone())
+                .await?
+                .into_bitmap(),
+        );
         for h in 0..=header.height() {
             let curr_header = self.db().fetch_chain_header(h).await?;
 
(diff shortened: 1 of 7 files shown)


base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync/horizon_state_synchronization.rs
```diff
@@ -531,6 +532,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
 
         let mut prev_mmr = 0;
         let mut prev_kernel_mmr = 0;
+        let bitmap = Arc::new(self.db().get_complete_deleted_bitmap_at_tip().await?.into_bitmap());
```
stringhandler: 1. The db uses `fetch` as the verb, so let's keep it the same
2. The tip could have somehow changed between the call in line 529, so rather pass in the expected tip so that it can fail if the tip has somehow changed
```suggestion
        let bitmap = Arc::new(self.db().fetch_complete_deleted_bitmap_at(header.hash()).await?.into_bitmap());
```

base_layer/core/src/base_node/sync/rpc/service.rs
```diff
@@ -467,7 +466,27 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
                 if prev_header.height > end_header.height {
                     return Err(RpcStatus::bad_request("start index is greater than end index"));
                 }
+                // we need to construct a temp bitmap for the height the client requested
+                let (mut bitmap, tip_height) = self
+                    .db
+                    .get_complete_deleted_bitmap_at_tip()
```
stringhandler: See above comment, the tip may have changed in between db calls

base_layer/core/src/chain_storage/accumulated_data.rs
```diff
@@ -205,6 +205,31 @@ impl<'de> Visitor<'de> for DeletedBitmapVisitor {
...
+/// Wrapper struct to get a completed bitmap with the height it was created at
+#[derive(Debug, Clone)]
+pub struct CompleteDeletedBitmap {
+    deleted: Bitmap,
+    height: u64,
+}
+
+impl CompleteDeletedBitmap {
+    pub fn new(deleted: Bitmap, height: u64) -> CompleteDeletedBitmap {
+        CompleteDeletedBitmap { deleted, height }
```
stringhandler: Use a hash of the header instead of height if possible (or possibly in addition). This will help to reduce bugs in future

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1521,7 +1521,7 @@ impl BlockchainBackend for LMDBDatabase {
         &self,
         start: u64,
         end: u64,
-        deleted: &Bitmap,
+        deleted: Arc<Bitmap>,
```
stringhandler: This can still be a `&`

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -876,6 +871,13 @@ where B: BlockchainBackend
         let db = self.db_read_access()?;
         db.fetch_horizon_data()
     }
+
+    pub fn get_complete_deleted_bitmap_at_tip(&self) -> Result<CompleteDeletedBitmap, ChainStorageError> {
+        let db = self.db_read_access()?;
+        let deleted = db.fetch_deleted_bitmap()?.into_bitmap();
+        let height = db.fetch_chain_metadata()?.height_of_longest_chain();
```
stringhandler: These two calls make we worried. The height and hash that the deleted bitmap is created/update with should be stored alongside it so that it cannot get out of sync

stringhandler (dismissed):

stringhandler (approved):

PR #3144, opened 2021-07-28
test: cucumber for all wallet commands

Cifko:
## Description
Integration test for wallet commands. It doesn't cover all possible scenarios, but it does cover every command.


integration_tests/features/WalletCli.feature
```diff
@@ -0,0 +1,115 @@
...
+        Given I have a base node BASE
+        And I have wallet SENDER connected to base node BASE
+        And I have wallet RECEIVER connected to base node BASE
+        And I have mining node MINE connected to base node BASE and wallet SENDER
+        And mining node MINE mines 5 blocks
+        Then I wait for wallet SENDER to have at least 1000000 uT
+        And I wait 5 seconds
+        And I stop wallet SENDER
+        And I send 1000000 uT from SENDER to RECEIVER via command
+        And mining node MINE mines 5 blocks
```
SWvheerden: How does this test work?
A stopped wallet should not be able to send?

integration_tests/features/WalletCli.feature
```diff
@@ -0,0 +1,115 @@
...
+        Then the password of wallet WALLET is not kensentme
+        Then the password of wallet WALLET is changedpwd
+
+    Scenario: As a user I want to get balance via command
+        Given I have a base node BASE
+        And I have wallet WALLET connected to base node BASE
+        And I have mining node MINE connected to base node BASE and wallet WALLET
+        And mining node MINE mines 5 blocks
+        Then I wait for wallet WALLET to have at least 1000000 uT
+        And I stop wallet WALLET
```
SWvheerden, line 29: See comment below

integration_tests/features/WalletCli.feature
```diff
@@ -0,0 +1,115 @@
...
+    Scenario: As a user I want to set and clear custom base node for a wallet via command
+        Given I have a base node NODE1
+        And I have a base node NODE2
+        And I have wallet WALLET connected to base node NODE1
+        Then I set custom base node of WALLET to NODE2
+        And I clear custom base node of wallet WALLET
+
+    Scenario: As a user I want to change password via command
+        Given I have wallet WALLET connected to all seed nodes
+        When I stop wallet WALLET
```
SWvheerden, line 18: See comment below

integration_tests/features/WalletCli.feature
```diff
@@ -0,0 +1,115 @@
+Feature: Wallet CLI
+
+    Scenario: As a user I want to change base node for a wallet via command
+        Given I have a base node NODE1 connected to all seed nodes
+        And I have a base node NODE2 connected to all seed nodes
+        And I have wallet WALLET connected to base node NODE1
+        Then I change base node of WALLET to NODE2
```
SWvheerden, line 7: We need to add a check here to see if the base_node was changed. 
For example:
`Then wallet WALLET is connected tp NODE2`

SWvheerden (commented):
Looks good so far. 
I made a commit about the STOP command. I understand why its made, but it reads weird when you read the cucumber test. We should make the issue command shutdown and restart the node again. 

A few of the tests also just pass and fail if the command succeeded or not, not seeing if the actual command did what it was supposed to do.

integration_tests/features/WalletCli.feature
Cifko: I will remove the "And I stop wallet SENDER". But the wallet will be stopped anyway. The stop is included in the runCommand function of the wallet process. All these tests are via command, so you have to stop the wallet to be able to run it with the appropriate command.

integration_tests/features/WalletCli.feature
Cifko: It's above in my view of the PR :-) I don't know why the order is reversed... anyway, the same applies.

integration_tests/features/WalletCli.feature
Cifko: The same as above.

integration_tests/features/WalletCli.feature
Cifko: I did had that originally. It would be nice to have but I had remove it this morning, because it somehow caused the test to be flaky (I was not able to reproduce that in my local environment). It failed set base node, and set custom base node. One test never connected and one test did connect but the node reported not connected. I can create a ticket to upgrade these test and a TODO in the feature file. But I don't think it's worth investing effort to fix this issue. Let me know what you think. We can discuss this tomorrow morning.

integration_tests/features/WalletCli.feature
```diff
@@ -0,0 +1,115 @@
+Feature: Wallet CLI
+
+    Scenario: As a user I want to change base node for a wallet via command
```
stringhandler: For clarity, I suggest adding "command line" instead of "command" in all instances

```suggestion
    Scenario: As a user I want to change base node for a wallet via command line
```

stringhandler (dismissed):

stringhandler (approved):

PR #3150, opened 2021-07-30
fix: set robust limits for a busy blockchain

hansieodendaal:
## Description
Updated limits for the base node and wallet that would be robust for a busy blockchain - this was simulated with two sizeable stress tests of 15,000 transactions each, using 3 sender wallets to 3 receiver wallets, with 5,000 to between each wallet pair. Both tests returned a 100% mined as confirmed success ratio. These tests also utilized the new persistent dedup message cache.

The following previous stress test issues were addressed with this PR:
- Mempool view was considerably different between base nodes.
- Communications were breaking down on some base nodes with this huge amount of messages.
- Lots of message flood banning were taking place.
- Excessive numbers of SAF messages were being sent for the same transaction.
- Base node not responding after sync request failed.
- Console wallet broadcast status was not being updated correctly with many (thousands) transactions present.

The wallet **edit:**_transaction_ event broadcast channel was made configurable due to an increase from 200 to 5,500 not being enough as shown by the results below. (The mobile wallet can use the default or its own config settings for memory preservation.)

| Event                        | Received | Missed |
|------------------------------|----------|--------|
| ReceivedTransaction          | 943      | 4060   |
| ReceivedFinalizedTransaction | 1958     | 3045   |
| TransactionBroadcast         | 1961     | 3042   |
| TransactionMinedUnconfirmed  | 2013     | 2990   |
| TransactionMined             | 2059     | 2944   |

| Totals             | Value |
|--------------------|-------|
| Transactions mined | 5003  |
| Events missed      | 16081 |
| Lagging reads      | 800   |
| Events not fired   | 15281 |

## Motivation and Context
See above

## How Has This Been Tested?
System-level stress tests

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/mod.rs
+++ b/base_layer/wallet/src/output_manager_service/mod.rs
@@ -106,7 +106,7 @@ where T: OutputManagerBackend + 'static
         );
 
         let (sender, receiver) = reply_channel::unbounded();
-        let (publisher, _) = broadcast::channel(200);
+        let (publisher, _) = broadcast::channel(self.config.event_channel_size);
 
         // Register handle before waiting for handles to be ready
         let oms_handle = OutputManagerHandle::new(sender, publisher.clone());
--- a/common/config/presets/tari_config_example.toml
+++ b/common/config/presets/tari_config_example.toml
@@ -43,16 +43,16 @@ network = "weatherwax"
 #liveness_allowlist_cidrs = ["127.0.0.1/32"]
 
 # The buffer size constants for the publish/subscribe connector channel, connecting comms messages to the domain layer:
-# - Buffer size for the base node (min value = 30, default value = 100).
-#buffer_size_base_node = 100
-# - Buffer size for the base node wallet (min value = 300, default value = 1000).
-#buffer_size_base_node_wallet = 1000
+# - Buffer size for the base node (min value = 30, default value = 1500).
+#buffer_size_base_node = 1500
+# - Buffer size for the console wallet (min value = 300, default value = 50000).
+#buffer_size_console_wallet = 50000
 # The rate limit constants for the publish/subscribe connector channel, i.e. maximum amount of inbound messages to
 # accept - any rate attemting to exceed this limit will be throttled.
-# - Rate limit for the base node (min value = 5, default value = 20).
-#buffer_rate_limit_base_node = 20
-# - Rate limit for the base node wallet (min value = 5, default value = 20).
-#buffer_rate_limit_base_node_wallet = 20
+# - Rate limit for the base node (min value = 5, default value = 1000).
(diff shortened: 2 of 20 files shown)


base_layer/wallet/src/output_manager_service/mod.rs
```diff
@@ -106,7 +106,7 @@ where T: OutputManagerBackend + 'static
         );
 
         let (sender, receiver) = reply_channel::unbounded();
-        let (publisher, _) = broadcast::channel(200);
+        let (publisher, _) = broadcast::channel(3500);
```
philipr-za: shouldn't this use the new config?

hansieodendaal: It was actually only the transaction event channel size that was made configurable due to stats in the description; do you suggest having more of these configurable?

philipr-za: Ja, if we are doing it then I think we should do it for both.

hansieodendaal: Cool, making all those broadcast channels sizes configurable in the config file.

common/config/presets/tari_config_example.toml
```diff
@@ -116,20 +117,33 @@ console_wallet_db_file = "wallet/console-wallet.dat"
...
+transaction_event_channel_size = 25000
+# This is the size of the event channel used to communicate base node events to the wallet. A busy console
+# wallet doing thousands of bulk payments or used for stress testing needs a fairly big size (>3000) (default = 250).
+base_node_event_channel_size = 3500
+# This is the size of the event channel used to communicate output manager events to the wallet. A busy console
+# wallet doing thousands of bulk payments or used for stress testing needs a fairly big size (>3000) (default = 250).
+output_manager_event_channel_size = 3500
+# This is the size of the event channel used to communicate base node update events to the wallet. A busy console
+# wallet doing thousands of bulk payments or used for stress testing needs a fairly big size (>300) (default = 50).
+base_node_update_publisher_channel_size = 500
```
philipr-za, line 146: Just for the sake of not bloating this file further. I really think this one can come out as it is not a function of how many transaction's there are its a function of how many times a client makes the call.

philipr-za (commented):
I don't think the channel sizes should be in the config file, it really feels like we are bloating the config file with stuff that 99.99% of end users will not care about and the config file is getting insanely bloated.

I say put them in the config object with default values and then its possible to set them via environment variables if we need to.

stringhandler (dismissed):
Seems like a little too much config. E.g. I wouldn't even know a good value for a channel size, who exactly would want to tweak this? 

That said, I like the other parts of the PR and those fields can be removed from config later if we need to.

stringhandler (dismissed):

stringhandler (dismissed):

PR #3153, opened 2021-08-02
test: improve Verify all coinbases in hybrid mining are accounted for cucumber test

SWvheerden:
## Description

Improve the `Verify all coinbases in hybrid mining are accounted for` to be less flaky.
Improved the readability of the test. 
Made the following improvements:
Ensure that the monero merge mining does not mine more blocks than it should.
Rather than check if they are at the same height, check tips.
Due to the network topology, added a wait timer to ensure that the nodes have time to reorg and give the wallets time to invalidate coinbases.

Added in another improvement to the coinbase value selection. 
## Checklist:


integration_tests/helpers/transactionBuilder.js
```diff
@@ -167,13 +167,15 @@ class TransactionBuilder {
       flags: 0,
       maturity: 0,
     };
-    let key = Math.floor(Math.random() * 500 + 1);
+    let key = Math.floor(Math.random() * 500000000000 + 1);
```
philipr-za, line 170: Just checking but there was a reason this value was so low? @mikethetike?

SWvheerden: We never implemented wrap-around in cucumber, with these low numbers we did not have to care about wrap-around. JS does not have unsigned integers. 
See PR: #3043  and PR #2646

integration_tests/package-lock.json
```diff
@@ -1,4838 +1,8 @@
 {
   "name": "integration_tests",
   "version": "1.0.0",
-  "lockfileVersion": 2,
```
stringhandler, line 4: let's not go through this delete/undelete again please

integration_tests/helpers/transactionBuilder.js
stringhandler: Let's see if it's a problem. I can't remember the exact reason, but it had to do with adding and subtracting and modding, which I don't know if we still do

stringhandler (approved):

PR #3154, opened 2021-08-03
ON_HOLD: fix: improve wallet's base node service connection, aka "console wallet being offline"

hansieodendaal:
## Description
This PR made:
- the wallet's base node service connection (including obtaining RPC sessions) to be less flaky. i.e. recoverable more quickly;
- RPC connection errors to be recognized more explicitly as warnings in the logs.

_**Note:** This PR does not fix the root cause (maybe call it the root symptom?) of the problem, which was identified as a valid peer connection not always providing valid RPC sessions._

The major change brought about by this PR is to disconnect the connected base node whenever obtaining an RPC session gave an error so that a new connection could be established before retrying to obtain an RPC session. A minor change was to only re-dial the base node peer if the connection status is "not connected" for a slight efficiency improvement.

The attached logs show an 8 hour stretch for 7x base node - console wallet pairs that ran on 2x Windows 10 computers. The longest recorded "offline" times occurred in a stretch of ~27 minutes on two console wallets between 03:01 and 03:27, with intermittent successful connections in between those times, after which those two wallets' "online" status recovered. Previously, these wallets were measured as staying "offline" for hours (~5) on end.

As a point of interest, the internet connection was reset at ~07:53 for both computers, which can be seen in the logs as well.

[node_01.log](https://github.com/tari-project/tari/files/6922094/node_01.log)    |    [node_02.log](https://github.com/tari-project/tari/files/6922095/node_02.log)    |    [node_03.log](https://github.com/tari-project/tari/files/6922097/node_03.log)    |    [node_04.log](https://github.com/tari-project/tari/files/6922099/node_04.log)    |    [node_05.log](https://github.com/tari-project/tari/files/6922100/node_05.log)    |    [node_06.log](https://github.com/tari-project/tari/files/6922101/node_06.log)    |    [node_07.log](https://github.com/tari-project/tari/files/6922103/node_07.log)

## Motivation and Context
A new RPC session is created for each metadata request to the base node, whose result is used to display blockchain stats in the console wallet. On some systems (computers) the `wallet::base_node_service::chain_metadata_monitor` was not very robust resulting in console wallets being "offline" for extended periods. This was traced to a symptom with RPC sessions, where creating a new session on a valid peer connection would result in an RPC timeout time out without seemingly the ability to recover. 

## How Has This Been Tested?
Extended system-level testing with 7 base node - console wallet pairs on two Windows 10 computers where these "offline" symptoms were paramount.

## Checklist:

### DIFF
--- a/base_layer/wallet/src/base_node_service/monitor.rs
+++ b/base_layer/wallet/src/base_node_service/monitor.rs
@@ -38,6 +38,7 @@ use tari_comms::{
     peer_manager::NodeId,
     protocol::rpc::RpcError,
     PeerConnection,
+    PeerConnectionError,
 };
 use tari_core::base_node::rpc::BaseNodeWalletRpcClient;
 use tari_shutdown::ShutdownSignal;
@@ -56,6 +57,7 @@ pub struct BaseNodeMonitor<T> {
     connectivity_manager: ConnectivityRequester,
     event_publisher: BaseNodeEventSender,
     shutdown_signal: ShutdownSignal,
+    connected_peer: Option<PeerConnection>,
 }
 
 impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
@@ -74,6 +76,7 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
             connectivity_manager,
             event_publisher,
             shutdown_signal,
+            connected_peer: None,
         }
     }
 
@@ -90,8 +93,11 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
                     );
                     break;
                 },
-                Err(e @ BaseNodeMonitorError::RpcFailed(_)) | Err(e @ BaseNodeMonitorError::DialFailed(_)) => {
-                    debug!(target: LOG_TARGET, "Connectivity failure to base node: {}", e,);
+                Err(e @ BaseNodeMonitorError::RpcFailed(_)) |
+                Err(e @ BaseNodeMonitorError::PeerConnectionFailed(_)) |
+                Err(e @ BaseNodeMonitorError::DialFailed(_)) |
+                Err(e @ BaseNodeMonitorError::NoPeerConnection) => {
+                    warn!(target: LOG_TARGET, "Connectivity failure to base node: {}", e,);
                     debug!(
                         target: LOG_TARGET,
                         "Setting as OFFLINE and retrying after {:.2?}", self.interval
@@ -108,7 +114,6 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
                         target: LOG_TARGET,
                         "Base node has changed. Connecting to new base node...",
                     );
-
                     self.set_connecting().await;
                     continue;
                 },
@@ -130,12 +135,12 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
 
     async fn process(&mut self) -> Result<(), BaseNodeMonitorError> {
         let peer = self.wait_for_peer_to_be_set().await?;
-        let connection = self.attempt_dial(peer.clone()).await?;
+        self.make_peer_connection(peer.clone()).await?;
         debug!(
             target: LOG_TARGET,
             "Base node connected. Establishing RPC connection...",
         );
-        let client = self.connect_client(connection).await?;
+        let client = self.get_rpc_session().await?;
         debug!(target: LOG_TARGET, "RPC established",);
         self.monitor_node(peer, client).await?;
         Ok(())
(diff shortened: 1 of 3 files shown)


base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -176,14 +181,34 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
...
+    // Return a new RPC session; close the connection if the RPC session fails
+    async fn get_rpc_session(&self) -> Result<BaseNodeWalletRpcClient, BaseNodeMonitorError> {
+        let connection = self
+            .connected_peer
+            .as_ref()
+            .ok_or(BaseNodeMonitorError::NoPeerConnection)?;
+        let rpc_client: BaseNodeWalletRpcClient = match connection.clone().connect_rpc().await {
+            Ok(client) => client,
+            Err(e) => {
+                connection.clone().disconnect().await?;
```
philipr-za, line 207: So I think this points towards an issue with the management of the connections in the comms layer. This action here is something that should be done by the ConnectionManager when errors occur on that connection.

If this disconnection fixes the issue then it shows that there was something wrong with the connection that ideally the ConnectionManager should have detected.

I am not a fan of papering over the cracks in the domain layer on issues like this because there is clearly an issue in the management of the connections but we are treating the symptom in various places in the domain layer and thus hiding information that would reveal the comms layer issue.

@sdbondi @mikethetike thoughts?

sdbondi: If a connection has been "rudely" dropped by the remote peer, it is expected that there will be a delay, typically of a few seconds (depends on keep alive / when sends occur), before the peer connection recognises that the peer is gone. This is the nature of connections: one has to send something and not receive a TCP ACK. And then there's TOR, which may delay this further. We accept that there is a large cost to pinging constantly to check every few milliseconds. So software must be robust to the delay. 

Absolutely agree, that upon trying to establish a rpc session, if the connection is no longer valid, the peer connection should be in a disconnected state _at some point in the near future_. I have no reason to suspect that this isn't the case for TCP, but we don't exactly have a test where we hard kill a process to test this. However, I suspect that this could be a more complex problem when routed over tor - if you think about it, the final hop of the tor node would have to recognise the dropped connection and pass that information back through the circuit - we don't know how well tor handles this, but we can be certain there will be delays.

So what options do we have, assuming this assessment is accurate:
1. comms should disconnect a peer connection whenever an RPC timeout occurs
     - counter: timeouts are not _always_ indicative of a dropped connection - so disconnecting is likely unexpected and/or undesirable behaviour. Many components use the connection.
2. comms should almost instantly know that a connection is dropped 
   - counter: the reality is that this has a cost that is likely bad in general, and very bad for mobile
3. Additional robustness to this edge-case, the wallet should handle a timeout by requesting a new peer connection - if it times out again, then request a new peer connection and so on. If there is a way to reduce the delay over tor, this solution should live in comms but the wallet's behaviour stays as described regardless. As you know, I am working on wrapping this behaviour up in a component at the moment and this includes changes to the monitor.rs file.

Thanks @hansieodendaal for bringing this to light. I think it is valuable to create a test that reproduces this.

sdbondi: I understand that the base node may not have been rudely disconnected - if you want this to go in that is perfectly ok with me.

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -176,14 +181,34 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
         }
     }
 
-    async fn attempt_dial(&mut self, peer: NodeId) -> Result<PeerConnection, BaseNodeMonitorError> {
-        let conn = self.connectivity_manager.dial_peer(peer).await?;
-        Ok(conn)
+    // Establish a peer connection; only redial if the peer changed or is not connected
+    async fn make_peer_connection(&mut self, peer_node_id: NodeId) -> Result<(), BaseNodeMonitorError> {
+        let make_new_connection = match self.connected_peer.clone() {
```
stringhandler, line 186: These clones are not necessary, did you try using the `ref` keyword?

stringhandler (commented):
I don't think there's enough of a benefit here to merge this pr. I appreciate the increased logging, but I think there's too much additional code that will eventually need to be added everywhere. If the connect/disconnect code was wrapped in a BaseNodeRpcClient I would say we could do it, but this code would currently have to be repeated everywhere

base_layer/wallet/src/base_node_service/monitor.rs
hansieodendaal: To be clear, in this case, comms connectivity reports that the connection has not been dropped by the peer and is still valid. 

Each call in the wallet monitor to process base chain metadata starts with a dial peer to obtain the peer connection, then requesting an RPC client session, then communicating with the RPC client:
``` rust
1.)  let peer = self.wait_for_peer_to_be_set().await?;
2.)  let conn = self.connectivity_manager.dial_peer(peer).await?;       -> PeerConnection
3.)  let client = conn.connect_rpc().await?;                            -> BaseNodeWalletRpcClient
4.)  self.monitor_node(peer, client).await?;
```
When the RPC is misbehaving as described by this PR, (2.) would always succeed en (3.) would always fail, for hours on end. This boils down to `PeerConnection::is_connected()` returning `true` when an RPC client session cannot be established.

I agree with your point 3. above ("the wallet should handle a timeout by requesting a new peer connection"), as it is exactly what this PR proposes. In this case, there is only one peer in the list, so requesting a new peer connection entails disconnecting the current connection, which will force a new connection next time round.

sdbondi: > When the RPC is misbehaving as described by this PR, (2.) would always succeed en (3.) would always fail, for hours on end. 

Reproducing this in a test and/or isolating the cause is useful. 

> so requesting a new peer connection entails disconnecting the current connection

There are many reasons a connect_rpc could return an error including the remote peer being too busy and other valid reasons not to create an rpc session where the response of disconnecting the peer connection, potentially used by many components, is unexpected and incorrect. Otherwise, if the connection thinks it's working for hours on end but isn't, the cause should be found and fixed.

PR #3156, opened 2021-08-03
test: improve cucumber reliability of Massive multiple reorg test

SWvheerden:
## Description

Found that the test: Massive multiple reorg failed because the block height of 5 had an equal difficulty of one of height 10. https://35459-136459099-gh.circle-artifacts.com/0/integration_tests/temp/reports/cucumber_report.html
Simplified the test somewhat. 
Changed it to ensure that we would always end up with a single chain. 

## Checklist:


philipr-za (dismissed):
LGTM

stringhandler (dismissed):
Nice

hansieodendaal (approved):
LGTM

I did not realize when I wrote this originally that a sync is not forced if two different chains have exactly the same difficulty, so only checking the same chain tip after an additional 10 blocks have been mined after the competing chains were joined makes a lot of sense. The _"Then all nodes are on the same chain tip"_ step is also a nice addition and a much better check to verify the reorgs.

stringhandler (approved):

PR #3157, opened 2021-08-03
fix: fix search_kernel command

SWvheerden:
## Description

The search-kernel command in the base_node UI currently returns either an error, none or the block containing said kernel. 
The UI interprets a None as a successful command, but the block could not be found due to it being a pruned block. This is incorrect, as the returned block is always a `historical block`. `Historical blocks` can have pruned outputs in them. So when a block is pruned, the base_node will return the block with the information it can fill in, plus all pruned outputs. 

When a kernel is not found in the chain, the search commands returns Ok(None) and the UI now interprets this correctly as the kernel is not found. 

## Checklist:


philipr-za (approved):
LGTM

StriderDM (approved):
+1

stringhandler (approved):

PR #3159, opened 2021-08-04
feat: wallet connectivity service

sdbondi:
## Description

Adds a service responsible for wallet connectivity. This service
is responsible for and abstracts any complexity in the management of
the base node connections and RPC session management.

This PR makes use of this service in the base node montoring service but
does not "plumb" the WalletConenctivityService into the protocols. This
is left as a TODO, but we should expect this to remove many lines of
code and greaty simplify these protocols by removing the budren of
connection management in the various wallet components.

A number of simplifications on the peer connection and substream code,
debatably reducing places that bugs could hide.

## Motivation and Context

Follow on from #3152, making use of the pool in the wallet base node monitoring. The rest of the protocols should follow. 

## How Has This Been Tested?

Extensive service unit tests
Existing tests pass

Washing machine test on local wallet, with the broadcast protocol modified to use the wallet connectivity 
Base node RPC sessions max out at 10 each (2 wallets connected to same base node) even when pushing through 426 transactions.
image

Manually tested wallet connectivity states (Connecting -> Online | Connecting -> Offline | Connecting -> Online -> Offline -> Connecting -> Online)

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4932,6 +4932,7 @@ dependencies = [
  "futures-test",
  "lazy_static 1.4.0",
  "rand 0.8.4",
+ "tari_shutdown",
  "tempfile",
  "tokio 0.2.25",
 ]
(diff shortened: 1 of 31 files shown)


stringhandler (approved):
Nice. I prefer 'create_base_node_...` to `obtain_base_node..` but otherwise is good

sdbondi:
Yeah I named it that after initially naming it create_...  just because the semantics of this call is not to necessarily create one but to somehow obtain one ("I don't care how/when just give me an RPC connection") - from a pool of existing connections.

PR #3162, opened 2021-08-05
fix: ban peer when Merkle roots mismatch

sdbondi:
## Description

Ban peer if block validation fails during block body sync

## Motivation and Context

When this error is encountered, the peer sent an invalid block body given the current header chain.  
`DEBUG Block sync failed: Block validation failed: Block validation error: Mismatched MMR roots`
The node should ban the peer. Previously, the node would encounter this, then retry the same peer,
rinse and repeat - preventing the node from syncing.

## How Has This Been Tested?

Not explicitly tested - fairly basic change.

## Checklist:

### DIFF
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -22,7 +22,10 @@
 
 use super::error::BlockSyncError;
 use crate::{
-    base_node::sync::{hooks::Hooks, rpc},
+    base_node::{
+        sync::{hooks::Hooks, rpc},
+        BlockSyncConfig,
+    },
     chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend, ChainBlock},
     proto::base_node::SyncBlocksRequest,
     tari_utilities::{hex::Hex, Hashable},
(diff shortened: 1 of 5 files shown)


philipr-za (approved):
LGTM

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -267,4 +280,21 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
...
+
+    async fn ban_peer<T: ToString>(&mut self, node_id: NodeId, reason: T) -> Result<(), BlockSyncError> {
+        let reason = reason.to_string();
+        if self.config.sync_peers.contains(&node_id) {
+            debug!(
+                target: LOG_TARGET,
+                "Not banning peer that is allowlisted for sync. Ban reason = {}", reason
+            );
+            return Ok(());
+        }
```
SWvheerden, line 292: Should this logic rather not live in the connectivity manager? Here we don't ban a peer because it is allow listed, but I don't think we have the same logic with propagation, there that peer will get banned.

SWvheerden (approved):
Looks good, this was much needed. Just one question

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
sdbondi: This is the force sync peer setting specific to sync, not the seed peers. By default it's empty. So seed peers can get banned, but if you explicitly say "I want to force a sync only from peer X" we never ban them.

stringhandler (approved):

PR #3163, opened 2021-08-05
fix: improve prune mode to remove panics

SWvheerden:
## Description

When requesting pruned data on base_node, the node would panic as it tried to provide pruned data. 
Removed this panic and changed requests for `TransactionOutputs` to `PrunedOutputs` so that callee functions can decide what to do with the data. 
Fixed ignored pruned unit test. 
Added in cucumber test to prove nodes cant sync from pruned nodes and neither panics. 

## Checklist:


stringhandler (dismissed):
Nice. In future we might need to think of a better name for the PrunedOutput struct. Perhaps `PrunableOutput`? idk

stringhandler (approved):

PR #3164, opened 2021-08-05
fix: update handling of SAF message propagation and deletion

philipr-za:
## Description

This PR adds two changes to the way SAF messages are handled to fix two subtle bugs spotted while developing cucumber tests.

The first issue was that when a Node propagates a SAF message it was storing to other nodes in its neighbourhood the broadcast strategy it was using only chose from currently connected base nodes. This meant that if the Node had an active connection to a Communication Client (wallet) it would not just directly send the SAF message to that client but to other base nodes in the network region. This meant that the wallet would only receive new SAF message when it actively requested them on connection even though it was directly connected to the node.

This PR adds a new broadcast strategy called `DirectOrClosestNodes` which will first check if the node has a direct active connection and if it does just send the SAF message directly to its destination.

The second issue was a subtle problem where when a node starts to send SAF messages to a destination it would remove the messages from the database based only on whether the outbound messages were put onto the outbound message pipeline. The problem occurs when the TCP connection to that peer is actually broken the sending of those messages would fail at the end of the pipeline but the SAF messages were already deleted from the database.

This PR changes the way SAF messages are deleted. When a client asks a node for SAF message it will also provide a timestamp of the most recent SAF message it has received. The Node will then send all SAF messages since that timestamp that it has for the node and will delete all SAF messages from before the specified Timestamp.  This serves as a form of Ack that the client has received the older messages at some point and they are no longer needed.

## How Has This Been Tested?
Unit tests have been updated to test this functionality.

## Checklist:

### DIFF
--- a/comms/dht/src/store_forward/service.rs
+++ b/comms/dht/src/store_forward/service.rs
@@ -43,7 +43,7 @@ use futures::{
     StreamExt,
 };
 use log::*;
-use std::{cmp, convert::TryFrom, sync::Arc, time::Duration};
+use std::{convert::TryFrom, sync::Arc, time::Duration};
 use tari_comms::{
     connectivity::{ConnectivityEvent, ConnectivityEventRx, ConnectivityRequester},
     peer_manager::{NodeId, PeerFeatures},
@@ -76,7 +76,7 @@ impl FetchStoredMessageQuery {
         }
     }
 
-    pub fn since(&mut self, since: DateTime<Utc>) -> &mut Self {
+    pub fn with_messages_since(&mut self, since: DateTime<Utc>) -> &mut Self {
         self.since = Some(since);
         self
     }
@@ -85,6 +85,10 @@ impl FetchStoredMessageQuery {
         self.response_type = response_type;
         self
     }
+
+    pub fn since(&self) -> Option<DateTime<Utc>> {
+        self.since
+    }
 }
 
 #[derive(Debug)]
@@ -92,6 +96,7 @@ pub enum StoreAndForwardRequest {
     FetchMessages(FetchStoredMessageQuery, oneshot::Sender<SafResult<Vec<StoredMessage>>>),
     InsertMessage(NewStoredMessage, oneshot::Sender<SafResult<bool>>),
     RemoveMessages(Vec<i32>),
+    RemoveMessagesOlderThan(DateTime<Utc>),
     SendStoreForwardRequestToPeer(Box<NodeId>),
     SendStoreForwardRequestNeighbours,
 }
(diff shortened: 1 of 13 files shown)


SWvheerden (dismissed):
Looks good, this is a good fix

comms/dht/src/store_forward/service.rs
```diff
@@ -76,7 +76,7 @@ impl FetchStoredMessageQuery {
         }
     }
 
-    pub fn since(&mut self, since: DateTime<Utc>) -> &mut Self {
+    pub fn with_since(&mut self, since: DateTime<Utc>) -> &mut Self {
```
sdbondi: nit: Maybe `with_messages_since` reads a bit better.

sdbondi (commented):
Looks good - I had assumed that we keep the broadcast strats the same and just make closest send to the dest peer if a connection is available rather than creating a new one broadcast strat. (Perhaps a miscommunication, I had thought maybe a DirectOrClosest would be needed on the `connectivity_manager.select_connections` call, not as a broadcast strat) 

Any reason/use-case that we need to keep the previous `Closest` behaviour around?
A quick look, I can't see a reason to have a new strat. If that is the case, I'd say rather fix `ClosestNodes` to send directly if connected.

comms/dht/src/store_forward/service.rs
philipr-za: Sure, did read a bit weird.

philipr-za:
> A quick look, I can't see a reason to have a new strat. If that is the case, I'd say rather fix `ClosestNodes` to send directly if connected.

Cool, I was a bit hesitant to assume there was no use for it so I left it in as is, in case, but if you think it can be cut I will take it out. I think I will leave it called DirectOrClosestNodes just so that the name makes the behaviour clear.

comms/dht/src/store_forward/service.rs
philipr-za: Fixed

philipr-za:
> Looks good - I had assumed that we keep the broadcast strats the same and just make closest send to the dest peer if a connection is available rather than creating a new one broadcast strat. (Perhaps a miscommunication, I had thought maybe a DirectOrClosest would be needed on the `connectivity_manager.select_connections` call, not as a broadcast strat)
> 
> Any reason/use-case that we need to keep the previous `Closest` behaviour around?
> A quick look, I can't see a reason to have a new strat. If that is the case, I'd say rather fix `ClosestNodes` to send directly if connected.

So I looked at this a bit and I noted that the Join process still kind of needs the ClosestNodes without the direct option. I also think that perhaps there might be other applications where sending to a neighbourhood and not doing the direct case might be useful.

The Join process might need an update in the future when we move to k-buckets but I am going to leave that considering for the future.

stringhandler (approved):

sdbondi (approved):

PR #3173, opened 2021-08-09
fix: enforce unique commitments in utxo set

sdbondi:
## Description

Adds a unique commitment db index for the UTXO set as well as unique
commitment check in the block validator.

## Motivation and Context

Prevents the a consensus disagreement between pruned and unpruned nodes by
reusing a commitment.

## How Has This Been Tested?

- New unit test that tests adding a UTXO with the same commitment, but with different maturity to avoid a hash collision.
After that UTXO is spent, the duplicate UTXO is allowed to be included in a block.
- Archival sync 
- Pruned sync
- Sent a number of transactions
- Mined some blocks
- `rewind-blockchain` command

## Checklist:

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -62,6 +62,7 @@ use crate::{
             LMDB_DB_ORPHAN_PARENT_MAP_INDEX,
             LMDB_DB_TXOS_HASH_TO_INDEX,
             LMDB_DB_UTXOS,
+            LMDB_DB_UTXO_COMMITMENT_INDEX,
             LMDB_DB_UTXO_MMR_SIZE_INDEX,
         },
         BlockchainBackend,
@@ -96,13 +97,13 @@ type DatabaseRef = Arc<Database<'static>>;
 
 pub const LOG_TARGET: &str = "c::cs::lmdb_db::lmdb_db";
 
-struct OutputKey {
-    header_hash: HashOutput,
+struct OutputKey<'a> {
+    header_hash: &'a [u8],
     mmr_position: u32,
 }
 
-impl OutputKey {
-    pub fn new(header_hash: HashOutput, mmr_position: u32) -> OutputKey {
+impl<'a> OutputKey<'a> {
+    pub fn new(header_hash: &'a [u8], mmr_position: u32) -> OutputKey {
         OutputKey {
             header_hash,
             mmr_position,
@@ -110,7 +111,7 @@ impl OutputKey {
     }
 
     pub fn get_key(&self) -> String {
-        format!("{}-{:010}", self.header_hash.to_hex(), self.mmr_position)
+        format!("{}-{:010}", to_hex(&self.header_hash), self.mmr_position)
     }
 }
 
@@ -131,6 +132,7 @@ pub struct LMDBDatabase {
     kernel_excess_sig_index: DatabaseRef,
     kernel_mmr_size_index: DatabaseRef,
     output_mmr_size_index: DatabaseRef,
+    utxo_commitment_index: DatabaseRef,
     orphans_db: DatabaseRef,
     monero_seed_height_db: DatabaseRef,
     orphan_header_accumulated_data_db: DatabaseRef,
@@ -157,6 +159,7 @@ impl LMDBDatabase {
             kernel_excess_sig_index: get_database(&store, LMDB_DB_KERNEL_EXCESS_SIG_INDEX)?,
             kernel_mmr_size_index: get_database(&store, LMDB_DB_KERNEL_MMR_SIZE_INDEX)?,
             output_mmr_size_index: get_database(&store, LMDB_DB_UTXO_MMR_SIZE_INDEX)?,
+            utxo_commitment_index: get_database(&store, LMDB_DB_UTXO_COMMITMENT_INDEX)?,
             orphans_db: get_database(&store, LMDB_DB_ORPHANS)?,
             orphan_header_accumulated_data_db: get_database(&store, LMDB_DB_ORPHAN_HEADER_ACCUMULATED_DATA)?,
             monero_seed_height_db: get_database(&store, LMDB_DB_MONERO_SEED_HEIGHT)?,
@@ -225,13 +228,6 @@ impl LMDBDatabase {
                         witness_hash,
                         mmr_position,
                     )?;
-                },
-                InsertInput {
-                    header_hash,
-                    input,
-                    mmr_position,
-                } => {
-                    self.insert_input(&write_txn, header_hash, *input, mmr_position)?;
                 },
                 DeleteHeader(height) => {
                     self.delete_header(&write_txn, height)?;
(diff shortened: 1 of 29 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -685,13 +695,17 @@ impl LMDBDatabase {
...
+    ) -> Result<(), ChainStorageError> {
+        let hash_hex = block_hash.to_hex();
         debug!(target: LOG_TARGET, "Deleting block `{}`", hash_hex);
         debug!(target: LOG_TARGET, "Deleting UTXOs...");
-        let height = self
-            .fetch_height_from_hash(&write_txn, &hash)
-            .or_not_found("Block", "hash", hash.to_hex())?;
+        let height =
+            self.fetch_height_from_hash(&write_txn, &block_hash)
+                .or_not_found("Block", "hash", block_hash.to_hex())?;
```
```suggestion
                .or_not_found("Block", "hash", hash_hex)?;
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -847,17 +875,13 @@ impl LMDBDatabase {
...
-                to_hex(&output.commitment.as_bytes())
-            );
+            trace!(target: LOG_TARGET, "Inserting output `{}`", output.commitment.to_hex());
             self.insert_output(
                 txn,
                 block_hash.clone(),
                 header.height,
                 output,
-                (witness_mmr.get_leaf_count()? - 1) as u32,
+                (output_mmr.get_leaf_count() - 1) as u32,
```
SWvheerden: Just ensure these will always return the same count. I am not sure if output.mmr will return a smaller number if it actually gets pruned.  And witness_mmr.size() == output_mmr.size()

base_layer/core/src/validation/transaction_validators.rs
```diff
@@ -129,17 +129,15 @@ fn verify_not_stxos<B: BlockchainBackend>(tx: &Transaction, db: &B) -> Result<()
                 );
```
SWvheerden: Github does not let me comment on the above lines, but should we not replace this piece of code above to be the same as the block validation. The block validation now uses `db.is_unspent_output(&input.commitment)? {` while this still uses the deletedbitmap.

base_layer/core/tests/helpers/block_builders.rs
```diff
@@ -114,7 +114,7 @@ pub fn _create_act_gen_block() {
     let factories = CryptoFactories::default();
     let mut header = BlockHeader::new(consensus_manager.consensus_constants(0).blockchain_version());
     let value = consensus_manager.emission_schedule().block_reward(0);
-    let (mut utxo, key, _) = create_utxo(value, &factories, None, &TariScript::default());
+    let (mut utxo, key, _) = create_utxo(value, &factories, Default::default(), &TariScript::default());
```
SWvheerden:
```suggestion
    let (utxo, key, _) = create_utxo(value, &factories, OutputFeatures::create_coinbase(1), &TariScript::default());
```

base_layer/core/tests/helpers/block_builders.rs
```diff
@@ -114,7 +114,7 @@ pub fn _create_act_gen_block() {
...
     utxo.features = OutputFeatures::create_coinbase(1);
```
SWvheerden:
```suggestion

```

base_layer/core/tests/helpers/block_builders.rs
SWvheerden:
```suggestion
    let (mut utxo, key, _) = create_utxo(value, &factories, Default::default(), &script!(Nop));
```
We should probably do this as the default will be spendable.

SWvheerden (commented):
Looks good, a few small nits, and a question.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: Yeah they do, not sure why I used `witness_mmr` in the first place.

base_layer/core/src/validation/transaction_validators.rs
sdbondi: I did start to change this, but changed it back. tbh I wasn't too sure. 
Thinking about it now, I think we need the old check to test that the output was used as is (i.e. with the same maturity etc, by output hash) and is not spent AND check that the commitment is not already in the current UTXO set

base_layer/core/tests/helpers/block_builders.rs
sdbondi: These are working tests, so are you suggesting this as a general best practise?

base_layer/core/src/validation/transaction_validators.rs
sdbondi: On seconds thoughts, I think just using is_unspent_output is fine.

These are the spends we want to prevent:

```
Spending to same commitment in current UTXO set
U1 -> U2
U3 -> U2
-------------
Double spending output, even if has different output features (^)
U1 -> U2
U1^ -> U3
-------------
Spent to output with same commitment
U1 -> U1^
```

but not sure about these

```
(A)
U1 -> U2
U2 -> U1
-------------
(B)
U1 -> U2
U2 -> U1^
```
Currently (A) will fail validation because the hash exists in the MMR. It would also fail on the db level because we currently have a unique index on all outputs (`txos_hash_to_index_db`) but this is another possible point of dicussion.
(B) will work (but only as long as you change output features each time)

EDIT: we also need to disallow - which is why we need to check the output exactly matches
```
U1 -> U2
U2^ -> U3
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
sdbondi: `output_mmr.len()` will give you the pruned length (artificially created by taking `leaf_count - cardinality(deleted)`)
but if we ever decide to prune the MMR we could run into trouble. So I think I should revert this change.

base_layer/core/src/validation/transaction_validators.rs
sdbondi: Updated - The wallet uses the transaction validator to tell it (via rpc) if a transaction was mined or not, so we still need to query the output hash (fetch_output), but only if the input is not in the UTXO set - so will be a little more efficient because we don't need to check in the unspent case + we dont need the deleted bitmap. Just fixing tests.

SWvheerden: awesome, I hate the fact we still have this duplicated code. But we can't have the transactions and block validators running different rules. Mempool uses transactions validation to decide on the transactions to include on the block, while base nodes use the block validation to decide on blocks. If a tx passes tx_validation and fails block validation, its a problem.

sdbondi: Yeah not sure we're going to get around that for the moment - mostly because the mempool uses the transaction validation to tell the wallet things it needs to know (e.g was a transaction mined) and so needs to do extra things to determine this that is not necessary if just checking the "spendability" of a UTXO (i.e. what block validator cares about) - so for now I'll keep the duplicate code. They do actually have subtly different roles - so I don't think this is too bad if we make sure our tests check that the transaction validator never lets through anything that block validator will reject.

SWvheerden (dismissed):
LGTM

stringhandler (dismissed):

stringhandler (approved):

PR #3179, opened 2021-08-10
test: fix flaky unit test sending_transaction_and_confirmation

SWvheerden:
## Description

The unit test had 2 edge cases:

1. When the transaction had no change, the master_key index should be 0 and not 1. 
2. Inside of the builder we had an edge case of one check being `>=` while the next check being `>` which resulted in the tx builder thinking it did not have enough funds

## Checklist:


sdbondi (approved):
Good catch

stringhandler (approved):

PR #3180, opened 2021-08-10
test: fix wallet CLI cucumber tests (child process exited with code 107)

hansieodendaal:
## Description
1. When using the wallet CLI mode in cucumber integration tests, the wallet is not always in the correct state to send transactions, depending on the last known metadata status from the previously connected base node. This PR introduces exception handling for wallet CLI commands so that it will automatically retry to execute the command if it fails. _A follow-up PR should let the wallet wait for a base node connection before it executes certain commands._
2. Added a new test step to explicitly test transaction statuses in the wallet.
3. Fixed the UTXO scanning service so it would not go into an infinite loop trying to scan UTXOs if the blockchain does not have any UTXOs yet.
4. Fixed an erroneous error return in the output manager service.

## Motivation and Context
Cucumber tests involving wallet CLI commands were failing.

## How Has This Been Tested?
Wallet CLI cucumber tests.

## Checklist:

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/database.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/database.rs
@@ -700,10 +700,7 @@ where T: OutputManagerBackend + 'static
         let db_clone = self.db.clone();
         tokio::task::spawn_blocking(move || {
             match db_clone.write(WriteOperation::Remove(DbKey::AnyOutputByCommitment(commitment.clone()))) {
-                Ok(None) => log_error(
-                    DbKey::AnyOutputByCommitment(commitment.clone()),
-                    OutputManagerStorageError::ValueNotFound,
-                ),
+                Ok(None) => Ok(()),
                 Ok(Some(DbValue::AnyOutput(_))) => Ok(()),
                 Ok(Some(other)) => unexpected_result(DbKey::AnyOutputByCommitment(commitment.clone()), other),
                 Err(e) => log_error(DbKey::AnyOutputByCommitment(commitment), e),
--- a/base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
+++ b/base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
@@ -307,6 +307,11 @@ where TBackend: WalletBackend + 'static
             }
 
             let num_scanned = self.scan_utxos(&mut client, start_index, tip_header).await?;
+            if num_scanned == 0 {
+                return Err(UtxoScannerError::UtxoScanningError(
+                    "Peer returned 0 UTXOs to scan".to_string(),
+                ));
+            }
             debug!(
                 target: LOG_TARGET,
                 "Scanning round completed UTXO #{} in {:.2?} ({} scanned)",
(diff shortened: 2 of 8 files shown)


base_layer/wallet/src/output_manager_service/storage/database.rs
```diff
@@ -700,10 +700,7 @@ where T: OutputManagerBackend + 'static
         let db_clone = self.db.clone();
         tokio::task::spawn_blocking(move || {
             match db_clone.write(WriteOperation::Remove(DbKey::AnyOutputByCommitment(commitment.clone()))) {
-                Ok(None) => log_error(
-                    DbKey::AnyOutputByCommitment(commitment.clone()),
-                    OutputManagerStorageError::ValueNotFound,
-                ),
+                Ok(None) => Ok(()),
```
SWvheerden, line 703: Why remove the logging here?
If its confusing or provides false errors, perhaps reword the log.

base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
```diff
@@ -584,14 +589,16 @@ where TBackend: WalletBackend + 'static
...
                         warn!(
                             target: LOG_TARGET,
-                            "Failed to scan UTXO's from base node {}: {}", peer, e
+                            "Failed to scan UTXO's from base node {} ({}), waiting {}s", peer, e, wait_seconds
                         );
                         self.publish_event(UtxoScannerEvent::ScanningRoundFailed {
                             num_retries: self.num_retries,
                             retry_limit: self.retry_limit,
                         });
+                        time::delay_for(Duration::from_secs(wait_seconds)).await;
```
SWvheerden: We should not put a wait here. Rather put the wait where this run function is called from or on in the none part. 

Basically what's going to happen here is that say the peer has an error the wallet will wait 30 seconds before trying the next peer. In UTXO recovery there is more than 1 peer so we want to hop to the next peer as fast as possible. 

I strongly prefer putting the wait in the where this task is called from. This to me is a dirty hack to fix a problem we are effectively only getting on cucumber, and only in the logs, where the wallet cant scan yet because the base_node it is connected to has not finished booting up.

SWvheerden (commented):
Looks good so far, I question and a small concern.

base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
hansieodendaal: Yip, this delay is not good, removed it. The proper interval is already handled here in `utxo_scanning.rs`:
``` rust
        loop {
            futures::select! {
                _ = work_interval.select_next_some() => {
```

base_layer/wallet/src/output_manager_service/storage/database.rs
hansieodendaal: Logging was removed here because it is not an error. It is a valid condition that happens a lot and is not worth logging. @philipr-za, I know we discussed this; your opinion, please?

philipr-za: This is expected to happen a lot. Not finding the the output is not an error and I do not think it is notable in the logs. I would argue that if code is using this and not finding the output is a concern to that code that it should log the problem up at that layer.

SWvheerden (dismissed):
LGTM

SWvheerden (approved):

stringhandler (approved):

PR #3190, opened 2021-08-12
test: add wallet reorg cucumber tests

hansieodendaal:
## Description
Added wallet reorg cucumber tests for coinbase and normal
transactions. These will be failing until wallets can
handle reorgs properly - failing steps have been commented
out.

## Motivation and Context
See description

## How Has This Been Tested?
Cucumber integration tests.

## Checklist:

### DIFF
--- a/integration_tests/features/support/steps.js
+++ b/integration_tests/features/support/steps.js
@@ -11,7 +11,7 @@ const {
   getTransactionOutputHash,
   sleep,
   consoleLogBalance,
-  consoleLogCoinbaseDetails,
+  consoleLogTransactionDetails,
   withTimeout,
 } = require("../../helpers/util");
 const { ConnectivityStatus, PaymentType } = require("../../helpers/types");
@@ -1901,8 +1901,6 @@ Then(
       const transactionPending = await walletClient.isTransactionAtLeastPending(
         txIds[i]
       );
-      // let txnDetails = await wallet.getClient().getTransactionDetails(txIds[i]);
-      // consoleLogTransactionDetails(txnDetails, txIds[i]);
       expect(transactionPending).to.equal(true);
     }
   }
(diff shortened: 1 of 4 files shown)


integration_tests/features/support/steps.js
```diff
@@ -2569,28 +2543,34 @@ Then(
...
+  async function (transaction_type, walletName) {
     const wallet = this.getWallet(walletName);
     const walletClient = await wallet.connectClient();
-    console.log("\nListing all coinbase transactions: ", walletName);
-    const transactions = await walletClient.getAllCoinbaseTransactions();
+    var transactions;
+    var type;
+    if (transaction_type === "NORMAL") {
+      transactions = await walletClient.getAllNormalTransactions();
+      type = "NORMAL";
```
philipr-za, line 2562: Nit: why declare a new variable for this when `transaction_type` is === to the "NORMAL"?

integration_tests/features/support/steps.js
```diff
@@ -2569,28 +2543,34 @@ Then(
...
+    } else {
+      transactions = await walletClient.getAllCoinbaseTransactions();
+      type = "COINBASE";
```
philipr-za, line 2565: same here

philipr-za (approved):

integration_tests/features/support/steps.js
hansieodendaal: This makes the step resilient. If the scenario states anything else than "NORMAL", it will be treated as "COINBASE". Variable `transaction_type` can be anything the user writes, but `type` is controlled and used in the logical branches.

integration_tests/features/support/steps.js
hansieodendaal: As per the above comment

PR #3191, opened 2021-08-12
alternate approach to transaction validation

stringhandler:


base_layer/wallet/migrations/2021-07-05-13201407_metadata_signature/up.sql
```diff
@@ -4,15 +4,15 @@
 PRAGMA foreign_keys=off;
 DROP TABLE outputs;
 CREATE TABLE outputs (
-    id INTEGER NOT NULL PRIMARY KEY,
-    commitment BLOB NOT NULL,
+    id INTEGER NOT NULL PRIMARY KEY,  --auto inc,
+    commitment BLOB NULL,
```
philipr-za: Why remove this constraint?

base_layer/wallet/migrations/2021-07-05-13201407_metadata_signature/up.sql
```diff
@@ -4,15 +4,15 @@
...
     spending_key BLOB NOT NULL,
-    value INTEGER NOT NULL,
+    value BIGINT NOT NULL,
     flags INTEGER NOT NULL,
-    maturity INTEGER NOT NULL,
+    maturity bigint NOT NULL,
```
philipr-za:
```suggestion
    maturity BIGINT NOT NULL,
```

base_layer/wallet/migrations/2021-07-05-13201407_metadata_signature/up.sql
```diff
@@ -4,15 +4,15 @@
...
     status INTEGER NOT NULL,
-    tx_id INTEGER NULL,
-    hash BLOB NOT NULL,
+    tx_id bigint NULL,
```
philipr-za:
```suggestion
    tx_id BIGINT NULL,
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -162,10 +162,25 @@ where TBackend: OutputManagerBackend + 'static
         pin_mut!(request_stream);
 
         let mut shutdown = self.resources.shutdown_signal.clone();
-
+        // Should probably be block time or at least configurable
+        // Start a bit later so that the wallet can start up
+        let mut txo_validation_interval = interval_at(
```
philipr-za, line 167: Should def be configurable. 

My next question is does this process have to happen on the regular? It is going to thrash the battery and surely doesn't have to happen at such a high rate for output validation?

If we are going with this automagic approach we should at least allow the client to control the frequency of these polling events like we do with the power_mode in the transaction service to be able to lower the polling interval and save battery.

base_layer/wallet/src/output_manager_service/storage/database.rs
```diff
@@ -50,12 +50,36 @@ const LOG_TARGET: &str = "wallet::output_manager_service::database";
...
+        mark_deleted_in_block: Vec<u8>,
+    ) -> Result<(), OutputManagerStorageError>;
+
+    fn mark_output_as_unspent(&self, hash: Vec<u8>) -> Result<(), OutputManagerStorageError>;
+
     /// This method is called when a pending transaction is to be confirmed. It must move the `outputs_to_be_spent` and
     /// `outputs_to_be_received` from a `PendingTransactionOutputs` record into the `unspent_outputs` and
     /// `spent_outputs` collections.
-    fn confirm_transaction(&self, tx_id: TxId) -> Result<(), OutputManagerStorageError>;
+    fn confirm_transaction_encumberance(&self, tx_id: TxId) -> Result<(), OutputManagerStorageError>;
```
philipr-za, line 82: Nit: I don't really agree with the change to this name. Confirming the transaction actually removes encumberance.

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -305,6 +339,7 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
                     new_output.commit(&(*conn))?
                 },
                 DbKeyValuePair::PendingTransactionOutputs(tx_id, p) => {
+                    // TODO: Used only by coinbases, so should be renamed
```
philipr-za, line 356: This is used to keep track of spendable outputs that have been encumbered against a specific tx_id. Do you plan to do that in a different way because the logic to encumber outputs has not been changed in this PR thus far? If it is changing then I why would Coinbases be different?

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -1531,6 +1692,29 @@ impl Encryptable<Aes256Gcm> for KeyManagerStateSql {
...
+        let encrypted_master_key = encrypt_bytes_integral_nonce(&cipher, self.master_key.clone())?;
+        let encrypted_branch_seed =
+            encrypt_bytes_integral_nonce(&cipher, self.branch_seed.clone().as_bytes().to_vec())?;
+        self.master_key = encrypted_master_key;
+        self.branch_seed = encrypted_branch_seed.to_hex();
+        Ok(())
+    }
+
+    fn decrypt(&mut self, _cipher: &Aes256Gcm) -> Result<(), Error> {
+        unimplemented!("Not supported")
```
philipr-za, line 1721: So I agree we don't need to access the master key or branch seed yet while we only support a single wallet but if we ever have a secondary branch seed for a child wallet we will have to put this back.

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+            "Checking last mined TXO to see if the base node has re-orged"
+        );
+
+        loop {
+            if let Some(last_spent_output) = self
+                .db
+                .get_last_spent_output()
+                .await
+                .for_protocol(self.operation_id)
+                .unwrap()
```
philipr-za: unwrap

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+            {
+                let mined_height = last_spent_output.marked_deleted_at_height.unwrap(); // TODO: fix unwrap
+                let mined_in_block_hash = last_spent_output.marked_deleted_in_block.clone().unwrap(); // TODO: fix unwrap.
+                let block_at_height = self
+                    .get_base_node_block_at_height(mined_height, client)
+                    .await
+                    .for_protocol(self.operation_id)?;
+                if block_at_height.is_none() || block_at_height.unwrap() != mined_in_block_hash {
```
philipr-za:
```suggestion
                if block_at_height != Some(mined_in_block_hash) {
```

base_layer/wallet/src/transaction_service/protocols/transaction_broadcast_protocol.rs
```diff
@@ -497,56 +482,70 @@ where TBackend: TransactionBackend + 'static
...
-                    response.confirmations
-                );
-                return Ok(true);
-            }
             info!(
                 target: LOG_TARGET,
-                "Transaction (TxId: {}) detected as mined but UNCONFIRMED with {} confirmations",
-                self.tx_id,
-                response.confirmations
+                "Broadcast transaction detected as mined, will be managed by transaction validation protoocol"
```
philipr-za, line 487:
```suggestion
                "Broadcast transaction detected as mined, will be managed by transaction validation protocol"
```

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol_v2.rs
```diff
@@ -0,0 +1,417 @@
...
+        db: TransactionDatabase<TTransactionBackend>,
+        base_node_pk: CommsPublicKey,
+        connectivity_requester: ConnectivityRequester,
+        config: TransactionServiceConfig,
+        event_publisher: TransactionEventSender,
+    ) -> Self {
+        Self {
+            operation_id: 122, // Get a real tx id
+            db,
+            batch_size: 10,
```
philipr-za, line 79: Should come from config

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol_v2.rs
```diff
@@ -0,0 +1,417 @@
...
+            target: LOG_TARGET,
+            "Checking last mined transactions to see if the base node has re-orged"
+        );
+        loop {
+            if let Some(last_mined_transaction) = self
+                .db
+                .get_last_mined_transaction()
+                .await
+                .for_protocol(self.operation_id)
+                .unwrap()
```
philipr-za: unwrap

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol_v2.rs
```diff
@@ -0,0 +1,417 @@
...
+            {
+                let mined_height = last_mined_transaction.mined_height.unwrap(); // TODO: fix unwrap
+                let mined_in_block_hash = last_mined_transaction.mined_in_block.clone().unwrap(); // TODO: fix unwrap.
+                let block_at_height = self
+                    .get_base_node_block_at_height(mined_height, client)
+                    .await
+                    .for_protocol(self.operation_id)?;
+                if block_at_height.is_none() || block_at_height.unwrap() != mined_in_block_hash {
```
philipr-za:
```suggestion
                if block_at_height != Some(mined_in_block_hash) {
```

base_layer/core/src/base_node/rpc/service.rs
```diff
@@ -308,6 +325,112 @@ impl<B: BlockchainBackend + 'static> BaseNodeWalletService for BaseNodeWalletRpc
...
+
+        let metadata = self
+            .db
+            .get_chain_metadata()
+            .await
+            .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
+
+        let deleted_bitmap = self
+            .db
+            .fetch_complete_deleted_bitmap_at(metadata.best_block().clone())
```
SWvheerden, line 411: This is a note, this can be an expensive call to run if the agent requestion this call keeps asking for a header way back in the past

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -350,24 +365,53 @@ where TBackend: OutputManagerBackend + 'static
...
+                //         Err(OutputManagerProtocolError { id, error }) => {
+                //             warn!(
+                //                 target: LOG_TARGET,
+                //                 "Error completing UTXO Validation Protocol (Id: {}): {:?}", id, error
+                //             );
+                //         },
+                //     }
+                // });
+
+                let utxo_validation = TxoValidationTaskV2::new(
```
SWvheerden, line 406: nit: I would keep this as `TxoValidationTask`, and only start using a `TxoValidationTaskV2` when we have two at the same time with something akin to a hard fork where we need to use 2 versions at the same time. If we replace one with the other, I would just replace it keep the name of the replaced version, in this case `TxoValidationTask`

base_layer/wallet/src/transaction_service/storage/database.rs
```diff
@@ -363,6 +431,14 @@ where T: TransactionBackend + 'static
         Ok(*t)
     }
 
+    pub async fn get_last_mined_transaction(&self) -> Result<Option<CompletedTransaction>, TransactionStorageError> {
+        self.db.fetch_last_mined_transaction()
+    }
+
+    pub async fn fetch_unmined_transactions(&self) -> Result<Vec<CompletedTransaction>, TransactionStorageError> {
```
philipr-za, line 433: perhaps this should be called something like `fetch_unconfirmed_transactions` because it will return mined_unconfirmed which has mined in the name

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol_v2.rs
```diff
@@ -0,0 +1,417 @@
...
+                false,
+                mined_height,
+                mined_in_block.clone(),
+                num_confirmations,
+                num_confirmations >= self.config.num_confirmations_required,
+            )
+            .await
+            .for_protocol(self.operation_id)?;
+
+        if num_confirmations >= self.config.num_confirmations_required {
```
philipr-za, line 382: This feels like a weird Event to use for a lost coinbase that is actually not mined, perhaps a new kind of event for this case?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -660,10 +672,10 @@ where
                 .start_transaction_validation_protocol(retry_strategy, transaction_validation_join_handles)
                 .await
                 .map(TransactionServiceResponse::ValidationStarted),
-            TransactionServiceRequest::SetCompletedTransactionValidity(tx_id, validity) => self
-                .set_completed_transaction_validity(tx_id, validity)
-                .await
-                .map(|_| TransactionServiceResponse::CompletedTransactionValidityChanged),
+            /* TransactionServiceRequest::SetCompletedTransactionValidity(tx_id, validity) => self
```
philipr-za, line 653: This is used by the output manager to be able to set the validity of Import Transaction's if the output becomes invalid. This is needed because the faux transactions we create to represent imported UTXO's are not checked for validity but should be marked as invalid if the output itself is invalidated.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -276,9 +275,22 @@ where
             JoinHandle<Result<u64, TransactionServiceProtocolError>>,
         > = FuturesUnordered::new();
 
+        // Should probably be block time or at least configurable
```
philipr-za, line 276: Should take into account the power mode timeout that is currently set. These timeouts allow the wallet client to set a much lower interval to reduce polling frequency and save power.

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+        Ok(())
+    }
+
+    async fn update_received_outputs(
+        &self,
+        wallet_client: &mut BaseNodeWalletRpcClient,
+    ) -> Result<(), OutputManagerProtocolError> {
+        let unmined_outputs = self
+            .db
+            .fetch_unmined_received_outputs()
```
philipr-za, line 179: This doesn't seem correct to me. This function gets all outputs that have `mined_in_block` as null but this includes outputs that are being spent not just received?

Below if these outputs are mined then `update_output_as_mined` is called on them that gives them the Unspent status which seems wrong because they were outgoing funds. 

I see that the next stage of `update_spent_outputs` will correct this but  still the database state at this stage is incorrect.

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -415,7 +459,88 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
+            ))
+            .execute(&(*conn))
+            .num_rows_affected_or_not_found(1)?;
+
+        Ok(())
+    }
+
+    fn mark_output_as_unspent(&self, hash: Vec<u8>) -> Result<(), OutputManagerStorageError> {
+        let conn = self.database_connection.acquire_lock();
+        // Only allow updating of non-deleted utxos
```
philipr-za, line 537: not sure what this comment means?

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -954,6 +1077,13 @@ struct OutputSql {
     metadata_signature_nonce: Vec<u8>,
     metadata_signature_u_key: Vec<u8>,
     metadata_signature_v_key: Vec<u8>,
+    mined_height: Option<i64>,
+    mined_in_block: Option<Vec<u8>>,
+    mined_mmr_position: Option<i64>,
+    marked_deleted_at_height: Option<i64>,
+    marked_deleted_in_block: Option<Vec<u8>>,
+    received_in_tx_id: Option<i64>,
```
SWvheerden, line 1089: should this not be an i64?
You should always receive an output before you you have it, so you should receive it in some transaction. 
Or is this for the faucets and imported? If that's the case cant we add a special recive_id if they don't have a matching id?

base_layer/wallet/src/transaction_service/service.rs
SWvheerden, line 276: yeah, a time faster than the block time, does not make sense

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1481,30 +1493,39 @@ where
...
+        _retry_strategy: ValidationRetryStrategy,
         join_handles: &mut FuturesUnordered<JoinHandle<Result<u64, TransactionServiceProtocolError>>>,
     ) -> Result<u64, TransactionServiceError> {
         if self.base_node_public_key.is_none() {
             return Err(TransactionServiceError::NoBaseNodeKeysProvided);
         }
         trace!(target: LOG_TARGET, "Starting transaction validation protocols");
         let id = OsRng.next_u64();
-        let timeout = match self.power_mode {
+        let _timeout = match self.power_mode {
```
SWvheerden, line 1539: remove?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1901,39 +1929,43 @@ where
...
+        _join_handles: &mut FuturesUnordered<JoinHandle<Result<u64, TransactionServiceProtocolError>>>,
     ) -> Result<(), TransactionServiceError> {
         let completed_tx = self.db.get_completed_transaction(tx_id).await?;
 
         if completed_tx.status != TransactionStatus::Coinbase || completed_tx.coinbase_block_height.is_none() {
             return Err(TransactionServiceError::InvalidCompletedTransaction);
         }
 
-        let block_height = if let Some(bh) = completed_tx.coinbase_block_height {
+        let _block_height = if let Some(bh) = completed_tx.coinbase_block_height {
```
SWvheerden, line 1975: remove?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1901,39 +1929,43 @@ where
...
             bh
         } else {
             0
         };
 
-        let timeout = match self.power_mode {
+        let _timeout = match self.power_mode {
```
SWvheerden, line 1981: remove?

comms/src/protocol/rpc/client.rs
```diff
@@ -226,9 +226,9 @@ impl RpcClientConfig {
 impl Default for RpcClientConfig {
     fn default() -> Self {
         Self {
-            deadline: Some(Duration::from_secs(30)),
+            deadline: Some(Duration::from_secs(38)),
```
SWvheerden: reason for the 8 sec increase?

comms/src/protocol/rpc/client.rs
```diff
@@ -226,9 +226,9 @@ impl RpcClientConfig {
...
             deadline_grace_period: Duration::from_secs(10),
-            handshake_timeout: Duration::from_secs(30),
+            handshake_timeout: Duration::from_secs(33),
```
SWvheerden: and the 3 here?

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+                    chain_must_include_header: last_mined_header_hash.clone(),
+                    mmr_positions: batch.iter().filter_map(|ub| ub.mined_mmr_position).collect(),
+                })
+                .await
+                .for_protocol(self.operation_id)?;
+
+            for output in batch {
+                if deleted_bitmap_response
+                    .deleted_positions
+                    .contains(&output.mined_mmr_position.unwrap())
```
SWvheerden: can these unwraps fail?

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+                            deleted_bitmap_response.height_of_longest_chain,
+                            deleted_bitmap_response.best_block.clone(),
+                        )
+                        .await
+                        .for_protocol(self.operation_id)?;
+                }
+                if deleted_bitmap_response
+                    .not_deleted_positions
+                    .contains(&output.mined_mmr_position.unwrap()) &&
+                    output.marked_deleted_at_height.is_some()
```
SWvheerden, line 161: the word deleted here makes this difficult to follow. 
Am I correct in saying this case, the base_node said this utxo is still unspent, it has a mined_mmr height and the wallet thinks this utxo was spent at some height?

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+            let deleted_bitmap_response = wallet_client
+                .query_deleted(QueryDeletedRequest {
+                    chain_must_include_header: last_mined_header_hash.clone(),
+                    mmr_positions: batch.iter().filter_map(|ub| ub.mined_mmr_position).collect(),
+                })
+                .await
+                .for_protocol(self.operation_id)?;
+
+            for output in batch {
+                if deleted_bitmap_response
```
SWvheerden, line 145: this if checks if the base_node said this utxo was mined at x mmr height but was spent.

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task_v2.rs
```diff
@@ -0,0 +1,420 @@
...
+                    );
+                    break;
+                }
+            } else {
+                // No more outputs
+                break;
+            }
+        }
+
+        loop {
```
SWvheerden: wont this be more efficient to rather do this in a single loop. 
We ask for both the mined and unmined from the db.
We then ask for the block which is the furthest back and we repeat this till we found a height that the base_node still has.

philipr-za (commented):
So I like this approach to doing the validation. By adding some more info about the chain state when outputs and transactions are updated we are able to reduce the complexity of the polling and assumptions around chain state by quite a bit. It also means these calls can be batched quite nicely which will be a major load off.

- I do think we need to be smarter about the interval at which the validation is done. At the very least the interval should be configurable and linked to the power mode so that clients can choose to have a less frequent base node polling

## Output Validation:
- The output validation will set a pending output to be spendable and not encumbered without taking into account the confirmation time
- The get_balance call is still using the `tx_id` field and the PendingTransactionOutputs encumberance mechanism
- cancellation also uses the deprecated `tx_id` field
- Output validation is performed for all outputs including CancelledInbound which is a DoS vector as you could send a wallet many transactions and cancel them and the wallet will now include them in all batch validations.
- Seems like we should remove the pending_transaction_outputs table completely and rather make a new status for pending coinbases
- should remove the concept of Invalid outputs because they are now either spent, unspent or encumbered

## Transaction Validation:
- Should create and emit events for when reorgs occur that can cause a transaction's state to change
- If there are no active transactions we should reduce the frequency of these validations significantly
- PR doesn't invalidate imported Faux transactions when the imported output is invalidated

It is a WIP PR but lots of cleanup needed

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol_v2.rs
```diff
@@ -0,0 +1,417 @@
...
+impl<TTransactionBackend: TransactionBackend + 'static> TransactionValidationProtocolV2<TTransactionBackend> {
+    pub fn new(
+        db: TransactionDatabase<TTransactionBackend>,
+        base_node_pk: CommsPublicKey,
+        connectivity_requester: ConnectivityRequester,
+        config: TransactionServiceConfig,
+        event_publisher: TransactionEventSender,
+    ) -> Self {
+        Self {
+            operation_id: 122, // Get a real tx id
```
SWvheerden, line 77: What is this used for, tracking?
if so cant we just use a random number here?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1012,7 +1012,7 @@ impl InboundTransactionSql {
 
         if num_updated == 0 {
             return Err(TransactionStorageError::UnexpectedResult(
-                "Database update error".to_string(),
+                "Updating inbound transactions failed. No rows were affected".to_string(),
```
SWvheerden, line 998: why not use `num_rows_affected_or_not_found` here

SWvheerden (commented):
Looks good so far. I tried at the beginning to mark all the commented out and unimplemented so that we don't forget about them, but there is still too many so I stopped that halfway through, Those comments can be disregarded.

base_layer/wallet/src/output_manager_service/storage/database.rs
stringhandler: The idea here was to distinguish it from the 'Mined confirmed'. I think it was originally used when a transaction was confirmed, but I need to have the mined_height and hash to confirm it now. Let me have  a look and see if I can add it

hansieodendaal (commented):
This review focussed on system-level behaviour. 

_General observations are listed below:_
- Mini stress test to three receiving wallets of 2000 transactions each:
  - All transactions completed successfully in the receiving wallets
  - All transactions detected as mined confirmed in the receiving wallets

  **Update 2021/08/19 14:34**
- Mini stress test from three sending wallets of 2000 transactions each:
  - All transactions completed successfully in the sending wallets
  - All transactions detected as mined confirmed in the sending wallets

_Noteworthy findings are listed below:_
- Some transaction monitoring process steps are repeated unnecessarily after successful submission to the unconfirmed mempool, draining resources:
  - `Updated transaction ____ as unmined`
  - `Wallet Event Monitor received wallet transaction service event TransactionBroadcast`
  - `Updating transaction ____ as mined`
  - ` Wallet Event Monitor received wallet transaction service event TransactionMined`
- The wallet consumes a lot of CPU resources, >1x full-time CPU thread, when apparently doing nothing, noted to continue long after many transactions have been received.

  **Update 2021/08/19 14:34**
- I could not pick up any transaction events in the log files, i.e. messages containing `event for tx_id`. It seems the transaction service event stream is broken or publishing events are unreachable in the code (`async fn update_transaction_as_mined(`).
- Log message `trace!(target: LOG_TARGET, "Transaction Validation protocol has ended with result {:?}", join_result);` obfuscates warnings and errors; those should be logged as warning or errors:
  - `TRACE Transaction Validation protocol has ended with result Ok(Ok(122))`
  - `TRACE Transaction Validation protocol has ended with result Ok(Err(TransactionServiceProtocolError { id: 122, error: ConnectivityError { source: ConnectionFailed(ConnectFailedMaximumAttemptsReached) } }))`
  - `TRACE Transaction Validation protocol has ended with result Ok(Err(TransactionServiceProtocolError { id: 122, error: RpcError(HandshakeError(TimedOut)) }))`
  - The console wallet transaction statuses are not always updated, e.g. it shows completed when the log shows it is mined.

base_layer/wallet/migrations/2021-07-05-13201407_metadata_signature/up.sql
stringhandler: The schema.rs file has this as nullable, so this makes the DB align with the current rust and schema.rs. Fixing the rust is the correct place to do this, but was out of the scope of this already large PR

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
stringhandler: It's actually never set anywhere in the code at present, so the code would never run. EncumberedToBeSpent was updated by a different process and still is

PR #3192, opened 2021-08-13
Remove old unused integration tests

SWvheerden:
## Description

This PR removes a bunch of old unused integration tests. 

## Motivation and Context

We removed these tests because the helper function to make them function was getting too big and bloated and we could not keep updating them. All of these tests are covered in cucumber with much less mocking. All of these tests also require database access that we removed to make database access more secure. These are for example functions such as `spend_utxo` previously they were all exposed but they are now all internal LMDB functions and we only have access to functions such as add_block which will internally "spend" the utxos.

## Checklist:


hansieodendaal (approved):
LGTM

sdbondi (approved):

PR #3208, opened 2021-08-18
fix: improve p2p RPC robustness

sdbondi:
## Description

- Correctly handles edge case where the client can receive a response after
  the deadline + grace period has expired due to extreme latency. 
- Minor code simplifications
- Adds substream and RPC stress tests to comms
- Increase client-side deadline grace period. This is needed because at
  any point we could be transferring a lot of data, causing delays,
  which the client must tolerate.
- Minor performance optimisations (e.g removed usage of `.split()` which uses a BiLock)
- Change `was marked as offline but was already offline` to debug message

## Motivation and Context

Part of investigating RPC timeouts and RPC "getting into a bad state".

**Summary**

Basically what we're seeing is the limits of what the transport sockets (tor/tcp + noise + yamux) are able to provide.
I believe that the amount of data transferred often during DHT peer exchange + wallet UTXO sync + messaging + whatever else results in large delays that result in timeouts. The delays can result in the server sending a response before the deadline and the client receiving the response after the deadline + grace period - so I've increased the grace period. IMO we need to look at how we can reduce data transmission, possible prime targets are UTXO sync and reducing the data usage of DHT network discovery (less peers transferred, less often, optimal set reconciliation etc). Other potential optimisations may exist in the noise protocol implementation and yamux.

**Stress tests observations**

- Streaming ~100Mb over a local connection, the server can take many seconds (1s-15s) to transmit the data when with some contention. There is nothing unexpected about this, however I set a 4 second timeout with 200ms grace period and some contention over the socket, and so was seeing timeouts that I erroneously did not expect.
- ~~Reliability and performance appears better with many small messages than with fewer large payloads~~ 
- Performance is many times better in release mode, this is most likely because of the cryptography used in stream cipher of the noise protocol

```
running 1 test
'quick' is running...
'quick' completed in 0.01s
'basic' is running...
'basic' completed in 0.27s
'many_small_messages' is running...
'many_small_messages' completed in 1.69s
'few_large_messages' is running...
'few_large_messages' completed in 0.63s
'payload_limit' is running...
'payload_limit' completed in 3.15s
'high_contention' is running...
'high_contention' completed in 25.45s
'high_concurrency' is running...
'high_concurrency' completed in 41.70s
'high_contention_high_concurrency' is running...
'high_contention_high_concurrency' completed in 82.73s
test run ... ok
```

## How Has This Been Tested?

Stress tests test noise protocol + muxer + rpc performance

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4486,6 +4486,7 @@ dependencies = [
  "serde_json",
  "snow",
  "tari_common",
+ "tari_comms_rpc_macros",
  "tari_crypto",
  "tari_shutdown",
  "tari_storage",
--- a/base_layer/wallet/src/connectivity_service/handle.rs
+++ b/base_layer/wallet/src/connectivity_service/handle.rs
@@ -66,7 +66,7 @@ impl WalletConnectivityHandle {
     /// Obtain a BaseNodeWalletRpcClient.
     ///
     /// This can be relied on to obtain a pooled BaseNodeWalletRpcClient rpc session from a currently selected base
-    /// node/nodes. It will be block until this is happens. The ONLY other time it will return is if the node is
+    /// node/nodes. It will block until this happens. The ONLY other time it will return is if the node is
     /// shutting down, where it will return None. Use this function whenever no work can be done without a
     /// BaseNodeWalletRpcClient RPC session.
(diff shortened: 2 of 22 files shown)


stringhandler (dismissed):
Looks good

stringhandler (approved):

PR #3211, opened 2021-08-19
feat: allow DHT to be configured to repropagate messages for a number of rounds

sdbondi:
## Description

Use the dedup cache hit count to allow certain duplicate messages through a
configurable number of times. 

## Motivation and Context

~~This improves mempool synchronization.~~ Implements gossip repropagation that could be used for some message types in future.

## How Has This Been Tested?

New unit test. More manual system tests need to be done

## Checklist:


sdbondi:
@mikethetike lol, thanks

comms/dht/tests/dht.rs
```diff
@@ -536,6 +540,85 @@ async fn dht_propagate_dedup() {
...
+            .comms
+            .connectivity()
+            .dial_peer(node2.node_identity().node_id().clone())
+            .await
+            .unwrap();
+    }
+    // Pre-connect nodes, this helps message passing be more deterministic
+    connect_nodes(&mut node_A, &mut node_B).await;
+    connect_nodes(&mut node_B, &mut node_C).await;
+    connect_nodes(&mut node_C, &mut node_A).await;
```
hansieodendaal: Should this not rather read like this as per comments above?
```suggestion
    connect_nodes(&mut node_A, &mut node_C).await;
```

comms/dht/tests/dht.rs
```diff
@@ -536,6 +540,85 @@ async fn dht_propagate_dedup() {
...
+        .dht
+        .outbound_requester()
+        .propagate(
+            NodeDestination::Unknown,
+            OutboundEncryption::ClearText,
+            vec![],
+            out_msg.clone(),
+        )
+        .await
+        .unwrap();
```
hansieodendaal, line 685: B and C should receive?

comms/dht/tests/dht.rs
```diff
@@ -536,6 +540,85 @@ async fn dht_propagate_dedup() {
...
+        .await
+        .expect("Node expected an inbound message but it never arrived");
+    log::info!("Received message {}", msg.tag);
+
+    node_C
+        .dht
+        .outbound_requester()
+        .propagate(NodeDestination::Unknown, OutboundEncryption::ClearText, vec![], out_msg)
+        .await
+        .unwrap();
```
hansieodendaal: None should receive?

sdbondi: I didn't have time to complete this test, this can either go in or I will complete this test on Sunday.

comms/dht/tests/dht.rs
```diff
@@ -536,6 +540,85 @@ async fn dht_propagate_dedup() {
...
 
+#[tokio_macros::test]
+#[allow(non_snake_case)]
+async fn dht_repropagate() {
+    // Node C knows no one
+    let mut node_C = make_node(PeerFeatures::COMMUNICATION_NODE, []).await;
+    // Node B knows about Node C
+    let mut node_B = make_node(PeerFeatures::COMMUNICATION_NODE, [node_C.to_peer()]).await;
+    // Node A knows about Node B and C
+    let mut node_A = make_node(PeerFeatures::COMMUNICATION_NODE, [node_B.to_peer(), node_C.to_peer()]).await;
```
hansieodendaal: Is the intention not that (1) B sends to C, then (2) A sends the same message to B, who will propagate it to C, so that C will receive a duplicate message?

sdbondi: Correct, this test is not complete but does test that the message is received more than once

comms/dht/tests/dht.rs
hansieodendaal: Thanks for clearing that up, at least I am on the same page then

comms/dht/tests/dht.rs
sdbondi: Can do that, doesn't really make a difference here.

hansieodendaal:
Doing a quick stress test with this latest commit now

hansieodendaal (dismissed):
LGTM, works well, also performed system-level testing under stress/load conditions.

sdbondi:
@hansieodendaal Great to hear - thanks for the tests!

hansieodendaal (dismissed):
LGTM (better wording with `allowed_message_occurrences` thanks!)

comms/dht/examples/memorynet.rs
```diff
@@ -55,9 +55,9 @@ use std::{iter::repeat_with, time::Duration};
 use tari_comms::peer_manager::PeerFeatures;
 
 // Size of network
-const NUM_NODES: usize = 6;
+const NUM_NODES: usize = 30;
 // Must be at least 2
-const NUM_WALLETS: usize = 50;
+const NUM_WALLETS: usize = 5;
```
SWvheerden: Is there a reason we swapped the numbers around. The original numbers are much closer to how we currently run the network from the seed node perspective?

SWvheerden (commented):
The looks good, but I am very much against this approach for the following reasons.:
* The issue we have with the mempool synchronization is that the comms layer is not receiving the messages. The domain layer receives the messages from the comms layer every time. This PR allows that the comms layer forwards the messages more than once (3 times with the current setting) so that the domain layer has to decide multiple times what to do with the same message. 
* Currently this PR does not solve the problem of mempool synchronization across the network,  as the mempool will only propagate a transaction if it is the first time it has seen the transaction. So the second and third times the mempool sees the specific transaction, it will just reject it as duplicate and drop the message.
* If we want the domain layer to send out duplicate propagation we need to revalidate the entire message. For the mempool this means it needs to revalidate the rangeproof etc which is expensive to do, as it only rejects it as duplicate based on the excess of the kernel. This makes it very easy to get a node to forward "broken" transactions across the network for you if you just keep the kernel the same and change the rest of the transaction.
* The domain layer should not care about duplicated messages. It should only receive a message once. It should not change what processing and results happen if it receives a message more than once.
* If we send a message 3 times as is with this PR, we will effectively almost triple the network traffic. We will send a message out to 4 random peers from the connected peers. If we only have 4 connected peers, we will spam those same 4 peers multiple times. 

As far as I understand the problem and what this PR whats to accomplish is it wants a greater percentage of the network to see the message. How we select peers and the number of peers we forward to makes a very big difference on the network view of messages.  Rather than roll the dice to select 4 random peers 3 times, why not just roll the dice to select 12 peers once. This means you will not forward messages to duplicated peers, and your own domain layer does not have to do extra work to validate duplicate information.

comms/dht/examples/memorynet.rs
sdbondi: Can change them back - just for my testing with many base nodes

sdbondi:
@SWvheerden Thanks for the review. Definitely up for discussion. This can be turned off so I think the idea is to experiment. We can also experiment with sending to all connected peers (https://github.com/tari-project/tari/pull/3239)

hansieodendaal:
Linking on to the discussion by @SWvheerden. The attached trace log is interesting as it highlights those points exactly, looking from a final transaction broadcast to mempool perspective, created by using this PR branch. In summary:

1) This node: Finalized transaction received (from the base node connected to the sender wallet), sent to mempool
2) This node: mempool receives the transaction, insert it into the mempool & propagate to network
3) This node: Transaction propagated by comms
4) This node: Duplicate final transaction received by, sent to mempool again
5) This node: Mempool receives duplicate transaction, rejects it
6) This node: Duplicate final transaction received, dedup discards it
7) This node: Duplicate final transaction received, dedup discards it
8) This node: Duplicate final transaction received, dedup discards it
9) Receiver wallet: Submits finalized transaction to this base node
10) This node: Mempool receives duplicate transaction from receiver wallet, rejects it

[PR3211_review_from_mempool_perspective.log](https://github.com/tari-project/tari/files/7038894/PR3211_review_from_mempool_perspective.log)

sdbondi:
Updated this PR to make the mempool repropagate duplicate messages

sdbondi:
Think we need to evaluate if this is a better approach to just sending to all connected peers before putting effort into making the mempool work with this approach `validate -> add (if necessary) -> repropagate` . This process requires a bit fo a refactor actually, or we add logic to the dht which I'm not a fan of

hansieodendaal:
> @SWvheerden Thanks for the review. Definitely up for discussion. This can be turned off so I think the idea is to experiment. We can also experiment with sending to all connected peers (#3239)

That PR is definitely worth trying out as it will/should impact mempool sync across nodes, but I also understood that this PR might improve the general connectivity/connectedness factor in the network?

sdbondi:
I will change this to have a repropagation of 1 i.e is functionally equivalent to before the PR

comms/dht/src/config.rs
```diff
@@ -151,6 +156,7 @@ impl Default for DhtConfig {
             saf_max_message_size: 512 * 1024,
             dedup_cache_capacity: 2_500,
             dedup_cache_trim_interval: Duration::from_secs(5 * 60),
+            dedup_allowed_message_occurrences: 1,
```
hansieodendaal, line 159: **NIT:** We may need a TODO comment here to test or further implement this feature.

hansieodendaal (approved):
LGTM

Just one NIT below.

stringhandler:
@SWvheerden you still against this? Can I merge it?

stringhandler (approved):

SWvheerden:
@mikethetike I am still against the idea of repropagation, but as it stands this PR does not do that anymore, so this PR can go in. You can merge it

PR #3215, opened 2021-08-19
fix(wallet_ffi)add null check for `transport_type` in `create_comms_config` FFI function

philipr-za:
## Description
The `transport_type` argument is not checked is it null before it is used in this method. This PR adds the check and appropriate error response.

## How Has This Been Tested?
Not tested

## Checklist:


stringhandler (approved):

StriderDM (approved):
Nice catch

PR #3216, opened 2021-08-19
Add extra logging detail for wiremode warnings

delta1:
## Description

Minor log warning improvements for the case where a wire format byte is not received 

## Motivation and Context

Adding more log info for some observed warnings 

## How Has This Been Tested?

Replaced existing base node and started syncing a new node successfully 

## Checklist:

### DIFF
--- a/comms/src/connection_manager/listener.rs
+++ b/comms/src/connection_manager/listener.rs
@@ -30,7 +30,11 @@ use super::{
 };
 use crate::{
     bounded_executor::BoundedExecutor,
-    connection_manager::{liveness::LivenessSession, types::OneshotTrigger, wire_mode::WireMode},
+    connection_manager::{
+        liveness::LivenessSession,
+        types::OneshotTrigger,
+        wire_mode::{WireMode, LIVENESS_WIRE_MODE},
+    },
     multiaddr::Multiaddr,
     multiplexing::Yamux,
     noise::NoiseConfig,
@@ -56,6 +60,7 @@ use log::*;
 use std::{
     convert::TryInto,
     future::Future,
+    io::{Error, ErrorKind},
     sync::{
         atomic::{AtomicUsize, Ordering},
         Arc,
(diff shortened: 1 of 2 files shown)


stringhandler (commented):
So I think the actual problem here is in `read_wire_format`.

In my logs I see the warn "WARN  Peer at address '/ip4/127.0.0.1/tcp/55261' failed to send valid wire format" but I don't see either of the log messages from the branches of the match statements in `read_wire_format`, which leads me to believe the reason that it did not receive a wire format is due to a timeout (resulting in a None from the ?) 
I would prefer the log to reflect that a timeout occurred

delta1:
So in the errors I saw, `read_wire_format` is returning None due to a `unexpected end of file` error which I think is the socket being closed before the connection is established? Will look a bit more into it but I didn't see a timeout error.

hansieodendaal:
@delta1 I only saw one occurrence of `WARN  Failed to read first byte: unexpected end of file`.

@mikethetike I agree with your assessment, as a timeout in 
   `match time::timeout(time_to_first_byte, socket.read_exact(&mut buf)).await.ok()?` 
will result in `None` to be returned to 
   `match Self::read_wire_format(&mut socket, config.time_to_first_byte).await`. 
That error message results in 
   `WARN  Peer at address '/ip4/127.0.0.1/tcp/1429' failed to send valid wire format`, 
which is slightly misleading. The message should rather be something like 
   `WARN  Peer at address '/ip4/127.0.0.1/tcp/1429' failed to send its wire format before the timeout of 7s was reached`. 
I counted many of those _hidden timout_ errors.

delta1:
> @delta1 I only saw one occurrence of `WARN Failed to read first byte: unexpected end of file`.


Every instance in datadog for the past month is this warning, I can't find any timeouts.

delta1:
@hansieodendaal can you find any instances of the log message "deadline has elapsed" ? I think that's what the timeout would log in `read_wire_format`

delta1:
You can see in this [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b2ca5f929c4ec545a6ff5d8ff31b7275) that the timeout Elapsed error will display "deadline has elapsed" 

Please grep your logs for instances of "Failed to read first byte" so we can see what errors we're getting?

hansieodendaal:
> @hansieodendaal can you find any instances of the log message "deadline has elapsed" ? I think that's what the timeout would log in `read_wire_format`

Nothing like this, as the error is converted into a `None`

hansieodendaal:
> You can see in this [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b2ca5f929c4ec545a6ff5d8ff31b7275) that the timeout Elapsed error will display "deadline has elapsed"
> 
> Please grep your logs for instances of "Failed to read first byte" so we can see what errors we're getting?

I have deleted my old logs, but saw before that this error is very rare, also rare in new tests I am running.

delta1:
> Nothing like this, as the error is converted into a `None`

The error is first logged as a warning in `read_wire_format` 

image

delta1:
> I have deleted my old logs, but saw before that this error is very rare, also rare in new tests I am running.

Yep having a timeout is rare, but there are over 160 instances of "unexpected end of file" in datadog in the last month

hansieodendaal (approved):
LGTM

sdbondi (approved):
Nice looks like a great improvement to logs 👍

PR #3217, opened 2021-08-19
fix: bug in wallet base node peer switching

sdbondi:
## Description

- Connectivity retries connecting indefinitely, however previously if
  this continuously fails and the user updates their peer, the new peer will not be
  read until the previous peer was connected to. This PR fixes this.
- Add protocl information to comms RPC logs
- Cleanup peer state, making the wallet connectivity service the source
  of truth for the base node peer
- Adds an `Ack` flag to RPC protocol, this is not currently used but
  could be implemented in the client side in future if required without
  breaking the network (server supports it, we may choose in future to implement a client keep alive should it be needed,
 current server side impl easy and zero cost).

## Motivation and Context

Critical bug fix 

## How Has This Been Tested?

Switching the console wallet base node peer a number of times

## Checklist:

### DIFF
--- a/base_layer/wallet/src/base_node_service/service.rs
+++ b/base_layer/wallet/src/base_node_service/service.rs
@@ -50,7 +50,7 @@ pub struct BaseNodeState {
     pub updated: Option<NaiveDateTime>,
     pub latency: Option<Duration>,
     pub online: OnlineStatus,
-    pub base_node_peer: Option<Peer>,
+    // pub base_node_peer: Option<Peer>,
 }
 
 impl Default for BaseNodeState {
@@ -61,7 +61,6 @@ impl Default for BaseNodeState {
             updated: None,
             latency: None,
             online: OnlineStatus::Connecting,
-            base_node_peer: None,
         }
     }
 }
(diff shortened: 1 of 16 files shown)


base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -158,18 +157,7 @@ where T: WalletBackend + 'static
...
-        };
-
-        {
-            let mut lock = self.state.write().await;
-            *lock = new_state.clone();
-        }
-        self.wallet_connectivity.set_base_node(peer.node_id.clone()).await?;
-
-        self.publish_event(BaseNodeEvent::BaseNodeStateChanged(new_state));
+        self.wallet_connectivity.set_base_node(peer.clone()).await?;
```
philipr-za, line 160: This feels like it should happen in the wallet container `set_base_node_peer` function where all the other wallet services have their set_base_node method called.

sdbondi: Yeah agree,  will require a bit more of a refactor (`BaseNodeEvent::BaseNodePeerSet` event needs to be moved around / base node service needs to monitor the change and emit the event), wanted to get this in quicker.

stringhandler (approved):

PR #3223, opened 2021-08-20
fix: chain error caused by zero-conf transactions and reorgs

SWvheerden:
## Description

If a zero-conf transaction was in a block and the block is rewound. The block_chain backend will try to delete in input which was never marked as unspent as it was immediately spent. When we rewound this block we should check this and not try and delete an output that was never an unspent output on the chain. 

## Checklist:


sdbondi (approved):
Nice one! utAck

PR #3224, opened 2021-08-20
test: improve comms limits to combat network timeouts

hansieodendaal:
## Description
This PR:
- Improved various comms configuration settings to combat RPC connection and response timeouts
- Improved wire mode logging

The philosophy here is to rather wait for a connection or response than to abandon all and try again.

These settings were tested on two separate systems performing system-level tests where RPC timeouts and connection problems were previously prevalent:
- while base node/console wallet pairs were only monitoring the network or were linked to SHA3 or RandomX miners
- while performing a stress test and compiling Rust at the same time.

The former proved to run virtually without any errors while the latter registered some timouts, especially when performing Rust compilations (~ two orders of magnitude less).

**Edit:** 
- Fixed flaky cucumber test `Node should not sync from pruned node`
- Fixed magic numbers in unit test `test_txo_validation_rpc_timeout`

## Motivation and Context
See above

## How Has This Been Tested?
See above

## Checklist:

### DIFF
--- a/comms/src/connection_manager/dialer.rs
+++ b/comms/src/connection_manager/dialer.rs
@@ -514,7 +514,7 @@ where
                             .map_err(|_| ConnectionManagerError::WireFormatSendFailed)?;
 
                         let noise_socket = time::timeout(
-                            Duration::from_secs(30),
+                            Duration::from_secs(40),
                             noise_config.upgrade_socket(socket, ConnectionDirection::Outbound),
                         )
                         .await
--- a/comms/src/connection_manager/listener.rs
+++ b/comms/src/connection_manager/listener.rs
@@ -42,6 +42,7 @@ use crate::{
     protocol::ProtocolId,
     runtime,
     transports::Transport,
+    types::CommsPublicKey,
     utils::multiaddr::multiaddr_to_socketaddr,
     PeerManager,
 };
@@ -67,6 +68,7 @@ use std::{
     },
     time::Duration,
 };
+use tari_crypto::tari_utilities::hex::Hex;
 use tari_shutdown::ShutdownSignal;
 use tokio::time;
 
@@ -249,7 +251,7 @@ where
                     let result = Self::perform_socket_upgrade_procedure(
                         node_identity,
                         peer_manager,
-                        noise_config,
+                        noise_config.clone(),
                         conn_man_notifier.clone(),
                         socket,
                         peer_addr,
@@ -286,10 +288,12 @@ where
                     }
                 },
                 Ok(WireMode::Comms(byte)) => {
+                    let public_key = Self::remote_public_key_from_socket(socket, noise_config).await;
                     warn!(
                         target: LOG_TARGET,
-                        "Peer at address '{}' sent invalid wire format byte. Expected {:x?} got: {:x?} ",
+                        "Peer at address '{}' ({}) sent invalid wire format byte. Expected {:x?} got: {:x?} ",
                         peer_addr,
+                        public_key,
                         config.network_info.network_byte,
                         byte,
                     );
@@ -313,11 +317,13 @@ where
                     }
                 },
                 Err(err) => {
+                    let public_key = Self::remote_public_key_from_socket(socket, noise_config).await;
                     warn!(
                         target: LOG_TARGET,
-                        "Peer at address '{}' failed to send wire format. Expected network byte {:x?} or liveness \
-                         byte {:x?} not received. Error: {}",
+                        "Peer at address '{}' ({}) failed to send its wire format . Expected network byte {:x?} or \
+                         liveness byte {:x?} not received. Error: {}",
                         peer_addr,
+                        public_key,
                         config.network_info.network_byte,
                         LIVENESS_WIRE_MODE,
                         err
@@ -329,6 +335,32 @@ where
         // This will block (asynchronously) if we have reached the maximum simultaneous connections, creating
         // back-pressure on nodes connecting to this node
         self.bounded_executor.spawn(inbound_fut).await;
+    }
+
+    async fn remote_public_key_from_socket(socket: TTransport::Output, noise_config: NoiseConfig) -> String {
+        let public_key: Option<CommsPublicKey> = match time::timeout(
+            Duration::from_secs(30),
+            noise_config.upgrade_socket(socket, ConnectionDirection::Inbound),
+        )
+        .await
+        .map_err(|_| ConnectionManagerError::NoiseProtocolTimeout)
+        {
+            Ok(Ok(noise_socket)) => {
+                match noise_socket
+                    .get_remote_public_key()
+                    .ok_or(ConnectionManagerError::InvalidStaticPublicKey)
+                {
+                    Ok(pk) => Some(pk),
+                    _ => None,
+                }
+            },
+            _ => None,
+        };
+
+        match public_key {
+            None => "public key not known".to_string(),
+            Some(pk) => pk.to_hex(),
+        }
     }
 
     #[allow(clippy::too_many_arguments)]
--- a/comms/src/connection_manager/listener.rs
+++ b/comms/src/connection_manager/listener.rs
@@ -288,6 +288,9 @@ where
                     }
                 },
                 Ok(WireMode::Comms(byte)) => {
+                    // TODO: This call is expensive and only added for the benefit of improved logging and may lead to
+                    // TODO: DoS attacks. Remove later when not needed anymore or make it optional with a config file
+                    // TODO: setting.
                     let public_key = Self::remote_public_key_from_socket(socket, noise_config).await;
                     warn!(
                         target: LOG_TARGET,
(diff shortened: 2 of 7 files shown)


stringhandler (dismissed):
Ideally we wouldn't need these longer timeouts, but I think let's get back to a stable state and then we can reduce them

comms/src/connection_manager/listener.rs
```diff
@@ -296,9 +300,14 @@ where
                     }
                 },
                 None => {
+                    let public_key = Self::remote_public_key_from_socket(socket, noise_config).await;
```
sdbondi: `noise_config.upgrade_socket(socket, ConnectionDirection::Inbound),` - this performs the noise handshake with IO - it will always fail in this branch (unless someone decides to delay the first time they send a byte to perform the noise handshake only to be disconnected immediately after). This will always return `"public key not known".` after using up a socket and system resources.

comms/src/connection_manager/listener.rs
```diff
@@ -269,10 +271,12 @@ where
                     }
                 },
                 Some(WireMode::Comms(byte)) => {
+                    let public_key = Self::remote_public_key_from_socket(socket, noise_config).await;
```
sdbondi, line 294: I don't understand the rationale insofar as it relates to this PR "improving comms limits". 

1. All we know at this point is some connection from some/anywhere (browser, telnet, literally any TCP client) send a byte other than `0x46`.
2. In the happy/non-shenanigan case you are performing the noise socket protocol with peers that are configured on different networks so that your logs can have a public key of that node. 
3. The "connection is closed" logs I've seen on seed nodes are all related to RPC substreams (2+ days ago, none in the past day, though I have only checked one node). This part of the code executes at the inception of a `PeerConnection` before any RPC. Are we actually seeing this branch hit (unless a peer is not configured for weatherwax - which is the case it handles by denying the connection i.e. on the peer connection level, connection is closed == connection denied)?

comms/src/connection_manager/dialer.rs
```diff
@@ -514,7 +514,7 @@ where
                             .map_err(|_| ConnectionManagerError::WireFormatSendFailed)?;
 
                         let noise_socket = time::timeout(
-                            Duration::from_secs(30),
+                            Duration::from_secs(40),
```
sdbondi, line 517: I don't see a single instance in the seed node log of `"Noise protocol handshake timed out"` in the past month (or more) - it's ok if it is increased, hopefully there is a good reason why.

sdbondi (commented):
Increasing timeouts is fine as they will obviously reduce the probability of a timeout occurring with the current latency we're seeing (timeouts being a necessary part of connections in general, over the internet and esp over tor - but current latency are inordinately large. related: https://github.com/tari-project/tari/pull/3227). Agree with @mikethetike that we should seek to reduce the timeouts later. Just a thought, tokio's timeouts AFAIK do not take into account how busy the process/system is. They use real time whereas our code runs in cpu time. I wonder if a cpu-time based timeout would help here, essentially allowing more time if the system is being hammered.

I disagree with performing the (non-free IO and RTT) noise protocol handshake with a non-compliant non-authenticated connection just to get the public key (in the happy case) of a remote wallet/base node configured on the wrong network. I'm not convinced that is particularly useful and may add some unforeseen attack risk - is there a particular reason you added this, or was it because it seemed like good info to have?

hansieodendaal:
> I disagree with performing the (non-free IO and RTT) noise protocol handshake with a non-compliant non-authenticated connection just to get the public key (in the happy case) of a remote wallet/base node configured on the wrong network. I'm not convinced that is particularly useful and may add some unforeseen attack risk - is there a particular reason you added this, or was it because it seemed like good info to have?

This was a special request from @mikethetike to try and track down peers sending invalid wire formats all the time, but I do understand that it is expensive to get that information. We could maybe add a `TODO` note to remove these two calls at a later stage?

comms/src/connection_manager/dialer.rs
hansieodendaal: I saw this a couple in local tests between my laptop and desktop and especially under load, therefore the increase.

comms/src/connection_manager/listener.rs
hansieodendaal: Yes, I have seen ample of these messages coming through, see some sampled extracts below:
``` rust
2021-08-20 19:58:07.777564100 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1170' (be4d48ce68e4c7a05593159e9a11e75c0d646199b44115909e7bcf2faf7ef265) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 19:58:35.094426800 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1845' (be4d48ce68e4c7a05593159e9a11e75c0d646199b44115909e7bcf2faf7ef265) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 19:59:07.773171900 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/23093' (be4d48ce68e4c7a05593159e9a11e75c0d646199b44115909e7bcf2faf7ef265) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 19:59:34.088223100 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1193' (be4d48ce68e4c7a05593159e9a11e75c0d646199b44115909e7bcf2faf7ef265) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 20:08:22.288730400 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1243' (421557737326d3b2d355383094b5b9916dc62513bebacaddeb8afdb9877ae541) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 20:10:47.750326800 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/22301' (421557737326d3b2d355383094b5b9916dc62513bebacaddeb8afdb9877ae541) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 20:13:17.745209600 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1241' (421557737326d3b2d355383094b5b9916dc62513bebacaddeb8afdb9877ae541) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 20:17:47.710094300 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1944' (421557737326d3b2d355383094b5b9916dc62513bebacaddeb8afdb9877ae541) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 20:21:17.915394000 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1218' (f860e6654457bfcfd73f1a70c233b94dc1d139a0d1b2812c9d7e833ec409a568) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 21:11:52.915466300 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1033' (4e148b8edf987a108d2dc67ff379e5587a73bf09dc780ba4d21e7a8802d9d867) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 21:20:27.314824700 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/1803' (a8f95d9997d962fda575e2a93713a887fe2d44af2b2756d87ffaf696eeeb374e) sent invalid wire format byte. Expected 23 got: 6 
2021-08-20 21:22:58.413287200 [comms::connection_manager::listener] WARN  Peer at address '/ip4/127.0.0.1/tcp/23306' (a0b24a255230cb9307e3fc21510e278ef03f99339bacc632770be1590d909f0a) sent invalid wire format byte. Expected 23 got: 6 
```

comms/src/connection_manager/listener.rs
hansieodendaal: Yes, you are correct, I only saw `"public key not known"` in the logs and the edge case you mentioned also did not appear, so removing this line.

comms/src/connection_manager/listener.rs
hansieodendaal: > I don't understand the rationale insofar as it relates to this PR "improving comms limits".

Adding the public key to wire mode logging was not the major focus of the PR, just a side issue. The important change was that waiting for the first byte to arrive was upped from 7s to 45s as the former proved to be inadequate during load conditions.

hansieodendaal:
> Just a thought, tokio's timeouts AFAIK do not take into account how busy the process/system is. They use real time whereas our code runs in cpu time. I wonder if a cpu-time based timeout would help here, essentially allowing more time if the system is being hammered.

My guess is that using cpu-based timeouts will make matters worse as it will be totally unpredictable. 

Two interesting graphs relating to RPC session negotiation latency during load conditions as it is being logged below (_understood that this may involve other tasks as well as mentioned_). In this case, the sender and receiver wallets were compiled from different branches, and currently branch #3191 is a lot less optimized.

image

image

comms/src/connection_manager/listener.rs
sdbondi: Ok so why are these messages coming through? Are these nodes configured to the correct network? Briefly looking at git history, I think byte 6 is stibbons

stringhandler (approved):

PR #3228, opened 2021-08-22
feat: add support for forcing sync from seeds

Cifko:
## Description
Add support for forcing sync to a seed node. Specify index for the node from peer_seeds list.

## How Has This Been Tested?
Manually.

## Checklist:


applications/daily_tests/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "endOfLine": "auto"
```
sdbondi, line 2: Is this a problem on windows? Mixing line endings between files does not seem like a good idea, preferably any editor regardless of OS should be fine with `lf` line endings. Our policy is to use prettier defaults so that we avoid bike shedding.

sdbondi (commented):
Looks good - just one concern

applications/daily_tests/.prettierrc
Cifko: Without it on windows the prettier reformats the whole file. The LF endings are forced by git settings anyway, so this should not affect the commits. But without this setting the editing on windows mess up the compare during development. By the way I copied this form the integration folder.

sdbondi: Ah ok makes sense

applications/daily_tests/automatic_sync_test.js
```diff
@@ -9,38 +9,72 @@ const SyncType = {
...
   } else {
     syncType = SyncType.Archival;
   }
 
-  const { blockHeight, timeDiffMinutes, blockRate } = await run({
-    log: "./logs/base-node.log",
-    syncType,
-  });
+  // Check for forced index.
+  let index = process.argv.findIndex((element) => element === "forced");
```
sdbondi: I think now that we are using cli args with arguments, use yargs like  automated_recovery_test

applications/daily_tests/automatic_sync_test.js
```diff
@@ -9,38 +9,72 @@ const SyncType = {
...
+    config = config
       .toString()
       .replace("#pruning_horizon = 0", "pruning_horizon = 1000");
-    await fs.writeFile(
-      baseNode.baseDir + "/config/config.toml",
-      updated_config
+  }
+
+  // If we have forced index, rewrite the config.
+  if (options.forced !== undefined) {
```
stringhandler: Can an ENV not be used here instead?

stringhandler: e.g. TARI_WEATHER_WAX__FORCE_SYNC_PEERS

Cifko: There is no support for force sync env yet. But there will be after my second PR.
So I'm going to change the it to env.

applications/daily_tests/automatic_sync_test.js
Cifko: Done.

applications/daily_tests/automatic_sync_test.js
Cifko: Done

Cifko:
This **HAS**  to be pushed with the daily scripts. Because I switch to yargs and now it uses different notation for the arguments. I'm not aware of any other place where this script is called. Don't push if it's used somewhere else.

sdbondi (approved):
Nice - the [daily cron stuff](https://github.com/tari-project/tari/blob/development/applications/daily_tests/cron_jobs.js) actually passes in options to `run` so should be fine. Need to add/tweak cron to use the forceSyncPeer option, but can be in another PR.

PR #3230, opened 2021-08-23
test: cucumber forced sync

Cifko:
## Description
Add cucumber test that tests forced sync to single node.

## How Has This Been Tested?
npm test -- --name "Force sync many nodes agains one peer"

## Checklist:


Cifko:
The problem is that syncing 1000 nodes is not working. 999 works fine. Do we forbid sync over 1000 blocks (I don't think so). Haven't look at the error yet.
There are two options right now
1) set it to 999 blocks, create new ticket for the 1k sync and change it later to 1k
2) leave it at 1k and investigate now

integration_tests/features/Sync.feature
```diff
@@ -167,3 +167,24 @@ Scenario: Pruned mode network only
     Then node PNODE2 is at height 20
     Given I have a pruned node PNODE3 connected to node PNODE1 with pruning horizon set to 5
     Then node PNODE3 is at height 20
+
+  Scenario Outline: Force sync many nodes agains one peer
```
delta1, line 174:
```suggestion
  Scenario Outline: Force sync many nodes against one peer
```

delta1 (changes requested):
Looking good, just a typo. Testing locally now.

delta1 (dismissed):
Works well!

philipr-za:
> The problem is that syncing 1000 nodes is not working. 999 works fine. Do we forbid sync over 1000 blocks (I don't think so). Haven't look at the error yet.
> There are two options right now
> 
>     1. set it to 999 blocks, create new ticket for the 1k sync and change it later to 1k
> 
>     2. leave it at 1k and investigate now

I believe this must be a bug related to the pruning horizon which is by default 1000 blocks?

stringhandler (approved):

PR #3236, opened 2021-08-23
fix: block sync validation

SWvheerden:
Description
---
This fixes the block sync block body validation to validate the correct items
This consolidates all the syncing helper functions
This adds in unit tests for the validators

Motivation and Context
---
Block sync needs to validate all the required steps. This also consolidates all the possible validation functions to remove duplicated pieces of code between the validators so we use a single piece to test a specific consensus item. 

How Has This Been Tested?
---
Synced a node up to the tip height 22014. 
Added new unit tests. 
Ran all integration and unit tests.


stringhandler:
Adding a `do not merge` temporarily, just want to take a proper look through this

sdbondi (dismissed):
utAck - these tests are badly needed and I agree that we need the added checks in block sync.

stringhandler (dismissed):
Nice. utACK

sdbondi (approved):

PR #3237, opened 2021-08-24
fix: add periodic connection check to wallet connectivity service

sdbondi:
Description
---
- Adds a periodic check of the connection status and attempts a
  reconnect if no longer connected. Previously it was assumed that this
  can be done lazily because some caller will always call
  `obtain_base_node_wallet_rpc_client`, but this may not be the case. A
  periodic check is added.
- Clean up some state checking to use the wallet connectivity service.

Motivation and Context
---
Improves snappiness of the connectivity and chain state updates in the wallet

How Has This Been Tested?
---
Manually on the console wallet + existing tests


stringhandler (approved):
Seems ok (untested).

PR #3238, opened 2021-08-24
test: add tracing to comms via --tracing-enabled

stringhandler:
Description
---
Add tracing to comms to debug timings via the `--tracing-enabled` flag

Motivation and Context
---
It's currently difficult to understand the timings of network calls and errors in the application.

How Has This Been Tested?
---
Tested manually

### DIFF
--- a/comms/src/protocol/messaging/protocol.rs
+++ b/comms/src/protocol/messaging/protocol.rs
@@ -247,6 +247,7 @@ impl MessagingProtocol {
         Ok(())
     }
 
+    // #[tracing::instrument(skip(self, out_msg), err)]
     async fn send_message(&mut self, out_msg: OutboundMessage) -> Result<(), MessagingProtocolError> {
         let peer_node_id = out_msg.peer_node_id.clone();
         let sender = loop {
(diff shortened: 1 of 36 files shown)


comms/src/protocol/messaging/protocol.rs
```diff
@@ -247,6 +247,7 @@ impl MessagingProtocol {
         Ok(())
     }
 
+    // #[tracing::instrument(skip(self, out_msg), err)]
```
philipr-za, line 250: Maybe remove if its not being used but 🤷‍♂️

comms/src/protocol/rpc/client.rs
```diff
@@ -494,14 +505,17 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin + Send
 
         let (mut response_tx, response_rx) = mpsc::channel(10);
         if reply.send(response_rx).is_err() {
-            debug!(target: LOG_TARGET, "Client request was cancelled.");
+            event!(Level::WARN, "Client request was cancelled");
```
philipr-za, line 508: I believe these two lines are equivalent? The warn! is an event with this level so would remove one of these lines.

stringhandler: I didn't change all of the logs/warns to the tracing crate. I actually worried that we end up losing some logs, so I didn't take a blanket approach of swapping them out. At some point we should do it, but in the mean time there may be a period of duplicate logs

philipr-za (approved):
LGTM

PR #3239, opened 2021-08-24
fix: send transactions to all connected peers

sdbondi:
Description
---
 Send transactions to all connected peers as we do with block propagation

Motivation and Context
---
Alternative to #3211.

How Has This Been Tested?
---
Existing tests / single line code change


SWvheerden (approved):
LGTM this is a much better approach than #3211 
This way we send to every connected peer and we know it's going to reach every node

hansieodendaal (approved):
LGTM - I think we should try this out to see if it will alleviate the mempool sync issues.

**Note:** Please correct me if I am wrong, but #3211 in its current form might still have a purpose, just not solve the mempool sync issues?

sdbondi:
@hansieodendaal #3211 will repropagate messages without a destination (so must be gossiped to the whole network), currently the only messages that we have like this are cleartext transaction and block propagation messages. So I think we can definitely keep it, but perhaps disable it and try this approach first. In future, we can optimise the transaction propagation protocol.

philipr-za (commented):
Ja with the Dedup layer I think this is fine and will achieve faster saturation. Will the message be propagated further from these peers?

sdbondi:
@philipr-za Yeah, each unique valid transaction message received will hit the `handle_outbound_tx` (this time calling `flood`)  as before

philipr-za:
> @philipr-za Yeah, each unique valid transaction message received will hit the `handle_outbound_tx` (this time calling `flood`) as before

This really feels like overkill but I suppose its because our network is still so small so the connectivity graph is sparse. My intuition is that, as you said above, in the future when the size of the network grows we will be able to return to a more efficient gossip.

sdbondi:
@SWvheerden brought my attention to this paper which describes bitcoin's method (See `B Propagation Method`) - EDIT 😆 : https://www.gsd.inesc-id.pt/~ler/docencia/rcs1314/papers/P2P2013_041.pdf

Essentially it comes down to efficient protocol design where the initial inv message is small enough to be sent to everyone.

delta1:
> @SWvheerden brought my attention to this paper which describes bitcoin's method (See `B Propagation Method`) - essentially it comes down to efficient protocol design where the initial inv message is small enough to be sent to everyone.

Missing a link @sdbondi ?

SWvheerden:
https://www.gsd.inesc-id.pt/~ler/docencia/rcs1314/papers/P2P2013_041.pdf

haha, saw it's edited in.

delta1:
rofl thanks!

hansieodendaal:
Ran some system-level tests; this approach sending transactions to all connected peers definitely makes a big difference in mempool sync across nodes. Previously I had to connect a miner to ~4 of 7 base nodes I am running for tests so that all transactions could get mined, but with this change, I did not have to re-point the miner to any other base node apart from the one it is connected to by default.

delta1:
+1

hansieodendaal (approved):
LGTM

Definitely did the job during the recent stress test: "_Mempool view of the network drastically improved due to sending validated transactions to all connected peers instead of only to a select few. In the end, all monitored base nodes had an empty mempool with all broadcast transactions being mined without having to connect a miner to different base nodes afterwards._"

stringhandler (approved):

PR #3241, opened 2021-08-24
test: add sync against one test

Cifko:
Description
---
Add a script that syncs one node (always archival node) and then syncs **_n_**-nodes to the one created. It supports testing prune nodes as well (not the combinations).

Motivation and Context
---

How Has This Been Tested?
---
Manually _**node .\sync_against_one_test.js --nodes 5**_


Cifko:
I was not able to finish the test successfully. All nodes disconnects at tip ~247 (~6 minutes after connection init), at the same time.
From the base node:
> 13:06 WARN  Failed to read wire format byte within timeout of 7s. deadline has elapsed
13:06 WARN  Peer at address '/ip4/127.0.0.1/tcp/59586' failed to send wire format. Expected network byte 23 or liveness byte 46 not received. Error: timed out

They never recover, and they are not banned by the node.

stringhandler (approved):

stringhandler:
This is adding a new file, it can be merged

PR #3242, opened 2021-08-24
fix: remove openssl from windows runtime

hansieodendaal:
Description
---
- Removed OpenSSL installers and dependencies from Windows runtime
- Removed stdout as appender from console wallet's log4rs logger

Motivation and Context
---
- OpenSSL is not a Windows runtime dependency anymore
- The console wallet cannot handle log4rs logger messages to its stdout as it breaks the TUI

How Has This Been Tested?
---
Build the installer, ran an installer, ran all the executable components


delta1 (approved):
Nice one! Haven't tested but looks good.

PR #3248, opened 2021-08-25
fix: disable P2P transaction negotiation while recovery is in progress

philipr-za:
Description
---
Some weird behaviour was observed when a wallet would be busy with recovery and then receive transaction negotiation messages, either directly or via SAF.

The Recovery process is updating the Key Manager Indices and looking for commitments on the blockchain so to allow transaction negotiation during this time is dangerous as it might put duplicate commitments into the db and reuse spending keys.

This PR checks for the db key/value used to indicate Recovery progress before handling a transaction negotiation p2p message and if it is there the message is ignored with a log.

How Has This Been Tested?
---
Manual test


stringhandler:
Needs clippy fixed

stringhandler (dismissed):

PR #3255, opened 2021-08-27
feat: improve basenode switch from listening to lagging mode

SWvheerden:
Description
---
This PR changes the peer metadata push to listing mode speed to push every time it receives a chain metadata ping or pong message. 

Motivation and Context
---
This is introduced to allow a node to switch faster and not wait till it received it all the pings and pongs from a node.

How Has This Been Tested?
---

Run all unit tests and manually ran node.


base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -181,37 +182,29 @@ impl ChainMetadataService {
                     event.node_id
                 );
                 self.collect_chain_state_from_pong(&event.node_id, &event.metadata)?;
-
-                // All peers have responded in this round, send the chain metadata to the base node service
-                if self.peer_chain_metadata.len() >= self.peer_chain_metadata.capacity() {
```
sdbondi, line 186: Perhaps we can change this to a majority instead of all?

sdbondi (commented):
These changes make it extremely trivial for an attacker to switch base nodes to sync mode. This is the reason why it may be delayed sometimes, because all peers are required to 'vote' for a chain tip. Perhaps requiring a majority is a happy middle ground.

The listener currently assumes that the metadata it receives from the chain metadata service has "enough" data from peers i.e listener does not have to check that number of metadata entries to do it's work, chain metadata service will give it the right data.

SWvheerden:
Mmm, that's not quite how the current logic works. All this PR does is increase the frequency the metadata is sent. There is no voting etc for a chain tip. 

Currently, the metdata_service waits for all replies then send them all to the listing state to decide what to do with them. All it does is looks for the highest one, and compares that to its known peers, and its self. If any single one is higher it will go to lagging, and try and recover that block. 

This PR just increases the timing this all happens. The attack vector you described will definitely force a node to go to lagging, but if the data is invalid he will be banned. This whole process will just complete faster. But the logic of how and when it goes to lagging is completely unchanged.

sdbondi:
Ah that's right, there is no voting. The changes in this PR are fine (as is no worse than it was before, does decrease time to first syncing), but feels like we should get a majority consensus on the chain tip before kicking into sync mode. Either that, or switching to sync mode should not prevent mining unless the superior chain tip has been verified (something that header sync does). 

LGTM once clippy is fixed

stringhandler (approved):

stringhandler:
Yeah, I don't think it's the right place to protect against the attack that @sdbondi described. The proof of work + banning is a better solution to that problem. 

Is there any chance of the channel becoming too full from events? The Listening state is the only receiver I know about. Since it's a broadcast channel, I think it's fine?

sdbondi:
@mikethetike yeah broadcast channels have no backpressure (can never wait on send) and will discard older items if capacity is reached. We'd see `broadcast::error::RecvError::Lagged(n)` errors which we `debug!(...)` and continue in listening state. We may have to do some optimisations in the listener state if CPU is high because that even will trigger a lot more now.

PR #3256, opened 2021-08-28
feat: add Igor testnet

philipr-za:
Description
---
This PR adds support for the Igor testnet to the repo. This involves adding Igor to the Network Enum, adding a Igor generic block and adding a config file with the details of 4 Igor seed nodes (still to be rolled out)

Motivation and Context
---
We need a second testnet to test network switching

How Has This Been Tested?
---
Manually ran the network to generate seed nodes details.


common/config/presets/tari_igor_config.toml
```diff
@@ -0,0 +1,535 @@
+########################################################################################################################
+#                                                                                                                      #
+#                                      The Tari Network Configuration File                                             #
+#                                                                                                                      #
+########################################################################################################################
+
+# This file carries all the configuration options for running Tari-related nodes and infrastructure in one single
```
stringhandler, line 7: Should this not just be merged with the current config, just adding the igor settings?

stringhandler (approved):

PR #3257, opened 2021-08-28
fix: small display bug

Cifko:
Description
---
The escape sequence was eating up the string "Starting recovery at height: ".

How Has This Been Tested?
---
Manually/visually.


sdbondi (approved):
utAck

PR #3258, opened 2021-08-29
chore: tokio 1 and other crate upgrades

sdbondi:
Description
---
- upgrades to tokio 1.10
- upgrade multiaddr to 1.13
- updates select loops to use tokio::select!
- updates to use tokio mpsc and oneshot channels
- remove max_threads config
- removed tari_wallet dependency from tari base node
- moved emoji id library out of tari wallet into tari core (in order to
  remove dependency on `tari_wallet` for tari base node)
- Wait for bootstrap with mempool sync moved to the initializer
- Unit and integration test fixup
- Upgraded following crates that use or are required by tokio 1:
  `bytes`, `prost`, `tonic`, `reqwest`, `hyper`, `trust-dns-client`

~~Include changes from https://github.com/tari-project/tari/pull/3237~~ merged

Motivation and Context
---
Tokio runtime is perhaps the most critical dependency we have and was very out of date (was 0.2.x).
This PR takes advantage of bug fixes and optimisations of tokio 1.10.

How Has This Been Tested?
---
- Existing unit and integration tests run and pass
- Existing cucumber tests pass
- Ran all tari applications (base node, console wallet, miner, mm proxy, stratum transcoder)
- Ran a washing machine test on two upgraded wallets connected to an upgraded base node


applications/tari_app_grpc/src/conversions/com_signature.rs
```diff
@@ -21,7 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use std::convert::TryFrom;
-use tari_crypto::tari_utilities::ByteArray;
+use tari_core::crypto::tari_utilities::ByteArray;
```
SWvheerden, line 24: Why do we change from just straight up using tari_crypto to using the version exported from the tari_core?

applications/tari_base_node/src/main.rs
```diff
@@ -27,6 +27,7 @@
 #![deny(unused_must_use)]
 #![deny(unreachable_patterns)]
 #![deny(unknown_lints)]
+#![allow(dead_code)]
```
SWvheerden: what dead code are we allowing?

sdbondi:
Fixed pesky doc tests + one previously  broken cucumber test (actually still broken - won't check in this PR)

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -26,6 +26,7 @@ use crate::{
         helpers::{mean, median},
     },
 };
+use futures::{channel::mpsc, SinkExt};
```
sdbondi, line 29: Preempting some discussion here - tokio 1.10 mpsc channels do not implement Stream, so using futures channels is easiest - alternative is to use the wrappers provided in `tokio_utils::wrappers`, but this involves less code churn. When/if the Stream trait becomes standardised, it's easier to change the imports than to change the implementation not to use some wrappers.

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -995,32 +995,25 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
-                .iter()
-                .map(|x| x.to_string().as_bytes().to_vec())
-                .collect();
-            response = tari_rpc::SyncInfoResponse {
-                tip_height: info.tip_height,
-                local_height: info.local_height,
-                peer_node_id: node_ids,
-            };
-        }
+        let response = self
```
sdbondi, line 998: Bug fix: the call to `channel.recv().await` would wait for a channel update. - the state machine would typically do this if in listening state every few seconds (chain metadata service drives this). We just want to return whatever is the current state from this GRPC call.

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
```diff
@@ -160,7 +160,7 @@ impl Display for BaseNodeState {
 #[derive(Debug, Clone, PartialEq)]
 pub enum StateInfo {
     StartUp,
-    HeaderSync(BlockSyncInfo),
+    HeaderSync(Option<BlockSyncInfo>),
```
SWvheerden, line 163: why the option here?

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -483,7 +488,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         const COMMIT_EVERY_N_HEADERS: usize = 1000;
 
         // Peer returned no more than the max headers. This indicates that there are no further headers to request.
-        if self.header_validator.valid_headers().len() <= NUM_INITIAL_HEADERS_TO_REQUEST as usize {
+        if self.header_validator.valid_headers().len() < NUM_INITIAL_HEADERS_TO_REQUEST as usize {
```
SWvheerden, line 491: good edge case catch here, but this needs to be a separate PR.

applications/tari_app_utilities/Cargo.toml
```diff
@@ -33,3 +33,6 @@ features = ["transactions"]
 
 [build-dependencies]
 tari_common = {  path = "../../common", features = ["build", "static-application-info"] }
+
+[features]
```
stringhandler, line 37: This shows that there is an incorrect abstraction. This library is meant to house common functionality between the two (and more) apps. If there is code specific to the wallet, it needs to be moved into that app specifically

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -126,54 +127,34 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static
...
             mempool,
             num_synched: Arc::new(AtomicUsize::new(0)),
             permits: Arc::new(Semaphore::new(1)),
-            state_machine,
         }
     }
 
     pub async fn run(mut self) {
         info!(target: LOG_TARGET, "Mempool protocol handler has started");
-        while let Some(ref v) = self.state_machine {
```
SWvheerden, line 144: why remove this? The mempool sync should not run till the node has not finished syncing. It sync but it will reject a lot of transactions if it receives them before the basenode knows of the inputs.

applications/tari_app_utilities/Cargo.toml
```diff
@@ -33,3 +33,6 @@ features = ["transactions"]
...
+wallet = ["tari_wallet"]
```
stringhandler, line 39:
```suggestion
# TODO: This crate is supposed to hold common logic. Move code from this feature into the crate that is more specific to the wallet
wallet = ["tari_wallet"]
```

applications/tari_base_node/src/command_handler.rs
```diff
@@ -189,18 +190,8 @@ impl CommandHandler {
...
-                    info!(
-                        target: LOG_TARGET,
-                        "Error communicating with state machine, channel could have been closed"
-                    );
-                },
-                Some(data) => println!("Current state machine state:\n{}", data),
-            };
-        });
+        let watch = self.state_machine_info.clone();
+        println!("Current state machine state:\n{}", *watch.borrow());
```
stringhandler, line 194: Nice find

stringhandler (dismissed):

applications/tari_app_utilities/Cargo.toml
sdbondi: Yeah does need some sorting, I had finished the work with the base node but could not compile it because it depended on the wallet, so that's why these quick fixes exist - unfortunately because both the type and trait (`From<T>`) is foreign to tari_wallet we cannot just move the code. But we should sort this out.

applications/tari_base_node/src/command_handler.rs
sdbondi: 👍  The new tokio watch api makes it much more obvious when you are waiting for a change vs just getting the value - which is nice.

comms/src/pipeline/mod.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) use inbound::Inbound;
 mod outbound;
 pub(crate) use outbound::Outbound;
 
-mod translate_sink;
-pub use translate_sink::TranslateSink;
+// mod translate_sink;
+// pub use translate_sink::TranslateSink;
```
SWvheerden: remove?

comms/src/pipeline/sink.rs
```diff
@@ -35,22 +35,44 @@ impl<TSink> SinkService<TSink> {
...
+//
+//     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+//         Pin::new(&mut self.0).poll_ready(cx).map_err(Into::into)
+//     }
+//
+//     fn call(&mut self, item: T) -> Self::Future {
+//         let mut sink = self.0.clone();
+//         async move { sink.send(item).await.map_err(Into::into) }.boxed()
+//     }
+// }
```
SWvheerden: remove?

comms/src/transports/tcp.rs
```diff
@@ -25,44 +25,42 @@ use crate::{
...
 pub struct TcpTransport {
-    recv_buffer_size: Option<usize>,
-    send_buffer_size: Option<usize>,
+    // recv_buffer_size: Option<usize>,
+    // send_buffer_size: Option<usize>,
     ttl: Option<u32>,
-    #[allow(clippy::option_option)]
-    keepalive: Option<Option<Duration>>,
+    // #[allow(clippy::option_option)]
+    // keepalive: Option<Option<Duration>>,
```
SWvheerden, line 47: remove?

comms/src/transports/tcp.rs
```diff
@@ -25,44 +25,42 @@ use crate::{
...
-    #[doc("Sets `SO_RCVBUF` i.e the size of the receive buffer.")]
-    setter_mut!(set_recv_buffer_size, recv_buffer_size, Option<usize>);
-
-    #[doc("Sets `SO_SNDBUF` i.e. the size of the send buffer.")]
-    setter_mut!(set_send_buffer_size, send_buffer_size, Option<usize>);
+    // #[doc("Sets `SO_RCVBUF` i.e the size of the receive buffer.")]
+    // setter_mut!(set_recv_buffer_size, recv_buffer_size, Option<usize>);
+    //
+    // #[doc("Sets `SO_SNDBUF` i.e. the size of the send buffer.")]
+    // setter_mut!(set_send_buffer_size, send_buffer_size, Option<usize>);
```
SWvheerden, line 57: remove?

comms/src/transports/tcp.rs
```diff
@@ -25,44 +25,42 @@ use crate::{
...
 
     #[doc("Sets `IP_TTL` i.e. the TTL of packets sent from this socket.")]
     setter_mut!(set_ttl, ttl, Option<u32>);
 
-    #[doc("Sets `SO_KEEPALIVE` i.e. the interval to send keepalive probes, or None to disable.")]
-    setter_mut!(set_keepalive, keepalive, Option<Option<Duration>>);
+    // #[doc("Sets `SO_KEEPALIVE` i.e. the interval to send keepalive probes, or None to disable.")]
+    // setter_mut!(set_keepalive, keepalive, Option<Option<Duration>>);
```
SWvheerden, line 63: remove?

comms/src/transports/tcp.rs
```diff
@@ -81,9 +79,10 @@ impl TcpTransport {
 
     /// Apply socket options to `TcpStream`.
     fn configure(&self, socket: &TcpStream) -> io::Result<()> {
-        if let Some(keepalive) = self.keepalive {
-            socket.set_keepalive(keepalive)?;
-        }
+        // https://github.com/rust-lang/rust/issues/69774
+        // if let Some(keepalive) = self.keepalive {
+        //     socket.set_keepalive(keepalive)?;
+        // }
```
SWvheerden, line 85: remove?

comms/src/transports/tcp.rs
```diff
@@ -93,13 +92,13 @@ impl TcpTransport {
...
-        if let Some(send_buffer_size) = self.send_buffer_size {
-            socket.set_send_buffer_size(send_buffer_size)?;
-        }
+        // if let Some(recv_buffer_size) = self.recv_buffer_size {
+        //     socket.set_recv_buffer_size(recv_buffer_size)?;
+        // }
+        //
+        // if let Some(send_buffer_size) = self.send_buffer_size {
+        //     socket.set_send_buffer_size(send_buffer_size)?;
+        // }
```
SWvheerden, line 101: remove?

comms/src/transports/tcp.rs
```diff
@@ -108,10 +107,10 @@ impl TcpTransport {
 impl Default for TcpTransport {
     fn default() -> Self {
         Self {
-            recv_buffer_size: None,
-            send_buffer_size: None,
+            // recv_buffer_size: None,
+            // send_buffer_size: None,
             ttl: None,
-            keepalive: None,
+            // keepalive: None,
```
SWvheerden, line 113: remove?

comms/src/transports/tcp.rs
```diff
@@ -240,16 +201,15 @@ mod test {
...
-
-        assert_eq!(tcp.send_buffer_size, Some(123));
-        assert_eq!(tcp.recv_buffer_size, Some(456));
+        // tcp.set_send_buffer_size(123)
+        //     .set_recv_buffer_size(456)
+        tcp.set_nodelay(true).set_ttl(789);
+        // .set_keepalive(Some(Duration::from_millis(100)));
+
+        // assert_eq!(tcp.send_buffer_size, Some(123));
+        // assert_eq!(tcp.recv_buffer_size, Some(456));
```
SWvheerden, line 210: remove?

comms/src/transports/tcp.rs
```diff
@@ -240,16 +201,15 @@ mod test {
...
         assert_eq!(tcp.nodelay, Some(true));
         assert_eq!(tcp.ttl, Some(789));
-        assert_eq!(tcp.keepalive, Some(Some(Duration::from_millis(100))));
+        // assert_eq!(tcp.keepalive, Some(Some(Duration::from_millis(100))));
```
SWvheerden, line 213: remove?

SWvheerden (commented):
Looks good. This is much needed. 
I left a few questions here and there. Also, there is a lot of commented-out code. I would suggest either remove it, or if there is a reason you left the code, but a comment with it with the data and the reason so we know why it's commented out and in future, we can't remove or read it.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
sdbondi: https://github.com/tari-project/tari/pull/3264

applications/tari_app_grpc/src/conversions/com_signature.rs
sdbondi: Doesn't matter in this case - feel free to change it if you feel strongly

applications/tari_base_node/src/main.rs
sdbondi: Good catch - just needed to compile in release - will remove

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
sdbondi: Just allows us to show that we're in header sync state before we have all the data - pretty much exactly like BlockSyncStarting that was added while I was busy with this PR, after I rebased it - "header sync but no data yet"

base_layer/core/src/mempool/sync_protocol/mod.rs
sdbondi: TL;DR I moved this to the initializer to decouple the state machine service from this service. 

The state machine has to be passed into this service - in order to not require the state machine in tests, it was made into an option, which is clunky (some future developer: "what does passing None mean? ah well, I'll just pass None and see what happens - ok great it works <moves on>") . This precondition for this service running does not actually have much to do with the service itself - better if we decouple the state machine service from this service and add this check to the initializer as a precondition for this service running.

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

comms/src/transports/tcp.rs
sdbondi: This (and all the others in this file) could come back - tokio doesn't support these socket options at the moment. I'd rather leave them in

base_layer/core/src/mempool/sync_protocol/mod.rs
sdbondi: I had to change some that part of the code as part of the tokio 1 upgrade in any case

stringhandler (approved):

PR #3259, opened 2021-08-30
refactor: refactor wallet ffi cucumber tests

StriderDM:
Description
Refactored WalletFFI.feature into a working state, tested locally.

Further refactoring and dead code removal would be beneficial.

Motivation and Context
Necessary to get WalletFFI.feature working.

How Has This Been Tested?
Tested locally, each scenario tested with:
`./node_modules/.bin/cucumber-js --name "${scenario_name}"`


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2855,7 +2855,7 @@ pub unsafe extern "C" fn wallet_create(
         match TariPrivateKey::from_mnemonic(&(*seed_words).0) {
             Ok(private_key) => Some(private_key),
             Err(e) => {
-                error!(target: LOG_TARGET, "Mnemonic Error for given seed words: {}", e);
+                error!(target: LOG_TARGET, "Mnemonic Error for given seed words: {:?}", e);
```
philipr-za, line 2861: Ideally we would like to use the Display implementation for the Anyhow errors which has the human readable string.

integration_tests/features/WalletFFI.feature
```diff
@@ -1,129 +1,136 @@
...
+        And I have a ffi wallet FFI_WALLET connected to base node BASE
+        And I stop ffi wallet FFI_WALLET
+        And I wait 10 seconds
+        And I send 2000000 uT from wallet SENDER to wallet FFI_WALLET at fee 100
+        And I wait 5 seconds
+        And I restart ffi wallet FFI_WALLET
+        Then I wait for ffi wallet FFI_WALLET to receive 1 transaction
+        Then I wait for ffi wallet FFI_WALLET to receive 1 finalization
+        # Assume tx will be mined to reduce time taken for test, balance is tested in later scenarios.
+        # And mining node MINER mines 10 blocks
```
philipr-za, line 108: Feels like these two commented out lines are needed to actually test the scenario? Why are they omitted?

integration_tests/features/support/world.js
```diff
@@ -106,11 +106,11 @@ class CustomWorld {
     this.walletPubkeys[name] = walletInfo.public_key;
   }
 
-  async createAndAddFFIWallet(name, seed_words) {
+  async createAndAddFFIWallet(name, seed_words = null, passphrase = null) {
```
philipr-za, line 109: FYI I believe the `= null` is superfluous here, that is the default behaviour when an argument is not provided.

delta1: technically `seed_words` and `passphrase` would be `undefined` and not `null` without these default values - but since they are both falsy @philipr-za you can leave these defaults out

StriderDM: The difference between null and undefined is important for how they get used.  It needs to be programmatically empty rather than a reference is non-existent (very subtle difference). 

Have removed them here and done the check deeper in the wrapper though.

base_layer/wallet_ffi/src/lib.rs
StriderDM: Display was flagged as unimplemented, so changed it to debug.

integration_tests/features/WalletFFI.feature
```diff
@@ -1,129 +1,136 @@
...
 
-    Scenario: As a client I want to send Tari to a Public Key
-    # It's a subtest of "As a client I want to retrieve a list of transactions I have made and received"
-
-    Scenario: As a client I want to specify a custom fee when I send tari
-    # It's a subtest of "As a client I want to retrieve a list of transactions I have made and received"
-
-    Scenario: As a client I want to receive Tari via my Public Key while I am online
-    # It's a subtest of "As a client I want to retrieve a list of transactions I have made and received"
+    # It's just calling the encrypt function, we don't test if it's actually encrypted
```
stringhandler, line 6: These tests are so different from the previous file it makes it impossible to know what's changed. Did the order change or are these entirely new tests? Are the previous tests still intact?

stringhandler (approved):
There is so much code that's changed here it's impossible to tell if we've lost tests or code, but it's related to tests, so at worst we've lost coverage. Going to merge it because conflicts will be devastating, but in future, please try to avoid reordering existing code and try to add to files in a way that makes reviewing easier

PR #3260, opened 2021-08-30
fix: prevent immediate run of wallet recovery on cron script

sdbondi:
Description
---
When running cron, the two wallet recovery tests would run immediately
and not be awaited (Promises are eager in js)

Motivation and Context
---
Minor fix

How Has This Been Tested?
---
Script run and tests did not immediately start

### DIFF
--- a/applications/daily_tests/automatic_recovery_test.js
+++ b/applications/daily_tests/automatic_recovery_test.js
@@ -81,17 +81,17 @@ async function run(options = {}) {
       process: wallet.ps,
       outputStream: logfile,
       onData: (data) => {
-        let successLog = data.match(RECOVERY_COMPLETE_REGEXP);
-        let recoveredAmount = data.match(RECOVERY_WORTH_REGEXP);
-        if (successLog && recoveredAmount) {
-          let recoveredAmount = parseInt(recoveredAmount[1]);
-          if (recoveredAmount[2] === "T") {
+        let scannedMatch = data.match(RECOVERY_COMPLETE_REGEXP);
+        let recoveredAmountMatch = data.match(RECOVERY_WORTH_REGEXP);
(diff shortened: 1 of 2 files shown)


applications/daily_tests/automatic_recovery_test.js
```diff
@@ -84,14 +84,14 @@ async function run(options = {}) {
         let successLog = data.match(RECOVERY_COMPLETE_REGEXP);
         let recoveredAmount = data.match(RECOVERY_WORTH_REGEXP);
         if (successLog && recoveredAmount) {
-          let recoveredAmount = parseInt(recoveredAmount[1]);
+          let recoveredAmt = parseInt(recoveredAmount[1]);
```
SWvheerden: nit: I prefer the full typed out name, it's more readable than Amt.

SWvheerden (dismissed):
Looks good to me

applications/daily_tests/automatic_recovery_test.js
sdbondi: It's already used - you cannot rebind variables in js. I will change the variable names to be better

delta1: `let` is block scoped so you should be able to reuse the name if you really want to 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

stringhandler (dismissed):

applications/daily_tests/automatic_recovery_test.js
sdbondi: yeah, but both var names are in this scope :P

sdbondi: Changed it in any case to `recoveredAmountMatch` and `recoveredAmount`

stringhandler (approved):

PR #3262, opened 2021-08-30
fix: revert mining_node default logging config

philipr-za:
Description
---
The mining_node relies on its stdout logging for output for the binary and a recent global update to the logging filtered out the debug and info messages to the std out.

This PR updates the default logging config for the mining node so that debug and info messages are logged to stdout.

How Has This Been Tested?
---
Manually


delta1 (approved):
LGTM, untested

stringhandler (approved):
I suspect there will be errors and warns that are swallowed. I did spent a lot of time finding the sweet spot for logging which is undone here, but the other apps are probably more critical than this one is. If it's wrong we'll see soon enough

philipr-za:
This app doesn't use the majority of the stack. The issue was the majority of the useful miner output was not appearing in stdout.

PR #3264, opened 2021-08-30
fix: off-by-one causing "no further headers to download" bug

sdbondi:
Description
---
When entering the `synchonize_headers` function, a chain of headers has
been downloaded and validated but not committed. If there less than
1000 (not equal to as before), the function can exit without streaming
more as there are no more to send.

This PR correctly handles the case where the node is exactly 1000 headers behind
by: (1) correcting the off-by-one "no further headers to download" conditional and
(2) commiting headers before starting streaming if the PoW is stronger,
in case no further headers would be streamed.

Motivation and Context
---
Header sync ends prematurely when receiving exactly 1000 "pre-sync" headers.

How Has This Been Tested?
---
Manually - Sync from scratch.

### DIFF
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -487,46 +487,53 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
     ) -> Result<(), BlockHeaderSyncError> {
         const COMMIT_EVERY_N_HEADERS: usize = 1000;
 
-        // Peer returned no more than the max headers. This indicates that there are no further headers to request.
-        if self.header_validator.valid_headers().len() < NUM_INITIAL_HEADERS_TO_REQUEST as usize {
-            debug!(target: LOG_TARGET, "No further headers to download");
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -482,46 +482,53 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+        let pending_len = self.header_validator.valid_headers().len();
 
+        // Find the hash to start syncing the rest of the headers.
+        // The expectation cannot fail because the number of headers has been checked in determine_sync_status
+        let (start_header_height, start_header_hash) = self
+            .header_validator
+            .valid_headers()
+            .last()
+            .map(|h| (h.height(), h.hash().clone()))
+            .expect("synchronize_headers: expected there to be at least one valid pending header but there were none");
```
stringhandler: nit: this should be wrapped in a method

stringhandler (dismissed):

delta1 (dismissed):
utACK

stringhandler (approved):

PR #3265, opened 2021-08-30
feat: add ability to bypass rangeproof

stringhandler:
Description
---
Adds the ability to bypass rangeproof verification. 

Motivation and Context
---
Warning: This should not be done by default as it can cause a fork. By default this should always be set to verify rangeproofs, but in some scenarios, you want to disable it to quickly download a chain or run on a slim device. 

The rangeproof verification also takes the majority of time when profiling, so by disabling it, we can monitor other performance bottlenecks

How Has This Been Tested?
---
Manually

> Note that I disabled checking of rangeproofs during wallet sending because it adds little value to validate a rangeproof that you created


sdbondi (dismissed):
utAck - should help with debugging. Though rangeproof construction, rewinding and verification should be run in the blocking threadpool (preferably one spawn does batch processing) and I think there are some cases they aren't (e.g. range proof rewinding for every utxo in one-sided scanning and wallet recovery)

EDIT: also agree with enabling this for possible future light verification client purposes

philipr-za (dismissed):
Looks fine, though a style comment I prefer to try avoid passing around boolean flags like this. The readability of a method that just has a random true/false in the arguments is not great. Perhaps an enum or methods like `new_with_no_rangeproof_verification` or something?

base_layer/core/src/validation/transaction_validators.rs
```diff
@@ -40,17 +40,21 @@ pub const LOG_TARGET: &str = "c::val::transaction_validators";
...
+            factories,
+            bypass_range_proof_verification,
+        }
     }
 }
 
 impl MempoolTransactionValidation for TxInternalConsistencyValidator {
     fn validate(&self, tx: &Transaction) -> Result<(), ValidationError> {
-        tx.validate_internal_consistency(&self.factories, None)
+        tx.validate_internal_consistency(self.bypass_range_proof_verification, &self.factories, None)
```
SWvheerden, line 57: I dont think the mempool should have this option, this leaves the node to propagate bad transactions to the rest of the network.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -51,12 +51,17 @@ pub const LOG_TARGET: &str = "c::val::block_validators";
 #[derive(Clone)]
 pub struct OrphanBlockValidator {
     rules: ConsensusManager,
+    bypass_range_proof_verification: bool,
```
SWvheerden, line 54: This can be potentially dangerous. See main response.

base_layer/core/src/validation/block_validators.rs
```diff
@@ -311,15 +321,17 @@ fn check_mmr_roots<B: BlockchainBackend>(block: &Block, db: &B) -> Result<(), Va
 /// the block body using the header. It is assumed that the `BlockHeader` has already been validated.
 pub struct BlockValidator<B: BlockchainBackend> {
     rules: ConsensusManager,
+    bypass_range_proof_verification: bool,
```
SWvheerden, line 324: This is fine

SWvheerden (commented):
This looks good.  I agree with the base_node option to not check range proofs on blocks as they are "expensive" to create in the fact that you have to mine them first. 

But doing this for the mempool I disagree with. It leaves the node vulnerable to get banned by peers. In the scenario, you know the node does not check the range proof, so you send it a bad range proof. It won't check it, and re propagate the bad transactions to the rest of the network. This means the rest of the nodes will ban that node.  This attack is possible with block propagation as well. 

So I would caution against doing this for block and transaction propagation as it can influence the network security and not just your own local node. But doing this for block sync is fine.

base_layer/core/src/validation/transaction_validators.rs
stringhandler: Yeah I think I agree with this

base_layer/core/src/validation/block_validators.rs
stringhandler: I mentioned that this was dangerous in the description. It's kind of the point of this PR

stringhandler:
@philipr-za can you be more specific, perhaps with an example?

If time allowed, I would have made a struct that had a `bool` member for each validation to be allowed or disallowed, but in some cases there are only rangeproof checks, e.g. sending a transaction. I think in those methods it's actually probably better to just remove the line rather than allow it to be configured.

base_layer/core/src/validation/block_validators.rs
SWvheerden: Yes, but we have 2 distinct cases for dealing with blocks. 
1: `OrphanBlockValidator`, this only deals with propagated blocks
2: `BlockValidator` deals with syncing. 

So in the sync case, if you don't check range proofs, you can get a fork and only you end up on the bad chain. 
but in the propagation case, if you don't check the range proofs you will send out this bad block to your peers and then you can get banned. 

Thats why I suggest to only do this on sync, and or disable the propagation of the block if you don't check range proofs. That way you won't get banned for a bad block.

stringhandler: Yes, I agree with you it is dangerous and should not be run. 

So let's look at the use cases:
1. Fast sync
2. Lite mode clients

While enabling it only during fast sync,  the config setting is misleading and should be renamed to a specific `lite_mode` sync strategy or something like that.

If you take it out of normal operation, you prevent it's use in use case 2. For example, say you have a slow PI that is running this, but does not want to run block validation, for e.g. SPV. This is not possible.

stringhandler: This is really mostly for debugging and profiling. I thought about renaming the config to `debug_bypass_range_proof_verification`, but it really is an advanced feature and should not be toggled with. Also there are other features that are debug only. E.g. `max_difficulty`, `force_sync_peers`, `pruning_horizon at 1`

SWvheerden: I like the idea of naming it `lite_mode` or something. But if we go by bypassing range proof verification for block propagation as well, then should we not disable block propagation on that node. So that that node only asks for blocks, sends out blocks that are asked for. It will broadcast a block it receives from the network. This makes more sense in the lite SPV type mode. An SPV-type lite node should not be relied on for network integrity and the like.

stringhandler: Great you can do that in a follow up PR. This is meant as a debug only feature to measure profiling

sdbondi (approved):

PR #3266, opened 2021-08-30
refactor: enable compile without sqlite, move emoji id and common types to tari_common_types

stringhandler:
Description
---
It moves emoji id and common types to tari_common_types

Motivation and Context
---
The main problem here was a dependency on tari_base_node -> tari_wallet for `EmojiId`. Then EmojiId references PublicKey, so ended up moving a whole bunch around. 

> Note: Hidden in all of this is feature to compile SQLite without having it installed as a lib

How Has This Been Tested?
---
Manually


applications/tari_base_node/src/builder.rs
```diff
@@ -20,9 +20,11 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use crate::bootstrap::BaseNodeBootstrapper;
-use log::*;
 use std::sync::Arc;
+
```
sdbondi, line 24: nit: generally just keep imports together and let rustfmt sort out the order and formatting

sdbondi (commented):
Some untangling is quite badly needed. I would be cautious about splitting the imports, will always create inconsistencies and additional manual labour. This is rustfmt's job. I think we should consider using this rustfmt option https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=import#StdExternalCrate%5C%3A

base_layer/core/src/transactions/types.rs
```diff
@@ -21,97 +21,10 @@
...
-
-/// The type of output that `Challenge` produces
-pub type MessageHash = Vec<u8>;
+use tari_crypto::ristretto::dalek_range_proof::DalekRangeProofService;
 
 /// Specify the range proof type
 pub type RangeProofService = DalekRangeProofService;
 
 /// Specify the range proof
 pub type RangeProof = BulletRangeProof;
```
SWvheerden: Is there a reason why you left these to here and not move them over to comment::types as well?
`MAX_RANGE_PROOF_RANGE` is moved to common, so I would expect that to live with the other rangeproof types.

SWvheerden (commented):
Looks good, I only have one question.

applications/tari_base_node/src/builder.rs
stringhandler: Yeah I think this was caused by intellij's refactor. Will move it manually next time

base_layer/core/src/transactions/types.rs
stringhandler: I originally thought it had much deeper integration in core, but I've managed to merge bullet_proof into common_types as well

sdbondi (approved):

hansieodendaal (approved):
LGTM - tested the sqlite3 dependency for compile-time and runtime in Windows

PR #3267, opened 2021-08-30
fix: make logging less noisy

stringhandler:
Description
---
Remove logging of errors from tracing instrument macros.

Motivation and Context
---
Was reported as making the base node unusable. Hopefully we are not swallowing important information, but probably the right choice

How Has This Been Tested?
---
Manually

> ~~Note: This PR is based on #3266 to enable compilation without SQLite installed~~

### DIFF
--- a/comms/src/connection_manager/dialer.rs
+++ b/comms/src/connection_manager/dialer.rs
@@ -338,7 +338,7 @@ where
     }
 
     #[allow(clippy::too_many_arguments)]
-    #[tracing::instrument(skip(peer_manager, socket, conn_man_notifier, config, cancel_signal), err)]
+    #[tracing::instrument(skip(peer_manager, socket, conn_man_notifier, config, cancel_signal))]
     async fn perform_socket_upgrade_procedure(
         peer_manager: Arc<PeerManager>,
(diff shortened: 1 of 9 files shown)


philipr-za (dismissed):
Reviewed the `make logging of errors less noisy` commit and it looks good to me. 

I think this is a case that dialing is a behaviour that is expected to fail under normal operation which would mean its not a show stopping error that deserves to appear in the main UI. Should definitely still be visible in the logs/tracing though.

stringhandler:
Yeah I think the best situation would be logging the error higher up in situations where you definitely want the dial to succeed. In most cases the extra logging was due to the `err` member in the `#[instrument()]` attribute, which is very useful, but you can't specify the logging level

SWvheerden (dismissed):
LGTM,
I only reviewed the last commit as the rest looks like a duplicate of PR#3266

hansieodendaal (dismissed):
LGTM (not tested)
Reviewed the `make logging of errors less noisy` commit

stringhandler:
Cherry picked to reduce noise

SWvheerden (approved):
LGTM

sdbondi (approved):

PR #3270, opened 2021-08-31
fix: auto update continuously checks auto_update_check_interval is disabled

sdbondi:
Description
---
Continuously checks for updates when auto_update_check_interval is disabled.

Thanks @mikethetike. for finding it and for the fix 

Add check to if no auto update URIs are configured 

Motivation and Context
---
Bug fix

When check_interval is disabled, stream::empty() is used to disable the update checking, however it
returns None continuously when polled, causing the update to continuously be checked.

Also sets MissedTickBehaviour::Skip - which will prevent bursts of checks if intervals are missed

How Has This Been Tested?
---
Ran base node with auto_update_check_interval = 0 (or equivalently without this setting set)

### DIFF
--- a/base_layer/p2p/src/auto_update/service.rs
+++ b/base_layer/p2p/src/auto_update/service.rs
@@ -25,16 +25,18 @@ use crate::{
     auto_update::{AutoUpdateConfig, SoftwareUpdate, Version},
 };
 use futures::{future::Either, stream, StreamExt};
+use log::*;
 use std::{env::consts, time::Duration};
 use tari_common::configuration::bootstrap::ApplicationType;
 use tari_service_framework::{async_trait, ServiceInitializationError, ServiceInitializer, ServiceInitializerContext};
 use tokio::{
     sync::{mpsc, oneshot, watch},
     time,
+    time::MissedTickBehavior,
 };
 use tokio_stream::wrappers;
 
-const LOG_TARGET: &str = "app:auto-update";
+const LOG_TARGET: &str = "p2p::auto_update";
 
 /// A watch notifier that contains the latest software update, if any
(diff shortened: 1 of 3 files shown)


hansieodendaal (commented):
Suggest changing
```
const LOG_TARGET: &str = "app:auto-update";
```
to
```
const LOG_TARGET: &str = "app::auto_update";
```
as the former does not allow fine-tuning with log4rs

stringhandler:
> Suggest changing
> 
> ```
> const LOG_TARGET: &str = "app:auto-update";
> ```
> 
> to
> 
> ```
> const LOG_TARGET: &str = "app::auto_update";
> ```
> 
> as the former does not allow fine-tuning with log4rs

In fact, this file differs from the other files in the folder, which start with `p2p::auto_update`

base_layer/p2p/src/auto_update/service.rs
```diff
@@ -25,16 +25,18 @@ use crate::{
...
 use tari_service_framework::{async_trait, ServiceInitializationError, ServiceInitializer, ServiceInitializerContext};
 use tokio::{
     sync::{mpsc, oneshot, watch},
     time,
+    time::MissedTickBehavior,
 };
 use tokio_stream::wrappers;
 
-const LOG_TARGET: &str = "app:auto-update";
+const LOG_TARGET: &str = "app::auto_update";
```
stringhandler:
```suggestion
const LOG_TARGET: &str = "p2p::auto_update";
```

stringhandler (dismissed):

stringhandler (approved):

hansieodendaal (approved):
LGTM - tested

PR #3275, opened 2021-08-31
fix: update cucumber tests for walletffi.feature

StriderDM:
Description
Made types more explicit for ffi interface and easier to maintain going forwards.
Initialized InterfaceFFI in world rather than WalletFFIClient.
Updated steps in WalletFFI.feature
Updated some const variables in wrapper classes to be let.
Better handling of callback function pointer variables in wallet.
Added additional steps to tests.
Downgrade code to be compatible with nodejs v12.

Motivation and Context

How Has This Been Tested?
nvm use v12.22.6 && ./node_modules/.bin/cucumber-js features/WalletFFI.feature

Run log:
[run.log](https://github.com/tari-project/tari/files/7097096/run.log)

NodeJS version downgrade:
https://github.com/tari-labs/docker/pull/16

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,9 +127,15 @@ commands:
           name: Generate report
           command: cd integration_tests && node ./generate_report.js
           when: always
+      #- run:
+      #    name: Set the correct Node version for wallet FFI tests
+      #    shell: bash -l {0}
+      #    command: nvm install v12.22.6 && nvm use v12.22.6 && cd integration_tests && npm install
+      #    when: always
       - run:
-          name: Run ffi cucumber scenarios
+          name: Run FFI wallet library cucumber scenarios
           command: cd integration_tests && mkdir -p cucumber_output && node_modules/.bin/cucumber-js --tags "not @long-running and not @broken and not @flaky and @wallet-ffi" --format json:cucumber_output/tests_ffi.cucumber --exit
+          when: always
       - run:
           name: Generate report (ffi)
           command: cd integration_tests && node ./generate_report.js "cucumber_output/tests_ffi.cucumber" "temp/reports/cucumber_ffi_report.html"
--- a/integration_tests/features/support/steps.js
+++ b/integration_tests/features/support/steps.js
@@ -13,6 +13,7 @@ const {
   consoleLogBalance,
   consoleLogTransactionDetails,
   withTimeout,
+  waitForIterate,
 } = require("../../helpers/util");
 const { ConnectivityStatus, PaymentType } = require("../../helpers/types");
 const TransactionBuilder = require("../../helpers/transactionBuilder");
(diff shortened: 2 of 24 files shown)


.circleci/config.yml
```diff
@@ -129,7 +129,7 @@ commands:
           when: always
       - run:
           name: Run ffi cucumber scenarios
-          command: cd integration_tests && mkdir -p cucumber_output && node_modules/.bin/cucumber-js --tags "not @long-running and not @broken and not @flaky and @wallet-ffi" --format json:cucumber_output/tests_ffi.cucumber --exit
+          command: cd integration_tests && mkdir -p cucumber_output && touch cucumber_output/tests_ffi.cucumber && node_modules/.bin/cucumber-js --tags "not @long-running and not @broken and not @flaky and @wallet-ffi" --format json:cucumber_output/tests_ffi.cucumber --exit
```
stringhandler:
```suggestion
          command: cd integration_tests && mkdir -p cucumber_output && touch cucumber_output/tests_ffi.cucumber && node_modules/.bin/cucumber-js --tags "not @long-running and not @broken and not @flaky and @wallet-ffi" --format json:cucumber_output/tests_ffi.cucumber --exit
          when: always
```

integration_tests/features/support/steps.js
```diff
@@ -3826,15 +3826,15 @@ Then(
...
-      },
-      true,
-      700 * 1000,
-      5 * 1000,
-      5
-    );
+    let count = 0;
+
+    while (!(wallet.getCounters().received >= amount)) {
+      await sleep(1000);
```
stringhandler: Why is this better?

integration_tests/features/support/steps.js
```diff
@@ -3860,15 +3860,15 @@ Then(
...
-      },
-      true,
-      700 * 1000,
-      5 * 1000,
-      5
-    );
+    let count = 0;
+
+    while (!(wallet.getCounters().finalized >= amount)) {
+      await sleep(1000);
```
stringhandler: Same question

stringhandler (commented):
What's wrong with the waitFor method that it needs to be replaced. Wouldn't it be better to fix those?

.circleci/config.yml
stringhandler: This runs the FFI test even if the previous step failed

StriderDM: Thanks.

integration_tests/features/support/steps.js
StriderDM: Wrote the one like that and updated the others so they were all all running the same code, that's the only reason.

Would you like me to change it back to waitFor for all of them?

stringhandler: yes, unless there is a reason to not use waitFor?

stringhandler (dismissed):
utACK

stringhandler (approved):

PR #3285, opened 2021-09-02
fix: resolved design flaw in wallet_ffi library

StriderDM:
Description
Exposed TariTransactionKernel instead of TariExcess, TariExcessSignature and TariExcessPublicNonce. This was done as they are all referring to fields from the same object and also to reduce the amount of wrapper classes needs on the other side of the FFI boundary.

Added accessor methods to retrieve data from the opaque pointer for TariTransactionKernel and return the data as a type (in this instance a cstring) which can safely cross the FFI boundary.

Updated library header.
Updated rust tests.
Updated cucumber tests.
Updated comments.
Incremented library version.

Merge https://github.com/tari-project/tari/pull/3275 first.

Motivation and Context
---

How Has This Been Tested?
cargo test --all --all-features
nvm use v12.22.6 && ./node_modules/.bin/cucumber-js features/WalletFFI.feature

Screen Shot 2021-09-02 at 5 08 56 PM


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -265,6 +263,115 @@ pub unsafe extern "C" fn string_destroy(ptr: *mut c_char) {
...
+/// The ```string_destroy``` method must be called when finished with a string from rust to prevent a memory leak
+#[no_mangle]
+pub unsafe extern "C" fn transaction_kernel_get_excess_signature_hex(
+    kernel: *mut TariTransactionKernel,
+    error_out: *mut c_int,
+) -> *mut c_char {
+    let mut error = 0;
+    ptr::swap(error_out, &mut error as *mut c_int);
+    let mut result = CString::new("").unwrap();
+    ptr::swap(error_out, &mut error as *mut c_int);
```
philipr-za, line 344: duplicate line

StriderDM: Nice catch, thanks @philipr-za. Fixed.

stringhandler (approved):

PR #3295, opened 2021-09-05
feat: ping-peer command

sdbondi:
Description
---
`ping-peer [emoji id | pk | nodeid ]`

Motivation and Context
---
Allows latency tests pinging of peers using the liveness service

How Has This Been Tested?
---
Manually by running the command on a base node

### DIFF
--- a/applications/tari_base_node/src/command_handler.rs
+++ b/applications/tari_base_node/src/command_handler.rs
@@ -59,8 +59,15 @@ use tari_core::{
     tari_utilities::{hex::Hex, message_format::MessageFormat},
 };
 use tari_crypto::{ristretto::RistrettoPublicKey, tari_utilities::Hashable};
-use tari_p2p::auto_update::SoftwareUpdaterHandle;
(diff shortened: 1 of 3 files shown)


applications/tari_base_node/src/command_handler.rs
```diff
@@ -535,6 +544,42 @@ impl CommandHandler {
         });
     }
 
+    pub fn ping_peer(&self, dest_node_id: NodeId) {
+        let mut liveness = self.liveness.clone();
+
+        self.executor.spawn(time::timeout(Duration::from_secs(30), async move {
```
philipr-za, line 550: Maybe just a message if it hits the timeout?

sdbondi: Yeah first prize would be to improve liveness service to allow for a ping/pong result to be returned to a caller - for now, I think a message here will work - will add now

stringhandler (approved):
Nice, this will be useful

PR #3300, opened 2021-09-06
fix: dedup sql error when deleting many entries

sdbondi:
Description
---
Fixes ERROR Error when trimming message dedup cache:
ResultError(DatabaseError(__Unknown, "too many SQL variables"))

Motivation and Context
---
Bug fix when trimming many rows. Cleanup performance will also be better

How Has This Been Tested?
---
Existing test pass


stringhandler (approved):
Nice

PR #3304, opened 2021-09-06
feat: improve console wallet responsiveness

hansieodendaal:
Description
---
1. Improved console wallet responsiveness by introducing a [**edit** _debouncer with a_] cooldown period for calculating the balance, as this is a database call that grows linearly with the amount of UTXOs in the wallet and does not need to be repeated multiple times per second in certain scenarios. [**Edit** _All events that can change the balance and are captured in the wallet event monitor are followed by the `trigger_balance_refresh` command that will send an event to the debouncer, which in turn will collect all such events and only perform one balance update database query after the end of the cooldown period._] ~~A new base node state change (metadata) event will ensure the last request to calculate balance will be fulfilled.~~
2. Submit connectivity failed event if the metadata RPC requests time out so that 'Waiting for data...' will be displayed in the wallet.
3. Removed the RPC client ping call as a method when monitoring the base node to calculate latency and instead replaced it with the time measured to obtain metadata via the RPC client. 

Motivation and Context
---
1. Wallets with growing numbers of UTXOs become more and more sluggish due to repeated database calls to calculate the balance; this is much more pronounced when many broadcast monitoring protocols are running.
2. When the metadata RPC requests time out previous base node state would be displayed; this could carry on for 30 minutes or more with consecutive timeouts where the correct user feedback would be that the wallet is waiting for data.
3. The time measured for the ping call always follows the time it takes for the metadata call, which is evident as well when the base node gets busy and will often take longer than the metadata call. Having two longish RPC calls with no added value results in more RPC request timeouts than is necessary. See time measurement plots below:

   image
   image
   image
   image
   image

How Has This Been Tested?
---
System-level testing during a stressed coin split


base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -95,27 +106,37 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
...
                 .metadata
                 .ok_or_else(|| BaseNodeMonitorError::InvalidBaseNodeResponse("Tip info no metadata".to_string()))
                 .and_then(|metadata| {
                     ChainMetadata::try_from(metadata).map_err(BaseNodeMonitorError::InvalidBaseNodeResponse)
                 })?;
+            // Note: Do not use an RPC ping call to measure latency here, as the time measured for that always follows
+            // the time it takes for the metadata call, which is evident as well when the base node gets busy and will
+            // often take longer than the metadata call itself. Having two longish RPC calls with no added value
+            // results in more RPC request timeouts than is necessary.
+            let latency = Instant::now().duration_since(start);
```
sdbondi:
```suggestion
            let latency = start.elapsed();
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -441,6 +449,9 @@ pub struct AppStateInner {
     updated: bool,
     data: AppStateData,
     wallet: WalletSqlite,
+    balance_enquiry_cooldown_start: Instant,
+    balance_enquiry_cooldown_infringement: bool,
```
philipr-za: This is a over complication of a simple feature. Adding a lot of code to just produce a trace log of when this occurs seems unnecessary.

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -642,9 +658,30 @@ impl AppStateInner {
...
+            self.data.balance = balance;
+            self.updated = true;
+        } else {
+            self.balance_enquiry_cooldown_infringement = true;
+        }
+
+        Ok(())
+    }
+
+    pub async fn refresh_balance_check(&mut self) -> Result<(), UiError> {
```
philipr-za: I am struggling to understand the purpose of this?

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -160,6 +160,7 @@ impl WalletEventMonitor {
                                 match (*msg).clone() {
                                     BaseNodeEvent::BaseNodeStateChanged(state) => {
                                         self.trigger_base_node_state_refresh(state).await;
+                                        self.trigger_balance_refresh_check().await;
```
philipr-za: I don't understand what this is for. This is not an event that would result in the balance changing?

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -262,4 +263,12 @@ impl WalletEventMonitor {
         let mut inner = self.app_state_inner.write().await;
         inner.add_notification(notification);
     }
+
+    async fn trigger_balance_refresh_check(&mut self) {
```
philipr-za, line 267: I feel this is not needed.

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -95,27 +106,37 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
...
+            let start = Instant::now();
             let tip_info = client.get_tip_info().await?;
-
             let chain_metadata = tip_info
                 .metadata
                 .ok_or_else(|| BaseNodeMonitorError::InvalidBaseNodeResponse("Tip info no metadata".to_string()))
                 .and_then(|metadata| {
                     ChainMetadata::try_from(metadata).map_err(BaseNodeMonitorError::InvalidBaseNodeResponse)
                 })?;
+            // Note: Do not use an RPC ping call to measure latency here, as the time measured for that always follows
```
philipr-za: So my understanding of the RPC ping message is that it is supposed to be a measure of the round trip response without a database read so it is a better measure of network latency that should be divorced from the business of the base node database. If the metadata call and this are linked then that actually points to a problem on the base node that the RPC Pong response is somehow being blocked by Database activity.

common/config/presets/tari_config_example.toml
```diff
@@ -178,6 +178,8 @@ scan_for_utxo_interval = 180
 base_node_service_refresh_interval = 30
 # The maximum age of service requests in seconds, requests older than this are discarded
 base_node_service_request_max_age = 180
+# The cooldown period between balance enquiry checks in seconds, requests faster than this will be ignored
+balance_enquiry_cooldown_period = 60
```
philipr-za: 1 minute is huge!

common/src/configuration/utils.rs
```diff
@@ -100,6 +100,7 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
     cfg.set_default("wallet.base_node_query_timeout", 60).unwrap();
     cfg.set_default("wallet.base_node_service_refresh_interval", 5).unwrap();
     cfg.set_default("wallet.base_node_service_request_max_age", 60).unwrap();
+    cfg.set_default("wallet.balance_enquiry_cooldown_period", 5).unwrap();
```
philipr-za: I can buy 5 seconds

philipr-za (commented):
In principle I am ok with a cooldown on balance refresh however there is an issue because balance refresh is only triggered on events that would update the balance so what happens if the balance is updated 10 times in quick succession and the last trigger happens within a cooldown period. The balance will never be updated to reflect the change of that last event.

applications/tari_console_wallet/src/ui/state/app_state.rs
philipr-za: Is this in place to combat the edge case I mentioned in the review?

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
philipr-za: If this is for the edge case then this should rather be in the on_tick event

applications/tari_console_wallet/src/ui/state/app_state.rs
hansieodendaal: The purpose here is not a trace log. As you have observed, all requests in the cooldown period after the last accepted request will be ignored. To ensure that the last request in a cooldown period is processed if no more requests are made after that, we need a separate event to trigger such a check, such that the last ignored request in a cooldown period will be handled.

To summarize:
- regular calls to calculate the balance ends up with `refresh_balance()`;
- `refresh_balance()` will only perform the balance calculation if there is no cooldown period infringement and clear the `balance_enquiry_cooldown_infringement` flag, otherwise, it will set the `balance_enquiry_cooldown_infringement` flag;
- `refresh_balance_check()`, which runs in a separate thread with the regular `BaseNodeEvent::BaseNodeStateChanged`, will call upon `refresh_balance()` if the `balance_enquiry_cooldown_infringement` flag has been set.

applications/tari_console_wallet/src/ui/state/app_state.rs
hansieodendaal: Please see above

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
hansieodendaal: Please see above

philipr-za: I still think this should be in the on_tick event of the TUI. Here you are linking this to Base Node Event which has nothing to do with balance.

base_layer/wallet/src/base_node_service/monitor.rs
hansieodendaal: Really difficult to make sense of this. I could pick this up in the logs, and these may be from the same conversation or not but it is difficult to tell:

``` rust
CW:    00:15:41.694385500 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain RPC client 0 ms
   BN: 00:15:43.309834400 [comms::rpc] DEBUG [Peer=`ff03297bea5f3f548a85dd72c6`] Got request RequestID=12, Deadline=120s, Flags=(empty), Message=0 byte(s) 
CW:    00:16:56.568149800 [comms::rpc::client] TRACE Received response (64 byte(s)) from request #12 (protocol = t/bnwallet/1, method=5) in 75s
CW:    00:16:56.568175700 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain metadata from client 74873 ms
CW:    00:16:56.568225000 [comms::rpc::client] DEBUG Ping (protocol t/bnwallet/1) sent in 3.00µs
   BN: 00:16:58.072925300 [comms::rpc] TRACE RPC TIMING(REQ_ID=3): 'transmit' took 53.64s (SLOW)
   BN: 00:16:58.073036500 [comms::rpc] DEBUG [Peer=`ff03297bea5f3f548a85dd72c6` t/bnwallet/1] sending ACK response.
CW:    00:18:23.682751100 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain latency (ping) from client 87114 ms
   BN: 00:18:25.297436900 [comms::rpc] TRACE RPC TIMING(REQ_ID=3): 'transmit' took 85.17s (SLOW) 
```

What stands out is:
- the wallet measures 74.873s for a round trip metadata request; the base node reports a slow transmit of 53.64s
- the wallet measures 87.114s for a round trip ping request; the base node reports a slow transmit of 85.17s

If those timings are linked it would suggest being not blocked by database activity.

common/config/presets/tari_config_example.toml
hansieodendaal: Agreed

common/src/configuration/utils.rs
hansieodendaal: Lowering it to 1 s

base_layer/wallet/src/base_node_service/monitor.rs
sdbondi: Thanks - All the other places where `duration_since` was used too

applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
hansieodendaal: Agreed, done.

base_layer/wallet/src/base_node_service/monitor.rs
hansieodendaal: Ah.. missed those, thanks.

common/config/presets/tari_config_example.toml
```diff
@@ -178,6 +178,10 @@ scan_for_utxo_interval = 180
 base_node_service_refresh_interval = 30
 # The maximum age of service requests in seconds, requests older than this are discarded
 base_node_service_request_max_age = 180
+# The cooldown period between balance enquiry checks in seconds; requests faster than this will be ignored.
+# For specialized wallets processing many batch transactions this setting could be increased to 60 s to retain
+# responsiveness of the wallet with slightly delayed balance updates (default: 1)
+#balance_enquiry_cooldown_period = 1
```
philipr-za: I honestly still don't think this belongs here. It is soooooo niche as to be noise. @mikethetike thoughts?

common/config/presets/tari_igor_config.toml
```diff
@@ -178,6 +178,10 @@ scan_for_utxo_interval = 180
 base_node_service_refresh_interval = 30
 # The maximum age of service requests in seconds, requests older than this are discarded
 base_node_service_request_max_age = 180
+# The cooldown period between balance enquiry checks in seconds; requests faster than this will be ignored.
+# For specialized wallets processing many batch transactions this setting could be increased to 60 s to retain
+# responsiveness of the wallet with slightly delayed balance updates (default: 1)
+#balance_enquiry_cooldown_period = 1
```
philipr-za: as above

common/config/presets/tari_config_example.toml
hansieodendaal: Removed from the default config as proposed

common/config/presets/tari_igor_config.toml
hansieodendaal: Removed from the default config as proposed

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -141,6 +149,13 @@ impl AppState {
         Ok(())
     }
 
+    pub async fn refresh_balance_check(&self) -> Result<(), UiError> {
+        let mut inner = self.inner.write().await;
+        inner.refresh_balance_check().await?;
+        drop(inner);
```
sdbondi: nit: It is unidiomatic/unnecessary to explicitly drop variables at the end of a function body (it drops on line 157)

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -95,27 +106,38 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
...
+            // the time it takes for the metadata call, which is evident as well when the base node gets busy and will
+            // often take longer than the metadata call itself. Having two longish RPC calls with no added value
+            // results in more RPC request timeouts than is necessary.
+            // TODO: Find root cause of RPC ping calls taking approximately as long as RPC fetch meta data calls
+            let latency = start.elapsed();
 
-            let latency = client.ping().await?;
```
sdbondi, line 118: Think we are picking up differences across machines - in my tests the real ping was lower when for e.g. the base node was sending blocks to another base node and/or getting scanned by another wallet etc. .

Whereas, because get_tip_info requires the db, it would be slower in this case. The value presented as latency is not purely network latency and could be misleading. There are, as you'd expect, some cases where network performance takes a hit (and on your and some others machines it seems a lot worse). 

I did mention in my original PR that latency for get_tip_info would be more efficient and a better measure of actual base node performance, once we see a truer network latency value (which was instantly better in my case), but happy if we keep in mind the fact that database contention is included in the value presented in the console wallet as latency.

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -642,9 +665,25 @@ impl AppStateInner {
     }
 
     pub async fn refresh_balance(&mut self) -> Result<(), UiError> {
-        let balance = self.wallet.output_manager_service.get_balance().await?;
-        self.data.balance = balance;
-        self.updated = true;
+        if Instant::now().duration_since(self.balance_enquiry_cooldown_start) >
```
sdbondi:
```suggestion
        if self.balance_enquiry_cooldown_start.elapsed() >
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -450,13 +468,18 @@ impl AppStateInner {
         wallet: WalletSqlite,
         base_node_selected: Peer,
         base_node_config: PeerConfig,
+        balance_enquiry_cooldown_period: u64,
```
sdbondi: This should be a Duration type
```suggestion
        balance_enquiry_cooldown_period: Duration,
```

sdbondi (dismissed):
We should try not to create endless toml configuration that are meaningless to anyone that does not have intimate knowledge of the code. If we feel it necessary to tweak values endlessly, we should probably take a step back and think about a better way. 

The call to refresh_balance now does not necessarily refresh the balance. If the balance is being refreshed too often by some caller, perhaps the caller should handle that (debounce) correctly or mark the balance as stale so that it can be updated by the next tick (I think the latter is a good option). This is preferable to manually tweaking some values and creating unexpected side (non)effects in refresh_balance.

Otherwise, fine with the removal of ping. This can go in, but we should strive to create systems that work for users, rather than ones that have to be configured by engineers to work.

base_layer/wallet/src/base_node_service/monitor.rs
hansieodendaal: > ... happy if we keep in mind the fact that database contention is included in the value presented in the console wallet as latency.

Agreed, a proper comment in the code highlight this.

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -643,9 +665,23 @@ impl AppStateInner {
     }
 
     pub async fn refresh_balance(&mut self) -> Result<(), UiError> {
-        let balance = self.wallet.output_manager_service.get_balance().await?;
-        self.data.balance = balance;
-        self.updated = true;
+        if self.balance_enquiry_cooldown_start.elapsed() > self.balance_enquiry_cooldown_period {
```
stringhandler: There is no need for an additional method here, and the names are confusing. I would suggest removing the `refresh_balance_check` method and the `balance_enquiry_cooldown_infringement` variables as calling the `refresh_balance` method twice in quick succession has the same effect

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -643,9 +665,23 @@ impl AppStateInner {
...
+            self.balance_enquiry_cooldown_start = Instant::now();
+            self.balance_enquiry_cooldown_infringement = false;
```
stringhandler:
```suggestion
            
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -643,9 +665,23 @@ impl AppStateInner {
...
+            let balance = self.wallet.output_manager_service.get_balance().await?;
+            self.data.balance = balance;
+            self.updated = true;
+        } else {
```
stringhandler:
```suggestion
       }
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -643,9 +665,23 @@ impl AppStateInner {
...
+            self.balance_enquiry_cooldown_infringement = true;
```
stringhandler:
```suggestion
  
```

applications/tari_console_wallet/src/ui/state/app_state.rs
stringhandler: This can be removed. See above

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -95,27 +106,38 @@ impl<T: WalletBackend + 'static> BaseNodeMonitor<T> {
...
                 .metadata
                 .ok_or_else(|| BaseNodeMonitorError::InvalidBaseNodeResponse("Tip info no metadata".to_string()))
                 .and_then(|metadata| {
                     ChainMetadata::try_from(metadata).map_err(BaseNodeMonitorError::InvalidBaseNodeResponse)
                 })?;
+            // Note: Do not use an RPC ping call to measure latency here, as the time measured for that always follows
+            // the time it takes for the metadata call, which is evident as well when the base node gets busy and will
+            // often take longer than the metadata call itself. Having two longish RPC calls with no added value
+            // results in more RPC request timeouts than is necessary.
+            // TODO: Find root cause of RPC ping calls taking approximately as long as RPC fetch meta data calls
```
stringhandler: This seems more like a comment for a PR than a comment that is helpful for future developers. I would suggest removing it

common/src/configuration/global.rs
```diff
@@ -532,6 +533,15 @@ fn convert_node_config(
         .map(|seconds| seconds as u64)
         .map_err(|e| ConfigurationError::new(&key, &e.to_string()))?;
 
+    // The cooldown period between balance enquiry checks in seconds; requests faster than this will be ignored.
+    // For specialized wallets processing many batch transactions this setting could be increased to 60 s to retain
+    // responsiveness of the wallet with slightly delayed balance updates (default: 1)
+    let key = "wallet.balance_enquiry_cooldown_period";
```
stringhandler, line 539: nobody is ever going to change this to anything other than 1. I would say hardcoding it as 1 second is better in this case

stringhandler (changes requested):
I understand why you want to add this, but the implementation is so specific and adds so many extra tracking variables to achieve it that the extra orchestration code muddies up the other code. 

If you can find a way to achieve the same goal in a more elegant manner, I'd consider merging it

applications/tari_console_wallet/src/ui/state/app_state.rs
hansieodendaal: Implemented a balance check debouncer for these purposes - please see the latest commit

common/src/configuration/global.rs
hansieodendaal: The default is set to 1. The configuration setting is needed for stress testing or for running a large wallet where many events occur that can change the balance, where 1 s is not a good choice. These are average time measurements for the call to the `pub async fn get_balance` SQL queries to complete:
- ~15,000 total spent and unspent TXOs: 1 s
- ~125,000 total spent and unspent TXOs: 400 s

applications/tari_console_wallet/src/ui/state/debouncer.rs
```diff
@@ -0,0 +1,139 @@
...
+                balance.pending_outgoing_balance
+            );
+            let mut inner = self.app_state_inner.write().await;
+            if let Err(e) = inner.refresh_balance(balance).await {
+                warn!(target: LOG_TARGET, "Error refresh app_state: {}", e);
+            }
+        }
+        loop {
+            tokio::select! {
+                _ = &mut delay => {
```
stringhandler, line 83: If the desired behaviour is to cache the result for multiple calls within the cooldown period, this is not being achieved.

What will happen here is that many calls to this method will get slower and slower, blocking the callers.

Wouldn't it be better to change it to something like 

```

select! {
  balance_enquiry_events.recv() =>  {
   if  Instant::now() < cache_until_time { return cached_value }
  // get value
   cache_until_time = Instant::now() + self.balance_enquiry_cooldown_period;
   return value;
 }
  
}
```

stringhandler (approved):
Not 100% convinced, but the impact should be low. I think I can craft a simpler debouncer as I stated in the comments, but this can go in for now

applications/tari_console_wallet/src/ui/state/debouncer.rs
hansieodendaal: > If the desired behaviour is to cache the result for multiple calls within the cooldown period, this is not being achieved.

The desired behaviour is to postpone all calls to update the balance for a set cooldown period of time till the very end of the period (and of course then only perform one calculation) and not update it at the beginning of the period and use the cached value. This will ensure that the last request to update the balance will always be honoured if no more requests arrive after the current cooldown period. If the cooldown period is made much bigger for example increased to 60 s for large busy wallets, then not honouring the last balance update request in the cooldown period is very much noticeable in the TUI with displaying the incorrect balance.

> What will happen here is that many calls to this method will get slower and slower, blocking the callers.

If 1000 requests are sent to the debouncer (`broadcast`) channel in 1 s then only the last one will be remembered as it has size 1; all prior requests will be dropped. The select loop will only check for a new update request after the delay event, and the channel receive future will timeout if there is none. Thus no blocking calls.

Issue #3302, opened 2021-09-06
Tari Base Node doesn't run on Ubuntu 18.04


stringhandler:
logged via telegram group:

```
Samuel Harrison, [06.09.21 09:17]
Tari base node doesnt run on ubuntu 18.04 pc's

Samuel Harrison, [06.09.21 09:17]
It requires the latest glibc drivers of ubuntu 20.4 and above

Samuel Harrison, [06.09.21 09:19]
Tari devs, please remember backward compatibility when compiling
```

When running the binary

delta1:
Likely since we are building on 20.04, will attempt to change the build process to use 18.04

delta1:
error building croaring on 18.04 due to old clang version (6) - putting in a PR to change to `clang-10`

delta1:
Should be resolved by #3305 will test

PR #3305, opened 2021-09-06
ci: run builds and ci on ubuntu 18.04

delta1:
Description
---
Runs builds and tests on 18.04 to ensure we can still support it. EOL is April 2023 

Motivation and Context
---
#3302 

How Has This Been Tested?
---
github actions, and started built base node on ubuntu 18.04

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -7,7 +7,7 @@ on:
     branches:
       - ci-*
   schedule:
-    - cron: '05 00 * * *'
+    - cron: "05 00 * * *"
   workflow_dispatch:
     inputs:
       customTag:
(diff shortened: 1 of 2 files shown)


.github/workflows/base_node_binaries.yml
```diff
@@ -269,17 +269,20 @@ jobs:
...
-        with:
-          artifacts: "tari_*/**/*"
-          token: ${{ secrets.GITHUB_TOKEN }}
-          prerelease: true
-          draft: true
+# "The workflow is not valid. .github/workflows/base_node_binaries.yml (Line: 273, Col: 9): Unrecognized named-value: 'env'. Located at position 1 within expression: env.AWS_SECRET_ACCESS_KEY != '' ...
+  # create-release:
+  #   runs-on: ubuntu-latest
+  #   needs: builds
+  #   if: ${{ env.AWS_SECRET_ACCESS_KEY != '' && matrix.os != 'self-hosted' && startsWith(github.ref, 'refs/tags/v') }}
```
stringhandler: Should be this. The workflow is valid otherwise, but I want it to only run on tags

```suggestion
  #   if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```

stringhandler: Please uncomment the section and apply this change. Otherwise I'm happy to merge

delta1: Sure, I only commented it out since it broke the workflow

delta1: done as 2 commits

stringhandler (approved):

.github/workflows/base_node_binaries.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macos-10.15, windows-2019]
+        os: [ubuntu-18.04, macos-10.15, windows-2019]
```
StriderDM, line 27: Wondering if we shouldn't have both 18.04 and 20.04 for ubuntu in here.

PR #3308, opened 2021-09-07
fix: reduce overly-eager connection reaping for slow connections

sdbondi:
Description
---
Connection reaping has been improved for use with slower connections (network IO or slow thread scheduling):
- A connection is counted as used as soon as a request for a substream
  has been started.
- The minimum age of a connection before it can be reaped changed from 1
  minute to 20 minutes
- Reduces the refresh interval for pool refresh from every 30s to every 60s
- Set `MissedTickBehaviour::Delay` for refresh interval, as bursting is
  undesired

- fixes the stress_test example
- fix (rare) flakiness for `dial_cancelled` comms test

Motivation and Context
---
Connections should not be considered unused while they are attempting to establish a substream. This typically takes < 1 second over tor, however a slow network and/or thread starvation can cause substream establishment to take many seconds.
Previously, the connection would be counted as unused for that period, resulting in reaping while a substream is being established.

image

How Has This Been Tested?
---
Basic base node test 

image

### DIFF
--- a/comms/examples/stress/node.rs
+++ b/comms/examples/stress/node.rs
@@ -38,6 +38,7 @@ use tari_comms::{
     NodeIdentity,
     Substream,
 };
+use tari_shutdown::ShutdownSignal;
 use tari_storage::{
     lmdb_store::{LMDBBuilder, LMDBConfig},
     LMDBWrapper,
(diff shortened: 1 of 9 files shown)


philipr-za (approved):
Sounds like a possible cause for some of the weird behaviour where a connection is dropped while the higher layers think it is being established.

stringhandler (approved):

PR #3309, opened 2021-09-07
chore: fix 3 cucumber tests

Cifko:
Description
---
The test were failing, because the automated TXs were required, which should not be the case.

How Has This Been Tested?
---
npm test -- --name "Duplicate block is rejected"
npm test -- --name "Node should not sync from pruned node"
npm test -- --name "Wallets monitoring normal transactions after a reorg"

### DIFF
--- a/integration_tests/features/Sync.feature
+++ b/integration_tests/features/Sync.feature
@@ -103,15 +103,15 @@ Feature: Block Sync
     When I mine 15 blocks on PNODE2
     Then all nodes are at height 23
 
-  @broken
   Scenario: Node should not sync from pruned node
     Given I have a base node NODE1 connected to all seed nodes
-    Given I have a pruned node PNODE1 connected to node NODE1 with pruning horizon set to 5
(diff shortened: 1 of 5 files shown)


integration_tests/features/Sync.feature
```diff
@@ -103,10 +103,10 @@ Feature: Block Sync
     When I mine 15 blocks on PNODE2
     Then all nodes are at height 23
 
-  @broken
   Scenario: Node should not sync from pruned node
-    Given I have a base node NODE1 connected to all seed nodes
-    Given I have a pruned node PNODE1 connected to node NODE1 with pruning horizon set to 5
+    Given I do not expect all automated transactions to succeed
+    And I have a base node NODE1 connected to all seed nodes
+    And I have a pruned node PNODE1 connected to node NODE1 with pruning horizon set to 5
```
SWvheerden, line 108: Look at what I did here: you can turn it on and off again. It can be on here, and only turned off later in the test, and turned on again later

SWvheerden (approved):
LGTM

stringhandler (approved):

PR #3313, opened 2021-09-08
fix: always grow database when asked to resize

sdbondi:
Description
---
When a resize is detected as needed, we no longer need to do threshold
checking. Simply grow the database by the configured amount.

Increase the number of MAX_RESIZES in a transaction to accomodate very
large transactions (interestingly, needed during rewind-blockchain for >
1000 blocks, depending on configured resize growth).

- Fix incorrect error log in lmdb helper functions.
- Check for MAP_FULL in lmdb_replace

Motivation and Context
---
Bug fixes

How Has This Been Tested?
---
using rewind-blockchain on base node


philipr-za (approved):
LGTM

PR #3314, opened 2021-09-08
chore: update headers requests to use vec::with_capacity

SWvheerden:
Description
All the current comms requests for headers just a `Vec::new()` rather then a `Vec::with_capacity()`

Motivation and Context
We know the size of the result or can make a calculated guess about the size of the result so we can avoid resizing the vector the whole time by just using with_capacity. 

How Has This Been Tested?
Run all unit tests.


sdbondi (approved):
LGTM - most of these functions allow remote requests to allocate large/almost unbounded amounts of memory. I'm not sure how many of these requests are still in use vs have been replaced with RPC calls, so they should be reviewed and removed / check upper bounds of the number of blocks requested.

stringhandler (approved):

PR #3315, opened 2021-09-08
fix: wait couple rounds for no pings to send an event

Cifko:
Description
---
Wait couple rounds of `no ping` before sending the event.

How Has This Been Tested?
---
npm test -- --name "Full block sync with small reorg"
npm test -- --name "Pruned mode reorg simple"
npm test -- --name "Pruned mode reorg past horizon"

### DIFF
--- a/base_layer/core/src/base_node/chain_metadata_service/service.rs
+++ b/base_layer/core/src/base_node/chain_metadata_service/service.rs
@@ -43,12 +43,15 @@ use tari_comms::{
 use tari_p2p::services::liveness::{LivenessEvent, LivenessHandle, Metadata, MetadataKey};
 use tokio::sync::broadcast;
 
+const NUM_ROUNDS_NETWORK_SILENCE: u16 = 3;
+
 pub(super) struct ChainMetadataService {
     liveness: LivenessHandle,
     base_node: LocalNodeCommsInterface,
     peer_chain_metadata: Vec<PeerChainMetadata>,
     connectivity: ConnectivityRequester,
     event_publisher: broadcast::Sender<Arc<ChainMetadataEvent>>,
+    number_of_rounds_no_pings: u16,
 }
 
 impl ChainMetadataService {
@@ -69,6 +72,7 @@ impl ChainMetadataService {
             peer_chain_metadata: Vec::new(),
             connectivity,
             event_publisher,
+            number_of_rounds_no_pings: 0,
         }
     }
 
@@ -159,6 +163,7 @@ impl ChainMetadataService {
                     "Received ping from neighbouring node '{}'.",
                     event.node_id
                 );
+                self.number_of_rounds_no_pings = 0;
                 self.collect_chain_state_from_ping(&event.node_id, &event.metadata)?;
                 self.send_chain_metadata_to_event_publisher().await?;
             },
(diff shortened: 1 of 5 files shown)


base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -178,7 +182,11 @@ impl ChainMetadataService {
                     target: LOG_TARGET,
                     "New chain metadata round sent to {} peer(s)", num_peers
                 );
-                self.send_chain_metadata_to_event_publisher().await?;
+                // If there were no pings for awhile, we are probably alone.
+                self.number_of_rounds_no_pings += 1;
+                if self.number_of_rounds_no_pings >= 10 {
```
sdbondi: need to check that there are no pings (i.e. num_peers == 0)

sdbondi: Also suggest 3 rounds (~ 1 minute to decide on network silence) - the ping interval is currently set to 30s by default in the base node, so 10 will take too long 
```rust
const NUM_ROUNDS_NETWORK_SILENCE: u16 = 3; // updated from 5
```

```suggestion
                if self.number_of_rounds_no_pings >= NUM_ROUNDS_NETWORK_SILENCE {
```

Cifko: That's one of the questions I had prepared. If there are peers it should trigger the pings so the counter should be nullified.
Or should I nullify the counter when the num_peers are not zero? There are several ways to do this.
Should I nullify the `number_of_rounds_no_pings` once we call the `send_chain_metadata_to_event_publisher` or not.
i.e. should we send it everytime (since the limit is reached) or just from time to time?

sdbondi: Ah ok I see what you did, makes sense. I think that we should add a NetworkSilence event that gets emitted _every time_ there are `n` rounds in a row that have no responses before the next one starts. The reason is, because if we get pings, then stop getting pings `send_chain_metadata_to_event_publisher` will carry on sending old metadata from previous rounds whereas NetworkSilence is more explicit. Obviously the listener will need to handle this event in the same way that it handles `PeerChainMetadataReceived([])` (which will never be broadcast after this change). What do you think?

Cifko: Yeah, the NetworkSilence sounds really good. I've added the `ChainMetadataEvent::NetworkSilence` event.

SWvheerden: I also think we should reduce the `NUM_ROUNDS_NETWORK_SILENCE ` to be around 3. That's basically 90 seconds of zero inbound or outbound comms. 10 rounds is around 10 minutes. That wont work for cucumber. But 90 seconds might not either. We might want to make this a config option and for cucumber make it like 10 seconds or so.

SWvheerden (dismissed):
I like this approach so far. 

My only comment is about the number of rounds and the time delay.

stringhandler (dismissed):

base_layer/core/src/base_node/chain_metadata_service/service.rs
Cifko: It's 15 in the tests, 30 in the default_config in utils.rs (both for igor and weatherwax), Still 90 seconds is enough. I've change it to 3, and refactored it to the top as a constant.

stringhandler (approved):

PR #3319, opened 2021-09-08
[DO NOT MERGE] feat: add grpc support for seed words

Cifko:
Description
---
Add support for getting seed words via GRPC call.

Motivation and Context
---
I needed the seed words of a running wallet.

How Has This Been Tested?
---
I have small python GRPC client.

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -54,6 +54,8 @@ service Wallet {
     rpc ListConnectedPeers(Empty) returns (ListConnectedPeersResponse);
     // Cancel pending transaction
     rpc CancelTransaction (CancelTransactionRequest) returns (CancelTransactionResponse);
+    // Get seed words
+    rpc GetSeedWords(Empty) returns (GetSeedWordsResponse);
 }
 
 message GetVersionRequest { }
@@ -196,4 +198,8 @@ message CancelTransactionRequest {
 message CancelTransactionResponse {
     bool is_success = 1;
     string failure_message = 2;
+}
+
+message GetSeedWordsResponse {
+    repeated string seed_words = 1;
 }
--- a/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -437,6 +437,21 @@ impl wallet_server::Wallet for WalletGrpcServer {
             },
         }
     }
+
+    async fn get_seed_words(
+        &self,
+        _: Request<tari_rpc::Empty>,
+    ) -> Result<Response<tari_rpc::GetSeedWordsResponse>, Status> {
+        debug!(target: LOG_TARGET, "Incoming gRPC request to Get Seed Words");
+        let seed_words = self
+            .get_output_manager_service()
+            .get_seed_words()
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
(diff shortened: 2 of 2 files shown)


philipr-za (commented):
I feel this is a pretty large security risk. The seed words are able to generate every secret this wallet will ever use and this will make that accessible to any process on the same machine.

Cifko:
> I feel this is a pretty large security risk. The seed words are able to generate every secret this wallet will ever use and this will make that accessible to any process on the same machine.

Yeah, that's true. The main purpose was mainly for my convenience. We don't have to merge this. Putting a password as an input is also an option. Not super secure, but better. There are plenty of options to change the UI of the wallet to get this information, but that's overkill for debugging. In real life the user should have the seed words written down somewhere.

delta1:
@cifko there is a command line arg to write the seed-words to file `--seed-words-file-name` would that not suffice? 

I'm of 2 minds, because yes this is a security risk, but you can already use the unsecured grpc interface to send all the funds in the wallet. 

Also, for any other UI that would use grpc we would actually want to expose this (once grpc is authed in some way) 

🤷‍♂️

Cifko:
> @Cifko there is a command line arg to write the seed-words to file `--seed-words-file-name` would that not suffice?

To be honest I forgot we have that 🤣. 
Anyway, this is not our priority now. So will just leave this as do not merge. And we can return to this sometime in the future.

philipr-za:
> I'm of 2 minds, because yes this is a security risk, but you can already use the unsecured grpc interface to send all the funds in the wallet.

Yes, though just to discuss the relative degree of risk here. On one hand you can lose all the funds in your wallet, but the seed words actually represent all the keys that will ever be generated in the future too. So that being exposed is fundamentally a larger risk. This is partly why in all the crypto apps out there you get to see your seed words once and exactly once because of how sensitive they are.

delta1:
> Yes, though just to discuss the relative degree of risk here. On one hand you can lose all the funds in your wallet, but the seed words actually represent all the keys that will ever be generated in the future too. So that being exposed is fundamentally a larger risk. 

True, good point. 

> This is partly why in all the crypto apps out there you get to see your seed words once and exactly once because of how sensitive they are. 

Anecdotally, in my Green wallet I can retrieve the seed words at any time. I do think it should be possible (once authorized) to get your seeds words, since when a wallet is new and empty there's really no need to have them, but once it has some funds it's important to get them. The staged wallet security RFC also recommends this: 

> Stage zero - zero balance
> When the user has a zero balance, there's no risk in letting them skip securing their wallet.
> Therefore, Tari wallets SHOULD just skip the whole seed phrase ritual and let the user jump right into the action.

> Stage 1a - a reminder to write down your seed phrase
> Once the user's balance exceeds the MINIMUM_STAGE_ONE_BALANCE, they will be prompted to review and write down their seed phrase. The MINIMUM_STAGE_ONE_BALANCE is any non-zero balance by default. 

https://github.com/tari-project/tari/blob/02c5582a3c51d225a98bd8e935804ecf163b9217/RFC/src/RFC-0151_StagedWalletSecurity.md

Cifko:
> > Stage 1a - a reminder to write down your seed phrase
> > Once the user's balance exceeds the MINIMUM_STAGE_ONE_BALANCE, they will be prompted to review and write down their seed phrase. The MINIMUM_STAGE_ONE_BALANCE is any non-zero balance by default.

In our case the minimum stage can be one yat as well. Or any NFT in the future.

stringhandler:
I'm going to close this. idk if we still need the functionality. @Cifko if you feel strongly you can open it again

PR #3321, opened 2021-09-08
fix: handle stream read error case by explicitly closing the substream

sdbondi:
Description
---
- adds yamux stream id to logs to enrich rpc sesssion tracing info
- handle stream read error case by explicitly closing the substream
- updates rpc tests to use yamux

Motivation and Context
---
May allow us to diagnose where slowness in a RPC session/substream occurs 

How Has This Been Tested?
---
Tests updated


stringhandler (dismissed):
yes please

sdbondi:
ignored broken doc test (now needs yamux setup, previously a only MemorySocket was required)

stringhandler (approved):

PR #3334, opened 2021-09-10
fix: remove unnecessary range proof verify and fix test temp disk usage 

stringhandler:
Description
---
Remove an unnecessary rangeproof verify after we just created it

Note: Also reduced the size impact of LMDB temp databases that `cargo test` has when running on Windows machines.

Motivation and Context
---
There is no benefit in verifying a rangeproof immediately after creating it

How Has This Been Tested?
---


base_layer/core/src/transactions/transaction.rs
```diff
@@ -295,12 +295,12 @@ impl UnblindedOutput {
             sender_offset_public_key: self.sender_offset_public_key.clone(),
             metadata_signature: self.metadata_signature.clone(),
         };
-        // A range proof can be constructed for an invalid value so we should confirm that the proof can be verified.
-        if !output.verify_range_proof(&factories.range_proof)? {
-            return Err(TransactionError::ValidationError(
-                "Range proof could not be verified".into(),
-            ));
-        }
+        // // A range proof can be constructed for an invalid value so we should confirm that the proof can be verified.
```
sdbondi: > `// range proof can be constructed for an invalid value so we should confirm that the proof can be verified.`

Assuming the invalid value talked about here is a negative value? Something that is already an invariant because Microtari is backed by a u64.

sdbondi (commented):
Nice - maybe @philipr-za can confirm that this is sound and then we can remove the commented out code.

philipr-za (commented):
It does seem a bit paranoid to check the proof right after its been constructed with the same data. It came from the fact that you build the proof using the unblinded data and verify it with the commitment and the factory that produces the commitment here could be different to the commitment factory used within the tari_crypto crate during proof construction.

philipr-za (dismissed):

philipr-za (commented):
Approved, though I agree if the code is coming out then don't leave the comments in.

hansieodendaal (dismissed):
I agree with this change as the entire transaction including the range proof will be verified multiple times before being submitted to the mempool.

stringhandler:
Fixed tests by adding a simple check instead of running the full rangeproof validation. @philipr-za @hansieodendaal please have a look again

base_layer/core/src/transactions/transaction.rs
```diff
@@ -281,6 +282,12 @@ impl UnblindedOutput {
     }
 
     pub fn as_transaction_output(&self, factories: &CryptoFactories) -> Result<TransactionOutput, TransactionError> {
+        if factories.range_proof.range() < 64 && self.value > MicroTari::from(1u64.shl(&factories.range_proof.range()))
```
stringhandler: I think this is correct as opposed to `shl(....range() - 1)`

sdbondi: `factories.range_proof.range() < 64` will never be true in our case since we set the range to 64. 
`1u64 << 64` [will overflow](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a44daa637b2c86d392624a76d9653d98) so should be `range() - 1`.

stringhandler: it's actually true for 90% of the tests, that's why it's there. My reasoning is that if it is 64, then a u64 automatically fits in. 

The reason that this is here is because there are some tests that use a smaller range of 32. I'll have to check and see if `range() - 1` is correct in this case

philipr-za: so just trying `1u64.shl(64)` panics with `thread 'main' panicked at 'attempt to shift left with overflow'`.

1u64(63) will give you half the 64-bit range. 

If you are looking for the max value of an arbitrary bit depth you will need to shift up that many ones.

stringhandler: This is correct now that I've added the `=` to the `>=`. The case where the range is 64 is covered simply by the fact that you cannot include a higher value inside a u64.

philipr-za: I guess the short circuit protects us from the overflow

philipr-za (approved):

PR #3336, opened 2021-09-12
feat: rpc response message chunking

sdbondi:
Description
---
Adds server-side "chunking" protocol for large RPC responses.

Given a payload exceeding some threshold, comms RPC will split message payloads into a number (maximum 16) 
RPC messages and stream them to the client. The client will recombine the message payloads before emitting it to
the caller.  

BREAKING CHANGE: Backward compatible (client-side), an upgraded node may request from an older node and process
responses. Since an older node cannot process chunks, if a response contains any, the response will be invalid to the older node. Otherwise, if less than the payload threshold, the response is exactly as before. 

Motivation and Context
---
Large streamed message payloads (those seen in block sync) are often ~ 1.2 MiB over the wire. 
The [Bandwidth Delay product](https://www.wikiwand.com/en/Bandwidth-delay_product) is often low/very low for connections over tor resulting in large message frames will take many seconds to be completely received. 

This PR breaks message payloads into smaller chunks with less time between receiving a full frame on the client side. 

The algorithm for chunking is as follows:
```
THRESHOLD = 256kb 
SIZE_LIMIT = 384Kb
MAX_CHUNKS = 16
```
a. `payload <= THRESHOLD`: emit a single frame
b. `payload > THRESHOLD && payload <= SIZE_LIMIT `: emit a single frame containing the full payload
c. `payload > SIZE_LIMIT `: emit a one or more frames containing the next `SIZE_THRESHOLD` bytes, with the final frame either as per (a) or (b).

This is done to prevent an unnecessary/very small trailing chunk when a payload is only a handful of bytes over the threshold
e.g 256Kb (`THRESHOLD`) + 1 byte results in _one_ chunk not _two_.  384Kb (`SIZE_LIMIT`) + 1 kb results 
in 2 frames of 256Kb and 129Kb respectively. meaning each chunk has a maximum size of `SIZE_LIMIT` and the
trailing/last chunk has a minimum size of `SIZE_LIMIT - THRESHOLD  (128kb)`

How Has This Been Tested?
---

Some new unit tests, existing large message tests, manual archival sync using force sync

### DIFF
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -1003,11 +1003,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
             .state_info
             .get_block_sync_info()
             .map(|info| {
-                let node_ids = info
-                    .sync_peers
-                    .iter()
-                    .map(|x| x.to_string().as_bytes().to_vec())
-                    .collect();
+                let node_ids = info.sync_peers.iter().map(|x| x.to_string().into_bytes()).collect();
                 tari_rpc::SyncInfoResponse {
                     tip_height: info.tip_height,
                     local_height: info.local_height,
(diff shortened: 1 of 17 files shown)


stringhandler (approved):
Tested locally and works very well

PR #3338, opened 2021-09-13
feat: create initial asset checkpoint for console wallet

StriderDM:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/applications/tari_console_wallet/src/automation/command_parser.rs
+++ b/applications/tari_console_wallet/src/automation/command_parser.rs
@@ -61,6 +61,7 @@ impl Display for ParsedCommand {
             ClearCustomBaseNode => "clear-custom-base-node",
             RegisterAsset => "register-asset",
             MintTokens => "mint-tokens",
+            CreateInitialCheckpoint => "create-initial-checkpoint",
         };
 
         let args = self
@@ -131,6 +132,7 @@ pub fn parse_command(command: &str) -> Result<ParsedCommand, ParseError> {
         RegisterAsset => parser_builder(args).text().build()?,
         // mint-tokens pub_key nft_id1 nft_id2
         MintTokens => parser_builder(args).pub_key().text_array().build()?,
+        CreateInitialCheckpoint => parser_builder(args).pub_key().text().build()?,
     };
 
     Ok(ParsedCommand { command, args })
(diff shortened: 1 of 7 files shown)


applications/tari_console_wallet/src/automation/command_parser.rs
```diff
@@ -61,6 +61,7 @@ impl Display for ParsedCommand {
             ClearCustomBaseNode => "clear-custom-base-node",
             RegisterAsset => "register-asset",
             MintTokens => "mint-tokens",
+            InitialAssetCheckpoint => "create-initial-checkpoint",
```
stringhandler: Let's keep these names the same otherwise it gets confusing
```suggestion
            CreateInitialCheckpoint => "create-initial-checkpoint",
```

base_layer/core/src/transactions/transaction/mod.rs
```diff
@@ -183,6 +183,17 @@ impl OutputFeatures {
             ..Default::default()
         }
     }
+
+    pub fn for_checkpoint(merkle_root: Vec<u8>, asset_public_key: PublicKey) -> OutputFeatures {
+        Self {
+            flags: OutputFlags::ASSET_CHECKPOINT,
+            asset: Some(AssetOutputFeatures {
```
stringhandler: This feature should only be used for AssetRegistration. The `parent_public_key` is the field that should be used. The `unique_id` field should also be populated with a specific value. Perhaps [0000....000] or [11111...111]?
```suggestion
            asset: None
```

base_layer/core/src/transactions/transaction/mod.rs
```diff
@@ -229,6 +240,7 @@ bitflags! {
         // TODO: separate these flags
         const ASSET_REGISTRATION = 0b0000_1010; // Registration and also non-fungible
         const MINT_NON_FUNGIBLE = 0b0000_1100; // Mint and non-fungible
+        const ASSET_CHECKPOINT = 0b0000_1101;
```
stringhandler:
```suggestion
        const SIDECHAIN_CHECKPOINT = 0b0000_1101;
```

stringhandler (commented):
Looks good, some minor points

stringhandler (approved):

PR #3340, opened 2021-09-14
docs: rfc 0310 asset implementation

stringhandler:
WIP of RFC for Asset Implementation


RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
+# RFC-0310/AssetImplementation
```
delta1, line 1: Needs to be added to the summary/menu

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+## Asset Implementation Details
+
+![status: draft](theme/images/status-draft.svg)
+
+Maintainer(s): [mikethetike](https://github.com/mikethetike)
```
delta1, line 7: minor: I think the other RFCs use full name instead of handle

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+# Licence
+
+[ The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright <YEAR> <COPYRIGHT HOLDER | The Tari Development Community>
```
delta1: placeholders

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+## Disclaimer
+
+This document and its content are intended for information purposes only and may be subject to change or update
+without notice.
+
+This document may include preliminary concepts that may or may not be in the process of being developed by the Tari
+community. The release of this document is intended solely for review and discussion by the community regarding the
+technological merits of the potential system outlined herein.
+
+## Abstract
```
delta1: empty?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+## Related Requests for Comment
```
delta1, line 49:
```suggestion
## Related Requests for Comment

None.
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+## Proposal
+
+### Changes to outputs
+The following fields must be included in the TransactionOutput, inside the OutputFeatures:
+
+New flags to be added to OutputFlags
+
+| Name | Value | Description |
+| --- | --- | --- |
+| NON_FUNGIBLE | 0b1000_1000 | This UTXO contains Non Fungile data and must not be combined with other fungible UTXOs |
```
delta1: s/Fungile/Fungible/

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+| ASSET_REGISTRATION | 0b1100_1000 | This UTXO contains registration information for a new asset |
+| MINT |0b0100_0000 | This UTXO represents the creation of a new NFT, and is used during the  |
```
delta1: During the?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+| BURN | 0b0010_0000 | The `unique_id` in this UTXO can be spent as a normal input and must not appear in the output set unless accompanied by a `MINT` flag |
+| SIDECHAIN_CHECKPOINT | 0b1001_1000 | This UTXO is a checkpoint for a sidechain | 
+
+
+New fields added to `OutputFeatures`
+
+Note: this replaces some information in [RFC-0311](RFC-0310_AssetImplementation.md)
```
delta1: link is wrong

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+or the `ASSET_REGISTRATON` flag must also be set.
+
+Note: If the `parent_public_key` is populated and the `ASSET_REGISTRATION` flag is set, it must be treated as an asset registration 
+instruction. This allows sub assets to be grouped under a parent asset.
+
+**New Consensus Rule**
+
+When receiving a UTXO with the `MINT` and `NON_FUNGIBLE` flags set and the `ASSET_REGISTRATION` unset, the `minting_proof_signature`
+must be a valid commitment signature using the commitment of the UTXO that contains the currently valid asset registration feature
+for with the `unique_id` = `parent_public_key`. The message signed must be `Hash(commitment|script|unique_id|parent_public_key)`
```
delta1: "for with the" may be missing a word

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+and is only destroyed once the UTXO is spent.
+
+**New Consensus Rule**
+When receiving a transaction or block with an input with the `BURN` flag set, there must not exist a UTXO in the output set containing the same `unique_id`,`parent_public_key` pair, unless 
+that output has the `MINT` flag set.
+
+### Sidechain Checkpoints
+
+A sidechain checkpoint is a special type of NFT UTXO. Extra sidechain data can be specified in the `sidechain` field of the output features. Like 
+other NFT's, only one sidechain checkpoint UTXO with that unique_id can exist in the unspent set at one time. The conditions of
```
delta1:
```suggestion
other NFTs, only one sidechain checkpoint UTXO with that unique_id can exist in the unspent set at one time. The conditions of 
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+When receiving a transaction or block with an input with the `BURN` flag set, there must not exist a UTXO in the output set containing the same `unique_id`,`parent_public_key` pair, unless 
+that output has the `MINT` flag set.
+
+### Sidechain Checkpoints
+
+A sidechain checkpoint is a special type of NFT UTXO. Extra sidechain data can be specified in the `sidechain` field of the output features. Like 
+other NFT's, only one sidechain checkpoint UTXO with that unique_id can exist in the unspent set at one time. The conditions of 
+who can spend a checkpoint are determined by the TariScript on the checkpoint. In most cases it will be an `n of m` multisig condition 
+using the members of the Hotstuff committee, with `m` equal to the number of members of the committee and `n` equal to `m` minus 
+the number of failures the committee can tolerate.
```
delta1, line 178: not major - but I think `m of n` is in wider use than `n of m`

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+The unique_id can be a specificly chosen id, but can also be a random public key.
```
delta1:
```suggestion
The unique_id can be a specifically chosen id, but can also be a random public key. 
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+using the members of the Hotstuff committee, with `m` equal to the number of members of the committee and `n` equal to `m` minus 
+the number of failures the committee can tolerate.
+
+The unique_id can be a specificly chosen id, but can also be a random public key. 
+
+There is no limit to the number of different sidechain checkpoint tokens an asset has, but it will usually be one. 
+
+## New Base Node GRPC methods
+
+A base node should expose the following methods on via GRPC:
```
delta1:
```suggestion
A base node should expose the following methods via GRPC:
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+message TokenStatePair {
+  string name = 1;
+  bytes value = 2;
+}
+```
+
+## Checkpoints
+
+As part of the Commit phase, a validator node includes in its vote, a signature of the current state and instruction set
+in `locked_qc` that can be used as part of the threshold signature to spend the current checkpoint. This sign the state of the chain before the current instruction proposal is executed. Before
```
delta1:
```suggestion
in `locked_qc` that can be used as part of the threshold signature to spend the current checkpoint. This signs the state of the chain before the current instruction proposal is executed. Before
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+blocks in the sidechain have been created since the last checkpoint. If so, the leader assembles a transaction
+spending the previous checkpoint using the signatures it has obtained. 
+
+## Examples
+
+### NFT with sidechain metadata 
+
+Use case: I want to issue an NFT that has some immutable metadata and some mutable metadata.
+
+In  this example, let's pretend we are making a simple game YachtClicker in
```
delta1:
```suggestion
In this example, let's pretend we are making a simple game called YachtClicker in 
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+### NFT with sidechain metadata 
+
+Use case: I want to issue an NFT that has some immutable metadata and some mutable metadata.
+
+In  this example, let's pretend we are making a simple game YachtClicker in 
+which every time a user clicks on a Yacht, it increases in XP. The user can change the name of the yacht, 
+but only the asset issuer can award XP based on the clicks reported in the app. 
+
+Template used: [EditableMetadata](linktbd)
```
delta1: todo: linktbd

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+            "issuer_proof": "<com_sig of h(commitment |script) using asset registration commitment>"
+         },
+         "metadata": []
+      },
+      "script": "Nop"
+   }
+}
+```
+
+Note that this must be caller from Asset Issuer's wallet so that it can create the commitment signature.
```
delta1:
```suggestion
Note that this must be called from Asset Issuer's wallet so that it can create the commitment signature. 
```

RFC/src/RFC-0312_AssetTemplateEdtitableMetadata.md
```diff
@@ -0,0 +1,54 @@
+# RFC-0312-AssetTemplateEditableMetadaa
```
delta1, line 1: add to summary/menu

RFC/src/RFC-0312_AssetTemplateEdtitableMetadata.md
```diff
@@ -0,0 +1,54 @@
...
+
+![status: raw](theme/images/status-raw.svg)
+
+**Maintainer(s)**: [mikthetike](https://github.com/mikethetike)
+
+# Licence
+
+[ The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright <YEAR> <COPYRIGHT HOLDER | The Tari Development Community>
```
delta1: placeholders

RFC/src/RFC-0312_AssetTemplateEdtitableMetadata.md
```diff
@@ -0,0 +1,54 @@
+# RFC-0312-AssetTemplateEditableMetadaa
+
+## EditableMetadata Template - For simple assignment of metadata
+
+![status: raw](theme/images/status-raw.svg)
+
+**Maintainer(s)**: [mikthetike](https://github.com/mikethetike)
```
delta1, line 7: same minor point as above

delta1 (commented):
Looking good, found a few spelling/grammar points but the approach seems good to me

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: I don't want to add my full name if I can avoid it

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+| SIDECHAIN_CHECKPOINT | 0b1001_1000 | This UTXO is a checkpoint for a sidechain | 
+
+
+New fields added to `OutputFeatures`
+
+Note: this replaces some information in [RFC-0311](RFC-0310_AssetImplementation.md)
+
+| Name | Type | Description |
+| --- | --- | --- |
+| unique_id | bytes(32) | A unique id representing a token. This value must be present only when the NON_FUNGIBLE flag is set. The pair (parent_pub_key, unique_id) will be unique across the unspent set as described below |
```
sdbondi, line 82:
```suggestion
| unique_id | bytes(32) | A unique id representing a token. This value MUST be present only when the NON_FUNGIBLE flag is set. The pair (parent_pub_key, unique_id) will be unique across the unspent set as described below |
```

RFC/src/RFC-0310_AssetImplementation.md
CjS77: Mr Tike would prefer not to be doxxed

RFC/src/RFC-0310_AssetImplementation.md
SWvheerden: Perhaps we should add a cross-reference table to make it easy to see which fields are required for what flags.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+| parent_pub_key | bytes(32) |  A namespacing field that also allows a heirachy of assets. In most cases this will be the asset that a token belongs to, but future uses may be possible as well |
+| asset_registration | struct | Data for registering of an asset. This value **MUST** be present if ASSET_REGISTRATION flag is set |
```
SWvheerden, line 84: I am taking this as meaning that for every other flag, the data MUST not be present? Perhaps we can clarify this. A cross-reference table should make this easy.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+| asset_registration.name | bytes(64) | utf8 name of the asset. Optional |
+| asset_registration.template_ids | Vec<bytes(32)> |The templates that this asset exposes. Usually, at least one will be specified |
+| asset_registration.creator_sig | bytes(64) (32 Nonce + 32 sig) | A signature of the UTXO's `commitment` and `script` using the public key in `unique_id` to prove this registration was created in this UTXO |
+| asset_registration.checkpoint_unique_id | bytes(32) | A reference to the unique id reserved for checkpoint UTXOs. Optional |
+| asset_registration.checkpoint_frequency | uint32 | The frequency, in sidechain blocks (or other measure of time the sidechain uses) in which checkpoints are created |
+| mint.issuer_proof | bytes(96) (64 nonce + 32 sig) | A _ComSig_ proof proving that the owner of the UTXO containing the asset registration created this token
```
SWvheerden:
```suggestion
| mint.issuer_proof | bytes(96) (32 nonce + 64 sig) | A _ComSig_ proof proving that the owner of the UTXO containing the asset registration created this token 
```
The nonce in a comsig is a Pedersen commitment, so only 32 bytes. But the sig is 2 scalars, hence 32 + 32 = 64.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,605 @@
...
+
+
+Note: `unique_id` is not always required to be a public key
+
+### New Base Layer Consensus rules
+
+**New Consensus Rule**
+
+If a UTXO is received in the mempool or block with the `MINT` bit set in the flags, the containing transaction or block is valid only if 
+no other UTXO exists in the unspent set with the same `unique_id` and `parent_public_key`.
```
SWvheerden:
```suggestion
no other UTXO exists in the unspent set with the same `unique_id`,  `parent_public_key` pair. 
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+    // A comsig proving the caller owns the commitment in `caller_commitment`
+    // The challenge is the hash of the above fields, in order
+    bytes caller_sig = 6;
+    
+    // Fee fields TBD
+}
+
+```
+
+Upon receiving an instruction, a validator node must store it in its mempool and broadcast it to all other members of
```
SWvheerden, line 243: nit: Do we want to use the word mempool for the validator's memory storage of instructions?  The word mempool has become widely used in the blockchain's (aka baseLayer's) temp storage for transactions. Validator nodes technically don't deal with transactions but rather instructions. Perhals we should look into finding a unique name to help differentiate between the two?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+```
+
+> Note that the base layer tip used by the leader may be a number of blocks below the actual tip to minimise the 
+> number of reorgs and increase the chance that all committee members have the same view of the base layer
+
+The committee members process the proposal via Hotstuff, except that during the prepare and pre-commit stages, they must
+confirm that block in `base_layer_block` is still present in the base node that they are connected to. 
+
+If a validator node has reached the Commit stage, it must prevent the base layer from reorging by issuing a GRPC call to 
+`LockBlock`
```
SWvheerden, line 280: We need to properly think how this is implemented. This does open up a lot of attack vectors for a call like this. 

Should we not handle this the same way we handle a double spend or normal transaction for that matter. It's the validator node that cares that the checkpoint is in the main chain. A specific baseNode won't care about "Jhon's private checkpoint". If it gets reorged out they need to re-add it. Keeping old forked orphan chains breaks a lot of the block_chain ideas and needs special rules to cover those.

SWvheerden: Lets take this example:
There are 2 NFTs, `A` and `B`. 
At block height 1 Pete has 1 utxo with funds. 
At block height 2 Pete buys a NFT `A` with those funds. 
At block height 3 we have a checkpoint confirming Pete has NFT `A`

Now the block chain reorgs back to height 1
At block height 2-1 Pete buys a NFT `B` with those funds. 
At block height 3-1 we have a checkpoint confirming Pete has NFT `B`

So now at block height 3-1 what does Pete own? NFT `A `or `B` because according to the "LockedBlock" he owns `A`, but according to the main chain he owns `B`. Or does he own both and he just claimed a free NFT? When you want to buy NFT `A`, who owns it?

According to blockchain rules, NFT `A` is the person who owned `A` at height 1. As in the history Pete never bought `A`

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+The following fields must be included in the TransactionOutput, inside the OutputFeatures:
+
+New flags to be added to OutputFlags
+
+| Name | Value | Description |
+| --- | --- | --- |
+| NON_FUNGIBLE | 0b1000_1000 | This UTXO contains Non Fungile data and must not be combined with other fungible UTXOs |
+| ASSET_REGISTRATION | 0b1100_1000 | This UTXO contains registration information for a new asset |
+| MINT |0b0100_0000 | This UTXO represents the creation of a new NFT, and is used during the  |
+| BURN | 0b0010_0000 | The `unique_id` in this UTXO can be spent as a normal input and must not appear in the output set unless accompanied by a `MINT` flag |
```
philipr-za, line 72: This description is a bit confusing, is the `mint` supposed to be `BURN`?

SWvheerden (commented):
Looks good so far, few minor texts things. 

The one thing I either don't yet understand properly is the "lockedBlock" as it seems like it goes against anything blockchain.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+Note: this replaces some information in [RFC-0311](RFC-0310_AssetImplementation.md)
+
+| Name | Type | Description |
+| --- | --- | --- |
+| unique_id | bytes(32) | A unique id representing a token. This value must be present only when the NON_FUNGIBLE flag is set. The pair (parent_pub_key, unique_id) will be unique across the unspent set as described below |
+| parent_pub_key | bytes(32) |  A namespacing field that also allows a heirachy of assets. In most cases this will be the asset that a token belongs to, but future uses may be possible as well |
+| asset_registration | struct | Data for registering of an asset. This value **MUST** be present if ASSET_REGISTRATION flag is set |
+| asset_registration.name | bytes(64) | utf8 name of the asset. Optional |
+| asset_registration.template_ids | Vec<bytes(32)> |The templates that this asset exposes. Usually, at least one will be specified |
+| asset_registration.creator_sig | bytes(64) (32 Nonce + 32 sig) | A signature of the UTXO's `commitment` and `script` using the public key in `unique_id` to prove this registration was created in this UTXO |
```
philipr-za, line 87: does this imply that `unique_id` is a public key? It is not mentioned above.

philipr-za: I see it is mentioned below in more detail so perhaps just a mention of its public key nature in the table.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+| asset_registration.checkpoint_unique_id | bytes(32) | A reference to the unique id reserved for checkpoint UTXOs. Optional |
+| asset_registration.checkpoint_frequency | uint32 | The frequency, in sidechain blocks (or other measure of time the sidechain uses) in which checkpoints are created |
+| mint.issuer_proof | bytes(96) (64 nonce + 32 sig) | A _ComSig_ proof proving that the owner of the UTXO containing the asset registration created this token 
+| checkpoint.merkle_root | bytes(32) | Merkle root of the sidechain data. The format and meaning of this data is specific to the sidechain implementation |
+
+
+Note: `unique_id` is not always required to be a public key
```
philipr-za, line 95: Elaborate? My reading is that its a public key for asset registration but any random ID for a token?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+When receiving a UTXO with the `MINT` and `NON_FUNGIBLE` flags set, either the `parent_public_key` must be populated, 
+or the `ASSET_REGISTRATON` flag must also be set.
+
+Note: If the `parent_public_key` is populated and the `ASSET_REGISTRATION` flag is set, it must be treated as an asset registration 
+instruction. This allows sub assets to be grouped under a parent asset.
+
+**New Consensus Rule**
+
+When receiving a UTXO with the `MINT` and `NON_FUNGIBLE` flags set and the `ASSET_REGISTRATION` unset, the `minting_proof_signature`
```
philipr-za, line 153: `minting_proof_signature` is not mentioned anywhere else. Where does it live?

RFC/src/RFC-0310_AssetImplementation.md
philipr-za:
```suggestion
with the `unique_id`  that is the same as this minted output's `parent_public_key`. The message signed must be `Hash(commitment|script|unique_id|parent_public_key)`
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+must be a valid commitment signature using the commitment of the UTXO that contains the currently valid asset registration feature
+for with the `unique_id` = `parent_public_key`. The message signed must be `Hash(commitment|script|unique_id|parent_public_key)`
+
+Note: because this requires extra validation, the gram weight of a UTXO that carries the `MINT` flag should be higher.
+
+### Transferring a unique token
+
+The `MINT` flag is only set on the first UTXO a token appears in. When it is transferred to a new UTXO, it should only have the `NON_FUNGIBLE` flag set.
```
philipr-za, line 161: Don't we need some kind of consensus rule here to enforce that the `unique_id`s of any `NON_FUNGIBLE` flagged inputs  must appear in the outputs of this block?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+### Burning a unique token
+
+A token can be burnt by setting the `BURN` flag and then spending the UTXO. A token with the `BURN` set should be considered as valid as long as it is in the unspent set
+and is only destroyed once the UTXO is spent.
+
+**New Consensus Rule**
```
philipr-za, line 168: Do we perhaps need another consensus rule that enforces the burn that means you cannot undo the burn by spending to another output without the BURN flag?

RFC/src/RFC-0310_AssetImplementation.md
philipr-za: I think we have made ourselves used to saying n of m but a quick google confirms m of n is more widely used.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+A sidechain checkpoint is a special type of NFT UTXO. Extra sidechain data can be specified in the `sidechain` field of the output features. Like 
+other NFT's, only one sidechain checkpoint UTXO with that unique_id can exist in the unspent set at one time. The conditions of 
+who can spend a checkpoint are determined by the TariScript on the checkpoint. In most cases it will be an `n of m` multisig condition 
+using the members of the Hotstuff committee, with `m` equal to the number of members of the committee and `n` equal to `m` minus 
+the number of failures the committee can tolerate.
+
+The unique_id can be a specificly chosen id, but can also be a random public key. 
+
+There is no limit to the number of different sidechain checkpoint tokens an asset has, but it will usually be one.
```
philipr-za, line 182: I agree with this, I don't think there is any value in trying to use consensus to enforce anything about the checkpoints. They are purely there as on-chain commitments that users of the sidechain can point to and cannot be used for nay kind of enforcement by the main chain nodes.

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: Mempool is used in this RFC already: https://rfc.tari.com/RFC-0340_VNConsensusOverview.html#implementation

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: Let me expand a bit on the rational of it. The problem it is trying to solve is the fact that all committee members must have the same view of the blockchain when they are deciding if an instruction is valid or not. You could remove this probability by requiring commitments to be confirmed by X blocks before they can be used as a proof, but once a Hotstuff node commits to executing a set of instructions, it can't have the landscape underneath it change. It has already told the rest of the committee it will execute the instructions, so there has to be some sort of concept of locking.

Perhaps one way of doing it is to validate and pre-execute all the instructions in the Commit or PreCommit states and then the Decide phase is simply updating the state. I do think that the block that is at the tip should be communicated to the other committee nodes, so that they can confirm their view of the chain is the same

RFC/src/RFC-0310_AssetImplementation.md
philipr-za: I found you mention this down in the example but still think we should have a formal New Consensus Rule section here.

RFC/src/RFC-0312_AssetTemplateEdtitableMetadata.md
philipr-za:
```suggestion
# RFC-0312-AssetTemplateEditableMetadata
```

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+or block is valid only if:
+1. It contains exactly one input with the same `unique_id` and `parent_public_key`
+2. It contains no other UTXOs with the same `unique_id` and `parent_public_key`
+
+Note that the `ASSET_REGISTRATION` and `SIDE_CHAIN_CHECKPOINT` contain the `NON_FUNGIBLE` bit and will also contain 
+the `MINT` bit for the first instance of each of these.
+
+### Registering an asset
+
+A new asset is registered by creating a UTXO with the ASSET_REGISTRATION flag set, a new public key in `unique_id`,
```
sdbondi, line 116: Maybe expand on the usage of unique_id. If I understand correctly it can represent two types of data within the protocol. On registration, it is the public key of the registered asset and MUST be placed into the `parent_public_key` of the minted child token UTXOs. On a token UTXO it can be any sequence of bytes unique to the `parent_public_key` namespace.  Any reason we don't use a separate field for asset "class" public key?

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+### Minting a unique token
+
+A new token is created for an asset by creating a UTXO with the `MINT` and `NON_FUNGIBLE` flags set
+and the `mint_nonfungible` feature data set.
+
+**New Consensus Rule**
+
+When receiving a UTXO with the `MINT` and `NON_FUNGIBLE` flags set, either the `parent_public_key` must be populated, 
+or the `ASSET_REGISTRATON` flag must also be set.
```
sdbondi, line 146: typo: `ASSET_REGISTRATON`

I found this a little clearer
```suggestion
When receiving a UTXO with the `MINT` and `NON_FUNGIBLE` flags set, the `parent_public_key` and/or the `ASSET_REGISTRATION` flag must be set.
```

RFC/src/RFC-0310_AssetImplementation.md
sdbondi: Without an in-depth/practical understanding of hotstuff, doesn't the hotstuff protocol essentially allow for this? If the nodes do not agree on the next state in the `Decide` phase, there is still a valid way to abandon the view via the `NextView` interrupt. Should the tip change for `n - f` committee members, causing the current view to become invalid, nodes could broadcast `NextView` and the committee proceeds to the next view. If this comes too late and the view is passed, the reorg would need to be detected and instructions re-proposed for the next view i.e. pulled from the unconfirmed instruction mempool in a same/similar way to base layer transactions. 

[RFC 340](https://rfc.tari.com/RFC-0340_VNConsensusOverview.html#forks-and-byzantine-failures)
> If there is a temporary network partition, the chain may fork (up to a depth of three), but the protocol guarantees safety via the voting mechanism, and the chain will reconcile once the partition resolves.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+```json
+{
+   TransactionOutput: {
+      "features": {
+         // SIDECHAIN | MINT
+         "flags": "0b1101_1000",
+         // Special unique id,
+         "unique_id": "0x0000000000000000",
+         "parent_public_key": "<yacht_clicker_pub_key",
```
sdbondi, line 469: minor:
```suggestion
         "parent_public_key": "<yacht_clicker_pub_key>",
```
Missing in a few other places

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+The wallet will create a transaction containing a UTXO with the following `OutputFeatures` and publish
+to a base node.
+
+```json
+{
+  TransactionOutput: {
+    "features": {
+      "flags": "0b1100_1000", // ASSET_REGISTRATION
+      "unique_id": [
```
sdbondi, line 407: Should this be an array type?

sdbondi (commented):
Looks good, think this gives a good picture of the proposed solution. Mainly think that some clarification on the dual usage of `unique_id` is needed.

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+
+
+New fields added to `OutputFeatures`
+
+Note: this replaces some information in [RFC-0311](RFC-0310_AssetImplementation.md)
+
+| Name | Type | Description |
+| --- | --- | --- |
+| unique_id | bytes(32) | A unique id representing a token. This value must be present only when the NON_FUNGIBLE flag is set. The pair (parent_pub_key, unique_id) will be unique across the unspent set as described below |
+| parent_pub_key | bytes(32) |  A namespacing field that also allows a heirachy of assets. In most cases this will be the asset that a token belongs to, but future uses may be possible as well |
```
Cifko, line 83:
```suggestion
| parent_pub_key | bytes(32) |  A namespacing field that also allows a hierarchy of assets. In most cases this will be the asset that a token belongs to, but future uses may be possible as well |
```

RFC/src/RFC-0310_AssetImplementation.md
CjS77: These codes don't match the values in the code.

In the code, NON_FUNGIBLE is `0b1000`.

REGISTER is `0b10`, so NFT register would be `0b1010` etc

There are inconsistencies here and in the code, so we should tighten up the bitflag definitions.

RFC/src/RFC-0310_AssetImplementation.md
CjS77: We also need some additional lines in the RFC metadata, like "Deprecates RFC-xxx" and "Deprecated by RFC-XXX"

CjS77: If the bit fields are set up correctly, it should be easy to make clear definitions in the code (and this RFC).

`NFT_REGISTRATION = NON_FUNGIBLE | REGISTER`
`NFT_CHECKPOINT = NON_FUNGIBLE | CHECKPOiNT` 
etc

The bitflags crate actually has this functionality [built right in](https://docs.rs/bitflags/1.3.2/bitflags/).

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: Added

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: lol

RFC/src/RFC-0310_AssetImplementation.md
CjS77: Yes, this was also unclear for me.

RFC/src/RFC-0310_AssetImplementation.md
CjS77: The lock_base_node command feels unnecessary. 

We could require that as part of the COMMIT step, the Hotstuff committee members include the hash of the block they're building on; and Hotstuff consensus can be reached or declined in the usual way, but I'm not really sure that even this is necessary, as others have also mentioned.

If the chain re-orgs between the time that the COMMIT step is done and a transaction is broadcast, so be it, 
it is akin to a wallet broadcasting a transaction based on a slightly outdated view of the chain. The transaction is either still fine, or it gets rejected and the wallet reacts accordingly.

lock_base_node doesn't add any value in my view, because all it does is allow the committee members to have the same (possibly incorrect) view of the chain, and by broadcast time you have the same effect, since the base node is still allowed to reorg.

If there are persistent Byzantine failures due to different members having different views of the base chain, then we could have some sort of "resync" instruction that tries to get all committee members to converge on the same view.

If a committee transaction is rejected, the correct course of action for the committee would depend on the context and would need to be spec'd out a bit more. In any event, the committee is not waiting for the tx to be mined. It will immediately move onto the next instruction batch.

RFC/src/RFC-0310_AssetImplementation.md
CjS77: I'd love us to look for ways to conserve precious block space. The additional fields on OutputFeatures alone run to over 384 bytes (by very quick mental math) depending on the specifics.

There are a couple of signatures and fields here that I believe, with careful consideration, could be folded into the current UTXO fields, and offer some privacy for assets to boot.

E.g. The unique id might be added to the commitment as C = kG + vH + tT for instance.

RFC/src/RFC-0310_AssetImplementation.md
CjS77: In the example, it's just a hash though. So not always a pubkey

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+        ]
+      },
+      "metadata": [
+        protobuf
+        serialized
+        template
+        data
+        as
+        Vec<u8>
+      ]
```
CjS77, line 442:
```suggestion
      "metadata": [..] // protobuf serialized template data as Vec<u8>
```

and others

RFC/src/RFC-0310_AssetImplementation.md
```diff
@@ -0,0 +1,607 @@
...
+The wallet will create a transaction containing a UTXO with the following `OutputFeatures` and publish
+to a base node.
+
+```json
+{
+  TransactionOutput: {
+    "features": {
+      "flags": "0b1100_1000", // ASSET_REGISTRATION
+      "unique_id": [
+        generated_pubkey
```
CjS77, line 416: Is this the same as `<yacht_clicker_pub_key>` below?

RFC/src/RFC-0310_AssetImplementation.md
CjS77: `metadata` isn't mentioned anywhere prior to this. Is this the serialisation of the `fields` field above?

If so, it represents a potential scaling problem. Ignoring the fact that I'm limited by the base layer block size limit for now, what if I want to initialise a thousand fields for my asset?

CjS77 (commented):
Besides the specific comments left in the RFC, my overriding thinking while digesting this RFC is _will this implementation scale to handle millions of assets, each with millions of tokens and hundreds or thousands of descriptors / state variables / metadata, with millions of these tokens updating state many times a minute_?

And I think while this RFC has some great specific suggestions and sits on the back of a concrete, working PoC my sense is that in general, the answer is "not yet". 

How do we scale to millions of assets, billions of tokens and millions of state changes a day? 

One way to look at this is -- 1 MB/block * 30 / hr * 24 hr/day = 720 MB per day is the max data that every node on the network will be adding.  (It's roughly equivalent to 900k new standard MW UTXOs a day). This does not equate to millions of token mints and tens of millions of state changes _unless_ that state, and the tokens<sup>1</sup> are kept completely off chain.

What can the Tari Base layer handle in terms of assets? I would say that the full extent of the base layer's responsibility is to act as the primary ledger for Tari token transactions,  a global register for assets (just the asset, not the tokens) and a notary service for sidechains managing the assets.

All state, including tokens, must effectively live off chain.

<sup>1</sup>: Tari must allow token holders to "eject" tokens from side chains and keep them in cold storage, or as a means of transferring them to a different side chain, assuming direct SC<->SC transfers aren't possible. However, this is optional, and we should find a way to do this within the current framework of Tari base layer transactions, including the use TariScript. In addition, we should _disincentivise_ token holders to do this by making it relatively expensive to move a token on-chain by means of overweighted transaction fees for token storage; not because it's not a valid use case, but because the base layer simply cannot scale fast enough to make this something you "just do for the hell of it".

Like subatomic particles, tokens might pop into existence on a side-chain, move around a bit, and then get annihilated without the main chain ever being aware of it. This is a vital characteristic to achieve the scaling goals of Tari, and precludes the requirement that all mints must be recorded on the base layer.

Furthermore, tokens will need to be able to be minted at rates and in volumes that far exceed the block sizes and intervals of the main chain, further underlying the need for them to be created off-chain.

Coming slightly closer to home, I think the asset registration transaction described in this RFC is a really good foundation to work from. I suspect we can trim things down a little, but it's 100% in the right direction imho.

Notarisation / checkpointing is another key part of the base layer functionality. Again, given how little the base layer knows about the assets on side chains, the checkpoint transaction can likely also be shrunk significantly. There are also many details to iron out, for example what to do with cheats (and how we even define cheats). Some of the other RFCs need to be dusted off and revisited in this regard, since we've learnt a lot since those were first drafted.

I also understand that it's out of the scope of this RFC, but the topic of micropayments for validator nodes, atomic swaps for token purchases (99.999% of which should also be done off chain) and asset maintenance funding need to be covered.

There's a comment about decentralisation and censorship resistance (D&CR). This does not mandate everything to live on chain. It's perfectly doable to achieve D&CR on a suitably distributed side chain; one that's focussed on delivering D&CR for the asset that cares about that. Meanwhile, other assets that don't care as much (e.g. my grannie's biscuit shop's loyalty token) can sit on cheap, fast side-chains, comprising as few as one node.

In summary, this RFC is extremely valuable as a proof of concept, and for reigniting the conversation about how to achieve Tari's primary aim. Perhaps the RFC should place a link to this PR (how meta!) to preserve all the comments and discussion, and then I think it should be merged and added to the "canon".

RFC/src/RFC-0310_AssetImplementation.md
stringhandler: Having data be present or not present is actually a broader scope than this specific document. By disallowing extra data from appearing, you make it difficult to soft fork in changes in future. It's probably safer to allow extra fields (protobuf serialization will ignore them). Since in another RFC it is stated that features are paid for in fees, the user can add some arbitrary data, but must 1. sign it in the metadata_signature and 2. must include it in the weighting calculation

PR #3341, opened 2021-09-14
fix: remove explicit panic from rpc handshake on io error

sdbondi:
Description
---
Remove mistaken explicit panic

Motivation and Context
---
Forgot to remove panic while developing

How Has This Been Tested?
---
N/A

### DIFF
--- a/comms/src/protocol/rpc/handshake.rs
+++ b/comms/src/protocol/rpc/handshake.rs
@@ -148,7 +148,6 @@ where T: AsyncRead + AsyncWrite + Unpin
                 target: LOG_TARGET,
                 "IO error when sending new session handshake to peer: {}", err
             );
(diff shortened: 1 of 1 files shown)


stringhandler (approved):

delta1 (approved):
lmao

hansieodendaal:
Thanks.

I assume the `thread 'tokio-runtime-worker' panicked at 'ready_rx oneshot is never dropped without a reply: RecvError(())', tari-project/comms/src/protocol/rpc/client.rs:100:14` was as a result of the first panic that you fixed, being `thread 'tokio-runtime-worker' panicked at 'explicit panic', tari-project/comms/src/protocol/rpc/handshake.rs:151:13`.

PR #3342, opened 2021-09-14
docs: add CheckMultiSig/Verify to RFC-0202

delta1:
Description
---
Adds new opcodes for `CheckMultiSig` and `CheckMultiSigVerify` to RFC docs for TariScript opcodes 

Motivation and Context
---
tari-project/tari-crypto#59

How Has This Been Tested?
---
`mdbook serve`

### DIFF
--- a/RFC/src/RFC-0202_TariScriptOpcodes.md
+++ b/RFC/src/RFC-0202_TariScriptOpcodes.md
@@ -49,7 +49,7 @@ technological merits of the potential system outlined herein.
 
 ## Goals
 
-This Request for Comment (RFC) defines the opcodes that make up the TariScript scripting language and provides some 
+This Request for Comment (RFC) defines the opcodes that make up the TariScript scripting language and provides some
 examples and applicaitons.
 
 ## Related Requests for Comment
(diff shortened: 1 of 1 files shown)


RFC/src/RFC-0202_TariScriptOpcodes.md
```diff
@@ -325,6 +325,23 @@ push 0.
 Identical to [`CheckSig`](#checksigmsg), except that nothing is pushed to the stack if the signature is valid, and the
 operation fails with `VERIFY_FAILED` if the signature is invalid.
 
+##### CheckMultiSig(Msg)
+
+Pop the number $n$ indicating the number of allowed spenders, and then pop $n$ spender public keys off the
```
philipr-za, line 330: I don't think `spender` is the correct term here. You can use MultiSig for a number of use cases so I think it would be better to call them `signers`

stringhandler (approved):

RFC/src/RFC-0202_TariScriptOpcodes.md
delta1: Thanks will update

delta1:
draft while finalizing

RFC/src/RFC-0202_TariScriptOpcodes.md
```diff
@@ -325,6 +325,23 @@ push 0.
...
+stack. Then pop the number $m$ indicating the number of required spending signatures, and then pop $m$
+signatures from the stack provided as inputs to the script. If $m$ signatures of the possible $n$ public keys
```
hansieodendaal, line 332: It feels like there are too many `then`, `and then` sequences in these first two sentences.

delta1: Open to suggestions

stringhandler: This is fine. No need to change it

hansieodendaal (commented):
I feel it must be clear which part is the script (mined on the blockchain and not malleable) and which part is the input data. It should not be possible to determine which signatures make up the `n`, or what the `m` and `n` numbers are, by manipulating the input data. In essence, this should not be very much different from the way Bitcoin does it (see the [TLU summary on this](https://tlu.tarilabs.com/protocols/mimblewimble-mp-bp-utxo/MainReport.html#bitcoin--mtext-of-n--multisig-in-a-nutshell)).

RFC/src/RFC-0202_TariScriptOpcodes.md
SWvheerden: If we feel like it reads too difficult, perhaps make it a numbered list.

delta1: This needs to be changed to what was merged in tari-crypto anyway.

stringhandler:
Can this be moved to Ready to Review?

delta1:
No it's outdated

PR #3343, opened 2021-09-14
refactor: parallelized block validation

sdbondi:
Description
---

- all kernel validation for a block is handled by a single blocking
  thread
- all input validation for a block is handled by a single blocking
  thread
- output validation (incl. range proof validation) is done in parallel
  using a maximum of the configured number of blocking threads.
- reduction in the number of input, kernel and output loops
- adds new blockchain unit test infrastructure
- adds new unit tests

Motivation and Context
---
This PR aims to greatly reduce block sync times by optimising and parallelising
block validation.

How Has This Been Tested?
---
Unit tests
Manually base node sync


base_layer/core/src/validation/block_validators/async_validator.rs
```diff
@@ -0,0 +1,432 @@
...
+
+                if !input.is_mature_at(block_height) {
+                    warn!(
+                        target: LOG_TARGET,
+                        "Input found that has not yet matured to spending height: {}", block_height
+                    );
+                    return Err(TransactionError::InputMaturity.into());
+                }
+
+                match helpers::check_input_is_utxo(&*db, input) {
```
stringhandler, line 228: I think the header hash  perhaps previous_hash should be passed in here in the case where a block is added or removed from the chain while the validator is running
```suggestion
                match helpers::check_input_is_utxo(&*db, input, header.previous_hash) {
```

base_layer/core/src/validation/block_validators/async_validator.rs
```diff
@@ -0,0 +1,432 @@
...
+                let queue = queue.clone();
+                let range_proof_prover = self.factories.range_proof.clone();
+                let db = self.db.inner().clone();
+                task::spawn_blocking(move || {
+                    let db = db.db_read_access()?;
+                    let mut aggregate_sender_offset = PublicKey::default();
+                    let mut commitment_sum = Commitment::default();
+                    let mut outputs = Vec::new();
+                    let mut coinbase_index = None;
+                    while let Some((orig_idx, output)) = queue.lock().expect("lock poisoned").pop() {
```
stringhandler, line 295: This expect could be a problem

base_layer/core/src/validation/block_validators/async_validator.rs
stringhandler: Although I see the `db_read_access` above, so this is less of a worry, but a block may still be added before this task starts

base_layer/core/src/validation/block_validators/async_validator.rs
```diff
@@ -0,0 +1,432 @@
...
+                            }
+                            coinbase_index = Some(orig_idx);
+                        } else {
+                            // Lets gather the output public keys and hashes.
+                            // We should not count the coinbase tx here
+                            aggregate_sender_offset = aggregate_sender_offset + &output.sender_offset_public_key;
+                        }
+
+                        output.verify_metadata_signature()?;
+                        if bypass_range_proof_verification {
```
stringhandler:
```suggestion
                        if !bypass_range_proof_verification {
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -644,8 +644,8 @@ where B: BlockchainBackend
         let NewBlockTemplate { header, mut body, .. } = template;
         body.sort();
         let header = BlockHeader::from(header);
-        let mut block = Block { header, body };
-        let roots = self.calculate_mmr_roots(&block)?;
+        let block = Block { header, body };
+        let (mut block, roots) = self.calculate_mmr_roots(block)?;
```
SWvheerden, line 648: Why are moving the block in and out again? This look like calculate_mmr_roots is changing the block

base_layer/core/src/validation/block_validators/async_validator.rs
```diff
@@ -0,0 +1,432 @@
...
+                            output.verify_range_proof(&range_proof_prover)?;
+                        }
+
+                        helpers::check_not_duplicate_txo(&*db, &output)?;
```
stringhandler, line 316: Also might be worth checking that the chain has not changed before we start this check

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -659,13 +659,14 @@ where B: BlockchainBackend
...
     /// This function will panic if the block body is not sorted
-    pub fn calculate_mmr_roots(&self, block: &Block) -> Result<MmrRoots, ChainStorageError> {
+    pub fn calculate_mmr_roots(&self, block: Block) -> Result<(Block, MmrRoots), ChainStorageError> {
         let db = self.db_read_access()?;
         assert!(
             block.body.is_sorted(),
             "calculate_mmr_roots expected a sorted block body, however the block body was not sorted"
         );
-        calculate_mmr_roots(&*db, &block)
+        let mmr_roots = calculate_mmr_roots(&*db, &block)?;
```
SWvheerden, line 668: I know the assert! code is not part of this pr, but while we are changing the entire function should we not change this to return an error rather than panic the thread?

base_layer/core/src/validation/block_validators/orphan.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+/// This validator tests whether a candidate block is internally consistent.
+/// This does not check that the orphan block has the correct mined height of utxos
+
+/// This validator checks whether a block satisfies *all* consensus rules. If a block passes this validator, it is the
+/// next block on the blockchain.
+#[derive(Default)]
+pub struct BodyOnlyValidator;
```
stringhandler: Did you perhaps mean to put this struct in `body_only.rs` ?

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -180,7 +177,7 @@ impl ConsensusManagerBuilder {
     }
 
     /// Adds in a custom consensus constants to be used
-    pub fn with_consensus_constants(mut self, consensus_constants: ConsensusConstants) -> Self {
+    pub fn add_consensus_constants(mut self, consensus_constants: ConsensusConstants) -> Self {
```
SWvheerden, line 180: I like this, this is much more in line with what the function does than what the previous name implied.

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: It is not possible for the lock to get poisoned since it only exists while `pop()` is executing (if you didn't know lock poisoning is when a panic occurs while the lock guard exists i.e. the lock guard isn't safely dropped)

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: 🤦

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: Read lock prevents changes while the set of outputs is being validated

base_layer/core/tests/block_validation.rs
```diff
@@ -816,25 +812,19 @@ fn test_block_sync_body_validator() {
         &rules,
     );
     let new_block = db.prepare_block_merkle_roots(template).unwrap();
-    assert!(validator
-        .validate_body(&new_block, &*db.db_read_access().unwrap())
-        .is_err());
+    validator.validate_body(new_block).await.unwrap();
```
SWvheerden, line 815: Does this replace an is_err() with no error?

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: Yeah it's very unlikely, but possible if external code adds propagated blocks when the state machine is in block sync state (not confirmed that it does/doesn't). You fundamentally cannot share a single lock across threads, so we'll need to introduce that invariant here somehow either by ignoring propagated blocks while the base node is in sync mode as a rule / set a flag in the db so disable calls to `write` (i.e. read only mode). Open to suggestions.

base_layer/core/src/validation/block_validators/async_validator.rs
stringhandler: Fair enough

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: As you suggested, atomically checking the header hash vs the current header hash should also work. I'll add that as a final check before committing the new block.

stringhandler: Why not just check the height or hash of the tip after acquiring the read lock?

base_layer/core/src/validation/block_validators/async_validator.rs
stringhandler: (if you didn't know there are lines before the read lock where another thread can change the db) ;)

base_layer/core/src/validation/block_validators/async_validator.rs
sdbondi: We could do that too, but that will only save us starting an unnecessary validation that will fail. The main concern is when you release the read lock, post validation and acquire the write lock to commit the block - where it is possible for a data race: external code waiting on a write lock to write a different valid block to the tip of the chain, obtains the lock writes block A' to the tip, post validation acquires the write lock and writes the new tip A on top of A'. Come to think of it, I think we already prevent this on the database layer when inserting headers

stringhandler (approved):

SWvheerden (commented):
Looks good so far. I left a few questions. 

I synced to tip on this branch

PR #3344, opened 2021-09-14
refactor!: move unique_id and parent_public_key to outputfeatures

StriderDM:
Description
---
unique_id and parent_public_key has been moved to OutputFeatures instead of being individual fields on the Output itself.

Motivation and Context
---

How Has This Been Tested?
---


base_layer/core/src/proto/transaction.proto
```diff
@@ -75,10 +73,12 @@ message OutputFeatures {
     uint64 maturity = 2;
 
     bytes metadata = 3;
-
+    // Numbering 4-15 taken
```
stringhandler: My intention here was to use the recommendation here https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers but I think it will be confusing, so lets just use numbers from 4 onwards

```suggestion
    
```

base_layer/core/src/proto/transaction.proto
```diff
@@ -75,10 +73,12 @@ message OutputFeatures {
...
     AssetOutputFeatures asset = 16;
```
stringhandler:
```suggestion
    AssetOutputFeatures asset = 4
```

base_layer/core/src/proto/transaction.proto
```diff
@@ -75,10 +73,12 @@ message OutputFeatures {
...
     MintNonFungibleFeatures mint_non_fungible = 17;
     SideChainCheckpointFeatures sidechain_checkpoint = 18;
+    optional bytes unique_id = 19;
```
stringhandler:
```suggestion
    optional bytes unique_id = 5;
```

stringhandler (commented):
Looks good. Happy to merge when comments are addressed.

base_layer/core/src/proto/transaction.proto
StriderDM: Have re-numbered these fields

stringhandler (approved):

PR #3346, opened 2021-09-15
feat: send token

Cifko:
Description
---
When you are changing the amount, you can select token instead. I think the UI is very intuitive. Please do comment if you have better idea for the UI.

Motivation and Context
---

How Has This Been Tested?
---
Manually. After the transactions is mined, you need to recover the wallet to change the state of the token to unspent (the same behavior as with mint tokens and registering asset).


stringhandler (approved):
Tested locally. There were problems, but those are outside the scope of this pr

PR #3347, opened 2021-09-15
test: reduce max utxo query size

hansieodendaal:
Description
---
Reduced the wallet's `max_utxo_query_size`

Motivation and Context
---
Multiple `IO Error: frame size too big.` errors occurred in the RPC client with UTXO validation queries during console wallet and base node testing. This resulted in the RPC connection being closed and the UTXO validation task retrying the same thing.

**Edit: ** The error occurred in the wallet comms.

How Has This Been Tested?
---
System-level testing between the console wallet having a large wallet and the base node.


stringhandler:
Not sure is this is a bug? @sdbondi what do you think?

sdbondi:
@mikethetike @hansieodendaal Almost certainly caused by some peers are running the old code before breaking changes in PR #3336  - including a reduction in frame size, which makes sense given this error message. Running a node without `force_sync_peers` set to check.

sdbondi:
@hansieodendaal Have you tried setting your base node on the wallet to a base node running the latest code?

hansieodendaal:
@sdbondi, base node and wallet was on the latest code. The error occurred in the wallet comms due to queries from the console wallet to the base node that was too big, I guess probably due to changing transaction output data sizes. 

We may need to investigate why reducing the batch size was now again necessary, as it was reduced not long ago? I do not recall recent changes in the data model.

sdbondi:
@hansieodendaal ok I'll take a closer look at this and see where things need to be optimised

sdbondi:
@hansieodendaal @mikethetike So 3500 x ~1000 bytes(depends on tariscript) = 3.5MB which is still in the 4MB limit. This method should be streaming to ensure that responses are within 4MB. Having said this, I cannot see why this needs to be reduced.  I'll have to take a look at your logs.

sdbondi (approved):
Think this can go in, this RPC method needs to be streaming though. Less UTXOs in a batch means more request response rounds which means slower TX validation.

Testing locally I am not able to reproduce any issues with 3500, please feel free to message me with logs (making sure that both sides support chunking) and I'll see if I can find anything.

stringhandler (approved):

PR #3348, opened 2021-09-15
fix: update block explorer to use local grpc

stringhandler:
Fix text explorer to use local GRPC npm. The previous GRPC resulted in an index out of range error


delta1 (approved):
LGTM - untested

philipr-za (approved):
LGTM

PR #3349, opened 2021-09-15
fix: fix median timestamp index

SWvheerden:
Description
---
The current code will ask for the last 12 block heights and not 11. This code fixes it to only work on the last 11

Motivation and Context
---
The consensus rules are stated to work on the last 11 blocks and not 12. The current code incorrectly includes a second index.

How Has This Been Tested?
---
Unit tests


stringhandler (approved):
Technically a breaking change, but unlikely to have an effect in practice

PR #3351, opened 2021-09-15
ci: add wallet ffi to build

stringhandler:
Add wallet FFI to build. 

Previously this would fail on the cucumber step


sdbondi (approved):

StriderDM:
Looks good. 

We should also add the wallet ffi builds that compile libraries specific to the android and iOS devices to CI (the ones that run the mobile_build.sh script). This will show any compilation problems of the library specific to the device architectures.

Just a note.

delta1 (approved):

PR #3352, opened 2021-09-15
feat!: integration of new transaction and output validation

philipr-za:
Description
---

This PR cleans up and updates the new Output and Transaction validation strategies.

@mikethetike provided the first draft of this new validation approach for both Outputs and Transactions but the work needed some love to fully integrate into the wallet code. Many edge cases were caught and some bigger changes are made in this PR.

Some major changes are:
- Output validation now supports checking and reporting on number of confirmations
- The Pending Transaction database table is fully removed and this process is now handled fully within the Outputs table
- Validations are now triggered on receiving a new block from the base node rather than at an interval
- Base node `query_deleted` call is updated to return the block data of where a deletion occured

How Has This Been Tested?
---
Tests provided/updated and manually


stringhandler (approved):
Difficult to review everything in this pr, but I think I managed to go over the important pieces. Happy to merge and give it a try

PR #3353, opened 2021-09-15
ci: add wallet ffi to build and fix release build for wallet

delta1:
Description
---
- add wallet ffi build to ci 
- fix release build for wallet 

Motivation and Context
---
#3351

How Has This Been Tested?
---
- `cargo build -p tari_wallet --release` 
- `cargo build -p tari_wallet_ffi --release`

PR #3354, opened 2021-09-16
fix: dead_code lint error when base_node_feature is not set

sdbondi:
Description
---
Use `new_sorted_unchecked` function to keep clippy happy when base_node feature is not active
in tari_core.

Motivation and Context
---
Causing failed CI build (e.g. #3335 )

How Has This Been Tested?
---
CI must pass (FFI tests)


delta1:
Seems fine, I went the other way in #3353 adding the base_node feature to that function - is it that an issue?

sdbondi:
@delta1 Yeah I did that at first, but feels like the feature attrs are getting messier and messier. If you feel strongly one way we can close this.

delta1:
No I agree, I prefer this way 👍

stringhandler (approved):

PR #3357, opened 2021-09-16
fix: stop MTP attack

SWvheerden:
Description
---
This changes the timestamp of a new block template to be always greater than the median time past (MTP) of the past blocks as per consensus. 

Motivation and Context
---
It is possible for an attack to forward the median time to be greater than now, but less than the future time limit(FTL). This means that all valid blocks with the current timestamp will be rejected. This will ensure that ll newly created blocks always have a timestamp greater than the MTP. Due to consensus we also know that MTP will always be less than FTL. This ensures that all new block timestamps are `MTP < Timestamp < FTL`

For a description see: [Graft network](https://github.com/graft-project/GraftNetwork/pull/118#issuecomment-384599435)
and [MTP Attack (Jagernan)](https://github.com/zawy12/difficulty-algorithms/issues/30)

How Has This Been Tested?
---

All current unit tests pass.

### DIFF
--- a/applications/tari_mining_node/src/difficulty.rs
+++ b/applications/tari_mining_node/src/difficulty.rs
@@ -75,8 +75,14 @@ impl BlockHeaderSha3 {
             .chain(self.header.total_script_offset.as_bytes())
     }
 
-    pub fn set_timestamp(&mut self, timestamp: u64) {
-        self.timestamp = timestamp;
+    /// This function will update the timestamp of the header, but only if the new timestamp is greater than the current
+    /// one.
+    pub fn set_forward_timestamp(&mut self, timestamp: u64) {
+        // if the timestamp has been advanced by the base_node due to the median time we should not reverse it but we
+        // should only change the timestamp if we move it forward.
+        if timestamp > self.timestamp {
+            self.timestamp = timestamp;
+        }
     }
 
     pub fn random_nonce(&mut self) {
@@ -171,7 +177,7 @@ pub mod test {
             );
             timestamp = timestamp.increase(1);
             core_header.timestamp = timestamp;
-            hasher.set_timestamp(timestamp.as_u64());
+            hasher.set_forward_timestamp(timestamp.as_u64());
         }
     }
 }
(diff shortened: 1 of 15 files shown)


applications/tari_mining_node/src/difficulty.rs
```diff
@@ -76,7 +76,11 @@ impl BlockHeaderSha3 {
     }
 
     pub fn set_timestamp(&mut self, timestamp: u64) {
-        self.timestamp = timestamp;
+        // if the timestamp has been advanced by the base_node due to the median time we should not reverse it but we
+        // should only change the timestamp if we move it forward.
+        if timestamp > self.timestamp {
```
sdbondi, line 83: nit: I'd rather go for putting the check outside the function call (it's only used in one place anyhow). It isn't expected for `set_timestamp` not to set the timestamp. Alternatively, name the function `set_forward_timestamp` (or something) and explain in a docstring that is only sets the timestamp forward from what it is. This is a nit so feel free to ignore.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -643,7 +650,24 @@ where B: BlockchainBackend
     pub fn prepare_block_merkle_roots(&self, template: NewBlockTemplate) -> Result<Block, ChainStorageError> {
         let NewBlockTemplate { header, mut body, .. } = template;
         body.sort();
-        let header = BlockHeader::from(header);
+        let mut header = BlockHeader::from(header);
```
stringhandler, line 653: This method now does more than it says it does. It should be renamed or this logic should be moved elsewhere

sdbondi, line 653: nit: We may as well call this function `prepare_new_block(NewBlocktemplate) -> Block`

sdbondi (dismissed):
Nice catch - some nits that you can safely ignore :)

applications/tari_mining_node/src/difficulty.rs
SWvheerden: I agree good idea

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: I agree, will rename. That is what the function did anyway. All we do now is we do an extra check before we set the timestamp. Before this, the set_timestamp happened in `let header = BlockHeader::from(header);`

SWvheerden: I renamed the method to `prepare_new_block`. The timestamp was already set by this method, it was done inside of `let header = BlockHeader::from(header);`. All it does not is add an extra check to what it should set it and not just blanketly set it to now. But I do agree with the fact that the function name and what it does is misaligned.

stringhandler (approved):

PR #3360, opened 2021-09-17
fix: additional check for cancelled transactions

StriderDM:
Description
---
Additional check for cancelled transactions

Motivation and Context
---

How Has This Been Tested?
---
cargo test --all

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/app_state.rs
+++ b/applications/tari_console_wallet/src/ui/state/app_state.rs
@@ -586,6 +586,12 @@ impl AppStateInner {
                     self.refresh_balance().await?;
                     self.updated = true;
                     return Ok(());
+                } else if (tx.status == TransactionStatus::Pending || tx.status == TransactionStatus::Completed) &&
+                    tx.cancelled
+                {
+                    self.refresh_balance().await?;
+                    self.updated = true;
+                    return Ok(());
                 }
 
                 if let Some(index) = self.data.completed_txs.iter().position(|i| i.tx_id == tx_id) {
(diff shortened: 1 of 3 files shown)


base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -385,6 +385,19 @@ where TBackend: TransactionBackend + 'static
     }
 
     async fn receive_transaction_cancellation(&mut self, tx_id: TxId) {
+        match self.db.get_any_transaction(tx_id).await {
+            Ok(_) => {
+                // Make sure the transaction is cancelled
+                self.db.cancel_pending_transaction(tx_id).await.unwrap_or_default();
```
philipr-za, line 391: I disagree with this. This callback handler should not be contributing to changing state at all. It monitors the event streams and calls callbacks. If this is a bug then it really needs to be addressed at the point in the code that would produce these events. This is going to create a different behaviour in a mobile wallet using the FFI vs the console wallet.

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -586,6 +586,12 @@ impl AppStateInner {
                     self.refresh_balance().await?;
                     self.updated = true;
                     return Ok(());
+                } else if (tx.status == TransactionStatus::Pending || tx.status == TransactionStatus::Completed) &&
```
philipr-za, line 589: This is a needless repetition 
```Rust
    self.refresh_balance().await?;
    self.updated = true;
```
occur at the bottom of this function (Line 609) regardless of what happens here.

StriderDM:
Closing for https://github.com/tari-project/tari/pull/3369

PR #3363, opened 2021-09-17
ci: create separate mining artifact

stringhandler:
Description
---
Creates and uploads a separate mining artifact when building binaries

Motivation and Context
---
Since you can mine against a pool, you may only want to download the mining node

How Has This Been Tested?
---
Tested manually via Github action


delta1 (approved):
Nice idea, haven't tested this action but looks fine

philipr-za (approved):

PR #3364, opened 2021-09-17
fix: pressing b key should not clear existing base node in console wallet

StriderDM:
Description
---
Pressing b key should not clear existing base node, it should only activate the selection.
Additional check for selected().

Motivation and Context
---

How Has This Been Tested?
---
Manually


sdbondi (commented):
🎉  Tested and works

sdbondi (approved):

PR #3365, opened 2021-09-17
fix: relative paths for logs is now relative to data path instead of current execution directory

StriderDM:
Description
---
Relative paths for logs is now relative to data path instead of current execution directory

Motivation and Context
---
Logs are now stored in the expected location instead in the current working directory.

How Has This Been Tested?
---
Manually


sdbondi (approved):
Makes sense - utAck

PR #3369, opened 2021-09-20
fix: additional check for cancelled transactions

StriderDM:
Description
---
Check if a transaction has already been received and cancelled on accept transaction.

Motivation and Context
---

How Has This Been Tested?
---
cargo test --all


philipr-za (approved):
LGTM

PR #3370, opened 2021-09-20
fix: base node peer updates after being set

StriderDM:
Description
---
Fixed get_current_base_node_peer() cloning the watch instead of getting a reference to the watch and then cloning the peer.
Fixed some ownership nits in app_state.
Lifted return out of match in BaseNodeInputMode::Address.

Motivation and Context
---

How Has This Been Tested?
---
Manually

set_base_node_peer

set_base_node_peer_2

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/app_state.rs
+++ b/applications/tari_console_wallet/src/ui/state/app_state.rs
@@ -411,7 +411,7 @@ impl AppState {
     }
 
     pub fn get_required_confirmations(&self) -> u64 {
-        (&self.node_config.transaction_num_confirmations_required).to_owned()
+        self.node_config.transaction_num_confirmations_required
     }
 
     pub fn toggle_abandoned_coinbase_filter(&mut self) {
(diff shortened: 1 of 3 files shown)


applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -411,7 +411,7 @@ impl AppState {
     }
 
     pub fn get_required_confirmations(&self) -> u64 {
-        (&self.node_config.transaction_num_confirmations_required).to_owned()
+        (&self.node_config).clone().transaction_num_confirmations_required
```
stringhandler: ?

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -423,7 +423,7 @@ impl AppState {
     }
 
     pub fn unread_notifications_count(&self) -> u32 {
-        self.cached_data.new_notification_count
+        (&self.cached_data).clone().new_notification_count
```
stringhandler: Why the clone?

stringhandler (commented):
I'm not convinced these changes are for the better. They are adding clones for no reason

stringhandler:
I don't understand how this fixes the problem

PR #3243, opened 2021-08-24
feat!: make DhtHeader field non-malleable

philipr-za:
Description
---
This PR includes the relevant DhtHeader fields in the commitment of a Dht Message MAC signature so that they cannot be changed while in route.

The header fields included in the commitment are:
- Major version
- Minor version
- Destination
- Message type
- Message flags
- Expiry time (if exists)
- Ephemeral public key (if exists)

How Has This Been Tested?
---
Rust integration test provided and existing tests updated.

BREAKING CHANGE: Previous message signatures will not be valid

PR #3195, opened 2021-08-13
[DO NOT MERGE] fix: fix input_mmr type

SWvheerden:
## Description

Currently, the input_mmr is of type `MutableMmr` and this includes a bitmap. The input_mmr does not need a bitmap only needs a mountain range this fixes the typing to be `MerkleMountainRange`
## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I have squashed my commits into a single commit.

PR #3193, opened 2021-08-13
[DO NOT MERGE]feat: change kernel ordering

SWvheerden:
## Description

This changes the kernel order to be on the excess signature (nonce then signature)

## Motivation and Context

This locks the kernel into canonical ordering. 

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I have squashed my commits into a single commit.

PR #3372, opened 2021-09-21
feat: merge consensus breaking changes in #3195 #3193 with weatherwax compatibility

sdbondi:
Description
---
Block v1 (compatible with weatherwax at all heights):
- Kernel ordering using every kernel field
- Input MR commits to empty serialization of bitmap

Block v2 (Igor, breaking change)
- Kernel ordering (#3193)
- Input MR uses input hashes only without extaneous empty bitmap bytes
  (#3195)

Misc

- Add missing kernel order block validation
- Removed unnecessary transaction body sorting in wallet, it (still) is the responsibility of the GetNewBlockTemplate to assemble a final sorted block body.

Motivation and Context
---
Allow consensus breaking code to be merged without hard forking weatherwax. 
Igor will have to be reset.

To follow in subsequent PR: network breaking changes from DHT header malleability fix (#3243 ) 

How Has This Been Tested?
---
Archival sync weatherwax node from scratch, received propagated blocks
Washing machine between two weatherwax wallets

### DIFF
--- a/base_layer/core/src/transactions/transaction.rs
+++ b/base_layer/core/src/transactions/transaction.rs
@@ -926,7 +926,7 @@ impl From<CryptoFullRewindResult<BlindingFactor>> for FullRewindResult {
 /// [Mimblewimble TLU post](https://tlu.tarilabs.com/protocols/mimblewimble-1/sources/PITCHME.link.html?highlight=mimblewimble#mimblewimble).
 /// The kernel also tracks other transaction metadata, such as the lock height for the transaction (i.e. the earliest
 /// this transaction can be mined) and the transaction fee, in cleartext.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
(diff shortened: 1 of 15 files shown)


base_layer/core/src/transactions/transaction.rs
```diff
@@ -947,6 +947,76 @@ impl TransactionKernel {
...
+        if self.excess_sig.verify_challenge(excess, &c) {
+            Ok(())
+        } else {
+            Err(TransactionError::InvalidSignatureError(
+                "Verifying kernel signature".to_string(),
+            ))
+        }
+    }
+
+    pub fn deprecated_cmp(&self, other: &Self) -> Ordering {
```
stringhandler, line 969:
```suggestion
    // This method was used to sort kernels. It has been replaced, and should be removed in future
    pub fn deprecated_cmp(&self, other: &Self) -> Ordering {
```

stringhandler (dismissed):

stringhandler (approved):

PR #3375, opened 2021-09-21
refactor(wallet): simpler debounce mechanism

stringhandler:
Description
---
A simpler debounce mechanism for the wallet

Motivation and Context
---
@hansieodendaal in #3304 added a debounce service that potentially will queue up requests for a long time in the receive channel. This is a simpler implementation that achieves the same purpose. I have hard coded in the debounce time, since this is a visual timing that will feel smooth to the eye and configuring it per method and call is overkill.

How Has This Been Tested?
---
Manually


hansieodendaal (commented):
Hi @mikethetike, this is nicely generalized, but I feel we will be missing the purpose of the debouncer for the balance calculation specifically. I have two major comments as seen with system-level testing:

- We need the very last trigger to also execute the required function because the next trigger can be received after the update occurred but before the end of the debounce time thus leaving the last trigger request stale; the wallet could wait for a long time for the next trigger displaying the wrong information. (This was seen with the balance specifically.)
- Function `trigger_balance_refresh` specifically needs a configurable debounce time. With the system-level testing I have performed the `default_debounce_time` is too quick for this function for a large busy wallet; monitoring a couple of hundred transactions in various states will trigger this function continuously under stressed conditions as the balance calculation time quickly escalates to >> 1s. For that case, I was running with a 60 s delay between calculations to prevent continuous database queries.

stringhandler:
You are solving the wrong problem then. If the calculation takes >1s you should make the calculation not take more than 1s as a first step

PR #3382, opened 2021-09-23
chore: bundle sqlite by default for tests

stringhandler:
Description
---
Yet another attempt at bundling openssl and sqlite for easy debugging/testing on Windows

Motivation and Context
---
The current feature setup means that you cannot run `cargo test` in the `wallet_ffi` or any specific folder other than the root. I tried to follow the same way that other crates do this, but allowing a user to optionally specify whether a dependency is bundled or not, and not bundle it by default, but this makes running individual tests difficult. 

Since tests are run regularly, I made the default on wallet_ffi include the bundles, but then added `--no-default-features` to the mobile builds 

How Has This Been Tested?
---
Cargo test

### DIFF
--- a/base_layer/wallet_ffi/mobile_build.sh
+++ b/base_layer/wallet_ffi/mobile_build.sh
@@ -79,7 +79,7 @@ if [ -n "${DEPENDENCIES}" ] && [ -n "${PKG_PATH}" ] && [ "${BUILD_IOS}" -eq 1 ] 
   export PKG_CONFIG_PATH=${PKG_PATH}
   # shellcheck disable=SC2028
   echo "\t${CYAN}Building Wallet FFI${NC}"
-  cargo-lipo lipo --release > "${IOS_LOG_PATH}/cargo.txt" 2>&1
+  cargo-lipo lipo --release --no-default-features > "${IOS_LOG_PATH}/cargo.txt" 2>&1
   cd ../..
   cd target || exit
   # Copy the fat library (which contains symbols for all built iOS architectures) created by the lipo tool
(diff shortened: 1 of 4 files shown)


delta1:
LGTM just waiting on CI

delta1 (dismissed):

philipr-za (dismissed):
LGTM

base_layer/wallet_ffi/mobile_build.sh
```diff
@@ -392,7 +392,7 @@ EOF
       if [ "${MACHINE}" == "Mac" ]; then
         if [ "${MAC_MAIN_VERSION}" -le 10 ]; then
           if [ "${MAC_SUB_VERSION}" -ge 15 ]; then
-            cargo build --lib --release > "${ANDROID_LOG_PATH}/cargo_${PLATFORMABI}_${LEVEL}.txt" 2>&1
+            cargo build --lib --release --no-default-features > "${ANDROID_LOG_PATH}/cargo_${PLATFORMABI}_${LEVEL}.txt" 2>&1
           else
             cargo ndk --target ${PLATFORMABI} --android-platform ${LEVEL} -- build --release > "${ANDROID_LOG_PATH}/cargo_${PLATFORMABI}_${LEVEL}.txt" 2>&1
```
StriderDM: Are we only keeping support for the latest macOS? If not, this will need to be updated as well.

stringhandler: didn't realise there were so many logic branches here

StriderDM:
Looks like it could work, although I think we should do this before merging this in:
mobile_build.sh has to run and produce libraries for both android and iOS.
The wallet projects have to be built and run with these libraries (disable automatic download of dependencies or they will be overwritten) on the device or a simulator configured to use the device architectures we build for.

Cargo test in this case is unfortunately not enough to tell if this will work on the intended devices, the above will.

StriderDM:
There is also an alternative to this since the purpose of it is purely for tests. Since the libraries are not being static linked in to the wallet_ffi library by building with cargo, having the dependencies installed beforehand and in the path will be sufficient for the tests to run for wallet_ffi.

Just another option.

philipr-za (approved):

PR #3383, opened 2021-09-24
feat: update coinbase handling for new tx and output validation

philipr-za:
Description
---
The new tx validation protocol would continuously try and revalidate abandoned coinbase transactions and the TXO validation would try validate abandoned or reforged Coinbases indefinitely.

This PR updates the Tx validation handing of Coinbases to give them the Coinbase status when they are abandoned which means they won’t be revalidated but will still be checked during a reorg. The TXO validation has been updated so that ab abandoned coinbase can be marked as Abandoned when the transaction validation detects the abandonment so that the output will not be validated indefinitely.

A new test is provided and a number of cucumber tests for this branch have been fixed.

How Has This Been Tested?
---
Unit test provided and cucumber tests updated

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/models.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/models.rs
@@ -131,4 +131,5 @@ pub enum OutputStatus {
     ShortTermEncumberedToBeReceived,
     ShortTermEncumberedToBeSpent,
     SpentMinedUnconfirmed,
+    AbondonedCoinbase,
 }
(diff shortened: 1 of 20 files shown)


SWvheerden (approved):
LGTM

base_layer/wallet/src/output_manager_service/storage/models.rs
```diff
@@ -131,4 +131,5 @@ pub enum OutputStatus {
     ShortTermEncumberedToBeReceived,
     ShortTermEncumberedToBeSpent,
     SpentMinedUnconfirmed,
+    AbondonedCoinbase,
```
stringhandler:
```suggestion
    AbandonedCoinbase,
```

philipr-za: oof, nice catch

stringhandler: Happy with it otherwise

stringhandler (approved):

PR #3385, opened 2021-09-27
test: fix brokend test and remove @broken

Cifko:
Description
---
I've monitored the test and it was not stuck, it was just unlucky that one of the transaction was going to be processed last. Because they are not processed in the same order as we are checking them. I've doubled the timeout value.

How Has This Been Tested?
---
Manually.

### DIFF
--- a/integration_tests/features/StressTest.feature
+++ b/integration_tests/features/StressTest.feature
@@ -5,12 +5,12 @@ Feature: Stress Test
         Given I have a seed node NODE1
         And I have stress-test wallet WALLET_A connected to the seed node NODE1 with broadcast monitoring timeout <MonitoringTimeout>
         And I have a merge mining proxy PROXY connected to NODE1 and WALLET_A with default config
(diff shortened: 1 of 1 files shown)


integration_tests/features/StressTest.feature
```diff
@@ -21,59 +21,59 @@ Feature: Stress Test
...
         Examples:
-            | NumTransactions   | NumCoinsplitsNeeded   | NumNodes  | MonitoringTimeout |
-            | 10000             | 21                    | 3         | 300               |
+            | NumTransactions | NumCoinsplitsNeeded | NumNodes | MonitoringTimeout |
+            | 10000           | 21                  | 3        | 600               |
 
     @long-running
     Scenario: Simple Stress Test
-        Given I have a seed node NODE1
+        Given I have a seed node N3ODE1
```
delta1: intentional?

Cifko: lol, no ;-)
thanks

stringhandler (approved):

PR #3389, opened 2021-09-28
fix: reduce console wallet tui memory usage

hansieodendaal:
Description
---
Transactional information is copied to an internal buffer for the console wallet to use in its screen refresh tick cycle and updates to this information are event-driven. This PR reduces the memory overhead required to keep transactions in their various life cycle stages in the buffer memory.

Motivation and Context
---
Unnecessary memory was allocated to the internal buffer as the entire completed transaction body was kept while only the signature(s) and maturity is required. This is more significant for large wallets.

How Has This Been Tested?
---
System-level testing

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/app_state.rs
+++ b/applications/tari_console_wallet/src/ui/state/app_state.rs
@@ -877,6 +877,10 @@ pub struct CompletedTransactionInfo {
     pub direction: TransactionDirection,
     pub valid: bool,
     pub mined_height: Option<u64>,
+    pub is_coinbase: bool,
+    pub weight: u64,
+    pub inputs_count: usize,
+    pub outputs_count: usize,
 }
 
 impl From<CompletedTransaction> for CompletedTransactionInfo {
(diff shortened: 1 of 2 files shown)


philipr-za (approved):
LGTM

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -852,10 +861,72 @@ impl AppStateInner {
     }
 }
 
+#[derive(Clone)]
+pub struct CompletedTransactionInfo {
+    pub tx_id: TxId,
+    pub source_public_key: CommsPublicKey,
```
stringhandler, line 867: Can reduce memory usage even more by storing the text representation instead of the crypto primitive
```suggestion
    pub source_public_key: String,
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -852,10 +861,72 @@ impl AppStateInner {
...
+    pub destination_public_key: CommsPublicKey,
```
stringhandler, line 868: same

stringhandler (approved):

PR #3392, opened 2021-09-28
[for discussion] feat: get_balance mods to reduce memory foorprint

hansieodendaal:
Description
---
This PR:
- reduces simultaneous instances of potentially large heap memory allocations;
- parallelize the initial two database queries for pending transactions and unspent outputs.

Motivation and Context
---
The `pub async fn get_balance` function was identified as one place where a lot of memory build-up is occurring in the Ubuntu version of the console wallet. 

How Has This Been Tested?
---
System-level testing in Ubuntu 20.04 LTS


stringhandler (commented):
I think it will be better to replace this method with a single SQL query

PR #3394, opened 2021-09-29
refactor!: mapping for deleted mmr position to height/hash for perf

sdbondi:
Description
---
- add index to optimise query for block height from deleted mmr position
- update query_deleted rpc call to use new call
- check if database should be resynced to avoid confusion around
  db errors resulting from this change
- exit code message showing help for a required db resync

Motivation and Context
---
Base node RPC call to `query_deleted` is slow
 
How Has This Been Tested?
---
Updated blockchain db unit test, manually

BREAKING CHANGE: blockchain database will need to be resynced


stringhandler (approved):
Nice

PR #3406, opened 2021-10-02
refactor: rename validator node

stringhandler:
Description
---
Rename tari_dan_node to tari_validator_node

Note: I also allowed some defaults on the config file

Motivation and Context
---
To be more inline with RFCs

How Has This Been Tested?
---


Cifko:
LGTM. But there are some clippy issues.

PR #3407, opened 2021-10-04
feat: add unique id index to lmdb

StriderDM:
Description
---
Adds unique_id index to lmdb

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -67,6 +67,7 @@ use crate::{
             LMDB_DB_ORPHAN_HEADER_ACCUMULATED_DATA,
             LMDB_DB_ORPHAN_PARENT_MAP_INDEX,
             LMDB_DB_TXOS_HASH_TO_INDEX,
+            LMDB_DB_UNIQUE_ID_INDEX,
             LMDB_DB_UTXOS,
             LMDB_DB_UTXO_COMMITMENT_INDEX,
             LMDB_DB_UTXO_MMR_SIZE_INDEX,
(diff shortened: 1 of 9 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -445,6 +449,16 @@ impl LMDBDatabase {
             "utxo_commitment_index",
         )?;
 
+        if let Some(unique_id) = &output.features.unique_id.clone() {
```
sdbondi, line 452:
```suggestion
        if let Some(unique_id) = output.features.unique_id.as_ref() {
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -567,6 +581,10 @@ impl LMDBDatabase {
             "utxo_commitment_index",
         )?;
 
+        if let Some(unique_id) = input.features.unique_id.clone() {
```
sdbondi, line 584:
```suggestion
        if let Some(unique_id) = input.features.unique_id.as_ref() {
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -841,6 +859,9 @@ impl LMDBDatabase {
                     output.commitment.as_bytes(),
                     "utxo_commitment_index",
                 )?;
+                if let Some(unique_id) = output.features.unique_id.clone() {
```
sdbondi, line 862:
```suggestion
                if let Some(unique_id) = output.features.unique_id.as_ref() {
```

sdbondi (commented):
utAck - so the idea is that the hash preimage is `asset namespace/class || the token`?

delta1:
> utAck - so the idea is that the hash preimage is `asset namespace/class || the token`?

Sounds right, I think according to [proposed RFC] `(parent_pub_key, unique_id)` should be unique 

[proposed RFC]: https://github.com/tari-project/tari/pull/3340/files#diff-62c91e11620b9f0105f0c6418dbce59c6924086076081011ef45d38d61c05617R82

delta1 (changes requested):
as per the [proposed RFC](https://github.com/tari-project/tari/pull/3407#issuecomment-933309055) the key for the index should be `parent_public_key || unique_id`

delta1:
I'm happy to pick this up as part of the validations I'm doing @StriderDM 👍

delta1:
Included in #3416 thanks @StriderDM !

PR #3409, opened 2021-10-04
fix: don't display an error when there is no message to be read in tari_mining_node

StriderDM:
Description
---
Fixes a minor bug introduced by accident in PR#3396 causing terminal and logs to be peppered with errors for tari_mining_node (mining against a pool) where no action should be taken instead.

Motivation and Context
---
Prevents the following message from being incorrectly reported:
```
ERROR Error reading message: None
```

How Has This Been Tested?
---
Manually

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4820,7 +4820,7 @@ dependencies = [
 
 [[package]]
 name = "tari_wallet_ffi"
-version = "0.18.7"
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
utAck

stringhandler (approved):

PR #3411, opened 2021-10-04
feat: tx weight takes tariscript and output features into account [igor]

sdbondi:
Description
---
- adds consensus encoding for tari script and output features
- igor: tx weight takes TariScript and OutputFeatures metadata into
  account as per #3368
- igor: weights updated as per #3368
- add `TransactionWeight` parameters for `weatherwax` and `igor` networks
  and set them in consensus constants
- set max block weight for `igor` (+95% adjust to allow similar number of
  max inputs/outputs/kernels ~2Mb)
- decouple chain_storage from `consensus_contstants` module to allow
  wallet to use it
- wallet: minor simplification/dedup code of coin split logic
- wallet: default fee per gram depending on network
- wallet and base node tests updated as needed

Motivation and Context
---
See #3368

How Has This Been Tested?
---
Existing tests updated, manually checked normal and one-sided transaction sending between console wallets/mobile on watherwax and igor


base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -220,17 +233,18 @@ impl MempoolStorage {
     }
 
     /// Gathers and returns the stats of the Mempool.
-    pub fn stats(&self) -> Result<StatsResponse, MempoolError> {
+    pub fn stats(&mut self) -> Result<StatsResponse, MempoolError> {
```
SWvheerden, line 236: Why did we make this mut, this call is only supposed to read the mempool stats

base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -220,17 +233,18 @@ impl MempoolStorage {
...
             unconfirmed_txs: self.unconfirmed_pool.len(),
             reorg_txs: self.reorg_pool.len()?,
-            total_weight: self.unconfirmed_pool.calculate_weight(),
+            total_weight: self.unconfirmed_pool.calculate_weight(&weight),
         })
     }
 
     /// Gathers and returns a breakdown of all the transaction in the Mempool.
-    pub fn state(&self) -> Result<StateResponse, MempoolError> {
+    pub fn state(&mut self) -> Result<StateResponse, MempoolError> {
```
SWvheerden, line 247: Why did we make this mut, this call is only supposed to read the mempool state

base_layer/core/src/mempool/priority/mod.rs
```diff
@@ -21,10 +21,10 @@
...
-
-// Public re-exports
 pub use error::PriorityError;
+
+mod prioritized_transaction;
 pub use prioritized_transaction::{FeePriority, PrioritizedTransaction};
-pub use timelocked_transaction::{TimelockPriority, TimelockedTransaction};
+
+// mod timelocked_transaction;
+// pub use timelocked_transaction::{TimelockPriority, TimelockedTransaction};
```
SWvheerden: remove?

SWvheerden (commented):
Looks good

base_layer/core/src/mempool/mempool_storage.rs
sdbondi: Because of the way TtlCache works, specifically, the way it clears out expired items when you call some normally read-only methods like `len()` and `get()`. Before we had a double lock on the mempool storage which doesn't do anything. The only downside is a need for a write lock on the mempool instead of just a read when calling stats and state - but these calls aren't performance critical (i.e. pretty much only called from the bn prompt)

base_layer/core/src/mempool/mempool_storage.rs
sdbondi: As above

base_layer/core/src/mempool/priority/mod.rs
sdbondi: This code was dead before (not completely sure why there weren't any warnings). I wasn't sure if anyone would desperately miss it 😆   will remove

base_layer/core/src/transactions/transaction.rs
```diff
@@ -108,9 +111,21 @@ pub struct OutputFeatures {
 }
 
 impl OutputFeatures {
-    pub fn to_bytes(&self) -> Vec<u8> {
-        let mut buf = Vec::new();
-        bincode::serialize_into(&mut buf, self).unwrap(); // this should not fail
+    /// Encodes output features for a v1 block
+    pub fn to_v1_bytes(&self) -> Vec<u8> {
```
stringhandler, line 118: I think we should add a `version` to the output or perhaps OutputFeatures so that we know which version was used in the hash etc.

base_layer/core/src/transactions/transaction.rs
```diff
@@ -130,6 +145,29 @@ impl OutputFeatures {
     }
 }
 
+impl ConsensusEncoding for OutputFeatures {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, io::Error> {
+        let mut written = self.flags.consensus_encode(writer)?;
+        written += writer.write_varint(self.maturity)?;
```
stringhandler, line 154: Maybe something like this:
```suggestion
        written += writer.write_varint(1)?; // version so that it can be decoded with the correct version
        written += writer.write_varint(self.maturity)?;
```

base_layer/core/src/transactions/transaction.rs
sdbondi: Yeah think that makes sense - will do - I'll modify the hashes depending on the output version

sdbondi: The downside is additional storage and it turns this PR into a breaking one, blockchains will need to be resynced so not too bad

sdbondi: The default impl for  `OutputFeatures` no longer makes sense as the version depends on your network -> block version, which entails lots of changes. 

The intention was to remove the v1/v2 concept and just use consensus encoding at a convenient point (like a testnet reset).  But it may. be worth putting in the time to include "transaction/UTXO versioning" while maintaining backward-compatibility. Perhaps for another PR since this one is currently non-breaking for weatherwax?

base_layer/core/src/transactions/transaction.rs
sdbondi: Cool will add / hard code to 1

stringhandler (dismissed):

stringhandler (dismissed):

stringhandler (approved):

PR #3418, opened 2021-10-05
fix: auto update cucumber tests, reduced timeout

Impala123:
Description
---
This fixes issues with the auto update tests.

Motivation and Context
---
Make it work

How Has This Been Tested?
---
Using cucumber.

I did add a string to the dns.rs query_txt() result to actually test, since I can't modify DNS TEXT records...

push "base-node:macos-aarch64:99.8.12:6e77da12535e0ed7851761ae6c6729f2dd3ae8050840e8c08e10e279dadfde8a"

maybe would could add the macos-aarch64 to the DNS record?


stringhandler (approved):
Seems good. untested

delta1:
Hi @Impala123. Waiting on CI but in the meantime please update your PR title in line with the conventional commits spec 

https://www.conventionalcommits.org/en/v1.0.0/#summary

PR #3420, opened 2021-10-05
feat!: new transaction and output validation protocol

philipr-za:
Description
---
This PR is to merge the `tx-validation2` feature branch into development. This branch has been used to develop a new Output and Transaction validation protocols in the Wallet. This process relies on new functionality in the Base Node and as such for this branch’s features to work the seed nodes will need to be upgraded to use the code in this PR.

### Changes to the Base Node Wallet RPC service to support the new validation approach
- Add `query_deleted` RPC method to allow the wallet to request whether an output is deleted i.e. spent and in which block it was deleted
- Add `utxo_query` RPC method to allow for querying the status of a UTXO in more detail including what block it was mined in and the current tip state to aid in detecting reorgs
- Add a new index to the LMDB database mapping deleted MMR position to block height and hash.

### Changes to Wallet
- Output validation is now a single protocol that validates spent, unspent and invalid outputs replacing three separate protocols.
- We now store the block data (height and hash) of where an output was spent/received to make it possible to easily detect when a reorg has occurred. by iterating backwards through ordered outputs we can efficiently determine which outputs are affected by the reorg rather than revalidating all of them.
- Transaction Validation has been updated to replace the old Chain Monitoring protocol. Again by adding block data to each transaction it becomes possible to more efficiently detect reorg events and which transactions are affected by the reorg. This means we only have to revalidate those transactions and not all of them.
- We integrated the new Wallet Connectivity service into the new validation protocols to use the RPC connection pool instead of creating and managing our own RPC connections in the protocols.
- Validations are now triggered when the Wallet Base Node Service becomes aware of a new block rather than on a periodic interval.

### Wallet FFI
- TXO validation now only has a single launching method and a single feedback callback.

How Has This Been Tested?
---
Cucumber tests and Rust tests have been updated to work with this new features.


philipr-za:
Rather going to merge the feature branch than this squashed commit

PR #3424, opened 2021-10-06
feat: allow tor proxy to be bypassed for outbound tcp connections

sdbondi:
Description
---
- Adds `tor_proxy_bypass_for_outbound_tcp` setting that allows direct
  outbound connections to be made for TCP nodes, bypassing tor
- Mobile will have this set for now until it can be configured by the apps

Motivation and Context
---
By default, the most private mode is set and all traffic is routed through tor. With this option set 
outbound tcp traffic bypasses tor for outbound TCP peer addresses in exchange for better 
bandwidth and latency.

How Has This Been Tested?
---
Some unit tests + manually


stringhandler (approved):
Cool. Need to probably make it clear in the mobile apps that they are doing that

sdbondi:
Ah yeah sorry was a bit rushed with the description - happy to keep it like that or do you think it should be turned off?

stringhandler:
I think it's fine, but we need to add a config that can be set in the mobile apps

Issue #3423, opened 2021-10-05
tari_base_node: UnknownError on Fedora 34


therustmonk:
**Describe the bug**
Node failed with `UnknownError` on Fedora 34.

**To Reproduce**

1. Run `tor`:

```bash
$ sudo -u toranon tor --allow-missing-torrc --ignore-missing-torrc --clientonly 1 --socksport 9050 --controlport 127.0.0.1:9051 --log "notice stdout" --clientuseipv6 1
```

2. Run `tari_base_node`:

```bash
cargo run --bin tari_base_node
```

3. Node failed with:

```bash
Running `target/debug/tari_base_node`
Initializing logging according to "/home/denis/.tari/config/log4rs_base_node.yml"
00:10 WARN  DNS seed `seeds.weatherwax.tari.com` failed to resolve: Client failure: request timed out
00:10 ERROR Network discovery errored: Connectivity error: Internal actor response was unexpectedly cancelled. Waiting for 5s
00:10 WARN  Failed to mark offline time: JoinError(JoinError::Cancelled)
00:10 ERROR DhtConnectivity exited with error: ConnectivityError(ActorDisconnected)
UnknownError
00:10 ERROR Exiting with code (102): UnknownError
```

4. And there is a warning in the `tor`'s logs:

```bash
Oct 05 18:12:12.000 [notice] New control connection opened from 127.0.0.1.
Oct 05 18:12:12.000 [warn] Got authentication cookie with wrong length (0)
```

**Desktop (please complete the following information):**
Linux fedora 5.14.9-200.fc34.x86_64 #1 SMP Thu Sep 30 11:55:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Tor version 0.4.5.10.

therustmonk:
I already solved it and put this issue to help other Fedora users.

Fedora sets the value of **CookieAuthentication** parameter (of tor) to **1** in the `/etc/tor/torrc` config file.
Changing it to **0** and restarting `tor` solves the issue, but I'm not sure it's a secure way to solve that.

`man tor` fragment:

> **CookieAuthentication 0|1**
    If this option is set to 1, allow connections on the control port when the connecting process knows the contents of a file named "control_auth_cookie", which Tor will create in its data directory. This authentication method should only be used on systems with good filesystem security. (Default: 0)

delta1:
Nice find. There is an option to provide a tor password or cookie in config.toml so you don’t have to run tor without auth 👍 will link to it later

[...]

PR #3427, opened 2021-10-06
refactor: detailed error messages for users

therustmonk:
Description
---
This refactoring lets us improve the user experience if a node failed.

Motivation and Context
---
Related to #3423. These changes help us to fix the following issues.

Duplications like this:
[applications/tari_app_utilities/src/identity_management.rs#L58](https://github.com/tari-project/tari/tree/6578d1eb885578761803b6278123ad154c677d53/applications/tari_app_utilities/src/identity_management.rs#L58)

```rust
error!(
    target: LOG_TARGET,
    "Node identity information not found. {}. You can update the configuration file to point to a \
     valid node identity file, or re-run the node with the --create-id flag to create a new \
     identity.",
    e
);
return Err(ExitCodes::ConfigError(format!(
    "Node identity information not found. {}. You can update the configuration file to point to a \
     valid node identity file, or re-run the node with the --create-id flag to create a new \
     identity.",
    e
)));
```

Avoid logging in converting methods and let us derive `From` automatically:
[applications/tari_app_utilities/src/utilities.rs#L130](https://github.com/tari-project/tari/tree/6578d1eb885578761803b6278123ad154c677d53/applications/tari_app_utilities/src/utilities.rs#L130)

```rust
impl From<tari_common::ConfigError> for ExitCodes {
    fn from(err: tari_common::ConfigError) -> Self {
        error!(target: LOG_TARGET, "{}", err); 
        Self::ConfigError(err.to_string())
    }
}
```

Handle potential panics:
[applications/tari_app_utilities/src/identity_management.rs#L201](https://github.com/tari-project/tari/tree/6578d1eb885578761803b6278123ad154c677d53/applications/tari_app_utilities/src/identity_management.rs#L201)

```rust
pub fn save_as_json<P: AsRef<Path>, T: MessageFormat>(path: P, object: &T) -> Result<(), String> {
    let json = object.to_json().unwrap();
    if let Some(p) = path.as_ref().parent() {
```

_Not everything can be converted to JSON, and if we change the format of an object (and will have numbers as keys of maps, etc.), we can have panic there._

Later we can also implement the standard [`Termination`](https://doc.rust-lang.org/std/process/trait.Termination.html) trait (not stabilized yet) for this:
[applications/tari_app_utilities/src/utilities.rs#L86](https://github.com/tari-project/tari/tree/6578d1eb885578761803b6278123ad154c677d53/applications/tari_app_utilities/src/utilities.rs#L86)

```rust
impl ExitCodes {
    pub fn as_i32(&self) -> i32 {
        match self {
            Self::ConfigError(_) => 101,
            Self::UnknownError => 102,
            Self::InterfaceError => 103,
```

_And use `ExitCodes` as returned `Error` of the `main` function._

Improve specific errors checking:
[applications/tari_base_node/src/main.rs#L220](https://github.com/tari-project/tari/blob/6578d1eb885578761803b6278123ad154c677d53/applications/tari_base_node/src/main.rs#L220)

```rust
// todo: find a better way to do this
if boxed_error.to_string().contains("Invalid force sync peer") {
```

_It's the potential issue if we will change the error's message and forget to sync the text above._

How Has This Been Tested?
---

Not tested yet 🙃 (to be updated...)

### DIFF
--- a/applications/tari_app_utilities/src/identity_management.rs
+++ b/applications/tari_app_utilities/src/identity_management.rs
@@ -23,7 +23,13 @@
 use crate::utilities::ExitCodes;
 use log::*;
 use rand::rngs::OsRng;
-use std::{clone::Clone, fs, path::Path, string::ToString, sync::Arc};
+use std::{
+    clone::Clone,
+    fs,
+    path::{Path, PathBuf},
+    string::ToString,
+    sync::Arc,
+};
 use tari_common::configuration::bootstrap::prompt;
 use tari_common_types::types::PrivateKey;
 use tari_comms::{multiaddr::Multiaddr, peer_manager::PeerFeatures, NodeIdentity};
@@ -31,6 +37,7 @@ use tari_crypto::{
     keys::SecretKey,
     tari_utilities::{hex::Hex, message_format::MessageFormat},
 };
+use thiserror::Error;
 
 pub const LOG_TARGET: &str = "tari_application";
 
@@ -144,7 +151,7 @@ pub fn create_new_identity<P: AsRef<Path>>(
     path: P,
     public_addr: Multiaddr,
     features: PeerFeatures,
-) -> Result<NodeIdentity, String> {
+) -> Result<NodeIdentity, SaveAsJsonError> {
     let private_key = PrivateKey::random(&mut OsRng);
     let node_identity = NodeIdentity::new(private_key, public_addr, features);
     save_as_json(path, &node_identity)?;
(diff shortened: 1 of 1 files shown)


applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -165,29 +172,48 @@ pub fn recover_node_identity<P: AsRef<Path>>(
...
-        return Err(format!(
-            "Identity file, {}, does not exist.",
-            path.as_ref().to_str().unwrap()
-        ));
+        Err(LoadFromJsonError::NotExist(path.as_ref().to_owned()))
+    } else {
+        let contents = fs::read_to_string(path)?;
+        let object = T::from_json(&contents)?;
+        Ok(object)
     }
```
sdbondi, line 204: nit: not conditional with `else`

```rust
// meh
if !something {
dont_do_something();
} else {
do_something();
}

// Good
if something {
do_something();
} else {
dont_do_something();
}
```

IMHO it was better with the return since the purpose of `load_from_json` is over as soon as the path does not exist and if that is not the case then the function it probably doing too much, but I don't mind this at all #justnittin

sdbondi (approved):
Looks good!

sdbondi:
Suggest that `LoadFromJsonError` and `SaveToJsonError` can become one error type for the module.

stringhandler:
Already implemented in a different way

PR #3428, opened 2021-10-06
fix: allow env override of mining node pool settings

stringhandler:
Description
---
These configs were not override-able by env vars


philipr-za (approved):
LGTM

PR #3430, opened 2021-10-07
ci: auto patch crate versions on push to development

delta1:
Description
---

Adds github action to bump patch versions in all crates in the workspace on push to development branch. This will not push git tags nor releases, that is left as a manual task. 

Since wallet_ffi is already on `0.19.0`, this means all crates will move to `0.19.1` after this is implemented. Although a large jump, arguably better to get all versions matching and less confusion with wallet_ffi releases. 

- commits will be signed, **requires a new repo secret** containing the gpg secret key: `TARI_DEPLOY_SECRET_KEY`

Example commit after a successful PR merge here: https://github.com/delta1/tari/commit/13a9635d26fd81b5f8c439fcd18c64242ea877b6
Example CI run: https://github.com/delta1/tari/runs/3904018033?check_suite_focus=true 

(examples updated 2021-10-15 10:44)


.github/workflows/version.yml
```diff
@@ -0,0 +1,36 @@
...
+          pwd
+          /tmp/cws workspaces list
+      - name: bump patch version
+        run: |
+          /tmp/cws workspaces version patch \
+          --no-git-tag \
+          --no-global-tag \
+          --force "*" \
+          -y \
+          -m "ci: bump crate versions to %v" \
```
stringhandler:
```suggestion
          -m "ci: bump crate versions to %v [skip ci]" \
```

base_layer/p2p/Cargo.toml
```diff
@@ -10,13 +10,13 @@ license = "BSD-3-Clause"
 edition = "2018"
 
 [dependencies]
-tari_comms = { version = "^0.10", path = "../../comms" }
```
stringhandler: The versions on some crates are needed if you publish the crate to crates.io. Publishing ignores the path and the version must be present (and be a crate itself)

stringhandler (commented):
Two things: 
1. Please put the versions back in the lib crates that are on crates.io already. Versions in bin crates can be removed
2. Are the version commits signed?

base_layer/p2p/Cargo.toml
delta1: Oh I see, will try that I think there was an issue

delta1:
> Two things:
> 
> 1. Please put the versions back in the lib crates that are on crates.io already. Versions in bin crates can be removed
> 2. Are the version commits signed?

1. Will do so 
2. No I don’t think they are signed, will have to set up a new identity and repo secret etc. Will revert

.github/workflows/version.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: crazy-max/ghaction-import-gpg@v4
+        with:
+          gpg_private_key: ${{ secrets.TARI_DEPLOY_SECRET_KEY }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          git_push_gpgsign: false
+      - name: install
+        run: |
+          echo "this is a lot faster than building from source 🤷"
+          curl -o /tmp/cws -L https://github.com/delta1/cargo-workspaces/releases/download/v0.2.24/cargo-workspaces-v0.2.24-x86_64-linux-gnu
```
stringhandler: Happy to merge, but needs to be compiled from source and the commit locked down to a specific hash, otherwise it's possible to compromise that binary

delta1: Updated to use the release from [tari-project/cargo-workspaces](/tari-project/cargo-workspaces) and SHA hash is checked

stringhandler (approved):

PR #3435, opened 2021-10-08
fix: fix flakey rust tests

philipr-za:
Description
---
This PR aims to fix the following flakey tests:
1. `test_txo_validation`
2. `test_transaction_resending`

The fix for 1 was to put in an explicit wait for the validation protocol to complete, on the slower CI system is seems that the validation protocol was slow to update the database state before the main test queried the balance.

The fix for 2 is to increase the cool down and resend times to make them more resilient to the slower CI environment. The current times were quite tight which was fine on a local dev machine. This might still prove to be flakey in the future.

How Has This Been Tested?
---
These are fixes to tests


base_layer/wallet/tests/transaction_service/service.rs
```diff
@@ -3905,11 +3905,11 @@ fn test_transaction_resending() {
         .unwrap();
 
     assert!(alice_outbound_service
-        .wait_call_count(1, Duration::from_secs(4))
+        .wait_call_count(1, Duration::from_secs(5))
         .is_err());
 
-    // Wait for the cooldown to expire but before the resend period has elapsed see if a repeat illicts a reponse.
-    runtime.block_on(async { sleep(Duration::from_secs(2)).await });
+    // Wait for the cooldown to expire but before the resend period has elapsed see if a repeat illicts a response.
```
delta1, line 3911: lol

```suggestion
    // Wait for the cooldown to expire but before the resend period has elapsed see if a repeat elicits a response.
```

delta1 (approved):
LGTM untested

base_layer/wallet/tests/transaction_service/service.rs
stringhandler: lol

stringhandler (approved):

PR #3436, opened 2021-10-08
test: fix cucumber tests and config on validator node branch

stringhandler:


delta1 (approved):
LGTM as a first step, untested

PR #3437, opened 2021-10-08
fix!: remove consensus::encode::Encodable impl to upgrade monero-rs to 0.15.0

StriderDM:
Description
---
This PR changes FixedByteArray to FixedByteArray<N> and removes the Encodable trait where implemented in our codebase.

It also implements serialize and deserialize for [T,N] so that [u8;MAX_ARR_SIZE] can be used for the elem field in FixedByteArray struct.

Motivation and Context
---
Monero-rs has sealed the Encodable trait which prevents it from being implemented outside of the crate itself.

How Has This Been Tested?
---
cargo test --all

### DIFF
--- a/base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
+++ b/base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
@@ -21,40 +21,81 @@
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::crypto::tari_utilities::ByteArrayError;
-use monero::{
-    consensus::{encode, Decodable, Encodable},
-    VarInt,
-};
-use std::{convert::TryFrom, io, ops::Deref};
+use serde::{Deserialize, Serialize};
+use std::{convert::TryFrom, ops::Deref};
 use tari_crypto::tari_utilities::ByteArray;
 
-const MAX_ARR_SIZE: usize = 63;
+mod arrays {
+    use std::{convert::TryInto, marker::PhantomData};
 
-#[derive(Clone, Debug)]
-pub struct FixedByteArray {
-    elems: [u8; MAX_ARR_SIZE],
+    use serde::{
+        de::{SeqAccess, Visitor},
+        ser::SerializeTuple,
+        Deserialize,
+        Deserializer,
+        Serialize,
+        Serializer,
+    };
+    pub fn serialize<S: Serializer, T: Serialize, const N: usize>(data: &[T; N], ser: S) -> Result<S::Ok, S::Error> {
+        let mut s = ser.serialize_tuple(N)?;
+        for item in data {
+            s.serialize_element(item)?;
+        }
+        s.end()
+    }
+
+    struct ArrayVisitor<T, const N: usize>(PhantomData<T>);
+
+    impl<'de, T, const N: usize> Visitor<'de> for ArrayVisitor<T, N>
+    where T: Deserialize<'de>
+    {
+        type Value = [T; N];
+
+        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+            formatter.write_str(&format!("an array of length {}", N))
+        }
+
+        #[inline]
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where A: SeqAccess<'de> {
+            let mut data = Vec::with_capacity(N);
+            for _ in 0..N {
+                match (seq.next_element())? {
+                    Some(val) => data.push(val),
+                    None => return Err(serde::de::Error::invalid_length(N, &self)),
+                }
+            }
+            match data.try_into() {
+                Ok(arr) => Ok(arr),
+                Err(_) => unreachable!(),
+            }
+        }
+    }
+    pub fn deserialize<'de, D, T, const N: usize>(deserializer: D) -> Result<[T; N], D::Error>
+    where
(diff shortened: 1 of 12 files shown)


base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -163,33 +181,11 @@ mod test {
 
     #[test]
     fn serialize_deserialize() {
-        let data = consensus::serialize(&FixedByteArray::from_hex("ffffffffffffffffffffffffff").unwrap());
-        assert_eq!(data.len(), 13 + 1);
-        let arr = consensus::deserialize::<FixedByteArray>(&data).unwrap();
+        let data = bincode::serialize(&FixedByteArray::from_hex("ffffffffffffffffffffffffff").unwrap()).unwrap();
+        println!("{:?}", data);
```
philipr-za, line 185:
```suggestion
```

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -73,15 +114,15 @@ impl FixedByteArray {
     }
 }
 
-impl Deref for FixedByteArray {
+impl Deref for FixedByteArray<MAX_ARR_SIZE> {
```
philipr-za, line 117: Feels like you can make these implementations Generic N instead of for the specific MAX_ARR_SIZE and then just define that size where it is used.

sdbondi (commented):
Nice RE FixedArray, didn't know that const generics were stable - cool :) Though I think some `impl` blocks should remain size-agnostic as @philipr-za said - thought IMO not NB. 

I don't agree that these consensus-breaking changes are necessary. Consensus encoding to ensure that we have byte-for-byte control of the data in the `pow_data` field. Nothing prevents `bincode` from intentionally/unintentionally changing its binary format . I think the right approach to consensus critical encoding was already implemented here.  

I think for now we can simply take the `Encodable` trait from the crate over into the `monero_rx` module. Since it is part of Tari consensus IMO it makes sense to be a separate trait anyway. As a side note: We can still use `monero` crate's consensus encoding impl for each field in `MoneroPowData` as we have done, though technically we are coupling Monero's consensus to Tari's MM consensus which may/probably not be a minor inconvenience in future. 

Other context:

In #3411 I add a Tari-specific trait for consensus encoding that is pretty much exactly the same as `monero` crates'.  

The `Decodable` trait is still available. The `Encodable` trait was sealed because an invariance could not be preserved in `monero::consensus::encode::serialize` (see https://github.com/monero-rs/monero-rs/pull/48#issuecomment-869578047)

StriderDM:
@sdbondi Thanks, appreciated. I'll update it to be size-agnostic.

The breaking change comes from removing the encodable trait from fixed array and MoneroPowData, so it is only a breaking change on our side of the merged mining.

I do see your point with bincode, I don't see a problem with using consensus::{serialize,deserialize} from the monero crate, this can likely be used instead (think it was at one point recently if memory serves), but only for the monero structures within MoneroPowData. 

I don't think tightly coupling this is a good idea, especially since the monero structures can be easily stored as hex and then their serialization/deserialization is independent of MoneroPowData structure.

I'll see if I can rework this a bit more and look at the PR you mentioned again. Though not sure we can get around it being a breaking change entirely.

philipr-za:
I think that we should definitely be aiming for the philosophy that anything that could affect the formatting of data in a way that can break consensus should be brought explicitly under our control.

Whether or not this PR break consensus or not is not really the issue it is the fact that before we were using a serialization that Monero considered consensus critical and wouldn't change out of hand. This PR now uses Bincode for serialization and there is nothing stopping Bincode from changing its formatting out from under us which would break consensus with no action on our part. 

I don't think it is necessary to use monero's serialize/deserialize for this purpose, as we can see it can cause problems being coupled to an external crate. So I agree with @sdbondi that we should rather bring the `Encodable` trait into our `monero_rx` module and implement the serialization explicitly rather than delegating to bincode. @mikethetike thoughts?

sdbondi:
@StriderDM Cool no problem :)

> I don't think tightly coupling this is a good idea, especially since the monero structures can be easily stored as hex and then their serialization/deserialization is independent of MoneroPowData structure.

Hex is a base16 representation of binary data i.e does not enforce any data format on the data it represents. It is not appropriate to store in a blockchain header because it doubles byte size of the encoding vs the raw binary data.

The coupling to the `monero` crates structs is easily undone if necessary by copying the structs locally / creating local structs as needed that are encoded in the same binary format. I think it would be a good change but I maintain that this is not something we have to worry about now or even post-mainnet.

> Though not sure we can get around it being a breaking change entirely.

There is no reason for a breaking change here because of the monero crate upgrade. The only thing we might want to consider making a breaking change here is to add a version field for more flexibility later. Though, second thoughts, that is already covered by the Tari header `version` and the monero BlockHeader `major/minor` fields.

StriderDM:
@sdbondi Ok. We can do it the way you suggested if needed, no problem. Was under the impression we were going to keep it separate though I see this isn't the case. Closing this PR.

sdbondi:
@StriderDM I like the `FixedByteArray` changes and think that we should upgrade the `monero` crate, but there is no reason to break current consensus is all. You'd get my approval if you reopened inline with those goals.

philipr-za:
Yes, I agree with what this ticket set out to do but the way that it broke consensus was just a red flag. I think this should be reopened, and the encodable stuff brought into our code.

stringhandler:
@StriderDM please fill in the Motivation and Context section for this PR. It's not currently clear to me why this is needed.

PR #3438, opened 2021-10-08
fix: remove unnecessary wallet dependency

therustmonk:
Description
---
Removes `wallet` dependency from `tari_app_grpc` and `tari_app_utilities`.

Motivation and Context
---
`grpc` and `utilities` are common and don't contain relations with the wallet.
Dependency has been removed because it affects building time and complicates maintenance.

How Has This Been Tested?
---
CI only, because the changes don't affect any feature and change locations of some types only.


therustmonk:
curious, but it also related to #3427 and partially fixes it

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -729,9 +724,12 @@ impl TransactionBackend for TransactionServiceSqliteDatabase {
         let mut coinbase_txs = CompletedTransactionSql::index_coinbase_at_block_height(block_height as i64, &conn)?;
         for c in coinbase_txs.iter_mut() {
             self.decrypt_if_necessary(c)?;
-            let completed_tx = CompletedTransaction::try_from(c.clone()).map_err(|_| {
-                TransactionStorageError::ConversionError("Error converting to CompletedTransaction".to_string())
-            })?;
+            // TODO: WARNING! The change hides the intermediate error!
+            // Use `CompletedTransactionConversionError` ASAP!!!
+            let completed_tx = CompletedTransaction::try_from(c.clone())?;
```
sdbondi, line 881: The previous implementation hid the cause of the error so I would call your implementation an improvement :)

base_layer/wallet/src/error.rs
```diff
@@ -85,6 +86,16 @@ pub enum WalletError {
     UtxoScannerError(#[from] UtxoScannerError),
 }
 
+pub const ERROR_LOG_TARGET: &str = "tari::application";
```
sdbondi: It's `LOG_TARGET` everywhere else in the code base

```suggestion
pub const LOG_TARGET: &str = "tari::application";
```

sdbondi (dismissed):
Thanks for the PR - this definitely helps with some much needed code untwisting/decoupling - tested and LGTM

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
therustmonk: Ah, right. I'll improve that 👌

therustmonk:
rebased to `development`

stringhandler (dismissed):
Nice

sdbondi (approved):
LGTM

PR #3442, opened 2021-10-10
fix: u64->i64->u64 conversion; chain split height as u64

Impala123:
Description

removing an unnecessary conversion and using u64 for chain height 
---

How Has This Been Tested?
---

rust build & test


applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -326,15 +326,15 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
 
         let headers: Vec<u64> = if request.from_height != 0 {
             match sorting {
-                Sorting::Desc => ((cmp::max(0, request.from_height as i64 - num_headers as i64 + 1) as u64)..=
+                Sorting::Desc => (cmp::max(0, request.from_height - num_headers + 1)..=
```
stringhandler: Is there not a chance for an underflow here?

delta1: This needs a covering test, even if this logic is moved into a function.

delta1:
Seems you need to run cargo fmt

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
Impala123: should this then be  

`let headers: Vec<u64> = if request.from_height >= num_headers {`

num_headers is forced to be between 10 and 10_000
Means that from_height being 0 would go to the else

Or is it more desirable to check the underflow flag with overflowing_sub() ?

stringhandler (approved):

Issue #3410, opened 2021-10-04
Node should validate incoming SAF messages


SWvheerden:
A node should verify all incoming SAF messages if they are legit by either, or:

- Checking if they come from a nearby neighbor and request others
- OR preferably only accepting incoming SAF messages if it requested such messages.

By not checking these messages, it allows a malicious actor to control what the node downloads by preemptively forwarding messages to it. If SAF message can be used to "break" the node, it allows an attacker to kill a node by simply sending a message to it.

sdbondi:
Fixed in #3444 - only requested SAF responses are accepted

PR #3444, opened 2021-10-11
fix: check SAF message inflight and check stored_at is in past

### LINKED ISSUE #3412: Node should verify DHT message field "stored_at"
A node should verify and add some checking to the "stored_at" field of a SAF message and not just blindly use this as the "LastSafMessageReceived" field as this can cause a node to poison its own "LastSafMessageReceived" so that it never receives any SAF messages again.

sdbondi:
Description
---
- Keeps track of inflight SAF requests and only accepts responses for
  requests that are inflight
- Checks that `stored_at` is in the past
- Fixes #3412, #3410 

Motivation and Context
---
See #3412, #3410

How Has This Been Tested?
---
- New/existing unit/integration tests
- memorynet
- Manually


SWvheerden (approved):
LGTM, 

This does seem to somewhat limit the poisoning that can happen to the `LastSafMessageReceived` variable, it can still happen. But we are still trusting in, information received and calculated by another peer.

sdbondi:
Yeah if we look at the way it previously worked (before using a timestamp sent by a remote peer), there was no stored_at field but you essentially trusted that the peer gave you the latest messages, then marked the timestamp when the node shut down so that you would request messages from that time (while online you would receive propagated messages). There was still trust involved even without a timestamp. In the end, the protocol has to have the required rules to prevent abuse. 

I think that message deliverability could be improved by only updating timestamp (based on offline timestamp or stored_at) only after the appropriate peers have been queried rather than just a single peer, this would also further limit "poisoning". I tend to think that this would be easier to implement with a comms RPC interface in a similar way to block sync et al.

stringhandler (approved):

PR #3445, opened 2021-10-11
fix: add display_currency_decimal method

therustmonk:
Description
---
The inner type of `Tari` changed from `f64` to `decimal_rs::Decimal`.
Also, `display_currency` method was replaced with `format_currency` that adds separators to formatted strings.

Motivation and Context
---
`display_currency` method forces us to use `f64` that is not accurate for finance calculations.

How Has This Been Tested?
---
Updated unit tests.
**TODO:** Smoke test: run the wallet and the node.

### DIFF
--- a/applications/tari_console_wallet/src/automation/error.rs
+++ b/applications/tari_console_wallet/src/automation/error.rs
@@ -26,7 +26,7 @@ use chrono_english::DateError;
 use log::*;
 use tari_app_utilities::utilities::ExitCodes;
 use tari_core::transactions::{
-    tari_amount::{MicroTariError, TariConvertError},
+    tari_amount::{MicroTariError, TariConversionError},
     transaction::TransactionError,
 };
 use tari_wallet::{
@@ -44,8 +44,8 @@ pub const LOG_TARGET: &str = "wallet::automation::error";
 pub enum CommandError {
     #[error("Argument error - were they in the right order?")]
     Argument,
-    #[error("Tari value convert error `{0}`")]
-    TariConvertError(#[from] TariConvertError),
+    #[error("Tari value conversion error `{0}`")]
+    TariConversionError(#[from] TariConversionError),
     #[error("Transaction service error `{0}`")]
     TransactionError(#[from] TransactionError),
     #[error("Transaction service error `{0}`")]
--- a/applications/tari_console_wallet/src/automation/error.rs
+++ b/applications/tari_console_wallet/src/automation/error.rs
@@ -25,7 +25,10 @@ use std::num::{ParseFloatError, ParseIntError};
 use chrono_english::DateError;
 use log::*;
 use tari_app_utilities::utilities::ExitCodes;
-use tari_core::transactions::{tari_amount::MicroTariError, transaction::TransactionError};
+use tari_core::transactions::{
+    tari_amount::{MicroTariError, TariConvertError},
+    transaction::TransactionError,
+};
 use tari_wallet::{
     error::{WalletError, WalletStorageError},
     output_manager_service::error::OutputManagerError,
@@ -41,6 +44,8 @@ pub const LOG_TARGET: &str = "wallet::automation::error";
 pub enum CommandError {
     #[error("Argument error - were they in the right order?")]
     Argument,
+    #[error("Tari value convert error `{0}`")]
+    TariConvertError(#[from] TariConvertError),
     #[error("Transaction service error `{0}`")]
     TransactionError(#[from] TransactionError),
     #[error("Transaction service error `{0}`")]
--- a/base_layer/core/src/transactions/tari_amount.rs
+++ b/base_layer/core/src/transactions/tari_amount.rs
@@ -23,7 +23,7 @@
 use newtype_ops::newtype_ops;
 use serde::{Deserialize, Serialize};
 
-use crate::transactions::helpers::display_currency;
+use crate::transactions::helpers;
 use std::{
     fmt::{Display, Error, Formatter},
     iter::Sum,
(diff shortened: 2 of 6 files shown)


therustmonk:
The `experiment` commit contains `Trai(Decimal)` type.
It still uses `f64` all around, but we can replace it with Decimals. It should work.

Or how about removing `Tart(f64)` completely and use `MiniTari` everywhere with some special formatting methods?

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -219,17 +229,20 @@ impl From<Tari> for FormattedTari {
...
 }
 
 /// A convenience struct for representing full Tari. You should **never** use Tari in consensus calculations, because
 /// Tari wraps a floating point value. Use MicroTari for that instead.
-#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
-pub struct Tari(f64);
-
-newtype_ops! { [Tari] {add sub} {:=} Self Self }
-newtype_ops! { [Tari] {mul div rem} {:=} Self f64 }
+#[derive(
```
therustmonk, line 245: `newtype-ops` looks unmaintained. How about using `derive_more` instead?

therustmonk:
It's also possible to remove `Tari` type completely and use `MicroTari` with a special formatted like this:

```rust
pub struct TariFormat {
    MicroTari,
    Tari,
    Auto,
}

pub struct FormattedTari {
    format: TariFormat,
    separator: Option<char>,
}

impl MicroTari {
    pub fn formatted(&self, format: TariFormat, separator: Option<char>) -> FormattedTari {
        FormattedTari {
            format,
            separator,
        }
    }
}

impl Display for FormattedTari { /* etc. */ }
```

What do you think?

applications/tari_console_wallet/src/automation/error.rs
```diff
@@ -41,6 +44,8 @@ pub const LOG_TARGET: &str = "wallet::automation::error";
 pub enum CommandError {
     #[error("Argument error - were they in the right order?")]
     Argument,
+    #[error("Tari value convert error `{0}`")]
+    TariConvertError(#[from] TariConvertError),
```
delta1:
```suggestion
    #[error("Tari value conversion error `{0}`")]
    TariConversionError(#[from] TariConversionError),
```

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -140,26 +143,29 @@ impl std::str::FromStr for MicroTari {
...
-                .parse::<f64>()
+                .parse::<Decimal>()
                 .map_err(|e| MicroTariError::ParseError(e.to_string()))
                 .map(|v| {
-                    if v < 0.0 {
+                    if v.is_sign_negative() {
                         Err(MicroTariError::ParseError("value cannot be negative".to_string()))
                     } else {
-                        Ok(MicroTari::from(Tari::from(v.max(0.0))))
+                        // TODO: Check. It can't be `NaN` anymore. Still we need `.max(0.0)` check?
```
delta1, line 159: is this comment still current?

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -140,26 +143,29 @@ impl std::str::FromStr for MicroTari {
...
                 })?
         }
     }
 }
 
+// TODO: Implement `TryFrom` instead
 impl From<Tari> for MicroTari {
     fn from(v: Tari) -> Self {
-        MicroTari((v.0 * 1e6) as u64)
+        MicroTari((v.0 * 1_000_000u32).try_into().unwrap())
```
delta1: unwrap?

base_layer/core/src/transactions/tari_amount.rs
delta1: Yeah if we're already using `derive_more` now and it's easy to remove `newtype_ops` I think that's fine

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -243,28 +256,28 @@ impl Display for Tari {
     }
 }
 
-impl From<Tari> for f64 {
-    fn from(v: Tari) -> Self {
-        v.0
-    }
-}
+pub type TariConvertError = DecimalConvertError;
```
delta1:
```suggestion
pub type TariConversionError = DecimalConvertError;
```

delta1 (changes requested):
Looking good, just a few small things

base_layer/core/src/transactions/tari_amount.rs
therustmonk: No idea, why we compare the unsigned value with `0.0` 🤷‍♂️ I left the comment to double-check it together. How about removing that `max` call?

base_layer/core/src/transactions/tari_amount.rs
therustmonk: In the past, the value could be truncated silently, I thought it's better to panic here, but probably even better to replace the trait to `TryFrom`.

base_layer/core/src/transactions/tari_amount.rs
delta1: It looks like you did remove the max? Which is why I ask :) 

image

therustmonk: Ah, right, it's another one. I've removed it, because if a value is not less than 0.0 (greater or equal 0.0) it's always a win in `max(0.0)` comparison. I guess we had it to handle the `f64::NaN` case, but `Decimal` doesn't support `NaN` values at all and that comparison doesn't make sense anymore. If I'm correct about why we had that `max(0.0)` call.

therustmonk: https://docs.rs/decimal-rs/0.1.20/src/decimal_rs/parse.rs.html#266
https://docs.rs/decimal-rs/0.1.20/src/decimal_rs/parse.rs.html#321

stringhandler (approved):
Would have preferred to remove this function completely, but this is an improvement

PR #3446, opened 2021-10-11
feat: add sql query to obtain balance

hansieodendaal:
Description
---
- Replaced the `get_balance` function containing multiple sql queries with a single raw sql query.
- Removed redundant code (`fn fetch_pending_outgoing_outputs`) as a result of this change.
- Added more test points for time-locked balance.
- **Note:** To properly test transaction validation in `async fn test_txo_validation()`, access to the backend to obtain pending incoming transactions is needed via ` fn fetch_pending_incoming_outputs`, although that function is not used in production code anymore. If it is also removed other methods will have to be added to the backend to obtain the data for testing. Retaining the current function was chosen in lieu of adding other code.

Motivation and Context
---
Get balance used a lot of RAM and was really slow due to multiple database interactions.

Comparison of old vs. new query time performance for a wallet with 251,000 UTXOs in the database shown below:

- **Full scale**

image

- **Y-axis zoomed in**

image

How Has This Been Tested?
---
- Unit tests
- System level tests

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
@@ -23,6 +23,7 @@
 use crate::{
     output_manager_service::{
         error::OutputManagerStorageError,
+        service::Balance,
         storage::{
             database::{DbKey, DbKeyValuePair, DbValue, KeyManagerState, OutputManagerBackend, WriteOperation},
             models::{DbUnblindedOutput, KnownOneSidedPaymentScript, OutputStatus},
@@ -38,7 +39,7 @@ use crate::{
 };
 use aes_gcm::{aead::Error as AeadError, Aes256Gcm, Error};
 use chrono::{NaiveDateTime, Utc};
-use diesel::{prelude::*, result::Error as DieselError, SqliteConnection};
+use diesel::{prelude::*, result::Error as DieselError, sql_query, SqliteConnection};
 use log::*;
 use std::{
     convert::{TryFrom, TryInto},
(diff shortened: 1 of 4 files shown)


base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -1047,6 +1036,104 @@ impl OutputSql {
...
+        struct BalanceQueryResult {
+            #[sql_type = "diesel::sql_types::BigInt"]
+            amount: i64,
+            #[sql_type = "diesel::sql_types::Text"]
+            category: String,
+        }
+        let balance_query_result = if let Some(val) = tip {
+            let balance_query = sql_query(
+                "SELECT coalesce(sum(value), 0) as amount, 'available_balance' as category \
+                 FROM outputs WHERE status = ? \
```
stringhandler, line 1051: Should take the mined heights into account when the tip is specified
```suggestion
                 FROM outputs WHERE status = ?  and mined_height < $val and  deleted_at_height > $val\ 
```

philipr-za: I think this is a job for the validation code though, I think it is cleaner that this function is based purely on the status.

philipr-za (approved):
LGTM

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -1047,6 +1036,104 @@ impl OutputSql {
             .first::<OutputSql>(conn)?)
     }
 
+    /// Return the available, time locked, pending incoming and pending outgoing balance
+    pub fn get_balance(tip: Option<u64>, conn: &SqliteConnection) -> Result<Balance, OutputManagerStorageError> {
```
philipr-za, line 1040: It will be clearer what the `tip` argument is for if we rename it to something like `tip_for_timelock_calculation`. Currently it is not clear what the `tip` is for and that it actually affects whether the timelock balance is returned or not. The comment should also make the behaviour of the function clear when the tip is provided or not.

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
```diff
@@ -1047,6 +1036,104 @@ impl OutputSql {
...
+        #[derive(QueryableByName, Clone)]
+        struct BalanceQueryResult {
+            #[sql_type = "diesel::sql_types::BigInt"]
+            amount: i64,
+            #[sql_type = "diesel::sql_types::Text"]
+            category: String,
+        }
+        let balance_query_result = if let Some(val) = tip {
```
philipr-za, line 1048: the generic `val` here makes reading the code difficult below.

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
hansieodendaal: Will make work these two comments into a new small PR after the merge.

base_layer/wallet/src/output_manager_service/storage/sqlite_db.rs
hansieodendaal: As above

PR #3439, opened 2021-10-08
chore: improve cucumber tags and run time speed

SWvheerden:
Description
---

This PR improves the run time of certain tests to remove `waits` and make sure they have async functions to retest waitable events.
It adds a `@base-node` tag to all base node tests.
It adds a `@wallet` tag to all wallet tests.
It adds `@`  tags to all feature files.
Its adds and remove `@critical` tags to all required tests, adding where critical functionality is not tested and removing where it is double tested and or not critical to core functionality.

Motivation and Context
---
Made sure we can run all tests as short as possible. Tests are grouped with the following tags:
`@base-node` which can be used to run tests if verification of code base node functionality is desired
`@wallet` which can be used to run tests if verification of code wallet functionality is desired
`@critical` if a short test of all core functionality is desired.
`@long-running` if really detailed long-running tests are desired. Can take hours. 

Tests without a `@critical` and `@long-running` are tests that are required for passing CI, but will take longer to run.
Running test with `--tags @base-node AND @critical` or `--tags @wallet AND @critical` completes it run time below 10 minutes.

PR #2836, opened 2021-04-07
[tests] Attempt to replicate sync issue in integration tests

delta1:
## Description

- Update the cucumber mining node process to accept `mine_on_tip_only` environment variable
- Fix some issues in mining node regarding max-blocks, min/max difficulty, and add some extra logging
- New integration test that attempts to reproduce a chain corruption issue when mining against a node that is still doing an initial sync

## Motivation and Context

Related to #2829 and issues found manually when mining against a syncing node

## How Has This Been Tested?

Manually and attempted cucumber test

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Feature refactor (No new feature or functional changes, but performance or technical debt improvements)
* [x] New Tests
* [ ] Documentation

## Checklist:

* [x] I'm merging against the `development` branch.
* [x] I ran `cargo-fmt --all` before pushing.
* [ ] I ran `cargo test` successfully before submitting my PR.
* [x] I have squashed my commits into a single commit.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #3448, opened 2021-10-12
fix: add sanity checks to prepare_new_block

sdbondi:
Description
---

- Add basic sanity checks to `prepare_new_block`
- Check calc_median_timestamp invariant holds when used in `prepare_new_block`
- `get_new_block` GRPC call returns invalid argument error if invalid arguments are passed to `prepare_new_block` 
- Add new unit tests (tests previous panic) 

Motivation and Context
---

Possible panic observed in cucumber test is prevented 

```
Testing scenario "Verify all coinbases in hybrid mining are accounted for"
.seed: SEED_A
........[mmProxy client] Tip at 0 ...(stopping at 100)
stderr: thread 'tokio-runtime-worker' panicked at 'calc_median_timestamp: timestamps cannot be empty', base_layer/core/src/validation/helpers.rs:79:5

stderr: stack backtrace:

stderr:    0: std::panicking::begin_panic

stderr:    1: tari_core::validation::helpers::calc_median_timestamp
   2: tari_core::chain_storage::blockchain_database::BlockchainDatabase<B>::prepare_new_block
   3: tari_core::chain_storage::async_db::trace_log
   4: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut

stderr:    5: tokio::runtime::task::harness::Harness<T,S>::poll

stderr:    6: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

[mmProxy client] Tip at 35 ...(stopping at 100)
[mmProxy client] Tip at 95 ...(stopping at 100)
```

How Has This Been Tested?
---
New unit tests + existing tests


delta1 (dismissed):
Nice LGTM

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -664,21 +664,67 @@ where B: BlockchainBackend
...
         let min_height = header.height.saturating_sub(
             self.consensus_manager
                 .consensus_constants(header.height)
                 .get_median_timestamp_count() as u64,
         );
+
         let db = self.db_read_access()?;
-        let timestamps = fetch_headers(&*db, min_height, height)?
+        let tip_header = db.fetch_tip_header()?;
+        if header.height != tip_header.height() + 1 {
```
SWvheerden, line 688: While I am not against this check, I think this might break the "mine_on_tip_only" coonfig flag when this is set to false as this check might break then.  We either need to remove this check or remove that config/feature.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -664,21 +664,67 @@ where B: BlockchainBackend
...
+                func: "prepare_new_block",
+                arg: "template",
+                message: format!(
+                    "Expected new block template height to be {} but was {}",
+                    tip_header.height() + 1,
+                    header.height
+                ),
+            });
+        }
+        if header.prev_hash != *tip_header.hash() {
```
SWvheerden, line 699: See the above comment, if we just remove that check and keep the config/feature option then we can just check this against the prev hash if it exists and the height is correct. 

If we remove the config option, then this is fine

integration_tests/features/WalletMonitoring.feature
```diff
@@ -135,8 +135,8 @@ Feature: Wallet Monitoring
     And I have mining node MINER2 connected to base node NODE2 and wallet WALLET2
 
     When I co-mine <numBlocks> blocks via merge mining proxy PROXY1 and mining node MINER2
-    # This wait is here to give a chance for re-orgs to settle out
-    Then I wait 30 seconds
+#     This wait is here to give a chance for re-orgs to settle out
+    And I wait 30 seconds
```
SWvheerden: you can remove this line, this PR: #3439 removes this

SWvheerden (changes requested):
One of the sanity checks goes against a current config option/feature. We need to remove one of the two.

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: I think the answer here is to ensure that the height given is not more than `tip + 1` rather than not equal to.
Though, if the panic was not there it will fail another check afterwards (CannotCalculateNonTipMmr) - because we no longer have the full DeletedBitmap at each height (or would have to build it slowly) to construct the MMR. So I would say that if you wanted to mine behind the tip you should configure your base node not to sync to the tip somehow. I think that miner setting relates to whether is waits for the base node to indicate it's synced or not - which is slightly different.

SWvheerden: I had a look at the code, it does check this. This function is triggered on the option being true:
https://github.com/tari-project/tari/blob/278ac9353f525cb625f816313abed2aaf8c3b4db/applications/tari_mining_node/src/main.rs#L324

The fact that you mentioned it will fail later on means this is not used and we should remove this option

My opinion for the way forward would be to remove that option. Perhaps replace it with a `mine_without_initial_sync` or something that only checks the initial sync flag. And then always call this function.

sdbondi: AFAIK the miner never  tries to ask for a block template behind the tip - which is where these checks would come in. Not checking that the base node is synced before starting mining is a separate point, we can discuss removing that option in another PR I think - I'm sure there was a reason it was added, but on the face of it I don't see it as a must-have feature. (added in #2836) Also wouldn't say these checks break that feature.

Extra blah blah:

Remember that `prepare_new_block` (via the `get_new_block` grpc call) is typically (always) called with the a template from the get_next_block_template call. So if the base node is not synced, but you set the miner to mine it could obtain a valid new block template behind the "real tip", mine on that and submit it where the block would be added to the orphan pool / reorged in if stronger. There is obviously a race between the call to `get_next_block_template` and `get_new_block` and a new block coming in / more blocks being synced but that is the case anyway.

SWvheerden: I think you are right, while this would have technically broken the `mine_on_tip_only` config option, but that option does not and cannot work as it stands currently.  
The way the base_node and mempool work, it is not possible to get a template with transactions that is not at the local tip of the base_node. 

We should remove that config option, and if desired add a config option that lets you mine before you are synced or so. But that's not part of this PR and it should not be.

SWvheerden (approved):
LGTM

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: > while this would have technically broken the mine_on_tip_only config option

As above, it won't break it because the base node decides what the tip is. The option just determines if the miner checks it before mining or immediately asks for a template even if the bn isn't synced which could be mined and submitted (though would probably orphan)

You've asked me out of band to resolve this - so resolving :)

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Resolved in other comments

stringhandler (approved):

PR #3449, opened 2021-10-13
feat: improve logging for tari_mining_node

StriderDM:
Description
---
This PR improves the logging for tari_mining_node by making use of the log4rs target, it also sets the default logging level to info.
This PR also adds basic statistics for pool mining which are automatically displayed at a 20s interval.

Motivation and Context
---

How Has This Been Tested?
---
Manually


philipr-za (commented):
Tested this out and I have two comments:
Firstly, the stats are only displayed when a share is found and between finding a share there is no activity on the terminal at all. I would prefer a periodic output of the stats including the number of threads being used.

Secondly, the output is still linked to the log level of the miner directly. This means the log file also only has the minimal Info in it and if you change this level to Debug the terminal is flooded. Rather have a unique Target for the terminal output and then pipe the actual logs into the file.

philipr-za (commented):
Looking way better. Just one question, the estimated Sol/s number seems off? What unit is in? I think just include the unit for clarity.

StriderDM:
It is (S/s), which is the share rate as seen in the pool UI. Only counting the solutions found rather than how many hashes the miner is doing.

philipr-za:
> It is (S/s), which is the share rate as seen in the pool UI. Only counting the solutions found rather than how many hashes the miner is doing.

Def need to put hashrate, this is the key metric that a Miner cares about and tries to optimize.

philipr-za (approved):

PR #3450, opened 2021-10-13
WIP: Validate before dedup

Impala123:
Description
---
Reorder the layers to do validate / decrypt before dedup.

Motivation and Context
---
With the current order of layers, a malicious node could tamper with a message which would then be discarded by the validation layer. However the dedup cache currently stores this before it is discarded by validate. Thus any un-tampered version of the same message would no longer be processed.

How Has This Been Tested?
---
WIP, compiles but test fails, presumably because of the mockup message being invalid


sdbondi:
Nice - looking good. I've tested with the `memorynet` example (`cargo run --example memorynet --release`) and all looks like it's working. 

There are two more items to call this done (they can go in a subsequent PR if preferred): 

- [x] A failing test for the original issue. I would look to write this as a DHT crate integration test here `comms/dht/tests/dht.rs` with this scenario

```
create valid message `m1`
`NodeA` sends `m1'` to `NodeB` where `m1'` has an invalid DHT header but with same origin mac/identical body (not even sure if this is possible - maybe clearing/changing the ephemeral_pubkey field would do it)
`NodeA` sends `m1` to `NodeB`
assert that `NodeB` has received `m1` (i.e. has not deduped it) 
``` 

- [x] include the `origin_mac` field in the `hash_inbound_message` hash - this is so a malicious peer cannot craft an faux-encrypted message that passes validation and enters dedup cache but has the same body as the message meant for that peer, causing the peer to dedup the subsequent original message.

PR #3452, opened 2021-10-13
test: create cucumber test directory if necessary

CjS77:
The `run-tests.sh` script fails on a fresh install because the cucumber
output folder doesn't exist. This PR simple checks and creates the
folder if necessary.


delta1 (approved):
Good catch

stringhandler (approved):

PR #3455, opened 2021-10-14
chore: replace proxy mining with sha mining in tests

SWvheerden:
Description
---
Replaced merge mining proxy mining in most tests with sha3 miner. 
Added base_node state to grpc for use in the tests

Motivation and Context
---
The monero merge mining proxy makes a network call to the outside which can be unreliable. The sha3 miner only uses local information which is faster as well. 

We use sleeps that in the cucumber code to ensure that base_nodes have time to sync, we can just if they are in listening state to check if we can go to the next step

How Has This Been Tested?
---
Manual cucumber tests

### DIFF
--- a/integration_tests/features/Mempool.feature
+++ b/integration_tests/features/Mempool.feature
@@ -96,7 +96,7 @@ Feature: Mempool
     Then SENDER has TX1 in NOT_STORED state
     Then SENDER has TX2 in MINED state
 
-  @critical
+  @critical @flaky
   Scenario: Mempool clearing out invalid transactions after a reorg
     Given I do not expect all automated transactions to succeed
     Given I have a seed node SEED_A
--- a/integration_tests/features/Reorgs.feature
+++ b/integration_tests/features/Reorgs.feature
@@ -167,6 +167,8 @@ Feature: Reorgs
     And I connect node NODE_A1 to node NODE_A3 and wait 1 seconds
     And I connect node NODE_A2 to node NODE_A4 and wait 1 seconds
     And I connect node SEED_A1 to node SEED_A2 and wait <SYNC_TIME> seconds
+    Then node SEED_A1 is in state LISTENING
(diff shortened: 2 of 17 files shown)


integration_tests/features/Mempool.feature
```diff
@@ -96,7 +96,7 @@ Feature: Mempool
     Then SENDER has TX1 in NOT_STORED state
     Then SENDER has TX2 in MINED state
 
-  @critical
+  @critical @flaky
```
stringhandler, line 99: How is this regressing now?

stringhandler (commented):
I'm not keen on this. It makes the code more complicated instead of simpler. 

There are a number of tests that are now marked as flaky, an indication that this is a regression

integration_tests/features/Mempool.feature
SWvheerden: Its a new test, created in PR #3439 
It failed on CI this morning and I have not been able to get it to fail locally.

philipr-za (commented):
This is looking good. I like the addition of waiting on the base node state. One comment and we will tackle Flakey tests in the next sprint.

integration_tests/features/Reorgs.feature
```diff
@@ -197,6 +199,8 @@ Feature: Reorgs
     And I connect node NODE_B1 to node NODE_B3 and wait 1 seconds
     And I connect node NODE_B2 to node NODE_B4 and wait 1 seconds
     And I connect node SEED_B1 to node SEED_B2 and wait <SYNC_TIME> seconds
```
stringhandler, line 201: Perhaps we can remove this `wait <SYNC_TIME> seconds` and then keep the below

stringhandler (approved):
Ok, looking at this again I see that the state check generally replaces a wait X seconds. I think this can go in.

PR #3457, opened 2021-10-14
test: simplify cucumber tests

stringhandler:
Description
---
* removed cucumber 'I expect all transactions to succeed' steps
* better logging of warnings
* added default profile for cucumber tests

Motivation and Context
---
I wanted to remove long running tests from someone who runs a default `npm test` or `run-tests.sh`.

Also, there was a number of times where the base node would stall not output, while generating many warnings in the logs. This would cause circle ci to eventually kill it with "too long with no output". This looks like the problem was a bad block template getting created.

How Has This Been Tested?
---
npm test


philipr-za (commented):
Seems to have broken something in the FFI tests

philipr-za (approved):

PR #3459, opened 2021-10-15
fix: remove is_synced check for transaction validation

stringhandler:
Description
---
Remove the check that aborts transactions validation if the base node is not synced.

Motivation and Context
---
With this check in place, if the base node is syncing even one block when the wallet asks, it will abort the check. There most likely no relation between the block that the base node is syncing and the data that the wallet is asking for, and the base node may still be able to respond correctly. The logic in tx_validation already handles the fact that a base node may be in progress with a sync, and aborting it early is not necessary.

How Has This Been Tested?
---
Manually


philipr-za (dismissed):
LGTM

philipr-za (approved):

PR #3460, opened 2021-10-15
feat!: provide a compact form of TransactionInput

philipr-za:
Description
---

This PR updates the TransactionInput struct to support containing the duplicated Output data of the output it is spending OR just the hash of the output. This is done in the original TransactionInput as opposed to creating a new struct to allow for easier reuse of all the code that already supports the TransactionInput.

Methods are provided to:
- Create a new compact or full TransactionInput
- Return a compact form of the current TransactionInput
- Provide the output data that is referenced in the TransactionInput

This PR also updates the RPC definitions and all the conversion methods to support both forms of the TransactionInput

Motivation and Context
---

The reason for this change is to reduce the amount of duplicate data stored in the Blockchain db and sent across the wire during Block sync. Currently the input is stored with all the duplicated data from the output it is spending along side the spent output in the database which is a duplication of data. Now the TransactionInput is converted into its compact form with just a reference to the spent output before being stored in the database and before being sent across the wire. When the Input is read from the database or received then the associated output data is retrieved from the local datastore.

TODO: apply the compact form of the TransactionInput to transactions submitted to the base node mempool. This will further reduce data sent from wallets to the base nodes.

How Has This Been Tested?
---
cargo test

### DIFF
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -169,6 +169,8 @@ message TransactionInput {
     ComSignature script_signature = 7;
     // The offset public key, K_O
     bytes sender_offset_public_key = 8;
+    // The hash of the output this input is spending
+    bytes output_hash = 9;
 }
 
 // Output for a transaction, defining the new ownership of coins that are being transferred. The commitment is a
--- a/applications/tari_app_grpc/src/conversions/aggregate_body.rs
+++ b/applications/tari_app_grpc/src/conversions/aggregate_body.rs
@@ -27,14 +27,16 @@ use tari_utilities::convert::try_convert_all;
 
 use crate::tari_rpc as grpc;
 
-impl From<AggregateBody> for grpc::AggregateBody {
-    fn from(source: AggregateBody) -> Self {
-        Self {
+impl TryFrom<AggregateBody> for grpc::AggregateBody {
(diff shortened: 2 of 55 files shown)


base_layer/core/src/transactions/transaction.rs
```diff
@@ -430,13 +474,48 @@ impl TransactionInput {
...
+
+    /// Accessor method for the commitment contained in an input
+    pub fn script(&self) -> Result<&TariScript, TransactionError> {
+        match self.spent_output {
+            SpentOutput::OutputHash(_) => Err(TransactionError::MissingTransactionInputData),
+            SpentOutput::OutputData { ref script, .. } => Ok(script),
+        }
+    }
+
+    /// Accessor method for the commitment contained in an input
```
delta1: duplicated comments

stringhandler (commented):
Initial impression is that it looks good, but possibly worth testing on igor or a different network for a bit. Maybe we can set a wire mode?

delta1 (dismissed):
Fantastic work 👏 

i did some system level testing with a new chain on `igor`: 
- 2 base nodes 
- 2 wallets
- Mined with mining node 
- Merge mined with xmrig and proxy 
- Sent and mined transactions between wallets
- Sent and mined one-sided transactions between wallets

PR #3461, opened 2021-10-15
chore: improve cucumber tests to wait for broadcast

SWvheerden:
Description
---
Changed connect node to wait for both nodes to listening
Changed send transactions to wait for broadcast state. 
Removed some flaky and long-running tags.

Motivation and Context
---
This will test to run faster as more stable. Current SHA3 miner mines too fast for the wallets to finish sending. This PR market wait for the transaction to be broadcast before continuing. 
This makes the connect node to await listing state of both nodes to avoid using waits.

How Has This Been Tested?
---
Manual test of tests

### DIFF
--- a/integration_tests/features/support/steps.js
+++ b/integration_tests/features/support/steps.js
@@ -237,25 +237,29 @@ Given(
 );
 
 Given(
-  /I connect node (.*) to node (.*) and wait (.*) seconds/,
+  /I connect node (.*) to node (.*)/,
   { timeout: 1200 * 1000 },
-  async function (nodeNameA, nodeNameB, waitSeconds) {
-    expect(waitSeconds < 1190).to.equal(true);
+  async function (nodeNameA, nodeNameB) {
     console.log(
       "Connecting (add new peer seed, shut down, then start up)",
       nodeNameA,
       "to",
-      nodeNameB,
-      ", waiting for",
-      waitSeconds,
-      "seconds"
+      nodeNameB
     );
     const nodeA = this.getNode(nodeNameA);
     const nodeB = this.getNode(nodeNameB);
     nodeA.setPeerSeeds([nodeB.peerAddress()]);
     await this.stopNode(nodeNameA);
     await this.startNode(nodeNameA);
-    await sleep(waitSeconds * 1000);
+    await waitFor(
+      async () => {
+        let node_a_result = (await nodeA.get_node_state()) === "LISTENING";
+        let node_b_result = (await nodeB.get_node_state()) === "LISTENING";
+        return node_a_result && node_b_result;
+      },
+      true,
+      30 * 1000
+    );
   }
 );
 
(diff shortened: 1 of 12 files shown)


stringhandler (dismissed):

integration_tests/features/support/steps.js
```diff
@@ -1192,16 +1196,18 @@ Then(
...
+    this.lastResult = await waitFor(
+      async () => {
+        let tx_result = await client.transactionStateResult(sig);
+        console.log(
+          `Node ${node} response is: ${tx_result}, should be: ${pool}`
+        );
+        return tx_result === pool;
+      },
+      true,
+      1200 * 1000
```
hansieodendaal: **NIT**: This change - `20 * 60 * 1000` to `1200 * 1000` - makes it less obvious that the `waitFor` timeout for just less than the function timeout.

integration_tests/features/support/steps.js
```diff
@@ -237,25 +237,29 @@ Given(
...
-    await sleep(waitSeconds * 1000);
+    await waitFor(
+      async () => {
+        let node_a_result = (await nodeA.get_node_state()) === "LISTENING";
+        let node_b_result = (await nodeB.get_node_state()) === "LISTENING";
+        return node_a_result && node_b_result;
+      },
+      true,
+      30 * 1000
+    );
```
hansieodendaal, line 262: I am not sure that this will always be a guarantee that the nodes are connected and synced, as was the intention of the previous variable wait time after restart. Ideally, we should compare the best block hash of each node.

integration_tests/features/support/steps.js
```diff
@@ -1805,6 +1864,22 @@ When(
...
+          result =
+            result && sourceClient.isTransactionAtLeastBroadcast(tx_ids[i]);
+        }
+        return result;
+      },
+      true,
+      60 * 1000,
+      5 * 1000,
+      5
+    );
```
hansieodendaal, line 1881: This is a nice improvement overall where we do not want to send transactions as fast as possible.

integration_tests/features/support/steps.js
SWvheerden: The idea behind this is not to test if they are synced. It is to check if the nodes have checked their state and the other node's state and are happy with the state. 
For example, if the nodes have the same difficulty, they will not have the same block hash, but they will both be in listening mode. If you want to check if they are on the same tip, there is a step for that.

hansieodendaal: **Edit**: ~I still think it will be more beneficial to verify that nodes are then either synced to the same best block or have the same difficulty, which will make the tests more resilient. It is the same idea as verifying that transactions are broadcast before finishing the send transaction step.~ 

Maybe something like a successful 'dial-peer' both ways then? - Just an idea for a future PR.

philipr-za (approved):
LGTM

PR #3464, opened 2021-10-17
test: use TCP node for daily sync test

sdbondi:
Description
---
- uses TCP node (london) for sync tests
- `--force-sync-peers` option now takes the pubkey/address of the peer
  rather than a seed peer index
 - Bypass tor for TCP in sync test
 - Display sync peer in status line for header sync 
 - cleanup peer seeds and force sync peer JS array usage
 - make peer seeds optional (default to empty Vec) without having to specify a default for every network 
 - fix `When a new node joins the network, it should receive all peers` cucumber

Motivation and Context
---
The primary goal of the daily sync test is to provide visibility into the performance of the sync protocol.
Directly connecting to the same TCP node, bypassing the tor network, the network performance under which the sync protocol runs is more consistent.

How Has This Been Tested?
---
Run locally


stringhandler (dismissed):

sdbondi:
Just noticed that many cucumber tests failed, probably something in this PR - will check it out

sdbondi:
All cucumbers pass now - fixed this one in last commit
image

stringhandler (approved):

PR #3465, opened 2021-10-18
fix: typo in console wallet

selsta:
Noticed it while trying out the console wallet.


delta1 (approved):
Thanks!

stringhandler (approved):
Look to good me

PR #3466, opened 2021-10-18
fix: sha256sum isn't available on all *nix platforms

selsta:
`sha256sum` is part of GNU coreutils, which isn't available by default on e.g. macOS or FreeBSD. `shasum -a256` should work everywhere.


delta1 (approved):
LGTM, tested on macOS 👍

stringhandler (approved):
Awesome! Thanks

PR #3467, opened 2021-10-18
fix: remove unbounded vec allocations from base node grpc/p2p messaging

### LINKED ISSUE #3310: RPC command fetch_headers  generates in memory vector
The RPC command `fetch_headers` on the server-side will generate a vector of block headers in memory before trying to send it. 
Assuming the current tip height of around 20900, this will be of size 4.953 MB. Assuming only Sha3 blocks which are smaller than Monero blocks. This will grow in size as the blockchain grows, assuming 4mb max packets, to roughly block height 500 000. The server then needs a vector of size 119mb to keep all the headers in.

sdbondi:
Description
---
- replaces unbounded height lists with bound height ranges in internal base node service
  interfaces and p2p messaging
- removes unused and deprecated base node p2p messaging requests/responses
- load chain headers when requesting headers, this simplifies and
  reduces the database load when calling grpc get_blocks
- use iterator with almost no footprint for paging in grpc calls
- implement `DoubleSidedIterator` for `NonOverlappingIntegerPairIter`
- add overflow protection to `NonOverlappingIntegerPairIter` and
  additional tests

This PR does not contain any breaking changes, as no nodes using the base node p2p messaging interface except for block propagation (which is unchanged). GRPC protobuf contract is preserved.

**New dependency**
`either = "1.6.1"` added to `tari_base_node` crate and used for its `Either` iterator impl

Motivation and Context
---
- removing unused "outbound comms interface" request/responses reduces attack surface (TODO: remove this interface entirely - specifically the coupling for internal service requests and external p2p comms is troublesome, but mostly removed in the PR, but removing it entirely will make making changes to the base node service much easier)
- remote peers could request any number of heights allowing an external party to allocate memory without bounds 

Closes #3310 

How Has This Been Tested?
---
Existing GRPC tests
Manually tested block explorer api connected to local base node
Additional unit tests


base_layer/core/src/base_node/comms_interface/local_interface.rs
```diff
@@ -82,22 +83,31 @@ impl LocalNodeCommsInterface {
     }
 
     /// Request the block header of the current tip at the block height
-    pub async fn get_blocks(&mut self, block_heights: Vec<u64>) -> Result<Vec<HistoricalBlock>, CommsInterfaceError> {
+    pub async fn get_blocks(
+        &mut self,
+        start: u64,
```
stringhandler: I would much prefer the use of range here to avoid inclusive/exclusive inconsistency in methods
```suggestion
        range: Range<u64>,
```

base_layer/core/src/base_node/comms_interface/local_interface.rs
```diff
@@ -82,22 +83,31 @@ impl LocalNodeCommsInterface {
...
         }
     }
 
-    /// Request the block header of the current tip at the block height
-    pub async fn get_headers(&mut self, block_heights: Vec<u64>) -> Result<Vec<BlockHeader>, CommsInterfaceError> {
+    /// Request the block headers with the given range of heights. The returned headers are ordered from lowest to
+    /// highest block height
+    pub async fn get_headers(
+        &mut self,
+        start: u64,
```
stringhandler: Same

stringhandler (commented):
Nice, but let's rather change the `start` and `end` to use a range and avoid inconsistencies

base_layer/core/src/base_node/comms_interface/local_interface.rs
sdbondi: Makes sense, I originally tried to make it generic on `RangeBounds<u64>` but that won't work since `RangeBounds<u64>` cannot be made into a trait object.  Think it makes sense to force the use of `RangeInclusive` syntax e.g.`0..=10`. Will change

stringhandler (dismissed):

stringhandler (approved):

PR #3473, opened 2021-10-19
perf: set MDB_NOLOCK for peer and blockchain database

sdbondi:
Description
---
- Use `MDB_NOLOCK` flag for LMDB environment for peer db and blockchain db

Motivation and Context
---
In both cases we use `RwLock`s to manage concurrent access so any overhead on locking on LMDB is not required

From lmdb-zero:

> NOLOCK - Don't do any locking. If concurrent access is anticipated, the caller must manage all concurrency itself. For proper operation the caller must enforce single-writer semantics, and must ensure that no readers are using old transactions while a writer is active. The simplest approach is to use an exclusive lock so that no readers may be active at all when a writer begins.   

How Has This Been Tested?
---
Existing tests and manually


sdbondi:
Testing this, appears to work but DNM for now

stringhandler:
Interesting idea

sdbondi:
Double checked that LMDB transaction rollbacks still work. I ran a single test (`test_handle_possible_reorg_accum_difficulty_is_correct_case_1`) a few times with LMDB locking on and off, and was not really able to see any meaningful difference in execution time. Of course, this is not a good way to measure performance. That said, I have not observed any issues with the base node / comms with locking turned off so perhaps we should give this a try if only to save on the cost of [the reader table](http://www.lmdb.tech/doc/group__readers.html) 

> No reader table is used if the database is on a read-only filesystem, or if MDB_NOLOCK is set.

stringhandler:
Can merge after #3460

stringhandler (approved):

PR #3475, opened 2021-10-19
feat(wallet_ffi)!: add get_balance callback to wallet ffi

hansieodendaal:
Description
---
- Added get_balance callback to the wallet ffi callback handler that fires only if the balance has actually changed.
- Expanded the wallet ffi callback handler test framework to include a mock output manager request-response server.
- _**Update:** Added required methods and interfaces to the cucumber test framework._
- _**Update:** Fixed flaky wallet FFI cucumber tests._
- _**Update:** Fixed a bug in the wallet FFI cucumber test framework where the return type of `ref.types.ulonglong` did not correspond to the Rust return type and had a memory alignment problem._
- _**Update:** Fixed an issue whereby on a fast 8-core multi-tasking computer (e.g. AMD Ryzen 7 2700X) some of the wallet FFI cucumber tests did not complete properly after the test and went into an endless wait. The root cause of this issue has been traced down to incorrect use of synchronous calls to wallet FFI destroy methods where in actual fact those methods have async behaviour._
- _**Update:** Re-applied #3271._

~~The following anomaly exists when compiling the proposed `wallet_ffi/tests` module:~~
```
24 | use tari_wallet_ffi::callback_handler::CallbackHandler;
   |     ^^^^^^^^^^^^^^^ use of undeclared crate or module `tari_wallet_ffi`
```

_**Update**_ ~~Various code organizations have been tried, all with the same result. As an alternative, a working test and output manager service mock has been added into the test module in `callback_handler.rs`. Hopefully, the anomaly can be fixed. Duplicate code will be removed before the final commit.~~

Motivation and Context
---
- Mobile wallet efficiency.
- Resilient wallet FFI cucumber tests.

How Has This Been Tested?
---
- Expanded the current FFI `test_callback_handler` unit test.
- _**Update:** Ran all the wallet FFI cucumber tests to verify the new callback is working properly when using the wallet FFI library:_ 

```
2021-10-21T06:29:32.160Z callbackTransactionValidationComplete(9123501482775375388,0)
2021-10-21T06:29:32.161Z callbackBalanceUpdated: available = 0,  time locked = 0  pending incoming = 2000000 pending outgoing = 0
2021-10-21T06:29:32.263Z received Transaction with txID 14659183447022727953

...

2021-10-21T06:31:38.358Z Transaction with txID 14659183447022727953 was mined.
2021-10-21T06:31:38.358Z callbackBalanceUpdated: available = 2000000,  time locked = 2000000  pending incoming = 0 pending outgoing = 0
2021-10-21T06:31:38.359Z callbackTransactionValidationComplete(17755253868227079780,0)
```


philipr-za (commented):
So you don't actually need the mod.rs in the test module. Check out [Rust Test Organization ](https://doc.rust-lang.org/book/ch11-03-test-organization.html).

In terms of the test module accessing the crate the problem is caused by:
```
[lib]
crate-type = ["staticlib","cdylib"]
``` 
In the cargo.toml. https://github.com/rust-lang/cargo/issues/6659 mentioned adding "lib" to that list fixes it and I checked it does.

base_layer/wallet_ffi/wallet.h
```diff
@@ -464,6 +466,7 @@ struct TariWallet *wallet_create(struct TariCommsConfig *config,
                                  void (*callback_store_and_forward_send_result)(unsigned long long, bool),
                                  void (*callback_transaction_cancellation)(struct TariCompletedTransaction *),
                                  void (*callback_txo_validation_complete)(unsigned long long, unsigned char),
+                                 void (*callback_balance_updated)(struct TariBalance *),
```
StriderDM, line 471: You need a forward declaration of `struct TariBalance;` in this file or it will result in a linker error in the application trying to link against this.

Additionally:
You will need a method to free the memory for this struct.
Methods to retrieve the values for this struct as simple C types which can safely cross the FFI boundary.
Methods to set the values for fields within this struct (appears to be optional here).
Method to allocate memory for this struct (appears to be optional here).

base_layer/wallet_ffi/Cargo.toml
```diff
@@ -43,7 +44,7 @@ default-features = false
 features = ["transactions"]
 
 [lib]
-crate-type = ["staticlib","cdylib"]
+crate-type = ["staticlib","cdylib", "lib"]
```
StriderDM: I think this will have unintended consequences.

Looks like the cause of the problem is some tests are missing:
```
#[cfg(test)]
mod test {
...
}
```

philipr-za: I think when the duplication of the tests is resolved that will no longer be an issue. The thing I am worried about is whether adding that flag affects the libWallet build at all. @StriderDM would you mind just adding the "lib" flag to the current Dev branch and see if the build goes ok, please?

base_layer/wallet_ffi/wallet.h
philipr-za: Don't need to be able to set the members of the struct or allocate memory as its only every provided by the backend. So just need to be able retrieve the members and free the memory.

This is worth doing because I see at the moment you have to make a separate DB call for each member of the balance instead of a single db call and destructuring that result.

base_layer/wallet_ffi/Cargo.toml
hansieodendaal: After discussion with @StriderDM I decided to do away with adding `"lib"` to `crate-type` to enable the separate `tests` folder and to use the internal `#[cfg(test)] mod test {...` option. I propose investigating and repackaging tests could be a separate PR if needed.

StriderDM:
Nicely done @hansieodendaal. Well done. Just needs a format and the integration test should run, approved otherwise.

hansieodendaal:
> Nicely done @hansieodendaal. Well done. Just needs a format and the integration test should run, approved otherwise.

Thank you @StriderDM; your inputs were invaluable.

StriderDM (dismissed):

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -66,6 +70,23 @@ use tari_wallet::{
 
 const LOG_TARGET: &str = "wallet::transaction_service::callback_handler";
 
+/// This macro unlocks a Mutex or RwLock. If the lock is poisoned (i.e. panic while unlocked) the last value
```
philipr-za: Not needed

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -87,14 +108,17 @@ where TBackend: TransactionBackend + 'static
...
     callback_transaction_validation_complete: unsafe extern "C" fn(u64, u8),
     callback_saf_messages_received: unsafe extern "C" fn(),
     db: TransactionDatabase<TBackend>,
     transaction_service_event_stream: TransactionEventReceiver,
     output_manager_service_event_stream: OutputManagerEventReceiver,
+    output_manager_service: OutputManagerHandle,
     dht_event_stream: DhtEventReceiver,
     shutdown_signal: Option<ShutdownSignal>,
     comms_public_key: CommsPublicKey,
+    balance_cache: Arc<Mutex<Balance>>,
```
philipr-za: You don't need a mutex here, the callback handler doesn't access this value from different tasks.

base_layer/wallet_ffi/Cargo.toml
```diff
@@ -16,6 +16,7 @@ tari_p2p = { version = "^0.11", path = "../p2p" }
 tari_wallet = { version = "^0.11", path = "../wallet", features = ["c_integration"]}
 tari_shutdown = { version = "^0.11", path = "../../infrastructure/shutdown" }
 tari_utilities = "^0.3"
+tari_service_framework = { path = "../../base_layer/service_framework" }
```
philipr-za: This should be a Dev-dependancy as it is only used in tests.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1863,6 +1864,23 @@ pub unsafe extern "C" fn completed_transaction_destroy(transaction: *mut TariCom
     }
 }
 
+/// Frees memory for a TariBalance
```
philipr-za: This should be with the other methods to do with TariBalance

integration_tests/features/WalletFFI.feature
```diff
@@ -11,13 +11,15 @@ Feature: Wallet FFI
         And I have a ffi wallet FFI_WALLET connected to base node BASE
         # It's just calling the encrypt function, we don't test if it's actually encrypted
         And I set passphrase PASSPHRASE of ffi wallet FFI_WALLET
+        And I wait 5 seconds
```
philipr-za: Why are these needed? Can they be replaces with some sort of condition check? We have been on a long quest to remove these kind of waits in favour of hard condition checks. These kind of waits are where flakiness lie.

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -545,6 +626,116 @@ mod test {
...
+    use tari_shutdown::ShutdownSignal;
+    use tari_wallet::output_manager_service::{
+        error::OutputManagerError,
+        handle::{OutputManagerRequest, OutputManagerResponse},
+    };
+    use tokio::time::Instant;
+
+    /// This macro unlocks a Mutex or RwLock. If the lock is poisoned (i.e. panic while unlocked) the last value
+    /// before the panic is used.
+    macro_rules! acquire_lock {
```
philipr-za: If we cannot put this into the tests folder then at the very least lets put this module in a separate file in this folder.

integration_tests/features/WalletFFI.feature
hansieodendaal: Removing the waits

philipr-za:
Clippy says no

hansieodendaal:
> Clippy says no

Yip, still busy sorting out a flaky test

integration_tests/features/WalletFFI.feature
hansieodendaal: I believe the tests have now been made more predictive; see the latest commit.

philipr-za (dismissed):

base_layer/wallet_ffi/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "tari_wallet_ffi"
 authors = ["The Tari Development Community"]
 description = "Tari cryptocurrency wallet C FFI bindings"
 license = "BSD-3-Clause"
-version = "0.19.0"
+version = "0.20.0"
```
stringhandler: Please don't change versions. I need to prep the changelog and this will mess with that process

stringhandler (changes requested):

base_layer/wallet_ffi/Cargo.toml
hansieodendaal: Reverted

stringhandler (dismissed):

philipr-za (dismissed):

philipr-za (dismissed):

philipr-za (approved):

PR #3479, opened 2021-10-20
fix: fix recovery test reporting message

philipr-za:
Description
---
This PR fixes the message that is reported after a successful recovery test. It was formatted like a console.log(…) call but was refactored into a new function which ignored everything after the first argument.


delta1 (dismissed):
LGTM untested

stringhandler (approved):

stringhandler (dismissed):

stringhandler (approved):

PR #3482, opened 2021-10-22
ci: delete versioning action

delta1:
Description
---
Removes the versioning action, it is no longer required.


stringhandler (approved):
rip

PR #3483, opened 2021-10-22
fix: fix fetch tokens

stringhandler:
Description
---
Fix the fetch_tokens method. 

Also removed some of the re-exposed 'crypto' and 'tari_utilities' crates from tari_core.


delta1 (approved):
LGTM, untested

PR #3486, opened 2021-10-24
feat: add submit instruction comms rpc interface

sdbondi:
Description
---
Adds `submit_instruction` comms rpc method to validator node

Motivation and Context
---
Allow others to submit instructions to VNs via p2p comms

How Has This Been Tested?
---
Single unit test


stringhandler (approved):
Nice

PR #3488, opened 2021-10-24
fix: improve responsiveness of wallet base node switching

sdbondi:
Description
---

- clear previous base node's state if a new base node is selected
- interrupt sleep in bn monitor early if base node has changed
- interrupt get_tip_info RPC call early if base node has changed
- dont trigger set peer events if same peer is set

Motivation and Context
---
When the user selects a base node, the UI could appear slow when a sleep is in progress in the base node monitor.
During this time, the previous node's latency and height are displayed, which also gives the impression of an unresponsive UI.

How Has This Been Tested?
---
Manually by switching base nodes


philipr-za (approved):
LGTM

PR #3494, opened 2021-10-25
test: change timeouts in ci to reasonable values

therustmonk:
Description
---
Reduces timeouts to reasonable values. Used max values from reports of successful passes.

Motivation and Context
---
CI takes up to 4+ hours when issues happen and we have to wait when the full timeout expires.

How Has This Been Tested?
---
CI pass expected.


stringhandler (dismissed):
I'm a little worried that these might be difficult to maintain and it's not easy for a new developer to know what values to put in, but I'm happy to try it and see how it goes. 

Please convert from draft to full pr and I'll queue it for merging

stringhandler (dismissed):

stringhandler (approved):

PR #3496, opened 2021-10-26
feat: optimize get transactions query

hansieodendaal:
Description
---
- Optimized the get transactions query (`broadcast_all_completed_transactions`) for transactions that need to be broadcast/rebroadcast by sending a single diesel SQL query that only returns the result, instead of multiple queries that return all the transactions in the database with filtering and selection in the Rust code.
- Added a new unit test 'test_get_tranactions_to_be_rebroadcast'.

Motivation and Context
---
It is much more efficient to have a SQL query perform the filtering upfront.

How Has This Been Tested?
---
Unit tests, cucumber tests.


base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1625,7 +1625,7 @@ where
         }
 
         trace!(target: LOG_TARGET, "Restarting transaction broadcast protocols");
-        self.broadcast_all_completed_transactions(broadcast_join_handles)
+        self.broadcast_selected_completed_transactions(broadcast_join_handles)
```
stringhandler: I think the previous name was better. Having 'selected' in the name suggests you are expecting a selection to be passed to it

base_layer/wallet/tests/output_manager_service/service.rs
```diff
@@ -1381,6 +1381,23 @@ async fn test_txo_validation() {
         .await
         .unwrap();
 
+    // This is needed on a fast computer, otherwise the balance have not been updated correctly yet with the next step
+    let mut event_stream = oms.get_event_stream();
```
stringhandler, line 1385: weird, but ok

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: `broadcast_all_completed_transactions` is also not good as it is misleading, We select all valid, not cancelled, completed transactions with status `Completed` or `Broadcast`. I would actually like to change it to `rebroadcast_completed_and_broadcast_transactions`. Opinion?

philipr-za: Its not always a rebroadcast though. I think its fairly implicit that it will ignore invalid and cancelled transactions personally. Definitely shouldn't have `selected` in the title.

base_layer/wallet/src/transaction_service/storage/database.rs
```diff
@@ -424,6 +426,11 @@ where T: TransactionBackend + 'static
         self.db.fetch_unconfirmed_transactions()
     }
 
+    /// This method returns all completed transactions that must be re-broadcast
```
philipr-za:
```suggestion
    /// This method returns all completed transactions that must be broadcast
```

base_layer/wallet/src/transaction_service/service.rs
stringhandler: sounds good

base_layer/wallet/tests/output_manager_service/service.rs
```diff
@@ -1381,6 +1381,23 @@ async fn test_txo_validation() {
...
+    let delay = sleep(Duration::from_secs(10));
```
philipr-za, line 1386: You need the sleep even with the wait on the event? The event is only emitted after the database is updated? I agree with waiting on the event but the sleep seems unnecessary

philipr-za (commented):
Looking good, just one question is the test

base_layer/wallet/tests/output_manager_service/service.rs
hansieodendaal: You are correct in that we do not need an actual delay before the select loop. The sleep will not be executed unless it is `sleep(Duration::from_secs(10)).await`; this is just to pass on a delay to the select loop.

stringhandler (approved):

PR #3498, opened 2021-10-26
refactor: move payload processor and add wrapper for payload

stringhandler:
Description
---
Adds a wrapper to the payload in Hotstuff Messages to allow adding fields in future. For now just added the checkpoint in preparation for committees to vote on it.

Also moved the PayloadProcessor step to prepare. If the payload processor fails because the instructions are not valid at that point, it will not vote to prepare

How Has This Been Tested?
---
run locally


delta1 (approved):
LGTM, untested

PR #3501, opened 2021-10-27
fix: improve test Wallet should display transactions made

SWvheerden:
Description
---
This test can have the wallets fail to talk to each other before sending begins which results in the wallets talking over SAF. This is a much slower communication method which can result in the step `And I send 100000 uT from wallet WALLET_A to wallet WALLET_B at fee 100` failing to reach the broadcast stage. 
This PR moves the init step of Wallet_B much earlier to increase the changes that the two wallets should communicate directly.


philipr-za (approved):
LGTM

stringhandler (approved):

PR #3502, opened 2021-10-27
test: add trace logs to wallet's base node monitor

hansieodendaal:
Description
---
- Re-added trace logs to the console wallet's base node monitor to enable profiling of slow responses.
- Consistent fixed ms format rather than significant digits for log entries were used to enable easy creation of timing graphs with post-processing of the log files.

Motivation and Context
---
These trace logs are still needed for the improvement effort of the slow console wallet responses.

How Has This Been Tested?
---
Ran cargo clippy and cargo format.


sdbondi (approved):

delta1 (changes requested):
~~why is the `package-lock.json` file deleted? this is necessary to lock dependency versions, otherwise `npm ci` may install updated versions from what was run locally~~ 

sorry I read large diff as deleted

delta1 (approved):

PR #3505, opened 2021-10-27
feat!: implement new CipherSeed and upgrade encryption KDF

philipr-za:
Description
---
This PR adds a new CipherSeed implementation for use a seed for Key Derivation. The goal of the scheme is produce a wallet seed that is versioned, contains the birthday of the wallet, starting
entropy of the wallet to seed key generation, can be enciphered with a passphrase and has a checksum. During this process it was noted that we used a naive method to derive our database encryption key from the passphrase. This PR also updates that method to use Argon2 as a proper password hashing scheme that is not vulnerable to rainbow table brute forcing and timing attacks.

- Update db encryption key generation to use Argon2 KDF
- Persist the Argon2 salted hash in the DB to detect when encryption has been applied
- Implement a CipherSeed scheme based on aezeed that can be encoded using the Mnemonic seed words
- Integrate the new CipherSeed into the KeyManagers
- Update Wallet backend and Clients to use the new CipherSeeds

Motivation and Context
---
The CipherSeed scheme has three main benefits
- It contains the seed birthday which means we perform recoveries more efficiently and not scan the whole blockchain
- It contains a checksum to verify the seed phrase is correct
- It can be encrypted with a passphrase and decrypted and authenticated. We don’t current’y use a passphrase on the seeds. That will be future work.

How Has This Been Tested?
---
Test have been updated


base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -0,0 +1,377 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct CipherSeed {
+    version: u8,
+    birthday: u16,
+    pub entropy: [u8; CIPHER_SEED_ENTROPY_BYTES],
+    salt: [u8; CIPHER_SEED_SALT_BYTES],
+}
+
+impl CipherSeed {
+    pub fn new() -> Result<Self, KeyManagerError> {
```
SWvheerden: Is there a reason we return a result here?
As currently, it will always return an OK(CipherSeed)

SWvheerden (dismissed):
LGTM

base_layer/key_manager/src/cipher_seed.rs
philipr-za: Ah yes that is right, this is a remnant of a earlier version that did something that could break in here but this can be removed.

philipr-za:
Just a note, I manually tested this by creating a wallet, transacting and recovering using seed words with the new CipherSeed implementation and all good.

stringhandler (approved):

PR #3506, opened 2021-10-28
feat!: add deleted height to unique_id index

sdbondi:
Description
---
- adds the spend/deleted height to the asset id index when a unique id
  UTXO is spent
- head unique_id (ie. <asset_pk, unique_id> in current UTXO set) has
  "special" key in unique_id_index
- handle reorg code to update unique_id_index
- replace hex string keys with byte keys in kernel and input dbs
- add test for fetch_utxo_by_unique_id with deleted height param

Motivation and Context
---
Allow fetch_utxo_by_unique_id to query the "head" UTXO that existed at any height

How Has This Been Tested?
---
Unit tests 
Some tests are failing but I suspect this was already the case before this PR

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -414,7 +414,7 @@ where T: BlockchainBackend + 'static
                     for id in unique_ids {
                         let output = self
                             .blockchain_db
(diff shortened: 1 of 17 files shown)


delta1:
Looks great, yes some tests were failing before this. Just a lil clippy

sdbondi:
👌  I'll give some 📎 ❤️

PR #3509, opened 2021-10-28
feat: tracing for console wallet

StriderDM:
Description
---
This PR adds instrumentation to methods within the tari_console_wallet for tracing output in Jaeger. 

Also includes some clippy fixes with the latest update to the Rust toolchain

Motivation and Context
---
This will give an indication of how often which method is called and from which component.

How Has This Been Tested?
---
Manually

### DIFF
--- a/base_layer/wallet/src/base_node_service/monitor.rs
+++ b/base_layer/wallet/src/base_node_service/monitor.rs
@@ -41,6 +41,7 @@ use std::{
 use tari_common_types::chain_metadata::ChainMetadata;
 use tari_comms::protocol::rpc::RpcError;
 use tokio::{sync::RwLock, time};
+use tracing::instrument;
 
 const LOG_TARGET: &str = "wallet::base_node_service::chain_metadata_monitor";
 
@@ -110,6 +111,7 @@ where
         );
     }
 
+    #[instrument(name = "base_node_monitor::monitor_node", skip(self))]
     async fn monitor_node(&mut self) -> Result<(), BaseNodeMonitorError> {
         let mut base_node_watch = self.wallet_connectivity.get_current_base_node_watcher();
         loop {
--- a/base_layer/wallet/src/output_manager_service/master_key_manager.rs
+++ b/base_layer/wallet/src/output_manager_service/master_key_manager.rs
@@ -37,6 +37,7 @@ use tari_key_manager::{
     key_manager::KeyManager,
     mnemonic::{from_secret_key, MnemonicLanguage},
 };
+use tracing::instrument;
 
 const LOG_TARGET: &str = "wallet::output_manager_service::master_key_manager";
 
@@ -159,6 +160,10 @@ where TBackend: OutputManagerBackend + 'static
         Ok(script_key.k)
     }
 
+    #[instrument(
+        name = "key_manager::get_coinbase_spend_and_script_key_for_height",
+        skip(self, height)
+    )]
     pub async fn get_coinbase_spend_and_script_key_for_height(
         &self,
         height: u64,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4930,6 +4930,9 @@ dependencies = [
  "time",
  "tokio 1.13.0",
  "tower 0.3.1",
+ "tracing",
+ "tracing-futures",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/applications/tari_base_node/src/main.rs
+++ b/applications/tari_base_node/src/main.rs
@@ -388,7 +388,7 @@ async fn cli_loop(parser: Parser, mut shutdown: Shutdown) {
             res = &mut read_command_fut => {
                 match res {
                     Ok((line, mut rustyline)) => {
-                        if let Some(p) = rustyline.helper_mut().as_deref_mut() {
+                        if let Some(p) = rustyline.helper_mut() {
                             p.handle_command(line.as_str(), &mut shutdown);
                         }
                         if !shutdown.is_triggered() {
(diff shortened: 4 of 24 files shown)


base_layer/wallet/src/connectivity_service/handle.rs
```diff
@@ -58,6 +63,7 @@ impl WalletConnectivityHandle {
 
 #[async_trait::async_trait]
 impl WalletConnectivityInterface for WalletConnectivityHandle {
+    #[instrument(name = "wallet_connectivity_interface::set_base_node", skip(self, base_node_peer))]
     fn set_base_node(&mut self, base_node_peer: Peer) {
```
stringhandler: This method will clearly return instantly. There's no need to instrument it.

stringhandler (commented):
Feels like you have just applied the #[instrument] tag to every method without thinking about whether the method returns instantly or not. Have you had a look at the output? How many of these methods are on a critical path. I would suggest looking at the jaeger output and seeing which methods are useful.

I am worried that if we merge this PR we'll be removing these attributes forever.

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -86,11 +87,13 @@ where T: WalletBackend + 'static
     }
 
     /// Returns the last known state of the connected base node.
+    #[instrument(name = "base_node_service::get_state", skip(self))]
     pub async fn get_state(&self) -> BaseNodeState {
         self.state.read().await.clone()
     }
 
     /// Starts the service.
+    #[instrument(name = "base_node_service::start", skip(self))]
```
philipr-za: Cut this one

base_layer/wallet/src/base_node_service/handle.rs
```diff
@@ -65,17 +66,20 @@ impl BaseNodeServiceHandle {
         }
     }
 
+    #[instrument(name = "base_node_service_handle::get_event_stream", skip(self))]
```
philipr-za: cut

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -194,6 +196,7 @@ where
         Ok(())
     }
 
+    #[instrument(name = "base_node_monitor::map_state", skip(self, transform))]
```
philipr-za: cut, an in-memory operation

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -205,6 +208,7 @@ where
         self.publish_event(BaseNodeEvent::BaseNodeStateChanged(new_state));
     }
 
+    #[instrument(name = "base_node_monitor::publish_event", skip(self, event))]
```
philipr-za: cut

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -86,11 +87,13 @@ where T: WalletBackend + 'static
     }
 
     /// Returns the last known state of the connected base node.
+    #[instrument(name = "base_node_service::get_state", skip(self))]
```
philipr-za: Cut, in memory operation

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -121,6 +124,7 @@ where T: WalletBackend + 'static
         Ok(())
     }
 
+    #[instrument(name = "base_node_service::spawn_monitor", skip(self))]
```
philipr-za: cut, this just spawn. No major work done.

base_layer/wallet/src/connectivity_service/handle.rs
```diff
@@ -67,6 +69,7 @@ impl WalletConnectivityInterface for WalletConnectivityHandle {
         self.base_node_watch.send(Some(base_node_peer));
     }
 
+    #[instrument(name = "wallet_connectivity_interface::get_current_base_node_watcher", skip(self))]
```
philipr-za: cut

base_layer/wallet/src/connectivity_service/handle.rs
```diff
@@ -110,26 +118,35 @@ impl WalletConnectivityInterface for WalletConnectivityHandle {
         reply_rx.await.ok()
     }
 
+    #[instrument(name = "wallet_connectivity_interface::get_connectivity_status", skip(self))]
```
philipr-za: cut

base_layer/wallet/src/connectivity_service/handle.rs
```diff
@@ -110,26 +118,35 @@ impl WalletConnectivityInterface for WalletConnectivityHandle {
...
     fn get_connectivity_status(&mut self) -> OnlineStatus {
         *self.online_status_rx.borrow()
     }
 
+    #[instrument(name = "wallet_connectivity_interface::get_connectivity_watch", skip(self))]
```
philipr-za: cut

philipr-za (commented):
Ok, I still think there are many methods instrumented here that do not need to be. I started marking a few and mentioning why they shouldn't be instrumented but I am not going to do them all. Please apply those criterion to the rest.

Any method that is just:
- spawning a task (returns immediately),
- returning a channel,
- returning a piece of data from memory,
should not be instrumented.

What should be instrumented:
- Any method that does any appreciable work
- Any method that makes use of a scarce resource like the database or network

base_layer/wallet/src/output_manager_service/master_key_manager.rs
```diff
@@ -142,6 +143,7 @@ where TBackend: OutputManagerBackend + 'static
 
     /// Return the next pair of (spending_key, script_private_key) from the key managers. These will always be generated
     /// in tandem and at corresponding increments
+    #[instrument(name = "key_manager::get_next_spend_and_script_key", skip(self))]
```
stringhandler: I would not add any tracing to this file at the risk of leaking information

StriderDM: Have removed them, thanks.

philipr-za (dismissed):
LGTM

philipr-za (approved):

PR #3510, opened 2021-10-28
test: add unique_id block validator tests

sdbondi:
Description
---
- Add some tests for unique_id validation
- breaking change: Use version 1 blocks for localnet, igor and dibbler (kernel ordering, mmr without empty bitmap)
- Add mmr to test blocks in blockchain database so that concrete validation impls can be tested  

Motivation and Context
---
These are not complete as more work needs to happen to complete the validations

How Has This Been Tested?
---
Tests pass


stringhandler (approved):
Nice

PR #3512, opened 2021-10-29
feat: add a Rejected status to TransactionStatus

philipr-za:
Description
---
This status will indicate the transaction was rejected by the mempool for being invalid. In the past it was just cancelled which is also the case when a Coinbase Transaction is manually cancelled during mining because the blockchain has moved on.

This PR adds the new status and applies is when a transaction is rejected during the `transaction_broadcast_protocol`. The Console Wallet is updated to display the rejected status.

How Has This Been Tested?
---
cargo test and manual testing


sdbondi (approved):
Yup, this makes sense

stringhandler (approved):

PR #3514, opened 2021-10-30
chore: update toolchain to nightly-2021-09-18

CjS77:
This seemingly innoucous update required several changes due to changes
in how the compiler interprets dead code.

* A few variables were renamed _x where it looked like x might be used
  in upcoming PRs. YAGNI maybe, but I gave the benefit of the doubt in
these cases.
* Some crates that had LOADS of dead code (wallet ffi crate I'm looking at
  you), I just changed the `deny(dead_code)` to `warn(dead_code)`
* And in some places, I just nuked the dead code, which tidied up some
  heavy code in the wallet in particular. Since the tests pass, I
presume this is fine and there isn't a weird new compiler bug that is
missing where the code is needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 defaults:
-  rust_image: &rust_image quay.io/tarilabs/rust_tari-build-with-deps:nightly-2021-08-18
+  rust_image: &rust_image quay.io/tarilabs/rust_tari-build-with-deps:nightly-2021-09-18
 
 commands:
   test:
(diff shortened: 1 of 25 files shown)


delta1:
Looks like the GitHub actions will have to be updated. Maybe the version can be removed there and cargo will pick it up from the toolchain file? 

The docker images will also need updating, but that means it’s time for me to sort out tari-labs/docker#19

CjS77:
> Looks like the GitHub actions will have to be updated. Maybe the version can be removed there and cargo will pick it up from the toolchain file?

Ah, that makes sense! I was wondering why it wasn't grabbing the correct version.

> 
> The docker images will also need updating, but that means it’s time for me to sort out [tari-labs/docker#19](https://github.com/tari-labs/docker/issues/19)

The new docker image is up

CjS77:
I want to wait for #3518, which will sort out the clippy errors, and then will rebase and re-push this

CjS77:
Also needs #3515 to be merged first

stringhandler (approved):

PR #3519, opened 2021-10-31
test: various cucumber fixes

stringhandler:
Description
---

* improved reorg test to ensure max difficulty
* reduced various warn logs to info or debug
* allowed dns_seeds to be passed as a comma separated string when passed as an ENV var

Motivation and Context
---
The Mass Multiple Reorg cucumber test was flaky because it did not use explicit difficulties in the chains, so it had vague statements "All chains are on the same tip"

How Has This Been Tested?
---
Tested locally


sdbondi (dismissed):
Nice

integration_tests/features/support/steps.js
```diff
@@ -250,12 +250,18 @@ Given(
...
       async () => {
-        let node_a_result = (await nodeA.get_node_state()) === "LISTENING";
-        let node_b_result = (await nodeB.get_node_state()) === "LISTENING";
+        let nodeAState = await nodeA.get_node_state();
+        console.log(`${nodeNameA} state: ${nodeAState}`);
+        let nodeBState = await nodeB.get_node_state();
+        console.log(`${nodeNameB} state: ${nodeBState}`);
+        let node_a_result = nodeAState === "LISTENING";
+        let node_b_result = nodeAState === "LISTENING";
         return node_a_result && node_b_result;
```
hansieodendaal: I suggest not having these checks in here. It is already tested for in `baseNodeProcess.js/start -> baseNodeProcess.js/run` and includes other variants of possible console output as well.

I found that these additional checks let this step time out, possibly due to `await this.client.GetTipInfo().sendMessage({});` not returning directly after base node startup.

integration_tests/features/Reorgs.feature
```diff
@@ -140,23 +140,23 @@ Feature: Reorgs
   Scenario Outline: Massive multiple reorg
```
hansieodendaal, line 140: Nicely predictable

integration_tests/features/support/steps.js
```diff
@@ -1316,13 +1312,31 @@ When(
   }
 );
 
+When(
+  /I mine (.*) blocks on (.*) with difficulty (.*)/,
+  { timeout: 20 * 1000 },
```
hansieodendaal, line 1335: This time out may be a little low if we mine many blocks. Suggest 80s.

hansieodendaal (commented):
Looks good, just two suggestions to consider.

integration_tests/features/support/steps.js
stringhandler: Thanks I see that now

philipr-za (approved):
LGTM

integration_tests/features/support/steps.js
```diff
@@ -1364,11 +1351,14 @@ When(
   /mining node (.*) mines (\d+) blocks$/,
   { timeout: 1200 * 1000 }, // Must allow many blocks to be mined; dynamic time out below limits actual time
   async function (miner, numBlocks) {
-    const miningNode = this.getMiningNode(miner);
+    const miningNode = this.getMiningNode(miner);d
```
delta1, line 1354: this won't parse

stringhandler: lol, thought I fixed that

PR #3523, opened 2021-11-02
chore: upgrade cucumber

SWvheerden:
This upgrades cucumber to the latest up-to-date version 8.
This has the effect that IntelliJ's cucumber.js plugin works correctly.


stringhandler (approved):
Nice

PR #3527, opened 2021-11-03
fix: use time crate instead of chrono

therustmonk:
Description
---
Replaces `chrono` crate to `time` where possible.

Motivation and Context
---
`chrono` crate depends on `time`, but is not actively maintained.
There is a known security issue:
https://github.com/chronotope/chrono/pull/578

How Has This Been Tested?
---
In progress...


stringhandler (approved):
Nice

PR #3528, opened 2021-11-03
refactor!: remove outdated wallet_ffi balance methods

StriderDM:
Description
Removed outdated balance methods in wallet_ffi
Updated wallet.h
Updated cucumber tests

Motivation and Context
Removal of outdated methods

Merge https://github.com/tari-project/tari/pull/3547 first.

How Has This Been Tested?
nvm use 12.22.6 &&  node_modules/.bin/cucumber-js --profile "ci"  --tags "not @long-running and not @broken and @wallet-ffi"

### DIFF
--- a/integration_tests/helpers/ffi/wallet.js
+++ b/integration_tests/helpers/ffi/wallet.js
@@ -10,8 +10,17 @@ const Contacts = require("./contacts");
 const Balance = require("./balance");
 
 const utf8 = require("utf8");
+
+class WalletBalance {
+  available = 0;
+  timeLocked = 0;
+  pendingIn = 0;
+  pendingOut = 0;
+}
+
 class Wallet {
   ptr;
+  balance = new WalletBalance();
   log_path = "";
   receivedTransaction = 0;
   receivedTransactionReply = 0;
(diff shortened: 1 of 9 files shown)


base_layer/wallet_ffi/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "tari_wallet_ffi"
 authors = ["The Tari Development Community"]
 description = "Tari cryptocurrency wallet C FFI bindings"
 license = "BSD-3-Clause"
-version = "0.20.0"
+version = "0.21.0"
```
philipr-za: Please leave incrementing the version to Mike. His process for producing a release will deal with this and it gets incremented based on the Conventional Commit tag. In this case if this PR is breaking then there should be a ! in the title.

philipr-za (commented):
So I see that the various Balance helper's exist but that the only way to get a balance after these methods are removed is via the callbacks. I think there still needs to be an explicit method to request the balance. So please include a `wallet_get_balance(...)` method in this PR that returns the current TariBalance.

The title of this breaking PR should have a ! in it to indicate it is a breaking change i.e. `refactor!: remove outdated wallet_ffi balance methods`

base_layer/wallet_ffi/Cargo.toml
StriderDM: I sometimes overlook this accidentally so have made a point to remember it. Will revert it in light of the above though, thanks.

StriderDM:
@philipr-za Your last comment with regards to `wallet_get_balance(...)` is a change of scope for this item. I am inclined to agree and will make the necessary changes to this PR.

philipr-za (dismissed):

integration_tests/helpers/ffi/wallet.js
```diff
@@ -321,6 +337,22 @@ class Wallet {
     return result;
   }
 
+  getBalance() {
```
hansieodendaal, line 340: I think this method would be more aptly named `getBalanceFromCallback()` to minimize confusion

integration_tests/helpers/ffi/wallet.js
```diff
@@ -321,6 +337,22 @@ class Wallet {
...
+    return this.balance;
+  }
+
+  pollBalance() {
```
hansieodendaal, line 344: I think this method would be more aptly named `getBalanceFromWallet()` to minimize confusion (see related comment)

integration_tests/helpers/walletFFIClient.js
```diff
@@ -74,6 +74,10 @@ class WalletFFIClient {
     return this.wallet.getBalance();
   }
 
+  pollBalance() {
+    return this.wallet.pollBalance();
```
hansieodendaal, line 78: I think these methods would be more aptly named `getBalanceFromWallet()` to minimize confusion (see related comment)

hansieodendaal (dismissed):
LGTM

**Note:** See suggested name changes for future consideration.

philipr-za (approved):

PR #3536, opened 2021-11-04
fix: prevent race condition between block propagation and sync

sdbondi:
Description
---

- prevents base node from switching to sync mode before the block has a chance to arrive via propagation

Motivation and Context
---
Base node may switch to sync mode many times as the block arrives at peers, syncing a single block and returning to
listening state. It is more efficient to stay in listening state for a while to give the block chance to arrive via propagation.

How Has This Been Tested?
---
Manually tested:
- more than 1 behind tip - sync as normal
- startup 1 behind missed propagation, enters sync mode
- 1 behind, prevented from entering sync mode, block arrives shortly after 

image
Prevented 3 times in a few minutes on my base node.


sdbondi:
I've just noticed the `blocks_behind_before_considered_lagging` setting - think we should default this to 1 because this causes unnecessary network overhead

stringhandler:
> I've just noticed the `blocks_behind_before_considered_lagging` setting - think we should default this to 1 because this causes unnecessary network overhead

This setting is only intended to be used in debugging and tests. It should be set to 0 by default

sdbondi:
There is a race condition as described in the PR which puts the base node in header sync before the block arrives via propagation.

stringhandler (approved):
cool

sdbondi:
Yeah so maybe that setting has a slightly different use case. than this which would never need to wait > 1 - ok cool makes sense

PR #3538, opened 2021-11-05
feat: add ffi get mnemonic wordlist

hansieodendaal:
Description
---
Added an FFI method to get the entire mnemonic word list for a given language.

Motivation and Context
---
See above.

How Has This Been Tested?
---
- Added a unit test to test this functionality.
- Added a cucumber test to verify the mnemonic word list is retrieved.


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -860,6 +861,49 @@ pub unsafe extern "C" fn seed_words_create() -> *mut TariSeedWords {
...
+///
+/// ## Returns
+/// `TariSeedWords` - Returns the TariSeedWords instance containing the entire mnemonic wordlist for the requested
+/// language.
+///
+/// # Safety
+/// The ```string_destroy``` method must be called when finished with a string from rust to prevent a memory leak
+#[no_mangle]
+pub unsafe extern "C" fn get_mnemonic_wordlist_for_language(
+    language: *const TariMnemonicLanguage,
```
philipr-za: The problem with this is that how does the FFI client construct this enum? I think the best is to accept a c string here and parse it to the enum within the function. If the string is null or doesn't parse then you will return an error across the FFI boundary.

hansieodendaal: Agreed

base_layer/wallet_ffi/src/error.rs
```diff
@@ -90,6 +92,10 @@ impl From<InterfaceError> for LibWalletError {
                 code: 6,
                 message: format!("{:?}", v),
             },
+            InterfaceError::InvalidArgument(_) => Self {
+                code: 1,
```
delta1:
```suggestion
                code: 7,
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -860,6 +861,61 @@ pub unsafe extern "C" fn seed_words_create() -> *mut TariSeedWords {
...
+    error_out: *mut c_int,
+) -> *mut TariSeedWords {
+    use tari_key_manager::mnemonic_wordlists;
+
+    let mut error = 0;
+    ptr::swap(error_out, &mut error as *mut c_int);
+
+    let mut seed_words_vec = Vec::new();
+    if language.is_null() {
+        error = LibWalletError::from(InterfaceError::NullError("mnemonic wordlist".to_string())).code;
```
delta1, line 889:
```suggestion
        error = LibWalletError::from(InterfaceError::ArgumentError("Unknown mnemonic language".to_string())).code;
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -860,6 +861,61 @@ pub unsafe extern "C" fn seed_words_create() -> *mut TariSeedWords {
...
+        ptr::swap(error_out, &mut error as *mut c_int);
+    } else {
+        let language_string = CStr::from_ptr(language).to_str().unwrap();
```
delta1: is it safe to unwrap this?

philipr-za: Nothing is safe in an unsafe method! 

If this fails we need to panic because something is wrong with the underlying memory that is being passed to us so currently throughout the library we make that the clients problem.  We could make these safer though throughout the lib and return from the function gracefully. I will make a ticket to do that throughout though.

StriderDM: @philipr-za We can't panic through the FFI, it results in undefined behavior. We always have to gracefully handle an error.

base_layer/wallet_ffi/wallet.h
```diff
@@ -159,6 +159,13 @@ struct TariPrivateKey *private_key_from_hex(const char *hex, int *error_out);
 // Frees memory for a TariPrivateKey
 void private_key_destroy(struct TariPrivateKey *pk);
 
+/// -------------------------------- Mnemonic Word List  ------------------------------------------ ///
+// Create a TariMnemonicWordList instance containing the entire mnemonic wordlist for the requested language
+struct TariMnemonicWordList *mnemonic_word_list_for_language(const char *language, int *error_out);
+
```
StriderDM: You'll need a forward declaration at the top of this file:
```
struct TariMnemonicWordList;
```

You'll need two additional functions here with the implementation in rust:
```
unsigned int mnemonic_word_list_get_length(struct TariMnemonicWordList *mnemonic_word_list, int *error_out);

char * mnemonic_word_list_get_at(struct TariMnemonicWordList *mnemonic_word_list, int *error_out);
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -6217,6 +6306,38 @@ mod test {
...
+                let mnemonic_wordlist = match *(language) {
+                    TariMnemonicLanguage::ChineseSimplified => mnemonic_wordlists::MNEMONIC_CHINESE_SIMPLIFIED_WORDS,
+                    TariMnemonicLanguage::English => mnemonic_wordlists::MNEMONIC_ENGLISH_WORDS,
+                    TariMnemonicLanguage::French => mnemonic_wordlists::MNEMONIC_FRENCH_WORDS,
+                    TariMnemonicLanguage::Italian => mnemonic_wordlists::MNEMONIC_ITALIAN_WORDS,
+                    TariMnemonicLanguage::Japanese => mnemonic_wordlists::MNEMONIC_JAPANESE_WORDS,
+                    TariMnemonicLanguage::Korean => mnemonic_wordlists::MNEMONIC_KOREAN_WORDS,
+                    TariMnemonicLanguage::Spanish => mnemonic_wordlists::MNEMONIC_SPANISH_WORDS,
+                };
+                assert_eq!(
```
StriderDM, line 6337: This test needs to be written as if it were being called from C.

Would look similar to:
```
let size = word_list_get_length(mnemonic_wordlist_ffi)

for (let i = 0; i < size; i++)
{
    let word = word_list_get_at(mnemonic_wordlist_ffi);
    asset_eq!(word, sample_wordlist[i]);
}
```

StriderDM (dismissed):

base_layer/wallet_ffi/wallet.h
philipr-za: There is already a TariSeedWords object which is what we should be using for this.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -235,6 +236,9 @@ pub struct EmojiSet(Vec<ByteVector>);
 #[derive(Debug, PartialEq)]
 pub struct TariSeedWords(Vec<String>);
 
+#[derive(Debug, PartialEq)]
+pub struct TariMnemonicWordList(Vec<String>);
```
philipr-za: Please just reuse TariSeedWords

philipr-za: That way you don't need all the accessor and destroy functions to be repeated.

base_layer/wallet_ffi/wallet.h
delta1: I think the Word List (set of possible words) is different to TariSeedWords (recovery phrase words) here @philipr-za - but I might be wrong

delta1: @StriderDM is there an integration test we could add to ensure the header file has all the necessary declarations? checking it manually will eventually result in something getting missed

philipr-za: > I think the Word List (set of possible words) is different to TariSeedWords (recovery phrase words) here @philipr-za - but I might be wrong

I agree and if this wasn't an FFI interface I would prefer a distinctly named type but in FFI if you forward declare a new object now you also have to provide all the helper functions for a client to interact with it...when we could just reuse the TariSeedWords object which is the exact same thing with maybe a debatable not quite appropriate name.

base_layer/wallet_ffi/src/lib.rs
StriderDM: @philipr-za `TariSeedWords` is to get the words in and out of the wallet. `TariMnemonicWordList` looks like it refers to a dictionary all of words in the selected language. The former allows words to be added/removed, the latter does not. This may cause confusion if combined, so whilst the types are equivalent the usage is not.

hansieodendaal:
Although `TariSeedWords` and `TariMnemonicWordList` looks the same, they are different things. All methods that operate on `TariSeedWords` is not valid for `TariMnemonicWordList`, for example `seed_words_push_word`. One might retrieve the mnemonic word list and then continue to work with it as if it is the seed words.

base_layer/wallet_ffi/wallet.h
delta1: Oh I see your point, makes sense

philipr-za:
> Although `TariSeedWords` and `TariMnemonicWordList` looks the same, they are different things. All methods that operate on `TariSeedWords` is not valid for `TariMnemonicWordList`, for example `seed_words_push_word`. One might retrieve the mnemonic word list and then continue to work with it as if it is the seed words.

So you would rather repeat all the helper functions except that single one (which returns feedback that something is wrong)? We are dealing with a very clunky FFI interface here which is unsafe on many levels so needs careful handling by the client and this is a case where I believe that adding 3 completely repeated helpers functions for a SLIGHTLY more appropriate API interface is definitely the wrong move. 

Edit: To be clear the original PR already missed 2 of these functions which was get_length and get_at. I feel this already demonstrates the issue of adding more types that are not really necessary.

hansieodendaal:
Thank you @delta1, @philipr-za and @StriderDM for the review comments; applied all. Unit tests and cucumber test working properly.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -914,7 +990,7 @@ pub unsafe extern "C" fn seed_words_get_at(
         ptr::swap(error_out, &mut error as *mut c_int);
     } else {
         let len = (*seed_words).0.len();
-        if position > len as u32 {
+        if position >= len as u32 {
```
StriderDM, line 993: Why `>=`, i.e len of 32 has 0..31 positions.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -235,6 +236,9 @@ pub struct EmojiSet(Vec<ByteVector>);
 #[derive(Debug, PartialEq)]
 pub struct TariSeedWords(Vec<String>);
 
+#[derive(Debug, PartialEq)]
```
philipr-za: This should come out if its not being used.

base_layer/wallet_ffi/wallet.h
```diff
@@ -71,6 +71,8 @@ struct TariPendingInboundTransaction;
 
 struct TariTransportType;
 
+struct TariMnemonicWordList;
```
philipr-za: Not being used

hansieodendaal:
@StriderDM the code panics if `position == len`, thus the change from`>` to `>=`. I tested this by accident.

``` rust
        if position >= len as u32 {
            error = LibWalletError::from(InterfaceError::PositionInvalidError).code;
        } else {
            word = CString::new((*seed_words).0[position as usize].clone()).unwrap()
        }
```

StriderDM:
@hansieodendaal Good find. Yeah I realized that right after posting my comment.

philipr-za (approved):

PR #3546, opened 2021-11-08
fix: add decision step between header sync and pruned/archival

sdbondi:
Description
---

- removes pruned metadata sync logic from listening state
- adds `DecideNextSync` step that is responsible for finding a suitable peer for block/pruned sync
- chain metadata service includes peer latency
- DRY'd up chain metadata service ping/pong event handling
- take latency into account when selecting a sync peer   
- header sync orders the peers in ascending latency order

Motivation and Context
---
When transitioning to header sync, the listening state would filter out peers that have a suitable horizon height.
However any peer regardless of pruned height can provide headers for header sync.

How Has This Been Tested?
---
Sync tests
Manually


base_layer/core/src/base_node/state_machine_service/states/sync_decide.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/// Find the peer with the best latency
+fn find_best_latency<'a, I: IntoIterator<Item = &'a SyncPeer>>(iter: I) -> Option<SyncPeer> {
+    iter.into_iter()
+        .fold(Option::<&'a SyncPeer>::None, |current, sync_peer| match current {
+            Some(p) => match (p.latency(), sync_peer.latency()) {
+                (Some(_), None) => Some(p),
+                (None, Some(_)) => Some(sync_peer),
+                (Some(current), Some(l)) if current > l => Some(sync_peer),
```
stringhandler: is this correct?

sdbondi: Maybe can name those better, "if current peer's latency is worse (more) than this peer's latency, use this peer"

stringhandler: was just wondering about the `1` value, I expected `current > sync_peer` or is that a lower case L ?

sdbondi: lol, yeah L for latency, I'll change it - also see there is a test failing - will fix

stringhandler (approved):

PR #3552, opened 2021-11-09
feat: add atomic swap htlc sending and claiming

SWvheerden:
Description
---
This Pr adds in the ability to send and claim HTLC transactions. 
This is added to the console wallet UI and GRPC commands. 

UI:
`tari_console_wallet --command "init-sha-atomic-swap <Amount> <Bob pubkey> <Message>"`
This is used to initialize an atomic swap for the XTR side
`tari_console_wallet --command "finalise-sha-atomic-swap <Output hash> <pre_image hex>"`
This is used to finalize and claim an HTLC transaction for the XTR side. 

GRPC:
`rpc SendShaAtomicSwapTransaction(SendShaAtomicSwapRequest) returns (SendShaAtomicSwapResponse);`
This is used to initialize an atomic swap for the XTR side
`rpc ClaimShaAtomicSwapTransaction(ClaimShaAtomicSwapRequest) returns (ClaimShaAtomicSwapResponse);`
This is used to finalize and claim an HTLC transaction for the XTR side. 

Motivation and Context
---
This is the first step in allowing Atomic swaps for Tari and coins that have SHA256 as a pre_image. 
Currently, the available commands in this PR and tools available we can do the happy path of an atomic swap for:
Bitcoin 
Bitcoin Cash (Bitcoin ABC, Bitcoin Unlimited, Bitcoin XT)
Decred 
Litecoin
Monacoin
Particl
Qtum
Vertcoin
Viacoin
Zcoin

How Has This Been Tested?
---
Added unit and cucumber tests for the Tari side. 
Did a manual XTR BTC atomic swap
XTR UTXO hash: 
`186a95239d16ebf800cb3b7991521585b569253255002b009b16f11faab8d115`
BTC tx_ids: 
`1f770ce82aeb1f6c0d0345a6a88f1b3eb2e226595063491c93ad442625f678ae`
`2b6e2e2566577e8e52da685e2abb4252b98fc4a60dc9a1b09a56e6b003900bc9`


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -331,6 +335,23 @@ where
...
+                    .fetch_outputs_from_node(vec![output_hash])
+                    .await?
+                    .pop()
+                    .ok_or_else(|| OutputManagerError::ServiceError("Output not found".to_string()))?;
+
+                self.create_claim_sha_atomic_swap_transaction(output, pre_image, fee_per_gram)
+                    .await
+                    .map(OutputManagerResponse::ClaimShaAtomicSwapTransaction)
+            },
+            OutputManagerRequest::CreateClaimHTLCSwapTransaction(boxed) => {
```
```suggestion
            OutputManagerRequest::CreateClaimHtlcSwapTransaction(boxed) => {
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1159,6 +1180,119 @@ where
         Ok((tx_id, tx, fee, utxos_total_value))
     }
 
+    async fn fetch_outputs_from_node(
```
stringhandler, line 1187: I'm sure this function already exists?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -780,6 +801,142 @@ where
         Ok(())
     }
 
+    /// broadcasts a SHA-XTR atomic swap transaction
+    /// # Arguments
+    /// 'dest_pubkey': The Comms pubkey of the recipient node
+    /// 'amount': The amount of Tari to send to the recipient
+    /// 'fee_per_gram': The amount of fee per transaction gram to be included in transaction
+    pub async fn broadcast_sha_atomic_swap_transaction(
```
stringhandler: Why does this need a specific broadcast?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1159,6 +1180,119 @@ where
...
+            .as_bytes(),
+        )?;
+        let rewind_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
+        let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&rewind_key))?;
+        let rewound =
+            output.full_rewind_range_proof(&self.resources.factories.range_proof, &rewind_key, &blinding_key)?;
+
+        let rewound_output = UnblindedOutput::new(
+            rewound.committed_value,
+            rewound.blinding_factor.clone(),
```
stringhandler, line 1233: Since you are rewinding anyway, why set the blinding factor to a DH key? Can't the sender use any blinding factor?

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: I can probably think of a better name here, but it does the creation of the HTLC and then the broadcast.

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: Well, technically we could. For now, I kept its workings the same as the current one-sided implementation.  But from a practical standpoint, it makes no difference if the blinding factor is a DH key or it is random and only the rewinding keys used is a DH key.

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: I think it was removed when refactored the wallet to use streaming and the new validation 
This is currently the only function that calls the base_node function: `fn fetch_matching_utxos` outside of the base_node. There are other functions that get utxo's, but none that only requests them on the hash,

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: renamed this to `send_sha_atomic_swap_transaction`

base_layer/wallet/src/output_manager_service/service.rs
mikethetike: if you don't, you risk exposing every swap into and out of tari. A person interested in a specific pub key could easily brute force all the htlc's

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -331,6 +335,23 @@ where
                 .set_coinbase_abandoned(tx_id, abandoned)
                 .await
                 .map(|_| OutputManagerResponse::CoinbaseAbandonedSet),
+            OutputManagerRequest::CreateClaimShaAtomicSwapTransaction(output_hash, pre_image, fee_per_gram) => {
+                let output = self
```
philipr-za: Please put this in a method, trying to keep this Match as clean as possible for a future refactor.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -780,6 +798,142 @@ where
...
+            .map_err(|e| TransactionServiceProtocolError::new(tx_id, e.into()))?;
+        let fee = stp
+            .get_fee_amount()
+            .map_err(|e| TransactionServiceProtocolError::new(tx_id, e.into()))?;
+
+        // We want to return the htlc commitment so we track it, but we dont want to return the commitment of the change
+        let output = if tx.body().outputs()[0].script == script {
+            tx.body().outputs()[0].clone()
+        } else {
+            tx.body().outputs()[1].clone()
```
philipr-za: This potential for panic. You rely on the logic of the transaction builders to either give you 1 output with the correct script or 2 outputs but if, somehow, you get 1 output that does not have the correct script but there is no change then panic.

I would do this with a `.find()` mapping and error if it produces a none, that sidesteps the explicit indexing.

base_layer/wallet/tests/transaction_service/service.rs
```diff
@@ -908,6 +910,146 @@ fn recover_one_sided_transaction() {
...
+                    break;
+                },
+            }
+        }
+        assert!(found, "'TransactionCompletedImmediately(_)' event not found");
+    });
+
+    runtime.block_on(async move {
+        let (tx_id_htlc, htlc_fee, htlc_amount, tx) = bob_oms
+            .create_claim_sha_atomic_swap_transaction_with_output(output, pre_image, 20.into())
```
philipr-za: So I see this is using a method that is used nowhere else? The `command.rs` file doesn't use this version. I am guessing its so that you don't have to handle the base node UTXO request? I would really prefer you actually test that part by using a base node RPC mock. You would use `setup_transaction_service_no_comms` which provides you with a RPC mock

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -425,6 +451,46 @@ impl OutputManagerHandle {
         }
     }
 
+    pub async fn create_claim_sha_atomic_swap_transaction_with_output(
```
philipr-za: What is the purpose of this version of the method? I get the feeling it exists purely for the test? I would prefer you rather perform the test using a base node mock.

docs/src/btc_atomic_swap.md
```diff
@@ -0,0 +1,184 @@
...
+
+The following dependencies are required:
+
+* Tari wallet
+
+* Bitcoind (only for BTC atomic swaps, replace with other wallets for other coins)
+
+* Golang
+
+* decred atomic swap ci program
```
philipr-za: ci?

philipr-za (commented):
Looking really good, a couple minor comments.

hansieodendaal (dismissed):
LGTM from a code review and logic perspective - really nice.

philipr-za:
Changes look good, just needs a rebase. Quite a few conflicts.

philipr-za (approved):

PR #3534, opened 2021-11-03
feat: one-click installer - cli edition

CjS77:
# Tari Launchpad - CLI edition

# Code changes

- Updated configuration for gRPC endpoints to accept MultAddr format in addition to SocketAddr. Changes should be 100% backwards compatible.

a.k.a. _Tari one-click miner_.

Currently this only works on MacOS and Linux.
Contributions for  `quick_start.bat` are welcomed.

## Prerequisites

1. [Docker](https://docs.docker.com/get-docker/)
2. Set up some environment variables:
  - `DATA_FOLDER` (required). The path to store the Tari configuration files and logs. You can create multiple
     network setups, each in their own sandbox.
  - `TARI_NETWORK` (optional). Default=weatherwax. Specify the Tari network to connect to.
  - `START_TOR` (optional). Default=1. Whether we should start a Tor instance
  - `START_BASE_NODE` (optional). Default=1. Whether we should start a base node instance
  - `START_WALLET` (optional). Default=1. Whether we should start a wallet instance (GRPC only)
  - `START_MINER` (optional). Default=1. Whether we should fire up a SHA_3 solo miner.
  - `USE_OWN_MODEROD` (optional). Default=0. Whether we should start and sync our own Monero node.
  - `START_MONERO_MM` (optional). Default=1. Whether we should start merge-mining on Monero.

  - `TARI_WALLET_PASSWORD` (optional). Highly recommended. Default: tari. The password for your Tari wallet.
  - `TARI_WALLET__WEATHERWAX__TOR_CONTROL_AUTH` (optional). Default: tari. The password for the Tor control auth.
  - `TARI_MONERO_WALLET_ADDRESS` (optional). Though you'll donate funds to some random address if you don't set it.

## Quick Start

`./quick-start.sh`

And you should be golden.

## Interacting with the wallet and base node.

Any Tari client that supports gRPC can be connected to the node setup. With the default configuration, the base node
will be at `http://127.0.0.1:18142` and the wallet will be at `http://127.0.0.1:18143`.

[Kreya](https://kreya.app/) is a fairly user-friendly gRPC client that can give you a rudimentary UI to the docker rig
out of the box:

#### Getting block information

node_blocks

#### Wallet - whoami

wallet_id

#### Wallet - send transactions

wallet_send

## Viewing logs and configuration files

Logs and configuration files are stored in `{DATA_FOLDER}/{app}/log/*.log`.

So assuming your data folder is `/tmp/tari1/, the generated file structure looks like:

```text
[...]

PR #3557, opened 2021-11-10
feat: add support for MultiAddr in RPC config

CjS77:
Motivation and Context
---

Robust networking support in docker configurations requires us to be able to specify the gRPC address of
things like the node and wallet in terms of domains instead of raw IP addresses.

This adds support for MultiAddr specification of these addresses while still being 100%
backwards compatible.

Also included in #3534 

How Has This Been Tested?
---
No new tests added.


stringhandler (approved):
Nice. 

I think in future we should move `multiaddr_to_socketaddr` out of `comms` and into `common` so that we can remove the dependency on `comms`.

PR #3558, opened 2021-11-10
fix: stop leak of value of recovered output

SWvheerden:
Description
---
Updated the recovery of an output to not reuse the blinding factor to use as thescript_key, but rather generate a new key. 

Motivation and Context
---
If the blinding factor is reused as the script_key, on spending you will reveal k.G, this makes guessing v.H trivial and this leaks the k.G value. 

How Has This Been Tested?
---

### DIFF
--- a/base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
+++ b/base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
@@ -23,12 +23,16 @@
 use std::sync::Arc;
 
 use log::*;
-use tari_crypto::{inputs, keys::PublicKey as PublicKeyTrait, tari_utilities::hex::Hex};
-
-use tari_common_types::types::PublicKey;
+use rand::rngs::OsRng;
+use tari_common_types::types::{PrivateKey, PublicKey};
 use tari_core::transactions::{
     transaction::{TransactionOutput, UnblindedOutput},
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -91,13 +95,17 @@ where TBackend: OutputManagerBackend + 'static
             })
             .map(
                 |(output, features, script, sender_offset_public_key, metadata_signature)| {
+                    // Todo we need to look here that we might want to fail a specific output and not recover it as this
+                    // will only work if the script is a Nop script. If this is not a Nop script the recovered input
+                    // will not be spendable.
+                    let script_key = PrivateKey::random(&mut OsRng);
```
philipr-za, line 101: Shouldn't we derive this key from the key manager?

SWvheerden: We can, but this is in effect a single-use key and can be anything. We can also change this on actual spending of the recovered utxo,  so makes near enough no difference I think.

philipr-za: Alright, makes sense

philipr-za (approved):

PR #3562, opened 2021-11-11
fix: avoid implicit using of the time crate

therustmonk:
Description
---
This PR removes implicit `time` crate dependency where possible.

Motivation and Context
---
`chrono` crate uses a deprecated version of the `time` crate that has security issues.
As a result of the investigation, it was found that `chrono` supports `time` for compatibility,
but it's not required for most features and could be easily deactivated (by deactivating
default features).
All `time` relations removed except `tari_app_grpc` that depends on `tari_utilities` that also
needed the similar fix.

How Has This Been Tested?
---
CI


stringhandler (dismissed):

philipr-za (approved):
LGTM

PR #3566, opened 2021-11-12
test: reduce cucumber ci to critical only

stringhandler:
Hopefully by reducing this CI step to critical only, we'll have more green ticks for the check and will have less to maintain regularly. 

Ideally we would want to run all of the tests, but because some of the non-critical ones fail in a flaky manner, devs have started ignoring this step and making it required takes too long.


delta1 (approved):
I can't wait to see if this run passes on circle 🤣

philipr-za (approved):
I agree with this. Though at some point we should go through and confirm that there are no tests that should be critical and are being skipped. But lets get it green first!

PR #3567, opened 2021-11-12
feat: trigger time lock balance update when block received

philipr-za:
Description
---
This PR update the console wallet event monitor to trigger a balance refresh when a new block is received IF there is a timelocked balanced that could potentially changed based on the block height.

How Has This Been Tested?
---
cargo test


stringhandler (approved):
good idea

PR #3568, opened 2021-11-14
feat: implement multiple read single write for sqlite

hansieodendaal:
Description
---
- Implemented a generic system-wide pool connection manager for diesel SQLite connections that uses the modern Write-ahead Log (WAL) SQLite3 database mode to speed up database operations  (_see https://www.sqlite.org/wal.html_) with a configurable pool size.
- Changed the Wallet's SQLite database connection to make use of the pool connection manager.
- Refined SQLite profiling trace logs to only log an entry if the total time is > 1ms - this cut down on noisy entries as most database calls are now < 1ms.
- Notable SQLite pragma settings:
  - [PRAGMA schema.journal_mode](https://www.sqlite.org/pragma.html#pragma_journal_mode) = `WAL`
  - [PRAGMA schema.locking_mode](https://www.sqlite.org/pragma.html#pragma_locking_mode) = `NORMAL `
  - [PRAGMA schema.synchronous](https://www.sqlite.org/pragma.html#pragma_synchronous) = `| 2 | FULL`
  - [PRAGMA wal_autocheckpoint](https://www.sqlite.org/pragma.html#pragma_wal_autocheckpoint) = `1000` (the default)

Motivation and Context
---
The wallet's database connection was not optimal.

How Has This Been Tested?
---
- Unit tests
- Cucumber tests (`npm test -- --tags "not @long-running and not @broken"`)
- System-level tests with positive results all around (see discussion):
  - Configuration: Base node + console wallet + merge mining proxy + XMRig -> Mining
  - Test 1: Moderate coin split (22 transactions producing 3124 outputs)
  - Test 2: Moderate stress test (2x simultaneous `make-it-rain` transactions from one sender to two receiving wallets of 250 transactions each)

The graphs below show the SQLite WAL mode profiling for the above two tests (sender wallet only) with a pool size of 16 for all read and write database operations where the total time (acquire lock time + db operation time) was larger than 1ms. The wallet in question had a communications breakdown to its connected base node for a while after all transactions were completed and when it eventually gained an RPC connection validation protocols for all 500 transactions queried the base node as fast as possible, which panned out in some of those protocols waiting for a pooled connection. The actual time db operations will wait for a write operation to conclude is managed by SQLite and bunched within the `db op time` measurement.

image

image

### DIFF
--- a/base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
+++ b/base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
@@ -20,31 +20,24 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use crate::{
-    schema::{completed_transactions, inbound_transactions, outbound_transactions},
-    storage::sqlite_utilities::WalletDbConnection,
-    transaction_service::{
-        error::{TransactionKeyError, TransactionStorageError},
-        storage::{
-            database::{DbKey, DbKeyValuePair, DbValue, TransactionBackend, WriteOperation},
-            models::{CompletedTransaction, InboundTransaction, OutboundTransaction, WalletTransaction},
-        },
-    },
-    util::{
-        diesel_ext::ExpectedRowsExtension,
-        encryption::{decrypt_bytes_integral_nonce, encrypt_bytes_integral_nonce, Encryptable},
-    },
-};
-use aes_gcm::{self, Aes256Gcm};
-use chrono::{NaiveDateTime, Utc};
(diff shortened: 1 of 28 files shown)


base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -2567,77 +2605,86 @@ mod test {
...
-            direct_send_success: false,
-            send_count: 0,
-            last_send_timestamp: None,
-        };
-        let inbound_tx_sql = InboundTransactionSql::try_from(inbound_tx).unwrap();
-        inbound_tx_sql.commit(&conn).unwrap();
+        let mut pool = SqliteConnectionPool::new(db_path.clone(), 1, true, true, Duration::from_secs(60));
+        pool.create_pool()
+            .unwrap_or_else(|_| panic!("Error connecting to {}", db_path));
+        // Note: For this test the connection pool is setup with a pool size of one; the pooled connection must go out
```
philipr-za, line 2611: Nit: this is fine, just a note that you could explicitly drop() the pool rather than using a {} scope which I think is a bit more explicit and easy to use.

philipr-za (approved):
Very nice, looks good to me.

PR #3571, opened 2021-11-15
feat: get fee for transactions for stratum transcoder

StriderDM:
Description
---
Adds the ability for a fee to be requested for a transaction from the console wallet via the tari stratum transcoder. It makes use of the existing grpc call get_transaction_info and essentially returns a (TxId, Fee) collection of json objects.

Motivation and Context
---
Primarily used in the Tari implementation for Miningcore.

How Has This Been Tested?
---
Manually.


applications/tari_stratum_transcoder/src/proxy.rs
```diff
@@ -480,6 +480,64 @@ impl InnerService {
...
+                        request["id"].as_i64(),
+                        1,
+                        "`transactions` field is empty or an invalid type for transfer request.",
+                        None,
+                    ),
+                )
+            },
+        };
+
+        let mut grpc_transaction_info = Vec::new();
```
philipr-za, line 503: nit: just a note that this is a situation where a `Vec::with_capacity()` is a slight optimization

philipr-za (approved):
lgtm

PR #3572, opened 2021-11-15
test: make monerod stagenet usage resilient

hansieodendaal:
Description
---
Made `monerod` stagenet usage resilient for use in cucumber by testing publically available URLs from a pre-defined list before assigning the first available and responsive one to use. Flaky URLs have been pushed to the back of the list.

Motivation and Context
---
Cucumber merge mining tests failed because the currently hard-coded `monerod` was not available.

How Has This Been Tested?
---
Tested merge mining in cucumber, All the responsive URLs in the list have been tested to work with the merge mining proxy. Sample console output below:
``` rust
  >> Warn: `monerod` at http://3.104.4.129:18081 is not available!
  >> Info: `monerod` at http://singapore.node.xmr.pm:38081 is responsive and available
```


integration_tests/helpers/mergeMiningProxyProcess.js
```diff
@@ -38,7 +40,7 @@ class MergeMiningProxyProcess {
     // console.log("MergeMiningProxyProcess init - assign server GRPC:", this.grpcPort);
   }
 
-  run(cmd, args) {
+  run(cmd, args, monerodUrl) {
```
delta1, line 43: Personally I’d rather this key/value is just added to the args object it doesn’t need to be a new function parameter

delta1:
Nice idea!

integration_tests/helpers/mergeMiningProxyProcess.js
delta1: Sorry, args is for the command line so this is fine

hansieodendaal: Thanks for the comment, but I would rather not so that `args` can remain dedicated for the command line arguments.

stringhandler (approved):

integration_tests/helpers/mergeMiningProxyProcess.js
delta1: Sure I agree with that, but next time additional parameters are required then this should be refactored into an `opts` object

PR #3573, opened 2021-11-16
feat: add atomic swap refund transaction handling

SWvheerden:
Description
---
This Pr adds the following things:
Atomic swap refund transactions
Proper handling of the script context
Transaction spend priority
SQL query for filtering locked outputs
Fix broken get_balance tests

Motivation and Context
---
One of the features that are required for atomic swaps is to be able to process refunds. This PR now allows atomic swap refunds to be processed after the time lock of the HTLC has passed. This can happen in a manual way where the user manually asks the console-wallet via the CLI, or if the time lock has passed the wallet will automatically spend this UTXO first. 

One of the missing features of TariScript is the `scriptContext`, which allows scripts to validate the blockchain state such as block height that is required in this case. This is now piped in anywhere where the script is validated if possible. If the state cannot be gathered, the state is filled in with default values which is how it operated up to now. 

This adds in a transaction spend priority that can allow UTXO to be given a priority to be spent. This is the case for high-risk UTXO such as HTLC which more than one party can claim. Manual claim for these transactions is still advised as they can be claimed as soon as possible, but this adds a second automated state where if you want to spend some UTXO at least pick the ones that you need to spend more urgently than the other ones. 

The current `select_utxos` function uses a rust function to filter the maturity of the UTXOs to only select UTXOs it can spend. This PR completely swaps out the selection to be based completely on SQL queries which should be much faster. 

All current get_balance tests are broken in where the available and time-locked balance is asserted to be the same. This is not due to the function being broken but rather the test and the testing code. What happened is that tests use a tip_height of `u64::MAX` but this is too large as the database uses `i64::MAX`. The problem comes when `u64::MAX` is cast as `i64` as this is equal to `-1` and when the query `FROM outputs WHERE status = ? AND maturity > ? AND script_lock_height > ?` is run, it select everything as the tip is now `-1`. This ensures that the testing tip is selected as `i64::MAX` to let the queries run correctly.

How Has This Been Tested?
---
Added new tests.

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -54,12 +54,14 @@ service Wallet {
     rpc ListConnectedPeers(Empty) returns (ListConnectedPeersResponse);
     // Cancel pending transaction
     rpc CancelTransaction (CancelTransactionRequest) returns (CancelTransactionResponse);
-    // Will triggger a complete revalidation of all wallet outputs.
+    // Will trigger a complete revalidation of all wallet outputs.
     rpc RevalidateAllTransactions (RevalidateRequest) returns (RevalidateResponse);
     // This will send a XTR SHA Atomic swap transaction
     rpc SendShaAtomicSwapTransaction(SendShaAtomicSwapRequest) returns (SendShaAtomicSwapResponse);
     // This will claim a XTR SHA Atomic swap transaction
     rpc ClaimShaAtomicSwapTransaction(ClaimShaAtomicSwapRequest) returns (ClaimShaAtomicSwapResponse);
+    // This will claim a HTLC refund transaction
+    rpc ClaimHtlcRefundTransaction(ClaimHtlcRefundRequest) returns (ClaimHtlcRefundResponse);
 }
 
 message GetVersionRequest { }
@@ -122,6 +124,15 @@ message ClaimShaAtomicSwapRequest{
 }
 
 message ClaimShaAtomicSwapResponse {
+    TransferResult results = 1;
+}
+
+message ClaimHtlcRefundRequest{
+    string output_hash = 1;
+    uint64 fee_per_gram = 2;
+}
+
+message ClaimHtlcRefundResponse {
     TransferResult results = 1;
 }
 
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -41,12 +41,14 @@ use log::*;
 use serde::{Deserialize, Serialize};
 use std::{
     cmp::max,
+    convert::TryInto,
     fmt::{Display, Error, Formatter},
 };
 use tari_common_types::types::{
     BlindingFactor,
     Commitment,
     CommitmentFactory,
+    HashOutput,
     PrivateKey,
     PublicKey,
     RangeProofService,
@@ -55,6 +57,7 @@ use tari_crypto::{
     commitment::HomomorphicCommitmentFactory,
     keys::PublicKey as PublicKeyTrait,
     ristretto::pedersen::PedersenCommitment,
+    script::ScriptContext,
     tari_utilities::hex::Hex,
 };
 
@@ -342,6 +345,7 @@ impl AggregateBody {
     /// This function does NOT check that inputs come from the UTXO set
     /// The reward is the total amount of Tari rewarded for this block (block reward + total fees), this should be 0
     /// for a transaction
+    #[allow(clippy::too_many_arguments)]
     pub fn validate_internal_consistency(
         &self,
         tx_offset: &BlindingFactor,
@@ -349,6 +353,8 @@ impl AggregateBody {
         bypass_range_proof_verification: bool,
         total_reward: MicroTari,
         factories: &CryptoFactories,
+        prev_header: Option<HashOutput>,
+        height: Option<u64>,
     ) -> Result<(), TransactionError> {
         self.verify_kernel_signatures()?;
 
@@ -361,7 +367,7 @@ impl AggregateBody {
         self.verify_metadata_signatures()?;
 
         let script_offset_g = PublicKey::from_secret_key(script_offset);
-        self.validate_script_offset(script_offset_g, &factories.commitment)
+        self.validate_script_offset(script_offset_g, &factories.commitment, prev_header, height)
     }
 
     pub fn dissolve(self) -> (Vec<TransactionInput>, Vec<TransactionOutput>, Vec<TransactionKernel>) {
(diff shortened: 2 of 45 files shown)


applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -125,6 +127,15 @@ message ClaimShaAtomicSwapResponse {
     TransferResult results = 1;
 }
 
+message ClaimHtlcRefundRequest{
+    string output = 1;
```
philipr-za: What does this string represent? The commitment, the hash? I would add that to the name for clarity sake.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -425,12 +431,17 @@ impl AggregateBody {
...
         script_offset: PublicKey,
         factory: &CommitmentFactory,
+        prev_header: Option<HashOutput>,
+        height: Option<u64>,
     ) -> Result<(), TransactionError> {
         trace!(target: LOG_TARGET, "Checking script offset");
         // lets count up the input script public keys
         let mut input_keys = PublicKey::default();
+        let prev_hash: [u8; 32] = prev_header.unwrap_or_default().as_slice().try_into().unwrap_or([0; 32]);
+        let height = height.unwrap_or_default();
```
philipr-za, line 441: The fact that you are passing in `Some(u64::MAX)` for the height every where kind of implies that that should be the default if no height is supplied?

base_layer/wallet/src/base_node_service/mock_base_node_service.rs
```diff
@@ -94,7 +94,7 @@ impl MockBaseNodeService {
     }
 
     pub fn set_default_base_node_state(&mut self) {
-        let metadata = ChainMetadata::new(u64::MAX, Vec::new(), 0, 0, 0);
+        let metadata = ChainMetadata::new(i64::MAX as u64, Vec::new(), 0, 0, 0);
```
philipr-za, line 97: So this bug needs a bit of attention I think. Does the use of the `UNSIGNED` keyword in the sql migrations not fix this? It does seem that Sqlite does support u64 (https://www.sqlite.org/c3ref/int64.html) and the reason we are using i64 was actually due to the way Diesel generated the schema file. If the `UNSIGNED` sql keyword fixes this issue then lets go that route rather.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -339,6 +339,11 @@ where
                 self.claim_sha_atomic_swap_with_hash(output_hash, pre_image, fee_per_gram)
                     .await
             },
+
+            OutputManagerRequest::CreateHtlcRefundTransaction(output, fee_per_gram) => self
+                .create_refund_sha_atomic_swap_transaction(output, fee_per_gram)
```
philipr-za: feels like this method name should be changed to the more generic htlc terminology as its updated elsewhere.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1275,25 +1282,101 @@ where
...
+
+        trace!(target: LOG_TARGET, "Claiming HTLC refund with transaction ({}).", tx_id);
+        self.confirm_encumberance(tx_id).await?;
+        let fee = stp.get_fee_amount()?;
+
+        stp.finalize(KernelFeatures::empty(), &factories, None, self.last_seen_tip_height)?;
+
         let tx = stp.take_transaction()?;
 
+        self.resources.db.encumber_outputs(tx_id, Vec::new(), outputs).await?;
```
philipr-za, line 1382: don't forget to `confirm_encumberance(tx_id)`

base_layer/wallet/src/output_manager_service/storage/models.rs
```diff
@@ -104,12 +109,39 @@ impl Ord for DbUnblindedOutput {
...
+    Normal,
+    HtlcSpendAsap,
+    Unknown,
+}
+
+impl From<u32> for SpendingPriority {
+    fn from(value: u32) -> Self {
+        match value {
+            100 => SpendingPriority::HtlcSpendAsap,
+            500 => SpendingPriority::Normal,
```
philipr-za, line 123: These numbers are a bit arb but I agree with giving some space on either side of them. Seeing are they are arb maybe make it a u8 just to take up less space. Still gives plenty resolution.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -102,6 +104,20 @@ impl OutputSql {
...
+    pub fn fetch_unspent_outputs_for_spending(
+        tip_height: i64,
+        conn: &SqliteConnection,
+    ) -> Result<Vec<OutputSql>, OutputManagerStorageError> {
+        Ok(outputs::table
+            .filter(outputs::status.eq(OutputStatus::Unspent as i32))
+            .filter(outputs::script_lock_height.le(tip_height))
+            .filter(outputs::maturity.le(tip_height))
+            .order(outputs::spending_priority.asc())
+            .order(outputs::value.asc())
```
philipr-za: This must depend on the UTXOSelectionStrategy. There is a strategy that chooses Largest first. So would take that param into this call and adjust this query based on the strategy selected.

philipr-za (commented):
Looks really good so far, just a few comments.

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: I was going with trying to keep existing behavior. Currently, all scripts run with height = 0. The `Some(u64::MAX)` was only added and used in tests, not actual production code. 
This has a security implication, if we default to `u64::MAX` outside of tests, we might have scripts that should fail and pass. 
For example, if a wallet can't get a height for any reason, it should not assume that the tip is suddenly `u64::MAX` as this means that it will accept for example an HTLC refund transaction that should not be valid yet. This will leave observers to observe that the wallet is behaving badly.

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: Its about 4 lines up? But I moved it down.

base_layer/core/src/transactions/aggregated_body.rs
philipr-za: Alright, makes sense

base_layer/wallet/src/output_manager_service/service.rs
philipr-za: Ah, it should come after this statement which is why I didn't look above it.

philipr-za (dismissed):
Just some conflicts to sort.

philipr-za (dismissed):

philipr-za (approved):

PR #3574, opened 2021-11-16
docs: rfc 0250_Covenants

sdbondi:
Description
---
Adds RFC for covenants on Tari

[Rendered](https://demo.hedgedoc.org/s/Vwn0ZtWsS)

Motivation and Context
---
Allows sidechain checkpointing transactions and many other use-cases.

### DIFF
--- a/RFC/src/RFC-0250_Covenants.md
+++ b/RFC/src/RFC-0250_Covenants.md
@@ -1,0 +1,468 @@
+# RFC-0250/Covenants
+
+## Covenants
+
+![status: draft](theme/images/status-draft.svg)
+
+**Maintainer(s)**: [Stanley Bondi](https://github.com/sdbondi)
+
+# Licence
+
+[The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright 2021 The Tari Development Community
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
+
+1. Redistributions of this document must retain the above copyright notice, this list of conditions and the following
+   disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+   derived from this software without specific prior written permission.
+
+THIS DOCUMENT IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS", AND ANY EXPRESS OR IMPLIED WARRANTIES,
+INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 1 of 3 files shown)


RFC/src/RFC-0250_Covenants.md
```diff
@@ -0,0 +1,468 @@
...
+The following assumptions are made:
+1. Duplicate commitments within a block are disallowed by consensus _prior_ to covenant execution,
+2. all outputs in the output set are valid, and
+3. all inputs are valid spends, save for covenent checks.
+
+## Protocol modifications
+
+Modifications to the existing protocol and consensus are as follows:
+
+- the covenant is recorded in the transaction [UTXO],
```
SWvheerden, line 109: There is not really a thing such as a transaction, there are inputs and outputs, and kernels that are grouped together + some header info. 
Is this recorder, in the input, output, kernel, or the header (such show the offset and script offset is stored)

RFC/src/RFC-0250_Covenants.md
```diff
@@ -0,0 +1,468 @@
...
+- the covenant is committed to in the output and input hashes to prevent malleability,
+- transactions with covenants entering the mempool MUST be validated, and
+- each covenant in a block must be validated before being included in the block chain.
+
+### Transaction input and output changes
+
+A `covenant` field would need to be added to the `TransactionOutput` and `TransactionInput` structs
+and committed to in their hashes.
```
SWvheerden, line 117: and signatures

RFC/src/RFC-0250_Covenants.md
```diff
@@ -0,0 +1,468 @@
...
+    // data size: 32 bytes
+    Hash([u8; 32]),
+    // byte code: 0x02 
+    // data size: 32 bytes
+    PublicKey(RistrettoPublicKey),
+    // byte code: 0x03
+    // data size: 32 bytes
+    Commitment(PedersonCommitment),
+    // byte code: 0x04
+    // data size: 64 bytes
```
SWvheerden, line 159: Cant we reduce this to 32 bytes by just looking at the hash of a commitment, or output for that matter?
This is stored on the block chain and sent over the wire, we want this to be as small as possible.

RFC/src/RFC-0250_Covenants.md
```diff
@@ -0,0 +1,468 @@
...
+```
+
+##### filter_fields_hashed_eq(fields, hash)
+
+```yaml
+op_byte: 0x33
+args: [Fields, VarInt]
+```
+
+##### filter_relative_height(height)
```
SWvheerden, line 299: This command can be done as is with TariScript? If we check if the input is valid to be spent, that is completely covered by TariScript. If we want to check if the input has a set height then its a completely different thing,

RFC/src/RFC-0250_Covenants.md
```diff
@@ -0,0 +1,468 @@
...
+
+Spend within 10 blocks or burn
+
+```
+not(filter_relative_height(10))
+```
+
+Note, this covenant may be valid when submitted to the mempool, but invalid by the time it is put in a block for
+the miner.
+
```
SWvheerden, line 396: This is a good example of something you might want to do with this, but this is something I would argue we should rather do with TariScript. 

The difference being:
TariScript: Controlling when you may spend your already mined UTXO
Convenants: Controlling to what you may spend your UTXO. 

And here they focus on the rules relating to the input and output respectively and I think we should stick to these as far as possible to simplify things.

SWvheerden (commented):
This looks good, like the direction this is going in. Made a few clarification comments

RFC/src/RFC-0250_Covenants.md
sdbondi: Yeah that's the general direction I think is good, I think I included this as an "incase we need it" but agree with what you've said

RFC/src/RFC-0250_Covenants.md
sdbondi: It's mostly to support running one covenant if spent within 100 blocks and another after

```
xor(
     and(filter_relative_height(100), A),
     and(not(filter_relative_height(100)), B)   
)
```

stringhandler (approved):
Nice work. The RFC is really well written.

I think there are probably better examples, but I think these can and will be updated as the implementation is done and more real-world scenarios are clear.

RFC/src/RFC-0250_Covenants.md
sdbondi: Yeah, I'm 50/50 on needing that covenant because it means we have to revalidate before assembling the block. This isn't the best example. The positive 50% is thinking that it seems useful to be able to either run covenant `A` or `B` depending on maturity. This allows spending to outputs a certain way before 100 blocks and another way after.

Maybe something like a vault which allows you access to funds while locked in the vault (say for 100 blocks) but ensures that the covenant protects the funds for those 100 blocks (by allowing you to burn the funds if an attacker pwned your private keys, discouraging the attacker because you can ensure that no one gets the funds). I didn't manage to come up with an example of that covenant but I think it's possible.

PR #3575, opened 2021-11-16
feat(wallet): import utxo’s as EncumberedToBeReceived rather than Unspent

philipr-za:
Description
---

When the Wallet imports a UTXO it would set its status to Unspent making it immediately spendable.  However, imported outputs are missing a number of pieces of data from the chain that a traditionally received output would have. Completing a validation cycle fills in this data.

However, this caused some issues in the mobile clients when importing a UTXO and quickly spending it before the first validation was complete. There were unintended edge cases when a validation completed during an active transaction negotiation.

This PR solves this problem by importing utxo’s into the wallet via the Wallet.import_utxo(…) method in the EncumberedToBeReceived status and triggering a new validation immediately. This stops the output from being selected for a transaction before the validation completes which should happen quickly if the wallet is connected to a base node.

How Has This Been Tested?
---
cargo test


stringhandler (approved):
Good idea

PR #3577, opened 2021-11-16
feat: improve wallet connectivity status for console wallet

hansieodendaal:
Description
---
- Improved base node connection status feedback to the console wallet UI.
- Added disconnect logic for the wallet connectivity service to enable closing bad base node connections if RPC connections cannot be obtained, which will force the connection cycle to restart from scratch, rather than just waiting for the connection to become alive again.
- ~~Disabled obtaining latency information after obtaining tip information in the wallet's base node monitor, as it sometimes introduces huge additional delays.~~
~~_**Note:** Fixing this behaviour is marked for another PR._~~ _**(Fixed in PR #3579)**_

Motivation and Context
---
- Connectivity status was not always current in the console wallet under certain edge cases, observed when switching base nodes.
- Retrying to obtain an RPC connection from a bad base node connection is not efficient; this sometimes results in the base node staying offline for extended (> 30 minutes) durations.

How Has This Been Tested?
---
- System-level testing:
  - Switching base nodes and observing the connection status while (a) mining and (b) monitoring transactions to be mined.
  - Stress testing `make-it-rain`.
  - Stress testing `coin-split`.
  - Stress testing simaltaneous `make-it-rain` and `coin-split`.

See sample profiling measurements below depicting the added delay sometimes introduced by obtaining latency information, _**without the fix mentioned above**_:
``` rust
2021-11-16 14:18:07.490078500 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain RPC client 160 ms
2021-11-16 14:18:08.062610200 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain tip info in 572 ms
2021-11-16 14:18:08.062646600 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain latency info in 0 ms
2021-11-16 14:18:08.062944600 [wallet::base_node_service::chain_metadata_monitor] DEBUG Base node ece4c616a156c1e4fe34d94d8c Tip: 52127 (Synced) Latency: 572 ms
2021-11-16 14:18:37.491546100 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain RPC client 0 ms
2021-11-16 14:20:28.971178900 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain tip info in 111479 ms
2021-11-16 14:25:07.794774200 [wallet::base_node_service::chain_metadata_monitor] TRACE Obtain latency info in 278823 ms
[...]

PR #3579, opened 2021-11-16
fix: remove delay from last request latency call

sdbondi:
Description
---

- Allows the last_request_latency to be instantly queried even when a client is busy with another request. 

Motivation and Context
---
Ref #3577 

A delay could be experienced when calling get_last_request_latency because a single client runs on a single task and so can only handle one service request at a time.

How Has This Been Tested?
---
Added unit test


hansieodendaal (dismissed):
Looks really promising, thanks @sdbondi, will revert if I still see some funny behaviour, which I doubt.

stringhandler (dismissed):

PR #3580, opened 2021-11-16
refactor: remove tari_common dependency from tari_comms

sdbondi:
Description
---
- Remove tari_common as a dependency on tari_comms.
- Remove unused dependencies from tari_common

Motivation and Context
---
Tari common was added as a release dependency to tari_comms for a single error conversion. 
Tari comms should only rely on infrastructure crates.

How Has This Been Tested?
---
Code compiles


philipr-za (approved):
LGTM

stringhandler (approved):

PR #3584, opened 2021-11-17
fix: update daily test start times and seed phrase

philipr-za:
Description
---
- This PR updates the daily test star times to be from 1am so that the tests are done before standup.
- Updated the recovery seed phrase to a wallet with 1234567890uT for each spotting when recovery misses some outputs
- Updates the pruning horizon to 20.

How Has This Been Tested?
---
YOLO


delta1 (dismissed):
YOLO? YOLO.

sdbondi (approved):

stringhandler (approved):

PR #3587, opened 2021-11-18
fix: allow bullet proof value only rewinding off one-sided transaction

SWvheerden:
Description
---
This PR allows us to do bulletproof value-only rewinding on one-sided transactions.

Motivation and Context
---
Currently, it is not possible to do bulletproof rewinding on the value only on a one-sided transaction utxo due to the way the commitment blinding factor and bulletproof rewinding keys are created. 
Currently, the two bulletproof rewinding keys are created as:
```
 let rewind_key = PrivateKey::from_bytes(&hash_secret_key(&commitment_blinding_factor))?;
 let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&rewind_key))?;
```
This means that if you share the rewind key, which is used to do value only rewinding, that a person can calculate the blinding key which is used to do full rewinding and expose the commitment blinding factor.  by changing the calculation order we prevent this and only allow full rewinding by something who needs to be able to do this. 

```
 let rewind_key = PrivateKey::from_bytes(&hash_secret_key(&blinding_key ))?;
 let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&commitment_blinding_factor))?;
```

How Has This Been Tested?
---

All current test pass

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -1422,8 +1422,8 @@ where
                     )
                     .as_bytes(),
                 )?;
-                let rewind_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
-                let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&rewind_key))?;
+                let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Nice, makes sense 

This isn't so much a consensus change as something that could make previously 'undiscovered' one-side funds not redeemable. Not a huge deal in testnet, so think this can go in

philipr-za (approved):
Ja this looks good and almost no one is using one-side transactions at the moment so the breaking impact is very small.

PR #3588, opened 2021-11-18
fix: be more permissive of responses for the incorrect request_id

sdbondi:
Description
---
- add more intrumentation to tari_comms
- improve behaviour and logs around stream interruptions
- be more permissive of responses for the incorrect request_id
- reduce transaction_resend_period to 10 minutes from an hour @philipr-za 

Motivation and Context
---
Track more metrics to diagnose issues.
Formally recognise a stream interruption and continue the session vs a protocol violation which terminates the session. 

How Has This Been Tested?
---
Existing test, stress test involving 2 wallets and a base node


delta1 (dismissed):
LGTM

stringhandler (approved):
Awesome

PR #3589, opened 2021-11-18
feat: add wallet stub to collectibles app

delta1:
Description
---

- Adds logic to create wallet with password when collectibles is started
- Shows seed words to recover cipher seed 
- If wallet exists then user is prompted for password 
- Protects routes if wallet is not unlocked 

todo: 
- lock screen after inactivity
- hide accounts when locked
- allow multiple wallets
- wallet screen to retrieve seed words (after creation) 
- key manager to derive keys
- wallet recovery


stringhandler (approved):
Nice, looking good

PR #3590, opened 2021-11-18
feat: language detection for mnemonic seed words

StriderDM:
Description
---
Implementation of a basic language detection algorithm for mnemonic words

Motivation and Context
---
Improve language detection for mnemonic words

How Has This Been Tested?
---
cargo test --all


base_layer/key_manager/src/mnemonic.rs
```diff
@@ -158,11 +158,50 @@ pub fn from_bytes(bytes: Vec<u8>, language: &MnemonicLanguage) -> Result<Vec<Str
     Ok(mnemonic_sequence)
 }
 
+/// Detects the language based on the mnemonic words supplied
+fn detect_language(mnemonic_seq: &[String]) -> Result<MnemonicLanguage, MnemonicError> {
+    let mut languages = Vec::new();
+    for x in 0..23 {
```
sdbondi: off by one
```suggestion
    for x in 0..24 {
```

sdbondi (commented):
Think something like this may be simpler 

```rust
fn detect_language(words: &[&str]) -> Result<MnemonicLanguage, MnemonicError> {
    let mut langs = MnemonicLanguage::iterator().collect::<Vec<_>>();
    for word in words {
        langs = langs
            .drain(..)
            .filter(|lang| find_mnemonic_index_from_word(word, lang).is_ok())
            .collect();
        if langs.is_empty() {
          return  Err(MnemonicError::UnknownLanguage);
        }
        if langs.len() == 1 {
            return Ok(langs.remove(0));
        }
    }
    Err(MnemonicError::UnknownLanguage)
}
```

(probably could be made to have zero allocations but that's a bonus)

This language detection also needs to be in `MneumonicLanguage::from`

philipr-za (commented):
A bit of a brute force approach in my opinion. In 99% of cases the first word would be good enough. This approach checks all 24 words against all languages. But it probably works. The issue is I can't tell if it works because it is not tested at all. Please provide tests that show it working and testing the edge cases it is trying to solve.

Also you have left the old detection method in MnemonicLanguage::from(...) which is also where I would actually put this implementation.

StriderDM:
Thanks, @sdbondi, and @philipr-za. I have switched it to a scanning-based approach.

stringhandler (approved):

PR #3592, opened 2021-11-19
feat: standardize output hash for unblinded output, transaction output and transaction input

hansieodendaal:
Description
---
- Standardized output hash calculation for unblinded output, transaction output and transaction input.
- Refactored `transaction.rs` code layout to be modular in the file system. 
  _**Note:** The only real change in the code is the standardized use of `transactions/transaction_entities/mod.rs/pub fn hash_output(...)` called from `transactions/transaction_entities/transaction_input.rs`, `transactions/transaction_entities/transaction_output.rs` and `transactions/transaction_entities/unblinded_output.rs`._

Motivation and Context
---
Different output hash calculation implementations existed for output hash calculation for unblinded output vs. transaction output and transaction input.

How Has This Been Tested?
---
Unit testing.


SWvheerden (dismissed):
LGTM so far
I like the splitting up of the transaction file, that file had way too many structs and way too many lines.

hansieodendaal:
As a result of a discussion with @SWvheerden, I think it will be better to let the application layer determine the `CryptoFactories` to use, as well as to use the `Hashable` trait. I will make a separate commit for that proposal.

philipr-za (approved):
Looks good

PR #3596, opened 2021-11-19
feat: implement dht pooled db connection

hansieodendaal:
Description
---
Implemented the pooled SQLite db connection for the DHT storage in favour of the Rust based async read-write lock, which makes use of the much faster SQLite3 Write-ahead Log (WAL) mode.

Motivation and Context
---
Recent changes to the common implementation for SQLite pooled connections managed in the background via Diesel and SQLite3 enabled this improvement.

How Has This Been Tested?
---
- Unit tests
- Cucumber tests
- System-level tests [_**Edit:** Running a twice more intensive stress test than before (2x sender wallets with 3x 1000 transactions on one computer to 3x receiver wallets on another computer vs. 1x sender wallet with 3x 1000 transactions to 3x receiver wallets) the disk resource utilization was 90+% down from previous levels - ~100% disk utilization on both systems during all phases of the transactions life cycle down to <~5% with occasional spikes up to ~80%. All transactions reached mined confirmed in all 5 participating wallets.)_


comms/dht/src/dedup/dedup_cache.rs
```diff
@@ -72,96 +70,91 @@ impl DedupCacheDatabase {
...
+            Ok(1) => Ok(1),
+            Ok(n) => Err(StorageError::UnexpectedResult(format!(
+                "Expected exactly one row to be inserted. Got {}",
+                n
+            ))),
+            Err(diesel::result::Error::DatabaseError(kind, e_info)) => match kind {
+                DatabaseErrorKind::UniqueViolation => {
+                    // Update hit stats for the message
+                    {
+                        let conn = self.connection.get_pooled_connection()?;
```
sdbondi: Why not get the pool connection once in this function?

comms/dht/src/storage/error.rs
```diff
@@ -42,4 +43,12 @@ pub enum StorageError {
     MessageFormatError(#[from] MessageFormatError),
     #[error("Unexpected result: {0}")]
     UnexpectedResult(String),
+    #[error("Diesel R2d2 error: `{0}`")]
+    DieselR2d2Error(#[from] SqliteStorageError),
+}
+
+impl PartialEq for StorageError {
```
sdbondi: I think you can use the `PartialEq` derive macro

comms/dht/src/store_forward/database/mod.rs
```diff
@@ -82,173 +76,157 @@ impl StoreAndForwardDatabase {
...
-                        .filter(stored_messages::id.eq_any(message_ids))
-                        .execute(conn)?;
-                }
-                Ok(num_removed)
-            })
-            .await
+    pub(crate) fn truncate_messages(&self, max_size: usize) -> Result<usize, StorageError> {
+        let mut num_removed = 0;
+        let msg_count = {
+            let conn = self.connection.get_pooled_connection()?;
```
sdbondi: Same here, get the connection once

sdbondi (commented):
Looks good, A little concerned about blocking tokio task scheduling with synchronous calls, but understand that may be tricky with pooled connections. Let's see if we end up with any perf issues - it's in the wallet, so will probably be fine.

comms/dht/src/dedup/dedup_cache.rs
hansieodendaal: Yip, logical error on my part, as these operations run in series. (_Got confused by the fact that the pooled connection is held until the code goes out of scope._)

comms/dht/src/storage/error.rs
hansieodendaal: The `PartialEq` derive macro cannot be applied to some of the members here, i.e. `tokio::task::JoinError` and `tari_utilities::message_format::MessageFormatError`.

sdbondi: I suspect this results in an infinite recursion - by doing `self == other` you can calling `PartialEq::eq`. 
Try adding this to the bottom and it should stack overflow
```rust

#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn will_break() {
        let _ = StorageError::InvalidUnicodePath == StorageError::InvalidUnicodePath;
    }
}
```

sdbondi: If you want to check equality of an error in tests without implementing `PartialEq` you can use the `matches!`, a `match` or `tari_test_utils::unpack_enum!`  or remove if unused.

hansieodendaal:
> Looks good, A little concerned about blocking tokio task scheduling with synchronous calls, but understand that may be tricky with pooled connections. Let's see if we end up with any perf issues - it's in the wallet, so will probably be fine.

@sdbondi see edit about stress test results monitoring disk utilization. The change actually affects both the base node's and the wallet's DHT dedup databases.

sdbondi:
@hansieodendaal I'm happy about using a WAL - was talking specifically about tokio scheduler blocking, but if the calls are quick enough it shouldn't have a negative impact. I expect it will be fine.

EDIT: oh just saw the updated PR description - LGTM

sdbondi:
@hansieodendaal Just need to address the infinite recursion in PartialEq

comms/dht/src/storage/error.rs
hansieodendaal: Was actually not needed here, but I had to change it for `impl PartialEq for WalletStorageError`, also added a comparison test for `WalletStorageError`.

base_layer/wallet/src/error.rs
```diff
@@ -173,6 +173,75 @@ impl From<WalletStorageError> for ExitCodes {
 
 impl PartialEq for WalletStorageError {
     fn eq(&self, other: &Self) -> bool {
-        self == other
+        match other {
```
philipr-za: I definitely don't think this is needed. Rather just don't implement PartialEq. The reason is that anytime anyone adds a new error to this that derives from an error in another crate that doesn't implement PartialEq it will break this. We do not want to maintain this. You will see many errors in the system do not have this because its such a nightmare.

Anywhere that wants to check if the WalletStorageError equals something then they can use matches!() themselves.

base_layer/wallet/src/error.rs
```diff
@@ -173,6 +173,75 @@ impl From<WalletStorageError> for ExitCodes {
...
+        assert!(!compare_false);
+
+        let compare_true =
+            WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error(String::from("test1"))) ==
+                WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error(String::from("test2")));
+        assert!(compare_true);
+
+        let compare_true =
+            WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error(String::from("test1"))) ==
+                WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error(String::from("test1")));
```
sdbondi: and
```rust 
let compare_true = 
 WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error("test1".into()))
==
WalletStorageError::DieselR2d2Error(SqliteStorageError::DieselR2d2Error("something different".into()));
```
?

also true

sdbondi: Is there a reason we need PartialEq implemented on the error?

base_layer/wallet/src/error.rs
hansieodendaal: Ok, removing `PartialEq` from dependent errors.

base_layer/wallet/src/error.rs
hansieodendaal: Removed all of these as per the previous comment.

sdbondi (dismissed):
Nice looks good

philipr-za (dismissed):

philipr-za (approved):

PR #3598, opened 2021-11-21
fix: minor improvements to available neighbouring peer search

sdbondi:
Description
---
- Removes peers that were previously offline; a connection was re-attempted and fails.
- Instead of attempting to replace a disconnected peer in the neighbouring and random peer pools, attempt to reconnect to the disconnected peer. If that fails, the peer is replaced.
- neighbouring peers ordered by last seen (based on @mikethetike 's PR #2817)
- Xor distance uses u128 (based on @mikethetike 's PR #2817)
- add cooldown if encountering many failed connection attempts from DHT pool peers
- add last_seen directly to peer (w/ migration)

Motivation and Context
---
By removing peers from the peer manager when offline, the number of peers stored on the network should gradually decrease.  This will only happen once all/most peers are upgraded.

How Has This Been Tested?
---
Memorynet
Manually, base node operates normally and peer list reduces (until peer sync pulls more in again from nodes without this change)


hansieodendaal (dismissed):
Logic LGTM - not tested

sdbondi:
Going to work on various improvements to the strategy here, setting as draft

delta1 (approved):
LGTM, tested by starting to sync a node and seems good

PR #3600, opened 2021-11-22
chore: public_address for TCP node has higher priority from env/config

Cifko:
Description
---
The tcp `public_address` read from env/config (if set) instead of `base_node_id.json`.
I've added a comment to the newly generated `base_node_id.json` that this value is overwritten.
I've chosen json5 so it's backwards compatible.

How Has This Been Tested?
---
Manually.


delta1 (changes requested):
It looks okay but all of the unwraps should rather propagate errors

applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -203,7 +224,11 @@ pub fn save_as_json<P: AsRef<Path>, T: MessageFormat>(path: P, object: &T) -> Re
             fs::create_dir_all(p).map_err(|e| format!("Could not save json to data folder. {}", e.to_string()))?;
         }
     }
-    fs::write(path.as_ref(), json.as_bytes()).map_err(|e| {
+    let json_with_comment = format!(
+        "// The public address is overwritten by the config/environment variable\n{}",
```
delta1: I think this comment should probably specify what the ENV var is?

Cifko: I've added TARI_BASE_NODE__<network>__PUBLIC_ADDRESS into the comment.

Cifko:
> It looks okay but all of the unwraps should rather propagate errors

There was only one, and it's done. If you also meant the `match` on `applications/tari_app_utilities/src/identity_management.rs:159` that needs to stay there. Because that's used when you are creating the TCP node.

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -299,7 +299,7 @@ pub async fn init_wallet(
     );
 
     let node_address = match wallet_db.get_node_address().await? {
-        None => config.public_address.clone(),
+        None => config.public_address.clone().unwrap(),
```
delta1: And this one?

applications/tari_base_node/src/bootstrap.rs
```diff
@@ -184,11 +185,19 @@ where B: BlockchainBackend + 'static
...
+        let comms = initialization::spawn_comms_using_transport(comms, transport_type.clone()).await?;
         // Save final node identity after comms has initialized. This is required because the public_address can be
         // changed by comms during initialization when using tor.
-        identity_management::save_as_json(&config.base_node_identity_file, &*comms.node_identity())
-            .map_err(|e| anyhow!("Failed to save node identity: {:?}", e))?;
+        match transport_type {
+            TransportType::Tcp {
+                listener_address: _,
+                tor_socks_config: _,
+            } => {}, // Do not overwrite TCP public_address in the base_node_id!
```
sdbondi:
```suggestion
            TransportType::Tcp { .. } => {}, // Do not overwrite TCP public_address in the base_node_id!
```

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -299,7 +299,10 @@ pub async fn init_wallet(
     );
 
     let node_address = match wallet_db.get_node_address().await? {
-        None => config.public_address.clone(),
+        None => config
+            .public_address
+            .clone()
+            .ok_or(ExitCodes::ConfigError("node public address error".to_string()))?,
```
sdbondi: 📎

sdbondi (commented):
LGTM - minor comments

stringhandler (approved):

sdbondi (approved):

PR #3607, opened 2021-11-23
fix: seed word parsing

StriderDM:
Description
---
Moved `detect_language` into `MnemonicLanguage` and made it public.
Prevented a `TariSeedWords` object from becoming invalid in the event an invalid or inconsistent word was attempted to be pushed to it in wallet_ffi.
Differentiated between an invalid word and an inconsistent word.
Added word to the string of WordNotFound error.

Motivation and Context
---
General fixes

How Has This Been Tested?
---
cargo test --all

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -1020,11 +1020,13 @@ pub unsafe extern "C" fn seed_words_get_at(
 ///
 /// ## Returns
 /// 'c_uchar' - Returns a u8 version of the `SeedWordPushResult` enum indicating whether the word was not a valid seed
-/// word, if the push was successful and whether the push was successful and completed the full Seed Phrase
+/// word, if the push was successful and whether the push was successful and completed the full Seed Phrase.
+///  `seed_words` is only modified in the event of a `SuccessfulPush`.
 ///     '0' -> InvalidSeedWord
 ///     '1' -> SuccessfulPush
 ///     '2' -> SeedPhraseComplete
 ///     '3' -> InvalidSeedPhrase
+///     '4' -> NoLanguageMatch,
 /// # Safety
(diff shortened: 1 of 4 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1082,28 +1084,47 @@ pub unsafe extern "C" fn seed_words_push_word(
...
+        if (*seed_words).0.len() < 24 {
+            SeedWordPushResult::SuccessfulPush as u8
         } else {
             SeedWordPushResult::SeedPhraseComplete as u8
-        };
+        }
+    } else {
+        log::error!(
+            target: LOG_TARGET,
+            "Word in seed phrase is inconsistent: {:?}",
```
philipr-za: Do you mean that there is no language matching the words so far? If so I think this error is not very clear. If you saw it in the logs in isolation would you be able to determine what the problem is? Maybe something like "Current words in seed phrase do not match a supported language"

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1082,28 +1084,47 @@ pub unsafe extern "C" fn seed_words_push_word(
...
+            word_string
+        );
+        SeedWordPushResult::InconsistentSeedWord as u8
```
philipr-za: Maybe something like "SeedWordsDoNotMatchSingleLanguage"

base_layer/wallet_ffi/src/lib.rs
StriderDM: What is meant here is that by adding the word, the language cannot be determined anymore since there are no words that are either in one language or words common to at least one of them. The error will be UnknownLanguage.

Perhaps a pre-detection of the language prior to this and adding "...previously words were detected to be in `language`" will help?

philipr-za (approved):
Good catch

PR #3610, opened 2021-11-23
chore: change status line

Cifko:
Description
---
Added NodeId to status line.
And status line is printed with more info (currently RandomX) every 2 minutes.

How Has This Been Tested?
---
Manually.


Cifko:
Is there a better way than adding Mutex? In any other language it would be a 4 line change. But this looks crazy to me to do all that for one mutable variable. Another solution that came to my mind is to just grab the time and print it on every % 5 minute (so every fifth minute, but that looks wrong on different level).

stringhandler (approved):

stringhandler:
Yeah, but it's not a bad time to add a mutex around that command handler. Technically only one command should run at a time anyway

PR #3615, opened 2021-11-24
feat!: add tcp bypass settings for tor in wallet_ffi

StriderDM:
Description
---
This PR allows the TCP bypass settings for Tor to be set via wallet_ffi instead of being hard-coded.

Motivation and Context
---
As above.

How Has This Been Tested?
---
cargo test --all
nvm use 12.22.6 && node_modules/.bin/cucumber-js --profile "ci" --tags "not @long-running and not @broken and @wallet-ffi"


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2654,6 +2660,38 @@ pub unsafe extern "C" fn transport_tor_create(
 
     let identity = None;
 
+    let mut bypass_address: Option<Multiaddr> = None;
+    if !tor_proxy_bypass_address.is_null() {
```
sdbondi: I don't think that any ffi client needs this setting. This allows the client to specify a list of TCP addresses that will be directly connected to. The main use case is to connect a wallet to a local base node instead of going through tor, which only makes sense at the moment for the console wallet running on the same network/machine as a base node. It's ok to have this setting accessible but may create future confusion about how to use it (it should always be null in Aurora)

sdbondi (dismissed):
LGTM - though I perhaps wouldn't have exposed tor_proxy_bypass_address without a use case for it.

base_layer/wallet_ffi/src/lib.rs
StriderDM: @sdbondi 100%. It was included in the event it may be useful to us in debugging traffic and data usage issues. I can remove it, thoughts?

philipr-za: I agree that if these addresses are not of use to a mobile client then rather remove them because it does feel like these two parameters are related. @sdbondi so just to confirm the `tor_proxy_bypass_for_outbound` is of use to the mobile client correct? My understanding was that if its enabled the comms layer will skip Tor when connecting out to clients using a TCP address and that is a separate feature to the Bypass Address list?

StriderDM: @philipr-za Will do.

philipr-za (approved):

base_layer/wallet_ffi/src/lib.rs
sdbondi: @philipr-za That is correct, the other is a list of TCP/IP addresses that bypass tor, typically used when your console wallet and base node are running on the same machine.

PR #3617, opened 2021-11-25
WIP RFC XTR XMR swap

SWvheerden:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/RFC/src/RFC-0241_AtomicSwapXMR.md
+++ b/RFC/src/RFC-0241_AtomicSwapXMR.md
@@ -1,0 +1,264 @@
+# RFC-0240/Atomic Swap
+
+## Time-related Transactions
+
+![status: draft](theme/images/status-draft.svg)
+
+**Maintainer(s)**: [S W van Heerden](https://github.com/SWvheerden)
+
+# Licence
+
+[The 3-Clause BSD Licence](https://opensource.org/licenses/BSD-3-Clause).
+
+Copyright 2021 The Tari Development Community
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
+
+1. Redistributions of this document must retain the above copyright notice, this list of conditions and the following
+   disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+   derived from this software without specific prior written permission.
+
+THIS DOCUMENT IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS", AND ANY EXPRESS OR IMPLIED WARRANTIES,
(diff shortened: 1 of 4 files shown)


RFC/src/RFC-0241_AtomicSwapXMR.md
```diff
@@ -0,0 +1,238 @@
...
+* The blinding factor \\(k_i\\) for the Tari UTXO
+
+
+### Commitment phase
+
+This phase allows Alice and Bob to commit to the use of their keys. 
+Alice needs to provide bob the following:
+
+* Adaptor signature \\(b'_{Sra}\\) for \\(b_{Sra}\\)
+* Signature \\(a_{Sra}\\)
```
sdbondi, line 141: This is a commsig, right? So shouldn't the signature scalars be `u` and `v` rather than `a` and `b`?
```suggestion
* Signature \\(u_{Sra}\\)
```

RFC/src/RFC-0241_AtomicSwapXMR.md
```diff
@@ -0,0 +1,238 @@
...
+
+### XTR payment
+
+If Alice and Bob is happy with all the committed values up to know. Alice will create a Tari UTXO with the script mentioned above. 
+And because Bob already gave her the required signatures \\(a_{Srb}\\), \\(b_{Srb}\\) and \\(R_{Srb}\\) of the his part 
+of the aggregated signatures, she knows all the required information to spend this after the lock expired. 
+
+### XMR Payment
+
+If Bob cab see that Alice has published the Tari UTXO with the correct script, Bob can go ahead and publish the Monero UTXO
```
sdbondi, line 221:
```suggestion
If Bob cab see that Alice has published the Tari UTXO with the correct script and signature, Bob can go ahead and publish the Monero UTXO
```

RFC/src/RFC-0241_AtomicSwapXMR.md
SWvheerden: According to the original paper, yes. But I am sticking to the notation we used in the TariScript RFC: https://rfc.tari.com/RFC-0201_TariScript.html#transaction-output-changes
Here we used `a, b` because `v` is the symbol used for value in all MW documentation.

RFC/src/RFC-0241_AtomicSwapXMR.md
SWvheerden: There is no signature to check yet on the UTXO. It's just sitting there mined as an unspent output. 
The zero-knowledge proofs Bob needs to check for that UTXO, should have been checked before publishing the UTXO.

RFC/src/RFC-0241_AtomicSwapXMR.md
sdbondi: Ok sounds good

RFC/src/RFC-0241_AtomicSwapXMR.md
sdbondi: Surely to check that the agreed transaction was published you can't just look for the script, but this is a detail anyway so not important. The point is Bob finds it. Thanks for clarifying.

SWvheerden:
This has a security vulnerability in it. Closing this for now.

PR #3618, opened 2021-11-25
feat: add bulletproof rewind profiling

hansieodendaal:
Description
---
The following changes were made:
- Added bulletproof rewind profiling measurements.
- Improved wallet recovery resiliency with improved connection logic to the selected base node(s).
- Improved recovering of seed words usability.
- Improved some struct member names to be less confusing.
- Improved recovery UTXO streaming performance by tuning the chunking parameter.

**Results**
The following tests were run using standard Tor connectivity on a laptop with an Intel(R) Core(TM) i7-7820HQ CPU and basic Samsung NVMe storage device.

| Number of outputs               | 52654  | 52284  | 52753  | 52762 | 52762  |
| ------------------------------- | ------ | ------ | ------ | ----- | ------ |
| Chunk size                      | 10     | 100    | 250    | 125   | 125    |
| Stream form base node time (ms) | 197678 | 108757 | 236872 | 96274 | 115798 |
| -  per output (ms)              | 3.754  | 2.080  | 4.490  | 1.825 | 2.195  |
| -  normalized                   | 0.836  | 0.463  | 1      | 0.406 | 0.489  |
| Total processing time (ms)      | 77157  | 72653  | 69005  | 67170 | 68563  |
| -  per output (ms)              | 1.465  | 1.39   | 1.308  | 1.273 | 1.299  |
| -  normalized                   | 1      | 0.949  | 0.893  | 0.869 | 0.887  |
| Bulletproofs rewind time (ms)   | 10877  | 11040  | 10665  | 10854 | 10702  |
| -  per output (ms)              | 0.207  | 0.211  | 0.202  | 0.206 | 0.203  |
| -  normalized                   | 0.981  | 1      | 0.957  | 0.976 | 0.962  |

**Discussion**
- As can be seen, Bulletproof rewind performance was consistently measured at ~0.21ms per output.
- Total processing time, which includes all database processing and Bulletproof rewinding, averaged at about 1.35ms per output.
- Optimum streaming performance was found to be ~2ms per output.
- Output streaming performance was found to be closely related to the chunk size that are requested from the base node. Smaller (10) and larger (250) quantities were the worst performers, while a quantity in the range of 100 to 125 was approximately twice as fast. _For this PR a final chunk size value of 125 was chosen._

Motivation and Context
---
- Further code improvements can be based on informed decisions using the profiling information as basis.

How Has This Been Tested?
---
- Unit testing.
- Cucumber testing.
- System-level testing doing wallet recoveries.


applications/tari_console_wallet/src/recovery.rs
```diff
@@ -105,7 +105,8 @@ pub async fn wallet_recovery(wallet: &WalletSqlite, base_node_config: &PeerConfi
 
     let mut recovery_task = UtxoScannerService::<WalletSqliteDatabase>::builder()
         .with_peers(peer_public_keys)
-        .with_retry_limit(3)
+        // Do not make this a small number as wallet recovery needs to be resilient
+        .with_retry_limit(100)
```
philipr-za: We made this smaller on purpose, if you have gone through your whole peer list 3 times something larger is wrong that trying 97 more times is not going to solve.

base_layer/wallet/src/utxo_scanner_service/utxo_scanning.rs
```diff
@@ -254,11 +255,20 @@ where TBackend: WalletBackend + 'static
...
                     error: e.to_string(),
                 });
+                // No use re-dialing a peer that is not responsive for recovery mode
+                if self.mode == UtxoScannerMode::Recovery {
+                    if let Ok(Some(connection)) = self.resources.comms_connectivity.get_connection(peer.clone()).await {
+                        if connection.clone().disconnect().await.is_ok() {
+                            debug!(target: LOG_TARGET, "Disconnected base node peer {}", peer);
+                        }
+                    };
+                    sleep(Duration::from_secs(30));
```
philipr-za: It is very bad to sleep a thread in an async task. If you really want to sleep you should use the tokio async sleep.

philipr-za (commented):
Most these changes look fine, a couple comments though.

applications/tari_console_wallet/src/recovery.rs
hansieodendaal: Trying to connect to a single peer in the list when the base node connection was not responsive posed to be a problem. I think it is much better for the user to stop the recovery process eventually than to restart it every time and hope for a quick connection.

philipr-za: So it's a small enough issue that I don't mind it going in but I still think that trying a 100 times when you have tried 3 times across all your peers and that many failed attempts tells you there is something wrong with the peers or your setup and you need to take action.

So I don't mind this going in BUT we use this in the Daily tests for that exact purpose, to tell us there is a problem with recovery. So if you want this to be 100 you need to cater for it in the way the Daily recovery test is launched so that it only has 3 retries.

hansieodendaal: Added configurable retry limit with default value 3.

philipr-za (approved):

PR #3619, opened 2021-11-25
fix!: console wallet grpc_console_wallet_addresss config

StriderDM:
Description
---
This PR changes `grpc_console_wallet_address` to `grpc_address` and moves it into the `wallet` section of the config.

It also allows a grpc address to be specified via the `--grpc-address` command line argument for the console_wallet (using standard multiaddr formatting).

Additionally, fixes a bug due to a missing config default.

How Has This Been Tested?
---
cargo test --all
manually (`cargo run --bin tari_console_wallet -- --grpc-address /ip4/127.0.0.1/tcp/18144` + bloomRPC client)


common/src/configuration/utils.rs
```diff
@@ -245,7 +242,11 @@ pub fn default_config(bootstrap: &ConfigBootstrap) -> Config {
 
     cfg.set_default("wallet.base_node_service_peers", Vec::<String>::new())
         .unwrap();
-
+    cfg.set_default(
```
stringhandler: ?

StriderDM: Was removed in PR#3600, a new wallet would fail when trying to create it. Included both fixes, this one and `.unwrap_or_else(Multiaddr::empty)`.  One can be removed depending on which we would want to keep.

philipr-za: This was due to a bug introduced in #3600 but I don't think we need to add this default as it is a deprecated setting. In #3620 I provided an alternate fix to this issue so I think this can come out.

StriderDM: Will remove it.

philipr-za (approved):

PR #3622, opened 2021-11-26
docs: update RFC-0230_HTLC to use TariScript

philipr-za:
Description
---
This PR updates RFC-0230 from describing the thoeretical method that a traditional MW blockchain would need to use to implement an HTLC to using a TariScript based approach.


SWvheerden (dismissed):
LGTM

RFC/src/RFC-0230_HTLC.md
```diff
@@ -84,32 +72,50 @@ This adds the following requirement for a [base node]:
...
+CheckHeightVerify(4000)
+```
+
 #### Hashed Time-locked Contract
-Hashed time-locked contracts are a way of reserving funds for a certain payment, but they only pay out to the receiver if
-certain conditions are met. If these conditions are not met within a time limit, the funds are paid back to the sender.
+Hashed time-locked contracts ([HTLC]) are a way of reserving funds that can only be spent if a hash pre-image can be provided or
+if a specified amount of time has passed. The hash pre-image is a secret that can be revealed under the right conditions
+to enable spending of the UTXO before the time-lock is reached. The secret can be revealed directly by a party or by
+a party spending another output via a signature.
```
sdbondi:
```suggestion
to enable spending of the UTXO before the time-lock is reached. The secret can be directly exchanged between the parties or 
revealed to the other party by spending an output that makes use of an adapter signature.
```

sdbondi (dismissed):
LGTM

RFC/src/RFC-0230_HTLC.md
delta1: adapter -> adaptor

sdbondi (approved):

PR #3624, opened 2021-11-28
fix: use json 5 for tor identity (regression)

sdbondi:
Description
---

Use json 5 for save/load identity functions 

Motivation and Context
---
Save and load identity were using older version of the json parser which coul not parse the json comments.
This causes a new tor identity and onion address to be created on each base node startup.

How Has This Been Tested?
---
Onion address does not change on base node restart


hansieodendaal (approved):
LGTM

delta1 (approved):

PR #3625, opened 2021-11-29
feat: improve wallet responsiveness

hansieodendaal:
Description
---
Improved wallet responsiveness by cutting down on unnecessary wallet database calls:
- `TransactionEvent::TransactionValidationSuccess` was published regardless of any state change in the wallet database, which resulted in many unnecessary `fetch 'All Complete Transactions'` calls to the wallet database. This issue was optimized to only send the event when a state change was affected.
- A missing event was added when state changed due to reorgs.
- This should have a slight positive impact on mobile wallet battery usage.

Motivation and Context
---
Unnecessary wallet database calls to fetch all completed transactions were made each time the transaction validation protocol was run, especially when nothing changed in the database. This effect grew linearly with the amount of transactions in the database,

How Has This Been Tested?
---
- Unit tests
- Cucumber tests (`npm test -- --tags "not @long-running and not @broken"`)
- System level tests

### DIFF
--- a/applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
+++ b/applications/tari_console_wallet/src/ui/state/wallet_event_monitor.rs
@@ -119,7 +119,7 @@ impl WalletEventMonitor {
                                         self.trigger_balance_refresh();
                                         notifier.transaction_sent(tx_id);
                                     },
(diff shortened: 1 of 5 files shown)


philipr-za (approved):
LGTM

PR #3626, opened 2021-11-29
feat: add connect per site into the web extension

Cifko:
Description
---
Add connect feature to tari web extension. It allows user to grand permission to the extension. Currently there is only true/false permission, but the code is ready to handle specific permissions. 

Future work
---
- We need carefully audit the code to see if there is no backdoor (e.g. any site getting access without triggering the user confirmation). I currently don't see any, but the ports and window.postMessage that are used to communicate between the 3 parties involved (background.js, content.js, web).
- Disconnect feature, revoking access to the extension
- UI
- There is a popup flash when you didn't do the onboarding yet and you want to connect it to the website.

How Has This Been Tested?
---
Manually.


delta1 (approved):
LGTM untested

stringhandler (approved):

PR #3623, opened 2021-11-26
docs: rfc-0241 tari-monero atomic swap

SWvheerden:
Description
---
This document adds in RFC-0241 which describe Tari Monero atomic swaps

Motivation and Context
---
This is required if we want to be able to do Atomic swaps between Tari and Monero.

PR #3627, opened 2021-11-29
docs: tari-monero atomic swap additional method

SWvheerden:
Description
---
This provides an additional method to do atomic swaps with Monero without the complicated dance required for the adaptor signatures. This is an alternative method to #3623 

This adds in an md file: `RFC-0241_AtomicSwapXMR_v2.md` to describe the process but this is not linked to in the links so that normal browsing if the RFC's will not expose this, but this is documented.


SWvheerden:
Closing this for #3623
Moved all the information into that PR

PR #3628, opened 2021-11-29
fix!: multiple monerod addresses in tari merge mining proxy

StriderDM:
Description
---
This PR allows multiple monero daemons to be specified in the configuration of tari_merge_mining_proxy, this will help ensure that the proxy is still able to service requests in the event a monero daemon in the list were to become unreachable whether it be temporarily or permanently.

Motivation and Context
---
Reliability

How Has This Been Tested?
---
cargo test --all
Manually
nvm use 12.22.6 && node_modules/.bin/cucumber-js --profile "ci" --tags "not @long-running and not @broken and @merge-mining"

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,7 +614,7 @@ they are not enabled already:
   ```
   ```
   [base_node.weatherwax]
-  transport = "tor"
+  transpo*_r_*t = "tor"
   allow_test_addresses = false
   grpc_enabled = true
   grpc_base_node_address = "127.0.0.1:18142"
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -627,7 +627,20 @@ And then depending on if you are using solo mining or self-select mining you wil
 - For the Tari Merge Mining Proxy, under section **`merge_mining_proxy.weatherwax`**
   ```
   [merge_mining_proxy.weatherwax]
-  monerod_url = "http://monero-stagenet.exan.tech:38081"
+  monerod_url = [ # stagenet
```
philipr-za, line 630: There is a lot of repetition of these values here in the README. In fact I see there are quite a lot of config option in this document that I think shouldn't be here for cleanliness sake. I propose you put these options into the sample config file in the correct places, with some comments on what they are and then in the Readme just discuss them if they need to be discussed.

philipr-za (commented):
Generally looks fine but I think the repetition in the README needs a bit of a clean up.

philipr-za (approved):

PR #3629, opened 2021-11-30
feat(comms)!: add signature to peer identity to allow third party identity updates

sdbondi:
Description
---

- add signature to peers with update timestamp to allow updates
- update peer sync, identity, discovery and join protocols to validate signature (backward compatible)
- signature is optional, if not provided the peer addresses is not updated
- add migration for peer db
- remove old migrations
- add some unit tests
- include PR #3624 
- fix: use of `from_timestamp` on unvalidated data can cause a panic, replaced with `from_timestamp_opt`
- decouple comms cipher key from wallet and implement identity signing after loading (👁️ 🙏 @philipr-za )
- add list connected public keys to FFI (👁️ 🙏  @StriderDM)
- update cucumber tests as needed

Compatibility
PeerDB: All changes are forward compatible, once upgraded the previous peer db cannot be used.
DhtProtocol: fully compatible, identity signature is optional and will have the same effect as the current protocol if not provided
Identity Protocol: fully compatible, identity signature is optional but SHOULD be provided once upgraded
NodeIdentity json file: will be updated on node startup

Motivation and Context
---
Allow address changes to be securely propagated through the network by third party peers

How Has This Been Tested?
---
Some basic unit tests
Compatibility: manually with two upgraded base nodes connecting to existing weatherwax nodes, console wallet tested


applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -133,6 +131,10 @@ pub fn load_identity<P: AsRef<Path>>(path: P) -> Result<NodeIdentity, String> {
             e.to_string()
         )
     })?;
+    // Check weather the previous version has a signature and sign if necessary
```
delta1:
```suggestion
    // Check whether the previous version has a signature and sign if necessary
```

applications/tari_app_utilities/src/identity_management.rs
```diff
@@ -214,16 +195,15 @@ pub fn load_from_json<P: AsRef<Path>, T: MessageFormat>(path: P) -> Result<T, St
 ///
 /// ## Returns
 /// Result to check if successful or not, string will indicate reason on error
-pub fn save_as_json<P: AsRef<Path>, T: MessageFormat>(path: P, object: &T) -> Result<(), String> {
-    let json = object.to_json().unwrap();
+pub fn save_as_json<P: AsRef<Path>, T: Serialize>(path: P, object: &T) -> Result<(), String> {
+    let json = json5::to_string(object).unwrap();
```
delta1: unwrap -> map_err ?

delta1 (commented):
Nice looking really good, seems to be a build issue that's blocking CI

delta1:
Needs conflict resolution

sdbondi:
@delta1 Done

delta1:
That FFI test seems to be failing reliably. Going to try it out locally but I think there's may be a breaking change there?

delta1:
Or is it just a flaky test?

sdbondi:
@delta1 Seems worth looking into since this touches FFI - will take a look

EDIT: Will give it a second run, since it's a bit heavy to run on my machine right now

delta1:
Same issue reproduced locally @sdbondi

sdbondi:
@delta1 Ended up being an unrelated outbound messaging issue where the base node holds onto the outbound messaging stream for longer than it should after the peer disconnects (FFI_WALLET in this case), sending the SAF messages on a stale stream so it never arrives even though it was successfully requested by the new connection after the restart. Added a change that immediately terminates the outbound stream on peer disconnect which seems to have fixed it (and improved the overall reliability of the messaging protocol). Not clear why this becomes more prevalent in this PR. Thanks for picking up the failed FFI test.

PR #3631, opened 2021-11-30
test: removed stress test log target

hansieodendaal:
Description
---
Removed the temporary stress test log target that is not in use any more.

Motivation and Context
---
Dead code needs to be removed.

How Has This Been Tested?
---
Compile test


philipr-za (approved):
LGTM

PR #3636, opened 2021-12-01
fix: get-peer command works with public key again

sdbondi:
Description
---
Fix regression on get-peer, arg takes (partial) node id and a public key/emojiid

Motivation and Context
---

get-peer should work with full pk

How Has This Been Tested?
---
Manually


delta1 (approved):
Nice, tested - works great!

PR #3637, opened 2021-12-01
feat: bad block list for invalid blocks after sync

sdbondi:
Description
---
Maintain a list of "bad blocks" that should never be accepted
Remove pending headers if block sync fails to validate the block body against the (now invalid) headers.

Motivation and Context
---
(Partially) valid headers may become invalid (MMR) once the block body is synced. This PR removes those headers and adds the offending block to a banned block list.

How Has This Been Tested?
---
Some new unit tests, basic manual test


stringhandler (dismissed):

delta1:
Looks like a test might need updating?

stringhandler (dismissed):

therustmonk (dismissed):

stringhandler (approved):

PR #3638, opened 2021-12-02
feat: prevent banning of connected base node in wallet

StriderDM:
Description
---
This PR excludes the connected base node in the wallet from being banned.

Motivation and Context
---
Usability

How Has This Been Tested?
---
cargo test --all

### DIFF
--- a/comms/dht/src/connectivity/mod.rs
+++ b/comms/dht/src/connectivity/mod.rs
@@ -67,6 +67,7 @@ pub struct DhtConnectivity {
     config: DhtConfig,
     peer_manager: Arc<PeerManager>,
     node_identity: Arc<NodeIdentity>,
+    message_rate_immune_node: Arc<Option<NodeId>>,
     connectivity: ConnectivityRequester,
     dht_requester: DhtRequester,
     /// List of neighbours managed by DhtConnectivity ordered by distance from this node
@@ -102,6 +103,7 @@ impl DhtConnectivity {
             config,
             peer_manager,
             node_identity,
+            message_rate_immune_node: Arc::new(None),
             connectivity,
             dht_requester,
             metrics_collector,
@@ -245,17 +247,24 @@ impl DhtConnectivity {
             .await?;
 
         for (peer, mps) in nodes {
-            warn!(
-                target: LOG_TARGET,
-                "Banning peer `{}` because of flooding. Message rate: {:.2}m/s", peer, mps
-            );
-            self.connectivity
-                .ban_peer_until(
-                    peer,
-                    self.config.ban_duration,
-                    "Exceeded maximum message rate".to_string(),
-                )
-                .await?;
+            if *self.message_rate_immune_node != Some(peer.clone()) {
+                warn!(
+                    target: LOG_TARGET,
+                    "Banning peer `{}` because of flooding. Message rate: {:.2}m/s", peer, mps
+                );
+                self.connectivity
+                    .ban_peer_until(
+                        peer,
+                        self.config.ban_duration,
+                        "Exceeded maximum message rate".to_string(),
+                    )
+                    .await?;
+            } else {
+                warn!(
+                    target: LOG_TARGET,
+                    "Peer `{}` immune from banning. Message rate: {:.2}m/s", peer, mps
+                );
+            }
         }
         Ok(())
     }
@@ -469,6 +478,9 @@ impl DhtConnectivity {
             ConnectivityStateOffline => {
                 debug!(target: LOG_TARGET, "Node is OFFLINE");
                 self.refresh_peer_pools().await?;
+            },
+            SetMessageRateImmuneNode(node_id) => {
+                self.message_rate_immune_node = Arc::new(Some(node_id));
             },
             _ => {},
         }
--- a/comms/src/connectivity/manager.rs
+++ b/comms/src/connectivity/manager.rs
@@ -274,6 +274,11 @@ impl ConnectivityManagerActor {
                     error!(target: LOG_TARGET, "Error when banning peer: {:?}", err);
                 }
             },
+            SetMessageRateImmuneNode(node_id) => {
+                if let Err(err) = self.message_rate_immune_peer(&node_id).await {
+                    error!(target: LOG_TARGET, "Error when excluding peer from banning: {:?}", err);
+                }
+            },
             GetActiveConnections(reply) => {
                 let _ = reply.send(
                     self.pool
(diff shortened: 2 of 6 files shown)


comms/dht/src/connectivity/mod.rs
```diff
@@ -67,6 +70,7 @@ pub struct DhtConnectivity {
     config: DhtConfig,
     peer_manager: Arc<PeerManager>,
     node_identity: Arc<NodeIdentity>,
+    ban_exclusions: Arc<RwLock<Vec<NodeId>>>,
```
stringhandler: Can this not just be local? I think the other Arc's are shared amongst multiple different structs, whereas there should only be one DhtConnectivity

```suggestion
    ban_exclusions: Vec<NodeId>,
```

comms/src/connectivity/requester.rs
```diff
@@ -54,6 +54,7 @@ pub enum ConnectivityEvent {
     PeerConnected(PeerConnection),
     PeerConnectFailed(NodeId),
     PeerBanned(NodeId),
+    ExcludePeerBan(NodeId),
```
stringhandler: Is there something that subscribes to this event? I don't know if it's worth publishing it

stringhandler (commented):
I would prefer the term "allow_list" and "add_peer_to_allow_list" rather than "exclude_ban_peer".

comms/src/connectivity/requester.rs
philipr-za: It looks like this is being used to communicate to the DHT Connectivity Actor that doesn't have a direct handle because it doesn't provide a service API. However, I don't think this is really the ideal way to plumb this communication in. While it might make sense to add a peer to the Allowlist via the Comms::connectivity manager how do you manage that list afterwards in terms of seeing who is on it and removing peers from it?

philipr-za (commented):
This functionality would typically be called an Allowlist (whitelist in the past).

I don't think it is ideal to make the interface for adding to the allowlist a call via the Comms::connectivity manager. I think you should add service handle to the Dht::connectivity actor to build out a service API. The reason is that you will not just be adding to the allowlist but also querying who is on it and removing peers from it.

This PR adds to the allowlist but doesn't remove the previous base node from the allow list when a new one is selected as described in the ticket.

comms/dht/src/connectivity/mod.rs
```diff
@@ -67,6 +70,7 @@ pub struct DhtConnectivity {
     config: DhtConfig,
     peer_manager: Arc<PeerManager>,
     node_identity: Arc<NodeIdentity>,
+    message_rate_immune_node: Arc<RwLock<Option<NodeId>>>,
```
sdbondi: Does not need to be locked - this state belongs to one task/service

comms/src/connectivity/manager.rs
```diff
@@ -274,6 +274,11 @@ impl ConnectivityManagerActor {
                     error!(target: LOG_TARGET, "Error when banning peer: {:?}", err);
                 }
             },
+            MessageRateImmuneNode(node_id) => {
```
sdbondi: Should start with a verb
```suggestion
            AddMessageRateImmuneNode(node_id) => { // or AddPeerToAllowList in comms connectivity
```

sdbondi (commented):
Not sure why the base node would be sending so many messages to the wallet - that doesn't seem right. 

Assuming there is a valid reason to add this, an allowlist of peers that cannot be banned can either be added to comms connectivity (responsible for banning peer's it is asked to - "don't ever ban these peers") or DHT connectivity (responsible for checking message rates - "don't ban this peer for message rate violation"). 

I think the former makes more sense, and there is already a service API to work with.

philipr-za:
> Not sure why the base node would be sending so many messages to the wallet - that doesn't seem right.

Observed this occurring during a stress test and the base node was relaying SAF messages at a high rate.

sdbondi:
> Observed this occurring during a stress test and the base node was relaying SAF messages at a high rate.

SAF messages are sent in bundles of messages, should be `num_messages/batch_Size` messages so seems excessive - probably the protocol could be improved by rather RPC streaming SAF messages - in general, I would prefer that to adding new exceptions/rules for edge cases but happy for something like this to go in for now.

philipr-za:
> > Observed this occurring during a stress test and the base node was relaying SAF messages at a high rate.
> 
> SAF messages are sent in bundles of messages, should be `num_messages/batch_Size` messages so seems excessive - probably the protocol could be improved by rather RPC streaming SAF messages - in general, I would prefer that to adding new exceptions/rules for edge cases but happy for something like this to go in for now.

The sentiment is that there are several reasons that a wallet might ban a base node, with new ones being added in the future perhaps, but I would like to preserve the concept that a wallet will never ban the base node it has been explicitly set to use for any reason other than the WireMode byte. The WireMode byte though is enforced at a lower level I believe.

comms/dht/src/connectivity/mod.rs
```diff
@@ -245,17 +247,24 @@ impl DhtConnectivity {
...
-                "Banning peer `{}` because of flooding. Message rate: {:.2}m/s", peer, mps
-            );
-            self.connectivity
-                .ban_peer_until(
-                    peer,
-                    self.config.ban_duration,
-                    "Exceeded maximum message rate".to_string(),
-                )
-                .await?;
+            if *self.message_rate_immune_node != Some(peer.clone()) {
```
philipr-za, line 250: This should implemented as an allowlist in comms::connectivity

comms/src/connectivity/manager.rs
```diff
@@ -748,6 +753,13 @@ impl ConnectivityManagerActor {
         }
         Ok(())
     }
+
+    async fn message_rate_immune_peer(&mut self, node_id: &NodeId) -> Result<(), ConnectivityError> {
+        info!(target: LOG_TARGET, "Excluding peer {} from banning", node_id);
+        self.publish_event(ConnectivityEvent::SetMessageRateImmuneNode(node_id.clone()));
```
philipr-za, line 759: Using an event to communicate with a service is not a good design.

comms/src/connectivity/requester.rs
```diff
@@ -54,6 +54,7 @@ pub enum ConnectivityEvent {
     PeerConnected(PeerConnection),
     PeerConnectFailed(NodeId),
     PeerBanned(NodeId),
+    SetMessageRateImmuneNode(NodeId),
```
philipr-za, line 57: Rather than dealing with a specific case this should be an allowlist.

philipr-za (commented):
This should not be implemented in dht::connectivity where you need to communicate with the dht actor via an event stream. Please implement this as an Allowlist where the Bans are processed: in the comms::connectivity service. Again, we are not focusing on a single ban case, the idea is implement an allow list that prevents any bans from occurring for a node on the allow list.

StriderDM:
See PR #3642

PR #3639, opened 2021-12-03
feat: base_node switching for console_wallet when status is offline

StriderDM:
Description
---
This PR allows the console_wallet to periodically attempt to connect to another base_node in the list should it be found to be offline.

Motivation and Context
---
Usability.

How Has This Been Tested?
---
cargo test --all
manually


base_layer/wallet/src/base_node_service/mock_base_node_service.rs
```diff
@@ -100,6 +105,7 @@ impl MockBaseNodeService {
             is_synced: Some(true),
             updated: None,
             latency: None,
+            status: OnlineStatus::default(),
```
sdbondi: Not sure that a `Default` impl makes sense for `OnlineStatus`. I think rather be explicit:
```suggestion
            status: OnlineStatus::Connecting
```

applications/tari_console_wallet/src/ui/components/base_node.rs
```diff
@@ -42,9 +42,9 @@ impl BaseNode {
 impl<B: Backend> Component<B> for BaseNode {
     fn draw(&mut self, f: &mut Frame<B>, area: Rect, app_state: &AppState)
     where B: Backend {
-        let current_online_status = app_state.get_wallet_connectivity().get_connectivity_status();
+        let base_node_state = app_state.get_base_node_state();
```
sdbondi, line 46: I'm hesitant to duplicate the online status in the base node state, the wallet connectivity service is the source of truth for the connectivity status and bugs may creep in making sure the value in the base node state is always in-sync with what the connectivity service is saying.

base_layer/wallet/src/connectivity_service/service.rs
```diff
@@ -50,6 +50,12 @@ pub enum OnlineStatus {
     Offline,
 }
 
+impl Default for OnlineStatus {
```
sdbondi: Think this `Default` impl should be removed

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -181,6 +184,30 @@ impl AppState {
...
+                self.connectivity_check_cooldown = Some(Instant::now());
+                false
+            },
+        };
+
+        if update && self.wallet_connectivity.get_connectivity_status() == OnlineStatus::Offline {
+            let current = self.get_selected_base_node();
+            let list = self.get_base_node_list().clone();
+            let mut index: usize = list.iter().position(|(_, p)| p == current).unwrap_or_default();
+            if index == list.len() {
```
sdbondi: `index` can never equal the length of the list because it is an index in the list.
Should be 
```suggestion
            if index == list.len() - 1 {
```

or

```rust
let index = (index + 1) % list.len();
```

In both of these you need to make sure before that list isn't empty (i.e. `- 1 ` underflows or divide by zero) - pretty sure that the wallet exits with an error if that is the case, but thought I'd mention it in case

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -181,6 +184,30 @@ impl AppState {
...
+                index = 0;
+            } else {
+                index += 1;
+            }
+            let (_, next) = &list[index];
+            let _ = self.set_base_node_peer(next.clone()).await.unwrap_or_default();
```
sdbondi: Rather log this error

sdbondi (commented):
Looks good, some comments

applications/tari_console_wallet/src/ui/components/base_node.rs
StriderDM: I understand the hesitancy, though this is a draw method, we really should be using the variables from the state here.

base_layer/wallet/src/connectivity_service/service.rs
StriderDM: There is a Default::default() when creating a new state, so this cannot be removed since #derive(default) is marked as experimental on enums. I have been explicit with OnlineStatus::Connecting though.

applications/tari_console_wallet/src/ui/components/base_node.rs
philipr-za: I agree, the `draw` method shouldn't be making calls into the backend it should be using the cached values in the AppState

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -153,6 +153,7 @@ impl<B: Backend> App<B> {
 
     pub fn on_tick(&mut self) {
         Handle::current().block_on(self.app_state.update_cache());
+        Handle::current().block_on(self.app_state.check_connectivity());
```
philipr-za: This is not ideal. Connectivity state changes are all accompanied by events and so can be handled in the monitor when they arrive. The event based architecture means we do not need a complex polling + cooldown approach.

philipr-za (commented):
I don't agree with driving this process by a timed interval polling the connectivity status. It should be driven by the Connectivity Event stream.

applications/tari_console_wallet/src/ui/components/base_node.rs
sdbondi: The call is using a watch (it's not async), so it simply returns the currently set value - not sure if perf is the concern. In any case, no biggie since (for now) its set in one place by the monitor so happy if you would rather use the cached state.

applications/tari_console_wallet/src/ui/app.rs
StriderDM: Have removed the polling + cooldown.

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -125,11 +126,15 @@ where
                 timer.elapsed().as_millis()
             );
 
-            let base_node_id = match self.wallet_connectivity.get_current_base_node_id() {
+            let mut connectivity = self.wallet_connectivity.clone();
```
philipr-za: Don't understand why you need to clone this whole handle with all the channels and stuff inside it?

base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -48,6 +48,7 @@ pub struct BaseNodeState {
     pub is_synced: Option<bool>,
     pub updated: Option<NaiveDateTime>,
     pub latency: Option<Duration>,
+    pub status: OnlineStatus,
```
philipr-za: Oh I see I misinterpreted this in the previous iteration of this PR as being a struct that lives in the Console Wallet, It is definitely not right to duplicate this state in a separate service.

applications/tari_console_wallet/src/ui/components/base_node.rs
philipr-za: Excuse me I misread this comment in the previous version of this PR. I thought this struct lived only in the console wallet. It definitely should not be duplicated in the Base Node Monitoring Service state, that really isn't needed.

philipr-za (commented):
Looking better now that it is driven by the event. 

Something that is still missing from the original ticket spec is that the automatic switch should only happen is a custom base node is NOT set. If a custom base node is set then this should not happen.

Looking at it again I see now that the duplicated connectivity status was not just in the cached stated but all the way in the Base Node monitoring service which is not right.

philipr-za (dismissed):

philipr-za:
Clippy says no

philipr-za (approved):

PR #3640, opened 2021-12-06
feat(consensus)!: add tari script byte size limit check to validation

sdbondi:
Description
---
- Adds consensus rules for maximum script byte size

Breaking changes:
If an existing script is larger than 2048, that block/output will fail to validate on upgraded nodes. I have not explicitly checked this but the largest scripts used currently on testnet are one-sided payments and they are well within the limit.

Motivation and Context
---
See [@CjS77 comment](https://github.com/tari-project/tari-crypto/pull/65#issuecomment-971356808)

TODO: Ideally, in order to not negatively impact validation performance, we should not have to serialise the script again to gain its byte size.

How Has This Been Tested?
---
New unit test, basic manual test


delta1 (approved):
LGTM - agreed no current scripts in the chain should be anywhere near this size

PR #3641, opened 2021-12-06
feat!: use libtor to spawn in-process Tor for base node and console wallet

delta1:
Description
---
- base node and console wallet default to starting their own Tor instance
- 2 adjacent ports are selected from given range and tested to be available (this is for control and socks ports, not for the onion port)
- random password is generated and configured 
- data directory and logging file are temporary  
- new config option `use_libtor` for `[base_node.network]` and `[wallet.network]` defaults to true, when set to false allows Tor comms transport to be manually defined as before 

Motivation and Context
---
Improved user experience - ease of getting started without requiring separate Tor 

How Has This Been Tested?
---
Tested by setting up new base nodes and wallets and running them using libtor, and also tested that they still work using a system level Tor.
TODO: still to be tested on Windows

Breaking change: old config files *might* no longer work


delta1:
Going to need an update for build base node binaries CI

common/src/tor.rs
```diff
@@ -0,0 +1,204 @@
...
+        let temp = NamedTempFile::new()?.into_temp_path();
+        let file = temp.to_string_lossy().to_string();
+        instance.temp_file = Some(temp);
+        instance.log_destination = file;
+
+        Ok(instance)
+    }
+
+    /// Override a given Tor comms transport with the control address and auth from this instance
+    pub fn updated_comms_transport(&self, transport: CommsTransport) -> Result<CommsTransport, ExitCodes> {
```
sdbondi: `update_comms_transport`

common/src/tor.rs
```diff
@@ -0,0 +1,204 @@
...
+        tor.start_background();
+
+        loop {
+            tokio::select! {
+                _ = shutdown_signal.wait() => {
+                    info!(target: LOG_TARGET, "Shutdown signal received, dropping Tor");
+                    break;
+                }
+            }
+        }
```
sdbondi: ShutdownSignal is just a future that resolves when triggered

```suggestion
                shutdown_signal.wait().await;
                    info!(target: LOG_TARGET, "Shutdown signal received, dropping Tor");    
```

applications/tari_console_wallet/src/main.rs
```diff
@@ -108,6 +114,19 @@ fn main_inner() -> Result<(), ExitCodes> {
         return runtime.block_on(change_password(&global_config, arg_password, shutdown_signal));
     }
 
+    // Run our own Tor instance, if configured
+    if global_config.console_wallet_use_libtor &&
+        matches!(global_config.comms_transport, CommsTransport::TorHiddenService { .. })
+    {
+        let tor = Tor::randomize(19_050..20_000)?;
```
sdbondi, line 121: The tor port will not conflict with TCP ports (they are completely different protocols, tor port doesn't bind on a network socket) so the range can be wider - but otherwise, no problem keeping it in this range, though I may simplify this by removing it as an argument because basically the tor port doesn't matter/is just a number put on the end of the onion address.

 If I understand, this will change the address each time the node starts up. This may cause issues in connectivity because some nodes may not have up to date addresses (once #3629 goes in, it will update more quickly, but still eventual consistency, larger the network longer it may take for some nodes). It's better to change addresses fairly rarely (I would say just keep the address the same when possible).

common/src/tor.rs
```diff
@@ -0,0 +1,204 @@
...
+    let ports = port_range
+        .find(|port| ports_available(*port))
+        .map(|p| (p, p + 1))
+        .ok_or_else(|| io::Error::new(io::ErrorKind::AddrNotAvailable, "No ports available for Tor."))?;
+
+    Ok(ports)
+}
+
+/// Check if TCP `port` and `port + 1` are available to bind
+fn ports_available(port: u16) -> bool {
```
sdbondi, line 193: Ports on the tor protocol are not the same as TCP OS-level ports - they are "virtual" ports scoped to the hidden service address

sdbondi (commented):
Awesome, very cool that tor can be embedded. Changing the address each time will cause connectivity issues that can be hard to debug (nodes not connecting when they are online, not noticing the address has changed)

applications/tari_console_wallet/src/main.rs
delta1: No this is just the control and socks ports the tor identity is still used. Will just confirm that and make it more obvious

common/src/tor.rs
delta1: It’s for the local control and socks ports passed as arguments to libtor

common/src/tor.rs
delta1: Thanks I knew I was missing something lol

common/src/tor.rs
delta1: I had it that way before 🤣 will change

delta1:
Breaks the build on windows, looking for a solution

delta1:
> Awesome, very cool that tor can be embedded. Changing the address each time will cause connectivity issues that can be hard to debug (nodes not connecting when they are online, not noticing the address has changed)

Just to clarify - the public address remains the same and is loaded from the node identity file, it is just the Tor control port and socks port that are checked for availability in the port range and selected.

sdbondi:
> it is just the Tor control port and socks port that are checked for availability in the port range and selected.

Ok cool, sounds good. Then downgraded to a nit, maybe `Tor::randomize` could be `Tor::initialize` and remove the port range.

Warning about port bind test (from experience), the port may stay bound after the test even if it's dropped on linux (unless `SO_REUSEPORT` is used subsequently). You can let the OS assign an open port by using port 0

common/src/tor.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Two available adjacent TCP ports will be selected in the given
+    /// port range. These are scanned sequentially from start to end.
+    /// These ports are only for the control and socks ports, the onion address and port info are still loaded from the
+    /// node identity file.
+    pub fn randomize(port_range: Range<u16>) -> Result<Tor, ExitCodes> {
+        let mut instance = Tor::default();
+
+        // check for unused ports to assign
+        let (socks_port, control_port) = get_available_ports(port_range)?;
+        instance.socks_port = socks_port;
```
sdbondi, line 73: Just remembered, if you set both of these to `0` the OS will assign an open port.

delta1: In that case does the bind return the port number?

delta1:
> > it is just the Tor control port and socks port that are checked for availability in the port range and selected.
> 
> Ok cool, sounds good. Then downgraded to a nit, maybe `Tor::randomize` could be `Tor::initialize` and remove the port range.
> 
> Warning about port bind test (from experience), the port may stay bound after the test even if it's dropped on linux (unless `SO_REUSEPORT` is used subsequently). You can let the OS assign an open port by using port 0

Agreed sounds good

common/src/tor.rs
delta1: Okay I see I can use local_addr to get it from the listener. Will do

delta1:
Windows compilation issues would either require building via cygwin/mingw/msys2, or maintaining a fork and migrating the tor source to cmake for windows, both of which are undesirable for this feature. 

Closing this PR and will re-open as an opt-in feature for linux/macos only, unfortunately.

PR #3642, opened 2021-12-07
feat: prevent banning of connected base node in wallet

StriderDM:
Description
---

This PR excludes the connected base node in the wallet from being banned.

Motivation and Context
---

Usability

How Has This Been Tested?
---

cargo test --all

### DIFF
--- a/comms/src/connectivity/manager.rs
+++ b/comms/src/connectivity/manager.rs
@@ -93,6 +93,7 @@ impl ConnectivityManager {
             shutdown_signal: self.shutdown_signal,
             #[cfg(feature = "metrics")]
             uptime: Some(Instant::now()),
+            allow_list: vec![],
         }
         .spawn()
     }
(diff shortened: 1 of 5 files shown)


stringhandler (dismissed):

comms/src/connectivity/manager.rs
```diff
@@ -271,8 +273,20 @@ impl ConnectivityManagerActor {
                 let _ = reply.send(states);
             },
             BanPeer(node_id, duration, reason) => {
-                if let Err(err) = self.ban_peer(&node_id, duration, reason).await {
-                    error!(target: LOG_TARGET, "Error when banning peer: {:?}", err);
+                if !self.allow_list.contains(&node_id) {
```
philipr-za, line 276: Just add an info log message here that this Ban Request was ignored due to the AllowList so that this isn't ignored silently.

philipr-za (commented):
Looking good, one small comment

comms/src/connectivity/manager.rs
StriderDM: Will do.

philipr-za (approved):

PR #3651, opened 2021-12-08
feat: custom_base_node in config

StriderDM:
Description
---
This PR allows the custom_base_node_peer to be set via the config, env, or via the command line

Motivation and Context
---
Usability

How Has This Been Tested?
---
cargo test --all
manually


philipr-za (commented):
Sorry, I see that the ticket title was not correctly updated. The original idea evolved into just being able to override the stored custom base node via config file, command line switch or ENVAR. 

However, we do still need to perist a set custom base node if these three overrides are not being used so can you please put the db stuff back.

StriderDM:
@philipr-za Will do

PR #3652, opened 2021-12-08
refactor: change transaction mod name

stringhandler:
Description
---
Move transaction structs from `transaction_entities` to `transaction`

Motivation and Context
---
In a previous PR, these were moved from the single file `transaction` to `transaction_entities`. This was helpful but makes merging existing feature branches difficult. To make merges simpler, I have changed the name to `transaction` as a smaller step so that merges are less difficult.

How Has This Been Tested?
---
cargo test


philipr-za (approved):
lgtm

hansieodendaal (commented):
LGTM

PR #3654, opened 2021-12-10
feat: example to generate vanity node_ids

philipr-za:
Description
---
An example in tari_comms that generates vanity node_id's with a specified prefix. Used to generate node_ids for seed nodes.


sdbondi (dismissed):
LGTM

stringhandler (approved):

PR #3656, opened 2021-12-13
feat: covenants implementation

sdbondi:
Description
---

- Implementation of Covenants as per [RFC-0250](https://rfc.tari.com/RFC-0250_Covenants.html)
- Implement consensus encoding for some common types
- remove wrapper for consensus encoding tari script, just implement on the type 
- add consensus encoding impl for some common types

Motivation and Context
---
```rust
let covenant = covenant!(and(
    field_eq(
        @field::commitment, 
        @commitment(x),
    ), 
    fields_preserved(
        @fields(@field::features_unique_id, @field::features_parent_public_key)
    ))
);
covenant.execute(...)?;
```

Adding Covenants to inputs/outputs structs and validation is left for another PR - this is not a breaking change.

How Has This Been Tested?
---
Many new unit tests


base_layer/core/src/covenants/byte_codes.rs
```diff
@@ -0,0 +1,65 @@
...
+//  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+//  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//---------------------------------- ARG byte codes --------------------------------------------//
+pub(super) fn is_valid_arg_code(code: u8) -> bool {
+    (0x01..=0x0a).contains(&code)
```
delta1: Is `0x0a` used or should this be `0x01..0x0a` ?

base_layer/core/src/covenants/covenant.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+        CovenantTokenDecoder::new(&mut bytes).collect()
+    }
+
+    pub(super) fn write_to<W: io::Write>(&self, writer: &mut W) -> Result<usize, io::Error> {
+        CovenantTokenEncoder::new(self.tokens.as_slice()).write_to(writer)
+    }
+
+    pub fn validate(&self) -> Result<(), CovenantError> {
+        // TODO
```
delta1: is this still todo for this PR? or as part of the integration

base_layer/core/src/covenants/fields.rs
```diff
@@ -0,0 +1,329 @@
...
+        val.downcast_ref::<T>()
+    }
+
+    pub fn get_field_value_bytes(&self, output: &TransactionOutput) -> Vec<u8> {
+        use OutputField::*;
+        match self {
+            Commitment => output.commitment.to_consensus_bytes(),
+            Script => output.script.to_consensus_bytes(),
+            SenderOffsetPublicKey => output.sender_offset_public_key.to_consensus_bytes(),
+            Covenant => unimplemented!(),
```
delta1, line 107: Also to be implemented with the integration?

delta1 (commented):
Nice looks really great, just one possible off-by-1 in the range otherwise LGTM

base_layer/core/src/covenants/covenant.rs
sdbondi: So it's slightly towards a nice-to-have, because when it is executed and happens to be not valid, e.g. has invalid arguments for the filter, the block/txn validation will fail anyway - this is more about validating the correctness of the covenant which isn't so NB for now. Maybe I should remove this function.

base_layer/core/src/covenants/fields.rs
sdbondi: Yeah covenant field is added in the integration and this code is updated

delta1: Sweet

base_layer/core/src/covenants/byte_codes.rs
sdbondi: Good catch - 0x0a was removed at some point

delta1 (dismissed):

delta1:
🖇

stringhandler (approved):

PR #3661, opened 2021-12-15
fix: edge cases causing bans during header/block sync

sdbondi:
Description
---
Fixes the following edge cases:
- base node reorg while streaming blocks (including the reorged block) sends an error message to the peer to let it know to resolve reorg using header sync.
- lock add_block during header/block sync, sync protocols assume chain does not change while syncing
- check the local chain metadata at the time the base node kicked into header sync to decide to ban 

Motivation and Context
---
Fixes these erroneous ban conditions that were discovered during the recent stress test
1. Peer claimed an accumulated difficulty of 7632339559469659380317024 but validated difficulty was 7632339559469659380317024 <= local: 7632357293467858576437000.
2.  Banned sync peer because Peer claimed an accumulated difficulty of 7880400444975439614055394 but validated difficulty was 0 <= local: 7880400444975439
3. Peer sent hash for block header we do not have.

How Has This Been Tested?
---
Not explicitly tested, simple test syncing base node to tip


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -407,6 +407,15 @@ where T: BlockchainBackend + 'static
     ) -> Result<(), CommsInterfaceError> {
         let NewBlock { block_hash } = new_block;
 
+        if self.blockchain_db.inner().is_add_block_locked() {
```
stringhandler: nit: I prefer a name like `prevent_add_block` or `disable_add_block`

```suggestion
        if self.blockchain_db.inner().is_add_block_locked() {
```

stringhandler (dismissed):
Happy, let's see how it performs. The logic in the listening state is getting a bit hairy. We should do an audit of it sometime soon

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
sdbondi: Cool, I like `disable` better. As a consistency thing, I name functions that return a bool starting with `is_x` will rename to `is_add_block_disabled` and rename everything else along those lines

sdbondi:
Agreed, lots of edge cases that can potentially be handled more elegantly

PR #3666, opened 2021-12-17
feat: add GRPC call to search for utxo via commitment hex

SWvheerden:
Description
---
Added a GRPC call to search for the block a UTXO was mined in via the commitment.  UI command already exists, this adds a GPRC mapping to the same command. 

This fixes a UI bug that when a commitment is not found, says the the node is pruned and that the utxo was found.

### DIFF
--- a/applications/tari_app_grpc/proto/base_node.proto
+++ b/applications/tari_app_grpc/proto/base_node.proto
@@ -68,6 +68,8 @@ service BaseNode {
     rpc GetTipInfo(Empty) returns (TipInfoResponse);
     // Search for blocks containing the specified kernels
     rpc SearchKernels(SearchKernelsRequest) returns (stream HistoricalBlock);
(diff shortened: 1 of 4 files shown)


applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -289,6 +291,11 @@ message SearchKernelsRequest{
     repeated Signature signatures = 1;
 }
 
+// This is the request type for the Search Utxo rpc
+message SearchUtxosRequest{
+    repeated string commitments = 1;
```
stringhandler:
```suggestion
    repeated bytes commitments = 1;
```

stringhandler:
just needs fmt

stringhandler (dismissed):

stringhandler (approved):

PR #3682, opened 2022-01-05
refactor!: clean up #testnet reset TODOs

sdbondi:
Description
---
- cleans up a number of #testnet reset TODOs
- use consensus encoding for hash construction in relevant structs
- replace block version dependent kernel sorting
- input_mr without empty Bitmap data for all block versions

Breaking change because some igor/dibbler code used the v1 encoding

Motivation and Context
---
Clean up #testnetreset tags in preparation for testnet reset.

Some testnet reset tags remain in wallet transaction service 

How Has This Been Tested?
---
Existing tests pass

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -220,18 +220,13 @@ impl AggregateBody {
     }
 
     /// Sort the component lists of the aggregate body
-    pub fn sort(&mut self, version: u16) {
+    pub fn sort(&mut self) {
         if self.sorted {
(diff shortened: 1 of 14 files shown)


base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -220,18 +220,13 @@ impl AggregateBody {
     }
 
     /// Sort the component lists of the aggregate body
-    pub fn sort(&mut self, version: u16) {
```
delta1, line 223: looks like some tests need this updated too

PR #3684, opened 2022-01-06
feat: update faux transaction status when imported output is validated

philipr-za:
Description
---
Currently when a faucet UTXO is imported the Faux transaction that is created to track the event is created with the Imported status and reported to the FFI frontend as a Confirmed transaction. However, the output is not spendable until it has been validated. This PR changes the process so that initially the Faux transaction is reported as a Pending Inbound transaction to a FFI client and when a TXO validation is completed the transaction service will check if the output for the faux tx has become spendable at which point the Faux transaction will be marked as MinedConfirmed.

Also while writing the unit test for this code I thought of an easy way to make the Transaction Service testing harness much neater by adding a struct to collect all the testing interface components. This update account for most the lines in this PR.

How Has This Been Tested?
---
Rust unit tests provided


base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -2037,9 +2056,13 @@ impl UnconfirmedTransactionInfoSql {
                 completed_transactions::coinbase_block_height,
             ))
             .filter(
-                completed_transactions::mined_height
-                    .is_null()
-                    .or(completed_transactions::status.eq(TransactionStatus::MinedUnconfirmed as i32)),
+                completed_transactions::status
+                    .ne(TransactionStatus::Imported as i32)
```
SWvheerden, line 1995: Why are we excluding imported here?

SWvheerden (dismissed):
LGTM
I have one question regarding a SQL query. 
The testing harness is much improved with the struct, I like that.

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
delta1: A coinbase is never imported maybe?

philipr-za: because an imported transaction can never be "unconfirmed" it is either pending on confirmed.

philipr-za: For a bit more clarity this index method is used to pull transaction that need to be queried on a base node which imported transactions should not be.

delta1:
Looks good just a few clippys 📎

stringhandler (dismissed):

delta1 (approved):
LGTM, untested

PR #3688, opened 2022-01-10
feat: dibbler new genesis block with faucet utxos

delta1:
Description
---
Generates a new genesis block for `dibbler` network, with faucet utxos

Motivation and Context
---
Previous genesis block was missing faucet utxos

How Has This Been Tested?
---
- Relevant unit tests
- Built nodes, wallets, miner, started mining new chain, sent transactions between wallets
- Registered a new asset, minted tokens, but was unable to send token between wallets (could not find by unique id, I think that has changed recently)

### DIFF
--- a/applications/test_faucet/Cargo.toml
+++ b/applications/test_faucet/Cargo.toml
@@ -4,12 +4,14 @@ version = "0.22.1"
 authors = ["The Tari Development Community"]
 edition = "2018"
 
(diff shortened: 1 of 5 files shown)


SWvheerden (approved):
LGTM
Clippy is unhappy thou.

delta1:
> LGTM
> Clippy is unhappy thou.

yeah that's from the previous weatherwax merge I think

delta1:
#3690 fixes clippy

stringhandler (approved):

PR #3689, opened 2022-01-10
ci: add tor script to binaries bundle

stringhandler:
Add start tor script to bundled artifacts in Build Base Node Binaries Github Action


delta1 (dismissed):
LGTM untested

delta1 (approved):
yolo

PR #3693, opened 2022-01-11
feat: add version to Transacations

Cifko:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/base_layer/core/src/transactions/transaction/transaction_input.rs
+++ b/base_layer/core/src/transactions/transaction/transaction_input.rs
@@ -217,7 +217,7 @@ impl TransactionInput {
     /// features and script. This will ignore all other output and input fields
     pub fn is_equal_to(&self, output: &TransactionOutput) -> bool {
         match output.try_hash() {
-            Ok(hash) => self.output_hash() == hash,
+            Ok(hash) => self.output_hash() == Ok(hash),
             Err(_e) => false,
         }
     }
@@ -295,22 +295,25 @@ impl TransactionInput {
 
     /// Returns the hash of the output data contained in this input.
     /// This hash matches the hash of a transaction output that this input spends.
-    pub fn output_hash(&self) -> Vec<u8> {
-        match self.spent_output {
-            SpentOutput::OutputHash(ref h) => h.clone(),
-            SpentOutput::OutputData {
-                ref commitment,
-                ref script,
-                ref features,
-                ref covenant,
-                ..
-            } => HashDigest::new()
-                .chain(features.to_consensus_bytes())
-                .chain(commitment.as_bytes())
-                .chain(script.as_bytes())
-                .chain(covenant.to_consensus_bytes())
-                .finalize()
-                .to_vec(),
+    pub fn output_hash(&self) -> Result<Vec<u8>, String> {
+        match self.version {
+            0 => match self.spent_output {
+                SpentOutput::OutputHash(ref h) => Ok(h.clone()),
+                SpentOutput::OutputData {
+                    ref commitment,
+                    ref script,
+                    ref features,
+                    ref covenant,
+                    ..
+                } => Ok(HashDigest::new()
+                    .chain(features.to_consensus_bytes())
+                    .chain(commitment.as_bytes())
+                    .chain(script.as_bytes())
+                    .chain(covenant.to_consensus_bytes())
+                    .finalize()
+                    .to_vec()),
+            },
+            _ => Err("Newer version!".to_string()),
         }
     }
 
@@ -353,15 +356,15 @@ impl TransactionInput {
     }
 
     /// Return a clone of this Input into its compact form
-    pub fn to_compact(&self) -> Self {
-        Self::new(
+    pub fn to_compact(&self) -> Result<Self, String> {
+        Ok(Self::new(
             match &self.spent_output {
                 SpentOutput::OutputHash(h) => SpentOutput::OutputHash(h.clone()),
-                SpentOutput::OutputData { .. } => SpentOutput::OutputHash(self.output_hash()),
+                SpentOutput::OutputData { .. } => SpentOutput::OutputHash(self.output_hash()?),
             },
             self.input_data.clone(),
             self.script_signature.clone(),
-        )
+        ))
     }
 }
 
--- a/base_layer/core/src/transactions/transaction/transaction_input.rs
+++ b/base_layer/core/src/transactions/transaction/transaction_input.rs
@@ -25,11 +25,17 @@
 
 use std::{
     cmp::Ordering,
-    fmt::{Display, Formatter},
+    fmt::{self, Display, Formatter},
 };
 
 use blake2::Digest;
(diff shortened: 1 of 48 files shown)


base_layer/core/src/transactions/transaction/transaction_kernel.rs
```diff
@@ -61,7 +66,131 @@ pub struct TransactionKernel {
...
+                match version {
+                    0 => {
+                        let features = features.ok_or_else(|| de::Error::missing_field("features"))?;
+                        let fee = fee.ok_or_else(|| de::Error::missing_field("fee"))?;
+                        let lock_height = lock_height.ok_or_else(|| de::Error::missing_field("lock_height"))?;
+                        let excess = excess.ok_or_else(|| de::Error::missing_field("excess"))?;
+                        let excess_sig = excess_sig.ok_or_else(|| de::Error::missing_field("excess_sig"))?;
+                        Ok(TransactionKernel::new(features, fee, lock_height, excess, excess_sig))
+                    },
+                    _ => unimplemented!("new version needs implementing!"),
```
sdbondi: You'll need to return an error here (otherwise a peer can send an unsupported version and crash the node)

base_layer/core/src/transactions/transaction/transaction_kernel.rs
```diff
@@ -61,7 +66,131 @@ pub struct TransactionKernel {
...
+                match version {
+                    0 => {
+                        let features = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(1, &self))?;
+                        let fee = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(2, &self))?;
+                        let lock_height = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(3, &self))?;
+                        let excess = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(4, &self))?;
+                        let excess_sig = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(5, &self))?;
+                        Ok(TransactionKernel::new(features, fee, lock_height, excess, excess_sig))
+                    },
+                    _ => unimplemented!("new version needs implementing!"),
```
sdbondi: Return an error here

base_layer/core/src/transactions/transaction/transaction_input.rs
```diff
@@ -65,21 +70,119 @@ pub struct TransactionInput {
...
+                let version: u8 = version.ok_or_else(|| de::Error::missing_field("version"))?;
+                match version {
+                    0 => {
+                        let spent_output = spent_output.ok_or_else(|| de::Error::missing_field("spent_output"))?;
+                        let input_data = input_data.ok_or_else(|| de::Error::missing_field("input_data"))?;
+                        let script_signature =
+                            script_signature.ok_or_else(|| de::Error::missing_field("script_signature"))?;
+                        Ok(TransactionInput::new(spent_output, input_data, script_signature))
+                    },
+                    _ => unimplemented!("new version needs implementing!"),
```
sdbondi: Return an error

base_layer/core/src/transactions/transaction/transaction_input.rs
```diff
@@ -65,21 +70,119 @@ pub struct TransactionInput {
...
+                let version: u8 = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(0, &self))?;
+                match version {
+                    0 => {
+                        let spent_output = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(1, &self))?;
+                        let input_data = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(3, &self))?;
+                        let script_signature =
+                            seq.next_element()?.ok_or_else(|| de::Error::invalid_length(5, &self))?;
+                        Ok(TransactionInput::new(spent_output, input_data, script_signature))
+                    },
+                    _ => unimplemented!("new version needs implementing!"),
```
sdbondi: Return an error

base_layer/core/src/transactions/transaction/transaction_kernel.rs
```diff
@@ -88,15 +217,18 @@ impl Hashable for TransactionKernel {
...
-            .chain(&[self.features.bits()])
-            .chain(u64::from(self.fee).to_le_bytes())
-            .chain(self.lock_height.to_le_bytes())
-            .chain(self.excess.as_bytes())
-            .chain(self.excess_sig.get_public_nonce().as_bytes())
-            .chain(self.excess_sig.get_signature().as_bytes())
-            .finalize()
-            .to_vec()
+        match self.version {
+            0 => HashDigest::new()
```
sdbondi: We should include the version in the hash to prevent malleability

base_layer/core/src/transactions/transaction/transaction_output.rs
```diff
@@ -316,7 +465,10 @@ impl TransactionOutput {
 /// Implement the canonical hashing function for TransactionOutput for use in ordering.
 impl Hashable for TransactionOutput {
     fn hash(&self) -> Vec<u8> {
-        transaction::hash_output(&self.features, &self.commitment, &self.script)
+        match self.version {
```
sdbondi, line 368: I think we'll need to change the hashing to be fallible since it now depends on the correct version.
We'll also need to add consensus rules (block and transaction validators) to check that each incoming input,output and kernel is a supported version. 
Rather than change the Hashable trait, which should stay infallible, I think add a new fallible try_hash function.

base_layer/core/src/transactions/transaction/transaction_input.rs
```diff
@@ -65,21 +70,119 @@ pub struct TransactionInput {
     pub script_signature: ComSignature,
 }
 
+impl<'de> Deserialize<'de> for TransactionInput {
```
sdbondi: Maybe move the Deserialize `impl`s below the "main" impl for readability (here and other places)

sdbondi (commented):
Using serde for de/serialisation of objects in the db is ok but you will need to implement serialize too (you cannot convert to a new version because this potentially changes the hash of the input,output,kernel - each must be stored in the version they came in) 

Consensus encoding ensures that we have a consistent well-defined byte-for-byte encoding for structs/fields which form part of a hash  (e.g. output features) and we never construct a hash of kernel, input, output using serde, which is format agnostic and not appropriate for consensus.

Maybe this table is clearer:

| Use case                              	| Appropriate encoding                                                                                	|
|---------------------------------------	|-----------------------------------------------------------------------------------------------------	|
| Storage                              	| serde / consensus encoding / protobuf                                                               	|
| Network                               	| protobuf + consensus encoding where appropriate (e.g script, covenant, public keys, signatures etc) 	|
| Consensus (esp for hash construction) 	| consensus encoding                                                                                  	|

consensus encoding is appropriate for each use-case and, for storage, also means we can change field names, field order and even potentially add new fields without breaking the blockchain db. However that will require a much larger refactor, so happy with this approach which matches the previous approach with added versioning support.

PR #3694, opened 2022-01-11
docs: add rfc-0203 stealth addresses to the RFCs

philipr-za:
Description
---
This PR adds a new RFC for the use of stealth addresses in one-sided payments


RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,115 @@
...
+the [Cryptonote] whitepaper and by [Peter Todd] which went on to be used in Monero. These formulations were very similar 
+to the [BIP-32] style of address generation. Later in 2014 a developer called rynomster/sdcoin proposed a further 
+improvement to the scheme that he called the Dual-Key Stealth Address Protocol (DKSAP) that allowed for a separate 
+scanning key and spending key. Since then there have been many variations of DKSAP proposed, but generally they only 
+offer performance optimizations for certain scenarios or add an application-specific feature. For our application, DKSAP 
+will do the job.
+
+## Dual-key Stealth Addresses
+
+The Dual-key Stealth Address Protocol (DKSAP) uses two key-pairs for the recipient of a [one-sided transaction],
```
delta1: the link for `one-sided transaction` is missing

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,115 @@
...
+The Dual-key Stealth Address Protocol (DKSAP) uses two key-pairs for the recipient of a [one-sided transaction], 
+\\( A = a \cdot G \\) and \\( B = b \cdot G \\). Where \\( a \\) is called the scan key and \\( b \\) is the spend key.
+A recipient will distribute the public keys out of band to receive [one-sided payments].
+
+The protocol that a sender will use to make a payment to the recipient is as follows:
+1. Sender generates a random nonce key-pair \\( R = r \cdot G \\).
+2. Sender calculates a ECDH shared secret \\(c = H( r \cdot a \cdot G ) = H( a \cdot R) = H( r \cdot A) \\), where
+\\( H( \cdot ) \\) is a cryptographic hash function.
+3. The sender will then use \\( K_s = c \cdot G + B \\) as the last public key in the [one-sided payment] script. 
+4. The sender will communicate R to the sender by including it in the script so the recipient can parse it but
```
delta1:
```suggestion
4. The sender will communicate \\( R \\) to the receiver by including it in the script, so the recipient can parse it but 
```

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,115 @@
...
+4. The sender will communicate R to the sender by including it in the script so the recipient can parse it but 
+dropping it during script execution. This changes the script for a [one-sided payment] from 
+`PushPubkey(K_s)` to `PushPubkey(R) Drop PushPubkey(K_s)`.
+
+The recipient will need to scan the blockchain for outputs that contain scripts of the [one-sided payment] form, and when
+one is found they will need to do the following:
+1. Extract the nonce \\( R \\) from the script.
+2. Use the public nonce to calculate the shared secret \\(c = H( a \cdot R) \\)
+3. Calculate \\( K_s \\) and check if it exists in the script.
+4. If it exists, the recipient can produce the script signature required using the private key calcualted by \\( c + b \\).
```
delta1:
```suggestion
4. If it exists, the recipient can produce the script signature required by using the private key calculated by \\( c + b \\). 
```

RFC/src/RFC-0203_StealthAddresses.md
delta1: I think this whole sentence could be a bit clearer with something like: 

"The sender includes R in the output script for the receiver, but it is then dropped since it is not required for the script execution."

delta1 (dismissed):
Very nice, a few minor comments

delta1 (approved):

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Dual-key Stealth Addresses
+
+The Dual-key Stealth Address Protocol (DKSAP) uses two key-pairs for the recipient of a [one-sided payment], 
+\\( A = a \cdot G \\) and \\( B = b \cdot G \\). Where \\( a \\) is called the scan key and \\( b \\) is the spend key.
+A recipient will distribute the public keys out of band to receive [one-sided payments].
+
+The protocol that a sender will use to make a payment to the recipient is as follows:
+1. Sender generates a random nonce key-pair \\( R = r \cdot G \\).
+2. Sender calculates a ECDH shared secret \\(c = H( r \cdot a \cdot G ) = H( a \cdot R) = H( r \cdot A) \\), where
```
SWvheerden, line 87:
```suggestion
2. Sender calculates a ECDH shared secret \\(c = H( r * a \cdot G ) = H( a * R) = H( r * A) \\), where
```

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+\\( H( \cdot ) \\) is a cryptographic hash function.
```
SWvheerden, line 88:
```suggestion
\\( H( ... ) \\) is a cryptographic hash function.
```

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+3. The sender will then use \\( K_s = c \cdot G + B \\) as the last public key in the [one-sided payment] script. 
+4. The sender includes  \\( R \\) for the receiver but dropping it as it is not required during script execution. 
+This changes the script for a [one-sided payment] from `PushPubkey(K_s)` to `PushPubkey(R) Drop PushPubkey(K_s)`.
+
+The recipient will need to scan the blockchain for outputs that contain scripts of the [one-sided payment] form, and when
+one is found they will need to do the following:
+1. Extract the nonce \\( R \\) from the script.
```
SWvheerden, line 95: How does the script look, and where do we store R in the script.

SWvheerden (commented):
This looks good I made two adjustments to the math to make it more clear what's going on there.

The biggest hole I have left here is how the scripting looks. 

But I suspect we can adapt this to rather than calculate R as a separate nonce, we can use K_O as the nonce and reuse that. That way we don't have to store that in the script and extract that from the script. We can just read it from the output. And the script can stay something basing like `PushPubkey(K_S)`
K_O is supposed to be random anyway.

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+
+The Dual-key Stealth Address Protocol (DKSAP) uses two key-pairs for the recipient of a [one-sided payment], 
+\\( A = a \cdot G \\) and \\( B = b \cdot G \\). Where \\( a \\) is called the scan key and \\( b \\) is the spend key.
+A recipient will distribute the public keys out of band to receive [one-sided payments].
+
+The protocol that a sender will use to make a payment to the recipient is as follows:
+1. Sender generates a random nonce key-pair \\( R = r \cdot G \\).
+2. Sender calculates a ECDH shared secret \\(c = H( r \cdot a \cdot G ) = H( a \cdot R) = H( r \cdot A) \\), where
+\\( H( \cdot ) \\) is a cryptographic hash function.
+3. The sender will then use \\( K_s = c \cdot G + B \\) as the last public key in the [one-sided payment] script.
```
SWvheerden, line 89:
```suggestion
3. The sender will then use \\( K_S = c \cdot G + B \\) as the last public key in the [one-sided payment] script. 
```
Perhaps we can add this link as well: [Script Key](https://rfc.tari.com/Glossary.html#script-keypair)

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+\\( A = a \cdot G \\) and \\( B = b \cdot G \\). Where \\( a \\) is called the scan key and \\( b \\) is the spend key.
+A recipient will distribute the public keys out of band to receive [one-sided payments].
+
+The protocol that a sender will use to make a payment to the recipient is as follows:
+1. Sender generates a random nonce key-pair \\( R = r \cdot G \\).
+2. Sender calculates a ECDH shared secret \\(c = H( r \cdot a \cdot G ) = H( a \cdot R) = H( r \cdot A) \\), where
+\\( H( \cdot ) \\) is a cryptographic hash function.
+3. The sender will then use \\( K_s = c \cdot G + B \\) as the last public key in the [one-sided payment] script. 
+4. The sender includes  \\( R \\) for the receiver but dropping it as it is not required during script execution. 
+This changes the script for a [one-sided payment] from `PushPubkey(K_s)` to `PushPubkey(R) Drop PushPubkey(K_s)`.
```
SWvheerden, line 91:
```suggestion
This changes the script for a [one-sided payment] from `PushPubkey(K_S)` to `PushPubkey(R) Drop PushPubkey(K_S)`.
```

RFC/src/RFC-0203_StealthAddresses.md
```diff
@@ -0,0 +1,114 @@
...
+
+The recipient will need to scan the blockchain for outputs that contain scripts of the [one-sided payment] form, and when
+one is found they will need to do the following:
+1. Extract the nonce \\( R \\) from the script.
+2. Use the public nonce to calculate the shared secret \\(c = H( a \cdot R) \\)
+3. Calculate \\( K_s \\) and check if it exists in the script.
```
SWvheerden, line 97:
```suggestion
3. Calculate \\( K_S \\) and check if it exists in the script.
```

PR #3695, opened 2022-01-11
fix!: fix tx and txo validation callback bug in wallet FFI

philipr-za:
Description
---

A recent change to the transaction service events meant that the Transaction Validation callback was only being called when state changed and not on success like it used to.
It was also noted that the callbacks no longer reported when a validation failed.

This PR updates the Tx and TXO validation callbacks to be called on completion of the respective validation protocols and also report if it was a success or not. Due to the simplification of the validation protocols the callbacks were simplified so that the feedback is a boolean denoting success or failure.

How Has This Been Tested?
---
unit tests updated


stringhandler (dismissed):

SWvheerden (dismissed):

delta1:
Is that FFI integration test flaky? Maybe try rerun it

philipr-za:
> Is that FFI integration test flaky? Maybe try rerun it

I think its an issue in the Weatherwax branch. I realized I should retarget this PR to the dev branch anyway

stringhandler (approved):

PR #3698, opened 2022-01-11
ci: improve cargo test step

stringhandler:


delta1:
Doesn't seem to be any gains from this unfortunately

stringhandler:
It was not super effective

PR #3699, opened 2022-01-12
chore: change UTXO priority to Higher number = higher priority

SWvheerden:
Description
---
This PR changes the UTXO priority in the wallet to be higher number = higher priority. 

Motivation and Context
---
After some discussion in the team, it was decided that higher number = higher priority is preferred above lower number = higher priority and that default spend priority should be 0. 

How Has This Been Tested?
---
Adds a unit test to prove that the spending priority selects the correct UTXO.


delta1:
title lint wants you to remove the full stop

stringhandler (approved):

PR #3703, opened 2022-01-13
feat: enable optional libtor for macos/linux

delta1:
Description
---

Continues from #3641 

Optionally use a built-in Tor process for macos/linux only


common/src/tor.rs
```diff
@@ -0,0 +1,180 @@
...
+            temp_dir: None,
+            temp_file: None,
+        }
+    }
+}
+
+impl Tor {
+    /// Returns a new Tor instance with random options.
+    /// The data directory, passphrase, and log destination are temporary and randomized.
+    /// Two available adjacent TCP ports will be selected in the given
```
sdbondi: Think this comment needs to be updated. RE my comment on your previous PR, Ah I realise that you cannot give tor the 0 port and have to bind yourself because you need to know the port that the OS chooses. Let's see if it ever becomes an issue with already being bound then unbound / OS gives the port to some other process. Seems like that would be rare at worst. Otherwise, I suppose a configured port is the way to go.

sdbondi (dismissed):
Looks great - not tested

common/src/tor.rs
delta1: Good catch, thanks will update the comment. You'll see a bit lower down I do actually bind to `0` and then use `local_addr` to retrieve the port number

common/config/presets/console_wallet.toml
```diff
@@ -253,4 +209,4 @@ tor_control_auth = "none" # or "password=xxxxxx"
 # direcly over TCP. /ip4, /ip6, /dns, /dns4 and /dns6 are supported.
 # tor_proxy_bypass_addresses = ["/dns4/my-foo-base-node/tcp/9998"]
 # When using the tor transport and set to true, outbound TCP connections bypass the tor proxy. Defaults to false for better privacy
-# tor_proxy_bypass_for_outbound_tcp = false;
+# tor_proxy_bypass_for_outbound_tcp = false
```
delta1:
```suggestion
# tor_proxy_bypass_for_outbound_tcp = false;
```

scripts/test_in_docker.sh
```diff
@@ -2,7 +2,7 @@
 
 # Run the Tari test suite locally inside a suitable docker container
 
-IMAGE=quay.io/tarilabs/rust_tari-build-with-deps:nightly-2020-08-13
+IMAGE=quay.io/tarilabs/rust_tari-build-with-deps:nightly-2022-01-03
```
delta1:
```suggestion
IMAGE=quay.io/tarilabs/rust_tari-build-with-deps:nightly-2020-08-13
```

stringhandler (approved):

common/src/tor.rs
sdbondi: Yeah, but you still have to bind on the port yourself, then release it - giving the tor control port a 0 will allocate a port, but you don't have an easy way to get which port was chosen unless you could ask tor through the control port, but that is a mission.I think that on most OS's it should unbind the port in time for tor to rebind on it, I seem to remember having this problem long ago but maybe things have changed / windows problem 🤷  - I'm sure it'll be 💯

PR #3705, opened 2022-01-14
perf: reduce memory usage in service framework

sdbondi:
Description
---

- replace unbounded channel with single element bounded channel

Motivation and Context
---
The reply channel sends a single message and waits for a response. This means that buffering up many requests
in an unbounded channel is pointless as the waiting occurs in any case (whether sending on a full channel or waiting for a reply). No performance change observed, however, this will prevent any extra memory allocations esp in stress test conditions where many internal messages are buffered. It is unclear when, if ever, unbounded channels release that memory.

How Has This Been Tested?
---
Existing tests pass
Manually

### DIFF
--- a/base_layer/core/src/base_node/chain_metadata_service/service.rs
+++ b/base_layer/core/src/base_node/chain_metadata_service/service.rs
@@ -298,8 +298,8 @@ mod test {
         LocalNodeCommsInterface,
         reply_channel::TryReceiver<NodeCommsRequest, NodeCommsResponse, CommsInterfaceError>,
     ) {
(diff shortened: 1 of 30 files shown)


stringhandler (approved):

sdbondi:
Thinking about it more, the `Box::pin` created for each call will allocate the Future on the heap for each waiting call - An Unbounded channel handles memory much much better than that. And after running this for a while and with a stress test, a memory increase is observed. Closing this. There are better (but trickier) ways to optimise memory usage on this.

PR #3711, opened 2022-01-17
fix: update Weatherwax cucumber FFI tests to pass

philipr-za:
Description
---

This PR updates the CI profile of the FFI cucumber tests and add @critical tags to be the same as those in the Development branch so that they pass CI.


delta1 (approved):
lgtm

PR #3713, opened 2022-01-18
fix!: don't include duplicate unique ID in same block template

sdbondi:
Description
---

- Keeps track of unique IDs in the current unconfirmed transaction set
- Mempool will not select unique IDs for block templates
- Add basic mempool unit test
- Implement consensus encoding for remaining output features fields/structs (consensus breaking)
- Update cucumber JS encoding and allow unique id to be set for transactions
- Add cucumber test 
- Adds new unit tests for output features consensus encoding

Motivation and Context
---
A duplicate unique ID in the same block is invalid and will be rejected by the network, since nothing has changed in the mempool after the rejected block the same transactions could be selected for the next block and so on. For this reason, the base node should not place transactions with duplicate unique IDs in the block template.

How Has This Been Tested?
---
New unit and cucumber test


sdbondi:
broken cucumber fixed in #3716

SWvheerden (approved):
LGTM

stringhandler (approved):

PR #3717, opened 2022-01-19
fix: put libtor behind optional feature flag

delta1:
Description
---

Make libtor fully optional behind `libtor` feature flag since we are hitting a compilation issue on macos. 

magicalbitcoin/libtor-sys#33

### DIFF
--- a/common/config/presets/console_wallet.toml
+++ b/common/config/presets/console_wallet.toml
@@ -122,6 +122,7 @@ base_node_service_request_max_age = 180
 transport = "tor"
 
 # Spin up and use a built-in Tor instance. This only works on macos/linux and you must comment out tor_control_address below.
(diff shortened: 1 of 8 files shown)


SWvheerden (dismissed):
Tested, gets around the build issue with libtor

common/config/presets/console_wallet.toml
```diff
@@ -122,6 +122,7 @@ base_node_service_request_max_age = 180
 transport = "tor"
 
 # Spin up and use a built-in Tor instance. This only works on macos/linux and you must comment out tor_control_address below.
+# This requires that the base node was built with the optional "libtor" feature flag.
```
delta1:
```suggestion
# This requires that the wallet was built with the optional "libtor" feature flag.
```

stringhandler (approved):

PR #3719, opened 2022-01-20
chore: add latency to status line when syncing

Cifko:
Description
---
Add `with latency {:?}` to status line during header/block sync.

How Has This Been Tested?
---
Manually.


base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -273,7 +273,14 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
                 .commit()
                 .await?;
 
-            self.hooks.call_on_progress_block_hooks(block.clone(), tip_height, peer);
+            self.hooks.call_on_progress_block_hooks(
+                block.clone(),
+                tip_height,
+                peer,
+                client
+                    .get_last_request_latency()
```
sdbondi: Use the Duration type

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
```diff
@@ -345,7 +371,7 @@ impl Display for HorizonSyncInfo {
 #[derive(Clone, Debug, PartialEq)]
 pub enum HorizonSyncStatus {
     Starting,
-    Kernels(u64, u64, NodeId),
-    Outputs(u64, u64, NodeId),
+    Kernels(u64, u64, NodeId, Option<u32>),
+    Outputs(u64, u64, NodeId, Option<u32>),
```
sdbondi: Use the Duration type

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -284,6 +284,9 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                     split_info.local_tip_header.height(),
                     split_info.remote_tip_height,
                     sync_peer.node_id(),
+                    client
+                        .get_last_request_latency()
```
sdbondi: As discussed, `get_last_request_latency` only measures the last RTT which is not available during streaming, to measure time between incoming header/blocks/etc you'll need to use an `Instant` and `Instant::elapsed` to measure that yourself.

Cifko: We can change this after you latency PR to use the same metric.

stringhandler (approved):

PR #3724, opened 2022-01-20
fix(collectibles): fix setup and unlock

stringhandler:
Fix the user experience when running collectibles for the first time. 
* Hide sidebar during lock screen
* Remove Wallets from menu because they did not have a screen to view anything


delta1 (dismissed):
lgtm untested

delta1 (approved):

PR #3725, opened 2022-01-20
fix(comms): ensure signature challenge is constructed consistently

sdbondi:
Description
---
- Ensure timestamp is timezone independent
- Ensure multiple addresses are lexicographically ordered for the signature challenge
- Add extra info in peer validation errors 
- migration that clears out all signatures (allowing new ones to take their place if provided)

Motivation and Context
---
Some base nodes are being banned for propagating invalid peer signatures - it is still not clear how this happens but
since peer signatures are checked before being committed it must be an inconsistency between base nodes in how the challenge is being constructed. Two possible but unconfirmed issues are that the integer timestamp from the naive date-time is effected by timezones (this PR replaces with `DateTime<Utc>`) and/or the addresses were not ordered consistently (doubtful because all peers currently only ever advertise one address, this PR sorts them lexicographically)  

How Has This Been Tested?
---

Manually: ban still occurs, but it seems as if only the same few nodes somehow accepted an invalid signature, this PR will clear the invalid signatures and we can monitor from there.

### DIFF
--- a/comms/dht/src/network_discovery/error.rs
+++ b/comms/dht/src/network_discovery/error.rs
@@ -34,6 +34,6 @@ pub enum NetworkDiscoveryError {
     ConnectivityError(#[from] ConnectivityError),
     #[error("No sync peers available")]
     NoSyncPeers,
-    #[error("Sync peer sent invalid peer")]
+    #[error("Sync peer sent invalid peer: {0}")]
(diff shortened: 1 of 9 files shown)


delta1 (approved):
lgtm, untested

PR #3727, opened 2022-01-21
feat!: add recovery byte to output features

hansieodendaal:
Description
---
Added `recovery_byte` to `OutputFeatures`. The recovery byte will uniformly distribute all outputs in the blockchain into 255 "bins" that can be identified by the commitment of the output in combination with a master recovery byte key.

This change will come into effect on `dibbler` testnet once the `OutputFeatures` version is updated from `V0` to `V1`, which will result in a hard fork.

All funds will be retained after the hard fork.

Motivation and Context
---
The reasoning behind this is that it will enable a client to request only a specific subset of full-on outputs in a set of blocks. The commitment only takes up a fraction (less than 5%) of the byte space for a full-on output. A typical use case for this feature would be a wallet recovery client that needs to scan all outputs in the blockchain from the birthdate of the wallet to be recovered. The client would request all commitments in the set of blocks from a base node and identify all those that belong to their own "bin", after which a tailor made request can be submitted to retrieve the full-on output data for the identified commitments. This will result in a better than 20x saving in the amount of data transmitted for wallet recovery.

How Has This Been Tested?
---
- Unit tests
- Cucumber integration tests
- Performed system-level tests using a wallet that contains embedded output features in json text that does not contain the recovery byte.


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -142,8 +142,13 @@ impl ConsensusConstants {
 
     pub fn coinbase_weight(&self) -> u64 {
         // TODO: We do not know what script, features etc a coinbase has - this should be max coinbase size?
+        let output_features_estimate = OutputFeatures {
+            // This is a best guess to not underestimate metadata size
+            recovery_byte: u8::MAX,
```
philipr-za: This doesn't seem necessary. Whether the byte has a value of 0 or 255 it is still a byte in length.

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -111,18 +104,24 @@ impl SingleReceiverTransactionProtocol {
                 .construct_proof(spending_key, sender_info.amount.into())?
         };
 
+        let mut sender_features = sender_info.features.clone();
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, rewind_data);
+        if sender_features.recovery_byte != recovery_byte {
```
philipr-za: Nit: this will probably never be true right? If it is, have we saved any work here by doing the compare vs just setting the recovery byte no matter what?

base_layer/tari_stratum_ffi/src/lib.rs
```diff
@@ -280,41 +280,84 @@ mod tests {
...
+    // run with --nocapture flag to see output
+    fn get_hash() {
+        let block_hex = hex::decode(BLOCK_HEX).unwrap();
+        let mut block: Block = serde_json::from_str(&String::from_utf8_lossy(&block_hex).to_string()).unwrap();
+        block.header.nonce = NONCE;
+        let hash_hex = block.header.hash().to_hex();
+        println!("{}", hash_hex);
+    }
+
+    #[test]
```
philipr-za: Does this need a `#[test]` if it just being ignored?

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -295,6 +315,24 @@ impl OutputManagerHandle {
         }
     }
 
+    pub async fn add_rewindable_output(
```
philipr-za, line 342: Are `add_rewindable_output`,  `add_rewindable_output_with_tx_id`, `add_rewindable_unvalidated_output` only used for testing? I am assuming that they are because how would a normal client of this service be able to construct its own rewindable output externally.

So based on that I would say these three methods don't actually belong in API of this service. Now the question is how do you get these kind of outputs into the output manager for testing. The answer is that in the testing code you have access to the OMS database backend and you can insert them directly.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -335,6 +393,26 @@ impl OutputManagerHandle {
         }
     }
 
+    pub async fn add_rewindable_unvalidated_output(
```
philipr-za: I see this method is not even used in tests, I think it should come out

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1488,8 +1623,14 @@ where
                 .master_key_manager
                 .get_next_spend_and_script_key()
                 .await?;
-            let sender_offset_private_key = PrivateKey::random(&mut OsRng);
+            // TODO: Hansie - supply the correct recovery_byte here
```
philipr-za: Perhaps just clarify why this isn't the correct one?

philipr-za (commented):
Generally looking really good, couple comments but the main one is that the three methods added to the OMS API for adding a rewindable output seem like they don't belong outside of the tests.

Please keep further updates to this PR as their own commits so I can review the new stuff only.

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: It makes a difference for consensus encoding where 0 has a different size than 255.

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: I would counter that their counterparts, that is without the `rewindable` part, also need to be taken out of the API then. This is expanding on the pattern that is already there.

Your opinion?

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: Yeah, still busy with that one. That functionality needs to be tested though.

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: That was a note to self and I have already implemented the code, just need to remove the comment. Thanks

base_layer/tari_stratum_ffi/src/lib.rs
hansieodendaal: Good catch, those are just utilities to set up new fixed test data.

base_layer/wallet/src/output_manager_service/handle.rs
philipr-za: The counter parts are used for importing of outputs manually and by the UTXO scanner.

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
hansieodendaal: Let me simplify this throughout.

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: ~~As discussed, I will remove these from the API.~~

Maybe for a next PR.

base_layer/core/src/consensus/consensus_constants.rs
delta1: but it's encoded as a varint? cc @sdbondi

base_layer/core/src/consensus/consensus_encoding/integers.rs
```diff
@@ -47,6 +47,7 @@ macro_rules! consensus_encoding_varint_impl {
     };
 }
 
+consensus_encoding_varint_impl!(u8);
```
philipr-za: So this seems wrong for a Byte. Varint is more efficient for integers larger than a byte but for a byte half the time the value is actually 2 bytes.

base_layer/core/src/consensus/consensus_constants.rs
philipr-za: and for a byte value it shouldn't be a varint?

sdbondi: I doubt the recovery byte won't need to grow past 1 byte, so I suggest consensus encoding just writes the single byte (rather than varint) which means it will always be a fixed size regardless of the value.

base_layer/core/src/consensus/consensus_encoding/integers.rs
sdbondi: Agree with @philipr-za

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: Changed recovery byte encoding to always be encoded to 1 byte

base_layer/core/src/consensus/consensus_encoding/integers.rs
hansieodendaal: Yip, changed it

applications/tari_app_grpc/src/conversions/output_features.rs
```diff
@@ -62,6 +62,7 @@ impl TryFrom<grpc::OutputFeatures> for OutputFeatures {
             OutputFlags::from_bits(features.flags as u8)
                 .ok_or_else(|| "Invalid or unrecognised output flags".to_string())?,
             features.maturity,
+            features.recovery_byte as u8,
```
delta1: This will overflow if its larger than a u8 which might give unintended results? Probably best to use `try_from` in the same way as the version field above

base_layer/core/src/proto/transaction.rs
```diff
@@ -290,6 +290,7 @@ impl TryFrom<proto::types::OutputFeatures> for OutputFeatures {
             OutputFlags::from_bits(features.flags as u8)
                 .ok_or_else(|| "Invalid or unrecognised output flags".to_string())?,
             features.maturity,
+            features.recovery_byte as u8,
```
delta1: I think this should probably use `try_from` as well

base_layer/core/src/chain_storage/tests/blockchain_database.rs
```diff
@@ -779,6 +783,7 @@ mod fetch_utxo_by_unique_id {
             lock: 0,
             features: features
         )]);
+        features.set_recovery_byte(tx_outputs[0].features.recovery_byte);
```
delta1, line 786: What happens if someone creates an `OutputFeatures` and forgets to call `set_recovery_byte` ?

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -115,27 +132,87 @@ impl OutputFeatures {
...
+    /// with the value '0' never obtainable as it is being reserved as a default value.
+    pub fn create_unique_recovery_byte(commitment: &PedersenCommitment, rewind_data: Option<&RewindData>) -> u8 {
+        let commitment_bytes = commitment.as_bytes();
+        let recovery_key_bytes = if let Some(data) = rewind_data {
+            data.recovery_byte_key.as_bytes()
+        } else {
+            &[]
+        };
+        const RECOVERY_BYTE_SIZE: usize = 1;
+        let blake2_hasher = VarBlake2b::new(RECOVERY_BYTE_SIZE).expect("Should be able to create blake2 hasher");
```
delta1: Please add a comment about the potential panic 

> This can only panic if `VarBlake2b::new` is given an output size that is 0 or greater than 64

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -240,13 +317,31 @@ impl OutputFeatures {
     pub fn is_coinbase(&self) -> bool {
         self.flags.contains(OutputFlags::COINBASE_OUTPUT)
     }
+
+    fn consensus_encode_recovery_byte<W: Write>(recovery_byte: u8, writer: &mut W) -> Result<usize, io::Error> {
+        writer.write_all(&[recovery_byte][..])?;
```
delta1: What is `[..]` here?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -2011,7 +2015,9 @@ impl TryFrom<CompletedTransactionSql> for CompletedTransaction {
                 .map_err(TransactionKeyError::Destination)?,
             amount: MicroTari::from(c.amount as u64),
             fee: MicroTari::from(c.fee as u64),
-            transaction: serde_json::from_str(&c.transaction_protocol)?,
+            transaction: serde_json::from_str(&OutputFeatures::add_recovery_byte_to_serialized_data_if_needed(
```
hansieodendaal: ditto

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1485,7 +1485,9 @@ impl TryFrom<InboundTransactionSql> for InboundTransaction {
             tx_id: (i.tx_id as u64).into(),
             source_public_key: PublicKey::from_vec(&i.source_public_key).map_err(TransactionKeyError::Source)?,
             amount: MicroTari::from(i.amount as u64),
-            receiver_protocol: serde_json::from_str(&i.receiver_protocol)?,
+            receiver_protocol: serde_json::from_str(&OutputFeatures::add_recovery_byte_to_serialized_data_if_needed(
```
hansieodendaal: ditto

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -483,10 +484,12 @@ impl TryFrom<OutputSql> for DbUnblindedOutput {
     type Error = OutputManagerStorageError;
 
     fn try_from(o: OutputSql) -> Result<Self, Self::Error> {
-        let mut features: OutputFeatures =
-            serde_json::from_str(o.features_json.as_str()).map_err(|s| OutputManagerStorageError::ConversionError {
-                reason: format!("Could not convert json into OutputFeatures:{}", s),
-            })?;
+        let mut features: OutputFeatures = serde_json::from_str(
+            &OutputFeatures::add_recovery_byte_to_serialized_data_if_needed(o.features_json.clone()),
```
hansieodendaal: This should only be done at wallet startup, not every time transaction data is read from the db.

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1658,7 +1660,9 @@ impl TryFrom<OutboundTransactionSql> for OutboundTransaction {
                 .map_err(TransactionKeyError::Destination)?,
             amount: MicroTari::from(o.amount as u64),
             fee: MicroTari::from(o.fee as u64),
-            sender_protocol: serde_json::from_str(&o.sender_protocol)?,
+            sender_protocol: serde_json::from_str(&OutputFeatures::add_recovery_byte_to_serialized_data_if_needed(
```
hansieodendaal: ditto

hansieodendaal (commented):
**Changes requested:**
 The json text embedded in the wallet should only be done at wallet startup, not every time output or transaction data is read from the db.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -207,9 +224,11 @@ impl fmt::Display for OutputManagerRequest {
 
 /// API Reply enum
 #[derive(Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
```
delta1: Rather `Box` the TransactionOutput instead of allowing this clippy

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -583,7 +602,13 @@ where
...
     }
 
-    /// Add an unblinded output to the outputs table and marks is as `Unspent`.
+    pub fn calculate_recovery_byte(&mut self, spending_key: PrivateKey, value: u64) -> Result<u8, OutputManagerError> {
+        let commitment = self.resources.factories.commitment.commit_value(&spending_key, value);
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, Some(&self.resources.rewind_data));
+        Ok(recovery_byte)
+    }
+
+    /// Add an unblinded non-rewindable output to the outputs table and marks is as `Unspent`.
```
delta1:
```suggestion
    /// Add an unblinded non-rewindable output to the outputs table and mark it as `Unspent`.
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5244,7 +5244,24 @@ pub unsafe extern "C" fn wallet_import_utxo(
     };
 
     let public_script_key = PublicKey::from_secret_key(&(*spending_key));
-    // Todo the script_lock_height can be something other than 0, for example an HTLC transaction
+
+    let recovery_byte;
+    match (*wallet).runtime.block_on(
```
delta1:
```suggestion
    let recovery_byte = match (*wallet).runtime.block_on(
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5244,7 +5244,24 @@ pub unsafe extern "C" fn wallet_import_utxo(
...
+        (*wallet)
+            .wallet
+            .output_manager_service
+            .calculate_recovery_byte((*spending_key).clone(), amount),
+    ) {
+        Ok(v) => recovery_byte = v,
```
delta1:
```suggestion
        Ok(v) => v,
```

delta1 (commented):
Well done this was a lot of work. I have a few comments and questions, and my only real concern is that a developer must remember to update the recovery byte when creating OutputFeatures. I would vastly prefer it to be automatically updated whenever the commitment of the output is set or updated.

base_layer/core/src/chain_storage/tests/blockchain_database.rs
hansieodendaal: A typical flow to create an output is (only discussing the pertinent information):
- Select output features using a default/random recovery byte
- Calculate output weight & determine the fee
- Finalize the commitment information (i.e. select the blinding factor)
- Set the recovery byte with the commitment as input (i.e. finalize the output features)
- Calculate the partial or final metadata signature, also signing the output features
- Finalize the output as an unblinded output, using all previous information as inputs

Recovery byte checks is done in strategic places in the code base and automatically set whenever it needs to be updated within the various protocols. If someone adds functionality and forget to set the recovery byte as required for the protocol, recovery byte validation will not succeed the validation in `unblinded_output::as_transaction_output`,  `unblinded_output::as_rewindable_transaction_output` and `transaction_initializer::with_output`.

base_layer/core/src/transactions/transaction_components/output_features.rs
hansieodendaal: Yip, not needed

hansieodendaal:
>  I would vastly prefer it to be automatically updated whenever the commitment of the output is set or updated.

Not really possible to do this, please see detailed comment above.

base_layer/core/src/test_helpers/mod.rs
```diff
@@ -110,13 +110,13 @@ pub fn mine_to_difficulty(mut block: Block, difficulty: Difficulty) -> Result<Bl
     // When starting from the same nonce, in tests it becomes common to mine the same block more than once without the
     // hash changing. This introduces the required entropy
     block.header.nonce = rand::thread_rng().gen();
-    for _i in 0..10000 {
+    for _i in 0..20000 {
```
stringhandler, line 113: What is the reason for this? Even 10k is a really high value and makes tests run slowly

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -115,27 +132,88 @@ impl OutputFeatures {
...
+        let mut salt = 0u64;
+        loop {
+            blake2_hasher
+                .clone()
+                .chain(commitment_bytes)
+                .chain(recovery_key_bytes)
+                .chain(b"hash my recovery byte")
+                .chain(salt.to_le_bytes().as_slice())
+                .finalize_variable(|res| hash.copy_from_slice(res));
+            if hash[0] == OutputFeatures::RECOVERY_BYTE_DEFAULT {
```
stringhandler: This is not necessary, we can just allow 0's

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -115,27 +132,88 @@ impl OutputFeatures {
...
-    pub fn custom(flags: OutputFlags, metadata: Vec<u8>) -> OutputFeatures {
-        Self {
-            flags,
-            metadata,
-            ..Default::default()
+    /// Helper function to update the unique recovery byte if required and return updated output features
+    pub fn update_recovery_byte_if_required(
+        commitment: &PedersenCommitment,
+        rewind_data: Option<&RewindData>,
+        features: &OutputFeatures,
```
stringhandler, line 174: I would expect this signature to take a mut, since it implies that the intention was to change the features?
```suggestion
        features: &mut OutputFeatures,
```

A better signature would be `update_recovery_byte(&self, commitment: .., rewind_data:...) -> ()`

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -115,27 +132,88 @@ impl OutputFeatures {
...
+
+    /// Provides the ability to update the recovery byte after the commitment has become known
+    pub fn set_recovery_byte(&mut self, recovery_byte: u8) {
+        self.recovery_byte = recovery_byte;
+    }
+
+    /// Provides the ability to add the recovery byte to an older version of serialized json output features. We need
+    /// to add the non-optional 'recovery_byte' field to a persisted protocol that contains serialized json output
+    /// features, typically when reading it from the database, if it was created with 'OutputFeaturesVersion::V0'
+    pub fn add_recovery_byte_to_serialized_data_if_needed(protocol: String) -> String {
```
stringhandler, line 190: Please remove this. It will be very hard to maintain. 

Is it not better to just add a `serde(default = 0)` to the `recovery_byte`?

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -240,13 +318,31 @@ impl OutputFeatures {
     pub fn is_coinbase(&self) -> bool {
         self.flags.contains(OutputFlags::COINBASE_OUTPUT)
     }
+
+    fn consensus_encode_recovery_byte<W: Write>(recovery_byte: u8, writer: &mut W) -> Result<usize, io::Error> {
```
stringhandler, line 316: is this needed? Isn't u8 already consensus encodable?

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -240,13 +318,31 @@ impl OutputFeatures {
...
 
 impl ConsensusEncoding for OutputFeatures {
     fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, io::Error> {
         let mut written = self.version.consensus_encode(writer)?;
         written += self.maturity.consensus_encode(writer)?;
         written += self.flags.consensus_encode(writer)?;
+        match self.version {
+            OutputFeaturesVersion::V0 => (),
+            OutputFeaturesVersion::V1 => {
+                written += OutputFeatures::consensus_encode_recovery_byte(self.recovery_byte, writer)?;
```
stringhandler, line 337:
```suggestion
                written += u8::consensus_encode(self.recovery_byte, writer)?;
```

base_layer/core/src/transactions/transaction_components/transaction.rs
```diff
@@ -123,7 +123,10 @@ impl Transaction {
                 min_maturity,
                 input
                     .features()
-                    .unwrap_or(&OutputFeatures::with_maturity(std::u64::MAX))
+                    .unwrap_or(&OutputFeatures {
```
stringhandler, line 126: The previous code was better. Why not add the default to `with_maturity()`

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -196,6 +204,15 @@ impl UnblindedOutput {
             ));
         }
         let commitment = factories.commitment.commit(&self.spending_key, &self.value.into());
+
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, None);
+        if self.features.recovery_byte != recovery_byte {
```
stringhandler, line 209: This seems out of place. We are not checking any other values here. Please remove it, since it should either be somewhere else, or trust the data it has already

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -243,6 +260,14 @@ impl UnblindedOutput {
                 .map_err(|_| TransactionError::RangeProofError(RangeProofError::ProofConstructionError))?
         };
 
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, Some(rewind_data));
+        if self.features.recovery_byte != recovery_byte {
```
stringhandler, line 264: Same for here

base_layer/core/src/transactions/transaction_components/unblinded_output_builder.rs
```diff
@@ -63,6 +68,19 @@ impl UnblindedOutputBuilder {
         }
     }
 
+    pub fn update_recovery_byte_if_required(
```
stringhandler, line 71: this is at the builder stage, so no need to be cautious

```suggestion
    pub fn update_recovery_byte(
```

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -104,7 +104,8 @@ impl SingleReceiverTransactionProtocol {
                 .construct_proof(spending_key, sender_info.amount.into())?
         };
 
-        let sender_features = sender_info.features.clone();
+        let sender_features =
```
stringhandler: Why would the recovery byte not be correct here already?

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -214,6 +214,13 @@ impl SenderTransactionInitializer {
     ) -> Result<&mut Self, BuildError> {
         let commitment_factory = PedersenCommitmentFactory::default();
         let commitment = commitment_factory.commit(&output.spending_key, &PrivateKey::from(output.value));
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, self.rewind_data.as_ref());
+        if recovery_byte != output.features.recovery_byte {
+            self.clone().build_err(&*format!(
```
stringhandler: This clone feels wrong. Wouldn't this error get lost?

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -214,6 +214,13 @@ impl SenderTransactionInitializer {
     ) -> Result<&mut Self, BuildError> {
         let commitment_factory = PedersenCommitmentFactory::default();
         let commitment = commitment_factory.commit(&output.spending_key, &PrivateKey::from(output.value));
+        let recovery_byte = OutputFeatures::create_unique_recovery_byte(&commitment, self.rewind_data.as_ref());
```
stringhandler, line 217: Feels like this work has already been done. Why do we not add the commitment into the unblindedoutput and skip recreating it here?

base_layer/core/src/validation/helpers.rs
```diff
@@ -859,7 +859,10 @@ mod test {
         fn it_checks_the_input_maturity() {
             let input = TransactionInput::new_with_output_data(
                 TransactionInputVersion::get_current_version(),
-                OutputFeatures::with_maturity(5),
+                OutputFeatures {
```
stringhandler, line 862: See previous comments about this

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -106,6 +111,10 @@ pub enum OutputManagerRequest {
     ApplyEncryption(Box<Aes256Gcm>),
     RemoveEncryption,
     GetPublicRewindKeys,
+    CalculateRecoveryByte {
```
stringhandler, line 115: This seems strange to expose at this level. What is the use case for someone to only call this method without an output?

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -62,8 +63,12 @@ use crate::output_manager_service::{
 pub enum OutputManagerRequest {
     GetBalance,
     AddOutput((Box<UnblindedOutput>, Option<SpendingPriority>)),
+    // ToDo: This API request could probably be removed by expanding test utils if only needed for testing
+    AddRewindableOutput((Box<UnblindedOutput>, Option<SpendingPriority>, Option<RewindData>)),
```
stringhandler, line 67: How does a developer know to use this and not `AddRewindableOutputWithTxId`?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -694,8 +731,14 @@ where
         let input_data = inputs!(PublicKey::from_secret_key(&script_private_key));
         let script = script!(Nop);
 
+        let recovery_byte = self.calculate_recovery_byte(spending_key.clone(), value.as_u64())?;
+        let mut updated_features = features;
+        if updated_features.recovery_byte != recovery_byte {
```
stringhandler: This if statement is unnecessary

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -731,11 +774,17 @@ where
 
         let (spending_key, script_private_key) = self.get_spend_and_script_keys().await?;
 
+        let mut updated_features = single_round_sender_data.features.clone();
```
stringhandler: Is this correct? I don't think the sender's features should be updated

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -797,14 +846,13 @@ where
...
+        let output_features = OutputFeatures { ..Default::default() };
         let metadata_byte_size = self
             .resources
             .consensus_constants
             .transaction_weight()
             .round_up_metadata_size(
-                OutputFeatures::default().consensus_encode_exact_size() +
-                    script![Nop].consensus_encode_exact_size() +
-                    Covenant::new().consensus_encode_exact_size(),
+                output_features.consensus_encode_exact_size() + script![Nop].consensus_encode_exact_size(),
```
stringhandler: Are you sure the covenant size is 0?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -848,13 +896,13 @@ where
             unique_id,
             fee_per_gram,
         );
-        let output_features = OutputFeatures::default();
+        let output_features_estimate = OutputFeatures { ..Default::default() };
```
stringhandler: This is the same thing

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -73,12 +73,55 @@ pub struct OutputManagerSqliteDatabase {
 
 impl OutputManagerSqliteDatabase {
     pub fn new(database_connection: WalletDbConnection, cipher: Option<Aes256Gcm>) -> Self {
+        // TODO: Remove this migration for the next testnet reset (currently on Dibbler) #testnet_reset
+        let _ = OutputManagerSqliteDatabase::update_json_output_features_to_v1(database_connection.clone());
```
stringhandler: What about the outputs that are not v1?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -73,12 +73,55 @@ pub struct OutputManagerSqliteDatabase {
...
+
         Self {
             database_connection,
             cipher: Arc::new(RwLock::new(cipher)),
         }
     }
 
+    // TODO: Remove this migration for the next testnet reset (currently on Dibbler) #testnet_reset
+    fn update_json_output_features_to_v1(
```
stringhandler: This should not be done IMO. Newly created outputs should get the V1 output features only if they are created with output v1 or higher

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -83,10 +88,131 @@ pub struct TransactionServiceSqliteDatabase {
...
 impl TransactionServiceSqliteDatabase {
     pub fn new(database_connection: WalletDbConnection, cipher: Option<Aes256Gcm>) -> Self {
-        Self {
+        let res = Self {
             database_connection,
             cipher: Arc::new(RwLock::new(cipher)),
+        };
+
+        // TODO: Remove this migration for the next testnet reset (currently on Dibbler) #testnet_reset
+        let _ = TransactionServiceSqliteDatabase::update_json_output_features_to_v1(res.clone());
```
stringhandler: See previous comment

stringhandler (commented):
I would suggest trying the `#[serde(default= 0)]` and not changing any of the existing json in the faucet or sqlite db.

I'll try test this locally as well

base_layer/core/src/test_helpers/mod.rs
hansieodendaal: Tests were failing due to required difficulty not being reached in 10,000 iterations - the function will exit whenever the difficulty is reached.

base_layer/core/src/transactions/transaction_components/output_features.rs
hansieodendaal: If we allow `0` as a valid calculated `recovery_byte` the binning will be skewed when `V1` is introduced. The thinking here is that when a wallet request UTXOs to scan for recovery it should (a) retrieve all metadata for UTXOs with `recovery_byte != 0` (b) compile a list of potential outputs to scan based on the calculated `recovery_byte` (c) request full UTXO data for all UTXOs with `recovery_byte != 0` (to cover the `V0` outputs) as well as those with corresponding `recovery_byte` values.

This also allows us to have a valid default value that is not possible to attain when the `recovery_byte` is calculated. This is especially required for consensus encoding/decoding as `recovery_byte` cannot be part of the blockchain in `V0`  but we must be able to destructure encoded output features into `OutputFeatures`.

base_layer/core/src/test_helpers/mod.rs
stringhandler: Ok, let's keep an eye on it

base_layer/core/src/transactions/transaction_components/output_features.rs
stringhandler: Sounds like that will only be a problem for the current testnet. I vote for removing the code and allowing 0's. Any wallets that need to check for 0 before recovering can use the check `features.version >=1`

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: These are the features suggested by the sender for the receiver's output, then the receiver has to update the recovery byte.

base_layer/core/src/transactions/transaction_components/output_features.rs
hansieodendaal: Removing the reserved `RECOVERY_BYTE_DEFAULT` as discussed

base_layer/core/src/transactions/transaction_components/output_features.rs
hansieodendaal: Yip, we needed this. Also, see consensus encode and decoding for `OutputFeaturesVersion`.

base_layer/core/src/transactions/transaction_components/output_features.rs
hansieodendaal: As above

base_layer/core/src/transactions/transaction_components/transaction.rs
hansieodendaal: The reason is that using `with_maturity` gives one a sense that it is done and actually it is not, as `recovery_byte` is still missing.

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
hansieodendaal: This is one of three places where we validate that `recovery_byte` is calculated correctly; without this, it is possible to create a UTXO with the wrong recovery byte. If we allow that a wallet recovery may miss many funds thinking it is deep scanning all probable UTXOs.

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
hansieodendaal: As above

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
hansieodendaal: The spending key was not known before this point; this is the receiver building an output with the suggested features from the sender.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: Without the clone: `cannot move out of '*self' which is behind a mutable reference`

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: I think it has to do with the commitment factory that is needed to calculate the commitment, which will make creating an unblinded output a heavier operation.

In this example `commitment ` is needed for the metadata signature further on; it is just re-used here for the recovery byte calculation.

base_layer/core/src/validation/helpers.rs
hansieodendaal: As above

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: This was needed for the wallet FFI method `pub unsafe extern "C" fn wallet_import_utxo`as it could accept a null pointer for OutputFeatures and then assign default features, in which case the recovery byte had to be set correctly to pass the recovery byte validation further down the line. Coming to think of it, all input data should be coherent, so changing the FFI method to return an error in case of null pointer for OutputFeatures.

hansieodendaal: This is now only used for unit tests with a `ToDo` comment.

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: `AddRewindableOutput` is currently used in unit tests only, for example as in `async fn test_utxo_selection_no_chain_metadata()`.

`AddRewindableOutputWithTxId` is used in `pub async fn send_sha_atomic_swap_transaction` where `TxId` is chosen at the start of the protocol.

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: Yip, this is an error, reverting to the previous code.

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: The wording choice of `output_features_estimate` is important here to let the reader know at this stage these features are not the real features and only used to for fee estimation.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
hansieodendaal: The word choice `update_json_output_features_to_v1` is bad here, it should actually read `update_json_output_features_to_include_default_recovery_byte`

hansieodendaal: Going to replace this with

> Is it not better to just add a serde(default = 0) to the recovery_byte?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
hansieodendaal: Going to replace this with

> Is it not better to just add a serde(default = 0) to the recovery_byte?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
hansieodendaal: Going to replace this with

> Is it not better to just add a serde(default = 0) to the recovery_byte?

base_layer/wallet/src/output_manager_service/service.rs
sdbondi: Think he means `OutputFeatures { ..Default::default() };` is a convoluted way of writing the previous code `OutputFeatures::default()`

hansieodendaal: Yip, that was a refactoring remnant.

stringhandler (approved):

PR #3730, opened 2022-01-21
fix(core)!: add missing consensus encoding length byte rangeproof & covenants

sdbondi:
Description
---

- Adds consensus encode impl for RangeProof and ComSig
- Adds length varint for covenant
- Move version as first encoded byte
- Implement zero alloc hashing for outputs and witness hash
- Limit covenant and tariscript byte size
- Migration to set length byte in existing wallet transactions
- Ensure that an RPC message can never be zero length.
- covenant serde implementation does not use consensus encoding

Motivation and Context
---
Rangeproof is not fixed size and so requires a length varint. 

It is important to ensure that a remote peer cannot cause unbounded allocation, so various length checks are implemented. 

The wallet previously only ever stored a 0-byte covenant which is now invalid encoding.

Potential fix for WriteZero, ensuring that all rpc messages are non-zero in length (rare, but possible).

How Has This Been Tested?
---
Existing tests updated, manually - mined blocks on dibbler


base_layer/core/src/transactions/transaction/transaction_input.rs
```diff
@@ -353,7 +345,6 @@ impl TransactionInput {
                 .chain(self.script_signature.public_nonce().to_consensus_bytes())
```
SWvheerden: you don't have to split the comsig here anymore. You have added an implementation to consensus encode the entire comsig.

base_layer/core/src/transactions/transaction/transaction_kernel.rs
```diff
@@ -126,13 +126,13 @@ impl Hashable for TransactionKernel {
     /// $$ H(feature_bits | fee | lock_height | P_excess | R_sum | s_sum)
     fn hash(&self) -> Vec<u8> {
```
SWvheerden, line 131: Should this not use consensus encoding as well?
This hash is committed to in the kernel mmr, and every single implementation should be able to calculate this.

integration_tests/features/Mempool.feature
```diff
@@ -38,7 +38,7 @@ Feature: Mempool
     Then SENDER has TX1 in MINED state
     Then TX1 is in the MINED of all nodes
 
-  @critical
+  @critical @doit
```
SWvheerden: debug flag left in?

SWvheerden (commented):
This looks good. Made a comment or two. 
We can remove all the places where we split the comsig's and then encode them as this pr now adds an implementation to implement them directly. 

I also think we should use consensus encoding everywhere in the hashes and challenges as this is committed to on the blockchain, either in a MMR root, or just straight up in the output/input. And if we want another implementation to be able to verify this it needs to be able to decode/encode it.

stringhandler (approved):

PR #3731, opened 2022-01-21
feat(validator): add grpc identity rpc

stringhandler:
Add a gRPC method for getting the validator node's identity


delta1:
We need to gitignore the contents of the collectibles web-app/build folder but also .gitkeep it to satisfy the tauri build

delta1 (approved):

PR #3735, opened 2022-01-24
fix: prevent key leaking through derive debug impl

sdbondi:
Description
---
- Impl custom Debug for `UnblindedOutput` to prevent inadvertent leaking of private keys  

Motivation and Context
---
Secret keys would previously have been written to logs if debug formatting was used 

How Has This Been Tested?
---


stringhandler (approved):
Nice

PR #3736, opened 2022-01-24
chore: add fee per gram option to coin-split command

Cifko:
Description
---
Add option to change fee per gram in `coin-split`. The parameter is optional (default is 5µT)

How Has This Been Tested?
---
npm test -- --name "As a user I want to coin-split via command line"


stringhandler (approved):
golden

PR #3738, opened 2022-01-24
ci: switch to cargo cross for libwallet android builds

delta1:
Description
---

Use cargo cross in github actions to cross compile the libwallet android builds for x86 and aarch64. 

This also statically compiles openssl using the vendored feature of the openssl crate.

These builds still need to be tested to ensure they function correctly..

If the 32bit armv7 build is still required this will need some additional changes but can be done subsequently. 

image

### DIFF
--- a/.github/workflows/libwallet.yml
+++ b/.github/workflows/libwallet.yml
@@ -21,19 +21,31 @@ jobs:
         with:
           toolchain: nightly-2021-11-20
           override: true
+      - name: Install cross
+        run: |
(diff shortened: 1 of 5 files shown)


stringhandler (dismissed):
Happy, just need to fix the clippy

delta1:
> Happy, just need to fix the clippy

lol that looks new

delta1:
https://github.com/rustwasm/wasm-bindgen/issues/2774#issuecomment-1019143467

PR #3741, opened 2022-01-25
feat(sync): blockchain sync switches peers when max latency is exceeded

sdbondi:
Description
---
- Monitors sync peer latency and switches peers if latency exceeds maximum
- Add blocks/headers/kernels/outputs per second to status line
- Decouple horizon sync impl state machine for testability
- Extends unit test for dibbler genesis block to check chain balance (currently incorrect)

Motivation and Context
---
If a slow sync peer is encountered, sync times are naturally greatly impacted, so a faster peer should be selected.

Applicable sync peers are given to header sync which sorts them by latency and attempts to sync based on that order.
If a sync peer exceeds the initial maximum (3 secs) the next sync peer is used. If all sync peers exceed that latency, the latency is increased by 2 seconds and reattempted in that same order, unless there are less than 3 sync peers, in which case it kicks back to listening state so that more sync peers can be selected. This applies to header, block and horizon sync. 

How Has This Been Tested?
---
Manually, cucumber


delta1:
Looks good, what happens if all peers have too high a latency?

sdbondi:
@delta1 Updated the PR description

delta1 (dismissed):

delta1:
> @delta1 Updated the PR description

Thanks! looks good, I just want to get another set of eyes on it too. @mikethetike @philipr-za ?

stringhandler (dismissed):
Happy, but I need to test it locally

PR #3746, opened 2022-01-25
feat: show error if IPFS upload fails

sdbondi:
Description
---
- show error for failed file upload

Motivation and Context
---
File used to appear as if it was uploading forever, now an error appears

How Has This Been Tested?
---
Manually
image


delta1 (approved):
Nice LGTM, untested

PR #3747, opened 2022-01-25
fix: stratum and mining_node config

SWvheerden:
Description
---
Stratum transcoder does not read GRPC addresses from the config file.  This PR fixes the stratum transcoder to read the setting file and use the global config.

The mining node reads the GRPC addresses from the wallet and base node settings, but their server addresses. But it is not necessarily the servers you want to connect to. This PR makes sure that the mining node reads its own config to ensure it uses the correct settings. 

How Has This Been Tested?
---
Manual testing of settings

### DIFF
--- a/common/src/configuration/global.rs
+++ b/common/src/configuration/global.rs
@@ -819,24 +819,22 @@ fn convert_node_config(
 
     if application == ApplicationType::MiningNode {
         let key = "mining_node.base_node_grpc_address";
-        node_grpc_server = Some(
-            cfg.get_str(key)
-                .map_err(|e| ConfigurationError::new(key, None, &e.to_string()))
-                .and_then(|addr| {
-                    addr.parse::<Multiaddr>()
-                        .map_err(|e| ConfigurationError::new(key, Some(addr), &e.to_string()))
-                })?,
-        );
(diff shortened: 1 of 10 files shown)


common/src/configuration/global.rs
```diff
@@ -794,6 +794,55 @@ fn convert_node_config(
...
+        let pie = cfg.get_str(key);
+        dbg!(&pie);
+        node_grpc_server = Some(
+            pie.map_err(|e| ConfigurationError::new(key, None, &e.to_string()))
+                .and_then(|addr| {
+                    addr.parse::<Multiaddr>()
+                        .map_err(|e| ConfigurationError::new(key, Some(addr), &e.to_string()))
+                })?,
+        );
+        dbg!("hello from walle");
```
delta1: lol

SWvheerden: haha there was more than one, I just did not push them.

philipr-za (dismissed):
Looks fine, but you need to format your js

stringhandler:
Waiting on #3777 that has bigger config changes to see if this is still relevant

SWvheerden:
This is replaced by PR #3945

PR #3749, opened 2022-01-26
fix(core): fetch_header_containing_*_mmr functions now take a 0-based mmr position

sdbondi:
Description
---

- fetch_header_containing_kernel_mmr  and fetch_header_containing_utxo_mmr take a position starting from 0 inline with the mmr counts in the headers
- Updated calls (remove the + 1)
- Updated tests

Motivation and Context
---
The caller expects the position to be 0-based (e.g `Iterator::position` returns an index to the found element), the + 1 added to calls were used to account for this bug in the lmdb blockchain backend impl

How Has This Been Tested?
---

Updated unit tests, manual horizon sync


delta1 (approved):
Nice one, definite improvement

stringhandler (approved):

PR #3750, opened 2022-01-26
feat(collectibles): add form validation error when committee not set

sdbondi:
Description
---
- Add alert on save click when no committee members were added
- Adds eslint rules (rules chosen to limit lines of code changed, not based on any preference I have, defaults are probably best)

Motivation and Context
---
Dont submit the form unless at least one committee member is chosen

How Has This Been Tested?
---
Manually


delta1 (approved):
LGTM

PR #3751, opened 2022-01-26
fix: use of branch seed in key manager

Cifko:
Description
---
The branch seed was not used (even if it was mentioned in the comment).

Motivation and Context
---
Problem with doubling the tokens in collectibles because of the pub keys for asset and tokens were the same.

How Has This Been Tested?
---
Manually (fixed the double problem in collectibles)
Added new cargo test for the use of the branch seed.

### DIFF
--- a/base_layer/key_manager/src/key_manager.rs
+++ b/base_layer/key_manager/src/key_manager.rs
@@ -77,7 +77,12 @@ where
 
     /// Derive a new private key from master key: derived_key=SHA256(master_key||branch_seed||index)
     pub fn derive_key(&self, key_index: u64) -> Result<DerivedKey<K>, ByteArrayError> {
-        let concatenated = format!("{}{}", self.seed.entropy().to_vec().to_hex(), key_index);
+        let concatenated = format!(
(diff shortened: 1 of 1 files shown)


base_layer/key_manager/src/key_manager.rs
```diff
@@ -77,7 +77,12 @@ where
 
     /// Derive a new private key from master key: derived_key=SHA256(master_key||branch_seed||index)
     pub fn derive_key(&self, key_index: u64) -> Result<DerivedKey<K>, ByteArrayError> {
-        let concatenated = format!("{}{}", self.seed.entropy().to_vec().to_hex(), key_index);
```
delta1, line 80: @philipr-za does this indicate the current wallets are all deriving on the same branch??

delta1 (approved):
LGTM nice catch

base_layer/key_manager/src/key_manager.rs
philipr-za: Ja it would! This is a good catch.

PR #3752, opened 2022-01-26
chore: reclassify very long log message

CjS77:
The find_chain_split log message is an INFO message with no useful
_consumer_ information, dozens of lines long, hence should be DEBUG


delta1:
definitely agree, just needs cargo fmt

CjS77:
> definitely agree, just needs cargo fmt

doh!

sdbondi (approved):

stringhandler (approved):

PR #3760, opened 2022-01-27
feat: update the available balance in console wallet

philipr-za:
Description
---
Update the available balance in the console wallet to not include the time locked balance. 

Motivation and Context
---
This way the user doesn’t have to do a mental subtraction.

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_console_wallet/src/ui/components/balance.rs
+++ b/applications/tari_console_wallet/src/ui/components/balance.rs
@@ -46,15 +46,12 @@ impl<B: Backend> Component<B> for Balance {
         f.render_widget(block, area);
 
         let balance = app_state.get_balance();
-
+        let time_locked = balance.time_locked_balance.unwrap_or_else(|| MicroTari::from(0u64));
         let available_balance = Spans::from(vec![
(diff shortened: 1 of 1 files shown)


applications/tari_console_wallet/src/ui/components/balance.rs
```diff
@@ -46,15 +46,12 @@ impl<B: Backend> Component<B> for Balance {
         f.render_widget(block, area);
 
         let balance = app_state.get_balance();
-
+        let time_locked = balance.time_locked_balance.unwrap_or_else(|| MicroTari::from(0u64));
         let available_balance = Spans::from(vec![
             Span::styled("Available:", Style::default().fg(Color::Magenta)),
```
delta1, line 51: Is it worth changing "Available" to "Spendable" ?

delta1 (approved):
Totally agree

applications/tari_console_wallet/src/ui/components/balance.rs
philipr-za: I think it makes logical sense like this. The only reason we feel the need to differentiate is our context of how it's represented in the backend.

PR #3764, opened 2022-01-28
feat: re-use scanned range proofs

hansieodendaal:
Description
---
- Re-used scanned range proofs for one-sided payments and wallet recovery instead of re-calculating them every time
- Consistent creation of rewindable outputs in the wallet database

Motivation and Context
---
- Scanning of one-sided payments and wallet recovery was inefficient due to wasteful re-calculating of the range proof before adding the output to the wallet database, which in any case would never correspond to the range proof on the blockchain.
- A mixture of rewindable and non-rewindable outputs were created in certain transaction protocols; all outputs need to be rewindable.

How Has This Been Tested?
---
- Unit tests
- Cucumber tests (`npm test -- --tags "@critical"`)


philipr-za (approved):
LGTM

PR #3767, opened 2022-01-28
feat(ffi)!: add features, metadata_signature and sender_offset_public_key to import_utxo

sdbondi:
Description
---
- adds two arguments to `wallet_import_utxo` (features, metadata_signature and sender_offset_public_key)

Motivation and Context
---
Allows faucet utxos to be imported to the wallet db correctly

How Has This Been Tested?
---

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -109,7 +109,7 @@ use rand::rngs::OsRng;
 use tari_common_types::{
     emoji::{emoji_set, EmojiId, EmojiIdError},
     transaction::{TransactionDirection, TransactionStatus, TxId},
-    types::{ComSignature, PublicKey},
+    types::PublicKey,
 };
 use tari_comms::{
     multiaddr::Multiaddr,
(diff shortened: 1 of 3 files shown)


SWvheerden (dismissed):
This looks good.
I would add script_key and covenants now as well as require inputs. 
We leave todo's here to not break the mobile builds. But as we are requiring more info here we might as well add them now.

sdbondi:
@SWvheerden ok so I think we may as well do that, and if the wallet passes null in we use defaults - wasnt sure if adding new types to ffi is a pain or not - but looks like it's not a problem in unsafe land :)

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4836,6 +4842,31 @@ pub unsafe extern "C" fn wallet_import_utxo(
...
+
+    if script_private_key.is_null() {
+        error = LibWalletError::from(InterfaceError::NullError("script_private_key".to_string())).code;
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return 0;
+    }
+
+    if features.is_null() {
+        *features = OutputFeatures::with_maturity(0);
+    }
```
hansieodendaal, line 4865: This should be an error condition please, in preparation for the upcoming recovery byte. Default features will never be acceptable again. It will either be a function of the commitment alone, or of the commitment and a recovery byte key.

hansieodendaal (commented):
Just the one comment please

hansieodendaal (dismissed):
LGTM

stringhandler (approved):

PR #3776, opened 2022-01-31
chore: check genesis block during startup

Cifko:
Description
---
Check the genesis block from the binary is the same as the one in the DB.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -214,9 +214,9 @@ where B: BlockchainBackend
             difficulty_calculator: Arc::new(difficulty_calculator),
             disable_add_block_flag: Arc::new(AtomicBool::new(false)),
         };
+        let genesis_block = Arc::new(blockchain_db.consensus_manager.get_genesis_block());
         if is_empty {
             info!(target: LOG_TARGET, "Blockchain db is empty. Adding genesis block.");
-            let genesis_block = Arc::new(blockchain_db.consensus_manager.get_genesis_block());
             blockchain_db.insert_block(genesis_block.clone())?;
             let mut txn = DbTransaction::new();
             let body = &genesis_block.block().body;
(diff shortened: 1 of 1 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -230,6 +230,14 @@ where B: BlockchainBackend
             txn.set_horizon_data(kernel_sum, utxo_sum);
             blockchain_db.write(txn)?;
             blockchain_db.store_pruning_horizon(config.pruning_horizon)?;
+        } else if blockchain_db.get_chain_metadata()?.pruning_horizon() == 0 {
+            // Check the genesis block in the DB.
+            if !blockchain_db.block_exists(genesis_block.accumulated_data().hash.clone())? {
+                error!(target: LOG_TARGET, "Stored genesis block mismatch with the binary");
```
delta1: Please add the expected hash of the genesis block, and the hash of the block in the db to this log

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -230,6 +230,14 @@ where B: BlockchainBackend
...
+                return Err(ChainStorageError::CorruptedDatabase(
+                    "Genesis block mismatch.".to_string(),
```
delta1:
```suggestion
                    "Genesis block in database does not match the supplied genesis block in the code! Please delete and resync your blockchain database.".into(),
```

delta1 (commented):
Nice looking good, 2 small comments

base_layer/core/src/chain_storage/blockchain_database.rs
Cifko: Thanks. I'm really not good with wording errors :-)

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -230,6 +230,22 @@ where B: BlockchainBackend
             txn.set_horizon_data(kernel_sum, utxo_sum);
             blockchain_db.write(txn)?;
             blockchain_db.store_pruning_horizon(config.pruning_horizon)?;
+        } else if blockchain_db.get_chain_metadata()?.pruning_horizon() == 0 {
+            // Check the genesis block in the DB.
+            if !blockchain_db.block_exists(genesis_block.accumulated_data().hash.clone())? {
+                error!(
+                    target: LOG_TARGET,
+                    "Genesis block in database does not match the supllied genesis block in the code! Hash in the \
```
```suggestion
                    "Genesis block in database does not match the supplied genesis block in the code! Hash in the \
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -230,6 +230,22 @@ where B: BlockchainBackend
             txn.set_horizon_data(kernel_sum, utxo_sum);
             blockchain_db.write(txn)?;
             blockchain_db.store_pruning_horizon(config.pruning_horizon)?;
+        } else if blockchain_db.get_chain_metadata()?.pruning_horizon() == 0 {
```
stringhandler: The hash should be the same for pruned or archival nodes, so can't we exclude this check?

Cifko: Removed.

delta1 (approved):

PR #3778, opened 2022-01-31
fix: ffi library access violation error

philipr-za:
Description
---
Trying a couple things to fix this access violation error:
1. Using a type alias for Commitment for bindgen's sake
2. Being verbose about order of operations dereferencing the byte vectors

Not 100% either of these are the issue but giving them a try.


philipr-za:
Going to redo it under a new branch

PR #3781, opened 2022-01-31
feat: add logging and config to collectibles

Cifko:
Description
---
Added config and logging to tari collectibles.

Motivation and Context
---
Unite the behavior (configs, logging) of collectibles with other apps.

How Has This Been Tested?
---
Manually.


stringhandler:
The build process deletes applications/tari_collectibles/web-app/build/.gitkeep so just need to exclude that from commits manually until we find a way to prevent it getting deleted

common/config/presets/collectibles.toml
```diff
@@ -0,0 +1,6 @@
+[collectibles]
+########################################################################################################################
+#                                                                                                                      #
+#                                          Collectibles Configuration Options                                          #
+#                                                                                                                      #
+########################################################################################################################
```
hansieodendaal: An empty line is needed at the end of the file please, for when the individual files are stringed together to make the whole.

config/config.toml
```diff
@@ -147,4 +147,9 @@ peer_seeds = [
 committee = ["2ea0df3059caf4411624d6bf5b9c02238d607d2798c586b3e6c2a054da3f205a"] # cannot be of zero size
 phase_timeout = 30
 template_id = "EditableMetadata"
-
+[collectibles]
```
hansieodendaal: **Q:** Why do we have this duplicate configuration options, when we have the master config files in `common/config/presets`? I thought we did away with duplicate configs.

common/config/presets/collectibles.toml
```diff
@@ -0,0 +1,6 @@
+[collectibles]
```
hansieodendaal: This could go beneath the banner to be in line with the other configs?

common/config/presets/collectibles.toml
hansieodendaal: I think we need to add these values with their defaults but commented out:
```
# Description...
#validator_node_grpc_address = ...
# Description...
#base_node_grpc_address = ...
# Description...
#wallet_grpc_address = ...
```

config/config.toml
stringhandler: I think this file has been unintentionally included in source some how and should be deleted

stringhandler (dismissed):
Happy with this, let's fix the CI and it can be merged

config/config.toml
Cifko: Honestly, I don't know. But since the config is there, I added the collectibles part.

common/config/presets/collectibles.toml
Cifko: Done, I did it based on validator node config, which is this way. Will change that as well.

config/config.toml
Cifko: Removed the file.

Cifko: Ok, Mike explained, the file is deleted in this PR.

stringhandler (dismissed):

stringhandler (approved):

PR #3782, opened 2022-02-01
chore: remove debug info from get-mempool-state

sdbondi:
Description
---
Cleans up some debug info in `get-mempool-state` command

Motivation and Context
---
Debug info made `get-mempool-state` basically unreadable when txns had many inputs

How Has This Been Tested?
---
Manually


delta1 (approved):
LGTM

PR #3783, opened 2022-02-01
docs(rfc): consensus encoding

sdbondi:
Description
---

Adds RFC detailing current consensus encoding usage in the Tari base layer 

- Code changes: remove unnecessary allocations from metadata signature challenge


RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,346 @@
...
+signature challenges used in base-layer consensus. 
+
+## Related Requests for Comment
+
+* [RFC-0100: Base Layer](RFC-0100_BaseLayer.md)
+* [RFC-0120: Consensus](RFC-0120_Consensus.md)
+
+## Description
+
+A Tari base node must validate each block containing a [block header] as well as set of [transaction input]s, [transaction outputs] and [transaction kernel]s,
```
delta1: These links are missing

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Commitment
+
+A commitment is a [RistrettoPublicKey] and so has identical encoding.
+
+#### Schnorr Signature
+
+See the [TLU on Schnorr Signatures](https://tlu.tarilabs.com/cryptography/introduction-schnorr-signatures)
+
+A scnorr signature tuple is `<R, s>` where `R` is a [RistrettoPublicKey] and `s` is a the signature scalar wrapped in [RistrettoPrivateKey].
```
delta1:
```suggestion
A Schnorr signature tuple is `<R, s>` where `R` is a [RistrettoPublicKey] and `s` is a the signature scalar wrapped in [RistrettoPrivateKey].
```

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,346 @@
...
+2. `height` - `u64` converted to a fixed 8-byte array (little endian) 
+3. `prev_hash` - fixed 32-bytes 
+4. `timestamp` - UNIX timestamp converted to a fixed 8-byte array (little endian) 
+5. `input_mr` - fixed 32-bytes
+6. `output_mmr_size` - `u64` converted to a fixed 8-byte array (little endian)
+7. `witness_mr` - fixed 32-bytes
+8. `kernel_mr` - fixed 32-bytes
+9. `kernel_mmr_size` - `u64` converted to a fixed 8-byte array (little endian)
+10. `total_kernel_offset` - 32-byte Scalar, see [RistrettoSecretKey]
+11. `total_script_offset` - 32-byte Scalar, see [RistrettoSecretKey]
```
delta1: These links are missing

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,346 @@
...
+
+1. `merge_mining_hash` - As above
+2. `pow_algo` - enumeration of types of PoW as a single unsigned byte, where `Monero = 0x00` and `Sha3 = 0x01`
+3. `pow_data` - raw variable bytes (no length varint)
+4. `nonce` - the PoW nonce, `u64` converted to a fixed 8-byte array (little endian)
+
+#### Output Features 
+[Output Features]: #output-metadata-and-features "Output Features"
+
+```rust
```
delta1: mdbook test is failing, you'll need to `ignore` the code blocks 

```suggestion
```rust, ignore
```

delta1 (commented):
Looking good! A few small comments. Are those code changes supposed to be included with this PR?

sdbondi:
Thanks - updated the description, code changes is just to avoid unnecessary allocations

SWvheerden (dismissed):
LGTM

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Related Requests for Comment
+
+* [RFC-0100: Base Layer](RFC-0100_BaseLayer.md)
+* [RFC-0120: Consensus](RFC-0120_Consensus.md)
+
+## Description
+
+A Tari base node must validate each block containing a [block header] as well as set of transaction inputs, transaction outputs and transaction kernels,
+each containing a number of fields pertinent to their function within the [base layer]. The data contained within these structures need to be consistently encoded
```
stringhandler: I know data is plural, but this feels better to me. idk?
```suggestion
each containing a number of fields pertinent to their function within the [base layer]. The data contained within these structures needs to be consistently encoded
```

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+Note that nested structs are flattened and the order must be preserved to allow decoding.
+The `00` null byte is important so that for e.g. the `kind` bytes cannot be manipulated to 
+produce the same encoding as non-null `dob`.
+
+### Block Header 
+[block header]: #block-header "Block header"
+
+The block hash pre-image is constructed by first constructing the merge mining hash. Each encoding is concatenated in order as follows:
+
+1. `version` - `u16` converted to a fixed 2-byte array (little endian)
```
stringhandler: In the next iteration, `version` and `height` should be VarInts

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+  pub parent_public_key: Option<PublicKey>,
+  pub asset: Option<AssetOutputFeatures>,
+  pub mint_non_fungible: Option<MintNonFungibleFeatures>,
+  pub sidechain_checkpoint: Option<SideChainCheckpointFeatures>,
+}
+```
+
+Output features consensus encoding is defined as follows (in order):
+
+1. `version` - 1 unsigned byte. This should always be `0x00` but is reserved for future proofing.
```
stringhandler, line 219: Perhaps a varint rather?

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+- `commitment` - [RistrettoPublicKey]
+- `script` - byte length as [varint] + [TariScript]
+- `covenant` - byte length as [varint] + [Covenant]
+
+##### Witness hash
+
+The witness hash is appended to the witness Merkle tree.
+
+- `proof` - Raw proof bytes encoded using [dynamic vector] encoding
+- `metadata_signature` - [SchnorrSignature](#schnorr-signature)
```
stringhandler:
```suggestion
- `metadata_signature` - CommitmentSignature
```

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+    pub commitment: Commitment,
+    pub proof: RangeProof,
+    pub script: TariScript,
+    pub sender_offset_public_key: PublicKey,
+    pub metadata_signature: ComSignature,
+    pub covenant: Covenant,
+}
+```
+
+The canonical hash commits to the common data between an output and the input spending that output
```
stringhandler: Why not use the same hash as the one in the output merkle tree?

RFC/src/RFC-0121_ConsensusEncoding.md
```diff
@@ -0,0 +1,347 @@
...
+  pub output_version: TransactionOutputVersion,
+  pub features: OutputFeatures,
+  pub commitment: Commitment,
+  pub script: TariScript,
+  pub sender_offset_public_key: PublicKey,
+  pub covenant: Covenant, 
+}
+```
+
+The transaction input canonical hash pre-image is constructed as follows:
```
stringhandler, line 310: I would suggest using this instead:
input_version
output_hash
input_data
script_signature

stringhandler (commented):
Some questions. 

I think we should also remove any examples with Strings, because we have not included whether they would be null-terminated or not

RFC/src/RFC-0121_ConsensusEncoding.md
sdbondi: I don't have a strong opinion, 256 versions seems like plenty but obviously varint is completely future proof. As long as version < 128 we can change to varint later without it being a breaking change. Maybe the best argument is to have all integers are varint for consistency, but idk.

RFC/src/RFC-0121_ConsensusEncoding.md
sdbondi: It is the same hash. Maybe the wording can be improved?

RFC/src/RFC-0121_ConsensusEncoding.md
sdbondi: This just reflects what is in the code, but this makes sense since the input hash can be generated even with the input in 'compact mode' - I'll update here and add a TODO to the code

RFC/src/RFC-0121_ConsensusEncoding.md
stringhandler: Yeah, it really doesn't matter

RFC/src/RFC-0121_ConsensusEncoding.md
stringhandler: Yeah let's correct the problems in this RFC and then we'll create issues/bugs to fix the code

stringhandler (approved):

PR #3785, opened 2022-02-01
fix(comms): minor edge-case fix to handle inbound connection while dialing

sdbondi:
Description
---
- Notifies the dialler if an inbound connection has come in, so that it can notify the waiting diallers of the new connection

Motivation and Context
---
Noticed this edge case when a wallet start up and dials a base node, I then quickly manually dial the wallet from the same base node. The bn->wallet connection finishes first but the wallet still is in connecting state. The wallet does not immediately flip into connected state because it is waiting for the dial to finish, however this will also cause a tie-break resolution which may take longer. This PR resolves this by notifying the dialler of all inbound connections so that it can notify any waiting dials of the new connection immediately.

How Has This Been Tested?
---
Partially by existing unit tests, manually


delta1 (approved):
LGTM

PR #3788, opened 2022-02-02
fix(core)!: fix potential panic for sidechain merkle root with incorrect length

sdbondi:
Description
---
Fixes potential panic if a side-chain Merkle root of incorrect length is accepted by the blockchain.

Motivation and Context
---
Using a fixed 32-byte array forces any implementor to produce or validate the correct number of bytes. Esp. now that const generics are available, for performance and correctness, all hashes in the system should be fixed to 32 bytes and not held in a dynamic heap allocated vector.

I've changed the merkle_root type in SideChainCheckpoint because this is a relatively minor change.

This is a chain storage breaking change. 

How Has This Been Tested?
---
Existing tests, manually


base_layer/common_types/src/array.rs
```diff
@@ -0,0 +1,32 @@
...
+//  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+//  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use tari_utilities::ByteArrayError;
+
+pub fn copy_into_fixed_array<T: Default + Copy, const SZ: usize>(elems: &[T]) -> Result<[T; SZ], ByteArrayError> {
```
philipr-za, line 25: nice

philipr-za (approved):
Looks good, death to the unwrap!

base_layer/common_types/src/array.rs
sdbondi: :) Can probably be wrapped into a FixedHash type in future

stringhandler (approved):

PR #3789, opened 2022-02-02
fix: coinbase output recovery bug

philipr-za:
Description
---
A bug was discovered when recovering Coinbase outputs that was revealed when the Key Manager was fixed to actually use branch seeds. Coinbase output keys are derived on a separate key manager branch so when the UTXO scanner tried to update the key manager index it would not find the coinabse key in the main spending key branch which caused an error.

This PR updates the UTXO scanner to check if a recovered output has the coinbase flag or not and then searches on the Coinbase branch when looking for that outputs key index and script key.

How Has This Been Tested?
---
The PR updates the `Wallet recovery with connected base node staying online` cucumber test to include recovering some coinbase outputs.
The PR also updates the `Multiple Wallet recovery from seed node` cucumber test to work now that the coinbase issue is fixed. The test is also updated so that it recovered N distinct wallets to the same seed node where as before it would create N wallet with the same seed words and thus network identity.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/master_key_manager.rs
+++ b/base_layer/wallet/src/output_manager_service/master_key_manager.rs
@@ -20,6 +20,8 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::fmt::{Display, Error, Formatter};
+
 use futures::lock::Mutex;
 use log::*;
 use tari_common_types::types::{PrivateKey, PublicKey};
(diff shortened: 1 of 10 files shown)


base_layer/wallet/src/output_manager_service/master_key_manager.rs
```diff
@@ -41,14 +43,32 @@ use crate::{
...
+    Coinbase,
+    CoinbaseScript,
+    RecoveryViewOnly,
+    RecoveryBlinding,
+}
+
+impl Display for KeyManagerBranch {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> Result<(), Error> {
+        let response = match self {
+            KeyManagerBranch::Spend => "",
```
sdbondi, line 61: Should this be blank?

sdbondi (approved):
LGTM

base_layer/wallet/src/output_manager_service/master_key_manager.rs
philipr-za: Ja, in the beginning the first branch key was just blank but now it would be a breaking change to update that :P

PR #3791, opened 2022-02-02
chore: show error when not enough utxo for checkpoint

Cifko:
Description
---
Shows error when you don't have at least 2 utxos for asset creation.

How Has This Been Tested?
---
Manually.

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -44,6 +44,8 @@ service Wallet {
     rpc GetCompletedTransactions (GetCompletedTransactionsRequest) returns (stream GetCompletedTransactionsResponse);
     // Returns the balance
     rpc GetBalance (GetBalanceRequest) returns (GetBalanceResponse);
+    // Returns unspent amounts
+    rpc GetUnspentAmounts (Empty) returns (GetUnspentAmountsResponse);
     // Request the wallet perform a coinsplit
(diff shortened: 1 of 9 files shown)


applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -206,6 +208,10 @@ message GetBalanceResponse {
     uint64 pending_outgoing_balance = 3;
 }
 
+message GetUnspentOutputsResponse {
+    repeated UnblindedOutput outputs = 1;
```
stringhandler: This will expose the blinding factor.... can we provide a slimmer interface

SWvheerden (commented):
I am assuming that the whole intention of this PR is to make sure that the wallet has enough funds in enough utxo's to do checkpointing. 

Using the `get_unspent_amounts` call will get you the unspent utxos in the wallet, but those funds may be unspendable as they might be timelocked and or not confirmed.

Cifko:
> I am assuming that the whole intention of this PR is to make sure that the wallet has enough funds in enough utxo's to do checkpointing.
> 
> Using the `get_unspent_amounts` call will get you the unspent utxos in the wallet, but those funds may be unspendable as they might be timelocked and or not confirmed.

The uncofirmed were not there. But the timelocked were. I've added a new method to just get the amounts of utxos that spendable.

stringhandler:
This seems to be implemented in #3802

PR #3793, opened 2022-02-03
feat: add specific LibWallet error code for “Fee is greater than amount”

philipr-za:
Description
---
Surfacing a specific error code for the “Fee is greater than amount” Transaction Build error in the LibWalletError for the FFI interface.

### DIFF
--- a/base_layer/wallet_ffi/src/error.rs
+++ b/base_layer/wallet_ffi/src/error.rs
@@ -213,6 +213,12 @@ impl From<WalletError> for LibWalletError {
                 code: 204,
                 message: format!("{:?}", w),
             },
+            WalletError::TransactionServiceError(TransactionServiceError::OutputManagerError(
+                OutputManagerError::BuildError(ref s),
+            )) if s == &"Fee is greater than amount".to_string() => Self {
+                code: 212,
+                message: format!("{:?}", w),
+            },
             WalletError::TransactionServiceError(TransactionServiceError::OutputManagerError(_)) => Self {
                 code: 206,
                 message: format!("{:?}", w),
@@ -229,6 +235,7 @@ impl From<WalletError> for LibWalletError {
                 code: 211,
                 message: format!("{:?}", w),
             },
+
             // Comms Stack errors
             WalletError::MultiaddrError(_) => Self {
                 code: 301,


base_layer/wallet_ffi/src/error.rs
```diff
@@ -213,6 +213,12 @@ impl From<WalletError> for LibWalletError {
                 code: 204,
                 message: format!("{:?}", w),
             },
+            WalletError::TransactionServiceError(TransactionServiceError::OutputManagerError(
+                OutputManagerError::BuildError(ref s),
+            )) if s == &"Fee is greater than amount".to_string() => Self {
```
delta1, line 218: Not to be a pain, but this should probably be a const used in both places. Otherwise someone might change the other one without updating this one.

philipr-za: Would require a lot of const's, the Build Error is all strings, worst case here it will be caught by error 206

philipr-za: Ideally, BuildError should be an enum but not really looking to refactor it here

delta1: Sorry, I just meant making "Fee is greater than amount" a const &str and used here, and also the place where it's returned (transaction init?) 

```rust
pub const FEE_GT_AMOUNT_ERR: &str = "Fee is greater than amount";
```

Since you're relying on the strings being the same in both places

philipr-za: I would say that using const's is not the way to clean this up. It should be to refactor the TransactionError::BuildError to use the normal Anyhow Error paradigm we use everywhere else:

https://github.com/tari-project/tari/blob/development/base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs#L611

But that is not a very small job so happy to put it in the backlog.

delta1: That's fine I agree, the point was that if this string is changed there it will break your check here

delta1 (approved):

base_layer/wallet_ffi/src/error.rs
philipr-za: Ja I agree it's not ideal and have made a ticket to refactor BuildError based on the fragility this issue has revealed. Just didn't want to normalize the const strategy in the code base when the canonical solution should be to use Error enums.

delta1: Perfect

PR #3794, opened 2022-02-03
feat!: add scanned transaction handling for one-sided payments with callbacks

hansieodendaal:
Description
---
- Added separate statuses for scanned one-sided transactions to enable unique events and validation.
- Added validation for one-sided payments updating their mined height and confirmations.
- Added callbacks for scanned one-sided transactions (unconfirmed and confirmed)
- Updated the wallet FFI with the callbacks.

Motivation and Context
---
Currently when a one-sided output is found by scanning the blockchain a Faux transaction is created for it with the Imported status, the same as for a manually imported output.  Outputs found by scanning (for one-sided payments) should be handled differently to a manual import and their mined height and number confirmations should be validated.

How Has This Been Tested?
---
- Unit tests
- Cucumber tests
- System level testing


base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -388,6 +388,8 @@ where
         mined_height: u64,
         num_confirmations: u64,
     ) -> Result<(), TransactionServiceProtocolError> {
+        let is_scanned =
```
philipr-za: Instead of reusing this statement in many places make it a method on TransactionStatus i.e. `status.is_scanned()`

base_layer/common_types/src/transaction.rs
```diff
@@ -71,10 +77,56 @@ impl Display for TransactionStatus {
...
+            TransactionStatus::ScannedUnconfirmed => write!(f, "ScannedUnconfirmed"),
+            TransactionStatus::ScannedConfirmed => write!(f, "ScannedConfirmed"),
         }
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum ImportStatus {
+    /// This is an invalid transaction import status that will result in an error
+    Invalid,
```
philipr-za: I am not understanding why this variant exists?

base_layer/common_types/src/transaction.rs
```diff
@@ -71,10 +77,56 @@ impl Display for TransactionStatus {
...
+        match value {
+            TransactionStatus::Imported => Ok(ImportStatus::Imported),
+            TransactionStatus::ScannedUnconfirmed => Ok(ImportStatus::ScannedUnconfirmed),
+            TransactionStatus::ScannedConfirmed => Ok(ImportStatus::ScannedConfirmed),
+            _ => Err(TransactionConversionError { code: i32::MAX }),
+        }
+    }
+}
+
+impl Default for ImportStatus {
```
philipr-za: Does a default make sense for this kind of enum where you should really choose a specific value on creation?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1750,9 +1779,11 @@ impl CompletedTransactionSql {
         Ok(())
     }
 
-    pub fn set_as_unmined(&self, conn: &SqliteConnection) -> Result<(), TransactionStorageError> {
+    pub fn set_as_unmined(&self, conn: &SqliteConnection, is_scanned: bool) -> Result<(), TransactionStorageError> {
```
philipr-za: Do you need the `is_scanned` argument? The CompletedTransactionSql already has a status inside it so don't you just have to check if it was ScannedConfirmed?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1798,11 +1829,18 @@ impl CompletedTransactionSql {
         num_confirmations: u64,
         is_confirmed: bool,
         conn: &SqliteConnection,
+        is_scanned: bool,
```
philipr-za: is this needed? Can't you just check if it was Scanned{Unconfirmed/Confirmed}?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -2002,7 +2040,8 @@ impl UnconfirmedTransactionInfoSql {
                     .and(
                         completed_transactions::mined_height
                             .is_null()
-                            .or(completed_transactions::status.eq(TransactionStatus::MinedUnconfirmed as i32)),
+                            .or(completed_transactions::status.eq(TransactionStatus::MinedUnconfirmed as i32))
+                            .or(completed_transactions::status.eq(TransactionStatus::ScannedUnconfirmed as i32)),
```
philipr-za: Is this correct? This function is used by the validation protocol to request transaction that it can verify on the blockchain. However, ScannedUnconfirmed means it is a Faux transaction. I would say do not return these in the same method as you will have to filter them in the Validation protocol but rather make a separate DB method to get the ScannedUnconfirmed transactions.

philipr-za: In fact note on line 2039 I explicitly filter OUT imported transactions

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -491,7 +494,13 @@ where TBackend: WalletBackend + 'static
                     .scan_for_recoverable_outputs(outputs.clone())
                     .await?
                     .into_iter()
-                    .map(|v| (v, format!("Recovered on {}.", Utc::now().naive_utc())))
+                    .map(|v| {
+                        (
+                            v,
+                            format!("Recovered on {}.", Utc::now().naive_utc()),
+                            ImportStatus::Imported,
```
philipr-za: Shouldn't a recovered output be treated the same as a one-sided?

base_layer/wallet/src/wallet.rs
```diff
@@ -385,6 +385,7 @@ where
         sender_offset_public_key: &PublicKey,
         script_lock_height: u64,
         covenant: Covenant,
+        import_status: ImportStatus,
```
philipr-za: I am not sure this status should be exposed here, the appropriate status is really an internal piece of data. I think the reason you added it is that you are treating Recovered Outputs as Imported as One-sided as Scanned but I think that Recovered Output and One-sided should be Scanned and only the unblinded Faucet import should be Imported. That means you can take the status out of these public facing methods

base_layer/wallet/src/wallet.rs
```diff
@@ -429,14 +437,16 @@ where
...
     ) -> Result<TxId, WalletError> {
         let tx_id = self
             .transaction_service
-            .import_utxo(
+            .import_utxo_with_status(
                 unblinded_output.value,
                 source_public_key.clone(),
                 message,
                 Some(unblinded_output.features.maturity),
+                import_status.clone(),
```
philipr-za: This should just be Imported

base_layer/wallet/src/wallet.rs
```diff
@@ -401,7 +402,13 @@ where
 
         let tx_id = self
             .transaction_service
-            .import_utxo(amount, source_public_key.clone(), message, Some(features.maturity))
+            .import_utxo_with_status(
+                amount,
+                source_public_key.clone(),
+                message,
+                Some(features.maturity),
+                import_status.clone(),
```
philipr-za: This this should just be ScannedUnconfirmed

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -40,6 +40,8 @@
 //! `callback_transaction_mined` - This will be called when a Broadcast transaction is detected as mined via a base
 //! node request
 //!
+//! `callback_transaction_scanned` - This will be called when a one-sided transaction is detected as mined
+//!
```
philipr-za, line 48: Missing the description of unconfirmed method

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -40,6 +40,8 @@
 //! `callback_transaction_mined` - This will be called when a Broadcast transaction is detected as mined via a base
 //! node request
 //!
+//! `callback_transaction_scanned` - This will be called when a one-sided transaction is detected as mined
```
philipr-za: add `mined` into the callback name its not clear this callback is for when the scanned output becomes mined.

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -78,6 +80,8 @@ where TBackend: TransactionBackend + 'static
     callback_transaction_broadcast: unsafe extern "C" fn(*mut CompletedTransaction),
     callback_transaction_mined: unsafe extern "C" fn(*mut CompletedTransaction),
     callback_transaction_mined_unconfirmed: unsafe extern "C" fn(*mut CompletedTransaction, u64),
+    callback_transaction_scanned: unsafe extern "C" fn(*mut CompletedTransaction),
+    callback_transaction_scanned_unconfirmed: unsafe extern "C" fn(*mut CompletedTransaction, u64),
```
philipr-za: Are we using this callback to also indicate when the Scanned output is initially detected? I think that is ok but just checking that this event is emitted when the initially detection occurs?

philipr-za (commented):
So I feel like something is missing here. The scanned output faux transactions cannot be validated like normal transactions because we do not know their kernels. Their status needs to be driven by the status of the outputs, which for imported transaction was done using the `check_imported_transactions` task but I can't see anything similar for `ScannedUnconfirmed` outputs.

base_layer/common_types/src/transaction.rs
hansieodendaal: Agreed

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
hansieodendaal: Agreed

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
hansieodendaal: Agreed

base_layer/wallet_ffi/src/callback_handler.rs
hansieodendaal: Agreed

base_layer/wallet_ffi/src/callback_handler.rs
hansieodendaal: Renamed

base_layer/wallet_ffi/src/callback_handler.rs
hansieodendaal: Yes, that event is emitted when the initial detection occurs

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
hansieodendaal: Agreed

base_layer/common_types/src/transaction.rs
hansieodendaal: Agreed, removed

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
hansieodendaal: Good idea, doing it

base_layer/wallet/src/wallet.rs
hansieodendaal: Agreed

base_layer/wallet/src/wallet.rs
hansieodendaal: Agreed

base_layer/wallet/src/wallet.rs
hansieodendaal: Agreed

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
hansieodendaal: Agreed

base_layer/common_types/src/transaction.rs
```diff
@@ -27,6 +27,27 @@ pub enum TransactionStatus {
...
+            TransactionStatus::MinedUnconfirmed => false,
+            TransactionStatus::Imported => true,
+            TransactionStatus::Pending => false,
+            TransactionStatus::Coinbase => false,
+            TransactionStatus::MinedConfirmed => false,
+            TransactionStatus::Rejected => false,
+            TransactionStatus::FauxUnconfirmed => true,
+            TransactionStatus::FauxConfirmed => true,
+        }
+    }
```
philipr-za, line 50:
```suggestion
impl TransactionStatus {
    pub fn is_faux(&self) -> bool {
        match self {
            TransactionStatus::Imported |
            TransactionStatus::FauxUnconfirmed |
            TransactionStatus::FauxConfirmed => true,
            _ => false,
        }
    }
```

philipr-za (approved):
Looks good. My one comment is just a Nit so approving and going start a merge queue

base_layer/common_types/src/transaction.rs
hansieodendaal: Thanks, I will fix that as a side issue when I submit a new PR based on this merge.

PR #3798, opened 2022-02-03
fix(core): reduce one block behind waiting period

sdbondi:
Description
---
Reduces the time taken to switch to sync when one block behind (was 60s, now 20s)

Motivation and Context
---
With the improvement of block propagation, the time between detecting that a peer/s has progressed and receiving a new block is reduced, so it follows the "one block behind" waiting period should be reduced.

How Has This Been Tested?
---
Manually, minor change


philipr-za (approved):
lgtm

PR #3802, opened 2022-02-07
feat: read asset definitions from base layer

Cifko:
Description
---
Load assets from base layer. And periodically check for new assets.

How Has This Been Tested?
---
Manually.


stringhandler (commented):
looks good from a quick scan, but I'll need to test it locally

stringhandler (approved):

PR #3803, opened 2022-02-07
feat(validator_node): add get_sidechain_block p2p rpc method

sdbondi:
Description
---

- Add get_sidechain_blocks to validator node p2p rpc
- Use fixed 32-byte array for hashes on dan layer
- Implement FromStr and fallible integer conversions for TemplateId
- Remove asset_public_key from Instruction
- Add necessary chain db calls
- Add client call for get_sidechain_blocks
- Adds mocks and basic tests for get_sidechain_blocks validator node rpc

Motivation and Context
---
Allow client to fetch sidechain instruction state from vn

How Has This Been Tested?
---
Manually


dan_layer/core/src/types.rs
```diff
@@ -26,6 +26,8 @@ use tari_crypto::{
     tari_utilities::{ByteArray, ByteArrayError},
 };
 
+pub type Hash = [u8; 32];
```
stringhandler: Not used?

stringhandler (dismissed):
Nice. Thanks for cleaning up some of the dirty bits I left

dan_layer/core/src/types.rs
sdbondi: Don't think it's used, will remove in next PR (adding to client and some tests) if this is merged tomorrow

sdbondi: Removed in this PR since there was a conflict

delta1 (approved):

PR #3807, opened 2022-02-08
fix(wallet): fix aggressive disconnects in wallet connectivity

sdbondi:
Description
---
- Disconnects the wallet peer only if the pool is unable to create a new RPC session.
- Clear the `pool` when disconnecting, so that requests for a client session will wait
- Immediately return an error if dialling a peer that is banned 

Motivation and Context
---
Observed the wallet continuously disconnecting and reconnecting to base node, resulting in many of these message:
`Outbound messaging protocol failed for peer xxxxx: IO Error: 5e5731a0/4: connection is closed`

This is not a confirmed fix for the issue, because it is difficult to reproduce, but I suspect the problem was not clearing the 
pool in wallet connectivity when disconnecting was causing disconnect to be called whenever a peer session was requested by the monitor etc., sometimes timed in such a way that it would disconnect the new connection that was just established, and that would continue indefinitely. 

How Has This Been Tested?
---
Manually, switching base nodes, stop base node while connected then restart


delta1 (dismissed):
LGTM, untested

PR #3808, opened 2022-02-08
fix: improved image handling in collectibles

delta1:
Description
---

- Adds "back" and "reset" buttons to various parts of image selection in collectibles
- Adds a timestamp for the call to `ipfs files cp` in case of duplicate


sdbondi (approved):
Looks good - untested

PR #3810, opened 2022-02-08
refactor: split `ErrorCodes` into the `ExitError` and `ExitCode` pair

therustmonk:
Description
---
The PR splits `ErrorCodes` into the pair of `ExitCode` enum and `ExitError` struct.

Motivation and Context
---
The purpose of this change is to separate strings from error codes in order to later use more specific error types instead of strings or `ErrorCodes` directly to have detailed backtraces that will simplify and speed up debugging of the code.

How Has This Been Tested?
---
CI pass expected

### DIFF
--- a/applications/launchpad/backend/src/docker/error.rs
+++ b/applications/launchpad/backend/src/docker/error.rs
@@ -23,7 +23,7 @@
 
 use std::error::Error;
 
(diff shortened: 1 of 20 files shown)


sdbondi (approved):
Looks good, similar to `io::Error`

delta1 (approved):
Nice one, let's get this in before any conflicts

PR #3811, opened 2022-02-08
chore: change error propagation in validator node

Cifko:
Description
---
This PR changes all (that can fail) `unwraps` to `?`.

Motivation and Context
---
There were too many unwraps that did not propagate the errors. It's better if we want to handle non fatal errors in the code instead of panicking .

How Has This Been Tested?
---
Just by running the validator node and collectibles.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6973,6 +6973,7 @@ dependencies = [
  "thiserror",
  "tokio 1.17.0",
  "tokio-stream",
+ "tonic",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -338,7 +338,7 @@ impl ConsensusConstants {
         });
         vec![ConsensusConstants {
             effective_from_height: 0,
-            coinbase_lock_height: 360,
+            coinbase_lock_height: 3,
```
SWvheerden: ?

SWvheerden (changes requested):
This looks good. But did you leave your test code in?

base_layer/core/src/consensus/consensus_constants.rs
Cifko: heh, yeah, It's hard to test with 360 ;-)

stringhandler (dismissed):

stringhandler (approved):

PR #3813, opened 2022-02-09
ci: don't run binary builds daily

delta1:
Description
---

- removes binary build matrix daily, keeps on tags only 
- removes old scripts from libwallet build migration 
- removes libwallet daily build


leet4tari:
LGTM - but I wonder if reducing to builds only once a week and keep the builds on branches starting with ```builds-```?

PR #3814, opened 2022-02-09
fix: fix merge mining proxy pool mining

hansieodendaal:
Description
---
Fixed merge mining proxy pool mining:
- XMRig/merge_mininig_proxy error: 'required field "blocktemplate_blob" not found'
- config file was not read correctly;
- global config exited prematurely ;
- round-robin for `monerod_url` - start at the next entry in the list when encountering connection errors;
- updated cucumber integration test.

Motivation and Context
---
When used with Monero mainnet and pool mining, XMRig reported many errors and exited after mining for a while:
- XMRig/merge_mininig_proxy had an issue with pool mining:
   ```
   [2022-02-01 23:56:05.773]  origin   submitted to origin daemon (174/748)  diff 110685 vs. 311330
   [2022-02-01 23:56:06.135]  cpu      accepted (859/61) diff 40000 (361 ms)
   [2022-02-01 23:56:06.517] [127.0.0.1:7878] required field "blocktemplate_blob" not found
   ```
- Default merge mining proxy config on windows was wrong
- When `monerod_url` had an error new connection attempts were always started at the first entry instead of at the last used entry

How Has This Been Tested?
---
Cucumber tests for merge mining
System-level tests


applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -605,20 +642,41 @@ impl InnerService {
...
+                .read()
+                .expect("Read lock should not fail")
+                .clone();
+            match lock {
+                Some(url) => url,
+                None => "".to_string(),
             }
+        };
+        let mut iter = self.config.monerod_url.iter();
+        let _ = iter.find(|&x| x == &current_url);
```
philipr-za: This statement doesn't seem to do anything as the result is swallowed?

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -605,20 +642,41 @@ impl InnerService {
...
+        let mut count = 0;
+        loop {
+            count += 1;
+            if count > self.config.monerod_url.len() * 2 {
```
philipr-za: times 2? Is the intention to go through this list twice?

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: I am using this feature of `find` to start iterating at the next element in the list - from the help files:
``` rust
Stopping at the first true:
let a = [1, 2, 3];

let mut iter = a.iter();

assert_eq!(iter.find(|&&x| x == 2), Some(&2));

// we can still use `iter`, as there are more elements.
assert_eq!(iter.next(), Some(&3));
```

delta1: ah interesting! maybe just add a comment to that line - something like "advance the iterator to the position of current_url"

philipr-za: Ja I would have made this more readable by using the `position` method and then the `skip` method to start from that position in the loop.

delta1: even better

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: Yes, added a comment

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: I do not think using `skip` - which is superfluous - will add more value than the comment I have added, as advancing the iterator after the use of `position` and `find` is identical.

```rust
Basic usage:
let a = [1, 2, 3];

let mut iter = a.iter();

assert_eq!(a.iter().position(|&x| x == 2), Some(1));

// we can still use `iter`, as there are more elements.
assert_eq!(iter.next(), Some(&3));

```

philipr-za: The point would be to not use a `loop` but rather use the position to find how much to skip and then use a For statement on the iterator that is clearly defined in code using a skip and a chain statement to go through the list a second time.

It's much easier to be explicit on how your are iterating by using iterators and combinators.

Working on your basic example iterating through `a` starting from 2 and then going through the list a second time you would do this:

```Rust
for a.iter().skip(a.iter().position(|x| ==2).chain(a.iter()) {
// do stuff
}
```

applications/tari_merge_mining_proxy/src/proxy.rs
```diff
@@ -605,20 +642,43 @@ impl InnerService {
...
-                    let mut lock = self.last_available_server.write().expect("Write lock should not fail");
-                    *lock = Some(monerod_url.to_string());
-                    info!(target: LOG_TARGET, "Monerod server available: {:?}", uri.clone());
-                    return Ok(uri);
-                },
-                Err(_) => {
-                    warn!(target: LOG_TARGET, "Monerod server unavailable: {:?}", uri);
-                    continue;
-                },
+        let current_url = {
```
stringhandler: This will always be `"".to_string()` because of the above return statement, which makes the extra code here and the find below unnecessary

stringhandler (commented):
I'm not convinced this is an improvement. The fixes to the config file are good, but the other changes, specifically the one I highlighted, do not seem to make any difference to the existing code

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: Good pickup, I will fix the logic error here. If we have a troublesome `monerod` we do not want to connect to it straight away again even if it is possible; much better to try the next server in the list. `self.last_available_server` is zeroed if `async fn proxy_request_to_monerod` returns an error.

hansieodendaal: Done

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: Re-arranged the list to start from the next unused entry, expanded the iterator with the chained operator, removed the loop.

applications/tari_merge_mining_proxy/src/proxy.rs
hansieodendaal: See previous comment - removed the loop

stringhandler (approved):

PR #3817, opened 2022-02-10
chore: change NodeIdentity debug

Cifko:
Description
---
Remove secret_key from debug print for `NodeIdentity`. The display is using the same fields that are used in the `fmt::Display` trait


SWvheerden (dismissed):
LGTM

comms/src/peer_manager/node_identity.rs
```diff
@@ -221,3 +221,14 @@ impl fmt::Display for NodeIdentity {
         Ok(())
     }
 }
+
+impl fmt::Debug for NodeIdentity {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("NodeIdentity")
```
sdbondi, line 227: nit: Debug usually shows all fields, so I think include all fields here but redact the ones that should never be logged:
e.g. see `base_layer/core/src/transactions/transaction_components/unblinded_output.rs`

`.field("secret_key", &"<secret>")`

It's fine for the identity signature to be logged

sdbondi (commented):
Looks good, one nit

comms/src/peer_manager/node_identity.rs
sdbondi: Debug should show all fields, add
`.field("identity_signature", &*acquire_read_lock!(self.identity_signature))`

stringhandler (approved):

sdbondi (approved):

PR #3819, opened 2022-02-10
test: fix transaction output hash calculation in cucumber

hansieodendaal:
Description
---
- Fixed the transaction output hash calculation according to the recent consensus encoding rules.
- Removed `@broken` tags from 2x critical tests.

Motivation and Context
---
See above.

How Has This Been Tested?
---
Cucumber tests: 
- `As a wallet I want to submit a transaction`
- `As a wallet I cannot submit a locked coinbase transaction`


philipr-za (approved):
LGTM

PR #3820, opened 2022-02-10
chore!: change hash to use consensus encoding

Cifko:
Description
---
change hashes to use consensus encoding


base_layer/core/src/proof_of_work/proof_of_work.rs
```diff
@@ -89,6 +96,20 @@ impl Display for ProofOfWork {
     }
 }
 
+impl ConsensusEncoding for ProofOfWork {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, io::Error> {
+        let mut written = writer.write_varint(self.pow_algo.as_u64())?;
```
hansieodendaal: The `pow_algo` is an enum - could we not just treat it as a single byte instead of as a u64, similar to `impl ConsensusEncoding for OutputFeaturesVersion {`?
```suggestion
        writer.write_all(&[self.pow_algo as u8])?;
        let mut written = 1;
```

base_layer/core/src/proof_of_work/proof_of_work.rs
```diff
@@ -89,6 +96,20 @@ impl Display for ProofOfWork {
...
+        written += self.pow_data.consensus_encode(writer)?;
+        Ok(written)
+    }
+}
+
+impl ConsensusEncodingSized for ProofOfWork {
+    fn consensus_encode_exact_size(&self) -> usize {
+        self.pow_algo.as_u64().required_space() + self.pow_data.consensus_encode_exact_size()
```
hansieodendaal: As above:
```suggestion
        1 + self.pow_data.consensus_encode_exact_size()
```

hansieodendaal (commented):
Hi, just a comment of encoding to u64 vs. a single byte for an enum.
Also, this might be a breaking change. _**Edit:** Confirmed, the base node is banning all peers, so the title needs to change._

Looking good otherwise.

delta1:
If this breaks sync on the current chain then what purpose does it serve?

SWvheerden:
We are going to need it if we want to implement any hashing in another language, but for now immediately, it's not required.

delta1:
Yeah if it requires a chain reset it can't go in now.

hansieodendaal (dismissed):
LGTM

base_layer/core/src/blocks/block_header.rs
```diff
@@ -243,12 +246,11 @@ impl From<NewBlockHeaderTemplate> for BlockHeader {
...
 impl Hashable for BlockHeader {
     fn hash(&self) -> Vec<u8> {
-        HashDigest::new()
-            .chain(self.merged_mining_hash())
-            .chain(self.pow.to_bytes())
-            .chain(self.nonce.to_le_bytes())
-            .finalize()
-            .to_vec()
+        let mut hasher = HashWriter::new(HashDigest::new());
+        self.merged_mining_hash().consensus_encode(&mut hasher).unwrap();
```
sdbondi: This is not quite right because we still use a dynamic length for hashes, so consensus encoding will prepend 0x20 (32) to the encoding but hashes are fixed length so don't need that byte.

To deal with this, I suggest the following annoying 😛  code: (change var names as needed)

```rust
let hash = self.merge_mining_hash();
let mut hash_buf = [0u8; 32];
hash_buf.copy_from_slice(&hash[..32]);
hash_buf.consensus_encode(&mut writer).unwrap();
```

This will omit the redundant length byte from the encoding

sdbondi (commented):
Looks good, just the superfluous length byte for the hash

sdbondi:
In order for the new hashing to work with the current dibbler chain, we should only use the consensus hashing for block version > 2

sdbondi:
Updated code to continue support for v2 blocks and start using (block template) v3 blocks from height ~~19000~~ 23000  

Manually tested syncing from current dibbler

stringhandler (approved):

PR #3826, opened 2022-02-11
feat(validator-node): initial state sync implementation (partial)

sdbondi:
Description
---
- add OpLogs
- add comms RPC methods and client calls to sync validator node state
- add synchronising state and basic implementation to validator node state machine (currently always in sync due to incorrect Merkle root in initial checkpoint)
- fix: fetching checkpoint from base node returned incorrect field for asset public key
- add read only interface for state db

Motivation and Context
---
Validator node on a committee needs to sync their state after going offline.

How Has This Been Tested?
---
Manually, it didnt work because the merkle root of the initial checkpoint is incorrect. Sync component has been hardcoded to always say it is in sync for now.


applications/tari_validator_node/proto/dan/validator_node.proto
```diff
@@ -74,7 +74,48 @@ message GetSidechainBlocksRequest {
...
+  uint64 height = 2;
+}
+
+message GetStateOpLogsResponse {
+  repeated StateOpLog op_logs = 1;
+}
+
+message StateOpLog {
+  uint64 height = 1;
+  string operation = 2;
```
stringhandler, line 108: perhaps an enum here rather?

dan_layer/core/src/services/asset_processor.rs
```diff
@@ -89,12 +89,12 @@ impl AssetProcessor for ConcreteAssetProcessor {
         )
     }
 
-    fn invoke_read_method<TUnifOfWork: StateDbUnitOfWork>(
+    fn invoke_read_method<TUnitOfWork: StateDbUnitOfWorkReader>(
```
stringhandler, line 92: Nice. Perhaps we need a better name, but I like it

stringhandler (dismissed):
Very nice

dan_layer/core/src/services/asset_processor.rs
sdbondi: Yeah it's a bit long, open to suggestions/change later

applications/tari_validator_node/proto/dan/validator_node.proto
sdbondi: Yeah probably, I'm using this "weirdly" as a single char enum - `S = Set, D = Delete` etc. but an enum is more standard :P

stringhandler (approved):

PR #3835, opened 2022-02-14
feat!: add committee management utxo

delta1:
Description
---

- adds the ability to submit committee definition transactions/UTXOs to the base layer for managing sidechain committees
- adds committee definition features to output features
- ~bumps the output features version to V1 (this is a hard fork as older software won’t accept the new version)~ edit: keep V0 as current output features version and add mempool validation and consensus rules for versions
- clean up some wallet client code in collectibles
- bump the default clippy `too many args` to 12 and remove individual `allow`s 
- hardcode more of the genesis block instead of using defaults that might change 
- impl Display for MmrRoots 

update: 16 feb
specific changes: https://github.com/tari-project/tari/pull/3835/commits/d34af75b292aebc0ca2aca2a48bd9876204465de 

- not a hard fork until we opt in to new output features version number (keep V0 for now) 
- adds acceptable versions to consensus constants 
- adds mempool validation for acceptable versions in transactions 
- fixes a bug where UnblindedOutput would use the `current_version` instead of the given utxo version


delta1:
Integration tests JS probably needs an update

applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
```diff
@@ -55,6 +55,15 @@ impl WalletClient {
     Ok(())
   }
 
+  fn get_inner(
```
sdbondi: nit: convention when returning mutable references
```suggestion
  fn get_inner_mut(
```

base_layer/core/tests/helpers/nodes.rs
```diff
@@ -393,7 +393,7 @@ async fn setup_comms_services(
 }
 
 // Helper function for starting the services of the Base node.
-#[allow(clippy::too_many_arguments)]
+
```
sdbondi:
```suggestion
```

base_layer/wallet/src/assets/asset_manager.rs
```diff
@@ -242,6 +242,34 @@ impl<T: OutputManagerBackend + 'static> AssetManager<T> {
...
+                    committee_pub_keys.clone(),
+                    effective_sidechain_height,
+                    true,
+                ),
+            )
+            .await?;
+
+        let (tx_id, transaction) = self
+            .output_manager
+            .create_send_to_self_with_output(vec![output], 100.into(), None, None)
```
sdbondi: Fee_per_gram of 100 is a lot - the fees scale with the amount of metadata, outputs, inputs etc by themselves, so typically increasing the fee per gram isn't necessary. I see `create_registration_transaction` also has a fpg is 20 - I'd say that 5 -10 should probably be used everywhere unless a user sets it higher.

sdbondi (dismissed):
Very nice! Think 100 FPG is a quite high for the checkpoint

base_layer/wallet/src/assets/asset_manager.rs
delta1: Thanks @sdbondi well spotted, definitely agree

applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
delta1: Thanks 💯

base_layer/core/src/proto/transaction.rs
```diff
@@ -300,7 +301,8 @@ impl TryFrom<proto::types::OutputFeatures> for OutputFeatures {
                 Some(m) => Some(m.try_into()?),
                 None => None,
             },
-            features.sidechain_checkpoint.map(|a| a.try_into()).transpose()?,
+            features.sidechain_checkpoint.map(|s| s.try_into()).transpose()?,
+            features.committee_checkpoint.map(|c| c.try_into()).transpose()?,
```
stringhandler:
```suggestion
            features.committee_management.map(|c| c.try_into()).transpose()?,
```

stringhandler (dismissed):
Happy, just would prefer a better name to `committee_checkpoint`. Checkpoint is used for the sidechain to checkpoint it's data to the base layer. Having two terms for checkpoint will probably be confusing. 

I suggested `committee_management` but it's not great either. Open to other suggestions

base_layer/core/src/proto/transaction.rs
delta1: Cool, what about `committee_definition`

base_layer/core/src/transactions/transaction_components/output_features_version.rs
```diff
@@ -17,7 +17,7 @@ pub enum OutputFeaturesVersion {
 
 impl OutputFeaturesVersion {
     pub fn get_current_version() -> Self {
-        Self::V1
+        Self::V0
```
delta1: keep V0 for now until we opt in

base_layer/core/src/transactions/transaction_components/transaction_input_version.rs
```diff
@@ -8,10 +8,11 @@ use serde::{Deserialize, Serialize};
 
 use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized};
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, PartialOrd)]
 #[repr(u8)]
 pub enum TransactionInputVersion {
     V0 = 0,
+    V1 = 1,
```
delta1, line 15: This was added for testing and is unused for now

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -196,7 +196,8 @@ impl UnblindedOutput {
             ));
         }
         let commitment = factories.commitment.commit(&self.spending_key, &self.value.into());
-        let output = TransactionOutput::new_current_version(
+        let output = TransactionOutput::new(
+            self.version.clone(),
```
delta1: I think this was a bug, as calling these functions would overwrite the given utxo version with whatever `get_current_version` would return

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -242,7 +243,8 @@ impl UnblindedOutput {
                 .map_err(|_| TransactionError::RangeProofError(RangeProofError::ProofConstructionError))?
         };
 
-        let output = TransactionOutput::new_current_version(
+        let output = TransactionOutput::new(
+            self.version.clone(),
```
delta1: I think this was a bug, as calling these functions would overwrite the given utxo version with whatever `get_current_version` would return

delta1:
Just need to update some tests after the change back to V0

sdbondi (approved):
Not tested but looking good

stringhandler (approved):

PR #3837, opened 2022-02-14
feat: update console wallet tui

hansieodendaal:
Description
---
Updated console wallet TUI to better accommodate the expanding number of tabs:
- Moved base node status to the bottom of the screen, expanding tab selection to
  the entire width.
- Regained 1 line of real estate screen space at the bottom of the screen
  (with menu).
- Refined some help text to fit on the screen.
- Added a separate contacts tab to enable better handling of contacts not
  needing to access it via the transaction send tab.
- Improved column spacing on the contacts tab/pane.
- Improved the receive tab layout (moved connection details above the QR code).
- Removed contacts editing ability from the sent tab; only selection remains.

_**Some screenshots:**_

image

image

image

Motivation and Context
---
See above

How Has This Been Tested?
---
System level tests

### DIFF
--- a/applications/tari_console_wallet/src/ui/app.rs
+++ b/applications/tari_console_wallet/src/ui/app.rs
@@ -36,6 +36,7 @@ use crate::{
         components::{
             assets_tab::AssetsTab,
             base_node::BaseNode,
+            contacts_tab::ContactsTab,
             events_component::EventsComponent,
             log_tab::LogTab,
             menu::Menu,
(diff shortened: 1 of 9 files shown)


delta1:
Nice! Will take a look. I personally think that Events and Logs should be hidden and then enabled with some kind of “dev” or “debug” mode/key (but that’s a separate change)

hansieodendaal:
> Nice! Will take a look. I personally think that Events and Logs should be hidden and then enabled with some kind of “dev” or “debug” mode/key (but that’s a separate change)

Thanks and agreed, that should be a separate PR.

philipr-za (approved):
Tested, looks good

Issue #3675, opened 2021-12-29
Broken links in RFC description


AaronFeickert:
Some reference links for the RFC lifecycle are broken.

The links are here:
- https://github.com/tari-project/tari/blob/development/RFC/src/about.md?plain=1#L12
- https://github.com/tari-project/tari/blob/development/RFC/src/about.md?plain=1#L15

PR #3839, opened 2022-02-15
fix: update RFC links and README (#3675)

zhangcheng:
Description
---
Update in light of #3675.
For reason unknown to me, the link to RFC_template.md also has to appear in
SUMMARY.md. Otherwise, it will be a 404 as reported (somehow the HTML file is
not generated).

Motivation and Context
---
See above.

How Has This Been Tested?
---
Manual check.


sdbondi (approved):
Thanks LGTM

zhangcheng:
Just some doc text changed, it shouldn't cause CI test failure, right?

delta1:
> Just some doc text changed, it shouldn't cause CI test failure, right?

No I think it's unrelated

PR #3841, opened 2022-02-15
feat: adds get-mempool-tx command

therustmonk:
Description
---
Adds `get-mempool-state-tx` command that prints detailed information of a transaction.
Also, this PR refactors the commands parser to remove unnecessary type parameters from methods, simplify the code and make easier further switching to crates like `clap` or `nom`.

Motivation and Context
---
For debugging purposes, we need a command to get the full details of a transaction.

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_base_node/src/command_handler.rs
+++ b/applications/tari_base_node/src/command_handler.rs
@@ -24,6 +24,7 @@ use std::{
     cmp,
     fs::File,
     io::{self, Write},
+    str::FromStr,
     string::ToString,
     sync::Arc,
     time::{Duration, Instant},
@@ -62,6 +63,7 @@ use tari_p2p::{
     services::liveness::{LivenessEvent, LivenessHandle},
 };
 use tari_utilities::{hex::Hex, message_format::MessageFormat, Hashable};
+use thiserror::Error;
 use tokio::{
     runtime,
     sync::{broadcast, watch},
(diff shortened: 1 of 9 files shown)


applications/tari_base_node/src/command_handler.rs
```diff
@@ -376,11 +376,50 @@ impl CommandHandler {
     }
 
     /// Function to process the get-mempool-state command
-    pub fn get_mempool_state(&self) {
+    pub fn get_mempool_state(&self, full: bool, filter: Option<String>) {
```
sdbondi: Maybe instead of passing in `full`, you could show the transaction body if filter is Some and the Display impl of StateResponse if filter is None

applications/tari_base_node/src/parser.rs
```diff
@@ -747,3 +740,76 @@ impl Parser {
...
+
+#[derive(Debug, Error)]
+enum ArgsReason {
+    #[error("argument required")]
+    Required,
+    #[error("argument can't be parsed: {details}")]
+    NotParsed { details: String },
+}
+
+struct Args<'a> {
```
sdbondi: Great idea!

applications/tari_base_node/src/parser.rs
```diff
@@ -747,3 +740,76 @@ impl Parser {
...
+            .and_then(parse_emoji_id_or_public_key_or_node_id)
+            .map(either_to_node_id)
+    }
+
+    fn try_take_next<T>(&mut self, name: &'static str) -> Result<Option<T>, ArgsError>
+    where
+        T: FromStr,
+        T::Err: ToString,
+    {
+        match self.splitted.peek().map(|s| s.parse()) {
```
sdbondi: Why `peek` and not `next`? Peek wont progress the iterator so try_take_next will always return the same value

sdbondi (commented):
Nice idea RE the args, much better - some comments.

applications/tari_base_node/src/command_handler.rs
sdbondi: Reasoning, we probably dont want to display all of the transaction bodies as that will just scroll off the screen - only one at a time

applications/tari_base_node/src/parser.rs
```diff
@@ -747,3 +740,76 @@ impl Parser {
         self.command_handler.lock().await.list_reorgs();
     }
 }
+
+use std::str::SplitWhitespace;
```
sdbondi: This should move into it's own file and public functions marked as `pub`

stringhandler (approved):

PR #3842, opened 2022-02-15
feat: add persistence of transaction cancellation reason to wallet db

philipr-za:
Description
---
This PR adds the persistence of the reason a transaction was rejected or cancelled to the db. This is done by expanding the existing `cancelled` field on a CompletedTransaction too not just be a boolean but an integer encoded nullable Enum which will encode the reason for the cancellation. By expanding the information contained by this field we no longer need the `valid` field which was used to indicate when a Coinbase Transaction was abandoned and we just make that a cancellation reason. A diesel migration is provided to handle this change is db schema.

Some specific areas changed to handle the update:
1. The console wallet was updated to take this change into account when displaying transactions.
2. The transaction validation protocol was updated to handle coinbases using this field rather than the valid field
3. Transaction database methods were updated to take the change into account

In the FFI we do the following:
1. Remove the `completed_transaction_is_valid` method for the valid field that is taken out.
2. Added the `completed_transaction_get_cancellation_reason` method that returns whether the cancelled field is null as -1 or else the integer encoding of the TxCancellationReason enum.

```C
/// Gets the reason a TariCompletedTransaction is cancelled, if it is indeed cancelled
///
/// ## Arguments
/// `tx` - The TariCompletedTransaction
/// `error_out` - Pointer to an int which will be modified to an error code should one occur, may not be null. Functions
/// as an out parameter.
///
/// ## Returns
/// `c_int` - Returns the reason for cancellation which corresponds to:
/// | Value | Interpretation    |
/// |---  |---                  |
/// |  -1 | Not Cancelled       |
/// |   0 | Unknown             |
/// |   1 | UserCancelled       |
/// |   2 | Timeout             |
/// |   3 | DoubleSpend         |
/// |   4 | Orphan              |
/// |   5 | TimeLocked          |
/// |   6 | InvalidTransaction  |
/// |   7 | AbandonedCoinbase   |
/// # Safety
/// None
int completed_transaction_get_cancellation_reason(struct TariCompletedTransaction *transaction, int *error_out);
```

How Has This Been Tested?
---
Tests are updated to take this change into account.


applications/tari_app_grpc/proto/wallet.proto
```diff
@@ -166,7 +166,6 @@ message TransactionInfo {
     bytes excess_sig = 9;
     google.protobuf.Timestamp timestamp = 10;
     string message = 11;
-    bool valid = 12;
```
delta1, line 169: minor but maybe worth commenting this out instead to show that `12` is still reserved

applications/tari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -235,14 +239,12 @@ impl TransactionsTab {
                 format!("{}", local_time.format("%Y-%m-%d %H:%M:%S")),
                 Style::default().fg(text_color),
             )));
-            let status = if (t.cancelled || !t.valid) && t.status == TransactionStatus::Coinbase {
+            let status = if matches!(t.cancelled, Some(TxCancellationReason::AbandonedCoinbase)) {
```
SWvheerden, line 242: Why not do something akin to this:
```
let status_msg = if let Some(reason) = tx.cancelled {
                format!("Cancelled: {}", reason)
            } else {
                tx.status.to_string()
            };
```

applications/tari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -396,9 +395,9 @@ impl TransactionsTab {
             );
             let excess = Span::styled(tx.excess_signature.as_str(), Style::default().fg(Color::White));
             let confirmation_count = app_state.get_confirmations(&tx.tx_id);
-            let confirmations_msg = if tx.status == TransactionStatus::MinedConfirmed && !tx.cancelled {
+            let confirmations_msg = if tx.status == TransactionStatus::MinedConfirmed && tx.cancelled.is_none() {
```
SWvheerden, line 398: Nit: a match here would read easier than a nested if. 
Something like:
```
|match (tx.status, confirmation_count, tx.cancelled){
(TransactionStatus::MinedConfirmed, _,None) => ...
...
}
```

SWvheerden (approved):
LGTM
I just have a NIT about the nested if's. But they are fine as is.

delta1 (approved):

applications/tari_app_grpc/proto/wallet.proto
philipr-za: Hmm, that is true. Def how it should be done in the future though I don't think its a major issue at the moment as this is just used by the cucumber at this point.

stringhandler (approved):

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -334,3 +343,50 @@ impl From<WalletTransaction> for CompletedTransaction {
...
+pub enum TxCancellationReason {
+    Unknown,            // 0
+    UserCancelled,      // 1
+    Timeout,            // 2
+    DoubleSpend,        // 3
+    Orphan,             // 4
+    TimeLocked,         // 5
+    InvalidTransaction, // 6
+    AbandonedCoinbase,  // 7
+}
```
hansieodendaal, line 357: Discussion point for future reference:
- It might be cool to provide a bit more information to the user here, like the value of the time out or identifying which UTXO is implicated in the double spend when the culprit is just one of many dust inputs.

hansieodendaal (approved):
LGTM

Just a comment for future discussion.

PR #3843, opened 2022-02-16
ci: cargo test speedups

delta1:
Description
---

Reduces CI test time to approx 30 mins on cache hit, 40ish mins on cache miss

- caching: use [Swatinem/rust-cache](/Swatinem/rust-cache) for smarter caching 
- test: split test into compile and test run, hopefully gets rid of our `SIGILL` problems 
- remove test dependency on check nightly step since the cache was missed anyway 
- run tests in debug as it builds faster 

With thanks to https://matklad.github.io/2021/09/04/fast-rust-builds.html for the ideas


leet4tari:
@delta1 nice one! LGTM

stringhandler (approved):

PR #3844, opened 2022-02-17
feat(validator-node): committee proposes genesis block w/ instructions

sdbondi:
Description
---
- committee proposes genesis block/node with initial instructions as per template id
- state db is aware of which asset it is processing
- minor cleanup of some generics
- error handling

Motivation and Context
---
The committee can propose a genesis block with init instructions and agree on a genesis block and genesis merkle root

How Has This Been Tested?
---
Manually


stringhandler (approved):
Awesome. I would use the TSpecification instead of TOutboundService in future, but not a problem now

PR #3855, opened 2022-02-18
fix(validator-node): fix consensus stall after genesis

sdbondi:
Description
---
- Fixes consensus "stall" after genesis
- Clean up state generics using ServiceSpecification 

Motivation and Context
---
After genesis consensus would stay on ViewId(1), this has been fixed

How Has This Been Tested?
---
Manually, minting instructions processed


delta1 (approved):
LGTM

PR #3856, opened 2022-02-18
chore: consolidate fee_per_gram

SWvheerden:
Description
---
Consolidated the fee_per_gram to read from config file. This is used throughout the wallet where fee is not provided by the user. 

How Has This Been Tested?
---
Manual


applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -457,13 +454,14 @@ pub async fn make_it_rain(
                 }
                 let delayed_for = Instant::now();
                 let sender_clone = sender.clone();
+                let inner_config_clone = config_clone.clone();
```
philipr-za: I don't think you should clone the whole wallet config object here as it contains the cryptofactories which are quite large. You only need the fee config for these operations so I think rather specifically send that through into the `send_tari` and `send_one_sided` tasks below.

SWvheerden: Its not the complete config, that config only contains the `fee_per_gram `and `grpc address`. But Will modify to only take the fee_per_gram

philipr-za (approved):

PR #3857, opened 2022-02-19
feat: add contacts liveness service to base layer wallet

hansieodendaal:
Description
---
Implemented a liveness service for the base layer wallet's contacts that will send ping messages and respond with pong messages to indicate liveness information.

The next aspect of contacts liveness will be to integrate the events into the various wallet front ends.

Motivation and Context
---
There is a need to know which managed contacts in a wallet are online and when they have last been seen.

How Has This Been Tested?
---
- ~~**[WIP]**~~ Additional unit tests 
- System level testing between various console wallets


base_layer/wallet/src/config.rs
```diff
@@ -45,6 +45,7 @@ pub struct WalletConfig {
     pub base_node_service_config: BaseNodeServiceConfig,
     pub updater_config: Option<AutoUpdateConfig>,
     pub autoupdate_check_interval: Option<Duration>,
+    pub contacts_auto_ping_interval: u64,
```
philipr-za: Make this a Duration, a u64 is ambiguous. It could mean seconds or minutes or milliseconds but a Duration is explicit.

base_layer/wallet/src/config.rs
```diff
@@ -71,6 +73,7 @@ impl WalletConfig {
             base_node_service_config: base_node_service_config.unwrap_or_default(),
             updater_config,
             autoupdate_check_interval,
+            contacts_auto_ping_interval: contacts_auto_ping_interval.unwrap_or(90),
```
philipr-za: Is this 90 seconds? I think we should be more down in the 10-30 seconds range.

base_layer/wallet/src/contacts_service/service.rs
```diff
@@ -105,27 +154,192 @@ where T: ContactsBackend + 'static
...
+    }
+
+    async fn send_network_silence(&mut self) -> Result<(), ContactsServiceError> {
+        let _ = self
+            .event_publisher
+            .send(Arc::new(ContactsLivenessEvent::NetworkSilence));
+        Ok(())
+    }
+
+    fn resize_contacts_liveness_data_buffer(&mut self, n: usize) {
```
philipr-za, line 316: This feels like something that the Vec will do on its own?

philipr-za (commented):
Looking good so far

base_layer/wallet/src/config.rs
hansieodendaal: I thought the clients (via config) can decide what the duration should be to conserve mobile client's battery life.

base_layer/wallet/src/contacts_service/service.rs
hansieodendaal: Added a proper comment

base_layer/wallet/src/config.rs
philipr-za: It should def be configurable but here you are essentially choosing a default for when it's not explicitly set

hansieodendaal: Yes, I think it should be less frequent as a default. That being said, it is easy to change, so just say the word.

philipr-za: 90 seconds is so long, In a chat app would you want a latency of a full minute and a half before you are aware that a user is no longer online?

hansieodendaal: Ok, changed the default setting to 20s.

philipr-za (approved):

PR #3862, opened 2022-02-21
chore: update README's "Conversation channels" part

zhangcheng:
Description
---
Point to the relevant part of the website, so DRY with outdated information (such as IRC part).


delta1 (approved):
Thanks @zhangcheng

PR #3863, opened 2022-02-21
feat: add logging of cancelled outputs when transaction is rejected

philipr-za:
Description
---
This PR just adds some logging of which outputs were specifically cancelled when a transaction is reject by the base node. In addition when an output is marked as Unspent we also clear the Mined In Block field which will result in that output being revalidated the next time a validation is done.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
@@ -873,7 +873,7 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
 
         for output in outputs.iter() {
             if output.received_in_tx_id == Some(i64::from(tx_id)) {
-                debug!(
+                info!(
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -881,10 +888,18 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
                     &conn,
                 )?;
             } else if output.spent_in_tx_id == Some(i64::from(tx_id)) {
+                debug!(
```
delta1: maybe warn or info level? do these happen often?

philipr-za: You are right, on the transaction service side when a transaction is rejected the message is a warn. However these will fire when you do a manual cancellation so I think I will make these Info rather.

delta1 (approved):

PR #3869, opened 2022-02-23
chore: fix cargo udeps warnings in tari_metrics

sdbondi:
Description
---
- adds `optional` attribute to some dependencies in tari_metrics that are only used with certain features.
- `cargo udeps` no longer gives any warnings about unused dependencies.
- moves openssl-sys vendoring dependency to from comms to tari_p2p (used by tari_p2p dns-seed implementation) 

Motivation and Context
---
`cargo udeps` passes

How Has This Been Tested?
---
cargo +nightly udeps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,22 +122,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "antidote"
-version = "1.0.0"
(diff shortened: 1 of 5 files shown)


delta1 (approved):
LGTM

PR #3879, opened 2022-02-28
feat: key_management service

SWvheerden:
Description
---
This adds in a completely separate key management service to manage all keys with branches. 
Before this the wallet OMS used key manager inside of itself with only the spend branch's index saved to db. 
The key management service supports infinite branches and all will be saved to db. 

Motivation and Context
---
We need to keep track of all generated private key's as well s their branches. These keys will not only be used inside of the OMS, but also in other services. 

How Has This Been Tested?
---
Unit, integration, and manual tests.


base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -0,0 +1,110 @@
...
+    async fn apply_encryption(&self, cipher: Aes256Gcm) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner).write().await.apply_encryption(cipher).await
+    }
+
+    async fn remove_encryption(&self) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner).write().await.remove_encryption().await
+    }
+
+    async fn get_next_key(&self, branch: String) -> Result<PrivateKey, KeyManagerError> {
+        (*self.key_manager_inner).read().await.get_next_key(branch).await
```
philipr-za: should probably be a write() lock. Yes the interior mutability means this works but the process of getting a key does write a new index value.

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -0,0 +1,110 @@
...
+            .await
+    }
+
+    async fn find_key_index(&self, branch: String, key: &PrivateKey) -> Result<u64, KeyManagerError> {
+        (*self.key_manager_inner).read().await.find_key_index(branch, key).await
+    }
+
+    async fn update_current_key_index_if_higher(&self, branch: String, index: u64) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner)
+            .read()
```
philipr-za, line 108: think this should be a write lock

base_layer/wallet/src/key_manager_service/mock.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::{key_manager_service::KeyManagerInterface, types::KeyDigest};
+
+const LOG_TARGET: &str = "wallet::Key_manager_mock";
+const KEY_MANAGER_MAX_SEARCH_DEPTH: u64 = 1_000_000;
+use std::{collections::HashMap, sync::Arc};
+
+use crate::key_manager_service::{error::KeyManagerError, storage::database::KeyManagerState};
+
+#[derive(Clone)]
+pub struct KeyManagerMock {
```
philipr-za, line 45: Should probably in in the /tests/support folder

base_layer/wallet/src/key_manager_service/service.rs
```diff
@@ -0,0 +1,175 @@
...
+        KeyManagerInner {
+            key_managers: HashMap::new(),
+            db,
+            master_seed,
+        }
+    }
+
+    pub async fn add_key_manager(&mut self, branch: String) -> Result<(), KeyManagerError> {
+        if self.key_managers.contains_key(&branch) {
+            return Err(KeyManagerError::BranchAllreadyExists);
```
philipr-za:
```suggestion
            return Err(KeyManagerError::BranchAlreadyExists);
```

philipr-za: Though actually I think in these cases its better just to return `Ok(())`. The client asked for the branch to be created, it exists. The state they wanted is what exists.

base_layer/wallet/src/key_manager_service/service.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    pub async fn get_key_at_index(&self, branch: String, index: u64) -> Result<PrivateKey, KeyManagerError> {
+        let km = self
+            .key_managers
+            .get(&branch)
+            .ok_or(KeyManagerError::UnknownKeyBranch)?
+            .lock()
+            .await;
+        let key = km.derive_key(index)?;
+        self.db.set_key_index(branch, index).await?;
```
philipr-za: This feels dangerous, a client can destroy state here. I feel like we should have 2 distinct types of branches. managed indexing and manually indexing and you can only ask for a specific key from the latter type. This protects a client from inadvertly destroying the index state of a branch they think has managed index.

base_layer/wallet/src/key_manager_service/service.rs
```diff
@@ -0,0 +1,175 @@
...
+        Ok(())
+    }
+
+    pub async fn remove_encryption(&self) -> Result<(), KeyManagerError> {
+        self.db.remove_encryption().await?;
+        Ok(())
+    }
+
+    /// Return the Seed words for the current Master Key set in the Key Manager
+    pub async fn get_seed_words(
```
philipr-za, line 128: This feels like it actually belongs in the main Wallet DB where the master key is actually stored.

base_layer/wallet/src/key_manager_service/storage/database/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Holds the state of the KeyManager being used by the Output Manager Service
+#[derive(Clone, Debug, PartialEq)]
+pub struct KeyManagerState {
+    pub branch_seed: String,
+    pub primary_key_index: u64,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum DbKey {
```
philipr-za: We have been moving away from this interface. Just have a clear methods in the Database interface trait, the DbKey, Db Value, DbKeyValuePair and WriteOperation are just clunky abstractions that we don't really need.

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -807,9 +809,19 @@ pub async fn command_runner(
...
-                let mut rng = rand::thread_rng();
-                // todo add key from key manager service here
-                let (_, public_key) = PublicKey::random_keypair(&mut rng);
+                let key_manager = wallet.key_manager_service.clone();
+                match key_manager
+                    .add_new_branch(KeyManagerAssetBranch::Asset.to_string())
+                    .await
+                {
+                    Ok(()) => Ok(()),
+                    Err(KeyManagerError::BranchAllreadyExists) => Ok(()),
```
delta1: "Already" is spelled incorrectly in your error variant. 

But why does `add_new_branch` return an error if the branch already exists?

base_layer/wallet/src/assets/asset_manager.rs
```diff
@@ -368,3 +370,17 @@ pub struct AssetMetadata {
     description: String,
     image: String,
 }
+
+#[derive(Clone, Copy)]
+pub enum KeyManagerAssetBranch {
+    Asset,
```
delta1: IMO this is unnecessary, just use the string itself or a const

base_layer/wallet/src/error.rs
```diff
@@ -95,6 +96,8 @@ pub enum WalletError {
     UtxoScannerError(#[from] UtxoScannerError),
     #[error("Key manager error: `{0}`")]
     KeyManagerError(#[from] KeyManagerError),
+    #[error("Key service manager error: `{0}`")]
```
delta1:
```suggestion
    #[error("Key manager service error: `{0}`")]
```

base_layer/wallet/src/key_manager_service/error.rs
```diff
@@ -0,0 +1,88 @@
...
+use tari_utilities::{hex::HexError, ByteArrayError};
+
+use crate::error::WalletStorageError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum KeyManagerError {
+    #[error("Branch does not exist")]
+    UnknownKeyBranch,
+    #[error("Branch all ready exists")]
+    BranchAllreadyExists,
```
delta1:
```suggestion
    #[error("Branch already exists")]
    BranchAlreadyExists,
```

**But if this doesn't indicate an actual error then I don't think it should be necessary**

base_layer/wallet/src/key_manager_service/error.rs
```diff
@@ -0,0 +1,88 @@
...
+    TariKeyManagerError(#[from] KMError),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum KeyManagerStorageError {
+    #[error("Value not found")]
+    ValueNotFound,
+    #[error("Unexpected result: `{0}`")]
+    UnexpectedResult(String),
+    #[error("If an pending transaction does not exist to be confirmed")]
```
delta1:
```suggestion
    #[error("Pending transaction does not exist to be confirmed")]
```

base_layer/wallet/src/key_manager_service/error.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[error("This write operation is not supported for provided DbKey")]
+    OperationNotSupported,
+    #[error("Could not find all values specified for batch operation")]
+    ValuesNotFound,
+    #[error("Error converting a type: {reason}")]
+    ConversionError { reason: String },
+    #[error("Key Manager not initialized")]
+    KeyManagerNotInitialized,
+    #[error("Diesel R2d2 error: `{0}`")]
+    DieselR2d2Error(#[from] WalletStorageError),
```
delta1: Does `WalletStorageError` only ever produce an R2D2 error? This doesn't seem quite right

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    pub fn increment_index(
+        branch: String,
+        cipher: Option<&Aes256Gcm>,
+        conn: &SqliteConnection,
+    ) -> Result<u64, KeyManagerStorageError> {
+        Ok(match KeyManagerStateSql::get_state(&branch, conn) {
+            Ok(mut km) => {
+                if km.primary_key_index.len() > 8 {
+                    match cipher {
```
philipr-za: Look at the other wallet storage layers where we DRYed this check up using the `decrypt_if_necessary` method

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -0,0 +1,225 @@
...
+                        Some(cip) => km.decrypt(cip).map_err(KeyManagerStorageError::AeadError)?,
+                        _ => return Err(KeyManagerStorageError::ValueEncrypted),
+                    }
+                }
+                let mut bytes: [u8; 8] = [0u8; 8];
+                bytes.copy_from_slice(&km.primary_key_index[..8]);
+                let new_index = u64::from_le_bytes(bytes) + 1;
+                km.primary_key_index = new_index.to_le_bytes().to_vec();
+                if let Some(cip) = cipher {
```
philipr-za: In other wallet storage layers we used the `encrypt_if_necessary` to DRY this up

base_layer/wallet/src/key_manager_service/service.rs
```diff
@@ -0,0 +1,175 @@
...
+use tari_common_types::types::PrivateKey;
+use tari_key_manager::{
+    cipher_seed::CipherSeed,
+    key_manager::KeyManager,
+    mnemonic::{Mnemonic, MnemonicLanguage},
+};
+
+use crate::types::KeyDigest;
+
+const LOG_TARGET: &str = "wallet::Key_manager";
```
delta1:
```suggestion
const LOG_TARGET: &str = "wallet::key_manager";
```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub fn set_index(
+        branch: String,
+        index: u64,
+        cipher: Option<&Aes256Gcm>,
+        conn: &SqliteConnection,
+    ) -> Result<(), KeyManagerStorageError> {
+        match KeyManagerStateSql::get_state(&branch, conn) {
+            Ok(mut km) => {
+                km.primary_key_index = index.to_le_bytes().to_vec();
+                if let Some(cip) = cipher {
```
philipr-za: DRYed up as described above

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn decrypt(&mut self, _cipher: &Aes256Gcm) -> Result<(), String> {
+        unimplemented!("Not supported")
+        // let decrypted_master_key = decrypt_bytes_integral_nonce(&cipher, self.master_key.clone())?;
+        // let decrypted_branch_seed =
+        //     decrypt_bytes_integral_nonce(&cipher, from_hex(self.branch_seed.as_str()).map_err(|_| Error)?)?;
+        // self.master_key = decrypted_master_key;
+        // self.branch_seed = from_utf8(decrypted_branch_seed.as_slice())
+        //     .map_err(|_| Error)?
+        //     .to_string();
+        // Ok(())
```
philipr-za:
```suggestion

```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state_old.rs
```diff
@@ -0,0 +1,50 @@
...
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use chrono::NaiveDateTime;
+use diesel::{prelude::*, SqliteConnection};
+
+use crate::{key_manager_service::error::KeyManagerStorageError, schema::key_manager_states_old};
+
+// NB!!! This is a temporary migration file to convert existing indexes to new ones.
```
philipr-za:
```suggestion
// This is a temporary migration file to convert existing indexes to new ones.
```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state_old.rs
```diff
@@ -0,0 +1,50 @@
...
+// Todo Testnet remove
```
philipr-za:
```suggestion
// TODO Remove at next Testnet reset (currently on Dibbler)
```
Lets be clearer and less dramatic

base_layer/wallet/src/key_manager_service/service.rs
delta1: I really don't think this is an error state? This fn could just return an informational enum variant indicating `BranchCreated` or `BranchExists`

base_layer/wallet/src/key_manager_service/storage/database/backend.rs
```diff
@@ -0,0 +1,45 @@
...
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+// USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+use aes_gcm::Aes256Gcm;
+
+use crate::key_manager_service::{
+    error::KeyManagerStorageError,
+    storage::database::{DbKey, DbValue, WriteOperation},
+};
+
+/// This trait defines the required behaviour that a storage backend must provide for the Output Manager service.
```
delta1:
```suggestion
/// This trait defines the required behaviour that a storage backend must provide for the Key Manager service.
```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+    fn encrypt_if_necessary<T: Encryptable<Aes256Gcm>>(&self, o: &mut T) -> Result<(), KeyManagerStorageError> {
+        let cipher = acquire_read_lock!(self.cipher);
+        if let Some(cipher) = cipher.as_ref() {
+            o.encrypt(cipher)
+                .map_err(|_| KeyManagerStorageError::AeadError("Encryption Error".to_string()))?;
+        }
+        Ok(())
+    }
+
+    fn insert(&self, key_value_pair: DbKeyValuePair, conn: &SqliteConnection) -> Result<(), KeyManagerStorageError> {
```
philipr-za: Def think we have shown that this pattern is clunky. Check the newer methods in Transaction Service, its just easier to read and maintain to have explicit methods in the database backend trait.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+            DbKeyValuePair::KeyManagerState(km) => {
+                let mut km_sql = NewKeyManagerStateSql::from(km);
+                self.encrypt_if_necessary(&mut km_sql)?;
+                km_sql.commit(conn)?
+            },
+        }
+        Ok(())
+    }
+
+    fn do_migration(&self) -> Result<(), KeyManagerStorageError> {
```
philipr-za, line 91: Add a TODO: Remove in next testnet reset (current testnet: Dibbler)

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+            self.insert(spending_script_km, &conn)?;
+            KeyManagerStateSqlOld::delete(&conn)?;
+        }
+        Ok(())
+    }
+}
+
+impl KeyManagerBackend for KeyManagerSqliteDatabase {
+    #[allow(clippy::cognitive_complexity)]
+    fn fetch(&self, key: &DbKey) -> Result<Option<DbValue>, KeyManagerStorageError> {
```
philipr-za: as above, this pattern of abstracting this operation should just be separate explicit methods.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+                acquire_lock.as_millis(),
+                (start.elapsed() - acquire_lock).as_millis(),
+                start.elapsed().as_millis()
+            );
+        }
+
+        Ok(result)
+    }
+
+    fn write(&self, op: WriteOperation) -> Result<Option<DbValue>, KeyManagerStorageError> {
```
philipr-za: as with `fetch`

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
philipr-za: Which sort of implies this logic should live one layer higher in the KeyManagerSqliteDatabase logic.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+
+        Ok(None)
+    }
+
+    fn increment_key_index(&self, branch: String) -> Result<(), KeyManagerStorageError> {
+        let start = Instant::now();
+        let conn = self.database_connection.get_pooled_connection()?;
+        let acquire_lock = start.elapsed();
+        let cipher = acquire_read_lock!(self.cipher);
+        KeyManagerStateSql::increment_index(branch, cipher.as_ref(), &conn)?;
```
philipr-za: this logic should be here.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -128,10 +138,23 @@ where TBackend: OutputManagerBackend + 'static
...
 
+            let rewind_key = self
+                .master_key_manager
+                .get_key_at_index(KeyManagerOmsBranch::RecoveryViewOnly.to_string(), 0)
+                .await?;
+            let rewind_blinding_key = self
+                .master_key_manager
+                .get_key_at_index(KeyManagerOmsBranch::RecoveryBlinding.to_string(), 0)
+                .await?;
+            let rewind_data = RewindData {
```
philipr-za: Do this once when the recoverer is created. Here you are recalculating these keys for each output being recovered.

base_layer/wallet/src/output_manager_service/resources.rs
```diff
@@ -20,27 +20,54 @@
...
-    pub master_key_manager: Arc<MasterKeyManager<TBackend>>,
+    pub master_key_manager: TKeyManagerInterface,
     pub consensus_constants: ConsensusConstants,
     pub connectivity: TWalletConnectivity,
     pub shutdown_signal: ShutdownSignal,
+    pub rewind_data: RewindData,
+}
+
+#[derive(Clone, Copy)]
+pub enum KeyManagerOmsBranch {
```
philipr-za:
```suggestion
pub enum OutputManagerKeyManagerBranch {
```

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
philipr-za: In fact below I see you do calculate this data and put it in the resources...why calc it here?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -118,24 +126,38 @@ where
         shutdown_signal: ShutdownSignal,
         base_node_service: BaseNodeServiceHandle,
         connectivity: TWalletConnectivity,
-        master_seed: CipherSeed,
         node_identity: Arc<NodeIdentity>,
+        key_manager: TKeyManagerInterface,
     ) -> Result<Self, OutputManagerError> {
         // Clear any encumberances for transactions that were being negotiated but did not complete to become official
         // Pending Transactions.
+        OutputManagerService::<TBackend, TWalletConnectivity, TKeyManagerInterface>::initialise_key_manager(
```
philipr-za: You have put an unrelated operation between a comment and the line the comment is discussing...

base_layer/wallet/src/key_manager_service/storage/database/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use aes_gcm::Aes256Gcm;
+pub use backend::KeyManagerBackend;
+use log::*;
+
+use crate::key_manager_service::error::KeyManagerStorageError;
+
+const LOG_TARGET: &str = "wallet::key_manager_service::database";
+
+/// Holds the state of the KeyManager being used by the Output Manager Service
```
delta1: s/output/key/

base_layer/wallet/src/key_manager_service/storage/database/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+pub enum DbKeyValuePair {
+    KeyManagerState(KeyManagerState),
+}
+
+pub enum WriteOperation {
+    Insert(DbKeyValuePair),
+    Remove(DbKey),
+}
+
+/// This structure holds an inner type that implements the `OutputManagerBackend` trait and contains the more complex
```
delta1: s/output/key/

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl Encryptable<Aes256Gcm> for NewKeyManagerStateSql {
+    fn encrypt(&mut self, cipher: &Aes256Gcm) -> Result<(), String> {
+        let encrypted_index = encrypt_bytes_integral_nonce(cipher, self.primary_key_index.clone())?;
+        self.primary_key_index = encrypted_index;
+        Ok(())
+    }
+
+    fn decrypt(&mut self, _cipher: &Aes256Gcm) -> Result<(), String> {
+        unimplemented!("Not supported")
```
delta1, line 167: I guess rather just return this as an `Err` instead of panicing

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -147,6 +169,58 @@ where
         })
     }
 
+    async fn initialise_key_manager(key_manager: &TKeyManagerInterface) -> Result<(), OutputManagerError> {
+        OutputManagerService::<TBackend, TWalletConnectivity, TKeyManagerInterface>::initialise_key_manager_branch(
```
philipr-za:
```suggestion
        Self::initialise_key_manager_branch(
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -147,6 +169,58 @@ where
...
+        .await
+    }
+
+    async fn initialise_key_manager_branch(
+        branch: String,
+        key_manager: &TKeyManagerInterface,
+    ) -> Result<(), OutputManagerError> {
+        match key_manager.add_new_branch(branch).await {
+            Ok(()) => Ok(()),
+            Err(KeyManagerError::BranchAllreadyExists) => Ok(()),
```
philipr-za: As per my previous comment, this doesn't feel like it should be an error.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+            self.insert(spending_km, &conn)?;
+            self.insert(spending_script_km, &conn)?;
+            KeyManagerStateSqlOld::delete(&conn)?;
+        }
+        Ok(())
+    }
+}
+
+impl KeyManagerBackend for KeyManagerSqliteDatabase {
+    #[allow(clippy::cognitive_complexity)]
```
delta1: is this clippy allow necessary? doesn't seem like an overly complex fn

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    async fn get_next_key(&self, branch: String) -> Result<PrivateKey, KeyManagerError> {
+        (*self.key_manager_inner).read().await.get_next_key(branch).await
+    }
+
+    async fn get_key_at_index(&self, branch: String, index: u64) -> Result<PrivateKey, KeyManagerError> {
+        (*self.key_manager_inner)
+            .read()
+            .await
+            .get_key_at_index(branch, index)
```
philipr-za:
```suggestion
            .get_key_at_index(branch.to_string(), index)
```

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    async fn remove_encryption(&self) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner).write().await.remove_encryption().await
+    }
+
+    async fn get_next_key(&self, branch: String) -> Result<PrivateKey, KeyManagerError> {
+        (*self.key_manager_inner).read().await.get_next_key(branch).await
+    }
+
+    async fn get_key_at_index(&self, branch: String, index: u64) -> Result<PrivateKey, KeyManagerError> {
```
philipr-za:
```suggestion
    async fn get_key_at_index<T: Into<String>>(&self, branch: T, index: u64) -> Result<PrivateKey, KeyManagerError> {
```

I just noticed how many `.to_string()` calls there are and this would eliminate them for all client calls.

base_layer/wallet/src/output_manager_service/service.rs
delta1: the comment above this line is related to the call below it

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -617,14 +695,19 @@ where
...
+        &mut self,
         value: MicroTari,
         features: OutputFeatures,
     ) -> Result<UnblindedOutputBuilder, OutputManagerError> {
-        let (spending_key, script_private_key) = self
+        let spending_key = self
             .resources
             .master_key_manager
-            .get_next_spend_and_script_key()
+            .get_next_key(KeyManagerOmsBranch::Spend.to_string())
```
philipr-za: So this operation was done together before because these values need to use the same index. What happens if these two indices get out of sync somehow?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -927,27 +1020,32 @@ where
             "Building coinbase transaction for block_height {} with TxId: {}", block_height, tx_id
         );
 
-        let (spending_key, script_key) = self
+        let spending_key = self
```
philipr-za, line 987: This is supposed to be a key from the Coinbase branch using the height as the index.

delta1 (commented):
Nice work, just a few comments from my side

base_layer/wallet/src/output_manager_service/service.rs
philipr-za: furthermore, you have undone a huge amount of DRYing up by turning this into two calls here in the client service.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state_old.rs
delta1: agree and we should have a convention for this, I know we previously used `#testnet_reset` then it's easy to search

applications/tari_console_wallet/src/automation/commands.rs
SWvheerden: Because the key_manager does not know which branches it needs to have and not, it only keeps track of what is added. And this is just to let the user know when adds a new branch, this branch is already tracked, in other words, some service is using this exact branch and I am letting it up to the downstream provider to decide if this should be a hard error if the service shares a branch with some other service or if its okay. 

I was toying with the idea of adding a function `fn add_branch_if_not_exists`

base_layer/wallet/src/assets/asset_manager.rs
SWvheerden: Won't the asset manager need more than one branch of keys? That was the primary reason I chose an enum for that and not a constant. 
If not then lets go const?

base_layer/wallet/src/key_manager_service/error.rs
SWvheerden: See the above comment, it could be one or not, depending on the downstream user.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
SWvheerden: This should panic. You are never supposed to be able to decrypt a new_key_manager_state. 
It should not be possible. new_key_manager_state is only used when inserting a new key_manager_state into the DB, and there after if you retrieve it you will use key_manager_state.

base_layer/wallet/src/key_manager_service/handle.rs
SWvheerden: yes, but because the actual key_manager that updates the key sits behind a mutex, this is fine. 
A read lock here means that another service can ask to `get_next_key` for a different branch and still get that answer fine. 
As soon as you ask for the same branch, the mutex will stop you in the inner class.

base_layer/wallet/src/key_manager_service/handle.rs
SWvheerden: see above comment

base_layer/wallet/src/key_manager_service/mock.rs
SWvheerden: Moving it out to the integration folder means that no unit test can use the mock, only integration tests. It is behind a cfg(test) flag

base_layer/wallet/src/key_manager_service/storage/database/mod.rs
SWvheerden: I was trying to base this on how every other interface looks, but I am more than happy to move away from it as I don't like that interface.

base_layer/wallet/src/key_manager_service/service.rs
SWvheerden: removed that update.

base_layer/wallet/src/key_manager_service/service.rs
SWvheerden: Going to log this as a separate refactor. This PR is big as is. Currently the requests for the seed words pipe to the OMS which used to go to the key manager inside of the OMS. This is now piped to the current key manager.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state_old.rs
SWvheerden: changed the text and added the reset flag `#testnet_reset`

base_layer/wallet/src/key_manager_service/mock.rs
philipr-za: Some would argue that when you need other modules you are no longer writing unit tests but fair enough.

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    async fn apply_encryption(&self, cipher: Aes256Gcm) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner).write().await.apply_encryption(cipher).await
+    }
+
+    async fn remove_encryption(&self) -> Result<(), KeyManagerError> {
+        (*self.key_manager_inner).write().await.remove_encryption().await
+    }
+
+    async fn get_next_key<T: Into<String> + Send>(&self, branch: T) -> Result<(PrivateKey, u64), KeyManagerError> {
```
philipr-za: Just for future readability make this a struct with labelled fields, tuples like this are not easy to understand what data you are getting back.

base_layer/wallet/src/key_manager_service/service.rs
```diff
@@ -0,0 +1,176 @@
...
+{
+    pub fn new(master_seed: CipherSeed, db: KeyManagerDatabase<TBackend>) -> Self {
+        KeyManagerInner {
+            key_managers: HashMap::new(),
+            db,
+            master_seed,
+        }
+    }
+
+    pub async fn add_key_manager(&mut self, branch: String) -> Result<AddResult, KeyManagerError> {
```
philipr-za: This is a nit
```suggestion
    pub async fn add_key_manager_branch(&mut self, branch: String) -> Result<AddResult, KeyManagerError> {
```

philipr-za (commented):
Looks good, have one style comment I would like to see implemented please.

philipr-za (approved):
bors merge

delta1:
bors retry

PR #3889, opened 2022-03-02
feat: update console wallet notifications

philipr-za:
Description
---
This PR creates console wallet notifications for transaction events and also adds a hotkey to clear notifications.

A bug was also spotted in the way that Faux transactions were validated that resulted in the last Faux transaction confirmed event to be emitted after each validation. The validation was updated to only emit this event if the transaction was previously not confirmed.

### DIFF
--- a/applications/tari_console_wallet/src/ui/components/notification_tab.rs
+++ b/applications/tari_console_wallet/src/ui/components/notification_tab.rs
@@ -29,13 +29,21 @@ impl NotificationTab {
 
     fn draw_notifications<B>(&mut self, f: &mut Frame<B>, area: Rect, app_state: &AppState)
     where B: Backend {
+        let span_vec = vec![
+            Span::raw("Press "),
+            Span::styled("C", Style::default().add_modifier(Modifier::BOLD)),
+            Span::raw(" to clear notifications"),
+        ];
+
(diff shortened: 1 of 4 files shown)


applications/tari_console_wallet/src/ui/components/notification_tab.rs
```diff
@@ -4,9 +4,6 @@
 // are cleared.
 // Currently notifications are only added from the wallet_event_monitor which has
 // add_notification method.
-// TODO: auto delete old notifications. #LOGGED
```
philipr-za, line 7: This TODO is handled by the ability to manually clear the notifications, I don't think its ideal to do it automatically and notifications are not persisted so the manual option seems adequate.

applications/tari_console_wallet/src/ui/components/notification_tab.rs
```diff
@@ -4,9 +4,6 @@
...
-// TODO: add interaction with the notifications, e.g. if I have a pending transaction
```
philipr-za, line 8: I don't think this TODO is worthwhile. All the events come from a single pane so there is only one place to go which is one right arrow press away. There is also no mouse interactivity in the console wallet so the concept of "clicking" doesn't exist.

delta1 (approved):
Very nice

PR #3891, opened 2022-03-03
fix: revalidate rejected transactions when revalidating wallet database

SWvheerden:
Description
---
Changes the revelation SQL query to cause rejected transactions to be invalidated as well. 

Motivation and Context
---
Before this change, a wallet can have funds from a transaction that it believes wrongly was rejected. The revaluation will correctly cause that no funds are lost, but the UI displaying the transactions will still display that the transaction was rejected. This PR will fix this so that the wallet can correctly change this transaction to display mined. 

How Has This Been Tested?
---
Manual


philipr-za (approved):
bors merge

delta1:
bors retry

PR #3893, opened 2022-03-03
fix: console wallet display correct selected index and detail transaction

SWvheerden:
Description
---
This PR fixes a small bug when a new transaction is displayed in the list. The selected index is updated but the detailed transaction is not. This PR ensures that the selected index is set to the transaction displayed in the detail transaction section.

How Has This Been Tested?
---
manual

### DIFF
--- a/applications/tari_console_wallet/src/ui/components/transactions_tab.rs
+++ b/applications/tari_console_wallet/src/ui/components/transactions_tab.rs
@@ -172,6 +172,12 @@ impl TransactionsTab {
 
         let completed_txs = app_state.get_completed_txs();
         self.completed_list_state.set_num_items(completed_txs.len());
+        if let Some(detailed_tx) = &self.detailed_transaction {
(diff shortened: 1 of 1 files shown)


applications/tari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -172,6 +172,22 @@ impl TransactionsTab {
...
+                // index
+                let mut found_index = None;
+                let mut counter = 0;
+                for tx in &completed_txs {
+                    if detailed_tx.tx_id == tx.tx_id {
+                        found_index = Some(counter);
+                        break;
+                    }
+                    counter += 1;
+                }
```
delta1: There's a clippy warning here, but this could be done in much more idiomatic way with: 

```suggestion
                 let found_index = completed_txs
                    .iter()
                    .postition(|tx| tx.tx_id == detailed_tx.tx_id);
```

philipr-za:
clippy

delta1 (approved):

PR #3894, opened 2022-03-03
fix: update wallet logging config

philipr-za:
Description
---
Currently the tracing instrumentation is emitting as log records using the file names as the target. I have no idea how to stop it other than removing the tracing instrumentation. This behaviour is supposed to be an opt in feature but seems broken in Tracing.

So to combat this I updated the Root logger to rather just output to `other.log` which sends all these targetless records there and keeps the really log files clean.

I also updated the stdout appender to be a 10 meg rolling file that is not appended to and pointed the wallet logger at it so that the logging pane in the console wallet gets some useful info again.


delta1 (approved):
bors r+

delta1:
bors merge

PR #3898, opened 2022-03-07
fix: update metadata size calculation to use FixedSet.iter()

philipr-za:
Description
---
This is solving the following TODO: implement iter for FixedSet to avoid the clone

Iter() was implemented for FixedSet in the newest version of `tari_utilities` so this PR updates the main repo to point to the `main` branch of `tari_utilities` and then uses this iter() implementation to do what the TODO suggests which is to to avoid the clone by using the iterator.


stringhandler (approved):

delta1:
bors retry

delta1:
bors retry

PR #3902, opened 2022-03-08
fix(core): correctly filter pruned sync peers for block sync

sdbondi:
Description
---
Corrects filter predicate for full block sync peers

Motivation and Context
---
Rpc error: `Service returned an error: BadRequest: Requested full block body at height 6804, however this node has an effective pruned height of 15941` caused by peer incorrectly selecting pruned nodes. 

How Has This Been Tested?
---
Manually


delta1 (approved):
LGTM

PR #3903, opened 2022-03-08
feat: gracefully handle recovering a duplicate output in LibWallet

philipr-za:
Description
---
This PR addresses the issue of creating duplicate Faux transaction when an output that has previously been recovered is recovered again. This can happen during reorgs.
It is a small fix to the code that when outputs are scanned to see if they can be rewound only new unique outputs are returned to the caller and duplicate outputs are ignored.
The UTXO scanner will then only create new Faux transactions for the outputs that have not been previously recovered.

Two Rust tests were added to confirm the correct behaviour in output scanning.

This behaviour was already implemented for One-sided payments and a test existed that tested the ignoring of duplicates.

How Has This Been Tested?
---
Rust tests provided


SWvheerden (dismissed):
LGTM

stringhandler (dismissed):

philipr-za:
@mikethetike conflicts sorted

stringhandler (approved):

PR #3904, opened 2022-03-08
feat!: add less aggressive txn cancelling

hansieodendaal:
Description
---
- Added less aggressive transaction canceling, i.e. removed automatic transaction cancelling, when:
  - a wallet has difficulty discovering the receiving party (for direct P2P comms);
  - a wallet has difficulty discovering base node peers that could be used to send the transaction via store and forward;
  - tor is not available.
- Updated the wallet ffi with the new callbacks.
- Added a cucumber test to test sending transactions while offline.

Users will now manually cancel transactions stuck in pending due to the wallet continuing to be being offline.

_**Note:** Wallet ffi interface changed._

Motivation and Context
---
During a recent semi-stress test [[report here](https://github.com/tari-project/tari-data-analysis/blob/master/reports/stress_tests/20211221-make-it-rain/Stress%20test%20of%2020211221%20-%20analysis.md)] adverse connectivity and/or discovery symptoms caused ~55% of transactions to be abandoned. Attempted transactions should not be canceled if they failed to be sent both directly and via store and forward; a peer-to-peer connection may fail for valid reasons but failing to submit a transaction to a base node for store and forward delivery does not warrant the transaction to be canceled.

How Has This Been Tested?
---
System-level testing
Unit tests
Cucumber test: `Scenario: Wallet send transactions while offline`


base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1487,9 +1506,44 @@ where
...
+                    stage,
                     None,
                     self.last_seen_tip_height,
+                    sender_protocol,
+                );
+
+                let join_handle = tokio::spawn(protocol.execute());
+                join_handles.push(join_handle);
+            } else if stage == TransactionSendProtocolStage::Queued {
+                debug!(
```
philipr-za, line 1496: This branch is a perfect candidate for DRYing up this code.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2836,6 +2837,114 @@ pub unsafe extern "C" fn pending_inbound_transaction_destroy(transaction: *mut T
...
+
+/// Gets the direct send status of a TariTransactionSendStatus
+///
+/// ## Arguments
+/// `status` - The pointer to a TariTransactionSendStatus
+/// `error_out` - Pointer to an int which will be modified to an error code should one occur, may not be null. Functions
+/// as an out parameter.
+///
+/// ## Returns
+/// `c_uint` - Returns 1 if the transaction was sent directly, zero if it was not
```
philipr-za: This should be a bool

philipr-za: but actually I would say that to save having to create and maintain so many extra FFI methods why not just a single method that returns the enum encoded as unit?

philipr-za (commented):
Looks fine. I believe all the send scenarios are catered for in the tests? Cucumber or Rust?

Just think the FFI interface is not as clean as it could be.

hansieodendaal:
> Looks fine. I believe all the send scenarios are catered for in the tests? Cucumber or Rust?

Yes, I added a special cucumber test to test the queueing as well.

stringhandler (approved):

PR #3906, opened 2022-03-09
chore: clean up providing seed words from LibWallet

philipr-za:
Description
---
This PR just refactors where the seed words are provided from in LibWallet. Previously the KeyManager resided in the OutputManager Service and the key manager entropy was only contained there so the old `get_seed_words` methods all pointed into the OMS service. Since the CipherSeed is now stored in the main wallet database it is not longer required to access this data via the OMS so this PR cleans that up.

How Has This Been Tested?
---
manually


SWvheerden (commented):
LGTM
This is definitely the right type of PR. More deleted lines than added ones.

delta1 (approved):

stringhandler (approved):

PR #3907, opened 2022-03-10
fix(dht)!: fixes MAC related key vuln for propagated cleartext msgs

sdbondi:
Description
---
- changes challenge in signed messages to `e = H(P||R||H(m))` 
- remove minor from DHT versioning

Motivation and Context
---

This only affects _signed cleartext_ messages that are propagated through adversarial nodes.
Encrypted propagated messages are unaffected because only the recipient can decrypt the MAC and so cannot forge a signature.

Nonce malleability

```text
e = H(P||m)
<R, s> for P

Attacker chooses 
R' = s.G - eP
s' any scalar
and transmits <R', s'> for P

Verifier:
s'.G ?= R' + eP
      ?= s'.G - eP + eP
      ?= s'.G  (true for any s')
```

Related key 
```text
sig = <R, s> for P
e = H(R||m)

Attacker:
sig' = <R, s'> for P'
s' = s + a.H(R||m)
P' = P + a.G

Verifier:
s'.G ?= R + eP'
       ?= R + e(k + a).G
```

All signed cleartext and encrypted DHT messages are incompatible with current network:
Namely,
- Peer discovery
- Wallet Propagated/SAF messages

How Has This Been Tested?
---
New unit tests that verify OriginMac is secure against these vulnerabilities

Tested base node manually, and normal cleartext block and tx propagation messages continue to work

Tested wallet, direct send transactions are still compatible


comms/core/src/peer_manager/node_id.rs
```diff
@@ -74,13 +74,11 @@ impl NodeId {
...
             .expect("NodeId::byte_size() is invalid")
             .chain(bytes)
-            .finalize_variable(|hash| {
-                // Safety: output size and buf size are equal
-                buf.copy_from_slice(hash)
-            });
+            // Safety: output size and buf size are equal
+            .finalize_variable(&mut buf).unwrap();
         NodeId(buf)
     }
```
AaronFeickert, line 84: This hash invocation should use domain separation to avoid unexpected collisions elsewhere. This is supported directly using Blake2 personalization.

comms/dht/src/crypt.rs
```diff
@@ -105,29 +100,36 @@ pub fn create_origin_mac_challenge(header: &DhtMessageHeader, body: &[u8]) -> Ch
...
     expires: Option<EpochTime>,
     ephemeral_public_key: Option<&CommsPublicKey>,
     body: &[u8],
-) -> Challenge {
+) -> [u8; 32] {
     let mut mac_challenge = Challenge::new();
-    mac_challenge.update(&protocol_version.to_bytes());
-    mac_challenge.update(destination.as_inner_bytes());
+    mac_challenge.update(&protocol_version.as_bytes());
+    mac_challenge.update(destination.to_inner_bytes());
```
AaronFeickert, line 115: Is this guaranteed to have a fixed-length byte representation?

AaronFeickert (changes requested):

comms/dht/src/crypt.rs
AaronFeickert: Looks like this has a fixed-length representation, and types are differentiated using a flag.

AaronFeickert: It is possible, if desired, to use variable-length representations. In this case, the length (as a fixed-length representation!) should be prepended after the (fixed-length) flag.

sdbondi: That's right, each variant of `NodeDestination` has a single byte representation prepended to the encoding. Gotcha, both the NodeId and CommsPublicKey have fixed-length representations so I don't think we'll have a need for variable-length encodings here.

comms/dht/src/origin_mac.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+impl OriginMac {
+    /// Create a new signed [OriginMac](self::OriginMac) for the given message.
+    pub fn new_signed(signer_secret_key: CommsSecretKey, message: &[u8]) -> Self {
+        let (nonce_s, nonce_pk) = CommsPublicKey::random_keypair(&mut OsRng);
+        let signer_public_key = CommsPublicKey::from_secret_key(&signer_secret_key);
+        let challenge = construct_origin_mac_hash(&signer_public_key, &nonce_pk, message);
+        let signature = Signature::sign(signer_secret_key, nonce_s, &challenge)
+            .expect("challenge is [u8;32] but SchnorrSignature::sign failed");
```
stringhandler, line 58: I suppose this is basically infallible.

stringhandler (approved):

comms/dht/src/origin_mac.rs
sdbondi: It's 100% infallible with a 32-byte challenge and Ristretto - we should probably make an infallible version of sign that takes a [u8;32]

PR #3912, opened 2022-03-11
fix(block-sync): use avg latency to determine slow sync peer for block sync

sdbondi:
Description
---
Uses average blocks received per second to determine slow sync peer.
Display average block receive time on status line

Motivation and Context
---
Esp. for large blocks, there could be a single large delay that would previously immediately cause the next peer to be attempted. Using the average blocks received per second, a "blip" in the latency will not cause the sync peer to switch.

How Has This Been Tested?
---
Manually

### DIFF
--- a/base_layer/core/src/base_node/sync/config.rs
+++ b/base_layer/core/src/base_node/sync/config.rs
@@ -46,7 +46,7 @@ pub struct BlockchainSyncConfig {
 impl Default for BlockchainSyncConfig {
     fn default() -> Self {
         Self {
-            initial_max_sync_latency: Duration::from_secs(4),
+            initial_max_sync_latency: Duration::from_secs(10),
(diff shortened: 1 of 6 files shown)


base_layer/core/src/base_node/sync/config.rs
```diff
@@ -46,7 +46,7 @@ pub struct BlockchainSyncConfig {
 impl Default for BlockchainSyncConfig {
     fn default() -> Self {
         Self {
-            initial_max_sync_latency: Duration::from_secs(3),
+            initial_max_sync_latency: Duration::from_secs(4),
```
stringhandler: Let's bump this to 10 for now
```suggestion
            initial_max_sync_latency: Duration::from_secs(10),
```

stringhandler (dismissed):

stringhandler (approved):

PR #3918, opened 2022-03-16
chore: rewrite to_bytes_with_language

Cifko:
Description
---
I thought about it. And I decided to get rid of the bits array. If you don't like the bit shifting I can rewrite it back. But I personally find this easier to read.

Motivation and Context
---
Get rid of some ties to the tari_utilities. I can now (probably) remove 2 functions from tari_utilities.

How Has This Been Tested?
---
There is a cargo test `test_mnemonic_from_bytes_and_to_bytes` at the bottom of the file.


base_layer/key_manager/src/mnemonic.rs
```diff
@@ -23,8 +23,9 @@
 use std::{cmp::Ordering, slice::Iter};
 
 use strum_macros::{Display, EnumString};
-use tari_crypto::tari_utilities::bit::*;
+use tari_crypto::tari_utilities::bit::{bytes_to_bits, checked_bits_to_uint};
 
+// use tari_crypto::tari_utilities::bit::*;
```
sdbondi:
```suggestion

```

base_layer/key_manager/src/mnemonic.rs
```diff
@@ -211,20 +212,24 @@ pub fn to_bytes(mnemonic_seq: &[String]) -> Result<Vec<u8>, MnemonicError> {
...
+    let mut rest_bits: u8 = 0;
+
     for curr_word in mnemonic_seq {
-        match find_mnemonic_index_from_word(curr_word, language) {
-            Ok(index) => {
-                let curr_bits = checked_uint_to_bits(index, 11).ok_or(MnemonicError::IntToBitsConversion)?;
-                bits.extend(curr_bits.iter().cloned());
-            },
-            Err(err) => return Err(err),
+        let index = find_mnemonic_index_from_word(curr_word, language)?;
```
sdbondi, line 233: This is really clever :) A few notes:

1. It's a bit hard to follow the algo, could you write a brief comment on how it works in the function doc?
2. This takes in an arbitrary length mnemonic and this function is public and could be used to encode any length. I removed the 33 length conditional at the bottom and wrote a quick fuzzer and found some failure cases. Could you include this in your PR and see if you can fix it?

```rust
     #[test]
    fn fuzzer() {
        use rand::RngCore;
let start = 33; // NB: this can be < 33
        for len in start..1024 * 1024 {
            let mut secretkey_bytes = vec![0u8; len]; // RistrettoSecretKey::random(&mut OsRng).to_vec();
            OsRng.fill_bytes(&mut secretkey_bytes);
            let mnemonic_seq = mnemonic::from_bytes(secretkey_bytes.clone(), &MnemonicLanguage::English).unwrap();
            let mnemonic_bytes = mnemonic::to_bytes(&mnemonic_seq).unwrap();
            assert_eq!(secretkey_bytes, mnemonic_bytes, "failed len = {}", len);
        }
    }
```

 3. This code is not performance critical (it's run rarely) so prioritize correctness and soundness over perf

sdbondi (commented):
Looks good - it is still possible to break this so I've included a fuzz test

base_layer/key_manager/src/mnemonic.rs
Cifko: I've added arbitrary length.
But I have to rewrite the test to take only multiple of 11. Because the `from_bytes` is doing padding. So the `to_bytes` produce the padded value, which is sometimes longer (padded with 0 bytes).
Other options is to pad the secretkey_bytes with zeros to match the length of the mnemonic_bytes.  But that just complicates the test.

stringhandler (approved):

base_layer/key_manager/src/mnemonic.rs
sdbondi: Thanks :) Yeah happy with this approach

PR #3920, opened 2022-03-16
feat: add config setting to resize terminal

hansieodendaal:
Description
---
- Added a config setting to resize the base node terminal
- Added terminal titles to base node, console wallet,  merge mining proxy and minining node.

Motivation and Context
---
See avbove

How Has This Been Tested?
---
System-level testing


applications/tari_base_node/src/commands/cli.rs
```diff
@@ -105,14 +108,21 @@ fn multiline_find_display_length(lines: &str) -> usize {
 
 /// Try to resize terminal to make sure the width is enough.
 /// In case of error, just simply print out the error.
-fn resize_terminal_to_fit_the_box(width: usize, height: usize) {
-    if let Err(e) = execute!(stdout(), SetSize(width as u16, height as u16)) {
-        println!("Can't resize terminal to fit the box. Error: {}", e)
+fn apply_terminal_properties(width: usize, height: usize, resize_terminal: bool) {
```
stringhandler: Rather rename this function to what it does, remove the `if` and allow the caller to determine if they want to call it outside of the function. 

```suggestion
fn resize_terminal(width: usize, height: usize) {
```

applications/tari_base_node/src/commands/cli.rs
```diff
@@ -188,5 +198,7 @@ pub fn print_banner(commands: Vec<String>, chunk_size: i32) {
     }
     println!("{}", box_line(target_line_length, false));
 
-    resize_terminal_to_fit_the_box(target_line_length, height_to_resize);
+    if resize_terminal {
```
stringhandler, line 199: Since you are already checking this if, no need to pass in `resize_terminal` (See previous comment)

common/config/presets/base_node.toml
```diff
@@ -30,6 +30,8 @@ grpc_enabled = true
 grpc_address = "/ip4/127.0.0.1/tcp/18142"
 # Set to true to record all reorgs. Recorded reorgs can be viewed using the list-reorgs command.
 track_reorgs = true
+# Resize the base node terminal on startup equal to the banner size [default = true]
+#resize_terminal_on_startup = false
```
stringhandler: The default should be true

applications/tari_base_node/src/commands/cli.rs
hansieodendaal: Yip, remnant of last minute change...

stringhandler (approved):

PR #3925, opened 2022-03-17
refactor: remove unwraps from some modules

therustmonk:
Description
---
Remove `unwrap` calls from some modules.

Motivation and Context
---
The code contains `unwrap` calls that can lead to panic. Using results instead of unwraps makes the code safer.

How Has This Been Tested?
---
CI

### DIFF
--- a/base_layer/common_types/src/emoji.rs
+++ b/base_layer/common_types/src/emoji.rs
@@ -25,10 +25,7 @@ use std::{
     fmt::{Display, Error, Formatter},
 };
 
-use tari_crypto::tari_utilities::{
-    hex::{Hex, HexError},
-    ByteArray,
-};
+use tari_crypto::tari_utilities::{hex::Hex, ByteArray};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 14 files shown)


base_layer/common_types/src/emoji.rs
```diff
@@ -172,37 +164,45 @@ impl Display for EmojiId {
...
+    use anyhow::Error;
     use tari_crypto::tari_utilities::hex::Hex;
 
     use crate::{emoji::EmojiId, types::PublicKey};
 
     #[test]
-    fn convert_key() {
-        let pubkey = PublicKey::from_hex("70350e09c474809209824c6e6888707b7dd09959aa227343b5106382b856f73a").unwrap();
-        let eid = EmojiId::from_hex("70350e09c474809209824c6e6888707b7dd09959aa227343b5106382b856f73a").unwrap();
+    fn convert_key() -> Result<(), Error> {
```
sdbondi, line 186: No problem with this, but let's not change this everywhere. From experience, the unwraps in tests are more useful than returning a result from tests because they give you the line of the unwrap/panic whereas returning a result just gives you the failure - in longer tests, it becomes difficult to hunt for the line that failed.

Tests with results have their uses though (e.g cleanup in integration tests on failure)

base_layer/common_types/src/emoji.rs
```diff
@@ -97,15 +94,15 @@ impl EmojiId {
...
         let key = PublicKey::from_hex(hex_key)?;
         Ok(EmojiId::from_pubkey(&key))
     }
 
     /// Return the public key that this emoji ID represents
-    pub fn to_pubkey(&self) -> PublicKey {
-        let bytes = self.to_bytes();
-        PublicKey::from_bytes(&bytes).unwrap()
+    pub fn to_pubkey(&self) -> Result<PublicKey, EmojiIdError> {
+        let bytes = self.to_bytes()?;
```
sdbondi, line 104: Although in principle I agree with avoiding unwraps, It's worth noting that this function can never fail as the code currently stands because there are 256 emojis and u8 ranges from 0 to 256.

sdbondi (approved):
LGTM, some digressive comments - a clippy needs fixing

base_layer/common_types/src/emoji.rs
sdbondi: Ah nevermind, I see the input is a string - yes these unwraps are bad :)

sdbondi: Actually, is see the `byte_vec` code is duplicated in `str_to_pubkey` and `str_to_pubkey` is used in the wallet.
So this cannot panic previously because you can only construct the String from bytes using `EMOJI` from a public key 😥  - this Struct can be greatly simplified and byte_vec could safely be infallible - happy with the changes made here though

therustmonk:
Closed to group and split changes of separate crates into separate PRs.

PR #3926, opened 2022-03-17
fix: bug in block timing grpc method

delta1:
Description
---

- Fixes a bug where the grpc method for block timing was requesting the headers in the wrong order, resulting in wrong calculation 
- Removes the deprecated duplicate GetCalcTiming method 
- Un-ignores the chain balance test which is working again


base_layer/core/src/blocks/block_header.rs
```diff
@@ -168,16 +169,34 @@ impl BlockHeader {
         }
     }
 
+    /// Returns the [std::cmp::Ordering] of the timestamp of this header compared to the "other" header
+    pub fn compare_timestamp(&self, other: &BlockHeader) -> Ordering {
+        self.timestamp.as_u64().cmp(&other.timestamp.as_u64())
```
sdbondi: `EpochTime` has a `PartialOrd` impl that does this comparison.

base_layer/core/src/blocks/block_header.rs
```diff
@@ -168,16 +169,34 @@ impl BlockHeader {
...
+    }
+
     #[cfg(feature = "base_node")]
     pub fn into_builder(self) -> BlockBuilder {
         BlockBuilder::new(self.version).with_header(self)
     }
 
-    /// Given a slice of headers (in reverse order), calculate the maximum, minimum and average periods between them
+    /// Given a slice of headers, calculate the maximum, minimum and average periods between them.
+    /// Allocates in case the headers are in the wrong order.
```
sdbondi: This function always allocates for len >= 2 (`to_vec` creates a [copy of the slice and each element](https://github.com/rust-lang/rust/blob/051d1176b786aadd7d7c048f822cb6bfab00fe03/library/alloc/src/slice.rs#L209)) - but that is fine for this function.

Perhaps add a comment that headers should be ordered from oldest to youngest.

base_layer/core/src/blocks/block_header.rs
```diff
@@ -168,16 +169,34 @@ impl BlockHeader {
...
     pub fn timing_stats(headers: &[BlockHeader]) -> (u64, u64, f64) {
         if headers.len() < 2 {
             (0, 0, 0.0)
         } else {
+            let mut headers = headers.to_vec();
+
+            // ensure the slice is in reverse order
+            let ordering = headers[0].compare_timestamp(&headers[headers.len() - 1]);
```
sdbondi: nit: I don't think the compare_timestamp is generally useful enough
```suggestion
            let ordering = headers[0].timestamp.cmp(&headers[headers.len() - 1].timestamp);
```

sdbondi (commented):
LGTM - some minor comments

base_layer/core/src/blocks/block_header.rs
delta1: Whoops lol it did feel wrong

base_layer/core/src/blocks/block_header.rs
delta1: Yeah the wording isn't great, will update it.

base_layer/core/src/blocks/block_header.rs
```diff
@@ -168,16 +169,35 @@ impl BlockHeader {
         }
     }
 
+    /// Returns the [std::cmp::Ordering] of the timestamp of this header compared to the "other" header
+    pub fn compare_timestamp(&self, other: &BlockHeader) -> Ordering {
```
sdbondi: Think you can remove this?

delta1: Very strange I did remove it 🤷‍♂️

sdbondi (approved):

PR #3927, opened 2022-03-17
fix(sync): adds extra checks for sync stream termination

sdbondi:
Description
---
- adds extra checks in block sync RPC service for premature stream termination 
- uses rolling average for slow peer detection in header and horizon sync
- prevent more than one UTXO sync session per peer

Motivation and Context
---
Peers may end a sync session and quickly initiate a new one. Because the previous session could be still sending the batch of data for the previous session, the new session will be denied until the data has been fetched. This PR adds additional checks after loading data from the database but before sending it to check if the session has ended. However, this race condition still exists as it is inherent to the rule of only allowing one sync session per peer. In my tests with 600ms max permitted sync latency I did not manage to trigger the `Forbidden: Existing sync session found for this client. Only a single session is permitted` error, so this may still be an issue.  But these changes should lessen the chance of this happening. 

How Has This Been Tested?
---
Manually, existing tests


SWvheerden (dismissed):
LGTM
Tested with a sync from 0 up to the tip.

stringhandler (approved):

PR #3928, opened 2022-03-18
fix(comms)!: commit to public key and nonce in identity sig

sdbondi:
Description
---

- commit to public nonce and public key on identity signature 

Motivation and Context
---
Schnorr signatures are not secure unless public nonce and public key are committed to

How Has This Been Tested?
---
Existing tests (internal to identity signature)


SWvheerden (dismissed):
LGTM

stringhandler (approved):

PR #3930, opened 2022-03-18
feat: add mmr check to reconstructed block and mempool validation for unique excess signature

delta1:
Description
---

- Adds a check to the reconstructed block to ensure MMR roots match, if not then requests the full block from source peer
- Adds mempool validation to check submitted kernel excess signature does not already exist in the chain db

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -51,6 +51,7 @@ use crate::{
     consensus::{ConsensusConstants, ConsensusManager},
     mempool::Mempool,
     proof_of_work::{Difficulty, PowAlgorithm},
+    validation::helpers,
 };
 
 const LOG_TARGET: &str = "c::bn::comms_interface::inbound_handler";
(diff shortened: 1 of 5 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -617,9 +618,29 @@ where B: BlockchainBackend + 'static
 
         // NB: Add the header last because `with_transactions` etc updates the current header, but we have the final one
         // already
-        builder = builder.with_header(header);
+        builder = builder.with_header(header.clone());
+        let block = builder.build();
+
+        // Perform a sanity check on the reconstructed block, if the MMR roots don't match then it's possible one or
+        // more transactions in our mempool had the same excess/signature for a *different* transaction.
+        // This is extremely unlikely, but still possible. In case of a mismatch, request the full block from the peer.
```
sdbondi, line 626: I think this case is very possible:

Given these two spend to self transactions with inputs already in the utxo set (k_1, k_2), arbitrary scalar `x`
- Generate: TX1 : `k_1G + v_1H - ((k_2 - x)G + v_1H)` Excess: `(k_1 - k_2 + x)G`
- Generate TX2: `k_2G + v_1H - ((k_1 + 2k_2 - x)G + v_1H)`  Excess: `(k_1 - k_2 + x)G`
- reuse nonces (which leaks your key but oh well), reuse kernel offset, same fees for both txns

This produces a different commitment for the output, but the same excess sig. 

Would be nice to have an integration/unit test to test this case, but that can be in another PR

sdbondi (approved):
Awesome, think this handles the case well

PR #3931, opened 2022-03-18
feat: listen to terminal events in the watch mode

therustmonk:
Description
---
It adds an events stream to listens to key press events in the watch mode.

Another changes:
- `cli_loop` function replaced with the `CliLoop` struct that also moved to a separate module
- `status_loop` replaced with the `CliLoop` struct

Motivation and Context
---
We should suggest the user how to exit the watch mode.
Also the cli loop reduced and divided into several parts.

How Has This Been Tested?
---
Manually

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1451,15 +1451,16 @@ dependencies = [
 
 [[package]]
 name = "crossterm"
-version = "0.22.1"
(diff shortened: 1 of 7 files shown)


stringhandler (approved):
Nice

PR #3932, opened 2022-03-18
feat(metrics): add UTXO set size to base node metrics

sdbondi:
Description
---
- adds `tari_base_node::blockchain::utxo_set_size`
- splits `tari_base_node::blockchain::target_difficulty` into `target_difficulty_sha` and `target_difficulty_monero`
- changes `BlockchainDb::utxo_count` to return the number of UTXOs in the unspent set, rather than the number of outputs that were ever stored (differs depending on pruned height).
- adds peer list size metric

Motivation and Context
---
Monitor the UTXO set size and independent PoW algos.
fix: `utco_count` was meant to return the current UTXO set size 

How Has This Been Tested?
---
Local prometheus


delta1 (approved):
LGTM

PR #3934, opened 2022-03-21
fix: correct main path for wallet rpc client

sdbondi:
Description
---
- Fixes main path in package.json
- Minor improvements of debug logs in washing machine
- Washing machine waits for balance before coin split
- update package.json so that CI passes using node version in .nvmrc

Motivation and Context
---

Client libs complain about the incorrect main path

How Has This Been Tested?
---


delta1 (approved):
LGTM

PR #3935, opened 2022-03-21
test(core): reinstate inputs_not_mallaeble test

sdbondi:
Description
---
Reinstate/unignore `inputs_not_malleable` unit test

Motivation and Context
---
Test was broken 

How Has This Been Tested?
---
Test passes

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/transaction_input.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_input.rs
@@ -356,6 +356,8 @@ impl TransactionInput {
         }
     }
 
+    /// Sets the input's Tari script. Only useful in tests.
+    #[cfg(test)]
     pub fn set_script(&mut self, new_script: TariScript) -> Result<(), TransactionError> {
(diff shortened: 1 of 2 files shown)


base_layer/core/src/transactions/transaction_components/transaction_input.rs
```diff
@@ -356,6 +356,15 @@ impl TransactionInput {
         }
     }
 
+    pub fn set_script(&mut self, new_script: TariScript) -> Result<(), TransactionError> {
```
stringhandler, line 361:
```suggestion
  // Should only be used in tests
#[cfg(test)]
    pub fn set_script(&mut self, new_script: TariScript) -> Result<(), TransactionError> {
```

sdbondi: Put `#[cfg(test)]` on for good measure (EDIT: oh you wrote that :P)

stringhandler (approved):

PR #3936, opened 2022-03-22
fix: package.json and package-lock.json are not in sync

Cifko:
Description
---
We are getting errors of one of the test `CI npm packages` :
`'npm ci' can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with 'npm install' before continuing.`
So this PR provides the synced packages-json.lock

How Has This Been Tested?
---
Running `npm ci` after the fix.

### DIFF
--- a/applications/tari_collectibles/web-app/package-lock.json
+++ b/applications/tari_collectibles/web-app/package-lock.json
@@ -74,6 +74,15 @@
         }
       }
     },
+    "@babel/eslint-plugin": {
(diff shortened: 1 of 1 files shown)


sdbondi:
Also fixed in #3934

Cifko:
> Also fixed in #3934

Oh, ok, I will then for your PR to get in. Cancelling this one.

PR #3939, opened 2022-03-22
chore: allow null pointer as OutputFeatures for walletFFI

SWvheerden:
Description
---
This will allow a null pointer to be used as the output feature for the import utxo again. This will use the default features if null is used.

Motivation and Context
---
We need to create a proper mapping to the output features so that users can create output features to be used from outside. This will allow users to continue importing faucet utxos for the time being until that is created.

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -123,7 +123,7 @@ use tari_comms::{
 use tari_comms_dht::{store_forward::SafConfig, DbConnectionUrl, DhtConfig};
 use tari_core::{
     covenants::Covenant,
-    transactions::{tari_amount::MicroTari, CryptoFactories},
+    transactions::{tari_amount::MicroTari, transaction_components::OutputFeatures, CryptoFactories},
 };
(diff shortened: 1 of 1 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5205,11 +5205,11 @@ pub unsafe extern "C" fn wallet_import_utxo(
         return 0;
     }
 
-    if features.is_null() {
-        error = LibWalletError::from(InterfaceError::NullError("features".to_string())).code;
-        ptr::swap(error_out, &mut error as *mut c_int);
-        return 0;
-    }
+    let features = if features.is_null() {
+        OutputFeatures::default()
```
philipr-za, line 5209: @hansieodendaal is this alright? Will a recovery byte of 0 be treated correctly?

hansieodendaal: The features and metadata signature must be consistent. Default features are only allowed for `V0`

SWvheerden:
Lets rather do this properly.

PR #3942, opened 2022-03-22
refactor: move clippy lints to common file

stringhandler:
Description
---
* Move all clippy lints to a single file
* Update Contributing.md to inform new developers of process
* Replace `cargo clippy` with `cargo lints clippy` in CI workflow

Motivation and Context
---
To reduce duplicate code and the chance of missing lints on some projects. This (currently) requires developers to install and use `cargo-lints` and run `cargo lints clippy --all-targets` to get all errors.

I would have preferred a solution that did not require the installation of a different crate, but after much research and testing, there is no better (cleaner) way of doing this with only clippy and cargo.

How Has This Been Tested?
---
Ran `cargo lints clippy --all-targets`


stringhandler:
I'm creating this PR now, which should fail clippy. Hopefully I can get some help fixing the individual clippy warnings

lints.toml
```diff
@@ -0,0 +1,70 @@
...
+    # style
+    'clippy::explicit_into_iter_loop',
+    'clippy::explicit_iter_loop',
+    'clippy::if_not_else',
+    'clippy::match_bool',
+    'clippy::needless_pass_by_value',
+    'clippy::range_plus_one',
+    'clippy::struct_excessive_bools',
+    # perhaps this is a bit harsh
+    'clippy::too_many_lines',
```
Cifko: I think the default value of 100 is reasonable to have.

CjS77:
I like this a lot. 
I personally hate seeing lint configs in source code files, so this really butters my scones.

CjS77 (dismissed):

Cargo.toml
```diff
@@ -35,10 +35,18 @@ members = [
...
-#[profile.release]
-#debug = true
 
 # Shutdown when panicking so we can see the error, specifically for the wallet
 [profile.release]
 panic = 'abort'
+#debug = true
+
+
+#![cfg_attr(not(debug_assertions), deny(unused_variables))]
```
sdbondi, line 45: Can remove this?

stringhandler: yeah, added while I was trying stuff

stringhandler:
Continued by @Cifko in #3961

PR #3944, opened 2022-03-22
chore: restructure comms files into `core` crate

sdbondi:
Description
---
- moves `comms/` into `comms/core/`
- clean up some unused error variants 

Motivation and Context
---
The standard pattern for layers is to have a core crate in the same directory level as  related crates

How Has This Been Tested?
---
cargo test


sdbondi:
@mikethetike still need to be reviewed/approved 🙂

sdbondi:
@mikethetike still need to be reviewed/approved 🙂

delta1 (dismissed):

PR #3945, opened 2022-03-22
chore: deprecate stratum transcoder

SWvheerden:
Description
---
Miningcore used the stratum transcoder to do communication with the Tari base node and console wallet. 
This PR deprecates the stratum transcoder. Mining core is updated to only GRPC requests directly the base_node and console wallet. 

Updated GRPC methods to use consensus encoded bytes.

Renamed the `tari_stratum_ffi` to `tari_mining_helpers` as this more accurately reflects their usage as well as updated to use consensus encoded bytes rather than json encoded hex strings. 

Fix a bug in the covenants decoding.  Added consensus encoding and decoding to block and all dependants. 

Updated miner to use base64 encoded consensus bytes, rather than json encoded hex string.l


base_layer/core/src/blocks/block_header.rs
```diff
@@ -382,6 +384,48 @@ pub(crate) mod hash_serializer {
     }
 }
 
+impl ConsensusEncoding for BlockHeader {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<usize, io::Error> {
+        let mut written = self.version.consensus_encode(writer)?;
+        written += self.height.consensus_encode(writer)?;
+        written += self.prev_hash.consensus_encode(writer)?;
+        written += self.timestamp.as_u64().consensus_encode(writer)?;
+        written += self.output_mr.consensus_encode(writer)?;
```
sdbondi: For all hashes, we need to use fixed size encoding. Unfortunately, since we use a dynamic vec for hashes, this is unwieldy. The following code (from `BlockHeader::merge_mining_hash`) will do it:

```rust
            // TODO: Cleanup if/when we migrate to fixed 32-byte array type for hashes
            copy_into_fixed_array::<_, 32>(&self.output_mr)
                .map_err(....to io error....)
                .consensus_encode(&mut hasher)?;
```

For decoding:

```rust
let output_mr = <[u8;32] as ConsensusDecoding>::consensus_decode(reader)?;
header.output_mr = output_mr.to_vec();
```

sdbondi: Plan on moving to fixed hashes at some point

stringhandler (dismissed):

stringhandler (approved):

PR #3948, opened 2022-03-23
chore(core): make HashWriter API more ergonomic and similar to Digest

sdbondi:
Description
---
- adds chain method to `HashWriter` 

Motivation and Context
---
Improves ergonomics of `HashWriter` to remove the need to unwrap each call to consensus encode when using it

How Has This Been Tested?
---
Existing tests, manually to check no breaking changes


delta1 (approved):
Nice!

PR #3950, opened 2022-03-23
fix: fix Tor ID deserialization issue

philipr-za:
Description
---
Due to an issue with excluding the private key from the TorIdentity serialization that was fixed in https://github.com/tari-project/tari/pull/3946 any console wallets that ran the old code would crash on startup with a serialization error.

This PR adds some resilience to the startup process that if it cannot deserialize the old Tor Id it will log the issue and then use a newly generated one.

How Has This Been Tested?
---
Manually


sdbondi (approved):
LGTM - #3951 stores the identity sig in the case that the address is updated and node identity re-signed

sdbondi (approved):

PR #3951, opened 2022-03-23
fix(wallet): ensure that identity sig is stored on startup

sdbondi:
Description
---
- adds code to store the identity signature

Motivation and Context
---
Whenever the node address changes, the identity signature is resigned for the new address. This identity signature should be preserved and reloaded later.

Ref #3950 

How Has This Been Tested?
---
Manually: Checking that identity signature is stored and doesnt change between reboots


philipr-za (approved):
LGTM

PR #3956, opened 2022-03-24
fix: fix ffi import external utxo from faucet

hansieodendaal:
Description
---
- Fixed importing external UTXOs via the FFI _by changing the philosophy about having inconsistent recovery byte data in the wallet database_
- Renamed import UTXO methods to be more descriptive
- Added a unit test to test importing external UTXOs via the FFI

_**Note:** The FFI import UTXO method was renamed; this is a breaking change on the FFI interface._

Motivation and Context
---
See above

How Has This Been Tested?
---
Added a new unit test that test the import UTXO interface

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -121,10 +121,7 @@ use tari_comms::{
     types::{CommsPublicKey, CommsSecretKey},
 };
 use tari_comms_dht::{store_forward::SafConfig, DbConnectionUrl, DhtConfig};
-use tari_core::{
-    covenants::Covenant,
-    transactions::{tari_amount::MicroTari, transaction_components::OutputFeatures, CryptoFactories},
-};
+use tari_core::transactions::{tari_amount::MicroTari, CryptoFactories};
 use tari_crypto::{
     keys::{PublicKey as PublicKeyTrait, SecretKey},
     tari_utilities::ByteArray,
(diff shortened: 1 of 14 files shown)


SWvheerden (dismissed):
LGTM

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5205,11 +5210,11 @@ pub unsafe extern "C" fn wallet_import_utxo(
         return 0;
     }
 
-    if features.is_null() {
-        error = LibWalletError::from(InterfaceError::NullError("features".to_string())).code;
-        ptr::swap(error_out, &mut error as *mut c_int);
-        return 0;
-    }
+    let mut updated_features = if features.is_null() {
```
stringhandler: This is wrong. It should error

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5245,21 +5250,43 @@ pub unsafe extern "C" fn wallet_import_utxo(
 
     let public_script_key = PublicKey::from_secret_key(&(*spending_key));
 
+    // On the blockchain, 'V0' of the output features encode and decode without the recovery byte, whereas 'V1' onwards
+    // adds it in, however, for the wallet, the recovery byte must be consistent with the value commitment to be
+    // accepted in the wallet irrespective if it is 'V0' or 'V1'
+    let recovery_byte =
```
stringhandler: This should not be calculated. It should be imported as provided

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5245,21 +5250,43 @@ pub unsafe extern "C" fn wallet_import_utxo(
...
+                false,
+            )) {
+            Ok(val) => val,
+            Err(e) => {
+                error = LibWalletError::from(OutputManagerError(e)).code;
+                ptr::swap(error_out, &mut error as *mut c_int);
+                return 0;
+            },
+        };
+    updated_features.set_recovery_byte(recovery_byte);
```
stringhandler: If you change the data here, the hash will change

hansieodendaal:
Changed the philosophy about having inconsistent recovery byte data in the wallet database

stringhandler (approved):

PR #3958, opened 2022-03-25
feat: check file perms on identity files

sdbondi:
Description
---
- adds check for o+rw file permissions for node and tor identity files on unix family systems
- adds an error type for common identity module
- improves hint output on failure
- clean up duplicate tor help message in code
- adds "watch status" to rustyline history
- remove get LAN ip functionality, since it is not useful, remove `get_if_addrs` dependency

Motivation and Context
---
Checking for and rejecting open file permissions on identity files is good practice (a la  `~/.ssh`)

How Has This Been Tested?
---
Manually, setting incorrect permissions on node identity and tor identity
Shutting down tor and checking that the base node and wallet emit the correct message


delta1 (dismissed):
Looks good, untested

stringhandler (dismissed):

stringhandler (approved):

PR #3960, opened 2022-03-25
refactor(core): tari newtype wraps MicroTari

sdbondi:
Description
---
- uses MicroTari to back Tari newtype

Motivation and Context
---
This avoids a fair amount of complexity and footguns. I wanted to remove the decimal dependency but it come in handy for the parsing of decimal strings. The only API breaking change is that negatives are not supported (not currently used) but this would be fairly straight forward to add.

How Has This Been Tested?
---
New and updated unit tests, Manually checked the wallet


SWvheerden (dismissed):
LGTM

stringhandler (approved):

PR #3963, opened 2022-03-28
refactor: reduce unwrap calls in the tari_mining_node

therustmonk:
Description
---
The PR removes unnecessary `unwrap` calls from the `tari_mining_node` crate.

Motivation and Context
---
Avoiding potential panics and the code simplification.

How Has This Been Tested?
---
CI


applications/tari_mining_node/src/stratum/controller.rs
```diff
@@ -108,13 +108,11 @@ impl Controller {
     }
 
     fn send_message(&mut self, message: &str) -> Result<(), Error> {
-        if self.stream.is_none() {
-            return Err(Error::Connection(String::from("No server connection")));
-        }
+        let stream = self.stream()?;
         debug!(target: LOG_TARGET_FILE, "sending request: {}", message);
-        let _ = self.stream.as_mut().unwrap().write(message.as_bytes());
```
therustmonk, line 115: Why did we ignore these errors?

sdbondi: :man_shrugging: What you've done looks correct

applications/tari_mining_node/src/stratum/controller.rs
```diff
@@ -77,20 +77,20 @@ impl Controller {
     }
 
     pub fn try_connect(&mut self) -> Result<(), Error> {
-        self.stream = Some(Stream::new());
-        self.stream
-            .as_mut()
-            .unwrap()
-            .try_connect(&self.server_url, self.server_tls_enabled)?;
+        // TODO: Check should we `take` it here?
+        self.stream.take();
```
sdbondi: No need to take from the Option. Assigning to it will Drop the previous instance.

therustmonk: But if `try_connect` failed do we have to drop the previous connection this way?

applications/tari_mining_node/src/stratum/controller.rs
therustmonk: Wow, this change helped to find a potential issue. It could miss some data on writing:

```
error: written amount is not handled. Use `Write::write_all` instead
   --> applications/tari_mining_node/src/stratum/controller.rs:113:9
    |
113 |         stream.write(message.as_bytes())?;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::unused_io_amount)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount
```

sdbondi: Good catch - write will return an error if the socket has to wait to write, previous code would have written junk in some cases.

applications/tari_mining_node/src/stratum/controller.rs
sdbondi: I don't think it's an issue to connect out but think you're right the disconnect may happen after the new connection (I think rust will run the drop code first but it's not explicit or guaranteed). Can use take or 
```rust
// Ensure current connection is disconnected
self.stream = None;
```

sdbondi (dismissed):
Awesome!

stringhandler (approved):

PR #3964, opened 2022-03-28
feat(core): add tip height metric

sdbondi:
Description
---
- adds `base_node::blockchain::tip_height` metric to track current tip height

Motivation and Context
---
The main reason for this is currently the tip height is extracted from the logs using Loki which does not support the `delta` function. `delta` is needed to determine if the chain tip has moved or not within a timespan.

Following query allows an alert to be created that checks that nothing has changed within 20 minutes (delta of 0 over 20 min timespan)
`delta(tari_base_node::blockchain::tip_height[20m])`

How Has This Been Tested?
---
Manually: Metrics appearts in local prometheus instance


stringhandler (approved):

delta1 (approved):
Nice

sdbondi:
This PR gives @delta1 `delta()`

PR #3975, opened 2022-03-30
chore: fix "our blockchain is up-to-date" log

Cifko:
Description
---
Change from `info` to `debug` and change the text if we are ahead.


base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -369,10 +369,16 @@ fn determine_sync_mode(
             sync_peers: sync_peers.into_iter().cloned().map(Into::into).collect(),
         }
     } else {
-        info!(
+        debug!(
```
sdbondi, line 372: Feels like this should stay INFO level - @mikethetike what do you think?

delta1: I agree it should be info

Cifko: Something has to be done. It's spamming the logs. I have a base node running, out of 8585 lines of log lines in `base_layer.log`, there is 7672 with this message.
I can split the code into if else and print debug when we are ahead. Or we can print this message only when it changes (but that require more code modifications). If it would be C language I would just have local static, but it's not :-)

stringhandler: Yeah, I prefer debug. We already have info on status and other logs that provide the block data

stringhandler (approved):

PR #3978, opened 2022-03-30
fix: another fix for the recovery daily test

philipr-za:
Description
---
The daily recovery test was failing to do the webhook curl command due to an unescaped apostrophe in the message.

Also updated the seed words being used to a new one as I cannot yet figure out why the old seed words wont recover but feel its better to monitor new words for now that actually work.

How Has This Been Tested?
---
manually


delta1 (approved):
LGTM

stringhandler (approved):

PR #3982, opened 2022-03-30
feat: make wallet db calls synchronous rather than async

hansieodendaal:
Description
---
Removed spawn blocking calls for db operations from the wallet in the output manager service. (_This is the first PR in a couple of PRs required to implement this fully throughout the wallet code._)

We are using SQLite in WAL mode, which provides concurrent read access and single thread write access for the number of pooled connections configured. Implementing spawn blocking calls on top of that for every db operation is counter productive and easily results in interlock situations within the wallet code, as is the case currently with sending transactions in batch mode.

Although batch mode transactions still does not work with only this PR, it showed a definite improvement monitoring the action with tokio console; all interlocks are now within the transaction service and not in the output manager service anymore.

Motivation and Context
---
When sending multiple transactions in batch mode (with for example the `make-it-rain` command), the wallet locks up and no transactions are being sent.

How Has This Been Tested?
---
System level tests sending multiple transactions in batch mode to three receiving wallets.
Monitoring with tokio console.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/database/mod.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/database/mod.rs
@@ -103,8 +103,7 @@ where T: OutputManagerBackend + 'static
     }
 
     pub fn add_unspent_output(&self, output: DbUnblindedOutput) -> Result<(), OutputManagerStorageError> {
-        let db_clone = self.db.clone();
-        db_clone.write(WriteOperation::Insert(DbKeyValuePair::UnspentOutput(
+        self.db.write(WriteOperation::Insert(DbKeyValuePair::UnspentOutput(
             output.commitment.clone(),
             Box::new(output),
         )))?;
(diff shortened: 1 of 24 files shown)


base_layer/wallet/src/output_manager_service/storage/database/mod.rs
```diff
@@ -102,140 +102,107 @@ where T: OutputManagerBackend + 'static
...
     }
 
-    pub async fn add_output_to_be_received(
+    pub fn add_output_to_be_received(
         &self,
         tx_id: TxId,
         output: DbUnblindedOutput,
         coinbase_block_height: Option<u64>,
     ) -> Result<(), OutputManagerStorageError> {
         let db_clone = self.db.clone();
```
stringhandler: These clones can come out now. There are an artifact from having to `move` into the spawn.

stringhandler (commented):
Happy once the unnecessary `db.clone` calls are removed

stringhandler (approved):

CjS77:
A note after the fact.

I was curious to see how removing spawn blocking calls (with the assumption that they would become non-blocking) would help with locks, so I looked at this PR a little more closely.

Turns out that it's not what this PR does. Rather it's changing the DB requests from async to synchronous calls.

So it's even more blocking than before :) (but likely would help with locks).

I'm just updating the PR title to reflect this, to help identify the root purpose of the PR

hansieodendaal:
Personally, I would be more confused by the new title. I would have been happy with `feat: let sql in wal mode provide async db behaviour rather than application level spawn blocking (output manager service)`

PR #3993, opened 2022-03-31
refactor: dry grpc services of the validator node

therustmonk:
Description
---
These changes remove unwrapping grpc connections from options in every method.

Motivation and Context
---
Reducing amount of `unwrap` calls, replace with `Result` when possible.

How Has This Been Tested?
---
CI


stringhandler (approved):

therustmonk:
These change were moved to the #3998 PR, since some conflicts can't be solved easily after we added an extra field to the `BaseNodeClient` struct.

PR #3995, opened 2022-04-03
chore: obscure grpc error response

Cifko:
Description
---
Obscure errors returned by grpc on base node.  The flag for obscuring is coming from config (default is false).

### DIFF
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -92,6 +92,7 @@ pub struct BaseNodeGrpcServer {
     software_updater: SoftwareUpdaterHandle,
     comms: CommsNode,
     liveness: LivenessHandle,
+    report_grpc_error: bool,
 }
 
 impl BaseNodeGrpcServer {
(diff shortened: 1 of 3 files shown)


sdbondi:
Looks good, but I feel like it should be the other way around - you should opt in to sending error messages rather than opt out. maybe something like `grpc_enable_debug_responses`

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -104,8 +105,21 @@ impl BaseNodeGrpcServer {
...
     }
+
+    pub fn obscure_error_flag(&self) -> bool {
+        self.obscure_grpc_error
+    }
+}
+
+pub fn obscure_error(obscure: bool, status: Status) -> Status {
+    if obscure {
+        Status::unknown("Error has occurred. Details are obscured.")
```
stringhandler: The status code (int) should still be sent, just not the message

stringhandler (commented):
Happy, but the status code should still be propagated, only the message and details hidden

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
Cifko: Done.

Cifko:
> Looks good, but I feel like it should be the other way around - you should opt in to sending error messages rather than opt out. maybe something like `grpc_enable_debug_responses`

I rewrote it to `report_error` (default is false).

stringhandler (dismissed):

stringhandler (approved):

PR #3996, opened 2022-04-04
fix: estimated hashrate calculation is incorrect

mrnaveira:
Description
---
* Created new fields in the base_node grpc api for sha3 and monero hash rates, using an independent moving average for each one
* Kept the `estimated_hash_rate` field but now it corresponds to the sum of the sha3 and monero hash rates, so existing clients (i.e. web explorer) will not break
* Modified the text explorer to show the estimated hash rates

Motivation and Context
---
The current hash rate calculation is incorrect, we want to use a separated moving average for each PoW algo independently

How Has This Been Tested?
---
Unit testing for the moving average hash rate calculation


applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -170,9 +170,11 @@ message NewBlockTemplateRequest{
 message NetworkDifficultyResponse {
     uint64 difficulty = 1;
     uint64 estimated_hash_rate = 2;
-    uint64 height = 3;
-    uint64 timestamp = 4;
-    uint64 pow_algo = 5;
+    uint64 sha3_estimated_hash_rate = 3;
```
stringhandler: Rather add these fields at the end. numbers in protobuf should not change. 
```suggestion
    uint64 sha3_estimated_hash_rate = 6;
    uint64 height =3;
```

applications/tari_base_node/src/grpc/hash_rate.rs
```diff
@@ -0,0 +1,198 @@
...
+use std::collections::VecDeque;
+
+use tari_core::{
+    consensus::ConsensusManager,
+    proof_of_work::{Difficulty, PowAlgorithm},
+};
+
+/// The number of past blocks to be used on moving averages for (smooth) estimated hashrate
+/// The hash rate is calculated as the difficulty divided by the target block time
+const SHA3_HASH_RATE_MOVING_AVERAGE_WINDOW: usize = 12;
```
stringhandler, line 32: Need a comment here for why these values are different

applications/tari_base_node/src/grpc/hash_rate.rs
```diff
@@ -0,0 +1,198 @@
...
+            window_size,
+            hash_rates,
+            average: 0,
+        }
+    }
+
+    /// Adds a new hash rate entry in the moving average and recalculates the average
+    pub fn add(&mut self, height: u64, difficulty: Difficulty) {
+        // target block time for the current block is provided by the consensus rules
+        let target_time = self
```
stringhandler, line 64: Nice

applications/tari_base_node/src/grpc/hash_rate.rs
```diff
@@ -0,0 +1,198 @@
...
+        if self.hash_rates.is_empty() {
+            return 0;
+        }
+
+        let sum: u64 = self.hash_rates.iter().sum();
+        let count = self.hash_rates.len() as u64;
+        sum / count
+    }
+
+    pub fn get_average(&self) -> u64 {
```
stringhandler:
```suggestion
    pub fn average(&self) -> u64 {
```

stringhandler (commented):
Nice work. Happy to merge if you keep the protobuf numbers in NetworkDifficultyResponse

mrnaveira:
Implemented all suggested changes

stringhandler (approved):

PR #3999, opened 2022-04-05
test: write test for input, output and kernel malleability

mrnaveira:
Description
---
* Refactored existing **input** malleability test and added more tests for other input fields
    * Discovered that `version` is malleable. It's not included in the input hash but it should be to avoid this.
    * In general, the input hash does not follow the consensus encoding specified in [RFC-0121](https://rfc.tari.com/RFC-0121_ConsensusEncoding.html#transaction-input).
* Created a new unit test module that modifies the **output** of a block to check that alterations are detected by hashes
    * Discovered that `sender_offset_public_key` is malleable. It's not included in the output hash but it should be to avoid this.
    * Had to give pub visibility to a covenant related enum (`ConvenantToken`), it should be public anyways (thanks @sdbondi).
* Created a new unit test module that modifies the **kernel** of a block to check that alterations are detected by hashes.
    *  A couple of fields (`version` and `features`) could not be tested because they only accept a single constant value at the moment.

Motivation and Context
---
We want to make sure that input, output and kernel malleability are not allowed

How Has This Been Tested?
---
The new unit tests pass, but a couple of new tests  had to be ignored as they highlight that input and output hash should be fixed to include missing fields (`input.version` and `output.sender_offset_public_key`). I left those consensus changes out of the scope of this PR.


sdbondi (dismissed):
Very nice!

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -116,6 +116,11 @@ impl AggregateBody {
         &self.inputs
     }
 
+    /// Should be used for tests only. Get a mutable reference to the inputs
+    pub fn inputs_mut(&mut self) -> &mut Vec<TransactionInput> {
```
stringhandler, line 125:
```suggestion
   #[cfg(test)]
    pub fn inputs_mut(&mut self) -> &mut Vec<TransactionInput> {
```

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -131,9 +136,9 @@ impl AggregateBody {
         &self.kernels
     }
 
-    /// Should be used for tests only. Get a mutable reference to the inputs
-    pub fn inputs_mut(&mut self) -> &mut Vec<TransactionInput> {
-        &mut self.inputs
+    /// Should be used for tests only. Get a mutable reference to the kernels
+    pub fn kernels_mut(&mut self) -> &mut Vec<TransactionKernel> {
```
stringhandler, line 145:
```suggestion
   #[cfg(test)]
    pub fn kernels_mut(&mut self) -> &mut Vec<TransactionKernel> {
```

base_layer/core/src/transactions/transaction_components/mod.rs
```diff
@@ -100,13 +101,21 @@ pub(super) fn hash_output(
...
         .chain(features)
         .chain(commitment)
         .chain(script)
-        .chain(covenant)
-        .finalize()
+        .chain(covenant);
+
+    // previous versions were incorrectly ignoring the "sender_offset_public_key" in the hash
+    // so to maintain backwards compatibility we include it from a new "V2" version onwards
+    if version >= TransactionOutputVersion::V2 {
```
stringhandler: nice find

base_layer/core/src/transactions/transaction_components/transaction_input.rs
```diff
@@ -342,6 +351,12 @@ impl TransactionInput {
                     .chain(&self.input_data)
                     .chain(covenant);
 
+                // previous versions were incorrectly ignoring the input "version" in the hash
+                // so to maintain backwards compatibility we include it from a new "V2" version onwards
+                if self.version >= TransactionInputVersion::V2 {
```
stringhandler: This is correct, but I would prefer to remove the whole `cannonical_hash` method in favour of ConsensusEncode.

It's probably best to remove this addition and fix cannonical_hash properly in a different PR

stringhandler (commented):
The PR is good, but you changes highlight that the hashing of inputs (at least) are not aligned with the RFC. I think we should correct the hashing for TransactionInputVersion2 and merge it as one fix

base_layer/core/src/transactions/transaction_components/mod.rs
mikethetike: This is actually fine, the sender_offset_public_key is included in the witness hash

base_layer/core/src/transactions/transaction_components/transaction_input.rs
mrnaveira: Agree. I removed all consensus changes out of this PR to keep the scope small. It should be fixed in a different PR.

base_layer/core/src/transactions/transaction_components/mod.rs
mrnaveira: To keep the scope of this PR only on malleability tests, I reverted all consensus breaking changes (both for inputs and outputs). I believe they all should be done in a separate PR.

base_layer/core/src/transactions/aggregated_body.rs
mrnaveira: Sadly, the `#[cfg(test)]` macro cannot be currently added to those methods, as it makes them unavailable to integration tests. We can consider refactoring the tests or creating a cargo feature to workaround this limitation.

base_layer/core/src/transactions/aggregated_body.rs
mrnaveira: Sadly, the `#[cfg(test)]` macro cannot be currently added to those methods, as it makes them unavailable to integration tests. We can consider refactoring the tests or creating a cargo feature to workaround this limitation.

stringhandler (dismissed):

sdbondi (approved):

PR #3777, opened 2022-01-31
fix: make launchpad work with development

stringhandler:
Description
---
Update launchpad to work on windows and with the latest development. 
Also fixes the default configs for merge mining and mining node

Motivation and Context
---
Code was out of date after validator_node branch merge

How Has This Been Tested?
---
manually, cargo test

PR #4005, opened 2022-04-06
refactor!: config structs are directly mapped onto toml config

sdbondi:
Description
---

This PR builds from work done on #3777 by @philipr-za @mikethetike @delta1 @CjS77 

- refactors configuration to map directly to config structs on all apps
- refactor and simplify p2p configs
- split out various transport configs
- split out peer seeds from other p2p config so that it is common to all apps and overridable per network

FFI Breaking changes:
- `transport_type_destroy` renamed to `transport_config_destroy`. `transport_type_destroy` is still available but deprecated.
- uses TransportConfig instead of TransportType struct
- `network` argument removed from `comms_config_create`

Motivation and Context
---
Previous config was brittle and could easily break app configs when they are changed. This PR greatly reduces config maintenance by removing any mapping and default setting when changing configs. 

TODOs:
- update launchpad
- integration tests (in progress)

How Has This Been Tested?
---
Manually, testing each app
Existing tests
Code compiles


sdbondi:
Closing to bring back original commits

PR #4015, opened 2022-04-11
chore: remove #[ignored] tag from working tests

Cifko:
Description
---
Remove ignored test. 

How Has This Been Tested?
---
By running each test 3+ times. To make sure it's not flaky.


base_layer/p2p/src/peer_seeds.rs
```diff
@@ -199,7 +199,6 @@ mod test {
         use super::*;
         use crate::{dns::mock, DEFAULT_DNS_NAME_SERVER};
 
-        #[ignore = "This test requires network IO and is mostly useful during development"]
```
sdbondi, line 237: This test requires the DNS to be available and have some seeds in the TXT record which is why I ignored it. But we can keep it in and if `seeds.weatherwax.tari.com` ever goes away we'll fix it then.

sdbondi (approved):

PR #4023, opened 2022-04-11
fix: launchpad docker improvements

leet4tari:
Description
Fix docker tag and add notes for overrides
Bump monerod and xmrig versions
Add sub-version numbers for XMrig and MoneroD images
XMRig - Subversion and env version builds
MoneroD -  - Subversion and env version builds
Add check for required env DATA_FOLDER
Quote mixed Name volume paths - Use env for image repos, with defaults
Add env for docker compose path/runtime

Motivation and Context
Make lauchpad docker images easier to build and use

How Has This Been Tested?
Build locally and on Ubuntu VM


CjS77 (commented):
Looks good, but you need to include the options you've introduced into the tag we push to the repository.

So something like `base_node:version.arch.features`
Must be one word though.

Arch is always Linux, since it's docker. Are there cases where it would not be "native"?

leet4tari:
Thanks @CjS77 

Native is CPU ARCH as per rust compile flags. Main goal, which I might not have added details on, was making the docker image hosting more compatible (Mac 2014 support), reducing the CPU optimizations and making it as easy as possible for somebody to build locally with "native", while offer good and one liner defaults.

Not quite sure where you looking for the tagging?

I am keen on looking into ARM for the M1 support, but that might need quite a bit more investigation and support.

leet4tari:
Just adding a note about CPU compile flags - https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

CjS77:
In the build.sh file, you set the image tag with `-t tag`. The images we push must be differentiated by the various config options used in the build.

`latest` should be aliased to the most commonly used build

applications/launchpad/build_images.sh
```diff
@@ -1,28 +1,89 @@
...
+
+#TBN_ARCH=native
+TBN_ARCH=${TBN_ARCH:-x86-64}
+#TBN_FEATURES=avx2
+TBN_FEATURES=${TBN_FEATURES:-safe}
+
+# Docker tag URL
+TL_TAG_URL=${TL_TAG_URL:-quay.io/tarilabs}
+# Disable docker push if 0
+TL_TAG_PUSH=${TL_TAG_PUSH:-1}
```
stringhandler: let's make this default 0 rather for now

applications/launchpad/docker_rig/Readme.txt
```diff
@@ -0,0 +1,148 @@
...
+```
+
+Might be a chance that base_node id were not create or the first step missed, and there is a problem like
+```
+Y/n
+```
+In this case, stop the base_node with 
+```
+docker-compose stop base_node
+docker-compose run --rm base_node --create-id --init
```
stringhandler:
```suggestion
docker-compose run --rm base_node --init
```

applications/launchpad/docker_rig/config.toml
```diff
@@ -37,7 +37,8 @@ peer_seeds = [
 [base_node]
 network = "dibbler"
```
stringhandler: This whole file needs to be replaced. I think let's merge it as is, and then we can update it

applications/launchpad/docker_rig/Readme.txt
leet4tari: This is first run, before there is an ID. base_node will ask to create ID and wait forever for console input. I will test, but the above is note for init run only.

applications/launchpad/build_images.sh
leet4tari: done

applications/launchpad/docker_rig/Readme.txt
leet4tari: Updated tag to include arch and feature.

Removed latest linking and pushing.

leet4tari: Just a note, we should not use native for cloud images, because native probes local CPU for features. We could use a set CPU set, which is part of what I did in the build_images.sh

REPOSITORY                           TAG                                IMAGE ID       CREATED          SIZE
local-tarilabs/tari_sha3_miner       0.31.1-lp1-native-avx2             6c9742eb095b   4 minutes ago    140MB
local-tarilabs/tari_mm_proxy         0.31.1-lp1-native-avx2             c3a9d68491e9   12 minutes ago   124MB
local-tarilabs/tari_console_wallet   0.31.1-lp1-native-avx2             d21011b7b907   21 minutes ago   208MB
local-tarilabs/tari_base_node        0.31.1-lp1-native-avx2             497db468503a   31 minutes ago   215MB

leet4tari: Wondering if the tag should not also include the git commit hash?

applications/launchpad/build_images.sh
```diff
@@ -1,28 +1,82 @@
...
+if [[ "${TL_TAG_PUSH}" == "1" ]]; then
+  docker push ${TL_TAG_URL}/monerod:"${TL_VERSION}-v${MONERO_VERSION}-${TBN_ARCH}-${TBN_FEATURES}"
+fi
+
+echo "Building XMRig image v$TL_VERSION (XMRig v$XMRIG_VERSION) - $TL_VERSION-$XMRIG_VERSION"
+docker build -f docker_rig/xmrig.Dockerfile --build-arg VERSION="$TL_VERSION-$XMRIG_VERSION" \
+  --build-arg XMRIG_VERSION=$XMRIG_VERSION \
+  -t ${TL_TAG_URL}/xmrig:"${TL_VERSION}-${XMRIG_VERSION}-${TBN_ARCH}-${TBN_FEATURES}" ./../..
+if [[ "${TL_TAG_PUSH}" == "1" ]]; then
+  docker push ${TL_TAG_URL}/xmrig:"${TL_VERSION}-${XMRIG_VERSION}-${TBN_ARCH}-${TBN_FEATURES}"
```
CjS77, line 75: Remove `features` from tag, since this is not a rust image

applications/launchpad/build_images.sh
```diff
@@ -1,28 +1,82 @@
...
-
-docker build -f docker_rig/frontail.Dockerfile -t quay.io/tarilabs/frontail ./docker_rig
-docker tag quay.io/tarilabs/frontail:latest quay.io/tarilabs/frontail:$VERSION
-docker push quay.io/tarilabs/frontail:latest
-docker push quay.io/tarilabs/frontail:$VERSION
+
+echo "Building Monerod image v$TL_VERSION (Monerod v$MONERO_VERSION) - $TL_VERSION-v$MONERO_VERSION"
+docker build -f docker_rig/monerod.Dockerfile --build-arg VERSION="$TL_VERSION-v$MONERO_VERSION" \
+  --build-arg MONERO_VERSION=$MONERO_VERSION --build-arg MONERO_SHA256=$MONERO_SHA256  \
+  -t ${TL_TAG_URL}/monerod:"${TL_VERSION}-v$MONERO_VERSION-${TBN_ARCH}-${TBN_FEATURES}" ./../..
```
CjS77, line 65: Remove `features` from tag; this is not a Rust build

CjS77: Is this really a cross-platform image? If not, we should not include the `arch` in the tag either.

CjS77 (commented):
Picked up additional things and have commented.

In general, 

1.  It's cool to try get `docker-compose.yml` consistent, but this was really a PoC ahead of writing launchpad, so we can actually deprecate it. Thus I wouldn't spend too much effort trying to get it to work.
2. Same for `quick-start.sh`
3. I think we should _not_ try for multi-platform builds in this PR, since there are still some unknowns. Let's just have `linux/arm64` work 100% for this PR and then figure out multiplatform in a follow up.

applications/launchpad/build_images.sh
leet4tari: My initial idea was to match builds of third-party docker images with the build of TariLab suite at the time or maybe be able to build set versions, but still match the version to TL suite.  
  
```arch``` is X86 CPU feature set, not even selection between ARM vs X86.  
  
I think getting some of the bits together in a working state is useful and can be moved into more useful parts.

CjS77:
@leet4tari -- This PR was covered by #4076 right? So we can close this?

leet4tari:
Can be closed, I think too much has changed. I will pull texts and some info later.

PR #4031, opened 2022-04-13
test: uncomment DAN layer tests

mrnaveira:
Description
---
* Uncommented and adapted `test_simple_case` to be able to compile. The test was not doing anything due to a missing `await` in `start_replica`. The test must be kept ignored as there are many unimplemented features that are needed.
* Uncommented and adapted `basic_test_as_leader`, same as previous one, it needs to be kept ignored.
* Created multiple new mocks to support the tests

Motivation and Context
---
There were some ignored and commented out tests

How Has This Been Tested?
---
Code compiles. The tests are ignored and will not pass if ran, due to missing features in the DAN layer.


dan_layer/core/src/workers/consensus_worker.rs
```diff
@@ -406,42 +407,84 @@ mod test {
...
     };
 
     fn start_replica(
-        _inbound: MockInboundConnectionService<&'static str, &'static str>,
-        _outbound: MockOutboundService<&'static str, &'static str>,
-        _committee_manager: MockCommitteeManager,
-        _node_id: &'static str,
-        _shutdown_signal: ShutdownSignal,
-        _events_publisher: MockEventsPublisher<ConsensusWorkerDomainEvent>,
+        inbound: MockInboundConnectionService<RistrettoPublicKey, TariDanPayload>,
```
stringhandler, line 434: I intentionally tried to use strings here instead of the more complicated structs, if they can't be used, it means there is a type that uses a PublicKey instead of a generic somewhere

stringhandler (commented):
Thanks for fixing this. I'd prefer to keep the implementation with strings unless we really can't do it

dan_layer/core/src/workers/consensus_worker.rs
mrnaveira: The root cause of needing PublicKey is the ConsensusWorker implementation:

`impl<TSpecification: ServiceSpecification<Addr = PublicKey>> ConsensusWorker<TSpecification> {`

stringhandler (dismissed):

stringhandler (approved):

PR #4037, opened 2022-04-14
fix: weird behaviour of dates in base node banned peers

mrnaveira:
Description
---
* Changed the peer display format to output `offline_at` and `banned_until` as local time instead of UTC, fixing the “time in the past” issue
* Improved handling and display of permanent bans (when invoking `ban-peer` with no params). Note that this makes _existing_ permabans still display weird times, changes affect only new permabans
* Scope of the changes does seem to only affect log messages in other applications

Motivation and Context
---
There is some weird behaviour regarding display of date times on banned peers.
- Times are shown in UTC but without the timezone, so if the user is in UTC+N the ban end time may look like it's "in the past" for some values
- The default ban time is permanent, and this makes the "banned until" display the max datetime.

How Has This Been Tested?
---
Manually tested on a base node, by banning peers with different values

### DIFF
--- a/comms/core/src/utils/datetime.rs
+++ b/comms/core/src/utils/datetime.rs
@@ -22,15 +22,13 @@
 
 use std::time::Duration;
 
-use chrono::{DateTime, NaiveTime, Utc};
+use chrono::{DateTime, Local, NaiveDateTime, Utc};
 
 pub fn safe_future_datetime_from_duration(duration: Duration) -> DateTime<Utc> {
     let old_duration = chrono::Duration::from_std(duration).unwrap_or_else(|_| chrono::Duration::max_value());
-    Utc::now().checked_add_signed(old_duration).unwrap_or_else(|| {
-        chrono::MAX_DATE
-            .and_time(NaiveTime::from_hms(0, 0, 0))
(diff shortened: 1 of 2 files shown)


sdbondi (dismissed):
Looks good!

comms/core/src/utils/datetime.rs
```diff
@@ -48,6 +46,15 @@ pub fn format_duration(duration: Duration) -> String {
     }
 }
 
+pub fn format_local_datetime(datetime: &NaiveDateTime) -> String {
+    let local_datetime = DateTime::<Local>::from_utc(*datetime, Local::now().offset().to_owned());
+    format!("{}", local_datetime.format("%Y-%m-%d %H:%M:%S"))
```
sdbondi: nit:
```suggestion
    local_datetime.format("%Y-%m-%d %H:%M:%S")).to_string()
```

mrnaveira: good one, thanks

sdbondi (approved):

SWvheerden (approved):
Looks good to me
Spent about 5 mins on this.
Tested and run locally

PR #4038, opened 2022-04-14
refactor(dht): use CipherKey new type for diffie-hellman key 

sdbondi:
Description
---
- Document the DHT public interface
- Move forward layer to `outbound` module
- Remove unnecessary allocation from `encrypt` helper
- Return `CipherKey` new type from Diffie-Hellman helper function 
- Changes `encrypt` and `decrypt` to take in `CipherKey` newtype
- implement zeroize on drop for `CipherKey` new type

Motivation and Context
---
Documentation. 

`generate_ecdh_secret` returned a public key, which implies that it is safe to share publicly.
The `CipherKey` new type will zero it's contents before releasing it's memory.

The forward layer didn't have anything to do with SAF so shouldn't be located in that module.

How Has This Been Tested?
---
Existing tests pass
Manually, no breaking changes.

cargo doc --no-deps


hansieodendaal:
I started looking at this PR, not sure if the title is doing justice to the contents of the PR? As it is a mixed PR, I would rather put the emphasis on the code changes.

stringhandler (dismissed):
Nice

stringhandler (approved):

PR #4039, opened 2022-04-14
fix: only count base nodes in peers count in base node status

mrnaveira:
Description
---
Filter connection count in status command to only count base nodes

Motivation and Context
---
We only want to show connection count to base nodes in the base node status line. Currently it counts all connected peers regarding of type.

How Has This Been Tested?
---
Manually tested by listing all connections and then comparing it to the status line to see that it filters base nodes

### DIFF
--- a/applications/tari_base_node/src/commands/command/status.rs
+++ b/applications/tari_base_node/src/commands/command/status.rs
@@ -27,6 +27,7 @@ use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use clap::Parser;
 use tari_app_utilities::consts;
+use tari_comms::connectivity::ConnectivitySelection;
 
 use super::{CommandContext, HandleCommand};
 use crate::commands::status_line::{StatusLine, StatusLineOutput};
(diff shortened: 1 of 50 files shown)


applications/tari_base_node/src/commands/command/status.rs
```diff
@@ -94,7 +94,8 @@ impl CommandContext {
         );
 
         let conns = self.connectivity.get_active_connections().await?;
```
sdbondi: You can do this using 
```rust
let conns = self.connectivity.select_connections(ConnectivitySelection::all_nodes(vec![])).await?;
```
without having to filter it yourself

mrnaveira: totally right, I missed that method. I just updated the PR.

sdbondi (dismissed):

sdbondi (approved):

SWvheerden (approved):
Spent about 5 mins on this, 
Pulled and tested locally with a newly synced node.

stringhandler (approved):

PR #4040, opened 2022-04-14
fix(wallet): do not prompt for password if given in config

sdbondi:
Description
---
- uses cli password followed by configured password if specified

Motivation and Context
---
Should not prompt for password if given in config

How Has This Been Tested?
---
Manually


delta1 (approved):
lgtm

stringhandler (approved):

PR #4041, opened 2022-04-14
fix: launchpad build docker images

alex1307:
Description
Building docker images fail on Mac M1 Pro with error 137. 

Motivation and Context
Make launchpad docker images buildable and running.
Replaced the original `Tari` image with `rust:1.60-buster` and configured properly the build env.

How Has This Been Tested?
Build and test locally on Mac M1 Pro and Ubuntu 20.04 (LTS)x64


alex1307:
The PR contains wrong commits.

PR #4045, opened 2022-04-19
fix: makes header consensus encoding infallible

sdbondi:
Description
---

- removes `BlockHeader::consensus_encode` impl  errors

Motivation and Context
---
The ConnsensusEncode contract states that the implementation should never be the source of errors, only the writer may return errors.

How Has This Been Tested?
---
Existing tests
manually, basenode sync


SWvheerden (approved):
utACK

stringhandler (approved):

PR #4048, opened 2022-04-23
refactor(base-node): use existing peer feature methods to check if is a base or client node

mrnaveira:
Description
---
Refactored the `list-peers` and `list-connections` base node console commands to reuse the existing peer feature methods to check if is a base or client node.

Motivation and Context
---
Up until now,  the `list-peers` and `list-connections` base node console commands check if a node is a client/node by comparing values against enums, but there are already some convenient methods to check that.

How Has This Been Tested?
---
We don't have unit or integrations tests on console commands, so I tested by manually running the `list-peers` and `list-connections` to check if the results are correct.

### DIFF
--- a/applications/tari_base_node/src/commands/command/list_connections.rs
+++ b/applications/tari_base_node/src/commands/command/list_connections.rs
@@ -23,7 +23,6 @@
 use anyhow::Error;
 use async_trait::async_trait;
 use clap::Parser;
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
utAck

therustmonk (approved):

stringhandler (approved):

PR #4061, opened 2022-04-28
docs: add bearer token rfc

stringhandler:
An RFC for DAN layer bearer tokens

### DIFF
--- a/RFC/src/RFC_0388_BearerTokens.md
+++ b/RFC/src/RFC_0388_BearerTokens.md
@@ -65,17 +65,6 @@ DAN layer.
 This scheme is inspired by [Macaroons] in that it allows the bearer of a token to delegate a more restrictive token to
 another bearer. 
 
-## Quick Example: Invoking an instruction
-
-When calling a transfer, Alice creates a token for Dexter. 
-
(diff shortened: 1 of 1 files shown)


RFC/src/RFC_0388_BearerTokens.md
```diff
@@ -0,0 +1,130 @@
...
+Before starting execution of the instruction, the list of auth tokens must be validated to ensure that each token is more restrictive than
+the last.
+
+When executing instructions, caveats MUST be checked if they are relevant to the resources being acted upon. 
+For example, if a function requires a scope `transfer`, the most specific AuthToken must have that scope present.
+
+
+
+## Revoking Tokens
+Each contract SHOULD store a root nonce for each identity in the contract. To revoke a set of tokens, an identity owner
```
sdbondi, line 119: Another way we could do this is the token issuer stores root nonce/s and the token expiry . When revoking, the token issuer submits a revoke instruction with nonces and an expiry. The validator node stores the nonces until the expiry.

stringhandler:
Reopened here: https://github.com/tari-project/rfcs/pull/1

PR #4062, opened 2022-04-29
docs: confidential assets

Cifko:
Description
---
Confidential assets markdown document.

### DIFF
--- a/RFC/src/confidential_assets.md
+++ b/RFC/src/confidential_assets.md
@@ -1,0 +1,66 @@
+# Confidential assets
+
+The commitment is vH + rG + tF + sE
+
+- v - amount
+- r - blinding factor
+- t - token id
+- s - asset id
+
+## Getting tokens:
+
+---
+
+- minting
+
+```
+  message MintRequest {
+    bytes asset_id = 1;
+    bytes token_id = 2;
+    bytes signature = 3;
+    bytes commitment = 4;
+    bytes proofs = 5;
+  }
+```
+
+- asset_id - has to be clear for constitution check
+- token_id - same as above
+- signature - Schnorr signature for the key associated with the asset_id, we don't need to send the public key, VN should now it.
+- commitment
+- proofs:
+  - amount is in the range defined in the constitution (especially for NFTs the equality to one)
+
+## Sending FT
+
+---
+
+```
+  message TransferRequest {
+    bytes from = 1;
+    bytes to = 2;
+    bytes input_commitment = 3;
+    bytes output_commitment = 4;
+    repeated bytes excess_commitment = 5;
+    bytes proofs = 6;
+  }
+```
+
+- from - schnorr signature for the input commitment (we don't need to send the public_key, do we?)
+- to - destination public_key
+- input_commitment
+- output_commitment
+- excess_commitment - the change from the transaction
+- proofs :
+  - v > 0 in commitment
+  - v > 0 in the excess_commitment, in case the excess_commitment is there
+
+## Sending NFT
+
+---
+
+The commitment is the same as FT, but with fixed value of **v**.
+
+# Concerns
+
+- Should VN know which token_id are being minted? For NFT the check as to be there, but for FT they should probably not care
+- VN should know which asset is being transferred (they need to check the constitution)


SWvheerden (commented):
I writing this as a general comment in the direction of this RFC. 
As the RFC is currently written, is very much written as UTXO based, with he proposed sending, it does kina leak some info about the confidentially that we would many like to keep. 

What I would propose is that we can kinda add as many extra generators to the extended commitment as we want. We could add an account as well. This will kinda force every single tx to do conjoin every single time they do something, then we can just use plain old maths to make sure it all balances out, especially if we use fairly large scales(aka almost private key like) for the id's. 

Using MW type maths,
lets add an `aJ`, for account over generator J.
So now we have : `vH + rG + tF + sE + aJ`

Lets take a transfer example of Alice -> Bob
Alice has 2 tokens, with id: 500, 871. 
The asset ID is: 657
And her account is 876. 

Now she wants to transfer token 500 to Bob with account ID, 765.  Bob currently has 0 tokens, but has an account. 

The validator node can check this tx by doing:
(Allice UTXO_1) + (Bob UTXO_1)  - (Allice UTXO_2) - (Bob UTXO_2) = 0H + 0F + 0E + 0J + (r1-r2)G

This way the VN knows that no new account was created, no new token was created, only some token moved accounts.
For something like a NFT, we can even drop the value. As long as every NFT has an unique generated id.

Cifko:
> I writing this as a general comment in the direction of this RFC. As the RFC is currently written, is very much written as UTXO based, with he proposed sending, it does kina leak some info about the confidentially that we would many like to keep.
> 
> What I would propose is that we can kinda add as many extra generators to the extended commitment as we want. We could add an account as well. This will kinda force every single tx to do conjoin every single time they do something, then we can just use plain old maths to make sure it all balances out, especially if we use fairly large scales(aka almost private key like) for the id's.
> 
> Using MW type maths, lets add an `aJ`, for account over generator J. So now we have : `vH + rG + tF + sE + aJ`
> 
> Lets take a transfer example of Alice -> Bob Alice has 2 tokens, with id: 500, 871. The asset ID is: 657 And her account is 876.
> 
> Now she wants to transfer token 500 to Bob with account ID, 765. Bob currently has 0 tokens, but has an account.
> 
> The validator node can check this tx by doing: (Allice UTXO_1) + (Bob UTXO_1) - (Allice UTXO_2) - (Bob UTXO_2) = 0H + 0F + 0E + 0J + (r1-r2)G
> 
> This way the VN knows that no new account was created, no new token was created, only some token moved accounts. For something like a NFT, we can even drop the value. As long as every NFT has an unique generated id.

The UTXO will not be checked for existence? I mean Bob UTXO_1 doesn't exists, right? The `v=0` for this UTXO, it's there just to balance the accounts `aJ`, right? Also Alice UTXO_2 is `v=0` for the same reason. The VN will never know if `v=0` (or we can provide proof that is `v>0`) so we will not store UTXO that are zero. 
How does all the proofs look like? Alice need to provide proof that she know `a` for her account (Otherwise she can spent the Bob UTXO_2 because she know everything but the `a` value, right?) . Where does the `(r1-r2)` comes from? shouldn't it be `2*(r1-r2)` if `r1` is blinding factor for the inputs, and `r2` is for the outputs. Or zero if `r1` is for Alice and `r2` is from Bob. But it's non-interactive, so I don't know how Alice would get hands on the r2 (some generator with use of Bobs public key?).

SWvheerden:
> The UTXO will not be checked for existence? I mean Bob UTXO_1 doesn't exists, right? The v=0 for this UTXO, it's there just to balance the accounts aJ, right?

Bob UTXO_1 needs to exist for this to work correctly. He needs to open an account with the VNC, and they can mint him an empty UTXO for him (0V +  rG + 0F + 0E + aJ), He signs some sig with R, and the VNC can check that aJ is correct. 
Checking: `UTXO_1 = (0V +  0F + 0E + aJ)  + (rG)`

> How does all the proofs look like? Alice need to provide proof that she know a for her account (Otherwise she can spent the Bob UTXO_2 because she know everything but the a value, right?) 

No, She does not know r, v, or a from Bob's UTXO_2. She needs to sign with r thou to prove knowledge. She does know the difference in V's between Bob's UTXO_1 and UTXO_2. We need to fish this out further to see if it is possible to steal it faking by an r and making it balance, at first glance it does not seem like it, but we need to confirm. If we need to prove a, we could easily prove a by signing with it as well.

> Where does the (r1-r2) comes from? shouldn't it be 2*(r1-r2) if r1 is blinding factor for the inputs, and r2 is for the outputs. Or zero if r1 is for Alice and r2 is from Bob. 

Sorry this was me being lazy. Its the diff between the r values, so its (Alice_UTXO_1_r + Bob_UTXO_1_r -Alice_UTXO_2_r - Bob_UTXO_2_r). 

> But it's non-interactive, so I don't know how Alice would get hands on the r2 (some generator with use of Bobs public key?).

Yeah as this kinda stands it's not, but it is fully confidential. I don't know if we can modify it to be non-interactive and still be fully confidential? ie leaking some values to the VNC.

Cifko:
> > The UTXO will not be checked for existence? I mean Bob UTXO_1 doesn't exists, right? The v=0 for this UTXO, it's there just to balance the accounts aJ, right?
> 
> Bob UTXO_1 needs to exist for this to work correctly. He needs to open an account with the VNC, and they can mint him an empty UTXO for him (0V + rG + 0F + 0E + aJ), He signs some sig with R, and the VNC can check that aJ is correct. Checking: `UTXO_1 = (0V + 0F + 0E + aJ) + (rG)`

This can probably work if we are talking about single FT, but with NFT (or even with multiple FTs on the same asset), you would need UTXO for every token. 

> 
> > How does all the proofs look like? Alice need to provide proof that she know a for her account (Otherwise she can spent the Bob UTXO_2 because she know everything but the a value, right?)
> 
> No, She does not know r, v, or a from Bob's UTXO_2. She needs to sign with r thou to prove knowledge. She does know the difference in V's between Bob's UTXO_1 and UTXO_2. We need to fish this out further to see if it is possible to steal it faking by an r and making it balance, at first glance it does not seem like it, but we need to confirm. If we need to prove a, we could easily prove a by signing with it as well.
> 
> > Where does the (r1-r2) comes from? shouldn't it be 2*(r1-r2) if r1 is blinding factor for the inputs, and r2 is for the outputs. Or zero if r1 is for Alice and r2 is from Bob.
> 
> Sorry this was me being lazy. Its the diff between the r values, so its (Alice_UTXO_1_r + Bob_UTXO_1_r -Alice_UTXO_2_r - Bob_UTXO_2_r).
> 
> > But it's non-interactive, so I don't know how Alice would get hands on the r2 (some generator with use of Bobs public key?).
> 
> Yeah as this kinda stands it's not, but it is fully confidential. I don't know if we can modify it to be non-interactive and still be fully confidential? ie leaking some values to the VNC.

SWvheerden:
> This can probably work if we are talking about single FT, but with NFT (or even with multiple FTs on the same asset), you would need UTXO for every token.

No, I was thinking of summing them up. Hence me stating that the IDs need to be massive random scalars. They should not be incrementing normal numbers. Then the ids become almost like aggregated keys.

Cifko:
> > This can probably work if we are talking about single FT, but with NFT (or even with multiple FTs on the same asset), you would need UTXO for every token.
> 
> No, I was thinking of summing them up. Hence me stating that the IDs need to be massive random scalars. They should not be incrementing normal numbers. Then the ids become almost like aggregated keys.

But than you have to provide a proof for each UTXO that the sum of the ids (sum of `t`s, is actually a sum of `v` tokens from some set (distinct). Bu that's for NFTs only, If you have an asset with more than one FTs, then you can't distinguish the `v`s, you will have an UTXO with some `v` and e.g. two tokens `t1` and `t2`. How would you know that how many tokens `t1` you have ? You can say it's `v`, or `0` and `v` tokens `t2`. Or any linear combination.

SWvheerden:
> But than you have to provide a proof for each UTXO that the sum of the ids (sum of ts, is actually a sum of v tokens from some set (distinct). Bu that's for NFTs only, If you have an asset with more than one FTs, then you can't distinguish the vs, you will have an UTXO with some v and e.g. two tokens t1 and t2. How would you know that how many tokens t1 you have ? You can say it's v, or 0 and v tokens t2. Or any linear combination.

Mmm thats a good point about the values on different tokens for FT.  Why don't we do this then:

NFT:
We have no value, we just have a list of IDs, so no `vH ` on the commitment, that's if it's all unique, if it's not, then we still need the `vH `for value. 

FT:
We have no vH for the value of all tokens. We also don't have an id for the token. Rather we have `t1H`, `t2F `etc. Each token has its own value field with a unique generator.

Cifko:
> > But than you have to provide a proof for each UTXO that the sum of the ids (sum of ts, is actually a sum of v tokens from some set (distinct). Bu that's for NFTs only, If you have an asset with more than one FTs, then you can't distinguish the vs, you will have an UTXO with some v and e.g. two tokens t1 and t2. How would you know that how many tokens t1 you have ? You can say it's v, or 0 and v tokens t2. Or any linear combination.
> 
> Mmm thats a good point about the values on different tokens for FT. Why don't we do this then:
> 
> NFT: We have no value, we just have a list of IDs, so no `vH ` on the commitment, that's if it's all unique, if it's not, then we still need the `vH `for value.
> 
> FT: We have no vH for the value of all tokens. We also don't have an id for the token. Rather we have `t1H`, `t2F `etc. Each token has its own value field with a unique generator.

I'm afraid that if you sum everything into one UTXO, you will have some problems...e.g. you can't receive two transactions from two people at the same time. Because both will be spending your current UTXO. For person that maybe ok. But it would be impossible to build e.g. a store for NFTs/FTs.

SWvheerden:
The plan for the second layer is account-based, so if we want to keep account-based and confidential assets, we need to combine it all into a single utxo. 
If you want to send more than 1 single tx at a time, we can group them. But also because we won't really have long block times, it should be possible to send them out 1 at a time and have them finish quickly. And you can pipeline them, as you wait a few ms for the confirmation, you start negotiation on the next transaction.

hansieodendaal:
**_From a range proof perspective:_**
We always need `vH`, even if `v == 0`, otherwise we cannot have a range proof for the commitment - the range proof is for `v>=0` or `v>=v1` where `v1` is public and `>0`. In the same manner the owner of the commitment must know the remaining values for `r,t,s` otherwise the range proof cannot be constructed. `v` is like the amount of shares minted for a particular asset `s` with token id `t`, usually this will be `=1`.

hansieodendaal:
For commitment `C = vH + rG + tF + sE` the following must hold - `v`: scalar, `r`: point, `t`: point, `s`: point, `v>=0`. This means `r,t,s` will all be large prime numbers.

Strictly speaking the VNs should only know that:
- `v>=0` via the range proof or value made public
- `sE` is unique globally - this is a public value
- `tF` is unique globally - this is a public value

The minting transaction can disclosed and verified for the original amount of shares `v` for the specific token `t`, then the total supply for the NFT is also known.

PR #4064, opened 2022-04-29
fix(key-manager): remove floating point math from mnemonic code

sdbondi:
Description
---
- removes floating-point math from `mnemonic::from_bytes`
- masks "last" byte before conversion to u8 (clippy)

Motivation and Context
---
This fixes clippy cast warnings and avoids the overhead of working with floating-points 

How Has This Been Tested?
---
This module is tested well, these tests pass.


base_layer/key_manager/src/mnemonic.rs
```diff
@@ -185,20 +185,22 @@ pub fn from_bytes(bytes: &[u8], language: MnemonicLanguage) -> Result<Vec<String
 
     // Pad with zeros if length not divisible by 11
     let group_bit_count = 11;
-    let padded_size = ((bits.len() as f32 / group_bit_count as f32).ceil() * group_bit_count as f32) as usize;
+    let mut padded_size = bits.len() / group_bit_count;
```
Cifko, line 188: If you want, you can get rid of the `mut` by doing just `let padded_size = bits.len() + group_bit_count - 1 - (bits.len() - 1) % group_bit_count;`

base_layer/key_manager/src/mnemonic.rs
```diff
@@ -216,36 +218,37 @@ pub fn to_bytes(mnemonic_seq: &[String]) -> Result<Vec<u8>, MnemonicError> {
...
 /// We now have enough bits to get some output, we take 8 bits and produce output byte. We do this as long as we have at
 /// least 8 bits in the 'rest'.
 /// Sample of couple first steps:
 /// 1) the first output 'a' is last 8 bits from input 'A', we have leftover 3 bits from 'A'
 /// 2) We add 5 bits from 'B' to generate 'b', the leftover is 6 bits from 'B'
 /// 3) We add 2 bits from 'C to generate 'c', now we have 8 bits needed to generate 'd' and we have 1 bit leftover.
 pub fn to_bytes_with_language(mnemonic_seq: &[String], language: &MnemonicLanguage) -> Result<Vec<u8>, MnemonicError> {
-    let mut bytes: Vec<u8> = Vec::new();
-    let mut rest = 0;
+    const MASK: u64 = (1u64 << 8) - 1;
```
Cifko, line 232: you can use `u8::MAX.into()`, but it's matter of preference.

Cifko (approved):
Tested with `cargo test`

base_layer/key_manager/src/mnemonic.rs
SWvheerden: https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil
Does exactly what we want, but it's currently nightly only.

stringhandler (approved):

SWvheerden (approved):
5 minutes, 
Tested on new and old wallet

base_layer/key_manager/src/mnemonic.rs
sdbondi: Good point, like that better but oh well :)

PR #4065, opened 2022-04-29
refactor(comms): reduce length of long functions (clippy)

sdbondi:
Description
---
Reduces LOC for `handle_request` and `handle_connection_manager_event` methods
Allows long function for `NoiseSocket::poll_write_or_flush` 

Motivation and Context
---
These methods exceeded the linter's maximum of 100 LOC.
`poll_write_or_flush` is a poll function and so handling the continue loop case vs the return case makes putting it in a separate function a little unwieldy, and possibly will incur a small performance cost.

How Has This Been Tested?
---
Code compiles, Existing tests pass, running a base node

### DIFF
--- a/comms/core/src/noise/socket.rs
+++ b/comms/core/src/noise/socket.rs
@@ -364,6 +364,7 @@ where TSocket: AsyncRead + Unpin
 impl<TSocket> NoiseSocket<TSocket>
 where TSocket: AsyncWrite + Unpin
 {
+    #[allow(clippy::too_many_lines)]
     fn poll_write_or_flush(&mut self, context: &mut Context, buf: Option<&[u8]>) -> Poll<io::Result<Option<usize>>> {
         loop {
             trace!(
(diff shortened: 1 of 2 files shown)


Cifko (approved):
Tested by running tari_base_node

comms/core/src/noise/socket.rs
```diff
@@ -364,6 +364,7 @@ where TSocket: AsyncRead + Unpin
 impl<TSocket> NoiseSocket<TSocket>
 where TSocket: AsyncWrite + Unpin
 {
+    #[allow(clippy::too_many_lines)]
```
SWvheerden, line 367: NIT:
I would put the nice explanation you gave in the PR desc here as well.

SWvheerden (approved):
Review 10 minutes. 
Ensure that refactor did not change any logic and that this is purely a refactor. 
Does definitely improve the readability of the code.
Did not run base node as well as this was already done.

stringhandler (approved):

PR #4066, opened 2022-04-29
chore: use standard lint settings and fix lints

sdbondi:
Description
---
- updates lints.toml to match [tari-meta](https://github.com/tari-project/meta/blob/main/lints.toml)
- fixes a number of clippy minor lints

Motivation and Context
---
Tari repo uses standard linter settings EDIT: there are too many cast clippys to investigate, so that is left for another PR

Larger fixes are in #4063 #4064 #4065 which should be merged before this PR.

How Has This Been Tested?
---
`cargo lints clippy` passes


lints.toml
```diff
@@ -27,37 +25,31 @@ deny = [
...
+    'clippy::cast_precision-loss',
+    'clippy::cast_sign_loss'
 ]
-
-allow= [
-    # allow Default::default calls
-    'clippy::default_trait_access',
- # Generally when developers fix this, it can lead to leaky abstractions or worse, so
- # too many arguments is generally the lesser of two evils
- 'clippy::too_many_arguments'
```
stringhandler, line 62: I didn't realise this was removed from `meta`. It should be there as well

sdbondi: Good catch, will update here and in meta

sdbondi:
@SWvheerden @DenisKolodin Can I have a review on this one? It may be easier to check each of the commits individually.

applications/launchpad/backend/src/docker/workspace.rs
```diff
@@ -289,11 +289,11 @@ impl TariWorkspace {
         Ok(ids)
     }
 
-    /// Create and return a [`Stream`] of [`LogMessage`] instances for the `name`d container in the workspace.  
+    /// Create and return a [`Stream`] of [`LogMessage`] instances for the `name`d container in the workspace.
     pub fn logs(
         &self,
         container_name: &str,
-        docker: Docker,
+        docker: &Docker,
```
therustmonk, line 296: How about to put `Docker` into the `TariWorkspace`. We have to pass a reference to some methods calls, but it will me more convenient if we could keep the reference with the workspace.

1-st case, if you can put the `Docker` inside the `struct`:

```rust
pub struct TariWorkspace {
    name: String,
    config: LaunchpadConfig,
    containers: HashMap<String, ContainerState>,

    // Extra field
    docker: Docker,
}
```

It's the preferred way, but if a `Docker` instance aren't available at the moment of creation `TariWorkspace` you can add wrap it with an `Option`:

2-nd case:

```rust
pub struct TariWorkspace {
    name: String,
    config: LaunchpadConfig,
    containers: HashMap<String, ContainerState>,

    // Extra field
    docker: Option<Docker>,
}

impl TariWorkspace {

    // Methods to assign and access docker (not necessary if `Docker` is not optional)
    fn assign_docker(&mut self, docker: Docker) {
        self.docker = Some(docker);
    }

    fn docker_ref(&self) -> Result<&Docker, Error> {
        // Or `DockerWrapperError`
        self.docker.as_ref().ok_or_else(|| Error::msg("Docker not set to the TariWorkspace"))
    }
    
    fn docker(&self) -> Result<Docker, Error> {
        // Or `DockerWrapperError`
        self.docker.clone().ok_or_else(|| Error::msg("Docker not set to the TariWorkspace"))
    }
    
    pub fn logs(
        &self,
        container_name: &str,
    ) -> Option<impl Stream<Item = Result<LogMessage, DockerWrapperError>>> {

        // It returns `None` here if a docker reference is not available.
        let docker = self.docker_ref().ok()?;
        // Or use `Option` directly here (than `docker_ref` method is not necessary):
        let docker = self.docker.as_ref()?;

        let options = LogsOptions::<String> {
            follow: true,
            stdout: true,
            stderr: true,
            ..Default::default()
        };
        self.containers.get(container_name).map(move |container| {
            let id = container.id();
            docker
                .logs(id.as_str(), Some(options))
                .map(|log| log.map(LogMessage::from).map_err(DockerWrapperError::from))
        })
    }    
    
    pub async fn start_service(
        &mut self,
        image: ImageType,
        registry: Option<String>,
        tag: Option<String>,
    ) -> Result<String, DockerWrapperError> {

    	let docker = self.docker()?;
    	// Use docker here

    }
}
```

If we can do that, we can eliminate places where we have to pass `Docker` to methods directly.

therustmonk: Anyway, we can put that idea to a todo list as a variant of code layout. I keep doing review of other parts...

sdbondi: Thanks for the suggestion, probably good for another PR. This PR tries to avoid refactoring and focuses on fixing clippy errors/minor corrections in the least lines of code to get clippy to pass.

applications/tari_console_wallet/src/recovery.rs
```diff
@@ -78,6 +78,7 @@ pub fn get_seed_from_seed_words(seed_words: Vec<String>) -> Result<CipherSeed, E
 /// Recovers wallet funds by connecting to a given base node peer, downloading the transaction outputs stored in the
 /// blockchain, and attempting to rewind them. Any outputs that are successfully rewound are then imported into the
 /// wallet.
+#[allow(clippy::too_many_lines)]
```
SWvheerden, line 81: I cant comment on some of the lines below, but I am sure we can split this function into smaller ones. Like for example the loop in ln 120.
Add a todo we can split this into smaller functions in other prs

base_layer/core/src/base_node/rpc/service.rs
```diff
@@ -428,10 +428,8 @@ impl<B: BlockchainBackend + 'static> BaseNodeWalletService for BaseNodeWalletRpc
         let mut not_deleted_positions = vec![];
 
         for position in message.mmr_positions {
-            if position > u64::from(u32::MAX) {
-                return Err(RpcStatus::bad_request("position must fit into a u32"));
-            }
-            let position = position as u32;
+            let position =
+                u32::try_from(position).map_err(|_| RpcStatus::bad_request("All MMR positions must fit into a u32"))?;
```
SWvheerden, line 432: Haha one of the places that will break eventually.

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -182,6 +182,7 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
         Ok(connection)
     }
 
+    #[allow(clippy::too_many_lines)]
```
SWvheerden, line 185: add a todo here so we can split this in a future pr.

base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
```diff
@@ -301,6 +302,10 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
 
             kernel_hashes.push(kernel.hash());
 
+            // TODO: We mix u32 and u64 for the mmr length. This comes down to the use of a 32-bit croaring Bitmap.
+            //       Suggest should use u64 externally (u64 is in the header) and error on the database calls if they
+            //       are > u32::MAX. Remove the clippy exception once fixed.
```
SWvheerden, line 307: We should go over to u64 bitmap this is a ticking time bomb.

base_layer/core/src/mempool/mod.rs
```diff
@@ -78,9 +78,9 @@ use crate::transactions::transaction_components::Transaction;
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct StatsResponse {
-    pub total_txs: usize,
-    pub unconfirmed_txs: usize,
-    pub reorg_txs: usize,
+    pub total_txs: u64,
+    pub unconfirmed_txs: u64,
+    pub reorg_txs: u64,
```
SWvheerden, line 83: Why did we change this to u64?
These all come from vec len, which is usize which we now have to cast

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -482,9 +483,13 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin
...
 
-        let stats = self.mempool.stats().await?;
-        metrics::unconfirmed_pool_size().set(stats.unconfirmed_txs as i64);
-        metrics::reorg_pool_size().set(stats.reorg_txs as i64);
+        #[allow(clippy::cast_possible_truncation)]
+        #[allow(clippy::cast_possible_wrap)]
+        {
+            let stats = self.mempool.stats().await?;
+            metrics::unconfirmed_pool_size().set(stats.unconfirmed_txs as i64);
+            metrics::reorg_pool_size().set(stats.reorg_txs as i64);
```
SWvheerden, line 491: random question on metrics, we cant use usize here rather than i64?

SWvheerden (commented):
This looks good. 
The only thing I think we need to add is todo's and a good first issue in github so we can tackle some of these large functions. We have way too many of them.

sdbondi:
@SWvheerden definitely agree, the intention here was to mark them so that they can be fixed later. Searching for too_many_lines should do it so don't think we need a Todo above each of them.

base_layer/core/src/mempool/sync_protocol/mod.rs
sdbondi: Prometheus takes a signed integer, don't really have a choice.

applications/tari_base_node/src/bootstrap.rs
```diff
@@ -93,15 +93,15 @@ where B: BlockchainBackend + 'static
             .map(|s| SeedPeer::from_str(s))
             .map(|r| r.map(Peer::from).map(|p| p.node_id))
             .collect::<Result<Vec<_>, _>>()
-            .map_err(|e| ExitError::new(ExitCode::ConfigError, &e))?;
+            .map_err(|e| ExitError::new(ExitCode::ConfigError, e.to_string()))?;
```
therustmonk: `.to_string()` call is not necessary above. `ExitError` already does it:

```rust
impl ExitError {
    pub fn new<T: ToString>(exit_code: ExitCode, details: T) -> Self {
        Self {
            exit_code,
            details: Some(details.to_string()),
        }
    }
}

```

therustmonk (dismissed):
It's pretty huge, but I haven't serious admonitions here

applications/tari_base_node/src/bootstrap.rs
sdbondi: This is one of those cases where there doesn't seem to be a good answer. Clippy complains about needless_pass_by_value which is technically correct but makes the ergonomics of ExitError worse (esp. in the case of a double ref `str`). I first decided to simply pass in a String and let the caller convert as needed (often a String was passed in anyway so no conversion/allocation needed) but the static str case is worse e.g. `"my error msg".to_string()`. `T: Into<String>` would remove the `needless_pass_by_ref` issue but none of the error types implement this (though perhaps this is the best compromise). In the end it's such a minor issue and decided to move on to get the rest of the clippy fixes over the line.

base_layer/core/src/mempool/mod.rs
sdbondi: Agree, and this is a case where clippy may make the code worse. This saves on conversions and checks in the protobuf conversion code and ensures that this part works on 32-bit machines so I think its marginally safer (converting a usize to u64 on 32-bit machines is fine).

applications/tari_console_wallet/src/recovery.rs
stringhandler: No need for a todo, the clippy `allow` already shows that it needs to be split up

base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
stringhandler: The fix for the u64 bitmap is to add an array of bitmaps, I don't think it's as bad as you say, since when we get close to that number, we will need to add another element into the array

stringhandler (approved):

PR #4068, opened 2022-05-02
feat: contract storage in sqlite (proof of concept)

mrnaveira:
Description
---
* Created a new `diesel` migration in the existing `dan_layer` sqlite database.
    * The ` up.sql` script contains the instructions for creating the tables
    * The `sample_data.sql` script contains the instructions to insert sample data simulating a contract lifecycle, with 3 checkpoints, abandonment and quarantine.
* Comments: 
    * Decided to create a table `utxos` just for storing references to transactions in the blockchain. Not sure if it's necessary.
    * There are a few fields whose type and format is not clear: `initialization_arguments`, `runtime_specification`, `checkpoint_paramenters_change`, `sidechain_metadata_change`.
    * Expiration measurements for acceptance/changes done in relative block heights (thanks @hansieodendaal)
    * Regarding reogs, the DAN layer should wait for a set number of blocks before publishing the next step and be constantly monitoring for reorgs (again, thanks @hansieodendaal)

Motivation and Context
---
Proof of concept for contract data storage in SQLite

How Has This Been Tested?
---
Manually created a new sqlite database and then run `up.sql` to create the tables and `sample_data.sql` to populate them with a sample contract lifecycle.


hansieodendaal:
I really like all the comments in the sql creation file. 

To answer your question about timestamps, in a blockchain we usually measure "time" in block height, so time is relative to when something was mined.

I think the best way to deal with re-orgs for the initial registration, constitution and uptake transactions in a proof-of-work blockchain is to let things cool down a bit, for example wait for 100 blocks before the next step. Checkpoint transactions should only be published if the basis asset transactions (or their successor transactions) exist and are still unspent on the blockchain. The chain of all asset related transaction outputs must always be monitored to be in the blockchain with the correct spent/unspent status backwards from the latest one, and if a re-org is detected (some input or output does not exist anymore) all affected transactions should be re-submitted. If re-submitted transactions are rejected it means the "wallet's" view of the network is skewed and transaction inputs should be re-validated.

PR #4072, opened 2022-05-03
fix: support safe non-interactive mode

CjS77:
Non-interactive mode should NEVER prompt the user. This PR clears one
path where this happens: If a wallet does not exist and a password has
not been provided so that we can auto-create one. In this case, the
application should just exit.

Providing passwords on the command line is VERY bad practice, since
anyone with access to the machine can see the password in plaintext by
inspecting the running jobs.

The ability to read the password from the commaond-line, OR config file OR envar was
removed in a previous PR, and this is rectified here.

How Has This Been Tested?
---
Manually, with various CLI and envar combinations


sdbondi (approved):
utAck

stringhandler (approved):

PR #4073, opened 2022-05-03
feat: allow network to be set by TARI_NETWORK env var

sdbondi:
Description
---
Allows the base directory to be set using the `TARI_BASE_DIR` environment variable.
Allows the current network to be set by the `TARI_NETWORK` environment variable.
Allows the password to be passed in using an environment variable `TARI_WALLET_PASSWORD` in `tari_console_wallet`.

Motivation and Context
---
In some environments, it is desirable to use environment variables instead of passing cli arguments. 

How Has This Been Tested?
---
Setting env var `TARI_NETWORK=mainnet` and `dibbler` on the base node and wallet.
Setting env var `TARI_WALLET_PASSWORD = 'xxx'` on the console wallet

### DIFF
--- a/applications/tari_console_wallet/src/cli.rs
+++ b/applications/tari_console_wallet/src/cli.rs
@@ -39,7 +39,7 @@ pub(crate) struct Cli {
     /// Supply the password for the console wallet. It's very bad security practice to provide the password on the
     /// command line, since it's visible using `ps ax` from anywhere on the system, so always use the env var where
     /// possible.
-    #[clap(long)] // , env = "TARI_WALLET_PASSWORD")]
+    #[clap(long, env = "TARI_WALLET_PASSWORD", hide_env_values = true)]
     pub password: Option<String>,
     /// Change the password for the console wallet
     #[clap(long, alias = "update-password")]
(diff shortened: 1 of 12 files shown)


applications/tari_console_wallet/src/cli.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) struct Cli {
     /// Supply the password for the console wallet. It's very bad security practice to provide the password on the
     /// command line, since it's visible using `ps ax` from anywhere on the system, so always use the env var where
     /// possible.
-    #[clap(long)] // , env = "TARI_WALLET_PASSWORD")]
+    #[clap(long, env = "TARI_WALLET_PASSWORD", hide_env_values = true)]
```
therustmonk, line 42: How about to hide it completely with `#[clap(hide = true)]` ?

therustmonk (approved):

applications/tari_console_wallet/src/cli.rs
stringhandler: I think `hide_env_values` is correct here

stringhandler (approved):

PR #4078, opened 2022-05-04
wip: Decopling the contract between UI and Backend

alex1307:
Splitting the `Tauri` API contract from the underlying backend services.
Implement network_list service and expose it to `Tauri` command.

Motivation and Context
---

The current services is coupled with `Tauri` dependencies which makes the unit testing a challenge.
The current change splits the frontend contract and backend.
All `Tauri` commands supporting the UI are moved and all backend services are changed in order to not depend on \
`Tauri` UI bindings.

Implementing docker compose start, stop and ps commands.
Gather statistic and logs from running containers.

How Has This Been Tested?
---
Launchpad is started and all docker images are started and stopped. 
Logging, statistic work.
Started and stopped containers successfully.


applications/launchpad/backend/src/commands/network.rs
```diff
@@ -0,0 +1,44 @@
...
+    .into_iter()
+    .map(|enum_value| enum_value.to_string())
+    .collect()
+}
+
+#[test]
+fn network_list_test() {
+    let networks = ["dibbler".to_string(), "igor".to_string(), "mainnet".to_string()].to_vec();
+    assert_eq!(networks, enum_to_list(&TARI_NETWORKS));
+}
```
CjS77, line 44: Wrap in `#[cfg(test)]` mod please

applications/launchpad/backend/src/docker/models.rs
```diff
@@ -21,6 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 
+use core::fmt;
```
CjS77, line 24: use std::fmt for consistency

applications/launchpad/backend/Cargo.toml
```diff
@@ -21,6 +21,7 @@ tari_common = { path="../../../common"}
 bollard = "0.11.1"
 config = "0.13.0"
 env_logger = "0.9.0"
+lazy_static = "1.3.0"
```
CjS77, line 24: I notice a lot of people are ditching lazy_static for `thread_local` and `std::sync::once`. Possibly because lazy_static uses spin locks, or because it removes a dependency, I'm not sure.

applications/launchpad/backend/src/commands/api.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+
+///Sending messages to the Tauri
+pub fn send_tauri_message<P: Debug + Clone + Serialize>(
+    app: AppHandle<Wry>,
+    event: String,
+    payload: P,
+) -> Result<(), tauri::Error> {
+    debug!("sending: {:?} to {}", payload, event.clone());
```
CjS77, line 111: I don't think this clone is necessary. Check clippy

applications/launchpad/backend/src/commands/pull_images.rs
```diff
@@ -82,11 +73,11 @@ pub async fn pull_images(app: AppHandle<Wry>) -> Result<(), String> {
     Ok(())
 }
 
-async fn pull_image(image: ImageType, app: AppHandle<Wry>) -> Result<(), LauncherError> {
-    let state = app.state::<AppState>().clone();
-    let docker = state.docker.read().await;
+
+async fn pull_image<F>(image: ImageType, send_message: F) -> Result<(), LauncherError>
```
CjS77, line 77: Nice.

applications/launchpad/backend/src/docker/workspace.rs
```diff
@@ -466,44 +457,32 @@ impl TariWorkspace {
...
-            },
-        }
-        // Even if stopping failed (maybe it was already stopped), try and delete it
-        if delete {
-            match docker.remove_container(id.as_str(), None).await {
-                Ok(()) => {
-                    info!("Container {} deleted", id);
-                    container.set_deleted();
+        let about_to_delete = CONTAINERS.write().unwrap().remove(name);
+        if about_to_delete.is_some() {
```
CjS77, line 461: You don't log the message if we try and stop a non-existent container any more. To keep the same level of functionality you need to handle the `None` branch here and log the message from lines 472-475 on the left

CjS77: Also prefer this form
```
if let Some(container) = about_to_delete {
 // No need for: let container = about_to_delete.unwrap();
 ...
```

applications/launchpad/backend/src/docker/workspace.rs
```diff
@@ -466,44 +457,32 @@ impl TariWorkspace {
...
+            match docker.stop_container(id.as_str(), Some(options)).await {
+                Ok(_res) => {
+                    info!("Container {} stopped", id);
+                    if !delete {
+                        let updated = ContainerState::new(container.name().to_string().clone(),
+                            id.clone(),
+                            container.info().clone());
+                        
+                        CONTAINERS.write().unwrap().insert(name.to_string(), updated);
+                    }
```
CjS77, line 474: This is buggy. You never actually delete the container.

applications/launchpad/backend/src/docker/workspace.rs
```diff
@@ -466,44 +457,32 @@ impl TariWorkspace {
...
+            let options = StopContainerOptions { t: 0 };
+            match docker.stop_container(id.as_str(), Some(options)).await {
+                Ok(_res) => {
+                    info!("Container {} stopped", id);
+                    if !delete {
+                        let updated = ContainerState::new(container.name().to_string().clone(),
+                            id.clone(),
+                            container.info().clone());
+                        
+                        CONTAINERS.write().unwrap().insert(name.to_string(), updated);
```
CjS77, line 473: I'm not a fan of this. Just update the status.

applications/launchpad/backend/src/docker/workspace.rs
```diff
@@ -466,44 +457,32 @@ impl TariWorkspace {
...
-                warn!("Could not stop container {} due to {}", id, err.to_string());
-            },
-        }
-        // Even if stopping failed (maybe it was already stopped), try and delete it
-        if delete {
-            match docker.remove_container(id.as_str(), None).await {
-                Ok(()) => {
-                    info!("Container {} deleted", id);
-                    container.set_deleted();
+        let about_to_delete = CONTAINERS.write().unwrap().remove(name);
```
CjS77, line 460: There are problems with this rewrite. The original logic is better IMO. 
* Removing the state record is not the same as deleting the container
* Don't remove the container state object from the hashmap. Mark it as deleted. The idea here is to keep a record of deleted containers so that the front-end can differentiate between "destroyed" and "never existed".
* Update the fields if just stopping the container rather than removing and re-inserting. Doing this uses multiple locks and sort of defeats the point of having a "state" record in the first place.

applications/launchpad/backend/src/commands/service.rs
```diff
@@ -207,12 +177,110 @@ async fn create_default_workspace_impl(app: AppHandle<Wry>, settings: ServiceSet
...
+        .unwrap();
+
+    let found_state = match CONTAINERS.try_read().unwrap().get(service_name.as_str()){
+        Some(reference) => Some((*reference).to_owned()),
+        None => None,
+    };
+    
+    if found_state.is_some() {
+        
+        let container_state = found_state.unwrap();
```
CjS77, line 215:
```suggestion
    let Some(container_state) = match CONTAINERS.try_read().unwrap().get(service_name.as_str()){
        Some(reference) => Some((*reference).to_owned()),
        None => None,
    };
    

```

CjS77: I think you could even drop the cloned container_state this way

applications/launchpad/backend/src/commands/service.rs
```diff
@@ -207,12 +177,110 @@ async fn create_default_workspace_impl(app: AppHandle<Wry>, settings: ServiceSet
...
+        let stats_events_name = stats_event_name(container_state.id());
+        let log_events_name = log_event_name(container_state.name());
```
CjS77, line 217: Should check the container `status` and take appropriate actions

CjS77 (commented):
Some nice improvements here, but there are some bugs that need to be ironed out in the container management code.

PR #4081, opened 2022-05-05
fix(base-node): assign correct base dir to tor identity

sdbondi:
Description
---
Fixes regression: prepend the base directory to the tor identity file location.

Motivation and Context
---
The regression caused the tor identity, and therefore public address, to be shared amongst multiple base nodes.
This causes `DialedPublicKeyMismatch` errors that have been observed in logs.

How Has This Been Tested?
---
Manually checked that two base nodes have different tor identities and the file location is relative to the base dir.


SWvheerden (approved):
5 Minutes
LGTM
Tested and fixes the bug

stringhandler (approved):

PR #4083, opened 2022-05-06
fix(wallet): ensure block hash exists

brianp:
Description
---
Adds a quick data validation check before updating the database. The base node may return a response without a block_hash and the resulting db update creates an invalidatable scenario for the wallet.

Motivation and Context
---
An aurora wallet is failing to validate and causing problems

How Has This Been Tested?
---
It has not yet been tested.

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
@@ -322,7 +322,7 @@ where
                 .map_err(TransactionServiceError::ProtobufConversionError)?;
             let sig = response.signature;
             if let Some(unconfirmed_tx) = batch_signatures.get(&sig) {
-                if response.location == TxLocation::Mined {
+                if response.location == TxLocation::Mined && response.block_hash.is_some() {
                     mined.push((
                         (*unconfirmed_tx).clone(),
                         response.block_height,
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -323,12 +323,21 @@ where
...
-                    ));
+                    if response.block_hash.is_some() {
+                        mined.push((
+                            (*unconfirmed_tx).clone(),
+                            response.block_height,
+                            response.block_hash.unwrap(),
+                            response.confirmations,
+                        ));
+                    } else {
+                        debug!(
```
stringhandler:
```suggestion
                      warn!(
```

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -323,12 +323,21 @@ where
             let sig = response.signature;
             if let Some(unconfirmed_tx) = batch_signatures.get(&sig) {
                 if response.location == TxLocation::Mined {
-                    mined.push((
-                        (*unconfirmed_tx).clone(),
-                        response.block_height,
-                        response.block_hash.unwrap(),
-                        response.confirmations,
-                    ));
+                    if response.block_hash.is_some() {
```
brianp: I had to look into this a bit. I was concerned is_some() protects us from passing an empty option, but I was curious what would happen if our block hash was made of just an empty Vec<u8>. Was there a chance we could have created an empty vector and as a result diesel saved it in the db as null? 
It turns out this isn't a concern as diesel will use a specific null binary for an empty vector. Our example DB contains no data in the field so this must not be our current case, and likely not a concern here.

stringhandler (approved):

PR #4085, opened 2022-05-06
fix: prevent seed peer address from being overwritten unless newer

sdbondi:
Description
---
- Prevents seed peer addresses from being overwritten unless:
  -  The seed peer signs a peer update after the node starts up,
  - directly connected to the seed peer.
- Adds PeerValidator unit tests.

Motivation and Context
---
This is only a problem if the base node address has changed recently enough that it is not known to the rest of the network. The user has put the correct new address. The _unsigned_ base node peer is added to the peer DB.

Soon after, the DHT syncs peers, since the previous address was signed, the base node peer is overwritten with the previous address.

How Has This Been Tested?
---
Unit test for PeerValidator tests this behaviour

### DIFF
--- a/integration_tests/features/support/steps.js
+++ b/integration_tests/features/support/steps.js
@@ -408,10 +408,11 @@ Then(/(.*) should have (\d+) peers/, async function (nodeName, peerCount) {
 });
 
 Then(/(.*) has at least (\d+) peers/, async function (nodeName, peerCount) {
-  await sleep(500);
   const client = this.getClient(nodeName);
-  const peers = await client.getPeers();
-  expect(peers.length).to.be.greaterThanOrEqual(peerCount);
+  await waitForPredicate(async () => {
+    const peers = await client.getPeers();
+    return peers.length >= peerCount;
+  }, 10 * 1000);
 });
 
 When("I print the world", function () {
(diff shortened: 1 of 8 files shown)


SWvheerden (commented):
Looks good to me
The code does some refactoring, the refactoring makes it easier to read to code.  
Tested and works. 

It failing this integration test however ` When a new node joins the network, it should receive all peers`

integration_tests/features/support/steps.js
```diff
@@ -407,15 +407,12 @@ Then(/(.*) should have (\d+) peers/, async function (nodeName, peerCount) {
...
-  /(.*) should have at least (\d+) peers/,
-  async function (nodeName, peerCount) {
-    await sleep(500);
-    const client = this.getClient(nodeName);
-    const peers = await client.getPeers();
-    expect(peers.length).to.be.greaterThanOrEqual(peerCount);
-  }
-);
+Then(/(.*) has at least (\d+) peers/, async function (nodeName, peerCount) {
+  await sleep(500);
```
SWvheerden: This commit only changed some of the wording. This means this is most likely a flaky test. Should we not remove this sleep and put a longer wait and test till the condition matches in rather?
Something like this:
`await waitFor(
        async () => {`

sdbondi: The actual error on CI was a port conflict, pointing to a problem in the way we assign open ports in cucumber, not related to the test but I agree we should allow more than 500ms for the peers to sync so I've added get_peer polling as  suggested.

stringhandler (approved):

PR #4086, opened 2022-05-09
feat(daily-tests): use environment variable to set custom seed words

sdbondi:
Description
---
Allow custom seed words to be set for wallet recovery daily test by setting the `WALLET_RECOVERY_SEED_WORDS` environment variable.

Motivation and Context
---
Whenever the seed words in daily tests need to change, a code change is needed. The seed words should be configurable. The current hardcoded seed words are the default.

How Has This Been Tested?
---


SWvheerden (dismissed):
Looks good to me
Untested

stringhandler (approved):

PR #4087, opened 2022-05-10
fix(core): set reorg metrics on header sync rewind

sdbondi:
Description
---
Sets `tari_base_node::blockchain::reorgs` and `tari_base_node::blockchain::tip_height` metrics on header sync rewind.

Motivation and Context
---
Header sync is designed to handle syncing from nodes that are on a forked chain, and so can rewind and reorg the blockchain. This PR sets the correct metrics for a header sync rewind.

How Has This Been Tested?
---
Not explicitly tested in this PR (reorgs tested in general and in cucumber tests). This case is fairly rare and difficult to test manually (unit tests should be implemented for header sync! but this is out of scope). The correct metrics should be seen on grafana over the next month.


base_layer/core/src/base_node/state_machine_service/states/header_sync.rs
```diff
@@ -112,8 +113,13 @@ impl HeaderSyncState {
         });
 
         let local_nci = shared.local_node_interface.clone();
-        synchronizer.on_rewind(move |blocks| {
-            local_nci.publish_block_event(BlockEvent::BlockSyncRewind(blocks));
+        synchronizer.on_rewind(move |removed| {
+            if let Some(fork_height) = removed.last().map(|b| b.height()) {
```
SWvheerden: Are we sure `removed` is ordered by height ASC?
I think it's ordered by height DESC. 
But I don't think anywhere we guarantee or mention this. This might be an oversight.

SWvheerden (commented):
This looks good. I have a question about the order. 
I don't think we need to add a unit test to test this. But a cucumber test should be easy to test this.

base_layer/core/src/base_node/state_machine_service/states/header_sync.rs
sdbondi: It's ordered DESC (tip to fork height), `removed.last() -1` is the new tip height, this is documented in the `rewind_to_height` function in blockchain database. Agree it's not super clear.

sdbondi: We don't currently test metrics in cucumber. Updated the description.

SWvheerden (approved):
LGTM

stringhandler (approved):

PR #4088, opened 2022-05-11
feat(core)!: define OutputFlags for side-chain contracts

sdbondi:
Description
---
Defines OutputFlags for side-chain contracts.

Motivation and Context
---
Define flags required for side-chain contracts. Deprecated flags are kept and can be removed in implementation PRs
for the various side-chain components.

How Has This Been Tested?
---
Unit and cucumber tests updated as necessary

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/output_flags.rs
+++ b/base_layer/core/src/transactions/transaction_components/output_flags.rs
@@ -47,6 +47,14 @@ bitflags! {
         const CONTRACT_DEREGISTER   = 0x1000;
         /// Output is an abandoned contract checkpoint.
         const CONTRACT_ABANDONED    = 0x2000;
+
+        // TODO: Remove these deprecated flags
+        const NON_FUNGIBLE         = 0b0000_1000;
+        const ASSET_REGISTRATION   = 0b0000_0010 | Self::NON_FUNGIBLE.bits;
+        const MINT_NON_FUNGIBLE    = 0b0000_0100 | Self::NON_FUNGIBLE.bits;
+        const BURN_NON_FUNGIBLE    = 0b1000_0000 | Self::NON_FUNGIBLE.bits;
+        const SIDECHAIN_CHECKPOINT = 0b0001_0000 | Self::NON_FUNGIBLE.bits;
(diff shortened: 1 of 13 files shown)


hansieodendaal (dismissed):
LGTM - barring the CI failures

mrnaveira:
Same here, LGTM, I believe they're exactly the features we need for the contract lifecycle

base_layer/core/src/transactions/transaction_components/output_flags.rs
```diff
@@ -31,15 +31,20 @@ use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSi
...
+        /// Output is a coinbase output, must not be spent until maturity.
+        const COINBASE_OUTPUT       = 0x0001;
+        /// Output defines a side-chain contract.
+        const CONTRACT_DEFINITION   = 0x0100;
+        /// Output defines the constitution for a side-chain contract.
+        const CONTRACT_CONSTITUTION = 0x0200;
+        /// Output signals validator node acceptance to run a contract.
+        const CONTRACT_ACCEPT       = 0x0400;
+        /// Output is a contract checkpoint.
+        const CONTRACT_CHECKPOINT   = 0x0800;
```
hansieodendaal, line 44: We may need an additional flag for the initial checkpoint where all the contract acceptance UTXOs have to be spent into it - a slight nuance on the rules to be enforced.

mrnaveira: Isn't that going to be just the first checkpoint? Or does it need to be a custom output for that?

sdbondi: I think you're right and we might need an abandoned flag so the asset owner can spend an abandoned checkpoint into that state to activate the emergency keys

hansieodendaal: Agreed

sdbondi: I left off the INITIAL_CHECKPOINT flag for now because even if we have it, we'd have to check that it is an INITIAL checkpoint (we can't trust the flag), in doing so we have done the work to detect it.

stringhandler (dismissed):

stringhandler (approved):

PR #4090, opened 2022-05-11
fix: possible overflow in difficulty calculation (fixes #3923)

### LINKED ISSUE #3923: Downcasting in difficulty calculation can lead to overflow
The [difficulty calculation](https://github.com/tari-project/tari/blob/c0750b7496d62efdde43d9d053c3f49c51299c83/base_layer/core/src/proof_of_work/difficulty.rs#L119) has an overflow issue:

```rust 
pub(crate) fn big_endian_difficulty(hash: &[u8]) -> Difficulty {
        let scalar = U256::from_big_endian(hash); // Big endian so the hash has leading zeroes
        let result = U256::MAX / scalar;
        result.low_u64().into()
    }
```

Here, `scalar` is the target for the mining algorithm. The smaller the target, the more difficult it is to find a hash with the requisite number of leading zeroes.

The definition of difficulty is `TARGET_MAX / target`. Once target falls below 2^192, the difficulty will overflow to zero, since we downcast to 64 bits by truncating the highest portion of the actual difficulty.

Now, this is possibly only a theoretical problem since we need a ridiculous hashrate before this become an eventuality. We will certainly have native 256-bit integers long before this happens. 

However, to protect against weird things happening, it might be better to prevent the overflow anyway by returning `u64::max` if scalar <= 2^192.

jorgeantonio21:
Description
 - First attempt to tackle issue #3923

Motivation and Context
 - The main goal is to avoid possible overflow of difficulty, which even though extremely unlikely, it is a possible scenario. 

How Has This Been Tested?
 - So far, changes have not been tested. I am more than happy to get recommendations of tests. I haven't add them at the moment because it is not clear how to build an hash whose big endian is a scalar less than 2^192. I am therefore, very interested to understand this concept and add tests accordingly.

### DIFF
--- a/base_layer/core/src/proof_of_work/difficulty.rs
+++ b/base_layer/core/src/proof_of_work/difficulty.rs
@@ -129,10 +129,41 @@ pub mod util {
         let result = U256::MAX / scalar;
         result.low_u64().into()
     }
+
+    #[cfg(test)]
+    mod test {
+        use super::*;
+
+        #[test]
+        fn high_target() {
+            let target: &[u8] = &[
+                0xff, 0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+                0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+            ];
+            let expected = Difficulty::from(1);
+            assert_eq!(big_endian_difficulty(target), expected);
+        }
+
+        #[test]
+        fn max_difficulty() {
+            let target = U256::MAX / U256::from(u64::MAX);
+            let mut bytes = [0u8; 32];
+            target.to_big_endian(&mut bytes);
+            assert_eq!(big_endian_difficulty(&bytes), Difficulty::from(u64::MAX));
+        }
+
+        #[test]
+        fn stop_overflow() {
+            let target: u64 = 64;
+            let expected = u64::MAX;
+            assert_eq!(big_endian_difficulty(&target.to_be_bytes()), Difficulty::from(expected));
+        }
+    }
 }
+
 #[cfg(test)]
 mod test {
-    use crate::proof_of_work::difficulty::{Difficulty, util::big_endian_difficulty};
+    use crate::proof_of_work::difficulty::Difficulty;
 
     #[test]
     fn add_difficulty() {
@@ -149,25 +180,4 @@ mod test {
         let d = Difficulty::from(1_000_000);
         assert_eq!("1,000,000", format!("{}", d));
     }
-
-    #[test]
-    fn high_target() {
-        let target: &[u8] = &[0xff, 0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ];
-        let expected = Difficulty::from(1);
-        assert_eq!(big_endian_difficulty(target), expected);
-    } 
-
-    #[test]
-    fn max_difficulty() {
-        let target = u64::MAX;
-        let expected = u64::MAX;
-        assert_eq!(big_endian_difficulty(&target.to_be_bytes()), Difficulty::from(expected));
-    } 
-
-    #[test]
-    fn stop_overflow() {
-        let target: u64 = 64;
-        let expected = u64::MAX;
-        assert_eq!(big_endian_difficulty(&target.to_be_bytes()), Difficulty::from(expected));
-    } 
 }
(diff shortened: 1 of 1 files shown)


base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -118,6 +118,9 @@ pub mod util {
     /// This will provide the difficulty of the hash assuming the hash is big_endian
     pub(crate) fn big_endian_difficulty(hash: &[u8]) -> Difficulty {
         let scalar = U256::from_big_endian(hash); // Big endian so the hash has leading zeroes
+        if scalar <= U256::from(2).pow(U256::from(192)) {
+            return u64::MAX.into();
+        }
```
CjS77: It's simpler to do the U256 division in #124 first, and then check for overflow:

```
let scalar = U256::from_big_endian(hash);
let result = U256::MAX / scalar;
let result = result.min(u64::MAX.into())
```

jorgeantonio21: Thank you for the suggestion, I changed the code accordingly.

CjS77:
You can also write simple unit tests that are something like

```rust
#[test]
fn high_target() {
 let target: &[u8] = &[0xff, 0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ];
let expected = 1; # replace by the actual difficulty
assert!(big_endian_difficulty(target), expected);
} 

#[test]
fn max_difficulty() {
 let target = u64::MAX;
let expected = 1; # replace by the actual difficulty
assert!(big_endian_difficulty(target.to_be_bytes()), expected);
} 

#[test]
fn stop_overflow() {
 let target = 64;
let expected = u64::MAX; # replace by the actual difficulty
assert!(big_endian_difficulty(target.to_be_bytes()), expected);
} 
```

jorgeantonio21:
> ```rust
> #[test]
> fn high_target() {
>  let target: &[u8] = &[0xff, 0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ];
> let expected = 1; # replace by the actual difficulty
> assert!(big_endian_difficulty(target), expected);
> } 
> 
> #[test]
> fn max_difficulty() {
>  let target = u64::MAX;
> let expected = 1; # replace by the actual difficulty
> assert!(big_endian_difficulty(target.to_be_bytes()), expected);
> } 
> 
> #[test]
> fn stop_overflow() {
>  let target = 64;
> let expected = u64::MAX; # replace by the actual difficulty
> assert!(big_endian_difficulty(target.to_be_bytes()), expected);
> } 
> ```

I added these tests, with minor mods to get the correct values. Thank you for the suggestion

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -131,7 +132,7 @@ pub mod util {
 }
 #[cfg(test)]
```
CjS77, line 164: you'll need to move your tests into a new module inside the utils mod so that they get compiled under the same feature flag.

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -148,4 +149,25 @@ mod test {
...
+    #[test]
+    fn high_target() {
+        let target: &[u8] = &[0xff, 0xff, 0xff, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 ];
+        let expected = Difficulty::from(1);
+        assert_eq!(big_endian_difficulty(target), expected);
+    } 
+
+    #[test]
+    fn max_difficulty() {
+        let target = u64::MAX;
```
CjS77: This is actually the same test case as the one below. What you want here is a test that captures the path where the result in L#121 is equal to u64::MAX, such that the min function doesn't actually apply.

This should suffice:


            let target = U256::MAX / U256::from(u64::MAX);
            let mut bytes = [0u8; 32];
            target.to_big_endian(&mut bytes);
            assert_eq!(big_endian_difficulty(&bytes), Difficulty::from(u64::MAX));

CjS77 (commented):
Looking good. If you could finlise the test cases to improve the coverage slightly, and not keep feature flags consistent, we should be good.

(And a note that the `little_endian` version has the same bug, but this is for another PR)

CjS77:
I think you also need to run `cargo fmt -p tari_core` to pass the clippy tests here.

base_layer/core/src/proof_of_work/difficulty.rs
jorgeantonio21: Good point, I understand the logic and refactored the code accordingly. In this case, L#122 computes the min between the two values of `u64::MAX`, as represented by type `U256`, am I correct? 

If I am not missing something, we now have test coverage for each possible situation regarding the method `big_endian_difficulty`.

base_layer/core/src/proof_of_work/difficulty.rs
jorgeantonio21: I moved the tests to a test mod inside the utils mod, accordingly.

jorgeantonio21:
> I think you also need to run `cargo fmt -p tari_core` to pass the clippy tests here.

I did so, but for some reason I don't understand, the clippy tests continue to fail.

CjS77 (approved):
LGTM - assuming tests pass

PR #4092, opened 2022-05-11
fix(wallet header): remove function argument

brianp:
Description
---
As per https://github.com/tari-project/tari/commit/fcbdb0a424f66b9c77ca5334aec72290b07c7593 we removed the `network` argument from `comms_config_create` but forgot to remove it from the header file causing mobile builds to fail.

Motivation and Context
---
Corrective actions.

How Has This Been Tested?
---
It's been tested in the iOS build that went from not working to working.


sdbondi (approved):
Good catch!

PR #4093, opened 2022-05-12
test: coverage for chain_storage

therustmonk:
Description
---
Add coverage tests for `chain_storage`.

Motivation and Context
---
To increase the total test coverage.

How Has This Been Tested?
---
Unit tests

### DIFF
--- a/base_layer/core/src/chain_storage/async_db.rs
+++ b/base_layer/core/src/chain_storage/async_db.rs
@@ -399,3 +399,34 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self.db.write(transaction).await
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_helpers::blockchain::{create_new_blockchain, TempDatabase};
+
+    impl AsyncBlockchainDb<TempDatabase> {
+        pub fn sample() -> Self {
+            Self {
+                db: create_new_blockchain(),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn coverage_async_blockchain_db() {
+        let obj = AsyncBlockchainDb::sample();
+        obj.clone().into_inner();
+        obj.fetch_horizon_data().await.unwrap();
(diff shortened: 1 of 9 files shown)


base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -399,3 +399,34 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
...
+        obj.fetch_chain_header(0).await.unwrap();
+        obj.fetch_last_header().await.unwrap();
+        obj.clear_all_pending_headers().await.unwrap();
+        obj.fetch_last_chain_header().await.unwrap();
+        obj.cleanup_orphans().await.unwrap();
+        obj.cleanup_all_orphans().await.unwrap();
+        obj.prune_to_height(0).await.unwrap();
+        obj.get_stats().await.unwrap();
+        obj.fetch_total_size_stats().await.unwrap();
+        let _trans = obj.write_transaction();
```
SWvheerden, line 430: Did we go through these to ensure the cucumber tests cover them?
These tests only test that these functions do not return an error. Not that they produce a correct result. 
I think the test name should reflect this.

base_layer/core/src/chain_storage/horizon_data.rs
```diff
@@ -45,3 +45,17 @@ impl HorizonData {
         &self.utxo_sum
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn coverage_horizon_data() {
```
SWvheerden, line 54: See comment above. 
Here should be easily able to confirm that we do indeed return 0.h and 0.g

SWvheerden (commented):
This looks good. 
Most of these tests look fine, 
But some of them are artificially boosting the coverage. We are only testing that the functions return a result and don't send back an error. We don't actually test the function if returns the correct result.  We need to decide, case by case if we want leave the function as "untested" or blanked cover it using these test.

stringhandler (approved):

PR #4100, opened 2022-05-13
fix: better message for failed software update check

sdbondi:
Description
---
Changed the previous failure message for software updates to a warning and made the message clearer.

Motivation and Context
---
The current error message had a typo and looks like a more serious failure, however this happens currently because the
DNS has not been set up for dibbler.

This PR makes it clear that this is just a failure to check for software updates.

How Has This Been Tested?
---
Simple message change


SWvheerden (approved):
LGTM
Only a log text change.

stringhandler (approved):

stringhandler (approved):

PR #4101, opened 2022-05-13
fix: single allocation for block header consensus encoding

sdbondi:
Description
---
- implements `ConsensusEncodingSized` for `BlockHeader` to allow for single allocation when encoding to bytes
- uses `to_consensus_bytes` in `tari_mining_helper_ffi::inject_nonce`
- Use consensus encoding impl for EpochTime

Motivation and Context
---
Previously an unallocated Vec was passed into `consensus_encode` which leads to a number of allocations.
It is best practice, if the size is known, to allocate the vec once. This is implemented generically in `to_consensus_bytes`.

How Has This Been Tested?
---
Existing tests pass.


stringhandler (approved):
utACK

PR #4102, opened 2022-05-13
feat: hidden type to mask sensitive data in logs

brianp:
Description
---
Adding a wrapper type to mask sensitive data in debug, and log output.

Motivation and Context
---
To reduce attack vector surface area by ensuring log data doesn't contain private information.

How Has This Been Tested?
---
With a single test. It has not yet been applied in practice.

### DIFF
--- a/common/src/logging.rs
+++ b/common/src/logging.rs
@@ -21,7 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 
-use std::{fs, fs::File, io::Write, path::Path};
+use std::{fmt, fs, fs::File, io::Write, ops::DerefMut, path::Path};
 
 use crate::ConfigError;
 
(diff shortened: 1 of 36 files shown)


common/src/logging.rs
```diff
@@ -100,8 +100,46 @@ macro_rules! log_if_error_fmt {
...
+        &self.inner
+    }
+}
+
+impl<T> DerefMut for Hidden<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
```
sdbondi: Looks great! I would only add implementation `From<T> for Hidden<T>` . So you can go `42.into()` but up to you.

I think we should add this to the tari_utilities crate. This can also be included here and refactored out once tari_utilities has been released.

therustmonk: +1 for the `From<T>` implementation. Also you can add `into_inner(self) -> T` method to take the data out.

common/src/logging.rs
```diff
@@ -100,8 +100,46 @@ macro_rules! log_if_error_fmt {
...
+impl<T> fmt::Display for Hidden<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "[FILTERED]")
+    }
+}
+
+impl<T> std::ops::Deref for Hidden<T> {
```
therustmonk: How about to use [`derive_more`](https://crates.io/crates/derive_more) crate?
That'll make the implementation more compact: 

```rust
use derive_more::{From, Deref, DerefMut};

#[derive(Deref, DerefMut, From)]
pub struct Hidden(...);
```

stringhandler: I'm not keen on another crate, we end up having to audit all of these crates

SWvheerden (approved):

therustmonk:
Looks good.

The important notice: [`Hidden`][1] struct has `inner` field that will be reflected in the serialized data. Will it break backward compatibility?

[1]: https://github.com/brianp/tari_utilities/commit/f3ebe080bece63ba51e8be5ca2ceec8aaad8d586#diff-aac67a6c0179542071145a27eb9f5377976f423b090ac2ce436829c863340e12

brianp:
> will be reflected in the serialized data

Yes. I think it will break backward compatibility. I think the serialized structure may need something more custom to just output the inner value instead of a nested key/value. Maybe we need to treat it like [serde primitives](https://serde.rs/impl-serialize.html#serializing-a-primitive), and serialize only the inner value to ensure hidden isn't returned as a key, value pair.

brianp:
Update: `#[serde(transparent)]` on the Hidden type gets us there.

brianp:
Closing this PR. It works, and it shows the Hidden struct is useful but for legacy code it's high risk low reward to make these changes. The struct will be available in tari_utilities and we should use it moving forward but not worth the investment to retro-fit it unless someone is actually making changes in these areas.

PR #4103, opened 2022-05-13
fix(core): don't allow coinbase transactions in mempool

sdbondi:
Description
---
Rejects coinbase transactions from the mempool

Motivation and Context
---
Coinbase transactions should never be allowed in the mempool

How Has This Been Tested?
---
New unit test


SWvheerden (dismissed):
LGTM
Tested by syncing a new node, sending a transaction to aurora and back.

therustmonk (dismissed):

base_layer/core/src/validation/error.rs
```diff
@@ -114,6 +114,8 @@ pub enum ValidationError {
     CovenantError(#[from] CovenantError),
     #[error("Invalid or unsupported blockchain version {version}")]
     InvalidBlockchainVersion { version: u16 },
+    #[error("Output in transaction is marked as a coinbase")]
+    MempoolTransactionContainsCoinbase,
```
CjS77:
```suggestion
    #[error("Standard transaction contains coinbase output")]
    ErroneousCoinbaseOutput
```

CjS77: This is not necessarily just a mempool thing

CjS77:
IIRC this is the same check that nodes will make before deciding whether to rebroadcast a transaction.
If so, then that's the slight more important instance where this should get thrown out, and why I proposed renaming the error slightly.

Otherwise, I'm OK with the concept.

stringhandler (approved):

PR #4109, opened 2022-05-17
feat(wallet_ffi): new ffi method to create output features

mrnaveira:
Description
---
* Created a new FFI method to create an output features object, by passing the byte values as parameters.
* I decided to not expose the DAN layer features, as they are most likely going to change and also would require exposing more struct definitions.
* Updated the C header with the definition of the new method

Motivation and Context
---
Currently the mobile apps cannot properly import faucet UTXOs, as they are passing null values for output features and covenants to the `wallet_import_external_utxo_as_non_rewindable` method, which is not allowed. So the mobile apps needs new FFI methods to create both output features and covenants. This pull request deals with output features only.

How Has This Been Tested?
---
Created two new unit tests for the new FFI method

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -74,6 +74,7 @@ extern crate lazy_static;
 use core::ptr;
 use std::{
     boxed::Box,
+    convert::TryFrom,
     ffi::{CStr, CString},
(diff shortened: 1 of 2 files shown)


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #4112, opened 2022-05-18
 feat(launchpad): network_list and health_check commands

alex1307:
Description

    Implement api.rs module which provides the contract between Tauri UI and underlying backend services.
    Implement network_list tauri command and apply the same approach to image_list command.
    Move the commands to api module which should handle the contract between UI and backend.
    Implement health_check endpoint providing the actual container's statuses read from Docker.
    Separate the Container states from the workspace configuration.

    Motivation and Context
    The Container State struct should be used as a binding between image (ImageType) and docker containers(name, id).
    The actual docker status should be read from docker with the provided attributes(container id for example).
    Health check

    How Has This Been Tested?
    Container module provides create, update and delete test cases.
    То test health check I run docker and uncommented the test_health_check in module health_check.rs
    I run the app and check that base node and tor images are being started correctly.


alex1307:
The PR should be made in Altalogy Tari GitHub Repo

PR #4113, opened 2022-05-18
refactor(core): change ConsensusEncoding trait not to return written bytes

sdbondi:
Description
---
- Changes ConsensusEncoding trait not to return written bytes

Motivation and Context
---
Returning the number of bytes written from `consensus_encode` is unused, adds cruft to the API, and is error-prone.

This also uncovered some potential bugs where `write` was used instead of `write_all` (must use warnings) 
though this would only be an issue if writing to an `impl Write` that cannot always write all bytes in one go (e.g. network socket)

How Has This Been Tested?
---
Existing tests updated and pass


SWvheerden (approved):
LGTM
Only does the refactor stated and cleans up the tests.

stringhandler (approved):

PR #4114, opened 2022-05-18
feat: add FeePerGramStats to ffi library

Cifko:
Description
---
Add ffi function for getting the fee per gram stats.

How Has This Been Tested?
---
`npm test -- -p none --name "As a client I want to get fee per gram stats"`

### DIFF
--- a/base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
+++ b/base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
@@ -564,8 +564,7 @@ impl UnconfirmedPool {
             let mut total_fees = MicroTari::zero();
             let mut min_fee_per_gram = MicroTari::from(u64::MAX);
             let mut max_fee_per_gram = MicroTari::zero();
-            let mut last_count = 0;
-            for (i, key) in self.tx_by_priority.values().rev().enumerate().skip(offset) {
+            for key in self.tx_by_priority.values().rev().skip(offset) {
                 let tx = self.tx_by_key.get(key).ok_or(UnconfirmedPoolError::StorageOutofSync)?;
                 let weight = tx.weight;
 
@@ -574,23 +573,23 @@ impl UnconfirmedPool {
                 }
 
                 let total_tx_fee = tx.transaction.body.get_total_fee();
-                last_count = i + 1;
+                offset += 1;
                 let fee_per_gram = total_tx_fee / weight;
                 min_fee_per_gram = min_fee_per_gram.min(fee_per_gram);
                 max_fee_per_gram = max_fee_per_gram.max(fee_per_gram);
                 total_fees += total_tx_fee;
                 total_weight += weight;
             }
-            if last_count > 0 {
-                let stat = FeePerGramStat {
-                    order: start as u64,
-                    min_fee_per_gram,
-                    avg_fee_per_gram: total_fees / total_weight,
-                    max_fee_per_gram,
-                };
-                stats.push(stat);
+            if total_weight == 0 {
+                break;
             }
-            offset = last_count;
+            let stat = FeePerGramStat {
+                order: start as u64,
+                min_fee_per_gram,
+                avg_fee_per_gram: total_fees / total_weight,
+                max_fee_per_gram,
+            };
+            stats.push(stat);
         }
 
         Ok(stats)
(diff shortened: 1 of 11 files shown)


base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
```diff
@@ -581,15 +581,16 @@ impl UnconfirmedPool {
...
+                break;
             }
+            let stat = FeePerGramStat {
+                order: start as u64,
+                min_fee_per_gram,
+                avg_fee_per_gram: total_fees / total_weight,
+                max_fee_per_gram,
+            };
+            stats.push(stat);
             offset = last_count;
```
SWvheerden: Should this not increase?
This will only be able to count 2 weights worth of blocks deep.

SWvheerden (commented):
Looks good, I think there is a bug here thou. 
Perhaps we should test that case on unit or integration tests

base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
stringhandler: Does it not increase at line 577?

sdbondi: Yeah sounds right (`offset += last_count`), looks like I was a bit rushed with this one ;)

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -6430,6 +6432,275 @@ pub unsafe extern "C" fn log_debug_message(msg: *const c_char, error_out: *mut c
...
+) -> *mut TariFeePerGramStat {
+    let mut error = 0;
+    ptr::swap(error_out, &mut error as *mut c_int);
+    if fee_per_gram_stats.is_null() {
+        error = LibWalletError::from(InterfaceError::NullError("fee_per_gram_stats".to_string())).code;
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+    let len = fee_per_gram_stats_get_length(fee_per_gram_stats, error_out) as c_int - 1;
+    if len < 0 || position > len as c_uint {
```
sdbondi: nit: little cleaner
```suggestion
    let len = fee_per_gram_stats_get_length(fee_per_gram_stats, error_out);
    if len == 0 || position > len - 1 {
```

integration_tests/helpers/ffi/feePerGramStats.js
```diff
@@ -0,0 +1,32 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+const FeePerGramStat = require("./FeePerGramStat");
```
sdbondi:
```suggestion
const FeePerGramStat = require("./feePerGramStat");
```

sdbondi (commented):
Looks good - Just one `require` broken in cucumber code. With that fix, ran the new cucumber test and it passes.
Please also change the offset line to `offset += last_count`

base_layer/wallet_ffi/src/lib.rs
sdbondi: You'll also need to check error_out == 0

Cifko: Uff, I did copy paste most of these function from the other part of the code in this file. And now I see how many times we don't check :-) This file could you a big refactor...

base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
Cifko: I did refactor the code. I got rid of the last_count and enumerate as well. Please take a look.

base_layer/wallet_ffi/wallet.h
```diff
@@ -868,6 +870,8 @@ bool wallet_start_recovery(struct TariWallet *wallet, struct TariPublicKey *base
 /// None
 bool wallet_set_one_sided_payment_message(struct TariWallet *wallet, const char *message, int *error_out);
 
+struct TariFeePerGramStatsResponse *wallet_get_fee_estimates_per_block(struct TariWallet *wallet, unsigned int count, int *error_out);
```
sdbondi: You'll need to add the remaining fee_per_gram functions to the header

Cifko: Ahhh, there is so much stuff that we could just generate ;-) but it's not worth of doing so...

stringhandler (approved):

PR #4116, opened 2022-05-19
test: coverage for rolling average time struct

jorgeantonio21:
Description

- Add test coverage at `base_layer/core/src/common/...`

Motivation and Context

- Some additional coverage needs to be added to the code base . This PR adds a few of additional tests to cover logic regarding RollingAverageTime struct, etc.

How Has This Been Tested?

- We added tests for each of the present methods in the modified files.

### DIFF
--- a/base_layer/core/src/common/byte_counter.rs
+++ b/base_layer/core/src/common/byte_counter.rs
@@ -48,3 +48,24 @@ impl io::Write for ByteCounter {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::io::Write;
+
+    #[test]
+    fn write_test() {
+        let mut byte_counter = ByteCounter::new();
+        let buf = [0u8, 1u8, 2u8, 3u8];
+        let new_count = byte_counter.write(&buf).unwrap();
+        assert_eq!(byte_counter.get(), new_count);
+    }
+
+    #[test]
+    fn flush_test() {
+        let mut byte_counter = ByteCounter::new();
+        let flushed = byte_counter.flush().unwrap();
+        assert_eq!(flushed, ());
+    }
+}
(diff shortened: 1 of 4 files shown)


base_layer/core/src/common/byte_counter.rs
```diff
@@ -48,3 +48,24 @@ impl io::Write for ByteCounter {
...
+        let mut byte_counter = ByteCounter::new();
+        let buf = [0u8, 1u8, 2u8, 3u8];
+        let new_count = byte_counter.write(&buf).unwrap();
+        assert_eq!(byte_counter.get(), new_count);
+    }
+
+    #[test]
+    fn flush_test() {
+        let mut byte_counter = ByteCounter::new();
+        let flushed = byte_counter.flush().unwrap();
```
stringhandler: I think add some bytes to it as well

```suggestion
        let flushed = byte_counter.flush().unwrap();
```

stringhandler (dismissed):
Nice, just need the CI to pass

base_layer/core/src/common/byte_counter.rs
SWvheerden: Do something like this:
```
        let buf = [0u8, 1u8, 2u8, 3u8];
        let new_count = byte_counter.write(&buf).unwrap();
```

SWvheerden (commented):
This looks good, just add those bytes. 
Please run `cargo fmt` as well and push that up.

base_layer/core/src/common/byte_counter.rs
jorgeantonio21: Hi guys, thank you for the comments. I did add some bytes in the `flush_test` function, could you guys look it up and see if it is okay.

jorgeantonio21:
I did run the `cargo fmt` and applied the comments above

SWvheerden (approved):
LGTM

stringhandler (approved):

PR #4121, opened 2022-05-20
feat(core): adds constitution UTXO features

sdbondi:
Description
---
- adds `ContractConstitution` and related structs 
- implements consensus de/encoding for `ConsensusConstitution`
- implements `ConsensusEncoding` ,`Deref` and `DerefMut` for `MaxSizeVec`
- adds check for extraneous bytes to `check_consensus_encoding_correctness` 

Motivation and Context
---
Add initial structs for contract constitution UTXO. Adding this to OutputFeatures is left for a subsequent PR.

How Has This Been Tested?
---
Unit tests for consensus encoding


base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -67,6 +67,11 @@ pub struct OutputFeatures {
     #[serde(default)]
     pub recovery_byte: u8,
     pub metadata: Vec<u8>,
+    // TODO: Add these fields
+    // pub contract_id: Option<FixedHash>,
+    // pub constitution: Option<ContractConstitution>,
```
SWvheerden, line 72: Do we want to add flags for these as well?
Or are we going with the presence of them is enough?

base_layer/core/src/transactions/transaction_components/side_chain/contract_constitution.rs
```diff
@@ -0,0 +1,339 @@
...
+        })
+    }
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Deserialize, Serialize, Eq)]
+pub struct CommitteeChangeRules {
+    /// The maximum number of committee members that can be added.
+    pub max_new_committee_members: u32,
+    /// The maximum number of committee members that can be evicted.
+    pub max_evicted_committee_members: u32,
```
SWvheerden, line 177: Perhaps more options are better, but I don't see any reason why we would limit these two fields

SWvheerden (dismissed):
LGTM, 
I only left two general questions but it good is it stands

base_layer/core/src/transactions/transaction_components/output_features.rs
sdbondi: OutputFlags already exist for the constitution. Whenever a contract_id is present we likely want one of those flags to be set

base_layer/core/src/transactions/transaction_components/side_chain/contract_constitution.rs
sdbondi: Yeah wasn't sure what settings we want to add to this. Should only have settings that are useful since each setting translates into more validation complexity and impact performance. Suggestions welcome. Could also leave this out for now.

stringhandler (approved):

PR #4122, opened 2022-05-20
refactor: replace type alias with FixedHash newtype

sdbondi:
Description
---
- replaces type alias with FixedHash newtype
- implement consensus encoding for FixedHash
- removes duplicate fixed hash code from validator node

Motivation and Context
---
This newtype (originally added in the DAN core) is more generally useful and removes some of the cruft when using a 256-bit hash (e.g. `buf.copy_from_slice(...)` etc).

This type should be used for all new code that requires a hash, and eventually replace Vec<u8> (variable length, heap, not guaranteed to be well-formed) for hash types in BlockHeader etc.

How Has This Been Tested?
---
Existing tests pass


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #4131, opened 2022-05-24
fix(dht): optimisation, no decrypt if public key dest doesn't match

sdbondi:
Description
---
Don't attempt decryption if the public key destination does not match the node identity

Motivation and Context
---
Decryption is not necessary if the destination public key is specified and does not match.

How Has This Been Tested?
---
Manually, Memorynet and existing tests that check this case already pass.


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #4133, opened 2022-05-24
feat(wallet): new command to publish a contract definition transaction

mrnaveira:
Description
---
* Created a new transaction output features for the Contract Definition
    * For simplicity, I decided to leave out fields that could be more coupled to the template format.
    * The `contract_id` is calculated as a hash of the contents of the contract definition, in the line of what is described in the RFCs. I used `ConsensusHashWriter` to make it totally consistent.
    * I created a new Output Features version to signal the consensus breaking changes.
    * Had to update the genesis block struct to include the new features
    * Had to update the gRPC types to include the new features
* Created a new command in the tari console wallet (`publish-contract-definition`):
    1. Reads the contract definition from a JSON file. I defined auxiliary structs (`base_layer/wallet/src/assets/contract_definition.rs`) to decouple the file format from the output features.
    2. Created a new transaction with the `CONTRACT_DEFNITION` flag and the new Contract Definition output features.
    3. Publishes the corresponding UTXO into the network.
* Right now the base layer does not perform any check in the values in the contract definition. In the future we want to implement custom consensus rules for contract definition (like avoiding duplicates, spending requirements, etc.) but I purposely left that out of this PR for simplicity. 

Motivation and Context
---
As a user, I want to be able to publish a new contract definition into the network through the console wallet.

How Has This Been Tested?
---
* Created a new unit test for the consensus encoding of the new Contract Definition output features
* Created a new integration test for the wallet CLI that reads the contract definition from a JSON file and successfully publishes the transaction


base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,225 @@
...
+// Maximum number of functions allowed in a contract specification
+const MAX_FUNCTIONS: usize = u16::MAX as usize;
+
+// Fixed lenght of all string fields in the contract definition
+const STR_LEN: usize = 32;
+
+#[derive(Debug, Clone, Hash, PartialEq, Deserialize, Serialize, Eq)]
+pub struct ContractDefinitionFeatures {
+    pub contract_id: FixedHash,
+    pub contract_name: Vec<u8>,
```
sdbondi: I think rather use the fixed type since it's so small and saves a heap allocation
```suggestion
    pub contract_name: [u8; STR_LEN],
```

and perhaps a conversion function

```rust
// try return &str but I think rust will complain
pub fn contract_name_str(&self) -> Cow<'_, str> {
    String::from_utf8_lossy(&self.contract_name) /* .as_ref() -> &str */
}
```

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,225 @@
...
+    array::copy_into_fixed_array_lossy,
+    types::{FixedHash, PublicKey},
+};
+
+use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized, MaxSizeVec};
+
+// Maximum number of functions allowed in a contract specification
+const MAX_FUNCTIONS: usize = u16::MAX as usize;
+
+// Fixed lenght of all string fields in the contract definition
```
sdbondi:
```suggestion
// Fixed length of all string fields in the contract definition
```

integration_tests/features/WalletCli.feature
```diff
@@ -150,3 +151,14 @@ Feature: Wallet CLI
         And I create committee definition for asset on wallet WALLET via command line
         And mining node MINE mines 1 blocks
         Then WALLET is connected to BASE
+
+     Scenario: As a user I want to publish a contract definition via command line
```
brianp: Should we tag this scenario with something for like `@layertwo` or `@l2ud`? Something so we can scope or run specifically the use case defined specs for layer 2 work. Not needed for every new spec but for the ones where we're obviously completing a defined user scenario.

mrnaveira: Good idea, maybe `@dan_layer`? Just to keep consistent naming across the project

sdbondi (dismissed):
Looks good - approved with few minor comments

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,265 @@
...
+
+use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized, ConsensusHashWriter, MaxSizeVec};
+
+// Maximum number of functions allowed in a contract specification
+const MAX_FUNCTIONS: usize = u16::MAX as usize;
+
+// Fixed lenght of all string fields in the contract definition
+const STR_LEN: usize = 32;
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq, Hash)]
```
SWvheerden:
```suggestion
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
```
This is not the same Hash trait we use, this is the rust one `std::hash`

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,265 @@
...
+        Ok(Self {
+            contract_id,
+            contract_name,
+            contract_issuer,
+            contract_spec,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq, Hash)]
```
SWvheerden:
```suggestion
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
```

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,265 @@
...
+        let public_functions = MaxSizeVec::<PublicFunction, MAX_FUNCTIONS>::consensus_decode(reader)?.into_vec();
+
+        Ok(Self {
+            runtime,
+            public_functions,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq, Hash)]
```
SWvheerden:
```suggestion
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
```

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
```diff
@@ -0,0 +1,265 @@
...
+impl ConsensusDecoding for PublicFunction {
+    fn consensus_decode<R: Read>(reader: &mut R) -> Result<Self, Error> {
+        let name: Vec<u8> = <[u8; STR_LEN] as ConsensusDecoding>::consensus_decode(reader)?.to_vec();
+        let function = FunctionRef::consensus_decode(reader)?;
+
+        Ok(Self { name, function })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq, Hash)]
```
SWvheerden:
```suggestion
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
```

base_layer/wallet/src/assets/contract_definition.rs
```diff
@@ -0,0 +1,87 @@
...
+// USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use serde::{Deserialize, Serialize};
+use tari_common_types::types::{FixedHash, PublicKey};
+use tari_utilities::hex::Hex;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ContractDefinition {
+    pub contract_name: String,
+    pub contract_issuer: String,
```
SWvheerden: Should this not be a pubkey?

SWvheerden (commented):
This looks good. 
I just have an issue with the `derive(Hash)`
And a question about the one type

sdbondi (dismissed):
+1

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
mrnaveira: I did it because `output_features.rs` (which is the "parent" struct) derives the standard that Hash trait, so it makes it required to do it here as well

base_layer/wallet/src/assets/contract_definition.rs
mrnaveira: I agree. Even if that struct is just to decouple the features from the JSON file parsing, the PublicKey type is compatible with serde, so it can be used directly here. I pushed already the changes.

base_layer/core/src/transactions/transaction_components/contract_definition_features.rs
SWvheerden: mmm, wonder why it has it...
It looks like it could be removed, but I think that's for another PR

stringhandler (dismissed):
Nice. I want to note that the public functions may not look like that in the final form, but this is fine for a start

stringhandler (approved):

PR #4135, opened 2022-05-25
ci: docker image build for x86-64 & arm64 from x86-64

leet4tari:
Description
Build docker images for base_node, console_wallet, mm_proxy and sha3_miner on x86-64 for both x86-64 & arm64.

Cross-compiled in docker, not emulated. bash build script to follow. Builds on PR #4076, but needs randomx-rs at 1.1.11 and croaring and croaring-sys at 0.5.2 for cross-compiling to work.

Motivation and Context
Combined x86-64 and arm64 in single image from one CI service. 

How Has This Been Tested?
Build in a VM and on local hardware.

Images have not yet been tested, looking for feedback.

PR #4136, opened 2022-05-25
chore: update croaring and randomx

stringhandler:
Updates croaring and randomx


leet4tari:
LGTM - Basically I made similar changes in my testing for PR #4135

PR #4142, opened 2022-05-25
feat: add encrypted value in the transaction output

therustmonk:
Description
---
The PR adds extra encrypted field to `UnblindedOutput` structs.

What was done:
- Added `EncryptedValue` struct
- Implemented `ConsensusEncode`, `ConsensusDecode` and `ByteArray` for it
- Added the `encrypted_value` field of that type to the `UnblindedOutput` struct
- Added an extra parameter to metadata signature calculation methods
- Added an extra field to the `GRPC` protocol (`UnblindedOutput` type)
- Added an extra column and migrations for the `outputs` table
- The types `OutputSql` and `NewOutputSql` updated to store and restore the `EncryptedValue`

Important notes:
1. The value of `encrypted_value` passed as a parameter to the metadata signature evaluation but has not taken into account yet, because it will break the verification process of the `TransactionOutput` and we can complete that when the same `encrypted_value` field will be added to the `TransactionOutput` struct in the further PRs.
2. Real encryption process is not implemented yet. Instead of that I've added the method `todo_encrypt_from` that I used to mark the places where we have to put an encryption service and call it.

Motivation and Context
---
Add an encrypted amount to the `TransactionOutput`.

How Has This Been Tested?
---
CI

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/transaction_output.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_output.rs
@@ -63,6 +63,7 @@ use crate::{
         transaction_components::{
             full_rewind_result::FullRewindResult,
             rewind_result::RewindResult,
+            EncryptedValue,
             OutputFeatures,
             OutputFlags,
             TransactionError,
@@ -172,6 +173,8 @@ impl TransactionOutput {
             self.metadata_signature.public_nonce(),
             &self.commitment,
             &self.covenant,
+            // TODO: Use `self.encrypted_value` instead
+            &EncryptedValue::default(),
         );
         if !self.metadata_signature.verify_challenge(
             &(&self.commitment + &self.sender_offset_public_key),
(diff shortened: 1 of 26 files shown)


hansieodendaal:
I see `cargo clippy --all-targets -- -D warnings` picks up some errors, probably from an earlier merge, not too many, could you please fix it?

hansieodendaal (dismissed):
LGTM!

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -261,6 +266,7 @@ impl TransactionOutput {
                 .chain(sender_offset_public_key)
                 .chain(commitment)
                 .chain(covenant)
+                //.chain(encrypted_value)
```
stringhandler, line 269: Should this not be in here?

therustmonk: It is, but to add that I have to add `encrypted_value` to the TransactionOutput as well. I'm going to do that in the next PR to keep this one smaller.

stringhandler (approved):

PR #4147, opened 2022-05-30
fix: move peer dbs into sub folders

stringhandler:
Description
---
Add sub folders for each application's peer DB

Motivation and Context
---
If you run the base node and validator node in the same folder, you will get a file lock error. I added the same folder to the console wallet to keep it the same

How Has This Been Tested?
---
cargo test


mrnaveira (approved):
I downloaded the branch and manually ran both the base node and the validator node. The validator node now starts correctly without the peer_db issue.

PR #4148, opened 2022-05-30
feat: add an encrypted value to the TransactionOutput

therustmonk:
Description
---
The PR adds the `encrypted_value` field (of `EncryptedValue` type) to the `TransactionOutput` struct.
The same field also added to the `TransactionInput` for a case when input was produced from an output.

The changes also affected the genesis block, dibbler's faucet data, hashes and integration tests.

The value is not actually encrypted and now it's stored in the following format: u64le + 16 empty bytes.

Motivation and Context
---
Part of the BP+ features.

How Has This Been Tested?
---
CI tests (updated)

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
@@ -28,35 +28,27 @@ use std::io::{self, Read, Write};
 use serde::{Deserialize, Serialize};
 use tari_utilities::{ByteArray, ByteArrayError};
 
-use crate::consensus::{ConsensusDecoding, ConsensusEncoding};
+use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized};
 
 const SIZE: usize = 24;
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
-pub struct EncryptedValue {
-    /// value: u64 + tag: [u8; 16]
-    pub data: [u8; SIZE],
-}
+/// value: u64 + tag: [u8; 16]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+pub struct EncryptedValue(#[serde(with = "tari_utilities::serde::hex")] pub [u8; SIZE]);
 
 impl Default for EncryptedValue {
     fn default() -> Self {
-        Self { data: [0; SIZE] }
+        Self([0; SIZE])
     }
 }
 
 impl ByteArray for EncryptedValue {
     fn from_bytes(bytes: &[u8]) -> Result<Self, ByteArrayError> {
-        if bytes.len() == SIZE {
-            let mut this = Self::default();
-            this.data.copy_from_slice(bytes);
-            Ok(this)
-        } else {
-            Err(ByteArrayError::IncorrectLength)
-        }
+        ByteArray::from_bytes(bytes).map(Self)
     }
 
     fn as_bytes(&self) -> &[u8] {
-        &self.data
+        self.0.as_bytes()
     }
 }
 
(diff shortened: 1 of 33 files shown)


base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -67,20 +65,20 @@ impl EncryptedValue {
         let mut data: [u8; SIZE] = [0; SIZE];
         let value = amount.into().to_le_bytes();
         data[0..8].copy_from_slice(&value);
-        Self { data }
+        Self(data)
     }
 }
 
 impl ConsensusEncoding for EncryptedValue {
```
hansieodendaal, line 72: I think  a test would be in order for the encoding/decoding

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -67,20 +65,20 @@ impl EncryptedValue {
...
     }
 }
 
 impl ConsensusDecoding for EncryptedValue {
     fn consensus_decode<R: Read>(reader: &mut R) -> Result<Self, io::Error> {
         let data = <[u8; 24]>::consensus_decode(reader)?;
-        Ok(Self { data })
+        Ok(Self(data))
     }
 }
```
hansieodendaal: I think a test would be in order for the encoding/decoding

hansieodendaal (dismissed):
LGTM

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -32,31 +32,29 @@ use crate::consensus::{ConsensusDecoding, ConsensusEncoding};
...
 const SIZE: usize = 24;
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
-pub struct EncryptedValue {
-    /// value: u64 + tag: [u8; 16]
-    pub data: [u8; SIZE],
-}
+/// value: u64 + tag: [u8; 16]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+pub struct EncryptedValue(#[serde(with = "hex::serde")] pub [u8; SIZE]);
```
sdbondi: Any reason we use hex to serialize this value?
This essentially doubles the storage required to 48 bytes.
You could implement the `Hex` trait from tari_utilities if you want to convert this value to and from hex

sdbondi (commented):
Looks good - one question

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
therustmonk: I've added hex support to read values from the `dibbler_faucet.json` file. Ok, I'll try to make it more coherently with other types we use in the `TransactionOutput`.

sdbondi: :+1: I think you may need serialize_with, something like this https://stackoverflow.com/a/67148611/1077579
Essentially, using hex only when the serialisation is "human readable"

base_layer/core/src/transactions/transaction_components/hex_patch.rs
```diff
@@ -0,0 +1,105 @@
+use std::{fmt, marker::PhantomData};
```
hansieodendaal: The licence text is missing here

base_layer/core/src/transactions/transaction_components/mod.rs
```diff
@@ -57,6 +57,7 @@ mod committee_definition_features;
 mod encrypted_value;
 mod error;
 mod full_rewind_result;
+mod hex_patch;
```
hansieodendaal: Where is this used? Maybe it should move to `tari_utilities`?

base_layer/core/src/transactions/transaction_components/hex_patch.rs
therustmonk: Moved and fixed here: https://github.com/tari-project/tari_utilities/pull/43

base_layer/core/src/transactions/transaction_components/mod.rs
therustmonk: Yeah, I did that: https://github.com/tari-project/tari_utilities/pull/43

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -67,20 +59,54 @@ impl EncryptedValue {
...
         data[0..8].copy_from_slice(&value);
-        Self { data }
+        Self(data)
+    }
+
+    /// TODO: Replace this method with a real call of decryption service
+    /// that will produce a decrypted value from self.
+    pub fn todo_decrypt(&self) -> u64 {
+        let mut buffer = [0u8; 8];
+        (&mut buffer[0..8]).copy_from_slice(&self.0.as_slice()[0..8]);
```
sdbondi, line 69:
```suggestion
        buffer[0..8].copy_from_slice(&self.0[0..8]);
```

sdbondi (approved):
LGTM - not tested

PR #4149, opened 2022-05-31
feat(launchpad): stream wallet events to front-end

alex1307:
Description

Provide a new Tari event stream for:
	- transaction received (source: wallet)
	- transaction sent (source: wallet)
	- transaction cancelled (source: wallet)
	- transaction mined but unconfirmed (source: wallet)
	- transaction mined and confirmed (source: wallet)

Expose a streaming interface for all data we log using.
Create grpc TransactionEvent model and implement grpc streaming endpoint.
Forward logging data from notifier(app. logger) to the grpc streaming service.
Map the logging data to TransactionEvent model (Vec<String> -> TransactionEvent)

Motivation and Context
The launchpad users want to see the actual mining progress and all mining relalted events on the UI.
Hence we should forward the data/log streams from the underlying apps to the front-end.

How Has This Been Tested?

!!! TO_DO: The change IS NOT TESTED.
The change should be tested manually along with base node, tor, and launchpad.


stringhandler:
Retargeting to `testnet-dibbler`

PR #4145, opened 2022-05-27
feat(core): add contract acceptance utxo features

mrnaveira:
Description
---
- Adds `ContractAcceptance` struct to the side-chain output features
- Implements consensus encoding/decoding for `ContractAcceptance`
- Updates related gRPC types and conversions

Motivation and Context
---
Create structs for contract acceptance in the side-chain features.

How Has This Been Tested?
---
- New unit test for consensus encoding/decoding of the `ContractAcceptance`
- Existing unit/integration test pass

PR #4151, opened 2022-06-02
feat: contract acceptance publication

mrnaveira:
Description
---
- New gRPC method in the validator node for publishing a contract acceptance,
    - The only parameter for now is the `contract_id`
    - The validator node internally uses its own public key for building the acceptance.
    - The signature of the acceptance only has a mock value for now. When we have contract constitutions in place, we probably would want it to be the signature of a hash of a contract constitution identification.
    - Calls the new gRPC method in the wallet to publish the transaction
 - New gRPC method in the wallet to submit a contract acceptance. It builds and sends to the network a contract acceptance with the corresponding flags and features 

Motivation and Context
---
Validator nodes need to be able to publish contract acceptances, for contracts in which they are included as committee members. 

This PR continues [previous work](https://github.com/tari-project/tari/pull/4145) and provides a gRPC method in the validator node (and a corresponding one in the wallet) to publish contract acceptances. 

The functionality can be used in future work for implementing both auto-acceptance and manually acceptance of contract constitutions.

How Has This Been Tested?
---
New integration test that publishes and mines a contract acceptance through the new validator node gRPC endpoint


integration_tests/features/support/world.js
```diff
@@ -661,6 +661,13 @@ BeforeAll({ timeout: 2400000 }, async function () {
   await miningNode.init(1, 1, 1, 1, true, 1);
   await miningNode.compile();
 
+  /*
+  const vn = new ValidatorNodeProcess("compile", false, {}, null, null, "127.0.0.1:9999", "127.0.0.1:9998");
+  console.log("Compiling validator node...");
+  await vn.init();
+  await vn.compile();
+  */
+
```
SWvheerden, line 669: test code?

SWvheerden (commented):
This looks good. 
The only thing missing here, ut this could be a follow up PR, is validation on these transactions. 
All the stuff the base node needs to check

integration_tests/features/support/world.js
mrnaveira: Nice catch, I just updated the code to uncomment vn validation.

mrnaveira:
> This looks good. The only thing missing here, ut this could be a follow up PR, is validation on these transactions. All the stuff the base node needs to check

Correct. The work on layer 2 so far was intended to have a way to store all the features into the blockchain. We left the consensus rules/validation for future PRs, but it's something we will need to address soon.

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -656,6 +658,42 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+
+        let contract_id = FixedHash::try_from(message.contract_id)
+            .map_err(|err| Status::invalid_argument(format!("Invalid contract_id:{:?}", err)))?;
+
+        let validator_node_public_key =
+            PublicKey::from_bytes(message.validator_node_public_key.as_slice()).map_err(|e| {
+                Status::invalid_argument(format!("Validator node public key was not a valid pub key:{}", e))
+            })?;
+
+        let signature = Signature::try_from(message.signature.unwrap())
```
sdbondi: Not providing a signature will cause a panic

```suggestion
let signature = message.signature.ok_or_else(|| Status::bad_request("signature not provided"))?;
        let signature = Signature::try_from(signature)
```

base_layer/core/src/transactions/transaction_components/output_features.rs
```diff
@@ -294,6 +294,25 @@ impl OutputFeatures {
...
 
+    pub fn for_contract_acceptance(
+        contract_id: FixedHash,
+        validator_node_public_key: PublicKey,
+        signature: Signature,
+    ) -> OutputFeatures {
+        Self {
+            flags: OutputFlags::CONTRACT_ACCEPT,
+            sidechain_features: Some(
+                SideChainFeaturesBuilder::new(contract_id)
```
sdbondi: Suggest using the builder helper function on `SideChainFeatures` rather than importing the build struct directly
```suggestion
                SideChainFeatures::builder(contract_id)
```

sdbondi (commented):
Looks good, few minor comments

stringhandler (approved):

PR #4154, opened 2022-06-02
feat(wallet_ffi): add get utxos method

stringhandler:
It adds a method

### DIFF
--- a/applications/tari_console_wallet/src/automation/commands.rs
+++ b/applications/tari_console_wallet/src/automation/commands.rs
@@ -681,7 +681,7 @@ pub async fn command_runner(
                 println!("Emoji ID  : {}", emoji_id);
             },
             ExportUtxos => {
-                let utxos = output_service.get_unspent_outputs().await?;
+                let utxos = output_service.get_unspent_outputs(0, 10000, false).await?;
                 let count = utxos.len();
                 let sum: MicroTari = utxos.iter().map(|utxo| utxo.value).sum();
                 if parsed.args.is_empty() {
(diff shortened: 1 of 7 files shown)


applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -711,7 +711,7 @@ pub async fn command_runner(
                 println!("Total value of UTXOs: {}", sum);
             },
             CountUtxos => {
-                let utxos = output_service.get_unspent_outputs().await?;
+                let utxos = output_service.get_unspent_outputs(0,1000,false).await?;
```
SWvheerden, line 714: Is there a reason this is left filled in, and not taken from user commands?

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -221,7 +221,7 @@ impl wallet_server::Wallet for WalletGrpcServer {
     ) -> Result<Response<GetUnspentAmountsResponse>, Status> {
         let mut output_service = self.get_output_manager_service();
         let unspent_amounts;
-        match output_service.get_unspent_outputs().await {
+        match output_service.get_unspent_outputs(0, 1000, false).await {
```
SWvheerden, line 224: same here

SWvheerden (commented):
This looks good. 
The only question I have is about the fixed values for the commands and GRPC requests?

applications/tari_console_wallet/src/automation/commands.rs
stringhandler: I update this after #4612 is merged

PR #4160, opened 2022-06-03
fix(hotstuff): fix bug where decide state was listening for wrong message

stringhandler:
Description
---
Fixes the decide state listening for the wrong message type

Motivation and Context
---
I noticed that views were timing out and that the decide state would log  "Already received this message". This was happening because the leader would read the message that the replica was meant to, and then the replica would miss the message and the state would time out

How Has This Been Tested?
---
Manually


mrnaveira:
LGTM. As far as I can understand, I think the changes are consistent with what it's specified in the HotStuff original paper (the need for a `Decide` message type and replicas waiting for `Commit` message)

mrnaveira (approved):

sdbondi (approved):

PR #4162, opened 2022-06-03
feat(wallet): add help for wallet cli commands

stringhandler:
Description
---
Converts all wallet commands to use clap instead, providing help when called with `--help`

* Note: Also commented out the Script mode. I think we can remove this now as it's not used
* Note: Deleted all of the previous arg passing logic.
* Note: Deleted previous `register_asset` and other asset-y commands

Motivation and Context
---
This is IMHO a much cleaner and extendible way of calling commands and seeing what the required fields are. It also allows defaults to be added (e.g. default, overrideable messages)

How Has This Been Tested?
---
manually


sdbondi (dismissed):
Much better - just some minor 🖇️

sdbondi (approved):
Tested by manually running a command and works

PR #4168, opened 2022-06-07
fix(block-sync): check coinbase maturity

sdbondi:
Description
---
- Checks for correct coinbase maturity in block sync validator
- Adds unit tests for `check_coinbase_maturity` and `check_coinbase_reward` helpers

Motivation and Context
---
Coinbase maturity should be checked on block sync.

How Has This Been Tested?
---
Additional unit tests pass. Manually  by syncing on dibbler testnet.

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -349,7 +349,6 @@ impl AggregateBody {
     /// This function does NOT check that inputs come from the UTXO set
     /// The reward is the total amount of Tari rewarded for this block (block reward + total fees), this should be 0
     /// for a transaction
-
     pub fn validate_internal_consistency(
         &self,
         tx_offset: &BlindingFactor,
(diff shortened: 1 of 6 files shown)


CjS77:
It's a bit odd that we're asking the miners to add the maturity themselves.
I thought our code was such that if a coinbase output is being spent, we chack if it was mined > maturity blocks ago.

sdbondi:
There are currently no _special_ rules for spending a coinbase. We check that the maturity is correctly set at creation time as per consensus, however this check was missing from the block sync validation. 

We could add a separate spend rule for coinbases, however, this would require tracking`*` and fetching the mined height for each coinbase`**` vs statelessly comparing the maturity to the current height. 

It feels simpler to enforce the correct maturity for the miner at coinbase creation but open to changing the approach esp if there are downsides/oddness.

`*` UTXO mined_height is currently tracked but is informational and does not form part of the protocol/consensus rules
`**` Alternatively we could use the mmr pos of the coinbase input <= output_mmr_size in the header@{height-coinbase_maturity} to check maturity without tracking mined height.

CjS77 (approved):
I've been thinking on this. I agree with you that the pros outweigh the cons here. I initially balked because we have this policy of not adding redundant info to blocks (if it can be derived, let it be derived). But in this case, we're not adding a new field, we're using an existing one, for the purpose it was intended. 

And in fact, if a miner wants to lock up a UTXO for longer than the minimum maturity (I don't know why they would, but 🤷) then they are free to do so.

PR #4174, opened 2022-06-08
refactor(core)!: replace OutputFlags with OutputType

sdbondi:
Description
---
- Replaces `OutputFlags` With`OutputType` enum
- Updates transactionBuilder for integration tests
- Updates test faucet json files (The reason this PR is so massive!)

Motivation and Context
---
Many new types of UTXOs are being added to the protocol. The OutputFlags type uses 2 bytes to represent up to 16 different types. These flags could not meaningfully be combined which is the primary use case of using bitflags.

This PR replaces the 2 byte bitflags with a 1-byte enum that represents the various output types available.

**New dependency**
`serde_repr` - allows repr enums to be de/serialized with serde as an integer (https://serde.rs/enum-number.html)

How Has This Been Tested?
---
Existing tests updated as necessary


delta1:
Nice those output flags had definitely gotten unwieldy 

concept ack

stringhandler (approved):

PR #4175, opened 2022-06-08
fix(test): unifying dan layer integration tests

mrnaveira:
Description
---
* Solving cucumber tag duplication of `@dan_layer` and `@dan` tags, used only the `@dan` tag in all related tests
* Mark as broken all old tests for the DAN layer, they do not pass anymore
* Mark as broken the recent contract acceptance integration test. Some recent change did seem to broke it, it will be fixed in a separate PR in the future
* For name consistency, renaming the term _"constitution definition"_ for the more appropriate _"contract constitution"_
* Adapted the wallet CLI tests for contract definition and constitution to the new CLI (the `--command` option is no longer valid)
* Tagged the working DAN layer tests as `@critical`, to run them on new PR submissions to avoid breaking changes

Motivation and Context
---
It's desirable to be able to run all DAN related integration tests together under the same tag, and know if we are breaking existing functionalities when developing new features in the future.

How Has This Been Tested?
---
Now, the cucumber integration tests ran with `--tags "@dan and not @broken"` pass


therustmonk (dismissed):
LGTM

integration_tests/features/WalletCli.feature
```diff
@@ -152,26 +152,26 @@ Feature: Wallet CLI
...
+    @dan @critical
+    Scenario: As a user I want to publish a contract constitution via command line
         Given I have a base node BASE
         And I have wallet WALLET connected to base node BASE
         And I have mining node MINE connected to base node BASE and wallet WALLET
         And mining node MINE mines 4 blocks
         Then I wait for wallet WALLET to have at least 1000000 uT
-        And I create a "constitution-definition" from file "fixtures/constitution_definition.json" on wallet WALLET via command line
-        And mining node MINE mines 4 blocks
+        And I publish a contract constitution from file "fixtures/contract_constitution.json" on wallet WALLET via command line
```
sdbondi, line 174: The contract definition must exist before the constitution is published. In future this test will break, so I think you'll need to add a contract definition first in this test.

integration_tests/helpers/walletProcess.js
```diff
@@ -278,22 +280,26 @@ class WalletProcess {
...
       args.push("--log-config", this.logFilePath);
     }
+
     const overrides = this.getOverrides();
     Object.keys(overrides).forEach((k) => {
       args.push("-p");
       args.push(`${k}=${overrides[k]}`);
     });
+
+    // In the new cli commands are just arguments on the console wallet
```
sdbondi: nit: This comment probably won't make sense in a few weeks
```suggestion
    // Append command arguments 
```

sdbondi (dismissed):
Thanks LGTM - Ack (tested in CI)

Approving because comments are minor / all tests passing. We reapprove if suggested changes are made.

integration_tests/features/WalletCli.feature
sdbondi: This turns out to cause issues, resolving this because it can be fixed in another PR

mrnaveira: I agree. Seems like by publishing both the definition and constitution in the same test, the definition UTXO gets picked as an input for the constitution UTXO, and we don't want that. As the fix is a bit delicate, we can isolate it into a separated PR.

sdbondi (approved):

PR #4180, opened 2022-06-10
feat(wallet-ffi): adds FFI wallet_get_utxos() and test_wallet_get_utxos()

agubarev:
Motivation and Context
---
To allow client app (mobile) to get a paginated list of unspent outputs, filtered by a minimum value threshold.

How Has This Been Tested?
---
unit test

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -47,9 +47,9 @@
 //!     becoming a `CompletedTransaction` with the `Completed` status. This means that the transaction has been
 //!     negotiated between the parties and is now ready to be broadcast to the Base Layer. The funds are still
 //!     encumbered as pending because the transaction has not been mined yet.
-//! 3.  The finalized `CompletedTransaction' will be sent back to the the receiver so that they have a copy.
-//! 4.  The wallet will broadcast the `CompletedTransaction` to a Base Node to be added to the mempool. its status will
-//!     from `Completed` to `Broadcast.
+//! 3.  The finalized `CompletedTransaction` will be sent back to the the receiver so that they have a copy.
+//! 4.  The wallet will broadcast the `CompletedTransaction` to a Base Node to be added to the mempool. Its status will
+//!     move from `Completed` to `Broadcast`.
 //! 5.  Wait until the transaction is mined. The `CompleteTransaction` status will then move from `Broadcast` to `Mined`
 //!     and the pending funds will be spent and received.
 //!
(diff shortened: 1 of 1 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4093,6 +4106,99 @@ pub unsafe extern "C" fn wallet_get_balance(wallet: *mut TariWallet, error_out:
...
+    debug!(
+        target: LOG_TARGET,
+        "page = {:#?} page_size = {:#?} sort_asc = {:#?} dust_threshold = {:#?}",
+        page,
+        page_size,
+        sort_ascending,
+        dust_threshold
+    );
+
+    let rt = match Runtime::new() {
```
sdbondi: The wallet already has a runtime (it's an anti-pattern to create multiple runtimes)

`wallet.runtime`

sdbondi (commented):
Nice looks good - just one change RE the runtime.
Please also add details to the PR description

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4093,6 +4106,99 @@ pub unsafe extern "C" fn wallet_get_balance(wallet: *mut TariWallet, error_out:
...
+        Err(e) => {
+            error = LibWalletError::from(InterfaceError::TokioError(e.to_string())).code;
+            ptr::swap(error_out, &mut error as *mut c_int);
+            return ptr::null_mut();
+        },
+    };
+
+    match rt.block_on((*wallet).wallet.output_manager_service.get_unspent_outputs()) {
+        Ok(mut unblinded_outputs) => {
+            unblinded_outputs.sort_by(|a, b| match sort_ascending {
```
sdbondi: Ideally ordering and pagination should happen in the DB layer but we can refactor that later

base_layer/wallet_ffi/src/lib.rs
agubarev: Ah, that's right, removed new runtime initialization and using:
```rust
    match (*wallet)
        .runtime
        .block_on((*wallet).wallet.output_manager_service.get_unspent_outputs())
    {
        Ok(mut unblinded_outputs) => {
```

like in other places, thanks!

base_layer/wallet_ffi/src/lib.rs
agubarev: Yes, I have it planned after I get a better grasp of the project basics.

stringhandler (approved):

PR #4183, opened 2022-06-13
feat: generate wallet ffi header file automatically

stringhandler:
Description
---
Adds a `build.rs` and `cbindgen` to generate the header file automatically. I was not sure of the best location for this, so kept it the same as the previous `wallet.h` file, but renamed the file to match the crate (i.e. `tari_wallet_ffi.h`)

Motivation and Context
---
Currently, the `wallet.h` header file was manually edited. This obviously could lead to user error. Most notably, the `*_get_timestamp` methods were returning `c_longlong` but the header file indicated that they were using `c_ulonglong`. I'm not sure how long it has been like this, or which is correct, so I've changed the method to use `c_ulonglong` to match the previous header file.  

How Has This Been Tested?
---
Tested locally, but also asked @TruszczynskiA to compile iOs with the new header


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2437,15 +2418,15 @@ pub unsafe extern "C" fn completed_transaction_get_fee(
 pub unsafe extern "C" fn completed_transaction_get_timestamp(
     transaction: *mut TariCompletedTransaction,
     error_out: *mut c_int,
-) -> c_longlong {
+) -> c_ulonglong {
```
hansieodendaal, line 2421: I believe this is the correct choice here and ditto for the rest

hansieodendaal (dismissed):
LGTM

hansieodendaal:
Works really well for the auto header generation!

sdbondi (approved):

sdbondi:
This is great!

PR #4187, opened 2022-06-13
fix(validator): set tor_identity base path

sdbondi:
Description
---
sets the validator node tor identity file path relative to the base path if not absolute

Motivation and Context
---
All non-absolute paths should be relative to the base path

How Has This Been Tested?
---
Manually, tor identity file is relative to the base path


brianp:
Tested. The file ended up in the proper relative path. Looks good

brianp (approved):

stringhandler (approved):

PR #4185, opened 2022-06-13
refactor: constitution subcommands

brianp:
Description
---
This moves the previous command usage under the new contract subcommand and provides a constitution generator for quickly creating new template files.

Motivation and Context
---
Easier to use, and easier to read.

How Has This Been Tested?
---
~It hasn't. Another bug is now preventing the creation and publishing of contracts or constitutions.~

Manually tested. Both definition, and constitution can generate and publish (given the correct inputs)

PR #4188, opened 2022-06-13
feat(wallet): new command to publish a contract update proposal

mrnaveira:
Description
---
New subcommand in the wallet CLI (`contract publish-update-proposal`):
* Reads a proposal description from a JSON file
* Creates a transaction with the appropriate features
* Publishes the transaction into the network

Motivation and Context
---
We need a way to publish a contract update proposal into the blockchain, via a wallet command

How Has This Been Tested?
---
The new integration test pass


applications/tari_console_wallet/src/cli.rs
```diff
@@ -109,6 +109,7 @@ pub enum CliCommands {
     FinaliseShaAtomicSwap(FinaliseShaAtomicSwapArgs),
     ClaimShaAtomicSwapRefund(ClaimShaAtomicSwapRefundArgs),
     PublishConstitutionDefinition(PublishDefinitionArgs),
+    PublishContractUpdateProposal(PublishUpdateProposalArgs),
```
brianp: [We're refactoring all the contract related commands to subcommands](https://github.com/tari-project/tari/blob/ed1d7b38cce49538d2a504ccd8381eebf3e1b6d7/applications/tari_console_wallet/src/cli.rs#L209-L216) and this one should probably be included there as well. It'll likely have some conflicts with the refactoring [here as well](https://github.com/tari-project/tari/pull/4185)

brianp (commented):
Looks good. One comment about the command interface that will bring about some conflicts. 

Otherwise tested and it broadcasted fine. It was rejected for an issue with inputs but this is a known issue unrelated to this PR. At the point the update was broadcast properly I consider this a working PR.

applications/tari_console_wallet/src/cli.rs
mrnaveira: Thanks for the info, I just adapted the PR to the new command organisation (which is much cleaner IMO)

stringhandler (approved):

PR #4191, opened 2022-06-14
fix(validator-node): return error if contract_id empty for publish_contract_acceptance grpc

sdbondi:
Description
---
Returns invalid argument if contract id is not provided
Log error if publishing fails

Motivation and Context
---
If contract_id was not provided, a zero hash would be used and published.

How Has This Been Tested?
---
Manually


brianp (commented):
So the problem was more the use of `unwrap_or_default();` providing a zero hash?

How was it possible to pass an empty argument with grpc. Doesn't it normally validate the request structure. So a contract id needs to be present for the request to be accepted.

sdbondi:
@brianp All fields in protobuf are optional - so if you dont provide the contract_id, it will be, in the base of bytes, an empty vec - since that obviously isnt a 32-byte vec/slice the try_from will fail and the default will be used (i.e. zeros) which would then be passed to create constitution. 

This change will error out if the contract_id isnt 32-bytes (e.g was not passed in at all)

stringhandler (approved):

PR #4192, opened 2022-06-14
fix(test): integration test for validator node is broken

mrnaveira:
Description
---
* Returned a specific log line back to the standard output, to be able to detect that the validator node finished starting up from the integration tests
* Fixed the compilation of the validator node application on integration test startup
* Tagged the existing integration tests for contract acceptances as `@critical` so any broken changes on future pull requests will be spotted

Motivation and Context
---
We want to be able to run tests on validator node operations like contract acceptance. It was broken at some point.

How Has This Been Tested?
---
The existing integration test for contract acceptance, done through the validator node, now runs successfully


stringhandler (approved):

sdbondi (approved):
Nice!

brianp (approved):
👌🏻

PR #4091, opened 2022-05-11
refactor(core): upgrade randomx-rs to 1.1.11

sdbondi:
Description
---

- upgrades randomx-rs to 1.1.11
- removes redundant RwLock from `RandomXVMInstance` 
- remove _cache and _dataset fields as `RandomXVM` now keeps an owned copy of these instances

Motivation and Context
---
Includes improvements from previous versions of `randomx-rs`.
The RwLock in `RandomXVMInstance` was only used non-mutably which defeats the purpose of a RwLock.

~~NOTE: this PR includes cross-compilation changes in https://github.com/tari-project/randomx-rs/pull/43
These should be merged and released before merging this PR.~~ Merged

How Has This Been Tested?
---
Existing tests pass, manually

PR #4196, opened 2022-06-14
chore(deps): upgrade randomx

CjS77:
Back ported from https://github.com/tari-project/tari/pull/4091

Improve cross-compile in docker


CjS77:
Tests pass locally, so going to merge this

PR #4197, opened 2022-06-14
feat(wallet_get_utxos): wallet_get_utxos() and test_wallet_get_utxos()

agubarev:
Minor changes of  `wallet_get_utxos()` and added `destroy_tari_outputs()`

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -212,7 +212,6 @@ pub struct TariSeedWords(Vec<String>);
 #[derive(Debug, PartialEq)]
 pub struct TariPublicKeys(Vec<TariPublicKey>);
 
-#[repr(C)]
 pub struct TariWallet {
     pub wallet: WalletSqlite,
     pub runtime: Runtime,
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -47,9 +47,9 @@
 //!     becoming a `CompletedTransaction` with the `Completed` status. This means that the transaction has been
 //!     negotiated between the parties and is now ready to be broadcast to the Base Layer. The funds are still
 //!     encumbered as pending because the transaction has not been mined yet.
-//! 3.  The finalized `CompletedTransaction' will be sent back to the the receiver so that they have a copy.
-//! 4.  The wallet will broadcast the `CompletedTransaction` to a Base Node to be added to the mempool. its status will
-//!     from `Completed` to `Broadcast.
+//! 3.  The finalized `CompletedTransaction` will be sent back to the the receiver so that they have a copy.
+//! 4.  The wallet will broadcast the `CompletedTransaction` to a Base Node to be added to the mempool. Its status will
+//!     move from `Completed` to `Broadcast`.
 //! 5.  Wait until the transaction is mined. The `CompleteTransaction` status will then move from `Broadcast` to `Mined`
 //!     and the pending funds will be spent and received.
 //!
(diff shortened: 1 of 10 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -212,26 +212,31 @@ pub struct TariSeedWords(Vec<String>);
 #[derive(Debug, PartialEq)]
 pub struct TariPublicKeys(Vec<TariPublicKey>);
 
+#[repr(C)]
 pub struct TariWallet {
-    wallet: WalletSqlite,
-    runtime: Runtime,
-    shutdown: Shutdown,
+    pub wallet: WalletSqlite,
```
stringhandler, line 222: Are these needed?

agubarev: Not really, not yet.

agubarev: It doesn't do any harm though, but not much use too because the rest is still only just `typedef`s.

stringhandler (dismissed):

delta1:
Not really a fan of breaking the stable build with that #[feature] - it was a long journey to get stable compiling. 

Surely there’s a stable way to do the same? 🤷

agubarev:
> Not really a fan of breaking the stable build with that #[feature] - it was a long journey to get stable compiling.
> 
> Surely there’s a stable way to do the same? shrug

Thanks for pointing out and thanks to @sdbondi for the tip about `ManuallyDrop::new()`.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4189,6 +4213,24 @@ pub unsafe extern "C" fn wallet_get_utxos(
...
+/// ## Returns
+/// `()` - Does not return a value, equivalent to void in C
+///
+/// # Safety
+/// None
+#[no_mangle]
+pub unsafe extern "C" fn destroy_tari_outputs(x: *mut TariOutputs) {
+    if !x.is_null() {
+        Vec::from_raw_parts((*x).ptr, (*x).len, (*x).cap);
+        Box::from_raw(x);
```
sdbondi, line 4230: Interesting that you construct and drop the Vec AND Box - you would think that dropping the Vec is enough? Since the data is dropped in the line above (I think that is how rust does it if you don't bind to a variable) would Box::from_raw now not point to unallocated data?

agubarev: I'm not sure I completely understood your point.

agubarev: Boxed value is a struct containing internal pointer, its length and capacity. I'm following the [docs on into_raw_parts](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_raw_parts). I'm not sure that reconstructing and dropping the box would release the memory allocated for the internal array pointer, would it?

Is your concern that boxed value would not be properly dropped by going out of scope if its internal, pointed-to allocation is freed separately?

sdbondi: Ok I think that looks right, the len, cap and ptr are allocated on the heap with Box::into_raw(Box::new(...)) and that is still allocated.

sdbondi (dismissed):
Looks great - I like the general purpose output query interface

base_layer/wallet_ffi/src/lib.rs
sdbondi: Only saw your comments now - questions for my own interest, trying to understand as I'm not an `unsafe` expert 😄  

Thanks, makes sense to me now

sdbondi (approved):

agubarev:
closing to re-open with signed commits

PR #4205, opened 2022-06-15
feat(wallet): new cli commands to initialise proposals and amendments

mrnaveira:
Description
---
* Created a new `init-update-proposal` subcommand to create a proposal JSON file. 
    * The fields are very similar to a constitution, only the `proposal_id` is new.
    * For now, the `signature` is set to the default value. In the future we will need to implement a way for the user to be able to sign the proposal via a private key.
* Create a new `init-amendment` subcommand to create an amendment JSON file. 
    * The user is required to specify an update proposal JSON file path with the proposal to amend.
    * Asks for the `activation_window` to the user, which specifies the amount of blocks until the changes are enforced by the base layer.
    * For now, the list of signatures for each accepting committee member is set as empty. In the future we will need to retrieve those from the acceptance transactions themselves.
* Renamed the contract command handling function for name consistency.

Motivation and Context
---
We want an easy way to initialise both contract update proposals and amendments JSON files, to publish them later with other CLI commands.

How Has This Been Tested?
---
Manually


brianp (approved):
Looks good to me, untested.

stringhandler (approved):

PR #4207, opened 2022-06-16
feat: modularise ci workflows 

CjS77:
To make CI workflows more manageable and allow us to run some cheaper
tests on non-core development (e.g. launchpad), this PR does the
following:

- breaks out jobs into reuasable workflows
- adds several parameters to these workflows to DRY up the CI

- adds separate workflows for specific application branches while
  keeping the core CI flow the same.

- renames workflow files to be more indicative of when they run & what
they do.

- Delete some redundant workflows


.github/workflows/ci.yml
```diff
@@ -16,217 +16,51 @@ on:
...
-        with:
-          command: check
-          args: --release --package tari_wallet_ffi
+    uses:
+      ./.github/workflows/reusable-ci-jobs.yml
+    with:
+      check-ffi: true
+
+  check-stable:
+    uses:
```
stringhandler, line 32: Not sure if this works, but we need this to run the jobs on self-hosted for better run times
```suggestion
    runs-on: [ self-hosted, ubuntu18.04-high-cpu ]
    uses:
```

stringhandler: And same for other jobs

.github/workflows/reusable-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      test-wasm:
+        # flag to run any wasm tests. Includes key manager
+        required: true
+        type: boolean
+
+
+jobs:
+  tests:
+    name: test
+    runs-on: ubuntu-18.04
```
stringhandler, line 20: Or is it better to replace all of these?
```suggestion
    runs-on: [ self-hosted, ubuntu18.04-high-cpu ]
```

.github/workflows/ci.yml
CjS77: Probably, but I left everything as they're currently configured.

CjS77:
Nuke this for now. GH needs to add additional flexibility to Actions before this is realistic with self-hosted runners

PR #4210, opened 2022-06-17
feat(validator_node): add global db

brianp:
Description
---
This PR intorduces a new global database for the validator node. This is so it can store relevant state to help process contracts such as:
- the constitutions it's a part of so it can query acceptances
- the last scanned block hash so we don't always scan from genesis

Motivation and Context
---
Everytime the VN would scan for constitutions it would do so from the genesis block. It would then re-accept the same constitutions on every loop. We had no way to keep track of already scanned block, or what constitutions we're a part of. Long term we needed a solution and not a bandaid so we introduced a global state db.

How Has This Been Tested?
---
Manually. Ensuring the VN no longer eager accepts constitutions non-stop. (not actually tested yet).

### DIFF
--- a/dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
+++ b/dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
@@ -84,6 +84,8 @@ impl GlobalDbBackendAdapter for SqliteGlobalDbBackendAdapter {
                 })?,
         };
 
+        self.commit(&tx)?;
+
         Ok(())
     }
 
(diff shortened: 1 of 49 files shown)


dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
```diff
@@ -0,0 +1,105 @@
...
+            .execute("BEGIN EXCLUSIVE TRANSACTION;")
+            .map_err(|source| SqliteStorageError::DieselError {
+                source,
+                operation: "begin transaction".to_string(),
+            })?;
+
+        Ok(SqliteTransaction::new(connection))
+    }
+
+    fn set_data(&self, key: &[u8], value: &[u8]) -> Result<(), Self::Error> {
```
stringhandler, line 62: I think create 2 methods, one where you pass in a transaction, and one where you create a transaction

dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
```diff
@@ -0,0 +1,105 @@
...
+                    source,
+                    operation: "insert::metadata".to_string(),
+                })?,
+            Err(e) => return Err(e),
+        };
+
+        Ok(())
+    }
+
+    fn get_data(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
```
stringhandler, line 92: Same here

stringhandler (approved):

PR #4213, opened 2022-06-20
feat: swap dalek bulletproofs for bulletproofs-plus

hansieodendaal:
Description
---
Swapped out Dalek Bulletproofs for Tari Bulletproofs+ using the most basic range proof creation and verification

Motivation and Context
---

How Has This Been Tested?
---


base_layer/core/src/transactions/transaction_components/test.rs
```diff
@@ -137,12 +137,12 @@ fn range_proof_verification() {
         },
     }
 
-    let value = 2u64.pow(32) + 1;
+    let value = 2u64.pow(32);
```
stringhandler: This change looks a bit sus

hansieodendaal: Ok I will fix that test. It will fail with `construct_proof` if the value is too large.

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -610,27 +607,14 @@ impl OutputManagerHandle {
         }
     }
 
-    pub async fn get_outputs_by(&mut self, q: OutputBackendQuery) -> Result<Vec<UnblindedOutput>, OutputManagerError> {
```
stringhandler, line 613: This seems to be a bad merge

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -88,28 +89,24 @@ where
...
-                    return None;
-                }
-                let script_key = PrivateKey::random(&mut OsRng);
-                Some((
-                    UnblindedOutput::new(
+
+        let mut rewound_outputs: Vec<(UnblindedOutput, BulletRangeProof)> = Vec::new();
+        for output in outputs {
+            // TODO: Only outputs with scripts `== script!(Nop)` is recover-able - can this be improved?
+            if output.script != script!(Nop) {
```
stringhandler, line 96: Is this really the case?

stringhandler (approved):

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
hansieodendaal: The previous version of the code had this exact check - if a rewound was successful but the script was not `script!(Nop)` it dropped the output. Maybe this is latent defect? That is why I logged a `TODO`, but I will also register an issue.

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: O, that was a piece of unused code that I cleaned up.

PR #4215, opened 2022-06-21
refactor(validator_node): utilize existing transactions

brianp:
Share the active connection when fetching and setting data instead of
creating two different connections.

Description
---
This is a refactoring of the global state db to utilize existing connections when performing multiple functions. 

Motivation and Context
---
[Requested in the original PR](https://github.com/tari-project/tari/pull/4210). Helps to avoid db locked errors, and cleans up the amount of transactions used.

How Has This Been Tested?
---
Manually


dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
```diff
@@ -105,6 +106,26 @@ impl GlobalDbBackendAdapter for SqliteGlobalDbBackendAdapter {
         Ok(row.map(|r| r.value))
     }
 
+    fn get_data_with_connection(
+        &self,
+        key: &[u8],
+        connection: &SqliteConnection,
```
stringhandler: It's more useful to provide a transaction so that multiple calls use it before committing. Also, there is no method for getting a connection, but there is a method for getting a transaction. And finally, the connection can be retrieved from the transaction

```suggestion
        tx: &Self::BackendTransaction,
```

dan_layer/storage_sqlite/src/sqlite_global_db_backend_adapter.rs
```diff
@@ -38,6 +38,7 @@ impl SqliteGlobalDbBackendAdapter {
 
 impl GlobalDbBackendAdapter for SqliteGlobalDbBackendAdapter {
     type BackendTransaction = SqliteTransaction;
+    type Connection = SqliteConnection;
```
stringhandler: Not needed, see other comment

stringhandler (approved):

PR #4216, opened 2022-06-21
feat(validator_node): auto accept flag and better logging

brianp:
Description
---
This adds the `-a` and `--auto-accept` flag on start up to override the configuration setting.
Also cleans up info lines and stdout output to help get a sense of what the VN is doing.

Motivation and Context
---
Logging was poor and a startup flag > config changes

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_validator_node/src/dan_node.rs
+++ b/applications/tari_validator_node/src/dan_node.rs
@@ -38,7 +38,7 @@ use crate::{
     grpc::services::{base_node_client::GrpcBaseNodeClient, wallet_client::GrpcWalletClient},
 };
 
-const _LOG_TARGET: &str = "tari::validator_node::app";
+const LOG_TARGET: &str = "tari::validator_node::app";
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


stringhandler (approved):

applications/tari_validator_node/src/dan_node.rs
```diff
@@ -65,11 +65,16 @@ impl DanNode {
         let node = self.clone();
 
         if self.config.constitution_auto_accept {
+            println!("Contract constitution auto acceptance feature is TRUE");
```
stringhandler, line 68: Would prefer an `info!` but println is fine for now

brianp: I originally had it as info but wanted it displaying to stdout by default and it seems info! lines aren't currently.

brianp:
Closing this PR. Auto acceptance was temporarily removed in the checkpoints PR and won't return in a form that makes this PR worth keeping.

PR #4218, opened 2022-06-22
feat(wallet_ffi): add coin join and split

agubarev:
Description
---
Adding new FFI `wallet_coin_join()` function and working on unit tests for `wallet_coin_join()` and an already existing `wallet_coin_split()`

Motivation and Context
---
To enable users to join (collapse) UTXOs depending on the selected strategy, with default behavior to join outputs selected by the user.

How Has This Been Tested?
---
unit tests (WIP)


base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -101,10 +105,16 @@ pub enum OutputManagerRequest {
...
     GetUnspentOutputs,
+    GetOutputsBy(OutputBackendQuery),
     GetInvalidOutputs,
     ValidateUtxos,
     RevalidateTxos,
     CreateCoinSplit((MicroTari, usize, MicroTari, Option<u64>)),
+    CreateCoinJoin {
+        target_amount: MicroTari,
+        fee_per_gram: MicroTari,
+        commitments: Option<Vec<PublicKey>>,
```
stringhandler: Rather use an empty vec to represent "None"
```suggestion
        commitments: Vec<PublicKey>,
```

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -670,6 +699,26 @@ impl OutputManagerHandle {
         }
     }
 
+    pub async fn create_coin_join(
+        &mut self,
+        target_amount: MicroTari,
+        fee_per_gram: MicroTari,
+        commitments: Option<Vec<PublicKey>>,
```
stringhandler:
```suggestion
        commitments: Vec<PublicKey>,
```

agubarev: Done

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -218,20 +230,175 @@ pub struct TariWallet {
...
+    fn from_pubkey_vec(v: &mut Vec<PublicKey>) -> Result<Self, InterfaceError> {
+        Ok(Self {
+            tag: TariTypeTag::PublicKey,
+            len: v.len(),
+            cap: v.capacity(),
+            ptr: v.as_mut_ptr() as *mut c_void,
+        })
+    }
+
+    fn to_string_vec(&self) -> Result<Vec<String>, InterfaceError> {
```
sdbondi, line 319: The `to_*` functions should be marked as unsafe because they do not provide any additional safety guarantees than from_raw_parts (which is unsafe). If the data that TariVector points to is null/deallocated the Vec<String> will be invalid and the program will crash. Moreover if the vector is mutated (as you are free to do with an owned Vec, it could be reallocated and the pointer is no longer valid.

~~I think the method signature should be:
`unsafe fn as_string_slice(&self) -> Result<&[&str], InterfaceError>`~~ wont work

stringhandler (dismissed):
I would have preferred less duplication of code, but I'm not sure how often this functionality is actually going to be used, so let's try it and we can DRY  it up later

stringhandler (approved):

PR #4222, opened 2022-06-22
feat(wallet): adds contract_id to outputs db

sdbondi:
Description
---
- adds `contract_id` to outputs table in wallet db
- renames `flags` to `output_type` in wallet db

Motivation and Context
---
There are various use cases that will require to select outputs by contract_id.
The outputs table now exceeds 32 columns, so the `64-column-table` diesel feature needed to be added.

How Has This Been Tested?
---
Manually, contract id is populated


base_layer/wallet/Cargo.toml
```diff
@@ -37,7 +37,7 @@ chrono = { version = "0.4.19", default-features = false, features = ["serde"] }
 clear_on_drop = "=0.2.4"
 crossbeam-channel = "0.5.4"
 derivative = "2.2.0"
-diesel = { version = "1.4.8", features = ["sqlite", "serde_json", "chrono"] }
+diesel = { version = "1.4.8", features = ["sqlite", "serde_json", "chrono", "64-column-tables"] }
```
stringhandler, line 40: This usually increases compile time by a large amount. Can we drop a column maybe to fit into 32?

stringhandler: The CI didn't seem to take too long, so maybe not so bad

stringhandler (approved):

base_layer/wallet/Cargo.toml
sdbondi: Yeah we can drop  parent_pk and/or unique_id later and remove this feature.

PR #4223, opened 2022-06-22
test(dan): fix broken dan layer integration tests

brianp:
Description
---

- Fix any tests marked `@broken`
- Remove old layer 2 code
- Add more specific assertions for tests
- Add a CI workflow for dan layer tests to ensure PR's don't break existing functionality

Motivation and Context
---
We can't keep letting tests break and spending lots of time backdating the changes. Ensure they work, and continue to work moving forward.

How Has This Been Tested?
---
Oh- the tests have been tested alright


.github/workflows/dan_layer_integration_tests.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           command: build
           args: --release --bin tari_validator_node
       - name: npm ci
-        run: cd integration_tests && npm ci && cd node_modules/wallet-grpc-client && npm ci
+        run: cd integration_tests && npm ci && cd node_modules/wallet-grpc-client && npm install
```
delta1: Doing this means npm might resolve new packages instead of what’s in package-lock

brianp: Currently the package-lock is gitignored. So the best solution is likely to commit  it.

delta1: Yeah I thought so but I think Mike removed them a while back? 🤷‍♂️

stringhandler: lol it gets deleted every now and again. It should be included

brianp: Added!

integration_tests/features/ValidatorNode.feature
```diff
@@ -3,61 +3,45 @@
...
-        When I wait 30 seconds
-        When I register an NFT asset with committee of 4
-        And I mine 3 blocks
-        And I create 40 NFTs
-        And I mine 3 blocks
-
-    # Broken: needs a contract definition before publishing acceptance, however this is currently not easily done because
-    # GRPC methods need to be added and you cannot use the cli for a wallet while that wallet is already running
-    @dan @critical @broken
+    @dan @critical
```
stringhandler: No need to add the `@dan` tag if it's specified at the feature level already

stringhandler (dismissed):

stringhandler (approved):

PR #4225, opened 2022-06-22
feat: service for encrypting value

therustmonk:
Description
---
* The PR implements encryption of the `EncryptedValue` type.
* The new methods `encrypt_value` and `decrypt_value` added to the `OutputManagerService` (handle).
* Implemented `wallet_ffi` methods: `encrypted_value_encrypt` and `encrypted_value_decrypt`

Motivation and Context
---
The part of the BP+ feature.

How Has This Been Tested?
---
Unit tests + CI

### DIFF
--- a/applications/test_faucet/src/main.rs
+++ b/applications/test_faucet/src/main.rs
@@ -200,7 +200,7 @@ fn create_utxo(
     if !factories.range_proof.verify(&proof, &commitment) {
         panic!("Range proof does not verify");
     };
-    let encrypted_value = EncryptedValue::todo_encrypt_from(value);
+    let encrypted_value = EncryptedValue::default();
     let metadata_sig = TransactionOutput::create_final_metadata_signature(
         TransactionOutputVersion::get_current_version(),
         value,
--- a/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
@@ -25,10 +25,23 @@
 
 use std::io::{self, Read, Write};
 
+use blake2::Digest;
+use chacha20poly1305::{
+    aead::{Aead, Error, NewAead, Payload},
+    ChaCha20Poly1305,
+    Key,
+    Nonce,
+};
 use serde::{Deserialize, Serialize};
+use tari_common_types::types::{Commitment, PrivateKey};
+use tari_crypto::common::Blake256;
 use tari_utilities::{ByteArray, ByteArrayError};
+use thiserror::Error;
 
-use crate::consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized};
+use crate::{
+    consensus::{ConsensusDecoding, ConsensusEncoding, ConsensusEncodingSized},
+    transactions::tari_amount::MicroTari,
+};
 
 const SIZE: usize = 24;
 
@@ -52,23 +65,66 @@ impl ByteArray for EncryptedValue {
     }
 }
 
+#[derive(Debug, Error)]
+pub enum EncryptionError {
+    #[error("Encryption failed: {0}")]
+    EncryptionFailed(Error),
+}
+
+// chacha error is not StdError compatible
+impl From<Error> for EncryptionError {
+    fn from(err: Error) -> Self {
+        Self::EncryptionFailed(err)
+    }
+}
+
 impl EncryptedValue {
-    /// TODO: Replace this method with a real call of encryption service
-    /// that will produce an encrypted value from the given `amount`.
-    pub fn todo_encrypt_from(amount: impl Into<u64>) -> Self {
+    const TAG: &'static [u8] = b"TARI_AAD_VALUE";
+
+    pub fn encrypt_value(
+        encryption_key: &PrivateKey,
+        commitment: &Commitment,
+        value: MicroTari,
+    ) -> Result<EncryptedValue, EncryptionError> {
+        let aead_key = kdf_aead(encryption_key, commitment);
+        // Encrypt the value (with fixed length) using ChaCha20-Poly1305 with a fixed zero nonce
+        let aead_payload = Payload {
+            msg: &value.as_u64().to_le_bytes(),
+            aad: Self::TAG,
+        };
+        // Included in the public transaction
+        let buffer = ChaCha20Poly1305::new(&aead_key).encrypt(&Nonce::default(), aead_payload)?;
         let mut data: [u8; SIZE] = [0; SIZE];
-        let value = amount.into().to_le_bytes();
-        data[0..8].copy_from_slice(&value);
-        Self(data)
+        data.copy_from_slice(&buffer);
+        Ok(EncryptedValue(data))
     }
 
-    /// TODO: Replace this method with a real call of decryption service
-    /// that will produce a decrypted value from self.
-    pub fn todo_decrypt(&self) -> u64 {
-        let mut buffer = [0u8; 8];
-        (&mut buffer[0..8]).copy_from_slice(&self.0.as_slice()[0..8]);
-        u64::from_le_bytes(buffer)
+    pub fn decrypt_value(
+        encryption_key: &PrivateKey,
+        commitment: &Commitment,
+        value: &EncryptedValue,
+    ) -> Result<MicroTari, EncryptionError> {
+        let aead_key = kdf_aead(encryption_key, commitment);
+        // Authenticate and decrypt the value
+        let aead_payload = Payload {
+            msg: value.as_bytes(),
+            aad: Self::TAG,
+        };
+        let mut value_bytes = [0u8; 8];
+        let decrypted_bytes = ChaCha20Poly1305::new(&aead_key).decrypt(&Nonce::default(), aead_payload)?;
+        value_bytes.clone_from_slice(&decrypted_bytes[..8]);
+        Ok(u64::from_le_bytes(value_bytes).into())
     }
+}
+
+// Generate a ChaCha20-Poly1305 key from an ECDH shared secret and commitment using Blake2b
+fn kdf_aead(shared_secret: &PrivateKey, commitment: &Commitment) -> Key {
+    const AEAD_KEY_LENGTH: usize = 32; // The length in bytes of a ChaCha20-Poly1305 AEAD key
+    let mut hasher = Blake256::with_params(&[], b"SCAN_AEAD".as_ref(), b"TARI_KDF".as_ref());
+    hasher.update(shared_secret.as_bytes());
+    hasher.update(commitment.as_bytes());
+    let output = hasher.finalize();
+    *Key::from_slice(&output[..AEAD_KEY_LENGTH])
 }
 
 impl ConsensusEncoding for EncryptedValue {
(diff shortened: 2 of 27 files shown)


base_layer/core/src/transactions/encryption_service.rs
```diff
@@ -0,0 +1,70 @@
+//  Copyright 2022, The Tari Project
```
stringhandler: I'd prefer this file to be under `core/src/services`

base_layer/core/src/transactions/encryption_service.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[derive(Debug, Clone)]
+pub struct EncryptionServiceHandle {}
+
+impl EncryptionServiceHandle {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn encrypt_value(&self, value: MicroTari) -> EncryptedValue {
```
stringhandler: Probably needs a secret for the encryption as wel

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1168,20 +1181,28 @@ pub unsafe extern "C" fn encrypted_value_encrypt(
...
+
     if encrypted_value.is_null() {
         error = LibWalletError::from(InterfaceError::NullError("encrypted_value".to_string())).code;
         ptr::swap(error_out, &mut error as *mut c_int);
         return 0;
     }
 
-    let encrypted_value = (*encrypted_value).clone();
-    encrypted_value.todo_decrypt() as c_ulonglong
+    let value = (*wallet).wallet.encryption_service.decrypt_value(&*encrypted_value);
```
hansieodendaal: This method will throw an error if the value cannot be decrypted, which must be handled here

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -7763,9 +7782,67 @@ mod test {
...
+            );
 
             let amount = 1234u64;
-            let expected_encrypted_value = EncryptedValue::todo_encrypt_from(amount);
+            let encryption_service = EncryptionServiceHandle::new();
+            // TODO: Commitment? Encryption Key?
+            let encryption_key = todo!();
+            let commitment = todo!();
+            let expected_encrypted_value =
+                encryption_service.encrypt_value(&encryption_key, &commitment, amount.into());
```
hansieodendaal: This is not needed, as it is sufficient to test the round trip with `encrypted_value_encrypt` and `encrypted_value_decrypt`

base_layer/core/src/services/encryption_service.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    pub fn encrypt_value(&self, key: &PrivateKey, commitment: &Commitment, value: MicroTari) -> EncryptedValue {
+        const SIZE: usize = 24;
+        let mut data: [u8; SIZE] = [0; SIZE];
+        let value = u64::from(value).to_le_bytes();
+        data[0..8].copy_from_slice(&value);
+        EncryptedValue(data)
+    }
+
+    pub fn decrypt_value(&self, value: &EncryptedValue) -> MicroTari {
```
hansieodendaal: This method should throw an error if the value cannot be decrypted, as per Aaron's scanning poc

base_layer/core/src/services/encryption_service.rs
```diff
@@ -0,0 +1,72 @@
...
+// TODO: Rename to `EncryptionFactory`
+#[derive(Debug, Clone)]
+pub struct EncryptionServiceHandle {}
+
+impl EncryptionServiceHandle {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn encrypt_value(&self, key: &PrivateKey, commitment: &Commitment, value: MicroTari) -> EncryptedValue {
```
hansieodendaal: The `encryption_key` should be passed into this function

base_layer/core/src/services/encryption_service.rs
hansieodendaal: The `encryption_key` should be passed into this function

base_layer/core/src/services/encryption_service.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::transactions::{tari_amount::MicroTari, transaction_components::EncryptedValue};
+
+// use thiserror::Error;
+//
+// #[derive(Debug, Error)]
+// pub enum EncryptionServiceError {}
+
+// TODO: Rename to `EncryptionFactory`
+#[derive(Debug, Clone)]
+pub struct EncryptionServiceHandle {}
```
hansieodendaal: I propose to add Aaron's scanning poc code here to encrypt and decrypt the value with a `// TODO:` note to change it to use the `tari_crypto` methods when implemented there.

base_layer/core/src/services/encryption_service.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct EncryptionServiceInitializer {}
+
+impl EncryptionServiceInitializer {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+#[async_trait]
+impl ServiceInitializer for EncryptionServiceInitializer {
```
hansieodendaal: See comment above related to this not needing to be a true "service".

base_layer/core/src/services/encryption_service.rs
hansieodendaal: As per @stringhandler this should probably not be a true "service". As can be seen from Aaron's scanning poc, the value encryption and decryption is not complex and will only do one thing.

hansieodendaal: The encrypt and decrypt methods could probably just move back to `EncryptedValue`, as the only additional parameter needed will be the `encryption_key` if Aaron's scanning poc code is implemented directly into those methods.

hansieodendaal (commented):
Hi @DenisKolodin, some comments below. I think the required functionality is shaping nicely.

Having looked at Aaron's scanning poc together yesterday I believe we can simplify this to follow your original implementation philosophy without the need for a service. An encryption service in the true sense would need to offer more than just the value encryption.

comms/dht/src/crypt.rs
```diff
@@ -35,15 +35,16 @@ use tari_crypto::{
...
 
 use crate::{
     envelope::{DhtMessageFlags, DhtMessageHeader, DhtMessageType, NodeDestination},
     outbound::DhtOutboundError,
     version::DhtProtocolVersion,
 };
 
-#[derive(Debug, Clone, Zeroize, ZeroizeOnDrop)]
+#[derive(Debug, Clone, Zeroize)]
+#[zeroize(drop)]
```
therustmonk, line 47: In case if we use `chacha20poly1305` crate we should use `zeroize=1.4` to let cargo solve dependencies.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1732,10 +1757,11 @@ where
         let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
         // TODO: This may be usefull for the encrypted value decryption since this is an atomic swap -
         // TODO: when fixing 'todo_decrypt'
-        let _rewind_key = PrivateKey::from_bytes(&hash_secret_key(&blinding_key))?;
+        let rewind_key = PrivateKey::from_bytes(&hash_secret_key(&blinding_key))?;
+        let encryption_key = PrivateKey::from_bytes(&hash_secret_key(&rewind_key))?;
         // TODO: Fix this logic when 'todo_decrypt' - only commence if the tag is recognized
```
```suggestion
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1963,15 +1989,21 @@ where
                 )?;
                 let rewind_blinding_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
                 let recovery_byte_key = PrivateKey::from_bytes(&hash_secret_key(&rewind_blinding_key))?;
+                let encryption_key = PrivateKey::from_bytes(&hash_secret_key(&recovery_byte_key))?;
                 // TODO: Fix this logic when 'todo_decrypt' - only commence if the tag is recognized
```
```suggestion
```

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -97,15 +96,19 @@ where
                 continue;
             }
             // TODO: Fix this logic when 'todo_decrypt' - only commence if the tag is recognized
```
hansieodendaal:
```suggestion
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1732,10 +1757,11 @@ where
         let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
         // TODO: This may be usefull for the encrypted value decryption since this is an atomic swap -
         // TODO: when fixing 'todo_decrypt'
```
```suggestion
```

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1732,10 +1757,11 @@ where
         let blinding_key = PrivateKey::from_bytes(&hash_secret_key(&spending_key))?;
         // TODO: This may be usefull for the encrypted value decryption since this is an atomic swap -
```
```suggestion
        // TODO: This may be usefull for the encrypted value decryption since this is an atomic swap
```

applications/test_faucet/src/main.rs
```diff
@@ -200,7 +200,8 @@ fn create_utxo(
     if !factories.range_proof.verify(&proof, &commitment) {
         panic!("Range proof does not verify");
     };
-    let encrypted_value = EncryptedValue::todo_encrypt_from(value);
+    let encryption_keys = generate_keys();
+    let encrypted_value = EncryptedValue::encrypt_value(&encryption_keys.k, &commitment, value).unwrap();
```
hansieodendaal: I think we should be using the default value here, as faucet UTXOs are not recoverable without respending them.

hansieodendaal (commented):
Looking good apart from the the failing tests, also some comments

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -52,25 +65,66 @@ impl ByteArray for EncryptedValue {
...
+    pub fn decrypt_value(
+        encryption_key: &PrivateKey,
+        commitment: &Commitment,
+        value: &EncryptedValue,
+    ) -> Result<MicroTari, EncryptionError> {
+        let aead_key = kdf_aead(encryption_key, commitment);
+        // Authenticate and decrypt the value
+        let aead_payload = Payload {
+            msg: value.as_bytes(),
+            aad: b"TARI_AAD_SCAN".as_ref(),
```
stringhandler: I think it makes sense to use VALUE here, so that in future we can add extra fields
```suggestion
            aad: b"TARI_AAD_VALUE".as_ref(),
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1144,14 +1144,53 @@ pub unsafe extern "C" fn encrypted_value_as_bytes(
 /// memory leak
 #[no_mangle]
 pub unsafe extern "C" fn encrypted_value_encrypt(
```
stringhandler: I don't see the need for this method? I think it should be removed from the ffi interface

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1168,20 +1207,61 @@ pub unsafe extern "C" fn encrypted_value_encrypt(
 /// memory leak
 #[no_mangle]
 pub unsafe extern "C" fn encrypted_value_decrypt(
+    wallet: *mut TariWallet,
```
stringhandler: Same here, this method can be used to arbitrarily decrypt some data

stringhandler (commented):
Looks good. Happy to merge if we can remove: 
1. The FFI methods, they are not needed (and might even be a security risk)
2. The output manager handle methods (as far as I can see, they are only used by the FFI)

base_layer/wallet_ffi/src/lib.rs
hansieodendaal: Yip, this one was undecided in my head as well when I added it, so agree it can be removed along with the associated output manager handle.

base_layer/wallet_ffi/src/lib.rs
hansieodendaal: I would argue that we might keep this method. It is very handy for a wallet to be able to detect with a high degree of certainty which UTXOs it owns. That being said, a strategy of detect-ownership-then-request-all-information may very well be completely implemented in the base layer wallet and never touch the front-end.

W.r.t. "_arbitrarily decrypt some data_"; I think it is equally true or equally difficult for any client and not only clients using the FFI interface.

hansieodendaal (approved):
LGTM - all ffi integration tests and all unit tests passed locally

hansieodendaal (dismissed):
LGTM

PR #4234, opened 2022-06-27
feat: add sender to instructions

stringhandler:
Adds sender to DAN instructions

NOTE: I also created new schema.rs and config file for generating the global db with diesel.  See dan_layer/storage_sqlite/README.md


dan_layer/core/src/models/instruction.rs
```diff
@@ -78,6 +81,10 @@ impl Instruction {
         &self.args
     }
 
+    pub fn sender(&self) -> PublicKey {
```
sdbondi, line 84: nit: this naming convention usually means a reference to the item is returned.
Suggest either `to_sender_public_key` or return a reference.

sdbondi (approved):
LGTM (utAck)

brianp (approved):
Reviewed and looks fine. ~I tried running the migration changes but diesel_cli is refusing to compile so going with ut for now.~

Just had to say it wouldn't work and then it worked. 👌🏻

dan_layer/core/src/models/instruction.rs
stringhandler: Ah yeah, I thought it was `Copy` for some reason....

SWvheerden (approved):

PR #4239, opened 2022-06-28
fix(core)!: include issuer public key in contract id hash

sdbondi:
Description
---
- includes issuer public key ~~and commitment~~ in contract hash
- update command "wizard" code to fetch contract id from resulting transaction 
- simplify FixedString code
- removes unused Hashable trait impls

Motivation and Context
---
~~The contract ID should be a hash of all contract definition data. Adding the commitment allows contracts with the same name label to exist on the blockchain at the same time.~~
The contract name functions as a label and does not have to be globally unique. This PR essentially makes the
<issuer_pk, contract_name> pair unique. 

How Has This Been Tested?
---
Existing tests, manually

### DIFF
--- a/applications/tari_app_grpc/src/conversions/sidechain_features.rs
+++ b/applications/tari_app_grpc/src/conversions/sidechain_features.rs
@@ -24,7 +24,7 @@ use std::convert::{TryFrom, TryInto};
 
 use tari_common_types::types::{FixedHash, PublicKey, Signature};
 use tari_core::transactions::transaction_components::{
-    vec_into_fixed_string,
+    bytes_into_fixed_string,
     CheckpointParameters,
     CommitteeMembers,
     CommitteeSignatures,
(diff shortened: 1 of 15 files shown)


SWvheerden (commented):
I have my questions in allowing duplicate contracts. 
If this is all the same: 
```
contract_name
contract_issuer
contract_spec
```
Is it not essentially the same asset, just with a different owner? Do we want to allow this? I am thinking no, why do we want to allow copes of an asset to existing?  If you change the name or the spec, it's going to be unique? 

Another question I have is, is the contract def UTXO always going to stay static and never change? If it is, it kinda becomes hard to validate contract_id

sdbondi:
@SWvheerden There are a few ways to do this, but I dont think we want to tie the contract_id just to the contract_name as that functions as more of a label. This PR is more about tying the contract to a specific UTXO. i.e. you publish a contract utxo and you get a new unique contract_id on-chain. Also removing the need for the issuer to generate a new public key if a contract with the same "label" (contract_name) is used. However, the issuer generating a new key is not unreasonable.

I think I will change the PR to only include the definition UTXO data for now as that is not controversial.

stringhandler (approved):

PR #4240, opened 2022-06-28
refactor(base_layer/core): remove duplicate setting of recovery byte

sdbondi:
Description
---
Removes commitment calculation and setting of recovery byte in output manager. 

Motivation and Context
---
Noticed that recovery byte was needlessly set in the output manager. Recovery byte is finalized before the transaction is sent. 

How Has This Been Tested?
---
Existing tests


hansieodendaal (approved):
LGTM

stringhandler (approved):

PR #4241, opened 2022-06-29
feat(vn): record contract states

brianp:
Description
---
Add a contracts table to the global vn db so we can store the state of contracts the vn is a member of.

Motivation and Context
---
It will make it easier to go back and check known contracts that are awaiting quorum acceptance.

How Has This Been Tested?
---
Manually, and by fixing the previously broken spec.


sdbondi (dismissed):
Looks good (utAck)

stringhandler (approved):

PR #4245, opened 2022-06-30
fix: add cpp compat to wallet.h

stringhandler:
Changed the build.rs for cbindgen to add the following:

```
#ifdef __cplusplus
extern "C" {

#endif

#ifdef __cplusplus
}

#endif

```

Also renamed the file to `wallet.h`


sdbondi (dismissed):

sdbondi:
@stringhandler I think you need to update the file name in the mobile_build.sh script

PR #4260, opened 2022-07-04
feat: add range proof batch verification to validators

hansieodendaal:
Description
---
Added batch verification of range proofs where ever more than one range proof needed to be verified.

Motivation and Context
---
Batch verification range proofs is much faster than running linear proofs

How Has This Been Tested?
---
Unit tests
Integration tests
_System-level tests pending_

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/mod.rs
+++ b/base_layer/core/src/transactions/transaction_components/mod.rs
@@ -70,7 +70,7 @@ mod transaction_input;
 mod transaction_input_version;
 mod transaction_kernel;
 mod transaction_kernel_version;
-mod transaction_output;
+pub mod transaction_output;
 mod transaction_output_version;
 mod unblinded_output;
 mod unblinded_output_builder;
--- a/base_layer/core/src/transactions/transaction_components/transaction_output.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_output.rs
@@ -482,13 +482,14 @@ impl ConsensusDecoding for TransactionOutput {
     }
 }
 
+/// Performs batched range proof verification for an arbitrary number of outputs. Batched range proof verification gains
+/// above batch sizes of 2^8 = 256 gives diminishing returns, see <https://github.com/tari-project/bulletproofs-plus>,
+/// so the batch sizes are limited to 2^8.
 pub fn batch_verify_range_proofs(
     prover: &RangeProofService,
     outputs: &[&TransactionOutput],
 ) -> Result<(), RangeProofError> {
-    // Batched range proof verification gains above batch sizes of 256 gives diminishing returns,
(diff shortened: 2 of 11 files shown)


base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -476,3 +481,92 @@ impl ConsensusDecoding for TransactionOutput {
...
+                }],
+            });
+            proofs.push(output.proof.to_vec().clone());
+        }
+        index += power_of_two;
+        prover.verify_batch(proofs.iter().collect(), statements.iter().collect())?;
+    }
+    Ok(())
+}
+
```
SWvheerden, line 516: This is a pub function. Does this need to be public?
Either way, this function can benefit from docs. To understand what this function does, you need to read the comment on the function `batch_verify_range_proofs` and the tests for this function, this is not ideal.

SWvheerden (commented):
This looks good. 
I would just prefer adding docs to the two new functions. 
Especially the `power_of_two_chunk_sizes` one.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
hansieodendaal: Good point, fixed

base_layer/core/src/transactions/transaction_components/mod.rs
```diff
@@ -70,7 +70,7 @@ mod transaction_input;
 mod transaction_input_version;
 mod transaction_kernel;
 mod transaction_kernel_version;
-mod transaction_output;
+pub mod transaction_output;
```
stringhandler, line 73: rather use `pub use` to expose what you want to, otherwise all of these will eventually be `pub`

therustmonk (approved):

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -476,3 +481,98 @@ impl ConsensusDecoding for TransactionOutput {
...
+                    minimum_value_promise: 0,
+                }],
+            });
+            proofs.push(output.proof.to_vec().clone());
+        }
+        index += power_of_two;
+        prover.verify_batch(proofs.iter().collect(), statements.iter().collect())?;
+    }
+    Ok(())
+}
```
CjS77, line 515: This function should be in the bulletproofs+ crate. Think about it - Verifying batches of non power-of-two sets is something you'd be writing over and over if its not there.

hansieodendaal: Yes it could move there and be totally transparent.

hansieodendaal: I created an issue for it

stringhandler (approved):

PR #4261, opened 2022-07-04
feat(dan_layer)!: generate and add checkpoint signatures

sdbondi:
Description
---
- feat(dan_layer): generate and add checkpoint signatures
- feat(explorer): update text explorer to include signatures
-  feat(base_layer/wallet): update wallet command file formats
-  feat(base_layer/core): wires up committee signatures to grpc with conversions
-  feat(base_layer/core): adds SignerSignature struct for CommitteeSignatures
-  feat(base_layer/wallet): include committee signatures for checkpoint GRPC requests

Motivation and Context
---
DAN committees need to collect signatures to include in checkpoints. This PR adds the initial implementation of that.
The signature challenge is: 

`e = H_1(signer_public_key || public_nonce || H_2(contract_id||commitment||merkle_root||checkpoint_number))`

Although constructing a challenge with actual data is a TODO as that involves a bit of a refactor.

This PR also adds the `SignerSignature` struct that includes the signing public key along with the Schnorr signature tuple `<R, s>`. This is so that we know which if the validators signed. There may be an aggregate signature scheme that allows this.

How Has This Been Tested?
---
Manually, initial checkpoint contains signature.

### DIFF
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -313,7 +313,12 @@ message ContractAmendment {
 }
 
 message CommitteeSignatures {
-    repeated Signature signatures = 1;
+    repeated SignerSignature signatures = 1;
+}
+
+message SignerSignature {
+    bytes signer = 1;
+    Signature signature = 2;
 }
 
 // TODO: DEPRECATED
--- a/applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
+++ b/applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
@@ -117,12 +117,11 @@ impl WalletClient {
     merkle_root: Vec<u8>,
   ) -> Result<grpc::CreateInitialAssetCheckpointResponse, CollectiblesError> {
     let inner = self.get_inner_mut()?;
-    let committee = vec![];
     let request = grpc::CreateInitialAssetCheckpointRequest {
       // TODO: contract id
       contract_id: Vec::from_hex(asset_public_key)?,
       merkle_root,
-      committee,
+      committee_signatures: None,
     };
     let result = inner
       .create_initial_asset_checkpoint(request)
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -271,8 +271,7 @@ message RegisterAssetResponse {
 message CreateInitialAssetCheckpointRequest {
     bytes contract_id = 1;
     bytes merkle_root = 2;
-    repeated bytes committee = 3;
-    CommitteeSignatures committee_signatures = 4;
+    CommitteeSignatures committee_signatures = 3;
 }
 
 message CreateInitialAssetCheckpointResponse {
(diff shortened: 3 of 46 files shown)


applications/tari_app_grpc/proto/types.proto
```diff
@@ -313,7 +313,12 @@ message ContractAmendment {
 }
 
 message CommitteeSignatures {
-    repeated Signature signatures = 1;
+    repeated SignerSignature signatures = 1;
+}
+
+message SignerSignature {
+    bytes signer = 1;
```
SWvheerden, line 320: nit:
```suggestion
    bytes signer_pubkey = 1;
```

applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
```diff
@@ -117,12 +117,11 @@ impl WalletClient {
     merkle_root: Vec<u8>,
   ) -> Result<grpc::CreateInitialAssetCheckpointResponse, CollectiblesError> {
     let inner = self.get_inner_mut()?;
-    let committee = vec![];
     let request = grpc::CreateInitialAssetCheckpointRequest {
       // TODO: contract id
       contract_id: Vec::from_hex(asset_public_key)?,
       merkle_root,
-      committee,
+      committee_signatures: None,
```
SWvheerden, line 124: I take it this is still test/initial code?
Surely the initial checkpoint needs to be signed by the committee?

SWvheerden (commented):
This looks good. 
 have two questions, 
 including one on the challenge. 
 Why make the challenge hash (xx||hash(xx))?
 Surely we don't need the double hash, we can just concatenate all the data into a single hash once?

stringhandler (dismissed):
utACK.

stringhandler:
@SWvheerden I don't see a problem with the double hash. The "true" challenge is hashed separately and then the Nonce and Pub Keys are committed to in the signature to prevent reuse.

SWvheerden:
> I don't see a problem with the double hash. The "true" challenge is hashed separately and then the Nonce and Pub Keys are committed to in the signature to prevent reuse.

Dont get me wrong, I think it's perfectly fine as is, was wondering why double hash it, a single hash is supposed to be perfectly random.

sdbondi:
@SWvheerden basically it encapsulates that a SignerSignature must always commit to the public key and nonce, saves the caller from generating a nonce and potentially forgetting to include it in the challenge. The inner challenge will be domain separated.

SWvheerden:
> @SWvheerden basically it encapsulates that a SignerSignature must always commit to the public key and nonce, saves the caller from generating a nonce and potentially forgetting to include it in the challenge. The inner challenge will be domain separated.

Yeah signatures always need to commit to the pubkey and nonce, multisigs need to commit to the hash of the nonce, but that's not required here as its a single signer. 
So the whole reason for having a part inner hashed as well, is to have it be domain separated?

sdbondi:
> @SWvheerden So the whole reason for having a part inner hashed as well, is to have it be domain separated?

Yeah more or less.
From the perspective of the SignerSignature, you don't need to provide a hash but just an arbitrary length byte slice. 
We choose to hash first so that it is domain-separated. I don't see any reason not to do that.

applications/tari_collectibles/src-tauri/src/clients/wallet_client.rs
sdbondi: yeah this is in the collectibles app, which is a little in limbo at the moment

stringhandler (approved):

PR #4265, opened 2022-07-05
fix(vn): scan and save contracts without autoaccept

brianp:
Description
---
The VN was only scanning and saving contract if autoaccept was enabled. It should still scan and save, just not accept the contracts without autoaccept.

Motivation and Context
---
Noticed the wrong behaviour while working alongside the code.

How Has This Been Tested?
---
Integration tests still pass


SWvheerden (approved):
LGTM

stringhandler (approved):

sdbondi (approved):

PR #4267, opened 2022-07-06
feat: add mined_timestamp to wallet.db

Cifko:
Description
---
Add mined_timestamp to wallet.db

### DIFF
--- a/base_layer/wallet/migrations/2022-07-04-150000_add_mined_timestamp/up.sql
+++ b/base_layer/wallet/migrations/2022-07-04-150000_add_mined_timestamp/up.sql
@@ -1,2 +1,5 @@
 ALTER TABLE completed_transactions
     ADD mined_timestamp DATETIME NULL;
+
+ALTER TABLE outputs
+    ADD mined_timestamp DATETIME NULL;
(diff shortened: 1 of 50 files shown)


base_layer/wallet/migrations/2022-07-04-150000_add_mined_timestamp/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE completed_transactions
```
stringhandler, line 1: We should add it to the `outputs` table as well

base_layer/wallet/src/transaction_service/tasks/check_faux_transaction_status.rs
```diff
@@ -136,6 +136,7 @@ pub async fn check_faux_transactions<TBackend: 'static + TransactionBackend>(
                     tx.tx_id,
                     mined_height,
                     mined_in_block,
+                    0,
```
stringhandler, line 139: This should have a value, it's not only for faucet utxos

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -456,6 +460,7 @@ where
                 tx_id,
                 mined_height,
                 mined_in_block.clone(),
+                0,
```
stringhandler: Rather put the value that matches `mined_in_block` here for consistency

stringhandler (approved):
utACK. 

Seems correct from review

PR #4270, opened 2022-07-06
fix: fix config toml files

hansieodendaal:
Description
---
- Cleaned up config toml files
- Added most missing options
- Added documentation
- Removed redundant application config struct options

Motivation and Context
---
It was out of date and not accurate anymore

How Has This Been Tested?
---
- Cucumber integration tests made to work
- Performed system level testing for the base node and console wallet where each config setting was uncommented and the applications verified to start properly using the config.


common/config/presets/base_node.toml
```diff
@@ -37,59 +148,183 @@ identity_file = "config/base_node_id_dibbler.json"
 # - a "bridge" between TOR and TCP-only nodes
 # auxiliary_tcp_listener_address = "/ip4/127.0.0.1/tcp/9998"
 
+# Path to the LMDB data files
+#datastore_path = "peer_db"
```
stringhandler: This is not correct. This would fall under the `base_node.p2p` section, when I believe you are trying to set it under the `base_node` section

stringhandler (commented):
I'm not sure if the settings are in the correct place. 

I would suggest uncommenting all of the settings first to make sure there are no errors, then comment them again before checking in.

hansieodendaal:
> I'm not sure if the settings are in the correct place.
> 
> I would suggest uncommenting all of the settings first to make sure there are no errors, then comment them again before checking in.

Yes, exactly how I tested them, waded through them

common/config/presets/base_node.toml
hansieodendaal: This is for `P2pConfig::datastore_path` with default `datastore_path: PathBuf::from("peer_db")`, which goes under `base_node.p2p`

sdbondi (commented):
Thanks for adding field comments and the cleanup - LGTM

stringhandler (approved):

PR #4272, opened 2022-07-06
feat(vn): recognize abandoned state

brianp:
Description
---
This is a subtask for quarantining a contract. It lets the VN's monitoring the contract to flag it as abandoned (locally only) when checkpoints have been missed.

Motivation and Context
---
VN's need to be more aware of the status of contracts this is an initital step before quarantine happens.

How Has This Been Tested?
---

- [x] Tested manually
- [ ] ~Test written~ To much work to get contract state from the vn for a single spec


sdbondi (approved):
Nice LGTM (utAck)

Issue #4176, opened 2022-06-08
Hardcode a more recent zero point for wallet birthday


AaronFeickert:
A wallet's [birthday](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L61) is encoded as the number of days since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time), with a 2-byte capacity. This will hit its maximum value around the year 2060; however, this is wasteful, since no mainnet blocks have been mined between the epoch and now.

The following are recommended:
- Set a more recent zero point for the wallet birthday (the date of the eventual genesis block has been suggested)
- Hardcode this zero point as part of the [version](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L60) definition
- If (when?) the birthday field reaches its maximum value, you can define a new seed version and hardcode a new zero point

PR #4275, opened 2022-07-06
fix: add more recent zero point for wallet birthday (see issue #4176)

jorgeantonio21:
Description
--- Add logic to tackle issue #4176, which aims at improving birthday representation for Cipher Seed generation.

Motivation and Context
--- The goal of this PR is to resolve issue #4176.  The current version is supposed to allow a more recent genesis for wallet birthdays (it actually depends on the genesis block of the Tari network, we wish to work with). Moreover, it updates the version. Moreover, if the birthday does not fit within the `u16` type range, we update to a new version. 

How Has This Been Tested?
--- Add simple tests to check if logic is correct.


jorgeantonio21:
We introduce a `Birthday` type which encapsulates the logic to deal with a zero point in time, and updates this zero point in time 
 every `2^16` days. This is needed in order to have a well defined `birthday` for perpetuity. 

The current implementation relies on the genesis block of a chosen network. At the moment, we default to the `Dibbler` test-net, however this should be updated when the main-net is released. 

One can argue that this functionality is not at all required, and we should instead hardcode an initial zero point in time, and update it to the mainnet, at launch time.

stringhandler (approved):

PR #4289, opened 2022-07-08
feat: add quarantine and backup keys

SWvheerden:
Description
---
This PR adds the quarantine interval and backup_keys to the constitution as per the RFC. 

This also moves the side_chain features to boxed struct.

Motivation and Context
---
Adding the quarantine interval and backup_keys makes the output_features big enough that it causes a stack overflow error in some of the wallet tests. Since we will most likely add more features to the side_chain features, this is moved to a boxed struct so that it wont fail. 

How Has This Been Tested?
---
Unit tests


brianp (approved):
Tested locally that I could still publish constitutions etc. Everything worked. Code looks fine.

In the event we end up going a different direction how soon do you think we would have needed to wrap SideChainFeatures in a Box anyway?

Also worth noting I ran the failing spec locally and had no issues so that looks like it's just a CI thing.

base_layer/core/src/covenants/fields.rs
```diff
@@ -86,7 +86,7 @@ impl OutputField {
         self as u8
     }
 
-    pub(super) fn get_field_value_ref<T: 'static>(self, output: &TransactionOutput) -> Option<&T> {
+    pub(super) fn get_field_value_ref<T: 'static + std::fmt::Debug>(self, output: &TransactionOutput) -> Option<&T> {
```
sdbondi, line 89: Any reason for this debug trait bound?

base_layer/core/src/covenants/fields.rs
```diff
@@ -105,7 +105,7 @@ impl OutputField {
                 // sidechain_features because there is no reference to Option<FixedHash> for
                 // sidechain_features.contract_id to return.
                 match &output.features.sidechain_features {
-                    Some(SideChainFeatures { ref contract_id, .. }) => contract_id as &dyn Any,
+                    Some(v) => &v.contract_id as &dyn Any,
```
sdbondi, line 108: You could keep the pattern matching, and dereference the box in the match:
`match &*output.features.sidechain_features` 

What you have is fine though, unless it is the reason that you added the debug trait bound above?

base_layer/core/src/proto/transaction.proto
```diff
@@ -156,6 +157,9 @@ message RequirementsForConstitutionChange {
     // An allowlist of keys that are able to accept and ratify the initial constitution and its amendments. If this is
     // None, the constitution cannot be amended.
     CommitteeMembers constitution_committee = 2;
+    // An allowlist of keys that can be used in case of checkpoint abandonment. These keys can quarantine the the constitution
```
sdbondi, line 160:
```suggestion
    // An allowlist of keys that can be used in case of checkpoint abandonment. These keys can quarantine the constitution
```

base_layer/wallet/tests/transaction_service_tests/service.rs
```diff
@@ -1086,6 +1086,7 @@ async fn send_one_sided_transaction_to_self() {
 
 #[tokio::test]
 async fn manage_multiple_transactions() {
+    env_logger::init();
```
sdbondi, line 1089:
```suggestion
```

sdbondi (commented):
LGTM - very minor comments

Agree with the boxing of sidechain features, as it contributes to a significant portion of the static size of OutputFeatures.

base_layer/core/src/covenants/fields.rs
SWvheerden: that was for debugging, this should have been removed, missed it.

base_layer/core/src/covenants/fields.rs
SWvheerden: You cant deref Option<Box<T>>
Or at least the compiler does not want to allow me to do it.

SWvheerden:
```

error[E0614]: type `std::option::Option<Box<sidechain_features::SideChainFeatures>>` cannot be dereferenced
   --> base_layer/core/src/covenants/fields.rs:108:24
    |
108 |                 match &*output.features.sidechain_features {
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

sdbondi: oh yeah, output.features.sidechain_features.as_deref() is what you want

But just wanted to check that it's not somehow causing the Debug trait bound - what you have is 💯

sdbondi (approved):

sdbondi:
@SWvheerden `check_share` test is failing - please could you look into why that is?

PR #4292, opened 2022-07-08
chore: add stealth address to wallet

Cifko:
Description
---
Add a stealth address scanning and spending keys. Currently they are equal to wallet key. But that can change in the future.
This addresses will be used in sending and scaning. So if we change it future, we have to rewrite just one line of code and some UI. But no the whole scanning and sending.
The know stealth addresses table was also added. The table consist of hash of the stealth address (hash of `hex(scanning)::hex(spending)` key string. That hash is used as a primary key to the table.

### DIFF
--- a/comms/core/src/peer_manager/node_identity.rs
+++ b/comms/core/src/peer_manager/node_identity.rs
@@ -49,6 +49,8 @@ pub struct NodeIdentity {
     public_key: CommsPublicKey,
     features: PeerFeatures,
     secret_key: CommsSecretKey,
+    stealth_address_scanning_private_key: CommsSecretKey,
+    stealth_address_spending_private_key: CommsSecretKey,
     public_address: RwLock<Multiaddr>,
     #[serde(default = "rwlock_none")]
     identity_signature: RwLock<Option<IdentitySignature>>,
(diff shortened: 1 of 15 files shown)


comms/core/src/peer_manager/node_identity.rs
```diff
@@ -47,8 +47,12 @@ pub struct NodeIdentity {
     #[serde(deserialize_with = "deserialize_node_id_from_hex")]
     node_id: NodeId,
     public_key: CommsPublicKey,
+    stealth_address_scanning_public_key: CommsPublicKey,
+    stealth_address_spending_public_key: CommsPublicKey,
     features: PeerFeatures,
     secret_key: CommsSecretKey,
+    stealth_address_scanning_secret_key: CommsSecretKey,
+    stealth_address_spending_secret_key: CommsSecretKey,
```
SWvheerden: Why are we storing both the public and private keys?
Surely we only need to store the private ones?

SWvheerden (commented):
I just have a question about the public keys, surely we can just calculate them when needed?

comms/core/src/peer_manager/node_identity.rs
Cifko: Initially I was using the same patter as the surrounding code for public and secret key for the wallet. But you are right, and I deleted the stealth public keys everywhere. Only private are used now.

Cifko:
I've added known stealth addresses sql table to the PR.

comms/core/src/peer_manager/node_identity.rs
```diff
@@ -49,6 +49,8 @@ pub struct NodeIdentity {
     public_key: CommsPublicKey,
     features: PeerFeatures,
     secret_key: CommsSecretKey,
+    stealth_address_scanning_private_key: CommsSecretKey,
```
sdbondi, line 52: `NodeIdentity` is used for all network participants (base nodes, validator nodes etc) and the `comms` crate is for general purpose p2p networking so putting stealth address keys in the NodeIdentity is not appropriate.

sdbondi (commented):
Stealth address keys should be in a separate struct in the wallet and not bundled into the NodeIdentity as that applies to all network participants e.g. a base node does not need stealth keys. 
 
If you want to persist the most recent stealth keys use the `wallet_settings` table to create a `StealthKeys` struct and pass that in.

comms/core/src/peer_manager/node_identity.rs
Cifko: Yes. I'll close this PR, since this actually super complicates things. I will have to think this over.

PR #4294, opened 2022-07-08
fix(validator-node): only submit checkpoint if the leader

sdbondi:
Description
---
- changes consensus manager to only submit checkpoints if the leader
- emojifies some stdout logs
- minor fixes to log4rs sample

Motivation and Context
---
Only one checkpoint should be submitted, so to make things simpler we use the leader. 

How Has This Been Tested?
---
Manually


SWvheerden (approved):
LGTM

stringhandler (approved):

Issue #4182, opened 2022-06-11
`CipherSeed` MAC is keyed directly with low-entropy passphrase


AaronFeickert:
The MAC used in the `CipherSeed` construction is [keyed](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L214) (via `Blake2b` input concatenation) with the user-supplied passphrase, which should be assumed to be of low entropy.

A straightforward improvement is to extend the associated `Argon2` output and split it to generate two keys: one for the `ChaCha20` encryption, and the other for the MAC. While the MAC is very short (only 5 bytes in the current design), this approach can take some advantage of any key stretching provided by `Argon2`, in addition to preventing an attacker from performing any kind of precomputation on the MAC prior to deriving the key.

Additionally, the use of `Blake2b` for this MAC construction should use proper domain separation. Any future hashing API should be strongly considered for this.

hansieodendaal:
Wow, the pitfalls...

jorgeantonio21:
I tried to tackle this issue, the PR can be found [here](https://github.com/tari-project/tari/pull/4296).

AaronFeickert:
Another recommendation that came up recently in a discussion with @jorgeantonio21 was to use separate hash operations for turning the [primary salt](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L116) into the `Argon2` [salt](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L245) and `ChaCha20` [nonce](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L241). This follows the best practice of not reusing key- or nonce-type material across different cryptographic primitives. The primary salt should continue to be used directly when [computing the MAC](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L140), as there it serves as something akin to authenticated associated data.

PR #4296, opened 2022-07-09
fix: low entropy mac passphrase in cipher seed (see issue #4182) 

jorgeantonio21:
Description
---The following is an attempt to improve `CipherSeed` mnemonic generation by tackling MAC being keyed directly with a low entropy passphrase. We use proper domain separation to attain this.

Motivation and Context
--- The generation of MAC, within the context of a `CipherSeed` instance, is obtained through keying a (low) entropy passphrase. In order to reduce the chances of success of an attack involving offline key pre-computation, it is desirable to hash the passphrase, before MAC keying, using proper domain separation. The current PR is an attempt in this direction.

How Has This Been Tested?
--- With previous unit tests.


jorgeantonio21:
Currently, we use a `Blake256` to generate domain separation. It might be desirable to use some other hashing algorithm, for purposes I don't understand.

Also feel free to comment below, as many of these concepts are not completely understood by the creator of the current PR.

applications/tari_validator_node/src/grpc/services/wallet_client.rs
```diff
@@ -72,11 +72,11 @@ impl WalletClient for GrpcWalletClient {
         contract_id: &FixedHash,
         state_root: &StateRoot,
         checkpoint_number: u64,
-        checkpoint_signatures: &[SignerSignature],
+        checkpoint_signatures: Vec<SignerSignature>,
```
SWvheerden: Why change this to a vec?
We are not changing this in the function?

dan_layer/core/src/services/checkpoint_manager.rs
```diff
@@ -35,7 +35,7 @@ pub trait CheckpointManager {
         &mut self,
         checkpoint_number: u64,
         state_root: StateRoot,
-        signature: &[SignerSignature],
```
SWvheerden: same here, why change to a vec?

dan_layer/core/src/services/checkpoint_manager.rs
```diff
@@ -60,7 +60,7 @@ impl<TWallet: WalletClient + Sync + Send> CheckpointManager for ConcreteCheckpoi
         &mut self,
         checkpoint_number: u64,
         state_root: StateRoot,
-        signatures: &[SignerSignature],
```
SWvheerden: and here?

SWvheerden (commented):
Why did you change a bunch of &[T] to Vec<T>?

hansieodendaal:
Hi @jorgeantonio21, maybe you can squash all commits and rebase? There seems to be many unrelated changes in the code changes.

applications/tari_validator_node/src/grpc/services/wallet_client.rs
jorgeantonio21: I am sorry @SWvheerden, this was due to a merge conflict that I missed.

dan_layer/core/src/services/checkpoint_manager.rs
jorgeantonio21: Same merge conflict

dan_layer/core/src/services/checkpoint_manager.rs
jorgeantonio21: Same merge conflict

jorgeantonio21:
> Hi @jorgeantonio21, maybe you can squash all commits and rebase? There seems to be many unrelated changes in the code changes.

Hi @hansieodendaal, thanks for the advice. I did rebase and the merge conflicts seem to be tackled.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -255,6 +273,18 @@ impl CipherSeed {
         Ok(())
     }
 
+    fn apply_domain_separation(passphrase: &[u8], salt: &[u8], persona: &str) -> Result<Vec<u8>, KeyManagerError> {
+        if salt.len() > 16 || persona.len() > 16 {
+            return Err(KeyManagerError::DecryptionFailed);
+        }
```
hansieodendaal: This limitation is not required anymore with the use of the hashing API

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -34,14 +34,20 @@ use chacha20::{
...
 use tari_utilities::ByteArray;
 
 use crate::{
+    birthday::Birthday,
     error::KeyManagerError,
     mnemonic::{from_bytes, to_bytes, to_bytes_with_language, Mnemonic, MnemonicLanguage},
 };
 
 const CIPHER_SEED_VERSION: u8 = 0u8;
+const CIPHER_SEED_PERSONA: &str = "CIPHER_PERSONA";
```
hansieodendaal:
```suggestion
const CIPHER_SEED_PERSONA: &str = "com.tari.key_manager.cipher_seed";
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -95,10 +101,8 @@ pub struct CipherSeed {
 impl CipherSeed {
     #[cfg(not(target_arch = "wasm32"))]
     pub fn new() -> Self {
-        const SECONDS_PER_DAY: u64 = 24 * 60 * 60;
-        let days = u64::try_from(chrono::Utc::now().timestamp()).unwrap() / SECONDS_PER_DAY;
-        let birthday = u16::try_from(days).unwrap_or(0u16);
-        CipherSeed::new_with_birthday(birthday)
+        let birthday_data = Birthday::default();
```
CjS77: Why not just use `Birthday` instead of tearing it apart?

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -110,14 +114,17 @@ impl CipherSeed {
...
-    fn new_with_birthday(birthday: u16) -> Self {
+    fn new_with_birthday(birthday_data: Birthday) -> Self {
         let mut entropy = [0u8; CIPHER_SEED_ENTROPY_BYTES];
         OsRng.fill_bytes(&mut entropy);
         let mut salt = [0u8; CIPHER_SEED_SALT_BYTES];
         OsRng.fill_bytes(&mut salt);
 
+        let version = birthday_data.version();
+        let birthday = birthday_data.birthday();
+
```
CjS77: Instead of all this, just store a birthday instance

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use serde::{Deserialize, Serialize};
+use tari_common::configuration::Network;
+use tari_core::blocks::genesis_block::get_genesis_block;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Birthday {
+    zero_point_time: u64,
+    birthday: u16,
+    version: u8,
```
CjS77: The way `version` is calculated seems very weird. Why is the version dependent on the current time??

There are no docs to explain this.

edit: Ok, it's just named in a very misleading way with no docs. See comments below

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl Birthday {
+    pub fn default() -> Self {
+        let network = Network::Dibbler;
+        Self::new(network)
+    }
```
CjS77: impl Default!

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    pub fn new(network: Network) -> Self {
+        let current_time = Self::current_time_in_seconds();
+        Self::new_from_current_time(network, current_time)
+    }
+
+    fn new_from_current_time(network: Network, current_time: u64) -> Self {
```
CjS77:
```suggestion
    fn new_from_time(network: Network, time: u64) -> Self {
```

if it was the current time you were using, you wouldn't need to pass the time in

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+        const SECONDS_PER_DAY: u64 = 24 * 60 * 60;
+        const PERIOD_LENGTH: u64 = u16::MAX as u64 + 1; // 2^16
+
+        let mut zero_point_time = Self::get_network_genesis_time(network);
+
+        let days = (current_time - zero_point_time) / SECONDS_PER_DAY;
```
CjS77: It would. be more idiomatic to use `Duration` and `Instant`

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+        let birthday = u16::try_from(days % PERIOD_LENGTH).unwrap();
+        let version = u8::try_from(days / PERIOD_LENGTH).unwrap();
```
CjS77: ????

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    fn new_from_current_time(network: Network, current_time: u64) -> Self {
+        const SECONDS_PER_DAY: u64 = 24 * 60 * 60;
+        const PERIOD_LENGTH: u64 = u16::MAX as u64 + 1; // 2^16
+
+        let mut zero_point_time = Self::get_network_genesis_time(network);
+
+        let days = (current_time - zero_point_time) / SECONDS_PER_DAY;
+        let birthday = u16::try_from(days % PERIOD_LENGTH).unwrap();
```
CjS77: Is there a reason it's not `(days % PERIOD_LENGTH) as u16`?

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+        let version = u8::try_from(days / PERIOD_LENGTH).unwrap();
+
+        zero_point_time += PERIOD_LENGTH * u64::from(version);
```
CjS77: I prefer `PERIOD_LENGTH * (version as u64)` but simply stylistic.

base_layer/key_manager/src/birthday.rs
CjS77: Ohhh! This is an epoch counter. 

Please rename it some other than version (epoch / period for example), and provide a doc string explaining what it is for.

base_layer/key_manager/src/birthday.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.zero_point_time
+    }
+
+    pub fn current_time_in_seconds() -> u64 {
+        u64::try_from(chrono::Utc::now().timestamp()).unwrap()
+    }
+
+    pub fn get_network_genesis_time(network: Network) -> u64 {
+        get_genesis_block(network).block().header.timestamp.as_u64()
+    }
```
CjS77: I think it would be better to not have the dependency on `tari_core::blocks::genesis_block` at all. 
Better to pass in the genesis time as a parameter to the constructor. Then this module is more independent / decoupled. Then you ALSO can drop the dependence on `Network` it looks like. Bonus!

base_layer/key_manager/src/birthday.rs
CjS77: And then wouldn't it be simpler to make the birthday a u32/u64 than manually reconstruct what is effectively a u24?

CjS77 (commented):
Reviewed Birthday module primarily

base_layer/key_manager/src/birthday.rs
jorgeantonio21: That is definitely a possibility. However, I don't feel confident that the birthday genesis time might depend on the API use. Since we are primarily interested in using `Birthday` to set up a timestamp to scan the blockchain, it might be good to have a consistent genesis time for birthday. I am possibly missing something in your commentary.

Moreover, following [issue](https://github.com/tari-project/tari/issues/4176), it is referred that one should have an hard-coded birthday genesis time, and it is suggested to use the timestamp of the genesis network. I also really want to remove the dependency on the `Network`, as well. One could just hard-code a recent date (like now) and set it for the future use of the `Birthday` type.

base_layer/key_manager/src/birthday.rs
jorgeantonio21: Definitely so ! I am going to change it :)

base_layer/key_manager/src/birthday.rs
jorgeantonio21: Much more readable this way, I am going to change it.

base_layer/key_manager/src/birthday.rs
jorgeantonio21: I am going to add docs. When I implemented the code, I had in mind the third suggestion described [here](https://github.com/tari-project/tari/issues/4176). 

From what I understood, the `version` field in `Birthday` should really work as an epoch counter. This would be used later on to index the `version` in `CipherSeed`. Following this approach, versioning is epoch-dependent.

Do you think this approach makes sense  ?

base_layer/key_manager/src/birthday.rs
jorgeantonio21: Done

base_layer/key_manager/src/birthday.rs
jorgeantonio21: I added a more detailed comment below.

base_layer/key_manager/src/birthday.rs
jorgeantonio21: I agree. I only implemented this way for testing purposes. I am going to refactor it conveniently.

base_layer/key_manager/src/birthday.rs
jorgeantonio21: If one decides to keep the versioning for `CipherSeed` to be dependent on the epoch. Another approach would be to migrate the version computation to the `CipherSeed` impl, thus make it more explicit the use case we have in mind.

hansieodendaal (commented):
LGTM, just the failing "test key manager wasm" tests left to fix

base_layer/key_manager/src/birthday.rs
AaronFeickert: The intent of the overall `version` field should be an unambiguous identification of all parts of the protocol that all software agrees on. So if you were to change `Argon2` parameters or migrate to a different MAC algorithm, you'd define a new version; any software that knows about it will do the right thing, and any software that doesn't will simply produce an error.

I don't, however, think it's a good idea to attempt to compute a dynamic version number from the birthday. The intent of the comment in [this issue](https://github.com/tari-project/tari/issues/4176) was that once enough time passes such that a given birthday could potentially overflow, then you (or whoever is doing development that far into the future!) would manually define a new version that accounts for this. Doing it dynamically feels like a risky approach that could cause conflicts with any non-birthday protocol changes requiring new versions.

base_layer/key_manager/src/birthday.rs
AaronFeickert: The idea behind hardcoding all parameters is to ensure that all implementations and installations agree on them unambiguously. Then you don't need to worry about differences in defaults or other computed parameters; everyone is working from the same playbook.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -65,7 +72,15 @@ pub const CIPHER_SEED_MAC_BYTES: usize = 5;
 /// checksum    4 bytes
 ///
 /// In its enciphered form we will use the MAC-the-Encrypt pattern of AE so that the birthday and entropy will be
```
AaronFeickert, line 81: To be clear, MAC-then-encrypt is in general not a safe approach to authenticated encryption. It can be shown to be safe under certain designs, like a stream cipher with a strong MAC, but it isn't clear that the truncated hash function approach here strictly meets this requirement.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -65,7 +72,15 @@ pub const CIPHER_SEED_MAC_BYTES: usize = 5;
...
-/// encrypted. The version and salt are associated data that are included in the MAC but not encrypted.
+/// encrypted. Moreover, instead of generating the MAC directly using the user provided low entropy passphrase, we
+/// shall instead produce a 32-byte hashing of it. In reality, such 32-byte hashing is obtained by applying Argon2,
```
AaronFeickert: It's much more than just a hash of the password, which would not be safe for this purpose. The intent is to use a password-based key derivation function to generate a derived key of higher effective entropy through the use of a carefully-designed function like `Argon2` that's built for this purpose. Then we use separate portions of that derived key for the MAC and for `ChaCha20` encryption.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -130,19 +145,18 @@ impl CipherSeed {
...
-            .expect("Should be able to create blake2 hasher; will only panic if output size is 0 or greater than 64");
-        let mut hmac = [0u8; CIPHER_SEED_MAC_BYTES];
-        blake2_mac_hasher
-            .chain(plaintext.clone())
-            .chain([CIPHER_SEED_VERSION])
-            .chain(self.salt)
-            .chain(passphrase.as_bytes())
-            .finalize_variable(|res| hmac.copy_from_slice(res));
+        // generate the hash MAC
+        let mut hmac = Self::generate_hmac(
```
AaronFeickert: For clarity I would rename this, since [HMAC](https://datatracker.ietf.org/doc/html/rfc2104) has a specific definition that isn't being used here. In this design the MAC is generated from a hash function, but it's not strictly HMAC.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -230,25 +243,29 @@ impl CipherSeed {
...
 
     fn apply_stream_cipher(data: &mut Vec<u8>, passphrase: &str, salt: &[u8]) -> Result<(), KeyManagerError> {
-        let argon2 = Argon2::default();
         let blake2_nonce_hasher: VarBlake2b = VarBlake2b::new(size_of::<Nonce>())
             .expect("Should be able to create blake2 hasher; will only panic if output size is 0 or greater than 64");
 
         let mut encryption_nonce = [0u8; size_of::<Nonce>()];
+
+        // TODO: use hashing API to generate suitable domain separation instead
+        let label = b"chacha20_salt";
```
AaronFeickert: Is this domain separation label too generic? Other separators elsewhere are more detailed to avoid any accidental reuse elsewhere.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -230,25 +243,29 @@ impl CipherSeed {
...
         blake2_nonce_hasher
+            .chain(label)
```
AaronFeickert: A safer approach (and one used by the hashing API) is to ensure that any variable-length input to the underlying hash function be prepended with a fixed-length encoding of the input's length. This avoids accidental or malicious collisions. If you're taking a belt-and-suspenders approach, it's best to do this even for fixed-length inputs, in case they change later.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -230,25 +243,29 @@ impl CipherSeed {
...
-        // will use key stretching to produce a longer nonce for the passphrase hash and the encryption nonce.
-        let salt_b64 = SaltString::b64_encode(&encryption_nonce)?;
+        // we take the last 32 bytes of the generated derived encryption key for Chacha20 cipher, see
+        // documentation
+        let derived_encryption_key: Vec<u8> = Self::generate_domain_separated_passphrase_hash(passphrase, salt)?
+            .into_iter()
+            .rev()
+            .take(32)
+            .rev()
+            .collect();
```
AaronFeickert: Is there a more straightforward and/or more idiomatic way to do this?

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +281,107 @@ impl CipherSeed {
...
+impl CipherSeed {
+    fn generate_hmac(
+        birthday: &[u8],
+        entropy: &[u8],
+        cipher_seed_version: &[u8],
+        salt: &[u8],
+        passphrase: &str,
+    ) -> Result<Vec<u8>, KeyManagerError> {
+        // birthday should be 2 bytes long
+        if birthday.len() != 2 {
```
AaronFeickert: Might be better to avoid the use of a magic number and move this length to a constant, in case it changes in a future version.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +281,107 @@ impl CipherSeed {
...
+            return Err(KeyManagerError::InvalidData);
+        }
+
+        // we take the first 32 bytes of the generated derived encryption key for MAC generation, see documentation
+        let passphrase_key: Vec<u8> = Self::generate_domain_separated_passphrase_hash(passphrase, salt)?
+            .into_iter()
+            .rev()
+            .take(32)
+            .rev()
+            .collect();
```
AaronFeickert: Doesn't this take the _last_ 32 bytes?

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +281,107 @@ impl CipherSeed {
...
+
+        // TODO: use hashing API to generate proper domain separation hash
+        let blake2_mac_hasher: VarBlake2b = VarBlake2b::new(CIPHER_SEED_MAC_BYTES).expect(
+            "Should be able to create blake2 hasher; will only panic if output size is 0 or greater than value",
+        );
+        let mut hmac = [0u8; CIPHER_SEED_MAC_BYTES];
+
+        blake2_mac_hasher
+            .chain(DOMAIN_SEPARATION_LABEL.as_bytes())
```
AaronFeickert: As noted elsewhere, ideally this should prepend the length of the label using a fixed-length encoding.

AaronFeickert (changes requested):

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Thank you for the clarification here, Aaron. I reused part of your explanation in the documentation, I hope it makes more clear the goal of the current refactor.

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: I rename it simply to `mac`.

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Definitely agree. I changed this to, following the current convention, `com.tari.key_manager.cipher_seed.chacha20_enc`. Similarly for the MAC generation label.

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Nice catch here :) !

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: I refactored it to use slicing, it makes the code cleaner and simpler to read. Thanks !

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Yes, definitely a bug. Corrected !

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: I refactored the code to use the hashing API and take the first `n` byte elements, depending on the use case.

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: I refactored the code to use the hashing API, this safety check should be taken care by the API.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +281,92 @@ impl CipherSeed {
     }
 }
 
+impl CipherSeed {
```
hansieodendaal, line 282: Using the `MacDomain` from the hasing API:
```suggestion
    // TODO: Remove when generic `HashingDomain::mac_hasher` implementation is ready
    fn mac_hasher<D: Digest + LengthExtensionAttackResistant>(
        domain_label: &'static str,
    ) -> DomainSeparatedHasher<D, MacDomain> {
        DomainSeparatedHasher::<D, MacDomain>::new(domain_label)
    }

    fn generate_mac(
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +281,92 @@ impl CipherSeed {
...
+
+        // salt should be 5 bytes long
+        if salt.len() != CIPHER_SEED_SALT_BYTES {
+            return Err(KeyManagerError::InvalidData);
+        }
+
+        // we take the first 32 bytes of the generated derived encryption key for MAC generation, see documentation
+        let passphrase_key = Self::generate_domain_separated_passphrase_hash(passphrase, salt)?[..32].to_vec();
+
+        let mac = DomainSeparatedHasher::<Blake256, GenericHashDomain>::new(DOMAIN_SEPARATION_LABEL)
```
hansieodendaal: Using the `MacDomain` from the hasing API:
```suggestion
         // TODO: Replace with generic `HashingDomain::mac_hasher` implementation
        let mac = Self::mac_hasher::<Blake256>(DOMAIN_SEPARATION_LABEL)
```

hansieodendaal (commented):
Hi, looking better and better. I have added a suggestion for using the `MacDomain` when generating the `mac`

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Great suggestion !

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -264,6 +282,100 @@ impl CipherSeed {
...
+        let passphrase_key = Self::generate_domain_separated_passphrase_hash(passphrase, salt)?[..32].to_vec();
+
+        // TODO: Replace with generic `HashingDomain::mac_hasher` implementation, when it is ready
+        let mac = Self::mac_hasher::<Blake256>(DOMAIN_SEPARATION_LABEL)
+            .chain(birthday)
+            .chain(entropy)
+            .chain(cipher_seed_version)
+            .chain(passphrase_key.as_slice())
+            .finalize()
+            .into_vec();
```
AaronFeickert: This is missing the salt.

AaronFeickert (changes requested):

base_layer/key_manager/src/cipher_seed.rs
jorgeantonio21: Totally my fault and a problematic bugs. Thanks for taking notice of it, I corrected accordingly!

hansieodendaal (dismissed):
LGTM

base_layer/key_manager/src/lib.rs
```diff
@@ -11,3 +11,29 @@ pub mod mnemonic_wordlists;
...
 #[cfg(feature = "wasm")]
 pub mod wasm;
+
+use tari_common::hashing_domain::*;
+
+/// The key_manager DHT domain separated hashing domain
+/// Usage:
+///   let hash = comms_dht_hash_domain().digest::<Blake256>(b"my secret");
+///   etc.
+pub fn comms_dht_hash_domain() -> HashingDomain {
```
hansieodendaal:
```suggestion
pub fn base_layer_key_manager() -> HashingDomain {
```

hansieodendaal (dismissed):
LGTM - one naming conflict

base_layer/key_manager/src/lib.rs
```diff
@@ -11,3 +11,29 @@ pub mod mnemonic_wordlists;
 #[allow(clippy::unused_unit)]
 #[cfg(feature = "wasm")]
 pub mod wasm;
+
+use tari_common::hashing_domain::*;
```
stringhandler: I'd prefer something more like this:

```suggestion

hash_domain!(KeyManagerHashDomain, "com.tari.keymanager");

pub fn create_hasher<D:Digest>(label: &'static str) -> DomainSeparatedHasher<....> {
   DomainSeparatedHasher::new::<D>(label)
}




```

stringhandler: or using the `create_hasher!()` macro directly

stringhandler: or finally, one method for each use case...
```suggestion
hash_domain!(KeyManagerHashDomain, "com.tari.keymanager");

pub fn create_passphrase_hasher() -> DomainSeparatedHasher<VarBlake, KeyManagerHashDomain> {}

pub fn create_cipher_seed_mac_hasher() -> Domain....
```

stringhandler (approved):
Happy, but will wait for confirmation from @AaronFeickert  before merging

AaronFeickert:
> Happy, but will wait for confirmation from @AaronFeickert before merging

A bit delayed (my apologies), but utACK.

PR #4297, opened 2022-07-11
feat(base_layer): checkpoint signature validation

mrnaveira:
Description
---
* New base layer validation of checkpoint signatures: verifies that ALL the signatures present in the checkpoint are valid
* Unified the error type for invalid signatures across all contract output types

Motivation and Context
---
The base layer needs to check if all the signatures in a checkpoint are valid.

This PR is based on previous work on checkpoints (see #4261  and #4285). The checkpoint signature is calculated as:
`e = H_1(signer_public_key || public_nonce || H_2(contract_id||commitment||merkle_root||checkpoint_number))`

But take into account that the `commitment` is still a mock value as of now we still don't have a method for creating a shared value.

How Has This Been Tested?
---
New unit test checks that invalid signatures are detected


SWvheerden (approved):
LGTM

sdbondi (approved):

PR #4303, opened 2022-07-12
feat(base_layer): checkpoint quorum validation

mrnaveira:
Description
---
Refactored the base layer committee validation for checkpoints:
* Now it's based on `HashSet` of VNC public keys: one for the constitution and other for the checkpoint
    * Surprisingly, clippy warns about `mutable_key_type` for using `PublicKey` types as the keys in the `HashSet`. I'm not sure where the mutability comes from. I decided to ignore that rule for the validation function
* The existing validation of members is now done by checking the _difference_ of the public key sets
* The new quorum validation is done by checking the _intersection_ of the public key sets

Motivation and Context
---
The base layer need to validate if the quorum of an incoming checkpoint is met. The minimum quorum for checkpoints is specified in the `ContractConstitution`, in the `checkpoint_params.minimum_quorum_required field`

How Has This Been Tested?
---
New unit test to check both successful and failing scenarios regarding quorum conditions


sdbondi (dismissed):
I like the use of HashSet for ease of code understanding, but it does result in some extra clones + heap allocations which could be avoided with a for loop. I think a HashSet<&PublicKey> would also work and use less memory (allocates only `n` words/usize).

In any case, happy for it to go in :+1:

sdbondi (approved):

PR #4307, opened 2022-07-14
feat: add sending to stealth address (command, grpc, gui)

Cifko:


stringhandler (approved):
Nonces should probably be saved to generate payment proofs, but this can be added later

PR #4310, opened 2022-07-14
feat: support for stealth addresses in one-sided transactions

agubarev:
Description
---
Added support of one-sided transactions with stealth addresses to the output scanner.

Motivation and Context
---
https://rfc.tari.com/RFC-0203_StealthAddresses.html

How Has This Been Tested?
---
manually, unit tests


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1892,159 @@ where
...
+                                .finalize()
+                                .as_ref(),
+                        )
+                        .unwrap();
+
+                        // calculating spending (public) key
+                        let ks = PublicKey::from_secret_key(&c) + public_key;
+
+                        if &ks != provided_spending_public.as_ref() {
+                            return None;
```
SWvheerden: I don't think return here will give the correct behaviour. 
If this case here is not correct, we should not drop all other correct UTXO and return back with none. This is in the middle of an iteration, I suspect you only want to continue here?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1892,159 @@ where
...
-                            known_one_sided_payment_scripts[i].input.clone(),
-                            known_one_sided_payment_scripts[i].private_key.clone(),
-                            output.sender_offset_public_key,
-                            output.metadata_signature,
-                            known_one_sided_payment_scripts[i].script_lock_height,
-                            output.covenant,
-                            output.encrypted_value,
-                        );
+        // ----------------------------------------------------------------------------
+        // scanning
```
SWvheerden: This function is now super long, 
Rather than make these comments, 
I would split the function into two.
One for scanning(this one can also potentially be divided into two) and one for processing.

SWvheerden (commented):
Looks good.

base_layer/wallet/src/output_manager_service/service.rs
agubarev: Yes, `return None` is to skip, to return from `filter_map` closure.  This function is to only find and add certain outputs, it's not supposed to pass anything through.

base_layer/wallet/src/output_manager_service/service.rs
agubarev: I thought about it, though the last function that adds scanned outputs would not be used for anything else. I have a TODO in place to make it shorter instead.

agubarev: Actually, you're right. I'll split it into two, because the 2 can be (possibly) useful for independent use.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1892,159 @@ where
...
+                        ) {
+                            Err(_) => None,
+                            Ok(spend_key) => Some((output.clone(), secret_key.clone() + c, spend_key)),
+                        }
+                    },
+
+                    _ => None,
+                }
+            })
+            .collect_vec();
```
sdbondi: Use this and remove itertools as dependency

```suggestion
            .collect::<Vec<_>>();
```

agubarev: it's a leftover from what I didn't use after all, though it's still nice to have around, and besides this service is already using it in dibbler. do you still want it out?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1891,159 @@ where
...
+                    // one-sided stealth address
+                    // NOTE: Extracting the nonce R and a spending (public aka scan_key) key from the script
+                    // NOTE: [RFC 203 on Stealth Addresses](https://rfc.tari.com/RFC-0203_StealthAddresses.html)
+                    [Opcode::PushPubKey(nonce), Opcode::Drop, Opcode::PushPubKey(provided_spending_public)] => {
+                        let secret_key = self.node_identity.secret_key().clone();
+                        let public_key = self.node_identity.public_key();
+
+                        // computing shared secret
+                        let c = RistrettoSecretKey::from_bytes(
+                            DomainSeparatedHasher::<Blake256, GenericHashDomain>::new("stealth address")
```
AaronFeickert: The [hashing API](https://github.com/tari-project/tari-crypto/blob/main/src/hashing.rs) suggests a descriptive reverse-DNS structure for domain separation labels.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1268,6 +1276,80 @@ where
...
+        >,
+    ) -> Result<TxId, TransactionServiceError> {
+        if self.node_identity.public_key() == &dest_pubkey {
+            warn!(target: LOG_TARGET, "One-sided spend-to-self transactions not supported");
+            return Err(TransactionServiceError::OneSidedTransactionError(
+                "One-sided-to-stealth-address spend-to-self transactions not supported".to_string(),
+            ));
+        }
+        let (nonce_private_key, nonce_public_key) = PublicKey::random_keypair(&mut OsRng);
+        let c = DomainSeparatedHasher::<Blake256, GenericHashDomain>::new("stealth address")
```
AaronFeickert: The [hashing API](https://github.com/tari-project/tari-crypto/blob/main/src/hashing.rs) suggests a descriptive reverse-DNS structure for domain separation labels.

base_layer/wallet/src/output_manager_service/service.rs
agubarev: we've changed label to `com.tari.stealth_address`, in both PRs; is that what you meant?

base_layer/wallet/src/output_manager_service/service.rs
stringhandler: I was also thrown by this. I think the filter_map is too large and you should probably rather use a simple for loop

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1891,159 @@ where
...
-                                return Err(err.into());
+        let scanned_outputs = outputs
+            .into_iter()
+            .filter_map(|output| {
+                // TODO: use MultiKey
+                match output.script.as_slice() {
+                    // ----------------------------------------------------------------------------
+                    // simple one-sided address
+                    [Opcode::PushPubKey(spending_key_public)] => {
+                        let keys = match self.resources.db.get_all_known_one_sided_payment_scripts() {
```
stringhandler: Calling this db method for every output is not necessary, rather move it outside the loop

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1891,159 @@ where
         Ok(())
     }
 
-    /// Attempt to scan and then rewind all of the given transaction outputs into unblinded outputs based on known
-    /// pubkeys
     fn scan_outputs_for_one_sided_payments(
```
hansieodendaal, line 2517:
```suggestion
    /// Attempt to scan and then rewind all of the given transaction outputs into unblinded outputs based on known
    /// pubkeys
    fn scan_outputs_for_one_sided_payments(
```

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: Agreed

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1888,97 +1891,175 @@ where
...
+                    }
+                },
+
+                // ----------------------------------------------------------------------------
+                // one-sided stealth address
+                // NOTE: Extracting the nonce R and a spending (public aka scan_key) key from the script
+                // NOTE: [RFC 203 on Stealth Addresses](https://rfc.tari.com/RFC-0203_StealthAddresses.html)
+                [Opcode::PushPubKey(nonce), Opcode::Drop, Opcode::PushPubKey(scanned_pk)] => {
+                    // computing shared secret
+                    let c = RistrettoSecretKey::from_bytes(
```
```suggestion
                    let stealth_key = PrivateKey::from_bytes(
```
We should use the generic types as far as possible

hansieodendaal (commented):
LGTM - just a couple of small comments

hansieodendaal:
Please add unit and cucumber tests (see `Scenario: Wallet sending and receiving one-sided transactions`) for one sided stealth addresses.

Cifko:
> Please add unit and cucumber tests (see `Scenario: Wallet sending and receiving one-sided transactions`) for one sided stealth addresses.

This is planned in next PR, because for that we need both PRs (this is scanning, and other one is sending) to be merged.

stringhandler:
Your PR has already been merged @Cifko

integration_tests/features/WalletTransactions.feature
```diff
@@ -32,6 +32,35 @@ Feature: Wallet Transactions
     Then all nodes are at height 30
     Then I wait for wallet WALLET_C to have at least 1500000 uT
 
+
```
stringhandler:
```suggestion
@critical
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -900,8 +942,204 @@ pub unsafe extern "C" fn private_key_from_hex(key: *const c_char, error_out: *mu
...
+/// as an out parameter.
+///
+/// ## Returns
+/// `*mut  TariPublicKey` - Returns a pointer to a `TariPublicKey`, which is a resulting sum of two public keys.
+/// Note that it returns ptr::null_mut() if either given key is null
+///
+/// # Safety
+/// The `public_key_destroy()` must be called when finished with a `TariPublicKey` to prevent a memory leak
+#[no_mangle]
+pub unsafe extern "C" fn private_key_add(
```
stringhandler: I don't think these are needed

stringhandler (dismissed):

stringhandler (approved):

PR #4312, opened 2022-07-14
refactor(base_layer): dan validation improvements

mrnaveira:
Description
---
* All helper functions for DAN validations are moved to `helpers.rs`, this cleans the code in each individual validator
* Improved some error handling:
    * Used the existing `MissingContractData` error variant when possible
    * Renaming `DuplicateUtxo` error type variant to `DuplicatedUtxo`
* Regarding unit testing:
    * Changed all error asserts to match the error variant instead of the error message
    * Used the existing `publish_contract` function when possible, to reduce boilerplate code
    * Refactored the available UTXOs to be an iterator instead of a vector. This allows us to avoid specifying each index when spending a UTXO (which is error prone as they can only be spent once).

Motivation and Context
---
During the development (and testing) of DAN base layer validations, some pain points were introduced. This PRs tries to clean some of them.

How Has This Been Tested?
---
All the existing unit tests for the DAN validations pass


sdbondi (approved):
Nice - the iterator makes much more sense :)

PR #4314, opened 2022-07-15
fix: transaction validation excess signature check

SWvheerden:
Description
---
Currently, only the transaction validator checks the rules that a transaction kernel must have a unique excess signature. But it only checks the first signature in the transaction. Transactions can have more than one kernel. 
This check is also not currently checked in block validation checks. 

This means that if a block with a duplicate excess signature is submitted, it will fail with an LMDB::error(duplicate key) and not with a validation error, although this is a validation constraint as per RFC 120. 

Motivation and Context
---
We need to check the kernel excess signature for uniqueness because this ensures that every kernel in the blockchain is unique.  By enforcing this, we ensure that replay attacks are impossible as the same kernel cannot be added to the blockchain again.

The excess signature is also used as a key index for the lookup of the kernels.


sdbondi (commented):
LGTM - we also have to check that duplicate kernel sigs are not in the current block. The DB should fail in that case, but the validator should have checked that before the DB layer.

stringhandler (approved):

Issue #4170, opened 2022-06-07
Wallet private key derivation is brittle


AaronFeickert:
[Wallet key derivation](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/key_manager.rs#L87-L92) relies on concatenation of seed entropy, a variable-length branch string, and an index. These values are hashed after concatenation using SHA-256, and parsed as a private key.

This approach is brittle:
- The concatenation can collide on distinct inputs, since arbitrary-length branch strings are allowed and the key index is not subject to a fixed-length (or even platform-independent) encoding
- The hash instantiation is not domain separated, which is unsafe if the inputs (or parts of them) are used elsewhere; it is therefore unsuitable as a key derivation function

The following are recommended:
- Use a modern key derivation function like `Blake2b` or `Blake3`
- Prepend the branch string length (where this length is itself in a fixed-length encoding) when encoding it as hash input
- Use a fixed-length and platform-independent encoding for the key index

Any future hashing API should be considered when addressing this issue.

jorgeantonio21:
A possible approach to this issue can be found [here](https://github.com/tari-project/tari/pull/4316). We make use of the current hashing [API](https://github.com/tari-project/tari-crypto/blob/main/src/hashing.rs), for proper domain separation. Moreover, current key index fixed length encoding is obtained by use of little endian representation of an `u64`.

PR #4316, opened 2022-07-17
fix!: improve wallet key derivation by use of proper domain separation (see issue  #4170)

jorgeantonio21:
Description
--- Add domain separation for wallet key derivation for type `KeyManager`.

Motivation and Context
--- The current wallet key derivation uses plain concatenation of variable length inputs. This procedure is known to not be collision resistant. A better methodology, as pointed [here](https://github.com/tari-project/tari/issues/4170), is to prepend the length of each datum to it before passing it through the hash function. We attain this by using the current hashing [API](https://github.com/tari-project/tari-crypto/blob/main/src/hashing.rs), which gives a suitable interface for domain separation, (which always prepends input length).

How Has This Been Tested?
--- Unit tests.


delta1:
Just out of interest this is a breaking change

AaronFeickert:
Length extension attack resistance isn't strictly a requirement here, as key derivation security doesn't depend on it. However, it is the case that any underlying hash function susceptible to length extension is probably a poor choice for other design or performance reasons anyway.

base_layer/key_manager/src/key_manager.rs
```diff
@@ -84,13 +87,17 @@ where
 
     /// Derive a new private key from master key: derived_key=SHA256(master_key||branch_seed||index)
```
AaronFeickert: This comment is no longer accurate, as `SHA-256` is not resistant to length extension attacks, and the input structure implied by the comment is out of date and no longer correct.

base_layer/key_manager/src/key_manager.rs
```diff
@@ -84,13 +87,17 @@ where
...
-            self.seed.entropy().to_vec().to_hex(),
-            self.branch_seed,
-            key_index
-        );
-        match K::from_bytes(D::digest(&concatenated.into_bytes()).as_slice()) {
+        // apply domain separation to generate derive key. Under the hood, the hashing api prepends the length of each
+        // piece of data for concatenation, reducing the risk of collisions due to redundance of variable length
+        // input
+        let derive_key = DomainSeparatedHasher::<D, GenericHashDomain>::new(DOMAIN_SEPARATION_LABEL)
+            .chain(self.seed.entropy().to_vec().to_hex())
```
AaronFeickert: Any reason to use a hex encoding here? It seems unnecessary, and the use of the hashing API is already a breaking change.

AaronFeickert (changes requested):

base_layer/key_manager/src/key_manager.rs
jorgeantonio21: Absolutely ! Thank you for taking notice of it !

base_layer/key_manager/src/key_manager.rs
jorgeantonio21: You are totally right Aaron, one can simply chain the seed entropy as it is.

stringhandler (dismissed):

base_layer/key_manager/src/lib.rs
```diff
@@ -11,3 +11,25 @@ pub mod mnemonic_wordlists;
...
+    HashingDomain::new("base_layer.key_manager.cipher_seed.mac_generation")
+}
+
+pub fn base_layer_key_manager_argon2_encoding() -> HashingDomain {
+    HashingDomain::new("base_layer.key_manager.cipher_seed.argon2_encoding")
+}
+
+pub fn base_layer_key_manager_chacha20_encoding() -> HashingDomain {
+    HashingDomain::new("base_layer.key_manager.cipher_seed.chacha20_encoding")
+}
```
hansieodendaal: I see these functions are not being used in this PR; I suppose they were created for use elsewhere?

hansieodendaal (dismissed):
LGTM

base_layer/key_manager/src/lib.rs
hansieodendaal: Ah I see, used in #4296.

base_layer/key_manager/src/lib.rs
```diff
@@ -11,3 +11,25 @@ pub mod mnemonic_wordlists;
...
+
+pub fn base_layer_key_manager() -> HashingDomain {
+    HashingDomain::new("base_layer.key_manager.key_manager")
```
hansieodendaal:
```suggestion
    HashingDomain::new("base_layer.key_manager")
```

stringhandler (approved):

PR #4319, opened 2022-07-18
feat(base_layer)!: new output field for minimum value range proof

mrnaveira:
**Warning: this PR includes consensus breaking changes**

Description
---
* New field `minimum_value_promise` for:
    * Struct `UnblindedOutput`, the range proof gets calculated automatically
    * Struct `TransactionOutput`, the range proof for the minimum value is verified in the existing `verify_range_proof` function
* Adding the new field cascades into the following changes: 
    * Created a new `TransactionOutputVersion::V2` to isolate the breaking changes of the new field. Used to differentiate when calculating the consensus hashing.
    * Updated the gRPC definitions
    * Updated the `lmdb` block storage
    * For the wallet:
        * Updated multiple services
        * Updated the Sqlite storage (including a new migration)
        * Changed the FFI function `wallet_import_external_utxo_as_non_rewindable` with the new parameter
    * Used the default value whenever necessary: genesis block, faucet, etc.
    * Updated multiple test helpers across the project 

Motivation and Context
---
The new Bulletproof+ allows us to create range proofs to demonstrate that a commitment value is greater than a particular value. This is very useful for future stake validation.

The purpose of this PR is to include a convenience new field (`minimum_value_promise`) in transaction outputs to specify a minimum value for range proofs. Range proofs should be automatically created and verified with minimum changes to existing code. The default value should be 0 so it behaves as current proofs.

How Has This Been Tested?
---
* New unit test that builds a `TransactionOutput` from a `UnblindedOutput` with the new field, and verifies the range proof, metadata and mask

BREAKING CHANGE: Will not sync with existing dibbler genesis block


base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -231,30 +238,54 @@ impl UnblindedOutput {
...
+        };
+
+        let transaction_error = TransactionError::RangeProofError(RangeProofError::ProofConstructionError(
+            "Creating transaction output".to_string(),
+        ));
+
+        let proof_bytes = proof_bytes_result.map_err(|_| transaction_error.clone())?;
+
+        RangeProof::from_bytes(&proof_bytes).map_err(|_| transaction_error)
+    }
```
sdbondi: nit:
Think these errors can be improved. Not a huge deal at all, but this also allocates on the heap regardless of an error occuring.
```suggestion
        let proof_bytes = proof_bytes_result.map_err(|err|       TransactionError::RangeProofError(RangeProofError::ProofConstructionError(
            format!("Failed to construct range proof: {}", err)
        )))?;

        RangeProof::from_bytes(&proof_bytes).map_err(|_| 
        TransactionError::RangeProofError(RangeProofError::ProofConstructionError(
            "Rangeproof factory returned invalid range proof bytes".to_string(),
        )))
    }
```

The second error is currently impossible but we can keep this just in case there is a change.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -575,4 +608,22 @@ mod test {
             1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 64, 2, 1
         ]);
     }
+
+    #[test]
+    fn it_builds_correctly_from_unblinded_output() {
```
sdbondi, line 614: Would also be good to have the failing case where the output commitment amount is less than the stated minimum

sdbondi (commented):
Looks good - two minor comments

sdbondi (dismissed):
Nice!

applications/tari_app_grpc/src/conversions/transaction_input.rs
```diff
@@ -63,6 +66,7 @@ impl TryFrom<grpc::TransactionInput> for TransactionInput {
                 PublicKey::from_bytes(input.sender_offset_public_key.as_bytes()).map_err(|err| format!("{:?}", err))?;
             let covenant = Covenant::from_bytes(&input.covenant).map_err(|err| err.to_string())?;
             let encrypted_value = EncryptedValue::from_bytes(&input.encrypted_value).map_err(|err| err.to_string())?;
+            let minimum_value_promise = MicroTari::zero();
```
SWvheerden: I would not default this here. 
I would add this to the GRPC field as well and then map it here

base_layer/core/src/proto/transaction.rs
```diff
@@ -163,6 +163,7 @@ impl TryFrom<proto::types::TransactionInput> for TransactionInput {
                 sender_offset_public_key,
                 Covenant::from_bytes(&input.covenant).map_err(|err| err.to_string())?,
                 EncryptedValue::from_bytes(&input.encrypted_value).map_err(|err| err.to_string())?,
+                MicroTari::zero(),
```
SWvheerden: Dont default here, map it to grpc.
This is going to cause a hash difference between the input and output of the Output.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright 2018 The Tari Project
 //
-// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+// Redistribution an&&d use in source and binary forms, with or without modification, are permitted provided that the
```
SWvheerden:
```suggestion
// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
```

base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -113,6 +114,8 @@ impl SingleReceiverTransactionProtocol {
             .map_err(|_| TPE::EncryptionError)?
             .unwrap_or_default();
 
+        let minimum_value_promise = MicroTari::zero();
```
SWvheerden: nit: I would not default this here, but its probably fine

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -692,6 +692,7 @@ where
             &commitment,
             single_round_sender_data.amount,
         )?;
+        let minimum_value_promise = MicroTari::zero();
```
SWvheerden: This is dangerous as well, perhaps we should ad this to `TransactionSenderMessage`

base_layer/wallet/src/wallet.rs
```diff
@@ -445,6 +445,7 @@ where
             script_lock_height,
             covenant,
             encrypted_value,
+            MicroTari::zero(),
```
SWvheerden: Don't default here, it's going to break, import it from the utxo. This info comes from a utxo.
This is going cause all the import issues again, we had with the UTXO from recovery etc.

SWvheerden (changes requested):
This is going to break on running the actual test net. 
We can default the min_output_value everywhere. 
We have to properly map it in. 
Currently, its mapped to some outputs, but the inputs are all default. This will cause a hash imbalance and the outputs will not be recognised.

base_layer/wallet/src/wallet.rs
mrnaveira: Fixed. Although this change cascaded into modifying the wallet FFI function parameters for utxo import (`wallet_import_external_utxo_as_non_rewindable`)

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -575,4 +608,41 @@ mod test {
             1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 64, 2, 1
         ]);
     }
+
+    #[test]
```
hansieodendaal, line 613: Please add a test for batch verification with minimum value promises other than zero

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -92,6 +91,9 @@ pub struct TransactionOutput {
     pub covenant: Covenant,
     /// Encrypted value.
     pub encrypted_value: EncryptedValue,
+    /// The minimum value of the commitment that is proven by the range proof
+    #[serde(default)]
+    pub minimum_value_promise: MicroTari,
```
hansieodendaal, line 96: Please update `pub fn batch_verify_range_proofs(` to read in the `minimum_value_promise` from the output - currently `0` is assigned.

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -216,30 +223,59 @@ impl UnblindedOutput {
...
 
+    fn construct_range_proof(&self, factories: &CryptoFactories) -> Result<RangeProof, TransactionError> {
+        let proof_bytes_result = if self.minimum_value_promise.as_u64() == 0 {
+            factories
+                .range_proof
+                .construct_proof(&self.spending_key, self.value.into())
+        } else {
+            let extension_degree = ExtensionDegree::DefaultPedersen;
+            let secrets = vec![self.spending_key.clone(); extension_degree as usize];
+            let extended_mask = RistrettoExtendedMask::assign(extension_degree, secrets)?;
```
hansieodendaal:
```suggestion
            let extended_mask = RistrettoExtendedMask::assign(ExtensionDegree::DefaultPedersen, vec![self.spending_key.clone()])?;
```

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -216,30 +223,59 @@ impl UnblindedOutput {
...
+
+            let extended_witness = RistrettoExtendedWitness {
+                mask: extended_mask,
+                value: self.value.into(),
+                minimum_value_promise: self.minimum_value_promise.as_u64(),
+            };
+
+            factories
+                .range_proof
+                .construct_extended_proof(vec![extended_witness], None)
```
hansieodendaal:
```suggestion
                .construct_extended_proof(vec![extended_witness], seed_nonce)
```
These proofs must also contain the optional rewind the key

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -216,30 +223,59 @@ impl UnblindedOutput {
...
             self.metadata_signature.clone(),
             self.covenant.clone(),
             self.encrypted_value.clone(),
+            self.minimum_value_promise,
         );
 
         Ok(output)
     }
 
+    fn construct_range_proof(&self, factories: &CryptoFactories) -> Result<RangeProof, TransactionError> {
```
hansieodendaal:
```suggestion
    fn construct_range_proof(&self, factories: &CryptoFactories, seed_nonce: Option<PrivateKey>) -> Result<RangeProof, TransactionError> {
```
These proofs must also contain the optional rewind the key

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -216,30 +223,59 @@ impl UnblindedOutput {
         }
         let commitment = factories.commitment.commit(&self.spending_key, &self.value.into());
 
+        let range_proof = self.construct_range_proof(factories)?;
```
hansieodendaal:
```suggestion
        let range_proof = self.construct_range_proof(factories, None)?;
```

base_layer/core/src/transactions/transaction_components/unblinded_output.rs
```diff
@@ -216,30 +223,59 @@ impl UnblindedOutput {
...
+        })?;
+
+        RangeProof::from_bytes(&proof_bytes).map_err(|_| {
+            TransactionError::RangeProofError(RangeProofError::ProofConstructionError(
+                "Rangeproof factory returned invalid range proof bytes".to_string(),
+            ))
+        })
+    }
+
     pub fn as_rewindable_transaction_output(
```
hansieodendaal, line 289: This function must also have a call to `self.construct_range_proof`
Replace 
``` rust
            let proof_bytes = factories.range_proof.construct_proof_with_recovery_seed_nonce(
                &self.spending_key,
                self.value.into(),
                &rewind_data.rewind_blinding_key,
            )?;
```
with 
``` rust
            let proof_bytes = self.construct_range_proof(factories, Some(rewind_data.rewind_blinding_key))?;
```

hansieodendaal (changes requested):
Hi there,

Looking good, except some changes required where the range proofs are constructed and verified.

Regards

base_layer/wallet/src/wallet.rs
SWvheerden: It should have

stringhandler (approved):

PR #4322, opened 2022-07-19
feat: tari launchpad downstream merge

CjS77:
Merging source code changes required for tari_launchpad.

The launchpad source code itself is present in the PR, but this will be removed presently.

The only code that really needs to be reviewed here are changes to files outside of `applications/launchpad`

A follow-up PR will be removing the `applications/launchpad` folder into its own repo.


scripts/file_license_check.sh
```diff
@@ -2,7 +2,7 @@
 # run from the repo root
 
 rg -i "Copyright.*The Tari Project" --files-without-match \
-    -g '!*.{Dockerfile,asc,bat,config,config.js,css,csv,drawio,gitkeep,hbs,html,iss,json,lock,md,min.js,ps1,py,rc,scss,sh,sql,svg,toml,txt,yml,vue}' . \
+    -g '!*.{Dockerfile,asc,bat,config,config.js,css,csv,drawio,env,gitkeep,hbs,html,ini,iss,json,lock,md,min.js,ps1,py,rc,scss,sh,sql,svg,toml,txt,yml,vue,ts,tsx}' . \
```
stringhandler: The `ts` and `tsx` are incorrect here, but that can be fixed in a different PR

stringhandler (dismissed):
Reviewed everything outside of the `launchpad` folder and it looks good

scripts/file_license_check.sh
CjS77: Sure.
It would also be handy to have a script inject the license metadata into files that don't have it

stringhandler (dismissed):

Issue #4110, opened 2022-05-17
Message encryption does not use a KDF


AaronFeickert:
Message encryption currently [parses an ECDH shared secret directly](https://github.com/tari-project/tari/blob/development/comms/dht/src/crypt.rs#L49-L56) as a ChaCha20 symmetric encryption key. Using an ECDH shared secret in this way is nonstandard and brittle.

Recommendations:
- Change the return type of `generate_ecdh_secret` from `CipherKey` to another (or a new) type, to prevent it from being used as a ChaCha20 key directly.
- Run the shared secret through a domain-separated key derivation function (KDF), and use the output as the ChaCha20 key. In particular, `Blake2b` has a `persona` parameter that is suitable for domain separation; passing the ECDH key as the input to such an instantiation is suitable for use as a KDF.

AaronFeickert:
Superseded by #4138.

AaronFeickert:
The new hashing API should be used here.

Issue #4138, opened 2022-05-25
Message decryption relies (in part) on deserialization for validity


AaronFeickert:
When a node wishes to determine if it can decrypt an incoming encrypted message, it performs several steps; these include:
- Generating an ECDH shared secret using the provided ephemeral public key
- Decrypting the message signature ("origin MAC")
- Deserializing the signature
- Deserializing the signature public key
- Verifying the signature against the message and header
- Decrypting the message
- Deserializing the message

If the message was not encrypted for the node:
- The first two steps will always succeed
- The next two deserialization steps may succeed (possibly maliciously)
- Signature verification will (almost certainly) fail

It is possible to alter this construction such that messages not intended for a node will be rejected with less computational effort, and without relying on particular serialization formats.

To do so, the sending node does the following:
- Uses domain-separated KDFs (recommendation is `Blake2b`) with ECDH shared-secret input to produce a `ChaCha20` key and `ChaCha20-Poly1305` key, respectively:
  - `key_message = Blake2b(persona="Tari message key", input=[ECDH_shared_secret])`
  - `key_signature = Blake2b(persona="Tari signature key", input=[ECDH_shared_secret])`
 - Encrypts the message using `ChaCha20` with `key_message`
 - Generates the signature as before
 - Encrypts the signature using the `ChaCha20-Poly1305` AEAD construction with `key_signature`
 
The recipient node does the following:
- Uses domain-separated KDFs (recommendation is `Blake2b`) with ECDH shared-secret input to produce a `ChaCha20` key and `ChaCha20-Poly1305` key, respectively:
  - `key_message = Blake2b(persona="Tari message key", input=[ECDH_shared_secret])`
  - `key_signature = Blake2b(persona="Tari signature key", input=[ECDH_shared_secret])`
- Attempts to decrypt the encrypted signature using the `ChaCha20-Poly1305` AEAD construction with `key_signature`
  - If tag verification fails prior to decryption, the message was not intended for this node, so we abort
- Deserializes the signature as before
- Verifies the signature as before
- Decrypts the message using `ChaCha20` with `key_message`
- Deserializes the message as before

This approach adds an additional 16-byte `ChaCha20-Poly1305` tag to the message structure.

This issue relates to (and in part supersedes) [this issue](https://github.com/tari-project/tari/issues/4110).

AaronFeickert:
[...]

PR #4334, opened 2022-07-22
fix: hash echd secret

Cifko:
Description
---
Hash echd secret.

Motivation and Context
---
https://github.com/tari-project/tari/issues/4110


AaronFeickert:
When #4138 is addressed, there will be two separate keys derived from the ECDH shared secret: one is for `ChaCha20` message encryption, and the other is for `ChaCha20-Poly1305` authenticated signature encryption. To keep the intent and types clear, it may be better to have `generate_ecdh_secret` return a non-key type, and write separate key derivation functions that produce the appropriate derived keys.

AaronFeickert:
See #4336 for current related work!

jorgeantonio21:
Currently, [#4138](https://github.com/tari-project/tari/issues/4138) has been merged to development.

Issue #4333, opened 2022-07-21
Message challenge should use hashing API


AaronFeickert:
Message [challenges](https://github.com/tari-project/tari/blob/development/comms/dht/src/crypt.rs#L114-L133) should use the new hashing API.

Additionally, the nomenclature for this is confusing. First, the hash output is used as the input message for the message "origin MAC", which is actually a signature. Additionally, it is not really a challenge, but simply a binding representation of certain message data and metadata. It may be useful to rename this in a way that makes its intent and design more obvious.

PR #4336, opened 2022-07-22
fix: address points in issue #4138 and companions

jorgeantonio21:
Description
--- Fix issues #4138, #4333 and #4170.

Motivation and Context
--- Add domain separation and cipher authentication for out bound messaging

How Has This Been Tested?
--- Unit tests.


comms/dht/src/crypt.rs
```diff
@@ -43,17 +49,53 @@ use crate::{
     version::DhtProtocolVersion,
 };
 
+const DOMAIN_SEPARATION_CHALLENGE_LABEL: &str = "com.tari.comms.dht.crypt.challenge";
+const DOMAIN_SEPARATION_KEY_MESSAGE_LABEL: &str = "com.tari.comms.dht.crypt.challenge";
```
AaronFeickert: Domain separators must be unique.

comms/dht/src/crypt.rs
```diff
@@ -43,17 +49,53 @@ use crate::{
...
 pub struct CipherKey(chacha20::Key);
+pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
+
+// TODO:
+//  1. rename mac_challenge function
+//  2. check if output of generate_ecdh_secret has correct output size
 
 /// Generates a Diffie-Hellman secret `kx.G` as a `chacha20::Key` given secret scalar `k` and public key `P = x.G`.
-pub fn generate_ecdh_secret<PK>(secret_key: &PK::K, public_key: &PK) -> CipherKey
+pub fn generate_ecdh_secret<PK>(secret_key: &PK::K, public_key: &PK) -> [u8; 32]
```
AaronFeickert: This encoded output size is specific to the elliptic curve group used, and does not hold generically.

comms/dht/src/crypt.rs
```diff
@@ -43,17 +49,53 @@ use crate::{
...
 where PK: PublicKey + DiffieHellmanSharedSecret<PK = PK> {
     // TODO: PK will still leave the secret in released memory. Implementing Zerioze on RistrettoPublicKey is not
     //       currently possible because (Compressed)RistrettoPoint does not implement it.
     let k = PK::shared_secret(secret_key, public_key);
-    CipherKey(*Key::from_slice(k.as_bytes()))
+    let mut output = [0u8; 32];
```
AaronFeickert, line 63: This encoded size is specific to the elliptic curve group used, and is not generic.

AaronFeickert: It might be useful to define a generic `DiffieHellmanSharedSecret` type that holds the result of an ECDH exchange.

comms/dht/src/crypt.rs
```diff
@@ -43,17 +49,53 @@ use crate::{
...
+
+pub fn generate_key_message(data: &[u8]) -> CipherKey {
+    // domain separated hash of data (e.g. ecdh shared secret) using hashing API
+    let domain_separated_hash =
+        DomainSeparatedHasher::<Challenge, GenericHashDomain>::new(DOMAIN_SEPARATION_KEY_MESSAGE_LABEL)
+            .chain(data)
+            .finalize()
+            .into_vec();
+
+    // Domain separation uses Challenge = Blake256, thus its output has 32-byte length
```
AaronFeickert, line 73: It is possible to check and enforce this at compile time, in case later changes subtly break this?

comms/dht/src/crypt.rs
```diff
@@ -43,17 +49,53 @@ use crate::{
...
+
+pub fn generate_key_signature_for_authenticated_encryption(data: &[u8]) -> AuthenticatedCipherKey {
+    // domain separated of data (e.g. ecdh shared secret) using hashing API
+    let domain_separated_hash =
+        DomainSeparatedHasher::<Blake256, GenericHashDomain>::new(DOMAIN_SEPARATION_KEY_SIGNATURE_LABEL)
+            .chain(data)
+            .finalize()
+            .into_vec();
+
+    // Domain separation uses Challenge = Blake256, thus its output has 32-byte length
```
AaronFeickert, line 81: Is it possible to check and enforce this at compile time, in case later changes subtly break this?

comms/dht/src/crypt.rs
AaronFeickert: The use of a specific ECDH shared secret type would mean this function enforces this input type.

comms/dht/src/crypt.rs
AaronFeickert: The use of a specific ECDH shared secret type would mean this function enforces this input type.

comms/dht/src/crypt.rs
```diff
@@ -88,6 +144,27 @@ pub fn encrypt(cipher_key: &CipherKey, plain_text: &[u8]) -> Vec<u8> {
     buf
 }
 
+/// Produces authenticated encryption of the signature using the ChaCha20-Poly1305 stream cipher,
+/// refer to https://docs.rs/chacha20poly1305/latest/chacha20poly1305/# for more details.
+/// Attention: as pointed in https://github.com/tari-project/tari/issues/4138, it is possible
+/// to use a fixed length Nonce, with homogeneous zero data, as this does not incur any security
```
AaronFeickert: Not a fixed-length nonce, but a fixed nonce.

AaronFeickert (changes requested):

comms/dht/src/crypt.rs
```diff
@@ -248,4 +251,96 @@ mod test {
...
+
+        let nonce_ga = chacha20poly1305::Nonce::from_slice(&nonce);
+        let cipher = ChaCha20Poly1305::new(&key.0);
+
+        let encrypted = cipher
+            .encrypt(nonce_ga, signature)
+            .map_err(|_| DhtOutboundError::CipherError(String::from("Authenticated encryption failed")))
+            .unwrap();
+
+        assert_eq!(encrypted.len(), n + 16);
```
AaronFeickert, line 260: You could remove the magic number by getting the tag size programmatically.

comms/dht/src/crypt.rs
jorgeantonio21: Definitely ! Thank you !

comms/dht/src/crypt.rs
jorgeantonio21: Definitely, changed  !

comms/dht/src/crypt.rs
sdbondi: > It is possible to check and enforce this at compile time, in case later changes subtly break this?

I believe so. Noting that this will not really break sublty and will panic in `Key::from_slice` if not 32-bytes. There are many tests that would signal this breakage, so we can decide if the following is worthwhile.

`digest::Output<D>` returned from `finalize` implements `Into<[u8;32]>` for 32-byte outputs  - this could be used to generate a compile error if the `DomainSeparatedHash<D>` type exposed the inner output type e.g `into_inner(self) -> Output<D>`
```rust
 let domain_separated_hash =
        DomainSeparatedHasher::<Blake256, GenericHashDomain>::new(DOMAIN_SEPARATION_KEY_SIGNATURE_LABEL)
            .chain(data)
            .finalize();
// Will not compile if Into<[u8;32]> is not implemented.
let output_array : [u8;32] = domain_separated_hash.into_inner().into();
```

This requires adding the `into_inner` function or reimplementing (something like) `From<[u8;D::OutputSize::to_usize()]>` on `DomainSeparatedHash`.

AaronFeickert: @sdbondi How generic can that be made? It may be the case that other domain-separated hash function applications require output of a different length.

sdbondi: @AaronFeickert That is handled by the [`generic_array` crate](http://fizyk20.github.io/generic-array/generic_array/). Here, we are enforcing in this function (or else compile fails) that the hash output is correctly convertible to a 32-byte array representation.

Adding trait bounds are not necessary here as we use concrete types, but the trait bounds would be `where digest::Output<D>: Into<[u8;32]>`

This became possible with this change (https://github.com/tari-project/tari-crypto/pull/119) since we can now use  `digest::Output<D>`

```rust
let domain_separated_hash = DomainSeparatedHasher::<Blake256, GenericHashDomain>::new(DOMAIN_SEPARATION_KEY_SIGNATURE_LABEL) .chain(data) .finalize();
 // Will not compile if Into<[u8;32]> is not implemented. 
let output_array : [u8;32] = domain_separated_hash.into();
```

sdbondi: So as the tari_crypto domain hasher code currently stands, you'd need to disambiguate between the functions on the hasher type and the `impl Digest` functions with the same name.
e.g.
```rust
     let hasher = comms_dht_hash_domain_key_signature().chain(data);
    let domain_separated_hash =<DomainSeparatedHasher<CommsChallenge, DHTCommsHashDomain> as Digest>::finalize(hasher);
    let hash: [u8; 32] = domain_separated_hash.into();
```

Since we know that we're using Blake256 and breakage will be caught in tests - I think we can leave this as is

sdbondi (approved):
Ack 

Tested with memorynet

comms/dht/src/crypt.rs
sdbondi: @jorgeantonio21 Would like to have the well-defined DH return type, perhaps for a follow-up PR?

Issue #4137, opened 2022-05-25
Wallet database encryption does not bind to field keys


AaronFeickert:
Wallet database field key-value entries are secured in place: AES-GCM is used to encrypt values, with keys left in the clear. However, the value encryption does not bind this operation to the field key. An attacker could replace these values with other encrypted values taken from elsewhere in the database (or otherwise encrypted using the same AES-GCM key) without detection.

One mitigation is to use the field key as associated data passed to the encryption and decryption operations.

PR #4340, opened 2022-07-22
fix: wallet database encryption does not bind to field keys #4137

agubarev:
Description
---
Added `source_key` to `encrypt_bytes_integral_nonce()` and `decrypt_bytes_integral_nonce()` which are used to encrypt and decrypt values in the storage backend. Also, encrypted values are now suffixed with a MAC.

Motivation and Context
---
https://github.com/tari-project/tari/issues/4137
Wallet database field key-value entries are secured in place: AES-GCM is used to encrypt values, with keys left in the clear. However, the value encryption does not bind this operation to the field key. An attacker could replace these values with other encrypted values taken from elsewhere in the database (or otherwise encrypted using the same AES-GCM key) without detection.

One mitigation is to use the field key as associated data passed to the encryption and decryption operations.

How Has This Been Tested?
---
unit test


base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -150,25 +150,49 @@ pub struct KeyManagerStateUpdateSql {
     primary_key_index: Option<Vec<u8>>,
 }
 
+// WARNING: using the same source key prefix for NewKeyManagerStateSql and KeyManagerStateSql
```
stringhandler: `NewKeyManagerStateSql` is the insertable version fo `KeyManagerStateSql` so they should be the same

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -150,25 +150,49 @@ pub struct KeyManagerStateUpdateSql {
...
 impl Encryptable<Aes256Gcm> for KeyManagerStateSql {
+    fn source_key(&self, field_name: &'static str) -> Vec<u8> {
+        [b"KEY_MANAGER", self.branch_seed.as_bytes(), field_name.as_bytes()]
```
stringhandler: Create a const for `KEY_MANAGER`

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -150,25 +150,49 @@ pub struct KeyManagerStateUpdateSql {
...
+            .join(&0)
```
stringhandler: what is the reason for this extra `0` ?

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -150,25 +150,49 @@ pub struct KeyManagerStateUpdateSql {
...
+        )?;
 
         Ok(())
     }
 }
 
 impl Encryptable<Aes256Gcm> for NewKeyManagerStateSql {
+    fn source_key(&self, field_name: &'static str) -> Vec<u8> {
+        [b"KEY_MANAGER", self.branch_seed.as_bytes(), field_name.as_bytes()]
+            .join(&0)
```
stringhandler: same

base_layer/wallet/src/output_manager_service/storage/sqlite_db/new_output_sql.rs
```diff
@@ -121,15 +121,46 @@ impl NewOutputSql {
 }
 
 impl Encryptable<Aes256Gcm> for NewOutputSql {
+    fn source_key(&self, field_name: &'static str) -> Vec<u8> {
+        // WARNING: using `OUTPUT` for both NewOutputSql and OutputSql due to later transition without re-encryption
+        [b"OUTPUT", self.script.as_slice(), field_name.as_bytes()]
```
stringhandler: use const

base_layer/wallet/src/output_manager_service/storage/sqlite_db/new_output_sql.rs
```diff
@@ -121,15 +121,46 @@ impl NewOutputSql {
...
+            .join(&0)
+            .to_vec()
+    }
+
     fn encrypt(&mut self, cipher: &Aes256Gcm) -> Result<(), String> {
-        self.spending_key = encrypt_bytes_integral_nonce(cipher, self.spending_key.clone())?;
-        self.script_private_key = encrypt_bytes_integral_nonce(cipher, self.script_private_key.clone())?;
+        eprintln!(
```
stringhandler: remove dbg code

base_layer/wallet/src/util/encryption.rs
```diff
@@ -25,33 +25,86 @@ use aes_gcm::{
...
+    source_key: Vec<u8>,
+    ciphertext: Vec<u8>,
+) -> Result<Vec<u8>, String> {
+    if ciphertext.len() < AES_NONCE_BYTES + AES_MAC_BYTES {
         return Err(AeadError.to_string());
     }
-    let (nonce, cipher_text) = ciphertext.split_at(AES_NONCE_BYTES);
+
+    let (nonce, ciphertext) = ciphertext.split_at(AES_NONCE_BYTES);
+    let (ciphertext, stored_mac) = ciphertext.split_at(ciphertext.len() - AES_MAC_BYTES);
```
stringhandler: replace with `check_sub` or `saturating_sub` otherwise this could possibly cause a panic

base_layer/wallet/src/util/encryption.rs
```diff
@@ -25,33 +25,86 @@ use aes_gcm::{
...
     let nonce = GenericArray::from_slice(nonce);
-    cipher.decrypt(nonce, cipher_text.as_ref()).map_err(|e| e.to_string())
+    let plaintext = cipher.decrypt(nonce, ciphertext.as_ref()).map_err(|e| e.to_string())?;
+
+    let mut mac = DomainSeparatedHasher::<Blake256, MacDomain>::new("com.tari.storage_encryption_mac")
```
stringhandler: This doesn't look correct. You are decrypting before verifying the MAC

stringhandler (commented):
Looks good. Sorry for the terseness in the comments. This is a very important area, and my intention is not to be mean but to be clear.

Looks like you are on the right track. Here are some suggestions:

I would suggest creating a struct or mod with constants that wraps all the keys in a crate in one place so that we don't create the same key twice and have spelling errors.

The name "source_key" is a little confusing since there are other keys (e.g. spending_key) in the wallet. Perhaps a better name is `source_field` or `full_field_name`.

nit: Please remove package-lock.json from the commit set

sdbondi (approved):

PR #4341, opened 2022-07-25
feat: apply hashing api to tari_script and mmr crates

hansieodendaal:
Description
---
Applied the new `tari_crypto`  hashing API to `tari_script` and `mmr` crates.

Motivation and Context
---
Busy implementing the new `tari_crypto` hashing API throughout.

How Has This Been Tested?
---
Passed unit tests; some needed modification
Passed cucumber tests


base_layer/mmr/tests/support/mod.rs
```diff
@@ -22,23 +22,22 @@
 //
 
 use croaring::Bitmap;
-use digest::Digest;
 use tari_crypto::hash::blake2::Blake256;
-use tari_mmr::{Hash, HashSlice, MerkleMountainRange, MutableMmr};
+use tari_mmr::{mmr_hash_domain, Hash, HashSlice, MerkleMountainRange, MutableMmr};
 
-pub type Hasher = Blake256;
+pub type DigestAlgorithm = Blake256;
```
stringhandler: I don't think this is better. In fact I would prefer the type alias is dropped altogether

base_layer/wallet/src/transaction_service/error.rs
```diff
@@ -166,6 +167,8 @@ pub enum TransactionServiceError {
     BaseNodeNotSynced,
     #[error("Value encryption error: `{0}`")]
     EncryptionError(#[from] EncryptionError),
+    #[error("Two slices have different lengths")]
+    HashingDomainError(#[from] HashingDomainError),
```
CjS77: Rather keep the error message consistent with the type. 
```suggestion
    #[error("There was an error in generating a hash: {0}")]
    HashingDomainError(#[from] HashingDomainError),
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -989,7 +990,10 @@ where
         let tx_id = TxId::new_random();
         // this can be anything, so lets generate a random private key
         let pre_image = PublicKey::from_secret_key(&PrivateKey::random(&mut OsRng));
-        let hash: [u8; 32] = Sha256::digest(pre_image.as_bytes()).into();
+
+        let hash: [u8; 32] = tari_script_hash_domain()
+            .digest::<Sha256>(pre_image.as_bytes())
```
CjS77, line 995: While we're changing things, might be a good time to switch to `Blake256`, unless there's a good reason we're using SHA-2 here

CjS77: Can be done later btw, just marking the comment.

common/src/hashing_domain.rs
```diff
@@ -30,10 +31,10 @@ pub struct HashingDomain {
 }
 
 /// Error type for the pipeline.
-#[derive(Debug, Error)]
+#[derive(Debug, Clone, Error, PartialEq, Eq, Serialize, Deserialize)]
 pub enum HashingDomainError {
     #[error("Two slices have different lengths")]
-    CopyFromSlice,
+    SlicesHaveDifferentLengths,
```
CjS77: Is this error produced somewhere besides `as_fixed_bytes`?
If yes, add `DigestTooShort` variant
If no, change to `DigestTooShort` variant.

In any event, this would be far more instructive:
```
    #[error("Cannot create a slice of length {0} from a digest of length {1}.")]
    DigestTooShort(usize, size),
```

(Could also leave the parameters out)

infrastructure/tari_script/src/error.rs
```diff
@@ -49,6 +50,8 @@ pub enum ScriptError {
     VerifyFailed,
     #[error("as_hash requires a Digest function that returns at least 32 bytes")]
     InvalidDigest,
+    #[error("Two slices have different lengths")]
+    HashingDomainError(#[from] HashingDomainError),
```
CjS77: Same as previous comment in this respect.

CjS77 (commented):
Some little nits.

Otherwise this is an untested ACK.

Note: Look into implementing the `Digest` trait in the Hashing API rater than exposing a `digest` function in a wrapper here.

base_layer/wallet/src/transaction_service/service.rs
stringhandler: Let's do it now rather

stringhandler:
> Note: Look into implementing the Digest trait in the Hashing API rater than exposing a digest function in a wrapper here.

@CjS77 I was thinking the same thing and tried it: https://github.com/tari-project/tari-crypto/pull/119

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: This is part of the function called `send_sha_atomic_swap_transaction` for the "SHA-XTR atomic swap transaction", so in this case I think it should remain `Sha256`.

hansieodendaal:
Added some review changes while waiting for https://github.com/tari-project/tari-crypto/pull/119 to be approved/merged. This may have a substantial impact on the domain separated hashing implementation here.

hansieodendaal:
Closing this PR to follow principles in #4324

PR #4342, opened 2022-07-25
feat: build arm64 binaries from json matrix

leet4tari:
Description
Update supporting scripts to cross-compile build extra binaries for ARM64 OSX and Linux

Motivation and Context
Add ARM64 builds for OSX and Linux

How Has This Been Tested?
Build and run locally and in GHA from local fork


.github/workflows/base_node_binaries.json
```diff
@@ -0,0 +1,58 @@
...
+    "features": "safe"
+  },
+  {
+    "name": "windows-x64",
+    "runs-on": "windows-2019",
+    "rust": "stable",
+    "target": "x86_64-pc-windows-msvc",
+    "cross": false,
+    "target_cpu": "x86-64",
+    "features": "safe"
```
stringhandler, line 47: I think windows doesn't need safe... but could be wrong

buildtools/windows_inno_installer.iss
```diff
@@ -98,10 +99,10 @@ Source: "..\applications\tari_miner\windows\start_tari_miner.lnk"; DestDir: "{ap
 Source: "..\applications\tari_merge_mining_proxy\windows\start_tari_merge_mining_proxy.lnk"; DestDir: "{app}"; Flags: ignoreversion
 Source: "..\applications\tari_merge_mining_proxy\windows\start_xmrig.lnk"; DestDir: "{app}"; Flags: ignoreversion
 Source: "..\applications\tari_base_node\windows\start_tor.lnk"; DestDir: "{app}"; Flags: ignoreversion
-Source: "..\target\release\tari_base_node.exe"; DestDir: "{app}\runtime"; Flags: ignoreversion
-Source: "..\target\release\tari_console_wallet.exe"; DestDir: "{app}\runtime"; Flags: ignoreversion
-Source: "..\target\release\tari_miner.exe"; DestDir: "{app}\runtime"; Flags: ignoreversion
-Source: "..\target\release\tari_merge_mining_proxy.exe"; DestDir: "{app}\runtime"; Flags: ignoreversion
+Source: "D:\a\tari\tari\dist\tari_base_node.exe"; DestDir: "{app}\runtime"; Flags: ignoreversion
```
stringhandler: woops

stringhandler (commented):
Looks ok. Need to resolve conflicts

.github/workflows/base_node_binaries.json
leet4tari: CPU target set, if the windows binary is run on older hardware, I think we see a ```illegal cpu instruction```. We could really use the matrix build and target a faster CPU instruction set like avx2, but doubling up the builds might not add value?

buildtools/windows_inno_installer.iss
leet4tari: Referring to the ```D:``` hard codec location of the target build? I have no idea how to pass a variable into Inno ISS scripts.

.github/workflows/base_node_binaries.json
stringhandler: I think we generally only target Haswell and above. Anything older should compile from source.

I think we only use safe on macos because it used to fail

buildtools/windows_inno_installer.iss
stringhandler: https://jrsoftware.org/ispphelp/index.php?topic=isppcc

leet4tari: Added a default and command line override.

Also included tari_validator_node.exe to Windows package.

.github/workflows/base_node_binaries.json
leet4tari: ```target_cpu``` and ```features``` could be matrix build in future, but thinking we get the base in and then we can extend as needed or as requested?

Also not sure what features we want to enable in a build, but can be done per matrix set.

stringhandler (approved):

PR #4343, opened 2022-07-26
feat(dan): invocation of functions in templates

mrnaveira:
Description
---
* New template test that simulates a component with state, a constructor and get/set methods
    * For now we don't perform real calls to handle state and constructor calls as we don't have those implemented yet
* Renamed existing template test to a more proper "hello world" name
* Refactor of all template wasm compilation/execution/test into reusable functions, as much as possible without macros:
    * Template test initialization and invocation
    * Generation of template ABI code
    * Generation of template wasm code, allowing for multiple functions with different signatures

Motivation and Context
---
One of the next steps in template wasm compilation and execution is the implementation of a more complex example with multiple functions. We also want to refactor as much as possible all common code between examples

How Has This Been Tested?
---
The new units test for the templates pass


dan_layer/engine/tests/state/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        arguments: vec![Type::U32], // the component_id
+        output: Type::U32,  // the stored value
+    }];
+
+    generate_abi(template_name, functions)
+}
+
+#[no_mangle]
+extern "C" fn State_main(call_info: *mut u8, call_info_len: usize) -> *mut u8 {
+    let mut template_impl = TemplateImpl::new();
```
sdbondi, line 78: I like the idea and execution, but the problem with this approach is that it's dynamic, so each time main is called (main is called for every template invocation) a HashMap is allocated. 

This is fine for now though, but I think the macros will generate static code.

dan_layer/engine/tests/common/src/lib.rs
```diff
@@ -81,11 +85,10 @@ unsafe extern "C" fn tari_alloc(len: u32) -> *mut u8 {
     ptr
 }
 
-#[no_mangle]
-unsafe extern "C" fn tari_free(ptr: *mut u8) {
+pub unsafe fn tari_free(ptr: *mut u8) {
```
sdbondi: If `common::tari_alloc` are `extern`s they do not need to be "proxied" and exposed in the macro-generated code.

dan_layer/engine/tests/state/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    generate_main(call_info, call_info_len, template_impl)
+}
+
+extern "C" {
+    pub fn tari_engine(op: u32, input_ptr: *const u8, input_len: usize) -> *mut u8;
+}
+
+#[no_mangle]
+unsafe extern "C" fn tari_alloc(len: u32) -> *mut u8 {
```
sdbondi: These won't be needed if `common::tari_alloc` are `extern`s

sdbondi (dismissed):
Looks good, I think the approach can be refined from here.

dan_layer/engine/tests/state/src/lib.rs
mrnaveira: Agreed. The macros will most likely generate a `match` clause. For now I just wanted a clean way to reuse code.

sdbondi (approved):
LGTM :tada:

PR #4344, opened 2022-07-26
refactor: remove chrono as dependency

stringhandler:
Description
---
Remove dependency on chrono for birthday calculation in cipherseed

Motivation and Context
---
The wallet recovery uses a few extra days before the actual birthday when scanning, so being accurate to the timezone is not necessary, and certainly doesn't need an extra crate to calculate it.

How Has This Been Tested?
---
tests


stringhandler:
fyi @jorgeantonio21

sdbondi (approved):

PR #4358, opened 2022-07-28
feat(dan): basic template macro

mrnaveira:
Description
---
* New engine package to encapsulate the macro logic. In the future we may consider extracting it to a separate project or repo.
* The user template code handling is encapsulated into the `ast` module.
* Each section of the generated template code is split into different modules:
    * **definition**: generates a verbatim module of the user code, needed for the template macro to compile properly.
    * **abi**: generates the ABI section of the contract wasm, needed in the engine to know the contract function signatures.
    * **dispatcher**: executes a call invocation, by deciding which user-defined code to call and encoding the result.
    * **dependencies**: helper functions that need to be included into the wasm, as well as the `tari_engine` import signature.
* Current limitations:
    * The function arguments are not yet processed and encoded, so at the moment it only supports user-defined functions with no parameters. But return values are handled.
    * Struct fields and `self` is not handled yet
    * Calls to the tari engine import are not being done

Motivation and Context
---
Template developers need an ergonomic way of writing contract code. This PR aims to provide a new macro `template!` to allow easy development of new _basic_  contracts. And for _basic_ I mean no struct fields, no resource access and no composability, only stateless functions. Those features will be added in future PRs.

With this PR, the existing "Hello World" example is rewritten as:
```

template! {
    struct HelloWorld {}

    impl HelloWorld { 
        pub fn greet() -> String {
            "Hello World!".to_string()
        }
    }
}
```

How Has This Been Tested?
---
* The unit test for the "Hello World" contract example now is written using the new macro. It compiles and runs in our engine successfully
* New unit tests for some complex sections of the macro


dan_layer/engine/tests/hello_world/src/lib.rs
```diff
@@ -20,35 +20,14 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-// TODO: we should only use stdlib if the template dev needs to include it e.g. use core::mem when stdlib is not
-// available
+use tari_template_macros::template;
 
-use common::{generate_abi, generate_main, TemplateImpl};
-use tari_template_abi::{encode_with_len, FunctionDef, Type};
+template! {
```
sdbondi, line 25: The scoping is why we'll need to put something like  

```rust
#[tari::contract]
mod hello_world {
   pub struct HelloWorld { etc.}
}
```

If we want to avoid wrapping in `template! {...}`, unless there's another way to get the scoping (I can't think of any) - the main reason is that code completion on some editors breaks when wrapped in declarative macros.

sdbondi (approved):
A perfect start!

We can look avoiding the `template! {... }` wrapper in another PR.

CjS77:
Looks great. Why is it all in a tests folder though?

sdbondi:
Yeah, let's move the macro out of `tests` and then it should be good to merge

EDIT: on second thoughts, I want to prevent conflicts with https://github.com/tari-project/tari/pull/4350 since I already moved out the "common" crate that was in tests in that PR. Let's make things easy and merge this one and move it out later

dan_layer/engine/tests/hello_world/src/lib.rs
CjS77: This is fine IMO

It also allows some flexibility, because you can also write your contract in `my_contract.rs`, and then in `lib.rs` have

```rust
#[tari::contract]
mod my_contract;
```

PR #4361, opened 2022-07-29
refactor(dan): template macro as an attribute macro

mrnaveira:
Description
---
* Refactored the existing declarative macro (`template!`) into an attribute one (`#[template]`)
* Moved the macro project to the root of the `dan_layer` crate
* Improved unit testing of the generated ABI section of the wasm, to assert different function signatures.
* Used the `indoc!` macro in tests to make the code more readable

Motivation and Context
---
The existing template definition macro was created as a declarative macro, but code completion on some editors breaks when wrapped such types of macros.

The main goal of this PR is to transform the template macro into an attribute macro, to be used like:
```

#[template]
mod hello_world {
    struct HelloWorld {}

    impl HelloWorld { 
        pub fn greet() -> String {
            "Hello World!".to_string()
        }
    }
}
```

How Has This Been Tested?
---
* Existing tests pass


sdbondi (dismissed):
Nice :tada:

sdbondi (approved):

dan_layer/template_macros/src/template/abi.rs
```diff
@@ -101,9 +102,20 @@ mod tests {
...
+                    pub fn no_args_function() -> String {
+                        \"Hello World!\".to_string()
+                    }
+                    pub fn some_args_function(a: i8, b: String) -> u32 {
+                        1_u32
+                    }
+                    pub fn no_return_function() {}  
+                } 
+            }
+        "})
```
sdbondi, line 118: You can also use native rust  multiline strings 

```suggestion
      let input = TokenStream::from_str(
            r#"
            mod hello_world {
                struct HelloWorld {}
                impl HelloWorld {
                    pub fn greet() -> String {
                        "Hello World!".to_string()
                    }
                } 
            }
        "#,
        )
        .unwrap();
```

mrnaveira: The parser was giving me a hard time with native multilines due to some span issues. Using indoc we can avoid those kind of issues and it's a dev-dependency used only in tests, so my vote goes to indoc.

sdbondi: Thanks for the explanation - sounds good to me

PR #4362, opened 2022-07-29
fix(dht)!: add message padding for message decryption, to reduce message length leaks (fixes #4140)

### LINKED ISSUE #4140: Encrypted message lengths are trivially leaked
Message encryption and [decryption](https://github.com/tari-project/tari/blob/development/comms/dht/src/inbound/decryption.rs#L331-L332) are performed using the `ChaCha20` stream cipher. Because no padding is applied, the length of encrypted messages is trivially leaked, and is equal to the length of the resulting ciphertext. Additionally, header data like the message type is sent in the clear. An adversary may be able to infer additional information from this.

Mitigations include:
- Fixing the length of all messages, and padding messages to this length
- Discretizing the length of allowed messages, which may leak less information

It's important to ensure that no padding-related attacks are introduced by such mitigations.

jorgeantonio21:
Description
--- Message length is leaked while performing ChaCha20 encryption for message outbound/inbound. In order to mitigate this vulnerability, we use padding to every message before encryption. In this way, message length is always a multiple of a base length value.

Motivation and Context
--- Tackle issue #4140, see [here](https://github.com/tari-project/tari/issues/4140).

How Has This Been Tested?
--- Add unit tests


stringhandler (dismissed):
Could have also used a VarInt for the length, but since we're padding anyway, it's fine

comms/dht/src/crypt.rs
```diff
@@ -55,6 +55,9 @@ use crate::{
 pub struct CipherKey(chacha20::Key);
 pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
 
+const LITTLE_ENDIAN_U64_SIZE_REPRESENTATION: usize = 8;
+const MESSAGE_BASE_LENGTH: usize = 128;
```
sdbondi: A partial transaction message in my test was 4111 bytes, to obfuscate this from other messages, we'll need to make every message bigger than this. I suggest a `MESSAGE_BASE_LENGTH` of 6000.

comms/dht/src/crypt.rs
```diff
@@ -66,6 +69,49 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
+    let div_n_base_len = (n + k) / MESSAGE_BASE_LENGTH;
+    let output_size = (div_n_base_len + 1) * MESSAGE_BASE_LENGTH;
+
+    // join prepend_message_len | message | zero_padding
+    let mut output = prepend_to_message.to_vec();
+    output.append(&mut message.to_vec());
+    output.append(&mut [0u8; 128].to_vec());
+
+    // we have to truncate the vector to output_size
+    output.truncate(output_size);
```
sdbondi: This code has 6 heap allocations, try this instead (1 heap allocation):
```suggestion
    let mut output = Vec::with_capacity(output_size);
    output.extend_from_slice(&prepend_to_message);
    output.extend_from_slice(&message);
    output.extend(iter::repeat(0u8).take(output_size - n - k));
```

comms/dht/src/lib.rs
```diff
@@ -124,6 +124,10 @@ use tari_crypto::{hash_domain, hashing::DomainSeparatedHasher};
 
 hash_domain!(DHTCommsHashDomain, "comms.dht");
 
+pub fn comms_dht_message_hash() -> DomainSeparatedHasher<CommsChallenge, DHTCommsHashDomain> {
+    DomainSeparatedHasher::<CommsChallenge, DHTCommsHashDomain>::new("message_hash")
```
sdbondi:
```suggestion
    DomainSeparatedHasher::<CommsChallenge, DHTCommsHashDomain>::new_with_label("message_hash")
```

sdbondi (commented):
Looks good - some comments

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
+    fn decryption_fails_if_pad_message_prepend_is_modified() {
+        let pk = CommsPublicKey::default();
+        let key = CipherKey(*chacha20::Key::from_slice(pk.as_bytes()));
+        // long text makes last test case deterministic
+        let message = "This is my secret message, keep it secret !".as_bytes().to_vec();
+        let mut encrypted = encrypt(&key, &message);
+
+        // failure in case message length prepending has been modified such that resulting
+        // length is too big to fit within pad message length
+        encrypted[size_of::<Nonce>() + 1] += 1;
```
sdbondi: It's possible for the value at `encrypted[size_of::<Nonce>() + 1]` to be 255. So 255 + 1 will overflow causing the test to be flaky.

Also increasing the ciphertext byte won't necessarily result in an increase in the plaintext byte. This is observed when `unwrap_err` panics sometimes. 

Suggest rather testing the `get_original_message_from_padded_text` function directly and fudging with the padded length.

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
+        assert!(decrypt(&key, &encrypted)
+            .unwrap_err()
+            .to_string()
+            .contains("Original length message is invalid"));
+
+        encrypted[size_of::<Nonce>() + 1] -= 1;
+
+        // failure in case message length fits whithin pad message length, but its original length has been modified
```
sdbondi:
```suggestion
        // failure in case message length fits within pad message length, but its original length has been modified
```

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
             .contains("Authenticated decryption failed"));
     }
+
+    #[test]
+    fn pad_message_correctness() {
+        // test for small message
+        let message = &[0u8, 10, 22, 11, 38, 74, 59, 91, 73, 82, 75, 23, 59];
+        let prepend_message = message.len().to_le_bytes();
+        let pad_message = pad_message_to_base_length_multiple(message);
+        let pad = [0u8; 107];
```
sdbondi:
```suggestion
     let prepend_message = (message.len() as u64).to_le_bytes();
        let pad = iter::repeat(0u8)
            .take(MESSAGE_BASE_LENGTH - message.len() - prepend_message.len())
            .collect::<Vec<_>>();
```

comms/dht/src/crypt.rs
```diff
@@ -66,6 +69,49 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
     output
 }
 
+fn pad_message_to_base_length_multiple(message: &[u8]) -> Vec<u8> {
+    let n = message.len();
+    // little endian representation of message length, to be appended to padded message,
+    // assuming our code runs on 64-bits system
+    let prepend_to_message = (n as u64).to_le_bytes();
```
sdbondi: The maximum message size permitted by the protocol is 4MiB, so using u32 is plenty (you could have a 3-byte fixed length which gives a maximum length of 16MiB, but not too worried about 1 extra byte here)

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
+            pad_message[prepend_message.len()..prepend_message.len() + message.len()]
+        );
+        // pad is well specified
+        assert_eq!(pad, pad_message[prepend_message.len() + message.len()..]);
+
+        // test for large message
+        let message = &[100u8; 900];
+        let prepend_message = message.len().to_le_bytes();
+        let pad_message = pad_message_to_base_length_multiple(message);
+        let pad = [0u8; 116];
```
sdbondi:
```suggestion
 let message = &[100u8; MESSAGE_BASE_LENGTH * 8 - 100];
        let prepend_message = message.len().to_le_bytes();
        let pad = iter::repeat(0u8)
            .take((8 * MESSAGE_BASE_LENGTH) - message.len() - prepend_message.len())
            .collect::<Vec<_>>();
```

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
+            pad_message[prepend_message.len()..prepend_message.len() + message.len()]
+        );
+        // pad is well specified
+        assert_eq!(pad, pad_message[prepend_message.len() + message.len()..]);
+
+        // test for base message of multiple base length
+        let message = &[100u8; 1016];
+        let prepend_message = message.len().to_le_bytes();
+        let pad_message = pad_message_to_base_length_multiple(message);
+        let pad = [0u8; 128];
```
sdbondi:
```suggestion
 let message = &[100u8; MESSAGE_BASE_LENGTH * 9 - 123];
        let prepend_message = message.len().to_le_bytes();
        let pad = iter::repeat(0u8)
            .take((9 * MESSAGE_BASE_LENGTH) - message.len() - prepend_message.len())
            .collect::<Vec<_>>();
```

comms/dht/src/crypt.rs
```diff
@@ -325,4 +379,163 @@ mod test {
...
+    }
+
+    #[test]
+    fn decryption_fails_if_message_body_is_modified() {
+        let pk = CommsPublicKey::default();
+        let key = CipherKey(*chacha20::Key::from_slice(pk.as_bytes()));
+        let message = "My secret message, keep it secret !".as_bytes().to_vec();
+        let mut encrypted = encrypt(&key, &message);
+
+        encrypted[size_of::<Nonce>() + LITTLE_ENDIAN_U64_SIZE_REPRESENTATION + 1] += 1;
```
sdbondi: Use wrapping_add so that if the value is 255 it can wrap to 0
```suggestion
        let b = &mut encrypted[size_of::<Nonce>() + LITTLE_ENDIAN_U64_SIZE_REPRESENTATION + 1];
        *b = b.wrapping_add(1);
```

comms/dht/src/crypt.rs
jorgeantonio21: Done

comms/dht/src/crypt.rs
jorgeantonio21: Done

comms/dht/src/crypt.rs
jorgeantonio21: Amazing ! Thank you

comms/dht/src/crypt.rs
jorgeantonio21: Done

comms/dht/src/crypt.rs
jorgeantonio21: Done

comms/dht/src/crypt.rs
```diff
@@ -226,9 +277,10 @@ mod test {
...
...6d5954b3a8f66dd547d3b72565437ead511c9d5342406aace95e5cc31f2c6d618a24c219d0298f980a571ce29b999d20b3ed94a60e286ed7ddc647c439e3d421814da6d91c8b7d5b3fa70ec6a3c261ab9ee4ac779545edcc7db6df3345db26b91c5a997dac5e62b75dc05358821bab4fe65a049fe9ce8e537ae81a10dfcca0ef97c6cb95e3ff1573e5461f7b505e1678fbe97a41ab696b53f6ea09038ecda09eed34247251424b766306c0c64fd836274cf85fe0e0c19638127c2210b580c9194fe0cccac7ac80e3dde38e9cccd6a194ee923f4e73800bec0c77f60553f9c7c8413ea87d20c114d7b415fdd87fc55f273b1b3a9f9c71c4462d5b3f300daf0fc6c338278e5991e0c6de07a3c288d237df00325230be204f7b2bb7a127ac28b001e4225e910eeb9521f5af6cfeae1f18c08bf8ac9d1513c3794ba5b8ea9fb9a57825cb154fc1e9a9dddf809dd6bb11a207625b23b274344e7e0b7ca666e456735d5901f1341aca42e749183823b3debbd563aeebc68f9b15dce13d0fc1acef47d38d5967c2b6b3fe8ed69b180dbcbf17455ee6825641202ccd145c0a0a0f4091622338f48474e5838d8915f814eb87ad45e710b07f79f662c2120278ce05978d8a7aee20fc5661a08c072977ed878092e7183332b70c9c54db307c705e527f6fd2076e39c216b0490f552d52a109652958c62fc6bf7f913818dbdf5d92550779aae541d54d059d5844658422c17a24e374fa6f92e5a9fda87eee249747b9cd292043c9731d2c1d08d06eab030fb49e779cb58bf4f776d6aa0185db860007d8b2d0f7205dacb9201ac9538d2c37062f736b6b44e971e11500",
```
sdbondi, line 281: Could also use `include_str!` or `include_bytes!` - but this is completely fine as is :) 
```suggestion
            include_str!!(../../tests/assets/encrypted-msg.hex);
```

sdbondi (approved):
Thanks for the corrections - tested and works! :rocket:

jorgeantonio21:
> Thanks for the corrections - tested and works! rocket

Happy to hear that ! :)

PR #4373, opened 2022-08-02
feat: add domain hashing for consensus encoding

hansieodendaal:
Description
---
Added domain hashing for consensus encoding

Motivation and Context
---
Implementing the tari_crypt ohashing API

How Has This Been Tested?
---
- Passed unit tests
- Passed cucumber tests


base_layer/core/src/consensus/consensus_encoding/hash_writer.rs
```diff
@@ -23,26 +23,34 @@
...
+use tari_crypto::{hash::blake2::Blake256, hashing::AsFixedBytes};
 
-use crate::consensus::ConsensusEncoding;
+use crate::consensus::{ConsensusEncoding, ConsensusEncodingDomainHasher};
 
-#[derive(Clone)]
 pub struct ConsensusHashWriter<H> {
-    digest: H,
+    hasher: ConsensusEncodingDomainHasher<H>,
+    internal_state: Vec<u8>,
```
sdbondi: One of the primary goals of the consensus hash writer is to have zero heap allocations since hashes are done very often. This will result in many allocations as each element is added. I understand why you did this (due to the length being added to the domain-separated hash for each update). I suggest we simply add a static domain separation label to the consensus hashing and use a regular hash digest since consensus encoding already has canonical encoding/length encoding where applicable.

sdbondi (commented):
I feel that we should not use the length-prefixed domain hasher in this case. I imagine writing a spec for this hash where
we have to say to include the length twice e.g. `[length of bytes][[length of content][byte content]]` for every element.

It occurs to me that the DomainHasher is potentially breaking the Digest interface.
i.e 
```text
c = a||b;
H.chain(a).chain(b) != H(c)
```

base_layer/core/src/consensus/consensus_encoding/hash_writer.rs
hansieodendaal: Seems like a good idea

hansieodendaal:
Closed in favour of #4381

Issue #4365, opened 2022-07-29
Kernel signature does not commit to all kernel fields


SWvheerden:
Currently, the kernel does not commit to all kernel fields. 
We don't commit to the:

-  public excess
- Features

We also need to commit to the new burning features: https://github.com/tari-project/tari/pull/4364

SWvheerden:
PR: https://github.com/tari-project/tari/pull/4377

SWvheerden:
Fixed in https://github.com/tari-project/tari/pull/4377

PR #4377, opened 2022-08-02
feat!: fix kernel mutability

SWvheerden:
Description
---
Currently, the kernel has a mutability issue where the public excess, features, and new optional burn commitment is not committed to in the challenge.  
See issue: https://github.com/tari-project/tari/issues/4365

This is a breaking change as it changes every single kernel signature. 

This changes the challenge for the kernel to include those fields.  Because the kernel fields need to be signed by all parties, these needs fields must be decided at the start of the transaction. This required changes to the tx protocol as well. 

How Has This Been Tested?
---
Running all current unit tests and all critical cucumber tests.


base_layer/core/src/transactions/transaction_components/transaction_kernel.rs
```diff
@@ -142,6 +146,50 @@ impl TransactionKernel {
...
+        total_excess: &PublicKey,
+        fee: MicroTari,
+        lock_height: u64,
+        features: &KernelFeatures,
+        burn_commitment: &Option<Commitment>,
+    ) -> [u8; 32] {
+        let mut challenge = Challenge::new()
+            .chain(sum_public_nonces.as_bytes())
+            .chain(total_excess.as_bytes())
+            .chain(u64::from(fee).to_le_bytes())
```
stringhandler: Would be better to `chain(features.to_consensus_bytes())` kind of thing

stringhandler: I thought the features were a substruct like in the output. I suppose this is fine

stringhandler (dismissed):

integration_tests/log4rs/base_node.yml
```diff
@@ -82,7 +82,7 @@ root:
 loggers:
   # Route log events common to every application to all appenders
   tari::application:
-    level: debug
+    level: trace
```
sdbondi: Needs reverting

base_layer/core/src/transactions/transaction_components/transaction_kernel.rs
sdbondi: I would say that we should use consensus encoding and domain-separated hashing here.
Might need to wait for #4381

sdbondi: #4381 is merged

You can now write something like:
```rust
 DomainSeparatedConsensusHasher::<TransactionHashDomain>::new("kernel")
            .chain(sum_public_nonces)
            .chain(total_excess)
            .chain(fee)
            .chain(lock_height)
            .chain(features)
            // NOTE that you can (and should) pass `Option`s into consensus hashing
            .chain(burn_commitment)
            .finalize().into()

File: integration_tests/helpers/transactionBuilder.js [hunk 3]
```diff
@@ -25,14 +25,17 @@ class TransactionBuilder {
     return this.kv.private_key(id);
   }
 
-  buildChallenge(publicNonce, fee, lockHeight) {
+  buildKernelChallenge(publicNonce, publicExcess, fee, lockHeight, features) {
     const KEY = null; // optional key
     const OUTPUT_LENGTH = 32; // bytes
     const context = blake2bInit(OUTPUT_LENGTH, KEY);
     const buff = Buffer.from(publicNonce, "hex");
+    const buff2 = Buffer.from(publicExcess, "hex");
```
sdbondi: I added a DomainHasher class in JS

```javascript
let hash = new DomainHasher("com.tari.yada.yada.kernel")
  .chain(buff).chain(buff2)
// etc.
.finalize();
```

PR #4378, opened 2022-08-03
fix(key_manager): remove trailing '.' from hashing domains, fix WASM tests

sdbondi:
Description
---
- remove `tari_common` dependency from key manager as it does not support the wasm target.
- remove trailing `.` from hashing domains by using labels
- remove extra allocation in `CipherSeed::generate_mac`

Motivation and Context
---
`tari_common` includes `fs2` which does not support WASM targets causing [failed CI builds](https://github.com/tari-project/tari/runs/7645710771?check_suite_focus=true). This is currently breaking `development` branch tests.

Using a blank label appends a "." onto the domain, the current correct usage of the api is to use a label

How Has This Been Tested?
---
Existing tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7215,7 +7215,6 @@ dependencies = [
  "sha2 0.9.9",
  "strum",
  "strum_macros",
- "tari_common",
  "tari_common_types",
  "tari_crypto",
  "tari_utilities",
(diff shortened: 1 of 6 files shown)


stringhandler (approved):
Thanks

PR #4381, opened 2022-08-03
feat(base_layer/core): add domain hashing wrapper for consensus encoding

sdbondi:
Description
---
- adds `DomainSeparatedConsensusHasher` 
- renames `ConsensusHashWriter` to `ConsensusHasher` 
- removes `Write` implementation from `ConsensusHasher` 
- adds `TransactionHashDomain` hash domain 
- uses domain hashing for transaction metadata signature and script challenges

Motivation and Context
---
Creates an ergonomic and 0-alloc hasher for consensus encoded values.

Usage:
```rust
  DomainSeparatedConsensusHasher::<TransactionHashDomain>::new("metadata_signature")
            .chain(public_commitment_nonce)
            .chain(script)
            .chain(features)
            .chain(commitment)
            .chain(covenant)
            .chain(encrypted_value)
            .finalize()
```

How Has This Been Tested?
---
New unit test, existing tests use hashing


hansieodendaal:
This looks nice.

I think one of the issues with the failing integration tests is the hash calculation in 

-  `buildMetaChallenge(`
-  `buildScriptChallenge(`
-  `hashOutput(features, commitment, script, sender_offset_public_key)`

from what I could pick up. I was busy chasing some other failures after ^^ have been corrected before I closed #4373.

hansieodendaal:
Sorry, finger trouble closing this

sdbondi:
@hansieodendaal lol no problem :) Oh right, I'll need to update cucumber tests. Thanks

stringhandler (dismissed):

sdbondi:
Cucumber metadata signature is still broken, it seems to be the encrypted value is incorrect in the transactionBuilder. Going to leave that for another PR.

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -20,87 +20,93 @@
...
+    fn default() -> Self {
+        ConsensusHasher::new(HashDigest::new())
     }
 }
 
-impl<H: Update> Write for ConsensusHashWriter<H> {
-    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        self.digest.update(buf);
+/// This private struct wraps a Digest and implements the Write trait to satisfy the consensus encoding trait..
+#[derive(Clone)]
```
CjS77, line 80: Couldn't we provide a blanket `Write` impl on the underlying `DomainHasher`?

CjS77 (approved):
LGTM. See comment.

base_layer/core/src/consensus/consensus_encoding/hashing.rs
sdbondi: In this case we can't because the API for `Write` is meant to allow "chunked" writes (network sockets, files etc) and the `DomainSeparatedHasher` does not allow this.
```
c = a||b
H.chain(a).chain(b) != H.chain(c)
```

 This is the main reason why I chose to make the write interface internal to the hasher, because we know that ConseususEncoding writes each element in its entirety and our internal Write implementation never tells the caller to wait to write.

Issue #4356, opened 2022-07-28
Consolidate all sql migrations


SWvheerden:
The wallet has a lot of leftover migrations that are not used anymore. 
All of the migrations can be consolidated into single up/down files for each used table with the correct fields. 

The longer we leave this the more messier this becomes, ideally we want to only have a single set for when we do mainnet.

SWvheerden:
fixed in: https://github.com/tari-project/tari/pull/4415

PR #4384, opened 2022-08-04
chore: remove wallet key manager state migration

### LINKED ISSUE #4354: Remove wallet migration for testnet reset
The wallet has a migration to change how the key manage states are stored in the wallet. This needs to be removed before test-net reset. 
The relevant code is marked as:
`/// This is a temporary migration file to convert existing indexes to new ones.
/// Todo remove at next testnet reset (currently on Dibbler) #testnet_reset`

SWvheerden:
Description
---
Removes the wallet key manager state migration.
Removes key_manager_state_old db.

Motivation and Context
---
During the dibbler testnet, we changed how the key manager state is managed and used. This required us to recreate how the wallet stored the key manager states in the database. This required us to keep the old state and migrate over to the new storage scheme for the older wallets already running on the old one. Because we are resetting the testnet, we won't have older wallets, so we can remove the migration code.

Fixes: https://github.com/tari-project/tari/issues/4354


SWvheerden:
addresses : https://github.com/tari-project/tari/issues/4354

sdbondi (approved):

sdbondi:
Fixes https://github.com/tari-project/tari/issues/4356

PR #4386, opened 2022-08-04
fix: better coinbase handling (see issue #4353)

### LINKED ISSUE #4353: Better coinbase handling in wallet
The wallet generates new coinbases for mines when asked. This is done on the height and amount.

What the wallet does is when it is asked for a new coinbase it searches the db for a coinbase tx for that height and amount. If a match is found, that is returned. If a match is not found, it creates a new tx for that height and amount. But it also delete's all other coinbases for that height. 

This is a problem as some miner might still be mining that coinbase tx. The wallet should not just right of the bat delete all other coinbases. 

The other problem is that the miner generates an offset_private key out of random each time. This should be changed to a derived key(might use the current coinbase_pvt key) so that it is the same. 
It could happen if two miners ask for a new coinbase very close to each other that the wallet generates two coinbases only keeps one. Then the miner who won the block has its coinbases deleted, but the wallet thinks it has it due to the fact that the commitment matches.

jorgeantonio21:
Description
--- 
The current coinbase handling relies on height and amount. If a match is not found for that height and amount, then it creates a new tx, deleting all the existing coinbases for that height. This is a problem, as miners can be working concurrently on a (to-be deleted) coinbase. This PR addresses this issue (see #4353).

Motivation and Context
--- 
Fixes #4353.

How Has This Been Tested?
--- 
Unit and cucumber tests


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1010,16 +1010,17 @@ where
             None,
         )?;
 
-        // Clear any existing pending coinbase transactions for this blockheight if they exist
+        // Clear any existing pending coinbase transactions having this output hash, if they exist
         match self
             .resources
             .db
-            .clear_pending_coinbase_transaction_at_block_height(block_height)
+            .clear_pending_coinbase_transaction_with_hash(output.hash.as_slice())
```
SWvheerden: I dont think we have to clear this?
Rather than deleting the same utxo, we just see if its the same and retrieve that from the db and don't save a new one. 

Currently if its the same, it will delete the old one and replace it with an exact duplicate.

SWvheerden (changes requested):

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: We need to check this on the commitment, and not the complete output. The Rangeproof, and signatures have random nonce's in it. 
As of this PR, the commitment is deterministic though, so for each set of (fee + rewards), it will be unique.

sdbondi (commented):
Tested and works (had to fix the code a bit) - please fix the compile/clippy errors and we can merge

base_layer/wallet/src/output_manager_service/service.rs
jorgeantonio21: Changes to this were done, thanks !

sdbondi (approved):

PR #4397, opened 2022-08-05
fix: recover all known scripts

### LINKED ISSUE #4231: Only outputs with scripts `== script!(Nop)` is recover-able - can this be improved?
``` rust
            // TODO: Only outputs with scripts `== script!(Nop)` is recover-able - can this be improved?
            if output.script != script!(Nop) {
                continue;
            }
```
Is this really the case?

_Originally posted by @stringhandler in https://github.com/tari-project/tari/pull/4213#discussion_r905975950_

SWvheerden:
Description
---
This allows the wallet to recover all known scripts.

Motivation and Context
---
The wallet stores a list of all known scripts and inputs required to claim them. 
This PR allows the wallet to use this lest when searching for outputs to recover.

Fixes: https://github.com/tari-project/tari/issues/4231


sdbondi (dismissed):

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -104,14 +107,22 @@ where
...
                     output.recover_mask(&self.factories.range_proof, &self.rewind_data.rewind_blinding_key)?;
                 if output.verify_mask(&self.factories.range_proof, &blinding_factor, committed_value.into())? {
-                    let script_key = PrivateKey::random(&mut OsRng);
+                    let (input_data, script_key) = if let Some(index) = known_script_index {
+                        (
+                            known_scripts[index].input.clone(),
+                            known_scripts[index].private_key.clone(),
+                        )
+                    } else {
+                        let key = PrivateKey::random(&mut OsRng);
```
stringhandler, line 116: I'm not a fan of this. I understand it was in the previous PR, but surely there's no data needed for this? It's misleading to store a public key without storing the secret

SWvheerden: What do you mean dont store the private key?
We do, its the script_key. For a Nop, the input stack will then be the public key of that private key. 
So for Nop, which the else is, we store a random private key, as the script key(this is the key used to sign the input, aka script, aka K_s)
And then we store the input_data, that's required to "validly" claim the nop

stringhandler: ok I see it's saved in the db, I had missed that

stringhandler (approved):

PR #4415, opened 2022-08-08
fix: consolidate sql migrations (see issue #4356)

jorgeantonio21:
Description
--- 
Currenlty, wallet sql migrations have not been consolidated. This implies, that there exists many duplicate queries that are not anymore necessary. The current PR attempts to simplify the process and thus add logic to create the necessary sql tables once.

Motivation and Context
--- 
Tackle issue #4356.

How Has This Been Tested?
--- 
Own verification.


stringhandler (dismissed):
Looks good, will have to see if the tests on CI pass

sdbondi (approved):

PR #4425, opened 2022-08-09
fix(base-node): allow unused_mut for not(feature = "libtor")

sdbondi:
Description
---
Allows unused_mut for config variable if `not(all(unix, feature = "libtor"))`

Motivation and Context
---
mut is only required for `all(unix, feature = "libtor")`

How Has This Been Tested?
---
No clippy errors with/without libtor feature


hansieodendaal:
Do we need the same pattern for the validator node in `fn main_inner()` main.rs:107 ?

sdbondi:
@hansieodendaal no `config` is not used mutably in the vn main

stringhandler (approved):

PR #4431, opened 2022-08-10
fix: fix new block handling of non-tip blocks

### LINKED ISSUE #4352: Failed to handle incoming block
Base node seems to get a lot of these messages in normal operations
`00:45 ERROR Failed to handle incoming block message: CommsInterfaceError(ChainStorageError(CannotCalculateNonTipMmr("Block (#86354) previous hash is 7716f2d2327ea40cac3b1b8dd7350563438ccbf8b8a663610e92e03c60f95ba6 but the current tip is #86338 76b7c5d4826bc7c68fa6644a38768909b7c35b23882aedfad00216cc0941a450")))
00:45 ERROR Failed to handle incoming block message: CommsInterfaceError(ChainStorageError(CannotCalculateNonTipMmr("Block (#86353) previous hash is 2e860dafdb73501736ed25e79e88aafdee6903b419abea647bc0109e5b84137e but the current tip is #86338 76b7c5d4826bc7c68fa6644a38768909b7c35b23882aedfad00216cc0941a450")))
00:45 ERROR Failed to handle incoming block message: CommsInterfaceError(ChainStorageError(CannotCalculateNonTipMmr("Block (#86351) previous hash is aa8ad1ec115b36fae6e49e1274a21804218a1c690581f1cc0bdb0e3ff0758a2d but the current tip is #86338 76b7c5d4826bc7c68fa6644a38768909b7c35b23882aedfad00216cc0941a450")))
00:45 ERROR Failed to handle incoming block message: CommsInterfaceError(ChainStorageError(CannotCalculateNonTipMmr("Block (#86355) previous hash is 04a3e96ee3c9286b996e0bf8f2a1e02e0af1b25b4439bae71e65b87caec4acec but the current tip is #86338 76b7c5d4826bc7c68fa6644a38768909b7c35b23882aedfad00216cc0941a450")))`

We must revisit how incoming blocks are handled or suppress this error if it is not crucial.

SWvheerden:
Description
---
Currently, the base node cannot process a block that was received out of order (aka orphaned).
The `reconcile_block` function makes the assumption that the block is always on the current tip and tries to calculate the MMR roots of the block as a sanity check. The MMR roots can only be checked by a block that is built on the current local tip. 
This changes the `reconcile_block` function to first check orphan status, and if true, it will require the full block from the peer. 

This Pr also changes how the db lock is processed. Currently, it will print out a warn followed by an error. This suppresses the error and only prints the warn. 

How Has This Been Tested?
---
Manual

Fixes: https://github.com/tari-project/tari/issues/4352


applications/tari_console_wallet/src/main.rs
```diff
@@ -100,7 +100,7 @@ fn main_inner() -> Result<(), ExitError> {
     )?;
 
     #[cfg_attr(feature = "libtor", allow(unused_mut))]
-    let mut config = ApplicationConfig::load_from(&cfg)?;
+    let config = ApplicationConfig::load_from(&cfg)?;
```
delta1, line 103: #4425

sdbondi: #4425 merged

base_layer/core/src/base_node/service/service.rs
```diff
@@ -297,8 +297,14 @@ where B: BlockchainBackend + 'static
         task::spawn(async move {
             let result = handle_incoming_block(inbound_nch, new_block).await;
 
-            if let Err(e) = result {
-                error!(target: LOG_TARGET, "Failed to handle incoming block message: {:?}", e);
+            match result {
+                Err(BaseNodeServiceError::CommsInterfaceError(CommsInterfaceError::ChainStorageError(
+                    ChainStorageError::CannotAcquireFileLock,
```
sdbondi: This error only happens when you startup a node with the same base dir - the db lockfile is present. 
`Cannot acquire exclusive file lock, another instance of the application is already running`

Which error message are you trying to suppress?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1229,6 +1229,8 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(db: &T, block: &Block) -> Resul
 
     let metadata = db.fetch_chain_metadata()?;
     if header.prev_hash != *metadata.best_block() {
+        error!(target: LOG_TARGET, "Block hash: {}", header.hash().to_hex());
+        error!(target: LOG_TARGET, "Block : {}", block);
```
sdbondi: I think you meant to take these out / make into a more descriptive debug message

base_layer/core/src/base_node/service/service.rs
SWvheerden: good catch, wrong error.

stringhandler (commented):
I don't understand the need for this special case. Surely the mempool check will perform the same logic?

stringhandler (approved):

PR #4442, opened 2022-08-11
fix: cap unlimited rpc connection

Cifko:
Description
---
Setting -1 to rpc max count was setting value usize::MAX which is over the actual limit.

How Has This Been Tested?
---
Manually.


comms/core/src/protocol/rpc/server/mod.rs
```diff
@@ -175,7 +175,7 @@ impl RpcServerBuilder {
     }
 
     pub fn with_maximum_simultaneous_sessions(mut self, limit: usize) -> Self {
-        self.maximum_simultaneous_sessions = Some(limit);
+        self.maximum_simultaneous_sessions = if limit == usize::MAX { None } else { Some(limit) };
```
sdbondi: need to handle `usize::MAX - n`
```suggestion
              self.maximum_simultaneous_sessions = Some(cmp::min(limit,  BoundedExecutor::max_theoretical_tasks());
```

Add docs above explaining the behaviour

```rust
/// Sets the global maximum number of permitted RPC sessions. If the given number is greater than `usize::MAX >> 4`, then `usize::MAX >> 4` is used instead.
```

sdbondi (dismissed):

sdbondi (approved):

Issue #4395, opened 2022-08-05
Implement tari_crypto hashing API: mmr & wallet


hansieodendaal:
|      | Effective  GitHub hash: d25d726ed7e04a066135b80b1de9c019d928a75c |                                   |        |                                                              |                                                              |
| ---- | ------------------------------------------------------------ | --------------------------------- | ------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
| #    | Reference                                                    | Function/Area                     | Yes/No | URL                                                          | URL long                                                     |
| 18   | base_layer/mmr/src/common.rs:174                             | pub fn  hash_together<D: Digest>( | Yes    | [common.rs:174](https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/common.rs#L174) | https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/common.rs#L174 |
| 19   | base_layer/mmr/src/merkle_mountain_range.rs:137              | pub(crate) fn  hash_to_root(      | Yes    | [merkle_mountain_range.rs:137](https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/merkle_mountain_range.rs#L137) | https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/merkle_mountain_range.rs#L137 |
| 20   | base_layer/mmr/src/merkle_proof.rs:188                       | fn check_root<D:  Digest>(        | Yes    | [merkle_proof.rs:188](https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/merkle_proof.rs#L188) | https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/merkle_proof.rs#L188 |
| 21   | base_layer/mmr/src/mutable_mmr.rs:122                        | pub fn  get_merkle_root(          | Yes    | [mutable_mmr.rs:122](https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/mutable_mmr.rs#L122) | https://github.com/tari-project/tari/blob/development/base_layer/mmr/src/mutable_mmr.rs#L122 |
| 22   | base_layer/wallet/src/output_manager_service/service.rs:2725 | fn hash_secret_key(               | Yes    | [service.rs:2725](https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/output_manager_service/service.rs#L2725) | https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/output_manager_service/service.rs#L2725 |
[...]

PR #4445, opened 2022-08-11
feat!: apply hashing api to the mmr

hansieodendaal:
Description
---
Partial fulfillment for #4395:
- Added hashing API to the MMR by virtue of forcing the hashing type to be `D: Digest + DomainDigest` instead of only `D: Digest` at the lower level MMR methods and functions.
- base_layer/mmr/src/common.rs
- base_layer/mmr/src/merkle_mountain_range.rs
- base_layer/mmr/src/merkle_proof.rs
- base_layer/mmr/src/mutable_mmr.rs

Fixed cucumber function `const getTransactionOutputHash = function (output)`

Motivation and Context
---
Domain separated hashing as per the `tari_crypto` hashing API is needed for the MMR.

How Has This Been Tested?
---
- Passed unit tests
- Passed cucumber tests

### DIFF
--- a/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
@@ -35,10 +35,8 @@ use tari_common_types::types::{Commitment, RangeProofService};
 use tari_comms::{connectivity::ConnectivityRequester, peer_manager::NodeId};
 use tari_crypto::{
     commitment::HomomorphicCommitment,
-    hash::blake2::Blake256,
     tari_utilities::{hex::Hex, Hashable},
 };
-use tari_mmr::{MerkleMountainRange, MutableMmr};
 use tokio::task;
 
 use super::error::HorizonSyncError;
@@ -75,6 +73,10 @@ use crate::{
         TransactionOutput,
     },
     validation::{helpers, FinalHorizonStateValidation},
+    MutablePrunedOutputMmr,
+    PrunedKernelMmr,
+    PrunedOutputMmr,
+    PrunedWitnessMmr,
 };
 
 const LOG_TARGET: &str = "c::bn::state_machine_service::states::horizon_state_sync";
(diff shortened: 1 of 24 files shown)


base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
```diff
@@ -326,7 +331,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                     .fetch_block_accumulated_data(current_header.header().prev_hash.clone())
                     .await?;
                 let kernel_pruned_set = block_data.dissolve().0;
-                let mut kernel_mmr = MerkleMountainRange::<Blake256, _>::new(kernel_pruned_set);
+                let mut kernel_mmr = MerkleMountainRange::<BaseLayerKernelMmrHasherBlake256, _>::new(kernel_pruned_set);
```
stringhandler: I prefer this approach:
```
    let mut kernel_mmr = KernelMmr::new(kernels);
    let mut output_mmr = OutputMmr::new(outputs, deleted)?;
    let mut witness_mmr = WitnessMmr::new(range_proofs);
    let mut input_mmr = InputMmr::new(PrunedHashSet::default());
    ```
that I did in #4349

**Review by stringhandler — APPROVED** [review]

File: base_layer/mmr/benches/bench.rs [hunk 2]
```diff
@@ -34,7 +34,16 @@ mod benches {
     use blake2::Blake2b;
     use criterion::{criterion_group, BatchSize, Criterion};
     use digest::Digest;
-    use tari_mmr::MerkleMountainRange;
+    use tari_crypto::{hash::blake2::Blake256, hash_domain, hashing::DomainSeparatedHasher};
+    use tari_mmr::{Hash, MerkleMountainRange};
+
+    hash_domain!(
+        MmrBenchTestHashDomain,
+        "com.tari.tari_project.base_layer.mmr.bemches",
```
sdbondi, line 42: @hansieodendaal Just noticed a type-o here

sdbondi: Oh nm  just a test so it's not a problem :)

PR #4454, opened 2022-08-11
fix(ci): binary builds - remove tari_collectibles and lib deps for linux

leet4tari:
Description
Remove tari_collectibles from build. Also remove unnecessary Linux libraries for build.  

Motivation and Context
---

How Has This Been Tested?
Test builds in local fork all complete.


stringhandler (approved):
Thanks!

PR #4459, opened 2022-08-12
test: add ban test for dht

therustmonk:
Description
---
Adds a test to check ban login in DHT.

The test creates a `NodeId` with invalid signature and tries to offer it as a peer.

Motivation and Context
---
To cover ban logic with tests.

How Has This Been Tested?
---
CI


comms/dht/src/network_discovery/test.rs
```diff
@@ -151,6 +151,62 @@ mod state_machine {
...
+            .await;
+
+        connectivity_mock
+            .set_connectivity_status(ConnectivityStatus::Online(NUM_PEERS))
+            .await;
+        connectivity_mock.add_active_connection(connection).await;
+
+        // Checking banning logic
+        let invalid_peer = make_node_identity();
+        invalid_peer.replace_signature(make_node_identity().identity_signature_read().clone().unwrap());
```
sdbondi: Rather make a peer and set the signature - replace_signature is a little risky because the signature should be completely managed internally to NodeIdentity.

```rust
let mut peer = make_node_identity().to_peer();
peer.identity_signature = Some(invalid_peer.identity_signature_read().clone().unwrap());
```

stringhandler (approved):

Issue #4108, opened 2022-05-17
Tari Script should use Varints instead of u64's


stringhandler:
For example, CHECK_HEIGHT takes in a u64 which for the most part will be padded zeroes. It should use VarInt when serializing and deserializing

PR #4460, opened 2022-08-12
feat(tariscript)!: use varints in tari script (de)serialization

Cifko:
Description
---
Use varint for op codes that use 64bit numbers but in general will be low.

Motivation and Context
---
https://github.com/tari-project/tari/issues/4108

How Has This Been Tested?
---
Unit tests.


infrastructure/tari_script/src/op_codes.rs
```diff
@@ -623,18 +621,15 @@ mod test {
...
-            let s = &[val, 63, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3];
+            assert!(matches!(Opcode::read_next(&[val, 255]), Err(ScriptError::InvalidData)));
+            let s = &[val, 63, 1, 2, 3];
             let (opcode, rem) = Opcode::read_next(s).unwrap();
             assert_eq!(opcode, *op);
             assert_eq!(rem, &[1, 2, 3]);
             // Deserialise
             let mut arr = vec![1, 2, 3];
             op.to_bytes(&mut arr);
-            assert_eq!(&arr, &[1, 2, 3, val, 63, 0, 0, 0, 0, 0, 0, 0]);
```
stringhandler, line 637: This is weird that this passes. The op code is in the middle of the value. I was expecting 63 to occur before the value

stringhandler (changes requested):

sdbondi (dismissed):

infrastructure/tari_script/src/op_codes.rs
stringhandler: my bad, 63 is the height

stringhandler (dismissed):

stringhandler:
Just needs clippy to pass and can merge

infrastructure/tari_script/src/op_codes.rs
sdbondi: The existing test is confusing, 63 is the height value and `val` is the op code.
Should use the constants in tests and pass in a `val` and `expected_opcode` into `test_check_height`

sdbondi (approved):

Issue #4449, opened 2022-08-11
Incorrect ban reason saved here


stringhandler:
This code (find it pls) should save the ban reason as "BanReason::PeerSentInaccurateChainMetadata" (v0.35.0)

```rust 

Err(ref err @ BlockHeaderSyncError::PeerSentInaccurateChainMetadata { claimed, actual, local }) => {
                    warn!(target: LOG_TARGET, "{}", err);
                    self.ban_peer_long(node_id, BanReason::PeerCouldNotProvideStrongerChain {
                        claimed,
                        actual: actual.unwrap_or(0),
                        local,
                    })
                    .await?;
                },
```

delta1:
https://github.com/tari-project/tari/blob/6163d7baf71dbca7d21af09a204f5f6c5c515cd4/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs#L175

PR #4461, opened 2022-08-12
fix: wrong ban reason

Cifko:
Description
---
Change wrong ban.

Motivation and Context
---
https://github.com/tari-project/tari/issues/4449


sdbondi (approved):
Thanks

Issue #4413, opened 2022-08-08
Remove scary-looking user-facing errors that are not errors.


sdbondi:
Address the following error messages:

These should be reduced to info:

- `Remote peer closed rpc`
- - `ERROR Outbound messaging protocol failed for peer 0885662e727ebc0d4a80ac5cb9: IO Error: 1d407314/1: connection is closed`  
- ~~`WARN  Too many (40) connection failures, cooldown is in effect`~~

Add more TCP nodes to eliminate the following warning:
- `0% of nodes using tcp but 10% required` - No action, we should start some tcp nodes

Fix DNS to eliminate the following warning:
- ~~`WARN  Unable to check for software updates: DNS Client error: DNS Protocol error: io error: connection refused`~~ (software update DNS etc should be setup)

### Wallet log tab:

Reduce to info:
- `[wallet::assets::asset_manager] ERROR Found 0 owned outputs that contain constitution`

Investigate cause:
- `[wallet::storage::wallet] ERROR Cipher was not provided but Master Secret Key is encrypted`

PR #4462, opened 2022-08-12
chore: lower some log level messages

Cifko:
Description
---
Lower log level of some errors to info.

Motivation and Context
---
https://github.com/tari-project/tari/issues/4413

### DIFF
--- a/comms/core/src/protocol/messaging/outbound.rs
+++ b/comms/core/src/protocol/messaging/outbound.rs
@@ -102,6 +102,17 @@ impl OutboundMessaging {
                         "Outbound messaging protocol was unable to dial peer {}: {}", peer_node_id, err
                     );
                 },
+                Err(MessagingProtocolError::ConnectionClosed(err)) => {
+                    // Not sure about the metrics, but feels safer to keep on registering the error in metrics for now
+                    metrics::error_count(&peer_node_id).inc();
+                    debug!(
+                        target: LOG_TARGET,
+                        "Connection closed {}: {} {}",
+                        peer_node_id,
+                        err.kind(),
+                        err
+                    );
+                },
                 Err(err) => {
                     metrics::error_count(&peer_node_id).inc();
                     error!(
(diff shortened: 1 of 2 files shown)


comms/core/src/protocol/messaging/outbound.rs
```diff
@@ -104,7 +104,7 @@ impl OutboundMessaging {
                 },
                 Err(err) => {
                     metrics::error_count(&peer_node_id).inc();
-                    error!(
+                    info!(
```
sdbondi: There are some `io::Error`s that we want to be an error (most of the errors in `MessagingProtocolError`) and some that we don't (e.g. the remote closed the connection), but the `MessagingProtocolError::Io` case is where we want to exclude some errors.


I suggest adding a ConnectionClosed variant to MessagingProtocolError, and manually implement the conversion

```rust
impl From<io::Error> for MessagingProtocolError {
    fn from(err: Error) -> Self {
        match err.kind() {
            ErrorKind::ConnectionRefused |
            ErrorKind::ConnectionReset |
            ErrorKind::ConnectionAborted |
            ErrorKind::BrokenPipe |
            ErrorKind::UnexpectedEof => MessagingProtocolError::ConnectionClosed(err),
            _ => MessagingProtocolError::Io(err),
        }
    }
}
```

Then handle that variant here 
```rust
Err( MessagingProtocolError::ConnectionClosed(err)) => {
// Not sure about the metrics, but feels safer to keep on registering the error in metrics for now 
metrics::error_count(&peer_node_id).inc();

 debug!(
                                target: LOG_TARGET,
                                "Connection closed {}: {} {}",
                                peer_node_id,
                                err.kind(),
                                err
                            );
}
```

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -669,7 +669,7 @@ fn check_db_encryption_status(
                         WalletStorageError::InvalidEncryptionCipher
                     })?;
                 } else {
-                    error!(
+                    info!(
```
sdbondi: The message looks like it should be an error/warn. Under what conditions does it appear?

Cifko: Changed back to error. When the master key is encrypted and passphrase is not provided the wallet will fail to start. So I think error is appropriate in this case.

sdbondi (approved):

PR #4463, opened 2022-08-12
chore: remove Defunct L2 code

### LINKED ISSUE #4456: Remove unused L2 from Esmeralda
Esmeralda contains a lot of unused L2 code that we can remove.

SWvheerden:
Description
---
This removes all L2 code that is unused in the esmeralda chain. 

Fixes: https://github.com/tari-project/tari/issues/4456


applications/tari_console_wallet/src/automation/prompt.rs
```diff
@@ -25,13 +25,14 @@ use std::{io, io::Write, str::FromStr};
 use tari_utilities::hex::{Hex, HexError};
 
 use crate::automation::error::CommandError;
-
+#[allow(dead_code)]
```
sdbondi: I think either comment out the `mod prompt` in the parent module and add a comment saying that this is kept for future use _or_ add `#![allow(dead_code)]` to the top of this module with the same comment.

applications/tari_console_wallet/src/ui/components/send_tab.rs
```diff
@@ -477,14 +426,6 @@ impl<B: Backend> Component<B> for SendTab {
         self.balance.draw(f, areas[0], app_state);
         self.draw_send_form(f, areas[1], app_state);
 
-        if self.show_contacts {
```
sdbondi: I don't think this should be taken out

SWvheerden: good catch

stringhandler (approved):

PR #4466, opened 2022-08-15
refactor: split up app grpc proto files

sdbondi:
Description
---
Splits up `types.proto` into `transaction.proto`, `block.proto` and `sidechain_types.proto` 

Motivation and Context
---
`types.proto` had many somewhat unrelated types that could be better split up which makes them more maintainable.
It also makes it clear what types each protobuf service interface requires.

How Has This Been Tested?
---
Code compiles and tests pass


sdbondi:
~~Draft for now as #4463 should go in first and I'll resolve conflicts.~~ Ready

stringhandler (approved):

PR #4470, opened 2022-08-15
feat(core): add template registration sidechain features

sdbondi:
Description
---
- adds CodeTemplateRegistration output type
- adds TemplateRegistration to SideChainFeatures
- add grpc methods for registering a code template

Motivation and Context
---
Code templates need to be committed to on L1.

The `binary_url` and `repo_url` fields are not validated by base node consensus because: 
1. The rust implementation of Multiaddr does not currently support paths in http addresses e.g. `/dns4/github.com/tcp/443/http/tari-project/tari.git` is not well-formed however [the spec](https://multiformats.io/multiaddr/) says it should be.
2. Url parsing is complex and adding it to consensus code could introduce bugs (general/security/DoS)

Depends on #4466 
Depends on #4496 

How Has This Been Tested?
---
TODO: add cucumber tests for this output type


base_layer/core/src/proto/transaction.rs
```diff
@@ -427,3 +412,13 @@ impl TryFrom<Arc<Transaction>> for proto::types::Transaction {
         }
     }
 }
+
+// impl TryFrom<proto::types::Signature> for Signature {
```
SWvheerden: debug code?

base_layer/core/src/transactions/transaction_components/side_chain/sidechain_features.rs
```diff
@@ -24,26 +24,72 @@ use std::io::{Error, Read, Write};
...
+    transactions::transaction_components::CodeTemplateRegistration,
+};
 
-#[derive(Debug, Clone, Hash, PartialEq, Deserialize, Serialize, Eq)]
-pub struct SideChainFeatures {}
-
-impl SideChainFeatures {}
+#[derive(Debug, Clone, Hash, PartialEq, Deserialize, Serialize, Eq, Default)]
+pub struct SideChainFeatures {
+    pub template_registration: Option<CodeTemplateRegistration>,
```
SWvheerden: Should we not make this type an enum?

SWvheerden (commented):
Looks good

base_layer/core/src/transactions/transaction_components/side_chain/sidechain_features.rs
sdbondi: Yeah I think so - will change

stringhandler (dismissed):
seems to include a few merges from development

stringhandler (approved):

PR #4472, opened 2022-08-15
fix(core/covenants)!: update covenants to support OutputType enum

sdbondi:
Description
---
- adds new output type arg type
- adds field_eq support for filtering by output type 
- adds new test to check this case

Motivation and Context
---
- Covenants were still using the old `output_flags` field name 
- an OutputType argument literal was needed as filters like field_eq were not able to compare OutputTypes

The following syntax is supported:
`field_eq(@field::features_output_type, @output_type(Burn))`

How Has This Been Tested?
---
Additional unit tests


SWvheerden (dismissed):
LGTM

SWvheerden (approved):
LGTM

PR #4483, opened 2022-08-16
fix: fix transaction output hashing

hansieodendaal:
Description
---
 The newly introduced `class Blake256` was used and calculation of output features consensus bytes was consolidated. Some code organization improvements were done.

_**Note:** This change is also present in #4445._

Motivation and Context
---
Recent changes in the source code base required the cucumber test environment to be updated.

How Has This Been Tested?
---
When running locally: `npm test -- --tags "@critical and not @long-running and not @broken" --profile "none"`
```
26 scenarios (26 passed)
373 steps (373 passed)
18m27.251s (executing steps: 18m17.959s)
```

### DIFF
--- a/integration_tests/helpers/hashing.js
+++ b/integration_tests/helpers/hashing.js
@@ -82,7 +82,7 @@ module.exports = {
   domainHashers: {
     transactionKdf(label) {
       return new DomainHashing(
-        `com.tari.base_layer.core.transactions.v0.kdf.${label}`
+        `com.tari.base_layer.core.transactions.kdf.v0.${label}`
       );
     },
   },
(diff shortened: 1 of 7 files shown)


integration_tests/helpers/hashing.js
```diff
@@ -82,7 +82,7 @@ module.exports = {
   domainHashers: {
     transactionKdf(label) {
       return new DomainHashing(
-        `com.tari.base_layer.core.transactions.v0.kdf.${label}`
+        `com.tari.base_layer.core.transactions.kdf.v0.${label}`
```
sdbondi, line 85: Good catch, even though it's not used

sdbondi (approved):
LGTM  - thanks

PR #4485, opened 2022-08-16
fix: remove use of hardcoded value (see issue #4451)

### LINKED ISSUE #4451: Use config value here instead of hardcoded time
found in v0.35.0

```rust 
       match self.sync_peers(conn.clone()).await {
                        Ok(_) => continue,
                        Err(err @ NetworkDiscoveryError::PeerValidationError(_)) => {
                            warn!(target: LOG_TARGET, "{}. Banning peer.", err);
                            if let Err(err) = self
                                .context
                                .connectivity
                                .ban_peer_until(
                                    conn.peer_node_id().clone(),
                                    Duration::from_secs(60 * 60),
                                    err.to_string(),
                                )
                                .await
                            {
                                return err.into();
                            }
                        },
```

jorgeantonio21:
Description
--- Remove the use of hard-coded value [here](https://github.com/tari-project/tari/blob/development/comms/dht/src/network_discovery/on_connect.rs#L94).

Motivation and Context
--- 
Fixes #4451 

How Has This Been Tested?
--- Manually.


jorgeantonio21:
Fixes: #4451.

comms/dht/src/network_discovery/on_connect.rs
```diff
@@ -91,7 +91,7 @@ impl OnConnect {
                                 .connectivity
                                 .ban_peer_until(
                                     conn.peer_node_id().clone(),
-                                    Duration::from_secs(60 * 60),
+                                    Duration::from_secs(DURATION_SYNC_PEERS),
```
SWvheerden: Dont make it a const, take it from the config, there is a config settings, `ban_short_duration`

SWvheerden (changes requested):
Making a const is better practice, but its a hardcoded value

comms/dht/src/network_discovery/on_connect.rs
jorgeantonio21: Great! I add this config value. Much better now :)

jorgeantonio21:
Remove `@dan` cucumber tests from `WalletCli.feature`, see PR #4490.

SWvheerden (approved):
LGTM

sdbondi (approved):

PR #4491, opened 2022-08-17
fix: grpc inconsistent serialization of keys (see issue #4224)

### LINKED ISSUE #4224: Inconsistent serialisation of keys in gRPC
In the wallet gRPC server:

```
    async fn identify(&self, _: Request<GetIdentityRequest>) -> Result<Response<GetIdentityResponse>, Status> {
        let identity = self.wallet.comms.node_identity();
        Ok(Response::new(GetIdentityResponse {
            public_key: identity.public_key().to_string().into_bytes(),
            public_address: identity.public_address().to_string(),
            node_id: identity.node_id().to_string().into_bytes(),
        }))
    }
```

The keys are a byte array of the _hex_ representation.

In the validator node gRPC server:

```
    async fn get_identity(
        &self,
        _request: tonic::Request<rpc::GetIdentityRequest>,
    ) -> Result<tonic::Response<rpc::GetIdentityResponse>, tonic::Status> {
        let response = rpc::GetIdentityResponse {
            public_key: self.node_identity.public_key().to_vec(),
            public_address: self.node_identity.public_address().to_string(),
            node_id: self.node_identity.node_id().to_vec(),
        };
        Ok(Response::new(response))
    }
```

It's the little-endian byte representation of the 256bit integer

I didn't check any other implementations, but there should be a consistent representation across the board to prevent confusion.

Personally, I prefer a String representation (64 byte hex string), or the little-endian byte representation (32 bytes). I don't like the 64 byte `bytes` representation at all.

jorgeantonio21:
Description
--- 
We address inconsistent serialization of keys across base node and wallet grpc servers. The first uses a little endian 32-byte representation of the keys, whereas the second a 64-byte hex representation. We opt to use a hex-representation for the public key on the wallet grpc side.

Motivation and Context
--- 
Fixes #4224.

How Has This Been Tested?
--- 
Manually.


jorgeantonio21:
fixes: #4224.

applications/tari_app_grpc/proto/network.proto
```diff
@@ -95,7 +95,7 @@ message SoftwareUpdate {
 message GetIdentityRequest { }
 
 message GetIdentityResponse {
-    bytes public_key = 1;
+    string public_key = 1;
```
stringhandler: bytes is actually correct here. see the fix below rather

applications/tari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -198,9 +198,9 @@ impl wallet_server::Wallet for WalletGrpcServer {
     async fn identify(&self, _: Request<GetIdentityRequest>) -> Result<Response<GetIdentityResponse>, Status> {
         let identity = self.wallet.comms.node_identity();
         Ok(Response::new(GetIdentityResponse {
-            public_key: identity.public_key().to_string().into_bytes(),
+            public_key: identity.public_key().to_string(),
```
stringhandler:
```suggestion
            public_key: Vec::from(identity.public_key().as_bytes())
```

jorgeantonio21: Thanks ! Why not just `identity.public_key().to_vec()` ?

stringhandler (approved):
Nice!

PR #4492, opened 2022-08-17
fix!: change monero consensus encoding an update for hardfork v15

SWvheerden:
Description
---
Monero-rs has sealed their encoding in the newer version. This means we need to encode the struct using our own encoding trait and only using the encoding for the monero versions. 
Upgrade to the newest monero v15 hardfork code.

Also point towards our own monero-rs fork so we can update code when we want to.

### DIFF
--- a/base_layer/core/Cargo.toml
+++ b/base_layer/core/Cargo.toml
@@ -54,7 +54,7 @@ integer-encoding = "3.0.2"
 lmdb-zero = "0.4.4"
 log = "0.4"
 log-mdc = "0.1.0"
-monero = { git = "https://github.com/tari-project/monero-rs.git", features = ["serde"], optional = true }
+monero = { git = "https://github.com/tari-project/monero-rs.git", branch = "main" , features = ["serde"], optional = true }
 newtype-ops = "0.1.4"
 num-traits = "0.2.15"
 num-derive = "0.3.3"
--- a/base_layer/core/Cargo.toml
+++ b/base_layer/core/Cargo.toml
@@ -54,7 +54,7 @@ integer-encoding = "3.0.2"
 lmdb-zero = "0.4.4"
 log = "0.4"
 log-mdc = "0.1.0"
-monero = { version = "^0.13.0", features = ["serde_support"], optional = true }
+monero = { git = "https://github.com/tari-project/monero-rs.git", features = ["serde"], optional = true }
 newtype-ops = "0.1.4"
 num-traits = "0.2.15"
 num-derive = "0.3.3"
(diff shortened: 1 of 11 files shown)


base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -109,33 +112,38 @@ impl ByteArray for FixedByteArray {
...
     }
 }
 
-impl Encodable for FixedByteArray {
-    fn consensus_encode<E: io::Write>(&self, e: &mut E) -> Result<usize, io::Error> {
-        self.as_slice().consensus_encode(e)
+impl ConsensusEncoding for FixedByteArray {
+    fn consensus_encode<W: Write>(&self, writer: &mut W) -> Result<(), io::Error> {
+        writer.write_all(&[self.len][..])?;
+        writer.write_all(&self.elems[0..self.len()])?;
```
sdbondi, line 140: Can make this encoding the same as the slice/vec<u8> encoding 
```suggestion
        self.as_slice().consensus_decode(writer)?;
```

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -109,33 +112,38 @@ impl ByteArray for FixedByteArray {
     }
 }
 
-impl Decodable for FixedByteArray {
-    fn consensus_decode<D: io::Read>(d: &mut D) -> Result<Self, encode::Error> {
-        #[allow(clippy::cast_possible_truncation)]
-        let len = VarInt::consensus_decode(d)?.0 as usize;
+impl ConsensusDecoding for FixedByteArray {
+    fn consensus_decode<R: Read>(reader: &mut R) -> Result<Self, io::Error> {
+        let mut buf = [0u8; 1];
```
sdbondi, line 117: I think this can be simplified a bit (decodes from slice/vec<u8> encoding)

```rust
  let len = u64::consensus_decode(reader)? as usize;
        if len > MAX_ARR_SIZE {
            return Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                format!("length exceeded maximum of {}-bytes for FixedByteArray: {}", MAX_ARR_SIZE, len),
            ));
        }
        let mut buf = [0u8; MAX_ARR_SIZE];
        reader.read_exact(&mut buf[..len])?;

        Ok(FixedByteArray{
            elems: buf,
            len: len as u8,
        })
```

sdbondi (commented):
LGTM - minor comments so far. Still want to look a little closer

base_layer/core/Cargo.toml
```diff
@@ -54,7 +54,7 @@ integer-encoding = "3.0.2"
 lmdb-zero = "0.4.4"
 log = "0.4"
 log-mdc = "0.1.0"
-monero = { version = "^0.13.0", features = ["serde_support"], optional = true }
+monero = { git = "https://github.com/tari-project/monero-rs.git", features = ["serde"], optional = true }
```
stringhandler: This must have a tag or a commit specified

stringhandler (dismissed):

Issue #4399, opened 2022-08-05
base_node panics with config setting: check_interval  = 0


hansieodendaal:
**Describe the bug**

This setting in the config.toml gets the base node to panic:
```
# This interval in seconds to check for software updates. Setting this to 0 disables checking.
check_interval = 0
```

**To Reproduce**

Run the base node with config option as above.

**Expected behavior**

Should not panic.

**Screenshots**

``` rust
thread 'tokio-runtime-worker' panicked at '`period` must be non-zero.', base_layer\p2p\src\auto_update\service.rs:98:36
stack backtrace:

                                            ▄▄▄▄▄▄▄                                                                                  
  ▄▄▄  ▓██▄     ░                      ▄▒█████████████▄                       ░▐  ░▐                           ▄▄▄▄          ▄▄▄▄    
  ▀███▄ ▀▓█▓     ░                   ▄▓█████████████████▌                      ░░░░ ▒ ░                      ███████▄      ▄███████  
 ▄▄ ▀▓██▄ ▓██▄      ▓███            ▓█████████████████████▄                  ░▄▄▒░▒▄▒░░░░░                 ▐█████████▌    ██████████ 
▐▓██▄ ▀███▄▐███▄   ▐▓▓▓            ▓▌  ▄▄▄▄  ░▀▀▀  ▄▄▄▄▄  ▓░               ▒▒▓▓▀▒▀▓▓▓▒░░░░░░               ███████████   ▐███████████
  ▀▓██▌ ▀███▓████▄▄██▓▒            ▓▓ ▐██ ░░   ▄  ░░ ███░▐▓▒             ▐▒▒▓▓▓▄ ▄▓▓▓▒▒░ ░░░              ▐▐█░   █████▌  ▀██   ▀█████
  ▄ ▐▓██▌░▓███████████▓           ▐▒▓▌▐███▄▄  ▓█▌ ▄▄████ ▓▓▒            ▌▒▒▓▓▓▓▓▓▓▓▓▓▒▒▒░ ░░                     ▐████▌        ░█████
 ▀██▓▄▐███████████████▓░           ▒▓▓▓▄▄▄▄▄▄█████▄▄▄▄▄▄█▓▒▒           ██▒▒▒▒▒▒▒▒▒▒▒▒▒░▒▒                  ░   ▄▒█████▒  ░    ▒██████
[...]

PR #4495, opened 2022-08-18
fix(base_node_config): check_interval is 0 made base node is panicked

hnidoaht-101:
Description
--
This PR served for fixing the bug described in #4399 

Motivation and Context
---
My target is finding a way to change it with as little change and minimal impact as possible. So here's how I can think of it. I'm not sure if this is the best solution. Any comments are welcome.

How Has This Been Tested?
---
Build runs in local fork, with `check_interval` = 0 and w/o `check_interval` setting.

### DIFF
--- a/base_layer/p2p/src/auto_update/service.rs
+++ b/base_layer/p2p/src/auto_update/service.rs
@@ -95,9 +95,13 @@ impl SoftwareUpdaterService {
     ) {
         let mut interval_or_never = match self.config.check_interval {
             Some(interval) => {
-                let mut interval = time::interval(interval);
-                interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
-                Either::Left(wrappers::IntervalStream::new(interval))
+                if interval.is_zero() {
+                    Either::Right(stream::empty())
+                } else {
+                    let mut interval = time::interval(interval);
+                    interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+                    Either::Left(wrappers::IntervalStream::new(interval))
+                }
             },
             None => Either::Right(stream::empty()),
         };


sdbondi (commented):
Thanks for the PR!

~~This doesn't address the root bug in the auto-update component.~~ Apologies, I didn't look properly and thought the change was in the config higher up. This change does indeed address the issue and this comment changes to a nit :)

nit: `code-readability`, `needless-complexity`
~~I suggest reverting and handle the edge case in `base_layer/p2p/src/auto_update/service.rs line: 97`~~
```rust
 let mut interval_or_never = match self.config.check_interval {
     // Add this line
        Some(interval) if interval.is_zero() => Either::Right(stream::empty()),
            Some(interval) => {
                let mut interval = time::interval(interval);
                interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
                Either::Left(wrappers::IntervalStream::new(interval))
            },
            None => Either::Right(stream::empty()),
        };
```

hnidoaht-101:
> Thanks for the PR!
> 
> ~This doesn't address the root bug in the auto-update component.~ Apologies, I didn't look properly and thought the change was in the config higher up. This change does indeed address the issue and this comment changes to a nit :)
> 
> nit: `code-readability`, `needless-complexity` ~I suggest reverting and handle the edge case in `base_layer/p2p/src/auto_update/service.rs line: 97`~
> 
> ```rust
>  let mut interval_or_never = match self.config.check_interval {
>      // Add this line
>         Some(interval) if interval.is_zero() => Either::Right(stream::empty()),
>             Some(interval) => {
>                 let mut interval = time::interval(interval);
>                 interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
>                 Either::Left(wrappers::IntervalStream::new(interval))
>             },
>             None => Either::Right(stream::empty()),
>         };
> ```

Thank you. How beautiful it is! I've pushed updates as your suggestion.

sdbondi (approved):
Thanks :tada:  - utAck

PR #4507, opened 2022-08-19
feat: add validator node registration

Cifko:


applications/tari_app_grpc/proto/base_node.proto
```diff
@@ -441,3 +443,7 @@ message MempoolStatsResponse {
     uint64 unconfirmed_weight = 4;
 }
 
+message ActiveValidatorNode {
+    bytes shard_key = 1;
```
stringhandler, line 445:
```suggestion
    bytes shard_key = 1;
    uint64 from_height = 2;
    uint64 to_height = 3;
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -838,6 +847,52 @@ where B: BlockchainBackend
...
+        let x = hashes.len();
+        if x == 0 {
+            // Empty tree return 0
+            return Ok(vec![0; 32]);
+        }
+        let prev_power_of_two = (x + 1).next_power_of_two() / 2;
+        let mut over = x - prev_power_of_two;
+        let mut i = prev_power_of_two - over;
+        let mut j = i;
+        while over > 0 {
```
stringhandler: It's better to use the MerkleMountainRange and just add all  of them into it as leaves. See how TransactionInputs are created

stringhandler (commented):
A good start, we should also add a block validation rule that checks that the merkle root is correct.
We should also add a signature to the output features using the ValidatorNode Public key to prove that it is the actual VN creating the output and not someone else

stringhandler (approved):

PR #4515, opened 2022-08-22
feat: merge grpc list headers commands

### LINKED ISSUE #4372: GRPC[base node]: ListHeaders and GetHeadersByHeight do the same thing
In the BaseNode GRPC, there are two methods:
1. ListHeaders
2. GetHeaderByHeight

These two methods do the same thing, or at least, you can call ListHeaders with a single height to get one header, but GetHeadersByHeight returns extra information, such as confirmations and rewards.

I suggest removing the single call and adding the extra fields to the ListHeader call

SWvheerden:
Description
---
The GRPC server has 2 methods, `getHeaderByHeight` and `getHeaderByHeight`. 
These 2 methods are especially duplicated, this removes `getHeaderByHeight` method and adds the additional fields that this method had to `getHeaderByHeight`

Fixes: https://github.com/tari-project/tari/issues/4372

### DIFF
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -139,7 +139,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
     type GetNetworkDifficultyStream = mpsc::Receiver<Result<tari_rpc::NetworkDifficultyResponse, Status>>;
     type GetPeersStream = mpsc::Receiver<Result<tari_rpc::GetPeersResponse, Status>>;
     type GetTokensInCirculationStream = mpsc::Receiver<Result<tari_rpc::ValueAtHeightResponse, Status>>;
-    type ListHeadersStream = mpsc::Receiver<Result<tari_rpc::BlockHeader, Status>>;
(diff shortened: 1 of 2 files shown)


sdbondi (approved):

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -402,25 +402,59 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
-                let result_headers = match handler.get_headers(start..=end).await {
+                let result_data = match handler.get_blocks(start..=end).await {
                     Err(err) => {
                         warn!(target: LOG_TARGET, "Internal base node service error: {}", err);
                         return;
                     },
                     Ok(data) => {
                         if is_reversed {
-                            data.into_iter().rev().collect::<Vec<_>>()
+                            data.into_iter()
```
sdbondi, line 412: Something like this should DRY up the code
```suggestion
                            let mut data_iter = data.into_iter();
                            if is_reversed {
                            data_iter = data_iter.rev();
                            }
                            data_iter.map(|chain_block| { // etc.
```

sdbondi: Oh that wont work - this is fine for now

stringhandler (approved):

Issue #4403, opened 2022-08-05
Wallet (console and FFI) should have setting to not choose outputs that reveal the address


stringhandler:
# Problem 

Wallets currently will choose the best outputs as inputs when spending, however since a lurking base node can generate a transaction graph of inputs to outputs with relative ease, a wallet may reveal its transaction history by including a (non-stealth address) one-sided payment. 

For example, an attacker wishing to know the transaction graph of a public key `PK_Alice`  can send a one-sided payment to `PK_Alice` using the Tariscript `Push(PK_Alice)`. At some point, Alice's wallet spends this transaction without realizing it.

# Possible solution
1. Could change the wallet to have a config setting, to not include one-sided payments by default when spending

PR #4516, opened 2022-08-22
feat: console and FFI should have setting to not choose outputs that reveal the address #4403

agubarev:
Description
---
https://github.com/tari-project/tari/issues/4403
Wallet (console and FFI) should have setting to not choose outputs that reveal the address #4403

Motivation and Context
---
Problem
Wallets currently will choose the best outputs as inputs when spending, however since a lurking base node can generate a transaction graph of inputs to outputs with relative ease, a wallet may reveal its transaction history by including a (non-stealth address) one-sided payment.

For example, an attacker wishing to know the transaction graph of a public key PK_Alice can send a one-sided payment to PK_Alice using the Tariscript Push(PK_Alice). At some point, Alice's wallet spends this transaction without realizing it.

Possible solution
Could change the wallet to have a config setting, to not include one-sided payments by default when spending

How Has This Been Tested?
---


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1390,6 +1409,9 @@ where
         // Attempt to get the chain tip height
         let chain_metadata = self.base_node_service.get_chain_metadata().await?;
 
+        // Respecting the setting to not choose outputs that reveal the address
+        selection_criteria.excluding_onesided = self.resources.config.autoignore_onesided_utxos;
```
```suggestion
      if self.resources.config.autoignore_onesided_utxos {
          selection_criteria.excluding_onesided = self.resources.config.autoignore_onesided_utxos;
          }
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -200,13 +202,22 @@ impl OutputSql {
             .filter(outputs::status.eq(OutputStatus::Unspent as i32))
             .order_by(outputs::spending_priority.desc());
 
+        eprintln!(
```
sdbondi: Debug print

base_layer/wallet/src/output_manager_service/storage/output_source.rs
```diff
@@ -0,0 +1,73 @@
...
+pub enum OutputSource {
+    Unknown,
+    Coinbase,
+    Recovered,
+    Standard,
+    OneSided,
+    StealthOneSided,
+    Refund,
+    AtomicSwap,
+}
```
sdbondi: The `#[default]` attribute is now stable 
```suggestion
// The source of where the output came from
#[derive(Copy, Clone, Debug, PartialEq, Display, Default)]
pub enum OutputSource {
    Unknown,
    Coinbase,
    Recovered,
    #[default]
    Standard,
    OneSided,
    StealthOneSided,
    Refund,
    AtomicSwap,
}
```

sdbondi (commented):
LGTM - just the eprintln to remove (this will bork the console wallet)

common/config/presets/d_console_wallet.toml
```diff
@@ -61,6 +61,13 @@
 #command_send_wait_timeout = 300
 #command_send_wait_stage = "Broadcast"
 
+# Wallets currently will choose the best outputs as inputs when spending, however since a lurking base node can
+# generate a transaction graph of inputs to outputs with relative ease, a wallet may reveal its transaction
+# history by including a (non-stealth address) one-sided payment.
+# If set to `true`, then outputs received via simple one-sided transactions, won't be automatically selected as
+# further transactions, but can still be selected individually as specific outputs.
+#autoignore_onesided_utxos = true
```
SWvheerden:
```suggestion
#autoignore_onesided_utxos = false
```
Nit: We tend to put the defaults in the config file.

SWvheerden (commented):
Looks good, just one nit

common/config/presets/d_console_wallet.toml
agubarev: good point, thanks

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -151,6 +152,7 @@ where
                 &self.rewind_data,
                 None,
                 Some(proof),
+                OutputSource::Recovered,
```
sdbondi, line 155: Even when recovered, the output should be marked according to its script (one-sided, stealth, standard, etc.)

sdbondi (dismissed):
LGTM, but it looks like recovered outputs are not categorized. This can be done in another PR.

stringhandler (dismissed):

PR #4520, opened 2022-08-23
fix: re-add user agents to base node and console wallet

sdbondi:
Description
---
Adds user agent for base node and console wallet

Motivation and Context
---
User-agents were previously advertised but got removed at some point

How Has This Been Tested?
---
Manually


stringhandler (approved):
ah, my bad

PR #4526, opened 2022-08-23
test: fix tests to run standalone in wallet

stringhandler:
Allow tests to be run individually, i.e. `cargo test -p tari_wallet` in the wallet.

The main reason this didn't work was that the wallet excluded the `base_node` feature from `tari_core`. I've added it as a dev-dependency, but not sure if it breaks the libwallet build


AaronFeickert:
Tested manually, and this fix appears to work for individual wallet tests. Have not tested any other build consequences.

sdbondi:
Unfortunately it appears that rust still has the issue where it uses features from dev dependencies in release builds.
Tested with `cargo build --package tari_wallet --release` and croaring is included in the build.

I think the following works in tari_core/lib.rs:

```rust
#[cfg(feature = "base_node")]
mod domain_hash {
    use tari_crypto::{hash::blake2::Blake256, hash_domain, hashing::DomainSeparatedHasher};
    use tari_mmr::{pruned_hashset::PrunedHashSet, Hash, MerkleMountainRange, MutableMmr};

    hash_domain!(
        KernelMmrHashDomain,
        "com.tari.tari_project.base_layer.core.kernel_mmr",
 // (snip..)
}
#[cfg(feature = "base_node")]
pub use domain_hash::*;
```

AaronFeickert:
Any manual tests/processes that should be run to confirm this works as expected?

stringhandler:
@AaronFeickert I need to fix the CI jobs

PR #4534, opened 2022-08-24
chore!: reset gen block for new faucets

SWvheerden:
This resets the Esmeralda gen block with a new timestamp and new faucets
Also reduces the coinbase_lock height to we can have easier tests temporarily


sdbondi (dismissed):

SWvheerden:
This should wait for #4537

sdbondi (approved):

PR #4525, opened 2022-08-23
feat: show the banned reason if a peer was banned in the contacts

SWvheerden:
Description
---
Currently, if a peer has been banned, the constants will only show the peer is offline. 
This PR queries the peer manager for the peer to ask if the peer was banned or not. When asking the contact service in the wallet for the online status of the contact, the service will now report if the peer was banned, as well the reason for the banning.

PR #4536, opened 2022-08-25
fix(cucumber): update get_online_status function signature in cucumber

sdbondi:
Description
---
updates get_online_status call in js api

Motivation and Context
---
get_online_status returns a string ptr. #4525 

How Has This Been Tested?
---
Cucumber tests


SWvheerden (dismissed):
LGTM

stringhandler (approved):

Issue #4535, opened 2022-08-25
Block consensus encoding and decoding does not work


SWvheerden:
Add the following test the block.rs, it should work:
```
#[cfg(test)]
mod test {
    use super::*;
    use crate::blocks::genesis_block::get_genesis_block;
    use crate::consensus::{ConsensusDecoding};
    use tari_common::configuration::Network;

    #[test]
    fn block_encode_decode(){
        let block = get_genesis_block(Network::LocalNet).block().clone();
        let mut block_bytes = Vec::new();
        block.consensus_encode(&mut block_bytes).unwrap();
        let decoded_block =  Block::consensus_decode(&mut block_bytes.as_slice()).unwrap();
        assert_eq!(block, decoded_block);

    }
}
```

PR #4537, opened 2022-08-25
fix(core): fix block consensus decoding and add tests

sdbondi:
Description
---
- fix consensus decoding for transaction output
- add tests for all impls of consensus encoding

Motivation and Context
---
Transaction output consensus decoding was missing encrypted value.
Tests picked this up.

How Has This Been Tested?
---
New tests


SWvheerden (approved):
LGTM

SWvheerden:
Fixes: https://github.com/tari-project/tari/issues/4535

stringhandler (approved):

PR #4549, opened 2022-08-26
feat(comms/rpc): restrict rpc session per peer #4497

### LINKED ISSUE #4497: [comms] Enforce maximum RPC sessions per peer
Currently any node can block new RPC connections on another node by taking up the max global RPC connections (default: 1000).

We need the change the RPC to have a total maximum RPC connections and a maximum per peer.

sdbondi:
Description
---
- restrict the max number of active rpc sessions per peer

Motivation and Context
---
fixes #4497 

How Has This Been Tested?
---
Additional unit tests

### DIFF
--- a/common/config/presets/c_base_node.toml
+++ b/common/config/presets/c_base_node.toml
@@ -183,7 +183,7 @@ track_reorgs = true
 # sessions.
 #rpc_max_simultaneous_sessions = 100
 # The maximum comms RPC sessions allowed per peer (default value = 10).
-#rpc_max_sessions_per_peer = 11
+#rpc_max_sessions_per_peer = 10
 
 [base_node.p2p.transport]
 # -------------- Transport configuration --------------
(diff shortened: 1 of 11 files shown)


common/config/presets/c_base_node.toml
```diff
@@ -182,6 +182,8 @@ track_reorgs = true
 # The maximum simultaneous comms RPC sessions allowed (default value = 100). Setting this to -1 will allow unlimited
 # sessions.
 #rpc_max_simultaneous_sessions = 100
+# The maximum comms RPC sessions allowed per peer (default value = 10).
+#rpc_max_sessions_per_peer = 11
```
SWvheerden:
```suggestion
#rpc_max_sessions_per_peer = 10
```
Is there a reason why the commented-out value is not the default?

SWvheerden (dismissed):
Looks good to me

common/config/presets/c_base_node.toml
sdbondi: Thanks - just a mistake, had it as one more than the maximum RPC pool size in the wallet, but setting it at 10 should be OK.

stringhandler (approved):

PR #4529, opened 2022-08-23
WIP: fix: remove unnecessary database encryption MAC

AaronFeickert:
Description
---
[PR 4340](https://github.com/tari-project/tari/pull/4340) adds an unnecessary MAC.

Motivation and Context
---
The existing AES-GCM AEAD includes an authentication tag, but it does not bind to the domain. This PR binds the domain to the existing AEAD tag, and removes the unnecessary MAC. It also improves the tests.

How Has This Been Tested?
---
Existing tests are updated.

PR #4550, opened 2022-08-26
fix!: replace AES-GCM with XChaCha20-Poly1305

AaronFeickert:
Description
---
This replaces AES-GCM, which is used for encrypted storage, with XChaCha20-Poly1305. It gets rid of a superfluous MAC. It updates tests to account for more failure modes.

This is an updated and cleaner version of [PR 4529](https://github.com/tari-project/tari/pull/4529) (which used ChaCha20-Poly1305) that should be easier to review.

Motivation and Context
---
The work in [PR 4340](https://github.com/tari-project/tari/pull/4340) binds field data into a MAC that is included with AES-GCM ciphertext. This binding is important and useful, but is done in the wrong place; AES-GCM already includes an authentication tag that is built in to the ciphertext and parsed automatically on decryption.

This work adds the field binding directly into the authenticated encryption as associated data, which is the standard way to include public context. It means the existing additional MAC is no longer needed, since malleated field data will fail the authentication phase of decryption.

Separately, the use of AES-GCM for authenticated encryption is suboptimal. Its nonce length is short enough that the use of random nonces can be risky if enough data is encrypted under a common key; if a nonce is ever reused, an attacker can use the resulting ciphertext to forge messages on arbitrary (even unused) nonces and the same key. While this seems unlikely to occur in the use cases of interest, there really isn't any good reason to use AES-GCM unless you have hardware support for it and speed is critical.

The XChaCha20-Poly1305 authenticated encryption construction uses longer nonces, and it is considered safe to generate them randomly. While this construction has not yet been standardized (an existing draft standard has expired), it is common. Its cryptographic sibling ChaCha20-Poly1305 uses the same nonce length as AES-GCM, and while nonce reuse is "somewhat less worse" than in AES-GCM, it would still be unsafe in this context if it occurred (forgeries are limited to existing nonces).

How Has This Been Tested?
---
Existing tests pass. New tests are included.

### DIFF
--- a/base_layer/wallet/src/storage/sqlite_db/wallet.rs
+++ b/base_layer/wallet/src/storage/sqlite_db/wallet.rs
@@ -22,15 +22,16 @@
 
 use std::{
     convert::TryFrom,
+    mem::size_of,
     str::{from_utf8, FromStr},
     sync::{Arc, RwLock},
 };
 
-use aes_gcm::{aead::generic_array::GenericArray, Aes256Gcm, KeyInit};
 use argon2::{
     password_hash::{rand_core::OsRng, PasswordHash, PasswordHasher, PasswordVerifier, SaltString},
     Argon2,
 };
+use chacha20poly1305::{Key, KeyInit, Tag, XChaCha20Poly1305, XNonce};
 use diesel::{prelude::*, SqliteConnection};
 use log::*;
 use tari_common_types::chain_metadata::ChainMetadata;
@@ -55,13 +56,7 @@ use crate::{
         sqlite_db::scanned_blocks::ScannedBlockSql,
         sqlite_utilities::wallet_db_connection::WalletDbConnection,
     },
-    util::encryption::{
-        decrypt_bytes_integral_nonce,
-        encrypt_bytes_integral_nonce,
-        Encryptable,
-        AES_MAC_BYTES,
-        AES_NONCE_BYTES,
-    },
+    util::encryption::{decrypt_bytes_integral_nonce, encrypt_bytes_integral_nonce, Encryptable},
     utxo_scanner_service::service::ScannedBlock,
 };
 
@@ -71,7 +66,7 @@ const LOG_TARGET: &str = "wallet::storage::wallet";
 #[derive(Clone)]
 pub struct WalletSqliteDatabase {
     database_connection: WalletDbConnection,
-    cipher: Arc<RwLock<Option<Aes256Gcm>>>,
+    cipher: Arc<RwLock<Option<XChaCha20Poly1305>>>,
 }
 impl WalletSqliteDatabase {
     pub fn new(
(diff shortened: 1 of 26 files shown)


AaronFeickert:
Note that existing applications of ChaCha20-Poly1305 using fixed nonces aren't affected by this. We certainly could move to XChaCha20-Poly1305 for consistency (and to avoid any accidental misuse of the code later) if desired, but it doesn't change security.

base_layer/wallet/src/util/encryption.rs
```diff
@@ -50,91 +48,118 @@ pub trait Encryptable<C> {
...
+            ])
+            .collect();
+        assert!(decrypt_bytes_integral_nonce(&cipher, b"correct_domain".to_vec(), ciphertext_with_evil_tag).is_err());
+
+        // Must fail if truncated too short
         assert!(decrypt_bytes_integral_nonce(
             &cipher,
-            b"correct_domain".to_vec(),
-            ciphertext[0..ciphertext.len().saturating_sub(32)].to_vec()
+            b"correct domain".to_vec(),
```
hansieodendaal, line 161: I think we need the `_` here, as `b"correct_domain"` is used throughout the test

hansieodendaal: I think we need the `_` here, as `b"correct_domain"` is used throughout the test
```suggestion
            b"correct_domain".to_vec(),
```

base_layer/wallet/src/util/encryption.rs
```diff
@@ -50,91 +48,118 @@ pub trait Encryptable<C> {
...
+            ciphertext[0..size_of::<XNonce>()].to_vec()
```
hansieodendaal: **NIT:** Maybe easier to see that the length is different by explicitly using it in the argument?
```suggestion
            ciphertext[0..(ciphertext.len() - 1)].to_vec()
```

hansieodendaal (dismissed):
LGTM, however, the use of `drop` is a bit weird. Just some small comments in one test.

Also, `pub fn decrypt_bytes_integral_nonce(` makes a lot more sense now. -:)

AaronFeickert:
> LGTM, however, the use of `drop` is a bit weird. Just some small comments in one test.
> 
> Also, `pub fn decrypt_bytes_integral_nonce(` makes a lot more sense now. -:)

Using `drop` was to make `clippy` happy. It complains otherwise.

base_layer/wallet/src/util/encryption.rs
AaronFeickert: Yep! This was a typo.

base_layer/wallet/src/util/encryption.rs
AaronFeickert: The idea was to show failure if the ciphertext is shorter than the minimum allowed length, which is the length of a nonce plus the length of a tag. That should trigger failure differently than your suggestion, which would fail due to malleated ciphertext because of how the ciphertext is parsed.

AaronFeickert: Updated to make the intent of the test clear.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -535,7 +530,7 @@ impl WalletBackend for WalletSqliteDatabase {
         }
 
         // Now that all the decryption has been completed we can safely remove the cipher fully
-        let _ = (*current_cipher).take();
+        std::mem::drop((*current_cipher).take());
```
sdbondi, line 533: nit: IMO this is the most idiomatic/simplest way to clear `current_cipher`.
```suggestion
        *current_cipher  = None;
```

sdbondi (approved):

stringhandler (approved):

PR #4551, opened 2022-08-29
fix(cucumber): handles listHeaders response correctly

sdbondi:
Description
---
- Fixes how listHeaders GRPC response is handled (since #4515)

Motivation and Context
---
#4515 returns the header in a nested struct, this PR updates the js code to handle that correctly.

How Has This Been Tested?
---
`Pruned node should prune outputs` passes

### DIFF
--- a/integration_tests/features/support/node_steps.js
+++ b/integration_tests/features/support/node_steps.js
@@ -363,18 +363,18 @@ Then(
         5 * height /* 5 seconds per block */
       );
       const currTip = await client.getTipHeader();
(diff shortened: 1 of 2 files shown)


sdbondi:
This is partially solved in #4552, however there are still cases where the new list headers return structure is not handled - I'll update the code here as needed, which may involve undoing #4552 (as getTipHeader is changed here to return the "non-nested" header). There were also a few other places where `getTipHeader` return type wasn't changed.

stringhandler (approved):

PR #4560, opened 2022-08-29
fix: remove unused dependencies

stringhandler:
remove unused dependencies identified by `cargo udeps`


sdbondi (dismissed):

sdbondi:
Looks like some dependencies need to be added only for wasm targets

PR #4563, opened 2022-08-29
fix: potential problem with not updating the OMS database

### LINKED ISSUE #4505: Wallet shows pending incoming funds of invalid coinbase
Wallet still shows pending incoming funds from coinbase transactions that were reorged out and don't exist.

SWvheerden:
Description
---
The wallet uses two services, the OMS and TMS. The coinbase handling only happens inside of the TMS. When the TMS validates the coinbase and decides that the coinbase has been abandoned, it will update the TMS and then update the OMS. 

If the OMS update fails, then the TMS will have been updated and the OMS not, causing the two databases to be out of sync. And because the logic that determines if a coinbase has to be updated lives with the TMS,  it will never now it still needs to update the OMS. Causing the OMS to always have pending_incoming output. 

This PR updates the flow to first update the OMS so when this fails, it can at a later date try to update the TMS. If the TMS fails, it will update the TMS to make sure its correct. 

Fixes: https://github.com/tari-project/tari/issues/4505

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
@@ -461,6 +461,23 @@ where
         mined_timestamp: u64,
         num_confirmations: u64,
     ) -> Result<(), TransactionServiceProtocolError<OperationId>> {
+        // This updates the OMS first before we update the TMS. If we update the TMS first and operation fail inside of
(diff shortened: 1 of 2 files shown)


stringhandler (dismissed):
Wish there was a better way of doing this, but suppose it's the best we can do now. 

Happy in principle, just need to get tests working

sdbondi (approved):

Issue #4555, opened 2022-08-29
`Command runner exits with 'Diesel error: database is locked'


hansieodendaal:
**Edit:** Title changed from `Wallet panics with 'Diesel error: database is locked'` to `Command runner exits with 'Diesel error: database is locked'` (_The wallet did not actually panic_)

Examples:

- Invalid command. Output manager error: `Output manager storage error: `Diesel error: `database is locked```
- Invalid command. Output manager error: `Key manager service error : Storage error: `Diesel error: `database is locked```

Probable (most likely) cause: 
- spawn blocking sql db calls 

The command script file (_condensed for legibility_):

```
# Let us stress the network

get-balance

# 1 of 140
coin-split 0.01T 499
# 2 of 140
coin-split 0.01T 499
.
.
.
# 140 of 140
coin-split 0.01T 499

```

stringhandler:
Needs some more detail

hansieodendaal:
Related PRs to fix this before another test is attempted: #4564, #4575, #4591 and #4597.

hansieodendaal:
**Feedback:**
- I could perform 200 coin spilts producing 99,800 UTXOs on each of 3 wallets successfully without encountering the `Diesel error: 'database is locked'` error.
- However, trying to perform the same test again afterward using the same wallets (with the added transaction and outputs data from the previous test) resulted in the `Diesel error: 'database is locked'` error.

hansieodendaal:
We have implemented all the recommended pragmas
Our error condition is described here: https://www.sqlite.org/rescode.html#locked
#4613 may help to fix the continued error

See related discussion in [stackoverflow](https://stackoverflow.com/questions/57123453/how-to-use-diesel-with-sqlite-connections-and-avoid-database-is-locked-type-of)

hansieodendaal:
See #4615 - `Diesel error: 'database is locked'` is not a non-recoverable error but the command runner stopped prematurely.

PR #4575, opened 2022-08-30
feat: remove spawn blocking calls from wallet db (contacts service)

hansieodendaal:
Description
---
- Removed spawn blocking calls for db operations from the wallet in the contacts service. (This is another PR in a couple of PRs required to implement this fully throughout the wallet code.)
- Reset the wallet's default db connection pool size back to 16 (from 5).

Motivation and Context
---
As per https://github.com/tari-project/tari/pull/3982 and https://github.com/tari-project/tari/issues/4555

How Has This Been Tested?
---
Unit tests
Cucumber tests
System-level test

### DIFF
--- a/base_layer/wallet/src/config.rs
+++ b/base_layer/wallet/src/config.rs
@@ -136,7 +136,7 @@ impl Default for WalletConfig {
             base_node_service_config: Default::default(),
             data_dir: PathBuf::from_str("data/wallet").unwrap(),
             db_file: PathBuf::from_str("db/console_wallet.db").unwrap(),
-            db_connection_pool_size: 5, // TODO: get actual default
+            db_connection_pool_size: 16, // Note: Do not reduce this default number
             password: None,
             contacts_auto_ping_interval: Duration::from_secs(30),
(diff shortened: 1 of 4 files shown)


base_layer/wallet/src/config.rs
```diff
@@ -136,7 +136,7 @@ impl Default for WalletConfig {
             base_node_service_config: Default::default(),
             data_dir: PathBuf::from_str("data/wallet").unwrap(),
             db_file: PathBuf::from_str("db/console_wallet.db").unwrap(),
-            db_connection_pool_size: 5, // TODO: get actual default
+            db_connection_pool_size: 16, // Note: Do not reduce this default number
```
stringhandler, line 139: this is really high.  I was expecting it to be only one or two connections at a time

stringhandler (approved):

base_layer/wallet/src/config.rs
hansieodendaal: From the start we have always tested with 16; this number slipped in somewhere. Effectively it makes read operation asynchronous.

PR #4578, opened 2022-08-30
fix(outbound): reduce messaging protocol error to debug

### LINKED ISSUE #4544: Reduce this error to debug: Outbound messaging protocol failed for peer 91d330040a825190783c5041c2: IO Error: b9310d3a/3: connection is close

willyrgf:
Description
---

Closes https://github.com/tari-project/tari/issues/4544

---

Is it really just that?


comms/core/src/protocol/messaging/outbound.rs
```diff
@@ -115,7 +115,7 @@ impl OutboundMessaging {
                 },
                 Err(err) => {
                     metrics::error_count(&peer_node_id).inc();
-                    error!(
+                    debug!(
```
sdbondi: We want to log other io errors not related to the connection closing.

Take a look at `comms/core/src/protocol/messaging/error.rs`  line 66 - I believe the error kind we're missing is WriteZero (the outbound channel could not write any bytes, presumably because the connection is closed)

stringhandler: We don't want to remove all errors, just the one where it's closed

willyrgf: I think I understand now. I commit a new change [here](https://github.com/tari-project/tari/pull/4578/commits/3318373d3d837d5cfc3d7cb745825368ecb36144) to address it.

Maybe that don't make sense for `MessagingProtocolError`, but might we can add `ErrorKind::WouldBlock` on this list too.

comms/core/src/protocol/messaging/error.rs
```diff
@@ -68,6 +68,10 @@ impl From<io::Error> for MessagingProtocolError {
             ErrorKind::ConnectionReset |
             ErrorKind::ConnectionAborted |
             ErrorKind::BrokenPipe |
+            ErrorKind::WriteZero |
+            ErrorKind::NotConnected |
+            ErrorKind::AddrInUse |
+            ErrorKind::AddrNotAvailable |
```
sdbondi: I would make  `ErrorKind::AddrInUse` and `ErrorKind::AddrNotAvailable` errors (this socket is a substream so these would not be expected at all and if they occur, we want to know). We only want to catch disconnection errors that can occur in normal operation. NotConnected seems like a good addition.

comms/core/src/protocol/messaging/outbound.rs
sdbondi: WouldBlock is handled internally by the AsyncRead implementation for the substream (Yamux). We want to keep these as narrow as possible to ensure that only "normal" disconnections are debug level logs and everything else is an error.

comms/core/src/protocol/messaging/error.rs
willyrgf: Thanks for the explanation @sdbondi.
It's done.

sdbondi (dismissed):
Thanks!

sdbondi:
One last thing, we require [commits to be signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

jorgeantonio21:
LGTM !

sdbondi (approved):

PR #4594, opened 2022-08-31
fix(dht): updates to message padding

AaronFeickert:
Description
---
[PR 4362](https://github.com/tari-project/tari/pull/4362) mitigates a metadata leak whereby encrypted messages are the same length as plaintext messages due to the use of a stream cipher. This work adds more complete length checks, such that padding can fail. It also more efficiently handles the edge case where no padding is needed.

Motivation and Context
---
To avoid directly leaking the length of plaintext messages after stream cipher encryption, [PR 4362](https://github.com/tari-project/tari/pull/4362) pads such messages to a multiple of a fixed base length after first prepending the original message length using a fixed encoding.

However, the following cases do not appear to be handled by the padding and unpadding code:
- The plaintext message length exceeds the fixed encoding length
- The ciphertext message is not long enough for extraction of the fixed encoding length
- The ciphertext message is not a multiple of the base length

Further, in the case where the message length (after length prepending) is exactly a multiple of the base length, an entire base length of padding is unnecessarily applied.

This work addresses these issues. The padding process now checks that the plaintext message does not exceed the limit enforced by the length encoding; as a result, it can now return an error that propagates to the encryption function caller. The padding algorithm has been simplified and now handles the multiple-of-the-base-length edge case by correctly applying no padding. The unpadding process now checks that it can safely extract the message length, and checks that the ciphertext message is a multiple of the base length.

How Has This Been Tested?
---
No test has been added for the case where the message length exceeds the limit allowed by the encoding, as this would imply very high memory usage (or swapping) exceeding 4 GB. Existing tests pass. A new test exercises the other failure modes.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -56,7 +56,6 @@ use crate::{
 pub struct CipherKey(chacha20::Key);
 pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
 
-const LITTLE_ENDIAN_U32_SIZE_REPRESENTATION: usize = 4;
 const MESSAGE_BASE_LENGTH: usize = 6000;
 
 /// Generates a Diffie-Hellman secret `kx.G` as a `chacha20::Key` given secret scalar `k` and public key `P = x.G`.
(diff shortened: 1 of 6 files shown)


AaronFeickert:
Manually tested the case where the plaintext message is too large for the length encoding.

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
-    // obtain length of original message, assuming our code runs on 64-bits system
-    let original_message_len = u32::from_le_bytes(le_bytes) as usize;
+    // Decode the message length
+    let mut encoded_length = [0u8; size_of::<u32>()];
+    encoded_length.copy_from_slice(&message[0..size_of::<u32>()]);
+    let message_length = u32::from_le_bytes(encoded_length) as usize;
 
-    if original_message_len > message.len() {
+    // The message is too short for the decoded length
+    if message_length + size_of::<u32>() > message.len() {
```
jorgeantonio21: I would be more restrictive and request it to be of same length, otherwise we error.

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
     }
 
-    // obtain original message
-    let start = LITTLE_ENDIAN_U32_SIZE_REPRESENTATION;
-    let end = LITTLE_ENDIAN_U32_SIZE_REPRESENTATION + original_message_len;
-    let original_message = &message[start..end];
+    // Remove the padding
+    let start = size_of::<u32>();
+    let end = start + message_length;
+    let unpadded_message = &message[start..end];
```
jorgeantonio21: One could further check that the remaining message bytes where actually from the default padded values, and have not been altered. However, even if the latter is the case, if chacha20 poly authentication is successful we are guaranteed that the actual message has not been modified via the attack. Correct me, if I am wrong.

comms/dht/src/crypt.rs
AaronFeickert: There's no way to tell if that's the case, as the actual message could having trailing zeros.

jorgeantonio21:
Very nice ! LGTM :)

PS: the failing tests seem to be unrelated with the current code refactoring, and probably are due to some previous PR merge to development.

comms/dht/src/crypt.rs
AaronFeickert: You're correct on both points. I considered this, but also concluded that it's more idiomatic and correct to use the authentication from the AEAD.

AaronFeickert:
Is it worth it (or even possible) to add the plaintext-is-too-long test in a way that won't horribly clog up the CI runners?

comms/dht/src/crypt.rs
jorgeantonio21: My suggestion was to have the following condition instead

`if message_length + size_of::<u32>() != message.len()`.

 If the `message_length + size_of::<u32>() < message.len()`, we might be in the case in which the message has been reduced, possibly by an attacker, is that right ?

comms/dht/src/crypt.rs
jorgeantonio21: Amazing ! Definitely, I remember you pointed out in some previous discussion, just wanted to confirm it.

comms/dht/src/crypt.rs
AaronFeickert: The `message` parameter is the padded message. Any message with padding applied would fail that check.

AaronFeickert: Just issued a commit that renames the parameter to `padded_message` for clarity.

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
-    // obtain length of original message, assuming our code runs on 64-bits system
-    let original_message_len = u32::from_le_bytes(le_bytes) as usize;
+    // Decode the message length
+    let mut encoded_length = [0u8; size_of::<u32>()];
+    encoded_length.copy_from_slice(&padded_message[0..size_of::<u32>()]);
+    let message_length = u32::from_le_bytes(encoded_length) as usize;
 
-    if original_message_len > message.len() {
+    // The message is too short for the decoded length
+    if message_length + size_of::<u32>() > padded_message.len() {
```
sdbondi: `message_length + size_of::<u32>()` can overflow (in release mode this wraps back to 0) so we need to add some checks to this input. This was an issue in the previous implementation too.

```suggestion
    let end = message_length.checked_add(size_of::<u32>()).ok_or_else(|| DhtOutboundError::PaddingError("Unpadded message length is too large".to_string()))?
    
     if end > padded_message.len() {
```

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
-fn get_original_message_from_padded_text(message: &[u8]) -> Result<Vec<u8>, DhtOutboundError> {
-    let mut le_bytes = [0u8; 4];
-    le_bytes.copy_from_slice(&message[..LITTLE_ENDIAN_U32_SIZE_REPRESENTATION]);
+fn get_original_message_from_padded_text(padded_message: &[u8]) -> Result<Vec<u8>, DhtOutboundError> {
+    // NOTE: This function can return errors relating to message length
+    // It is important not to leak error types to an adversary, or to have timing differences
+
+    // Assert that the padded message is a multiple of the base length
+    if padded_message.is_empty() || (padded_message.len() % MESSAGE_BASE_LENGTH) != 0 {
+        return Err(DhtOutboundError::CipherError("Bad padded message length".to_string()));
```
sdbondi: I think we should add a PaddingError variant to DhtOutboundError
```suggestion
        return Err(DhtOutboundError::PaddingError("Bad padded message length".to_string()));
```

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
-    // join prepend_message_len | message | zero_padding
-    let mut output = Vec::with_capacity(output_size);
-    output.extend_from_slice(&prepend_to_message);
-    output.extend_from_slice(message);
-    output.extend(std::iter::repeat(0u8).take(output_size - n - k));
-
-    output
+fn pad_message_to_base_length_multiple(message: &[u8]) -> Result<Vec<u8>, DhtOutboundError> {
+    // We require a 32-bit length representation
+    if message.len() > (u32::max_value() as usize) {
```
sdbondi: We can include a check here that ensures that `message_length + size_of::<u32>()` does not overflow/wrap around.
```suggestion
    if message.len() > (u32::max_value()  - size_of::<u32>()) as usize {
```

comms/dht/src/crypt.rs
```diff
@@ -70,45 +69,57 @@ pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPubli
...
+        return Err(DhtOutboundError::CipherError("Message is too long".to_string()));
```
sdbondi:
```suggestion
        return Err(DhtOutboundError::PaddingError("Message is too long".to_string()));
```

sdbondi (commented):
Thanks looks good. Some fairly minor comments.

comms/dht/src/crypt.rs
CjS77: You also have [`checked_sub`](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_sub) which is more idiomatic

sdbondi (approved):

stringhandler (approved):

PR #4602, opened 2022-09-01
feat: hide Coinbases that are in the process of being mined

### LINKED ISSUE #4583: Exclude Status == Coinbase from pending incoming

### LINKED ISSUE #4584: Hide coinbases when mining

SWvheerden:
Description
---
This PR hides coinbases by default that are in the process of being mined and not yet included in a mined block. 
It also removes them from the pending incoming balance. 

Motivation and Context
---
We changed how the coinbases are handled in the wallet. Previously the wallet only had a single pending coinbase utxo. This was then shown as pending incoming and that single coinbase utxo was shown on the completed transactions screen on the UI. It was removed after it was not mined for the height or changed to MInedUnconfirmed if actually mined. 

Wallets can now keep multiple coinbases for a specific height and keep them there till they can either confirm or deny they have been mined. When running multiple miners, this can mean that you have a coinbase utxo per miner (will only happen if the actual fees change per utxo). This floods the transaction screen with many many false transactions and makes the pending incoming balance useless as all of these will count towards that amount. 

How Has This Been Tested?
---
Manual

Fixes https://github.com/tari-project/tari/issues/4584
Fixes https://github.com/tari-project/tari/issues/4583


sdbondi (dismissed):

sdbondi (approved):
Tested and works as expected

Issue #4599, opened 2022-09-01
Wallet does not use "connected" base node


SWvheerden:
There exist cases where the wallet is connected to a specific user set base_node, and shows its connected. 
But when utxo validation process starts, it uses an rpc connection to a seed node defined in the config. 

This should not happen. When a wallet is connected to a node, it should only use connections to that specific node.

sdbondi:
@SWvheerden Please confirm if this still happens or if #4610 resolves it

SWvheerden:
I think we can close this, I cant seem to validate that the wallet uses any other node if it has a node set, even if the base node is closed

PR #4610, opened 2022-09-02
fix(wallet): detect base node change during txo_validation

sdbondi:
Description
---

Interrupt txo_validation_protocol and txo_validation_task if base node is changed

Motivation and Context
---
These long-running tasks continue to run using the same base node even if it has changed. This PR checks for base node changes and interrupts the tasks at the correct points.

Other tasks may also need to be interrupted in a similar way.

Ref #4599 - this may fix this issue, but more info is needed to confirm

How Has This Been Tested?
---
Manually, changing the base node and checking that the tasks end.

### DIFF
--- a/base_layer/wallet/src/connectivity_service/mock.rs
+++ b/base_layer/wallet/src/connectivity_service/mock.rs
@@ -69,6 +69,11 @@ impl WalletConnectivityMock {
         self.base_node_watch.send(Some(base_node_peer));
     }
 
+    pub async fn base_node_changed(&mut self) -> Option<Peer> {
+        self.base_node_watch.changed().await;
+        self.base_node_watch.borrow().as_ref().cloned()
+    }
+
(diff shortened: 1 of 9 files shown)


stringhandler (dismissed):

stringhandler:
That test looks like it broke in this pr

sdbondi:
Looks like there are some problems here that need to be figured out - #4599 needs more info because it's unclear if the symptom noted in the issue is because the txo validation has begun before the base node is switched or after

sdbondi:
The base node watch is a little tricky because there is no way to say "I've seen the value"  without awaiting on `.changed()` - has_changed() doesn't cut it (i.e. it will be changed again the next time the protocol starts . So instead I poll `changed()` with the main future and interrupt if base node has changed and that seems to work better

stringhandler (approved):

PR #4612, opened 2022-09-02
fix(console_wallet): use cli.non_interactive instead of prompt to show seed words

### LINKED ISSUE #4568: console-wallet: seed words don't show up if using the --password flag
When starting a new console wallet with `--password 'xxx'` the seed words do not show

willyrgf:
Description
---

When starting a new console wallet with --password 'xxx' the seed words do not show. 

It's happen because of the property used to decide show or not the seed words its based on prompt instead of --non-interactive arg.

More detailed here: https://github.com/tari-project/tari/issues/4568

---

Closes https://github.com/tari-project/tari/issues/4568

### DIFF
--- a/applications/tari_console_wallet/src/main.rs
+++ b/applications/tari_console_wallet/src/main.rs
@@ -47,6 +47,7 @@ use tari_key_manager::cipher_seed::CipherSeed;
 #[cfg(all(unix, feature = "libtor"))]
 use tari_libtor::tor::Tor;
 use tari_shutdown::Shutdown;
+use tari_utilities::SafePassword;
 use tracing_subscriber::{layer::SubscriberExt, Registry};
 use wallet_modes::{command_mode, grpc_mode, recovery_mode, script_mode, tui_mode, WalletMode};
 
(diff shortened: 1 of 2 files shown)


stringhandler:
Looks good. Will test it locally

applications/tari_console_wallet/src/main.rs
```diff
@@ -219,6 +221,13 @@ fn main_inner() -> Result<(), ExitError> {
     result
 }
 
+fn get_password(config: &ApplicationConfig, cli: &Cli) -> Option<SafePassword> {
+    cli.password
+        .as_ref()
+        .or(config.wallet.password.as_ref())
```
jorgeantonio21, line 227: The logic was already in the code previously. But is it safe in case the user doesn't provide `passwor` via the cli, we are getting it from the `config.wallet` ?

jorgeantonio21: Previously, this logic was placed within `fn main_inner()`, file `applications/tari_console_wallet/stc/main.rs`. It might be potentially risky to use this function outside that given context.

sdbondi (approved):
Tested and works - thanks!

applications/tari_console_wallet/src/main.rs
sdbondi: Looks ok to me - not sure exactly what the concern is but since the password bytes aren't moved around the stack (only the pointer Box<[u8]> is, not the contents of the pointer), and the function is private I don't think there is any additional risk. Feel free to reopen if you disagree :)

jorgeantonio21: Perfect, thanks for the explanation @sdbondi ! :)

PR #4614, opened 2022-09-05
feat(ci): merge non-critical & long-running CI into one workflow

leet4tari:
Description:
Match cargo build args with CI
Take advantage of GHA caching
Split out Ubuntu package dependencies into a bash script for reuse
Include binaries in report artifacts

Motivation and Context:
Reduce workflows and make easier to read and re-use

How Has This Been Tested?
Run in local fork - both non-critical & long-running


sdbondi (approved):
LGTM

PR #4615, opened 2022-09-06
feat: add wallet command runner error handling

hansieodendaal:
Description
---
Added error handling to the wallet command runner so that the system can gracefully recover rather than stop executing.
This will also help to sort out the issues in #4555 as those long scripts exit prematurely.

Motivation and Context
---
Command execution stopped prematurely.

How Has This Been Tested?
---
System level tests


stringhandler (approved):
Nice, I had noticed that make-it-rain was erroring, but nothing was printed

PR #4624, opened 2022-09-07
fix: remove unused dependencies

sdbondi:
Description
---
- removes unused dependencies found by `cargo udeps`
- remove `tari_common_types` dep from tari_script


infrastructure/libtor/Cargo.toml
```diff
@@ -7,16 +7,13 @@ license = "BSD-3-Clause"
 [dependencies]
 tari_common = { path = "../../common" }
 tari_p2p = { path = "../../base_layer/p2p" }
+tari_shutdown = { version = "^0.38", path = "../shutdown"}
 
 derivative = "2.2.0"
+libtor = "46.9.0"
```
delta1, line 13: This might fail on windows

sdbondi: tari_libtor is now an optional dependency in the base node/console wallet - so is the problem when `--all-features` is used?

sdbondi: ~~I added an explicit compile failure for windows builds that use `libtor` - although I'm not sure that is a great idea as windows may be supported one day~~ reverted the commit, I think it's ok as long as someone doesn't build for windows with the libtor feature enabled

stringhandler (approved):
lol thanks. I would have eventually gotten my PR, but this is quicker

infrastructure/libtor/Cargo.toml
delta1: @sdbondi https://github.com/delta1/tari/runs/8225310476?check_suite_focus=true#step:15:853

sdbondi: Ah so the problem is that it's trying to build libtor on windows. Wonder how easy it would be to exclude that package from the build for windows e.g `cargo build --workspace --exclude tari_libtor`

delta1: yep that's why I had put it under `[target.'cfg(unix)'.dependencies]` 

:shrug:

sdbondi: nice hack :) we can do it that way, but having a crate that builds no code in it is a bit weird. I'd rather not include the crate in the build at all if it's not used, i.e. you cant compile tari_base_node and tari_console_wallet with --features libtor if using windows

cargo udeps sees it an a unused dependency, which is a little erroneous but, I'd ideally rather not compile the crate at all in the normal rust way i.e. with tari_libtor as an optional dependency

PR #4639, opened 2022-09-08
fix: replace Luhn checksum with DammSum

### LINKED ISSUE #4638: Emoji ID checksum is somewhat weak
The checksum used for emoji ID encoding is a version of the [Luhn algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm) that is [generalized](https://en.wikipedia.org/wiki/Luhn_mod_N_algorithm) to even-parity dictionaries. This algorithm cannot detect all transpositions.

Since better alternatives exist, this should be replaced.

AaronFeickert:
Description
---
This replaces the [generalized Luhn](https://en.wikipedia.org/wiki/Luhn_mod_N_algorithm) checksum algorithm for emoji ID encoding with [DammSum](https://github.com/cypherstack/dammsum), and refactors `EmojiId` to be cleaner and easier to understand.

Fixes [issue 4638](https://github.com/tari-project/tari/issues/4638).

Motivation and Context
---
Emoji IDs are encoded with a checksum that is intended to detect simple errors; these include single character substitutions and single transpositions of adjacent characters. The generalized Luhn checksum algorithm cannot detect all transpositions, though it has the benefit of being a single character.

While checksums of longer length and greater complexity can provide more comprehensive error detection, this comes at the cost of a longer overall emoji ID encoding and (often) additional structures like lookup tables.

DammSum is a simple single-character checksum algorithm based on the [Damm algorithm](https://en.wikipedia.org/wiki/Damm_algorithm) that provably detects the two desired error types in their entirety. It requires no lookup tables and uses only a small number of bitwise operations.

Other options based on Reed-Solomon or CRC designs may be considered as well as alternatives.

How Has This Been Tested?
---
Existing tests should pass. New randomized tests are added that comprehensively check failure modes and error detection.


base_layer/common_types/src/dammsum.rs
```diff
@@ -0,0 +1,210 @@
...
+
+// Fixed for a dictionary size of `2^8 == 256`
+const COEFFICIENTS: [u8; 3] = [4, 3, 1];
+
+/// Compute the DammSum checksum for an array, each of whose elements are in the range `[0, 2^8)`
+pub fn compute_checksum(data: &Vec<u8>) -> u8 {
+    let mut mask = 1u8;
+
+    // Compute the bitmask (if possible)
+    for bit in COEFFICIENTS {
```
jorgeantonio21, line 53: Why not `COEFFICIENTS.iter().map(mask += 1u8 << bit);`

base_layer/common_types/src/emoji.rs
```diff
@@ -22,22 +22,34 @@
...
+/// An emoji ID is a 33-character emoji representation of a public key that includes a checksum for safety.
+/// Each character corresponds to a byte; the first 32 bytes are an encoding of the underlying public key.
+/// The last byte is a DammSum checksum of all preceding bytes.
+///
+/// Because the emoji character set contains 256 elements, it is more compact (in character count, not necessarily
+/// in display width!) than other common encodings would provide, and is in theory easier for humans to examine.
+///
+/// An emoji ID can be instantiated either from a public key or from a string of emoji characters, and can be
+/// converted to either form as well. Checksum validation is done automatically on instantiation.
+
```
jorgeantonio21, line 72: Maybe add an example, similar to the one there was in the docs before ?

jorgeantonio21:
Great ! LGTM

base_layer/common_types/src/dammsum.rs
```diff
@@ -0,0 +1,210 @@
...
+        result ^= *digit; // add
+        let overflow = (result & (1 << 7)) != 0;
+        result <<= 1; // double
+        if overflow {
+            // reduce
+            result ^= mask;
+        }
+    }
+
+    result as u8
```
jorgeantonio21: Why do we need `result as u8` and not just `result` ? `result was of type `u8` all along.

base_layer/common_types/src/emoji.rs
AaronFeickert: Sure, I can do that. I wasn't sure if there would be additional changes, like network encoding, that would render specific examples invalid. But might as well!

base_layer/common_types/src/dammsum.rs
AaronFeickert: This is just an artifact of when `result` was a `u16` to hold the overflow. Will change.

sdbondi (dismissed):
Nice, agree with the changes to EmojiId

sdbondi (approved):

PR #4649, opened 2022-09-08
fix(core/sync): handle deadline timeouts by changing peer

### LINKED ISSUE #4648: RPC deadline should not fail sync, rather next sync peer should be selected

sdbondi:
Description
---
If the RPC deadline is reached, immediately move on to the next sync peer.

Motivation and Context
---
Fixes #4648

How Has This Been Tested?
---
Manually, header, pruned and block sync


stringhandler (approved):
Nice, can probably even be tighter with these numbers, but let's see

stringhandler:
And perhaps make the max_latency higher

PR #4656, opened 2022-09-09
fix: fee estimate

SWvheerden:
Description
---
Allows fee estimate to always return an estimated fee. 

Motivation and Context
---
Currently, when calling fee estimate and you do not have enough funds, or funds pending, the fee estimate will error. 
We do not require such a high level of accuracy when calculating a fee estimate for an external call for an estimation. 
When the wallet does not have enough funds available, the API will return the fee estimate for default with 1 input and 1 kernel. 

How Has This Been Tested?
---
unit test

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -865,7 +865,7 @@ where
                     Covenant::new().consensus_encode_exact_size(),
             );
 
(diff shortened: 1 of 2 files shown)


stringhandler (dismissed):
Yeah, think this is better

stringhandler:
Test failing though

stringhandler (approved):

PR #4657, opened 2022-09-09
fix(comms/messaging): fix possible deadlock in outbound pipeline

sdbondi:
Description
---
- Fixes possible rare deadlock when broadcasting many messages due to internal channel
- Reduce number of inbound and outbound pipeline workers
- Greatly reduce buffer size between inbound messaging and the inbound pipeline to allow for substream backpressure
- Adds "last resort" timeout in outbound pipeline

Motivation and Context
---

The outbound pipeline could deadlock when all pipeline workers are busy, and the outbound sink service is full, causing the pipeline to wait for both a free executor slot and a free slot to send on the channel

How Has This Been Tested?
---
Memorynet, Manually: wallet stress tests (2 x wallets, 2 x base nodes), checked SAF message exchange

### DIFF
--- a/base_layer/p2p/src/config.rs
+++ b/base_layer/p2p/src/config.rs
@@ -95,8 +95,6 @@ pub struct P2pConfig {
     /// The maximum number of concurrent outbound tasks allowed before back-pressure is applied to outbound messaging
     /// queue
     pub max_concurrent_outbound_tasks: usize,
-    /// The size of the buffer (channel) which holds pending outbound message requests
-    pub outbound_buffer_size: usize,
     /// Configuration for DHT
     pub dht: DhtConfig,
     /// Set to true to allow peers to provide test addresses (loopback, memory etc.). If set to false, memory
@@ -131,9 +129,8 @@ impl Default for P2pConfig {
             transport: Default::default(),
             datastore_path: PathBuf::from("peer_db"),
             peer_database_name: "peers".to_string(),
-            max_concurrent_inbound_tasks: 50,
-            max_concurrent_outbound_tasks: 100,
-            outbound_buffer_size: 100,
+            max_concurrent_inbound_tasks: 4,
+            max_concurrent_outbound_tasks: 4,
             dht: DhtConfig {
                 database_url: DbConnectionUrl::file("dht.sqlite"),
                 ..Default::default()
--- a/comms/core/src/pipeline/outbound.rs
+++ b/comms/core/src/pipeline/outbound.rs
@@ -20,11 +20,14 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{fmt::Display, time::Instant};
+use std::{
+    fmt::Display,
+    time::{Duration, Instant},
+};
 
 use futures::future::Either;
 use log::*;
-use tokio::sync::mpsc;
+use tokio::{sync::mpsc, time};
 use tower::{Service, ServiceExt};
 
 use crate::{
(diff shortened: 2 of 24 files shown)


base_layer/p2p/src/config.rs
```diff
@@ -131,9 +129,8 @@ impl Default for P2pConfig {
             transport: Default::default(),
             datastore_path: PathBuf::from("peer_db"),
             peer_database_name: "peers".to_string(),
-            max_concurrent_inbound_tasks: 50,
-            max_concurrent_outbound_tasks: 100,
-            outbound_buffer_size: 100,
+            max_concurrent_inbound_tasks: 4,
+            max_concurrent_outbound_tasks: 4,
```
SWvheerden, line 133: With the latency we seem to be getting on tor sometimes, is this perhaps not a bit too low by default?

SWvheerden (commented):
Looks good

base_layer/p2p/src/config.rs
sdbondi: This is the number of concurrent inbound/outbound tasks that the pipeline (i.e. DHT middleware) runs.
Tor latency is what it is. I think these two parts of the question don't really have anything to do with each other.
If anything, higher latency = slower message rate = concurrency is used less. 

The logs indicate that the limits of 4/4 are not hit that often, even in stress test conditions, and usage is cleared extremely quickly. One main main reason for reducing the concurrency is that if something locks up, the limit is hit quickly instead of hiding the issue for extended periods of time, possibly leading to us never discovering them at all. Another reason is to limit the busy-ness of the node under stress conditions, which under high loads, leads to peer backpressure (all senders have to reduce the rate that they send messages)

sdbondi: We may have to increase concurrency, but I propose that we keep it low and increase if there is an issue. The limits that were there were way too high (should never really be higher than the number of cores) and led to the problems mentioned as well as too much buffering of messages.

hansieodendaal, line 133: These  default values should also be updated in the 2x config files please

comms/core/src/pipeline/outbound.rs
```diff
@@ -93,16 +96,26 @@ where
...
                     self.executor
                         .spawn(async move {
                             let timer = Instant::now();
                             trace!(target: LOG_TARGET, "Start outbound pipeline {}", id);
-                            if let Err(err) = pipeline.oneshot(msg).await {
-                                error!(
-                                    target: LOG_TARGET,
-                                    "Outbound pipeline {} returned an error: '{}'", id, err
-                                );
+                            match time::timeout(Duration::from_secs(30), pipeline.oneshot(msg)).await {
```
hansieodendaal, line 103: This makes a lot of sense in the event that an outbound pipeline is locked up.

Why not duplicate this pattern for the inbound pipeline as well, as the safety net overhead is cheap we have nothing to lose? It might just highlight difficult-to-find deadlocks there as well.

base_layer/p2p/src/config.rs
hansieodendaal, line 133: These default values should also be updated in the 2x config files please

base_layer/p2p/src/config.rs
```diff
@@ -95,8 +95,6 @@ pub struct P2pConfig {
     /// The maximum number of concurrent outbound tasks allowed before back-pressure is applied to outbound messaging
     /// queue
     pub max_concurrent_outbound_tasks: usize,
-    /// The size of the buffer (channel) which holds pending outbound message requests
-    pub outbound_buffer_size: usize,
```
hansieodendaal, line 99: This reduction in config is nice, as all the `outbound_buffer_size` in the config files were a bit confusing.

hansieodendaal (commented):
Nice - please see the comments below.

comms/core/src/protocol/messaging/extension.rs
```diff
@@ -39,9 +39,9 @@ use crate::{
     runtime::task,
 };
 
-/// Buffer size for inbound messages from _all_ peers. This should be large enough to buffer quite a few incoming
-/// messages before creating backpressure on peers speaking the messaging protocol.
-pub const INBOUND_MESSAGE_BUFFER_SIZE: usize = 100;
+/// Buffer size for inbound messages from _all_ peers. Is the message consumer is slow to get through this queue,
+/// sending peers will start to experience backpressure (this is a good thing).
```
hansieodendaal, line 43: Not quite sure about this change. It feels that at least this should be configurable as we might have different roles for base nodes requiring different buffer sizes.

comms/core/src/protocol/messaging/extension.rs
```diff
@@ -39,9 +39,9 @@ use crate::{
...
+pub const INBOUND_MESSAGE_BUFFER_SIZE: usize = 10;
```
hansieodendaal, line 44: Not quite sure about this change. It feels that at least this should be configurable as we might have different roles for base nodes requiring different buffer sizes.

hansieodendaal: Otherwise LGTM

sdbondi: In my tests I had this as 1, I made it 10 because there would potentially be a small performance gain by having some buffering, however more than a small amount of buffering is pointless*. There is no reason I can think of to make this user configurable, this is completely internal to the system and should just work.

*EDIT: also is detrimental to memory usage (100 x 4mb messages = 400Mb of memory usage) and delays peer backpressure (allowing DoS attacks)

hansieodendaal: Perfect

base_layer/p2p/src/config.rs
SWvheerden: sounds good. Thanks

stringhandler (approved):
Nice

Issue #4620, opened 2022-09-06
[console wallet] Balance not updated correctly with multiple massive input transactions


hansieodendaal:
![Image](https://user-images.githubusercontent.com/39146854/188632189-26ff6022-a3bd-46e1-b52c-8c3b61a57a27.png)

hansieodendaal:
This may be related to the issue. With wallet recovery of a fairly large wallet, the balance should reflect pending incoming until the status has changed all the way from `Imported` to `FauxConfirmed`.

![Image](https://user-images.githubusercontent.com/39146854/189101503-02af4292-3b63-4f8e-9616-2e5b9937b3d0.png)

SWvheerden:
@hansieodendaal 
Did the wallet fix these issues by it self? 
Or do you still have the wallet that has this broken?

hansieodendaal:
The wallets could not fix themselves

SWvheerden:
Going to close this for now, as this is an old issue, and we made multiple fixes around this. 
We should reopen this  if we encounter this again

PR #4659, opened 2022-09-11
fix: balance not updated correctly with multiple massive input transactions #4620

agubarev:
Description
---
https://github.com/tari-project/tari/issues/4620

Motivation and Context
---
Balance not updated correctly with multiple massive input transactions #4620

How Has This Been Tested?
---
manually

### DIFF
--- a/base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
+++ b/base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
@@ -26,6 +26,7 @@ use chacha20poly1305::XChaCha20Poly1305;
 use chrono::NaiveDateTime;
 use derivative::Derivative;
 use diesel::{prelude::*, sql_query, SqliteConnection};
+use itertools::Itertools;
 use log::*;
 use tari_common_types::{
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -371,106 +372,110 @@ impl OutputSql {
...
+                        .filter(|x| {
+                            matches!(
+                                OutputStatus::try_from(x.status).unwrap(),
+                                EncumberedToBeSpent | ShortTermEncumberedToBeSpent | SpentMinedUnconfirmed
+                            )
+                        })
+                        .map(|x| MicroTari(x.amount as u64))
+                        .sum(),
+                }
+            },
```
SWvheerden, line 478: This logic here, needs to happen in SQL. 
This is a regression to how it was before we fixed the performance issues here. 
If you have many outputs, this will be slow. 
If you compare speed, SQL >> RUST. 
There should be no filtering done in RUST here.

agubarev:
another time

PR #4674, opened 2022-09-13
fix: deadlock

### LINKED ISSUE #4668: [base node] Block chain db locks up with read-write lock
Sometimes the block chain db locks up with interlock - this was observed in a big stress test with a base node serving a dedicated wallet, a SHA3 miner and an XMRIG miner.

``` rust
2022-09-13 00:40:56.936229100 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6576
2022-09-13 00:40:56.936246000 [c::cs::database] TRACE [add_block] acquired write access db lock for block #6576 in 600.00ns
2022-09-13 00:40:57.717735500 [c::cs::database] TRACE [add_block] released write access db lock for block #6576 
2022-09-13 00:58:58.408781900 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6577
2022-09-13 00:58:58.408788900 [c::cs::database] TRACE [add_block] acquired write access db lock for block #6577 in 500.00ns
2022-09-13 00:58:58.926669600 [c::cs::database] TRACE [add_block] released write access db lock for block #6577 
2022-09-13 01:06:23.420061300 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:06:23.420069100 [c::cs::database] TRACE [add_block] acquired write access db lock for block #6578 in 800.00ns
2022-09-13 01:06:23.944608200 [c::cs::database] TRACE [add_block] released write access db lock for block #6578 
2022-09-13 01:06:46.663132800 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:06:47.201773700 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6579
2022-09-13 01:08:23.542110800 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:10:13.128309400 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:19:51.308916800 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:22:17.251338900 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:24:06.834677600 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:27:47.983194000 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:28:46.756211400 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:29:13.694456600 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:44:57.777042700 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:45:40.835611500 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
2022-09-13 01:46:56.703743600 [c::cs::database] TRACE [add_block] waiting for write access to add block block #6578
```

SWvheerden:
Description
---
This fixes a deadlock in the code. 

The `RwLock` allows a mutex to exist that allows concurrent reads but blocks concurrent writes, this behaves differently depending on the OS.
On Linux systems: The lock favors reads, meaning that as long as there exists a reader lock, it will allow a new reader lock to open. This means that the system can cause starvation of writers. 
On Mac/Win the lock has equal ordering, this means that as soon as a writer queues for a lock, all additional readers will be blocked till after the writer has acquired and released its lock. 

This behavior can be dangerous if recursive locks are used, as was the case here. 

At about the same time, a block was submitted, and a template was constructed for a new miner. 
The `add_block` process requires a write lock, while the `block template` process requires a read lock. 
The `template process` was first in acquiring a lock on the read, followed shortly by the `add_block` on the blocking for a write. But the deadlock was caused after the `add_block` blocked for a write, the `block template` required an additional read_lock on the calculation of the mmr roots. And thus, the entire `block_chain db `class is deadlocked. 

Fixes: https://github.com/tari-project/tari/issues/4668

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -807,8 +807,8 @@ where B: BlockchainBackend
         if median_timestamp > header.timestamp {
             header.timestamp = median_timestamp.increase(1);
         }
-        let block = Block { header, body };
-        let (mut block, roots) = self.calculate_mmr_roots(block)?;
+        let mut block = Block { header, body };
+        let roots = calculate_mmr_roots(&*db, &block)?;
         block.header.kernel_mr = roots.kernel_mr;
         block.header.kernel_mmr_size = roots.kernel_mmr_size;
         block.header.input_mr = roots.input_mr;


sdbondi (approved):
Great find, the rust docs now (not sure since when) have an example of this

https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html

sdbondi:
Took the liberty of proving/understanding the spurious deadlock, running this on playpen sometimes deadlocks (timeout), but on my local Linux machine it always deadlocks.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c8d94e50a665a4c15d13e3b3886703d8

Same behaviour for tokio async locks. 

This only happens when obtaining a read lock more than once at the same time from the same thread.

stringhandler (approved):
Nice find. I feel like there might be more

PR #4675, opened 2022-09-13
fix: reinsert transactions from failed block

stringhandler:
Description
---
After removing transactions from the mempool from a failed validation, reinsert them to keep the valid ones

Motivation and Context
---
Currently, the implementation discards all transactions when validation fails. This is a pretty heavy approach, because the block may be incorrect in the header. There may even be an attack where a malicious user crafts a bad block and removes all transactions in the mempool.

In this approach, the transactions are reinserted into the mempool. If any of them are now invalid (e.g. double spends) they will be discarded

How Has This Been Tested?
---
existing tests, CI


sdbondi (dismissed):
Makes sense

hansieodendaal (dismissed):
LGTM - I agree with this approach

SWvheerden:
This looks fine, but it does kinda seem like what's the point in removing them in the first place. 
They are going to be re-added through the same process as when they were first added. 

This will only catch the case where the mempool failed to process a mined block, but then again I don't know if this will catch it.

comms/dht/src/outbound/message.rs
```diff
@@ -191,12 +191,13 @@ impl fmt::Display for DhtOutboundMessage {
...
             f,
-            "\n---- Outgoing message ---- \nSize: {} byte(s)\nType: {}\nPeer: {}\nHeader: {}\n{}\n----",
+            "\n---- Outgoing message ---- \nSize: {} byte(s)\nType: {}\nPeer: {}\nHeader: {}\n{}\n----\n{:?}\n",
             self.body.len(),
             self.dht_message_type,
-            self.destination_node_id,
+            self.destination,
             header_str,
             self.tag,
+            self.body
```
sdbondi, line 200: Maybe the length rather?

comms/dht/src/outbound/message_params.rs
```diff
@@ -102,6 +104,13 @@ impl SendMessageParams {
         Default::default()
     }
 
+    pub fn with_debug_info(&mut self, debug_info: String) -> &mut Self {
+        if let Some(ref mut params) = self.params {
+            params.debug_info = Some(debug_info);
+        }
```
sdbondi:
```suggestion
        self.params_mut().debug_info = Some(debug_info);
```

base_layer/core/src/mempool/service/service.rs
```diff
@@ -144,41 +143,6 @@ impl MempoolService {
...
-            debug!(
-                target: LOG_TARGET,
-                "Transaction with Message {} from peer `{}` not processed while busy with initial sync.",
-                tx_msg.dht_header.message_tag,
-                tx_msg.source_peer.node_id.short_str(),
-            );
-            return;
-        }
-        let inbound_handlers = self.inbound_handlers.clone();
-        task::spawn(async move {
```
sdbondi, line 171: No issues that I can see with not spawning. You won't have write lock contention because the only services that call mempool.insert AFAIK are the mempool service and mempool sync.

Outbound propagation uses an unbounded channel, so can never block on send.

comms/dht/src/logging_middleware.rs
```diff
@@ -86,7 +91,7 @@ where
     }
 
     fn call(&mut self, msg: R) -> Self::Future {
-        trace!(target: LOG_TARGET, "{}{}", self.prefix_msg, msg);
+        info!(target: LOG_TARGET, "{}{:?}", self.prefix_msg, msg);
```
sdbondi: Perhaps debug?

sdbondi (commented):
LGTM, not for this PR, but in favour of splitting the comms logs up, maybe DHT+messaging, rpc and the rest. 

I don't think outputting the byte contents of all outgoing messages is very useful

comms/dht/src/logging_middleware.rs
sdbondi: lol, meant debug-level logging (going by the guide of whether a user would be interested in this information)

PR #4676, opened 2022-09-14
fix(comms): simplify and remove possibility of deadlock from pipelines and substream close

sdbondi:
Description
---
- Simplify outbound pipeline by removing the [pipeline] -> [messaging] channel
- Pipe outbound messages directly to the messaging protocol instead of through the outbound pipeline
- Fix rare lockup when calling yamux control close 

Motivation and Context
---
The outbound pipeline needed to poll two channels in order to make progress, some code branches in the outbound pipeline may need to use the other channel, and if that channel is full and the number of concurrent outbound tasks are full, a deadlock will occur. This case has not been directly observed, but is technically possible so should be eliminated.

This PR removes the [pipeline] -> [messaging] channel, making the outbound pipeline only have to poll one channel. It also directly pipes `OutboundMessage`s to the messaging protocol.

EDIT: I believe I've found the root cause. The connectivity manager would rarely "lock up" causing the pipelines to lock up (both pipelines require calls to connectivity manager). I traced this in the logs and found that the last thing the connectivity manager does is resolve a tie break before locking up. This involves disconnecting one of the peer connections, and it appeared this future, extremely rarely, did not resolve. Digging deeper from there, I was able to track down a flaw in the substream close procedure, write a test that reproduces it and make a fix.

How Has This Been Tested?
---
Number of ~1000-2000tx stress tests, leaving base nodes overnight (none of these are conclusive but no issues were encountered)

### DIFF
--- a/comms/core/src/pipeline/outbound.rs
+++ b/comms/core/src/pipeline/outbound.rs
@@ -62,15 +62,17 @@ where
             // Pipeline IN received a message. Spawn a new task for the pipeline
             let num_available = self.executor.num_available();
             if let Some(max_available) = self.executor.max_available() {
-                // Only emit this message if there is any concurrent usage
-                if num_available < max_available {
-                    debug!(
-                        target: LOG_TARGET,
-                        "Outbound pipeline usage: {}/{}",
-                        max_available - num_available,
-                        max_available
-                    );
-                }
+                log!(
+                    target: LOG_TARGET,
+                    if num_available < max_available {
+                        Level::Debug
+                    } else {
+                        Level::Trace
+                    },
+                    "Outbound pipeline usage: {}/{}",
+                    max_available - num_available,
+                    max_available
+                );
             }
             let pipeline = self.config.pipeline.clone();
             let id = current_id;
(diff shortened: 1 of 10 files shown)


comms/core/src/pipeline/outbound.rs
```diff
@@ -58,111 +50,67 @@ where
...
+            if let Some(max_available) = self.executor.max_available() {
+                // Only emit this message if there is any concurrent usage
+                if num_available < max_available {
+                    debug!(
+                        target: LOG_TARGET,
+                        "Outbound pipeline usage: {}/{}",
+                        max_available - num_available,
+                        max_available
+                    );
+                }
```
hansieodendaal: Just so that we have all logs available if we want to look at trace level logs.
```suggestion
                if num_available < max_available {
                    debug!(
                        target: LOG_TARGET,
                        "Outbound pipeline usage: {}/{}",
                        max_available - num_available,
                        max_available
                    );
                } else {
                    trace!(
                        target: LOG_TARGET,
                        "Outbound pipeline usage: {}/{}",
                        max_available - num_available,
                        max_available
                    );
                }
```

hansieodendaal (dismissed):
LGTM, I like the simplified approach.

(_I am going to run a biggish stress test with this code, maybe it will solve #4637_)

stringhandler (approved):

PR #4677, opened 2022-09-14
fix: potential race condition between add_block and sync

SWvheerden:
Description
---
This fixes a potential race condition. 

It is possible for `add_block` to pass the `is_add_block_disabled()`, start doing orphan validation (which can take quite long). 
While this is happening, `sync` sets the `add_block_disabled` flag and acquires a read_lock to then do pre-processing to determine sync mode, etc. While this is busy, `add_block` asks for a write_lock. 
`Add_block` gets its write_lock before `sync` gets its write_block because of the prioritization of RWLock. 

Also moved the `if db.contains(&DbKey::BlockHash(block_hash))` before the orphan validation as this is a much cheaper operation.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -875,17 +875,11 @@ where B: BlockchainBackend
         }
 
         let new_height = block.header.height;
-        // Perform orphan block validation.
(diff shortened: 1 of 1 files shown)


stringhandler (approved):
I like this.

stringhandler:
Merging, can fix the test in a follow up pr

PR #4682, opened 2022-09-15
fix(p2p): remove DETACH flag usage

sdbondi:
Description
---

Remove DETACH flag from tor hs initialization.

Motivation and Context
---
DETACH was previously added as an attempt to improve tor connection reliability, but allowing tor to keep the hidden service registered even after the node has shut down. However, this was never really confirmed to improve reliability. Since this causes tor to fail an assertion, and since testing shows that connections still function as expected, it is removed in this PR.

 
How Has This Been Tested?
---
Stopping and starting nodes and checking the tor log does not throw an error


stringhandler (approved):
yip, agree with this

Issue #4680, opened 2022-09-15
[console wallet] UTXO validation fails in an endless loop


hansieodendaal:
The UTXO Validation Protocol repeatedly for the attached wallet - after it fails it always restarts at exactly the same point and then goes into an endless validation loop.

For this wallet, a couple of thousand UTXOs were stuck at statuses 2 (`EncumberedToBeReceived`), 3 (`EncumberedToBeSpent`), 6 (`UnspentMinedUnconfirmed`) and 9 (`SpentMinedUnconfirmed`). Investigating the issue showed that UTXO validation kept on failing in an endless loop. To further test this, all of those UTXO statuses were changed to 0 (`Unspent`) out of band and then the wallet was restarted to let UTXO validation start again and mark those UTXO as 1 (`Spent`) or 5 (`Invalid`), bt this also did not work.

_**Note:** If one recovers this wallet it holds 20,735 UTXOs, whereas the wallet under investigation has 24,252 UTXOs._

The cycle of events below:
``` rust
2022-09-15 06:58:18.271204800 [wallet::output_service::txo_validation_task] DEBUG Starting TXO validation protocol with peer bbf4b4ba6fd3bf1c57c46920bd (Id: 14752336761847096741)
2022-09-15 06:58:18.271222300 [wallet::output_service::txo_validation_task] DEBUG Checking last mined TXO to see if the base node has re-orged (Operation ID: 14752336761847096741)
2022-09-15 06:58:18.290688200 [wallet::output_service::txo_validation_task] DEBUG Last mined transaction is still in the block chain according to base node. (Operation ID: 14752336761847096741)
2022-09-15 06:58:18.310425300 [wallet::output_service::txo_validation_task] DEBUG Last mined transaction is still in the block chain according to base node (Operation ID: 14752336761847096741).
2022-09-15 06:58:18.648396000 [wallet::output_service::txo_validation_task] DEBUG Asking base node for location of 100 unconfirmed outputs by hash (Operation ID: 14752336761847096741)
2022-09-15 06:58:18.652313900 [wallet::output_service::txo_validation_task] DEBUG Base node returned 100 outputs as mined and 0 outputs as unmined (Operation ID: 14752336761847096741)
[...]

PR #4626, opened 2022-09-07
fix(wallet): queued transaction broadcast

sdbondi:
A quick and naive implementation of a queued broadcast worker,  POC not to be merged

Description
---

Motivation and Context
---

How Has This Been Tested?
---

PR #4689, opened 2022-09-16
feat: improve txo validation logic

### LINKED ISSUE #4669: Investigate wallet validation trigger
From the log file it looks the wallet is triggering transactions validation to run about every second:
[log-2022-09-12_16-31-47.txt](https://github.com/tari-project/tari/files/9553777/log-2022-09-12_16-31-47.txt)

Look for validation related to: TxId: 590546471493860469)
It looks like it validates it every second, as well as updates the database more than once per query.

### LINKED ISSUE #4670: wallet output table broken row
Somehow the utxo validation service does not update a row properly.
We need to verify that we catch all instances that happen like this and prevent it. 
```
[wallet::output_manager_service] WARN  Error completing UTXO Validation Protocol (Id: 14741583085246359368): InconsistentDataError("Output marked as mined, but mined_height or mined_in_block was empty")
```

We need to make sure updates are atomic

hansieodendaal:
Description
---
Fixed atomic mined height update error and implemented a recovery safety net for transaction output validation.

Motivation and Context
---
Transaction validation was failing in an endless loop under certain conditions. (See #4680, #4669, #4670)

Fixes https://github.com/tari-project/tari/issues/4669
Fixes https://github.com/tari-project/tari/issues/4670

How Has This Been Tested?
---
Passed unit tests.
Passed cucumber tests.
System level testing.

Co-Authored-By: SW van Heerden swvheerden@gmail.com


comms/core/examples/stress/node.rs
```diff
@@ -23,6 +23,8 @@
 use std::{convert, net::Ipv4Addr, path::Path, sync::Arc, time::Duration};
 
 use rand::rngs::OsRng;
+// 'HsFlags' used in macro
+#[allow(unused_imports)]
```
sdbondi: This has been fixed

SWvheerden:
Fixes: https://github.com/tari-project/tari/issues/4669
Fixes: https://github.com/tari-project/tari/issues/4670

sdbondi:
Fixes https://github.com/tari-project/tari/issues/4669
Fixes https://github.com/tari-project/tari/issues/4670

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -551,7 +554,27 @@ where
         let mut shutdown = self.resources.shutdown_signal.clone();
         let mut base_node_watch = self.resources.connectivity.get_current_base_node_watcher();
         let event_publisher = self.resources.event_publisher.clone();
+        let validation_in_progress = self.validation_in_progress.clone();
         tokio::spawn(async move {
+            // Note: We do not want the validation task to be queued
+            let mut _lock = match validation_in_progress.try_lock() {
+                Ok(val) => val,
+                _ => {
+                    if let Err(e) = event_publisher.send(Arc::new(OutputManagerEvent::TxoValidationInternalFailure(id)))
```
sdbondi: I think this will fail a lot because the validation task has been designed to run for multiple outputs. We should use a channel and a long running service if we want one to run at a time (like in #4626 )

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -4182,7 +4182,7 @@ pub unsafe extern "C" fn wallet_create(
     callback_faux_transaction_unconfirmed: unsafe extern "C" fn(*mut TariCompletedTransaction, u64),
     callback_transaction_send_result: unsafe extern "C" fn(c_ulonglong, *mut TariTransactionSendStatus),
     callback_transaction_cancellation: unsafe extern "C" fn(*mut TariCompletedTransaction, u64),
-    callback_txo_validation_complete: unsafe extern "C" fn(u64, bool),
+    callback_txo_validation_complete: unsafe extern "C" fn(u64, u64),
```
sdbondi, line 4190: I think the JS code will need to be updated

base_layer/wallet/src/output_manager_service/service.rs
sdbondi: But if this works then I think we see and perhaps implement that later if needed

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -551,7 +554,26 @@ where
         let mut shutdown = self.resources.shutdown_signal.clone();
         let mut base_node_watch = self.resources.connectivity.get_current_base_node_watcher();
         let event_publisher = self.resources.event_publisher.clone();
+        let validation_in_progress = self.validation_in_progress.clone();
         tokio::spawn(async move {
+            // Note: We do not want the validation task to be queued
+            let mut _lock = match validation_in_progress.try_lock() {
```
stringhandler, line 560: This is very difficult to understand the logic here, can it not be simplified?

stringhandler (commented):
It is not entirely clear how this solves the problem. I would also suggest using a semaphore instead, so that we could have X threads doing validation instead of just one. 

There are a lot of warnings spread across the changes making it unclear where the actual fix is.

stringhandler:
Reviewing again, perhaps a semaphore is not correct here. Mutex is fine, but please try simplify the code

stringhandler (approved):

PR #4692, opened 2022-09-19
fix(core): calculate validator node MR at prepare block stage

sdbondi:
Description
---
- calculates validator node MR in `calculate_merkle_roots` function
- allows tari_core and wallet to compile without base node feature
- removes validator_node_mr param from `BlockHeader::from_previous`
- removes validator_node_mr from NewBlocktemplate
- removes validator node mr validation task from async validator
- adds validator node mr validator to `check_merkle_roots`
- removes unused get_validator_mr function from blockchain db
- checks validator node mr in genesis block sanity check
- removes panic from grpc conversion code

Motivation and Context
---
Validator node MR is created and checked in a different way from other merkle roots, this PR brings that code inline with other the current merkle root code + number of minor improvements.

How Has This Been Tested?
---
Existing tests - TODO: write validator node registration tests for blockchain db and block validators

### DIFF
--- a/applications/tari_app_grpc/proto/block.proto
+++ b/applications/tari_app_grpc/proto/block.proto
@@ -62,7 +62,7 @@ message BlockHeader {
     // Sum of script offsets for all kernels in this block.
     bytes total_script_offset = 15;
     // Merkle root of validator nodes
-    bytes validator_node_merkle_root = 16;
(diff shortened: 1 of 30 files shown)


stringhandler (approved):
Took only a quick look, if it's incorrect, it'll come out in testing

PR #4694, opened 2022-09-19
feat: improve the TMS validation process

SWvheerden:
Description
---
Adds a mutex on the TMS validation process to stop multiple validations running per time.

Motivation and Context
---
Validation on the transactions is run when the Base_node state changes, which is an automated event triggered when a base_node lags, a new block is received, or the wallet connects to a new base node. The other path for this is user triggered where the user manually triggers a validation. 

Triggering more than one validation at a time is a waste of resources and network traffic as the wallet now asks the same basenode for the same data more than once.

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
@@ -214,13 +214,23 @@ where
         );
         let op_id = self.operation_id;
         while let Some(last_mined_transaction) = self.db.fetch_last_mined_transaction().for_protocol(op_id)? {
-            // we can use expect here as fetch_last_mined_transaction filters on mined_height and mined_in block is some
             let mined_height = last_mined_transaction
                 .mined_height
-                .expect("Tx validation mined height is missing");
+                .ok_or_else(|| {
+                    TransactionServiceError::ServiceError(
+                        "fetch_last_mined_transaction() should return a transaction with a mined_height".to_string(),
+                    )
+                })
+                .for_protocol(op_id)?;
             let mined_in_block_hash = last_mined_transaction
                 .mined_in_block
-                .expect("Tx validation mined in block is missing");
+                .ok_or_else(|| {
+                    TransactionServiceError::ServiceError(
+                        "fetch_last_mined_transaction() should return a transaction with a mined_in_block hash"
+                            .to_string(),
+                    )
+                })
+                .for_protocol(op_id)?;
 
             let block_at_height = self
                 .get_base_node_block_at_height(mined_height, client)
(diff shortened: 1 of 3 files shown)


base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
```diff
@@ -214,23 +214,13 @@ where
...
+            // we can use expect here as fetch_last_mined_transaction filters on mined_height and mined_in block is some
             let mined_height = last_mined_transaction
                 .mined_height
-                .ok_or_else(|| {
-                    TransactionServiceError::ServiceError(
-                        "fetch_last_mined_transaction() should return a transaction with a mined_height".to_string(),
-                    )
-                })
-                .for_protocol(op_id)?;
+                .expect("Tx validation mined height is missing");
```
hansieodendaal: I do not believe this is an improvement. In my mind the previous code that returned an error was better, but still not ideal.

I would prefer that we just log an error message and continue to the next _last mined transaction_ should we have data inconsistencies. This way the protocol will not stop and do an endless repeat should we have a single row with problematic data.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -2199,8 +2201,15 @@ where
...
-
+        let validation_in_progress = self.validation_in_progress.clone();
         let join_handle = tokio::spawn(async move {
+            let mut _lock = validation_in_progress.try_lock().map_err(|_| {
+                debug!(
+                    target: LOG_TARGET,
+                    "Transaction Validation Protocol (Id: {}) spawned while a previous protocol was busy, ignored", id
+                );
+                TransactionServiceProtocolError::new(id, TransactionServiceError::TransactionValidationInProgress)
+            })?;
```
hansieodendaal, line 2212: I agree with this safeguard.

hansieodendaal (commented):
I agree with the added safeguard to ignore concurrent requests, but have some doubts about handling the inconsistent data errors (see below). 

Any opinions, anyone?

base_layer/wallet/src/transaction_service/protocols/transaction_validation_protocol.rs
SWvheerden: I removed this as the TMS defines a mined transaction as a tx with a `mined_height` and `mined_block`. 
The above query we ask to get us a transaction with these properties explicitly asks the db for this. 

See: https://github.com/tari-project/tari/blob/92ceef998dc25392b6c76e2d4a86120eff5ebc33/base_layer/wallet/src/transaction_service/storage/sqlite_db.rs#L1069-L1075

So the only reason this can be None is if somehow the memory got corrupted between asking the db and us now getting the value.

stringhandler: this can't be an expect

stringhandler:
How does a user manually trigger a validation?

Issue #3115, opened 2021-07-20
Mempool view is different across nodes


hansieodendaal:
**Describe the bug**
With a recent stress test, the mempool view was different across nodes, resulting in many transactions not being mined.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run a 36,000 transaction stress test from 6 sending wallets to 10 receiving wallets, with each wallet connected to its own base node.

**Expected behaviour**
The mempool view should be fairly consistent across base nodes for the stress test duration, i.e. until all transactions are mined.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OSX, Windows

**Smartphone (please complete the following information):**
 - n/a

**Additional context**

hansieodendaal:
This behaviour was reproduced in recent smaller stress tests - 20,000 transactions from 2x sender wallets to 2x receiver wallets.
- Restarting base node does not help
- Restarting wallet does not help
- Repointing the wallet to a different base node gets the wallet to resubmit its missing transactions to the new base node, thus improving the mempool view across different nodes

hansieodendaal:
Some results about mempool sync behaviour during 2x consecutive stress tests. There were 2x sender and 2x receiver wallets with their own base nodes on one computer, and a base node + SHA3 miner + XMRig miner together with 3x monitor nodes on another computer. A total of 20,000 negotiated transactions were sent.

The primary setting under investigation was `[base_node.mempool] service.initial_sync_num_peers` 
All nodes continued to run throughout the test, nothing was restarted.

**`[base_node.mempool] service.initial_sync_num_peers = 2`**

Both the sender and receiver nodes had virtually the same mempool stats throughout due to the connected wallets


![Image](https://user-images.githubusercontent.com/39146854/191287124-e493938a-0901-4092-bbdc-5653eb9bdd14.png)



**`[base_node.mempool] service.initial_sync_num_peers = 5`**

In the 2nd test the rest of the network was much more in tune, which is the behaviour we want, although it does not make sense that `[base_node.mempool] service.initial_sync_num_peers` can have that influence.

![Image](https://user-images.githubusercontent.com/39146854/191287160-f46d8bb0-9132-438e-a6c4-3475335e4339.png)

hansieodendaal:
Using `v0.38.5` I performed a stress test with interactive transactions to peer wallets, 3x senders (16,000 transactions each) on one computer to 4x receivers (12,000 transactions each) on another computer (48,000 transactions in total).

[...]

PR #4706, opened 2022-09-20
fix: sync handling and increase reorg speed in mempool

SWvheerden:
Description
---
This PR updates the handling reorgs and syncs in the mempool. 

The reorg code is updated to remove unnecessary checks when handling reorgs.

The sync code is completely redone to now only listen for a sync completed status. As of now the mempool ignores the rewind event as the mempool knows this is not the final state, either the base node will forward to a new tip or reset itself to the old tip. The base node upon receiving sync completed event will re-evaluate all transactions inside of it and reset it state to the correct state. 

There will be a followup pr to add mempool sync

### DIFF
--- a/base_layer/core/src/mempool/mempool_storage.rs
+++ b/base_layer/core/src/mempool/mempool_storage.rs
@@ -191,14 +191,13 @@ impl MempoolStorage {
         // validation. This is important as invalid transactions that have not been mined yet may remain in the mempool
         // after a reorg.
         let removed_txs = self.unconfirmed_pool.drain_all_mempool_transactions();
-        // this returns all orphaned transaction, but because we know the new blocks are already added, they will still
-        // be orphaned, we can drop them here
-        let _ = self.insert_txs(removed_txs);
+        // Try to add in all the transactions again.
+        self.insert_txs(removed_txs);
         // Remove re-orged transactions from reorg  pool and re-submit them to the unconfirmed mempool
         let removed_txs = self
             .reorg_pool
             .remove_reorged_txs_and_discard_double_spends(removed_blocks, new_blocks);
-        let _ = self.insert_txs(removed_txs);
+        self.insert_txs(removed_txs);
         Ok(())
     }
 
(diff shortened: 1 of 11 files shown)


base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -186,45 +186,48 @@ impl MempoolStorage {
...
+        // transactions here as we will need to revalidate them again after the sync was done and the mempool and
+        // blockchain does not yet know how the blocks will look.
+        let removed_txs = self.unconfirmed_pool.drain_all_mempool_transactions();
+        // lets save to the reorg pool
+        self.reorg_pool.insert_all(current_tip, removed_txs);
+        Ok(())
+    }
 
+    /// After a sync event, we need to try to add in all the transaction form the reorg pool.
+    pub fn process_sync(&mut self) -> Result<(), MempoolError> {
```
stringhandler, line 235: it would be better to trigger a mempool sync here. The reorg pool might only have 5 blocks worth

SWvheerden: This begs the question if the base_node enters sync, should we not rather just dump the mempool, await listening status, and then sync the mempool?

SWvheerden: will add this in a second pr

hansieodendaal:
Helps to alleviate issues logged in #3115 as observed with system-level stress testing

base_layer/core/src/mempool/service/inbound_handlers.rs
```diff
@@ -173,13 +173,9 @@ impl MempoolInboundHandlers {
...
             },
             ValidBlockAdded(_, _) => {},
-            BlockSyncRewind(removed_blocks) => {
-                self.mempool
-                    .process_reorg(removed_blocks.iter().map(|b| b.to_arc_block()).collect(), vec![])
-                    .await?;
-            },
-            BlockSyncComplete(tip_block) => {
-                self.mempool.process_published_block(tip_block.to_arc_block()).await?;
+            BlockSyncRewind(_) => {},
```
hansieodendaal, line 176: Agreed

base_layer/core/src/mempool/service/inbound_handlers.rs
```diff
@@ -173,13 +173,9 @@ impl MempoolInboundHandlers {
...
+            BlockSyncComplete(_,_) => {
+                self.mempool.process_sync().await?;
```
hansieodendaal, line 178: Agreed

hansieodendaal (dismissed):
LGTM

I thoroughly tested this code with system-level stress testing:
- spending dust - filling up blocks with not so many transactions but each having many many inputs

ToDo: 
- normal transactions
- coin splits

sdbondi (approved):

stringhandler (approved):
Seems fine

PR #4712, opened 2022-09-21
feat: improve bn command mode timeouts

hansieodendaal:
Description
---
Assigned a base node command mode time-out based on the typical time it will take to return back to the command line for specific commands.

Motivation and Context
---
Differentiation in the time-out is needed for the various commands, for example, `check-db` and `rewind-blockchain` need a lot more time to complete than `version` or `whoami`. The specific case encountered was that the previous blanket allowed time-out of `70s` was not sufficient to perform `check-db`  or `get-mempool-stats` for a busy (stressed) block chain and mempool.

How Has This Been Tested?
---
System-level testing.

### DIFF
--- a/applications/tari_base_node/src/commands/command/mod.rs
+++ b/applications/tari_base_node/src/commands/command/mod.rs
@@ -222,14 +222,13 @@ impl CommandContext {
                 Command::HeaderStats(_) |
                 Command::SearchUtxo(_) |
                 Command::SearchKernel(_) |
-                Command::Quit(_) |
-                Command::Exit(_) => 30,
-                // Depending on the state and busyness of the mempool, these commands can take a fairly long time
-                Command::Status(_) |
-                Command::Watch(_) |
                 Command::GetMempoolStats(_) |
                 Command::GetMempoolState(_) |
-                Command::GetMempoolTx(_) => 180,
+                Command::GetMempoolTx(_) |
+                Command::Status(_) |
+                Command::Watch(_) |
+                Command::Quit(_) |
+                Command::Exit(_) => 30,
                 // These commands involve intense blockchain db operations and needs a lot of time to complete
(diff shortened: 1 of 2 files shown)


therustmonk (dismissed):

applications/tari_base_node/src/commands/command/mod.rs
```diff
@@ -193,8 +193,50 @@ impl CommandContext {
...
+                Command::SearchUtxo(_) |
+                Command::SearchKernel(_) |
+                Command::Quit(_) |
+                Command::Exit(_) => 30,
+                // Depending on the state and busyness of the mempool, these commands can take a fairly long time
+                Command::Status(_) |
+                Command::Watch(_) |
+                Command::GetMempoolStats(_) |
+                Command::GetMempoolState(_) |
+                Command::GetMempoolTx(_) => 180,
```
sdbondi: I would make the mempool and status much shorter, since they should complete quickly. It's annoying to have to wait 70 seconds (nevermind 180) for it to complete before you can use the prompt again. I'd say make it 5 seconds.

sdbondi: Ideally, we should handle a ctrl+c signal and abort the command without any timeouts

sdbondi: Ideally, we should handle a ctrl+c signal and abort the command without any timeouts

hansieodendaal: My train of thought was to at least display something. With my testing, even 180 s is sometimes not enough. Under normal circumstances, it will return quickly and by the off chance that a mempool query takes longer sufficient time is allowed.

hansieodendaal: I proposed something that could solve the problem, please see the latest commit.

sdbondi (dismissed):

sdbondi (approved):
Thanks LGTM

PR #4714, opened 2022-09-21
fix(core): use compact inputs for block propagation

sdbondi:
Description
---
- Send compact inputs when the full block is requested
- Only load full inputs when required

Motivation and Context
---
Full blocks are requested if any transactions are missing. This PR reduces the size of those full blocks.
This also improves slightly optimizes block sync by removing the need to load inputs for blocks that are being sent. 

How Has This Been Tested?
---
Modified existing rust integration test to include a single transaction, forcing reconcile block to request the full block and testing the fetching of full inputs from the local db.


base_layer/core/src/chain_storage/db_transaction.rs
```diff
@@ -202,6 +202,9 @@ impl DbTransaction {
     /// Insert a "chained" orphan block.
     /// The transaction will rollback and write will return an error if the orphan already exists.
     pub fn insert_chained_orphan(&mut self, orphan: Arc<ChainBlock>) -> &mut Self {
+        if orphan.block().body.inputs().iter().any(|b| b.is_compact()) {
+            panic!("Compact inputs are not supported in the blockchain");
```
stringhandler: Panic seems wrong here

stringhandler (commented):
Seems fine, other than the one panic

base_layer/core/src/chain_storage/db_transaction.rs
sdbondi: Whoops good catch, that was just an extra check for me while testing that I forgot to take out (co-pilot wrote the text :))

stringhandler (approved):

PR #4724, opened 2022-09-22
fix(comms/rpc): increase max frame size limit for rpc requests

sdbondi:
Description
---
Increases max frame size for RPC from 2MiB to 3MiB

Motivation and Context
---
A wallet may need submit transactions greater than 2MiB in size

How Has This Been Tested?
---
Existing tests pass

### DIFF
--- a/comms/core/src/protocol/rpc/mod.rs
+++ b/comms/core/src/protocol/rpc/mod.rs
@@ -30,7 +30,7 @@ mod test;
 
 /// Maximum frame size of each RPC message. This is enforced in tokio's length delimited codec.
 /// This can be thought of as the hard limit on message size.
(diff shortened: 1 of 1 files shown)


hansieodendaal (commented):
I think we should ensure that a negotiated & finalized transaction can always be submitted to a base node.

For example, a unit test can verify max transaction constants (`MAX_TRANSACTION_INPUTS`, `MAX_TRANSACTION_OUTPUTS`) to calculate an approximate maximum transaction size (using normal transaction inputs and outputs) and compare it to RPC_MAX_FRAME_SIZE.

hansieodendaal (approved):
LGTM

_(I made an issue for the unit test so we do not forget, not part of this PR)_

stringhandler (approved):

PR #4726, opened 2022-09-22
feat(core): store and fetch templates from lmdb

mrnaveira:
Description
---
* Created a new `lmdb` database for template registrations, with methods for inserting and fetching them.
* Base layer stores in the database all templates that appear in new blocks.
* New gRPC method `GetTemplateRegistrations` to retrieve all new templates since a specific block height

Motivation and Context
---
In [previous work](https://github.com/tari-project/tari/pull/4470) we added template registration to UTXO sidechain features.

The next step is for the base layer to store and index all template registrations that appear in blocks, as well as provide a gRPC method to retrieve them.

As the main query for templates will come from the base layer scanner in the Validator Node, the gRPC query method should allow to filter all new templates since a specific block height.

How Has This Been Tested?
---
Manually ran the base layer and perform a gRPC query via Postman.


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2495,6 +2533,17 @@ impl BlockchainBackend for LMDBDatabase {
                 value: public_key.to_hex(),
             })
     }
+
+    fn fetch_template_registrations(&self, from_height: u64) -> Result<Vec<TemplateRegistration>, ChainStorageError> {
+        let txn = self.read_transaction()?;
+        lmdb_filter_map_values(&txn, &self.template_registrations, |tr: TemplateRegistration| {
```
sdbondi, line 2583: We can optimise this query by making using a compound key <reg_height, hash> - but we can optimise this later

sdbondi (dismissed):

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
mrnaveira: Agree, I put a comment for future improvement

stringhandler (approved):

PR #4730, opened 2022-09-23
feat: trigger mempool sync on lag

SWvheerden:
Description
---
Adds in the ability to retrigger the mempool sync if a larger re_org or block sync occurred. 

Motivation and Context
---
As of #4706 the mempool will now properly handle re-orgs and block syncs. 
But when the base_node is not at the "correct" tip its mempool will reject all invalid transactions. In order to get those transactions back, we need to trigger a mempool sync. 
This adds in a new config setting to control the maximum number of blocks the base_node added during a reorg or sync before it starts a sync process. 
The assumption is that in most cases, the blocks removed will be less or the same as the blocks added, so we only need to look at the blocks added. 

Open Questions
---
Do we just sync to one peer, or do we sync to `initial_sync_num_peers` peer?

Prerequisite
----
Requires: #4706

### DIFF
--- a/base_layer/core/src/mempool/sync_protocol/initializer.rs
+++ b/base_layer/core/src/mempool/sync_protocol/initializer.rs
@@ -34,7 +34,7 @@ use tokio::{sync::mpsc, time::sleep};
 use crate::{
     base_node::{comms_interface::LocalNodeCommsInterface, StateMachineHandle},
     mempool::{
-        sync_protocol::{MempoolSyncProtocol,  MEMPOOL_SYNC_PROTOCOL},
+        sync_protocol::{MempoolSyncProtocol, MEMPOOL_SYNC_PROTOCOL},
         Mempool,
         MempoolServiceConfig,
     },
@@ -102,7 +102,7 @@ impl ServiceInitializer for MempoolSyncInitializer {
                 }
                 log_mdc::extend(mdc.clone());
             }
-           let base_node_events = base_node.get_block_event_stream();
+            let base_node_events = base_node.get_block_event_stream();
 
             MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity, base_node_events)
                 .run()
--- a/base_layer/core/src/mempool/sync_protocol/initializer.rs
+++ b/base_layer/core/src/mempool/sync_protocol/initializer.rs
@@ -32,9 +32,9 @@ use tari_service_framework::{async_trait, ServiceInitializationError, ServiceIni
 use tokio::{sync::mpsc, time::sleep};
 
 use crate::{
-    base_node::StateMachineHandle,
+    base_node::{comms_interface::LocalNodeCommsInterface, StateMachineHandle},
     mempool::{
-        sync_protocol::{MempoolSyncProtocol, MEMPOOL_SYNC_PROTOCOL},
+        sync_protocol::{MempoolSyncProtocol,  MEMPOOL_SYNC_PROTOCOL},
         Mempool,
         MempoolServiceConfig,
     },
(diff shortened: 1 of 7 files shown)


base_layer/core/src/mempool/sync_protocol/initializer.rs
```diff
@@ -103,9 +107,9 @@ impl ServiceInitializer for MempoolSyncInitializer {
                 }
                 log_mdc::extend(mdc.clone());
             }
-
-            MempoolSyncProtocol::new(config, notif_rx, connectivity_event_subscription, mempool)
-                .run()
+            let block_event_stream = base_node.get_block_event_stream();
+            MempoolSyncProtocol::new(config, notif_rx, connectivity_event_subscription, mempool, connectivity)
```
sdbondi: I would get rid of the events receiver param and get it inside the run function now that we're passing in connectivity

```suggestion
            MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity)
```

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -174,6 +182,55 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static
...
+                }
+            },
+            _ => {
+                return;
+            },
+        }
+        // we need to make sure the service can start a sync
+        if self.num_synched.load(Ordering::SeqCst) >= self.config.initial_sync_num_peers {
+            self.num_synched.fetch_sub(1, Ordering::SeqCst);
+        }
```
sdbondi: nit: slightly better use of atomic ordering
```suggestion
        if self.num_synched.load(Ordering::Aquire) >= self.config.initial_sync_num_peers {
            self.num_synched.fetch_sub(1, Ordering::Release);
        }
```

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -174,6 +182,55 @@ where TSubstream: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static
         }
     }
 
+    async fn spawn_handle_block_event(&mut self, block_event: &BlockEvent) {
```
sdbondi: total nit :P - I would say this function just handles a block event, the fact that the protocol needs to be spawned is an implementation detail. 
```suggestion
    async fn handle_block_event(&mut self, block_event: &BlockEvent) {
```

base_layer/core/src/mempool/sync_protocol/initializer.rs
```diff
@@ -103,9 +107,9 @@ impl ServiceInitializer for MempoolSyncInitializer {
                 }
                 log_mdc::extend(mdc.clone());
             }
-
-            MempoolSyncProtocol::new(config, notif_rx, connectivity_event_subscription, mempool)
-                .run()
+            let block_event_stream = base_node.get_block_event_stream();
+            MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity)
+                .run(block_event_stream)
```
sdbondi: I think we can keep it consistent with the the way other dependencies are added
```suggestion
            let block_event_stream = base_node.get_block_event_stream();
            MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity, block_event_stream)
                .run()
```

base_layer/core/src/mempool/sync_protocol/initializer.rs
```diff
@@ -103,9 +104,12 @@ impl ServiceInitializer for MempoolSyncInitializer {
...
             }
-
-            MempoolSyncProtocol::new(config, notif_rx, connectivity_event_subscription, mempool)
-                .run()
+            let streams = MempoolSyncStreams {
+                block_event_stream: base_node.get_block_event_stream(),
+                connectivity_events: connectivity_event_subscription,
+            };
+            MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity)
+                .run(streams)
```
sdbondi: nit: I'm not a fan of this pattern. Rather use the connectivity handle to get the event stream inside the run function.
```suggestion
           let base_node_events = base_node.get_block_event_stream();

            MempoolSyncProtocol::new(config, notif_rx, mempool, connectivity, base_node_events)
                .run()
```

sdbondi (dismissed):
utAck - minor nits

stringhandler (dismissed):

stringhandler (approved):
utACK

PR #4733, opened 2022-09-23
fix!: update commitment signature

AaronFeickert:
Switches from `CommitmentSignature` to `CommitmentAndPublicKeySignature` for transaction authorization. Work in progress.

BREAKING CHANGE: Commitment signature proof has changed


applications/tari_app_grpc/proto/transaction.proto
```diff
@@ -69,7 +69,7 @@ message TransactionInput {
     // The script input data, if any
     bytes input_data = 5;
     // A signature with k_s, signing the script, input data, and mined height
-    ComSignature script_signature = 7;
+    ComAndPubSignature script_signature = 7;
```
stringhandler, line 72: I think this should still be ComSig

AaronFeickert: If the [description in RFC-0201](https://rfc.tari.com/RFC-0201_TariScript.html#transaction-input-changes) is still accurate, the intent is to assert knowledge of both the commitment opening and script private key, which necessitates the use of `CommitmentAndPublicKeySignature` even though the process isn't a multi-party operation. Otherwise you run into the same issue as with the transaction output use case, where you only assert the signer knows the sum of the commitment mask and script private key, and would need to ensure this doesn't result in any vulnerabilities elsewhere in the protocol.

SWvheerden: The assumption with RFC-0201 with the input and outputs are that the because the signer knows the aggregate keys, they know the individual as well. 

And we know even with multiparty signers that key cancelation is not a problem as they need to sign with the commitment blinding factor in the excess (so k needs to be known) and they need to calculate the correct script offset(again so k needs to be known)

I feel that a comsignature is 100% fine for both, but I am busy implementing this as a ComAndPubSignature for now.

stringhandler:
Closed in favour of #4943

Issue #4721, opened 2022-09-22
[console wallet/mempool] Mined transactions being invalidated due to `DuplicateKernelError`, `Validation failed due to already mined kernel`


hansieodendaal:
In some cases, a perfectly fine transaction that has been submitted to the mempool and being monitored by the wallet will be invalidated in the mempool whren it has been mined due to `DuplicateKernelError`, `Validation failed due to already mined kernel`, but then marked invalid in the wallet as well.

This behaviour surfaces during many coin splits.

stringhandler:
I think the broadcast protocol should not see this as an error.

PR #4735, opened 2022-09-26
fix: mined tx being invalid

Cifko:
Description
---
Fix mined transaction marked as invalid in the wallet.
https://github.com/tari-project/tari/issues/4721

How Has This Been Tested?
---
I wasn't able to reproduce the original bug. So only github actions tests.


SWvheerden (dismissed):
LGTM

stringhandler (approved):

PR #4737, opened 2022-09-26
feat: change priority in mempool to take into account age

SWvheerden:
Description
---
Changes the mempool to look at insertion time and not input maturity. 
These changes reflect the new updated [mempool rfc](https://github.com/tari-project/rfcs/pull/13)

Motivation and Context
---
The current code looks at the min input maturity, aka spending the oldest input, as an equalizer on the fee_per_gram when determining the transaction to pick for mining. This check is moot as even our default implementation of transaction sending uses 0 for all transactions except the coinbases which have their maturity checked on consensus rules. This means that most users will just keep that 0 as its not enforceable and, it will in most cases be 0, meaning this check is moot. 

This update adds in a local time as the equalizer for the same fee_per_gram so that we still choose the oldest transactions to put in a block. Adding in this behavior gives preference to older transactions and is desireable for the network as it means most transactions will be mined at some point, and it won't lead to stale transactions sitting in a mempool.


base_layer/core/src/mempool/priority/prioritized_transaction.rs
```diff
@@ -46,7 +47,11 @@ impl FeePriority {
         // Big-endian used here, the MSB is in the starting index. The ordering for Vec<u8> is big-endian and the
         // unconfirmed pool expects the lowest priority to be sorted lowest to highest in the BTreeMap
         let fee_priority = fee_per_byte.to_be_bytes();
-        let maturity_priority = (u64::MAX - transaction.min_input_maturity()).to_be_bytes();
+        let age = match SystemTime::now().duration_since(UNIX_EPOCH) {
+            Ok(n) => n.as_secs(),
+            Err(_) => 0,
+        };
+        let maturity_priority = (u64::MAX - age).to_be_bytes();
```
stringhandler:
```suggestion
        let age_priority = (u64::MAX - age).to_be_bytes();
```

stringhandler (approved):

PR #4738, opened 2022-09-26
ci: run coverage on prs

stringhandler:
Description
---
Run coverage on all PRs

Motivation and Context
---
Currently, this is only run after merge. This is often missed. With fewer PRs, and a more stable code base, I think we can run it on every PR

How Has This Been Tested?
---
Since this is CI related, I'll be using the PR itself to test

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -7,6 +7,11 @@ on:
     branches:
       - development
       - ci-coverage-*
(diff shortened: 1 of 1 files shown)


CjS77:
Hmm.. it's not a requirement metric, it's more a tracking metric, so running on every PR is just another very slow action to wait for on PRs INHO

PR #4742, opened 2022-09-27
feat: add multisig script that returns aggregate of signed public keys

hansieodendaal:
Description
---
Added an `m-of-n` multisig TariScript that returns the aggregate public key of the signatories if successful and fails otherwise. 

This is useful if the aggregate public key of the signatories is also the script public key, where signatories would work together to create an aggregate script signature using their individual script private keys.

Motivation and Context
---
To enhance the practicality of the  `m-of-n` multisig TariScript.

How Has This Been Tested?
---
Unit tests

Co-Authored-By: SW van Heerden swvheerden@gmail.com


stringhandler:
Needs a link to an RFC

hansieodendaal:
Please see https://github.com/tari-project/rfcs/pull/14 where the op codes have been added.

infrastructure/tari_script/src/script.rs
```diff
@@ -522,20 +529,25 @@ impl TariScript {
...
                 if !sig_set.contains(s) && !key_signed[i] && s.verify_challenge(pk, &message) {
                     key_signed[i] = true;
                     sig_set.insert(s);
+                    agg_pub_key = agg_pub_key + pk;
                     break;
                 }
             }
             if !sig_set.contains(s) {
-                return Ok(false);
+                return Ok(None);
```
jorgeantonio21, line 545: As an honest question, entering the current scope, means that the signature `s` hasn't signed the challenge set by `pk` and `message`. Shouldn't we error in this case, as this might be a case of a malicious execution ? Moreover, it would also allow to distinguish the case in which not enough signatures have been provided.

jorgeantonio21:
LGTM !

infrastructure/tari_script/src/script.rs
hansieodendaal: Error conditions are managed by the caller branch in `fn execute_opcode(`. It depends on the design of the opcode response.

infrastructure/tari_script/src/op_codes.rs
```diff
@@ -458,6 +466,13 @@ impl Opcode {
                 }
                 array.extend_from_slice(msg.deref());
             },
+            CheckMultiSigVerifyAggregatePubKey(m, n, public_keys, msg) => {
+                array.extend_from_slice(&[OP_CHECK_MULTI_SIG_VERIFY_AGGREGATE_PUB_KEY, *m, *n]);
+                for public_key in public_keys {
+                    array.extend(public_key.to_vec());
```
sdbondi: nit: Save an allocation
```suggestion
                    array.extend_from_slice(public_key.as_bytes());
```

sdbondi (commented):
Makes sense. One very minor improvement.

An alternative is make CHECK_MULTISIG_VERIFY push an aggregate public key (breaking change, but does not matter at this stage) and the user/script author could discard this using drop if needed but IDK this may be better and the difference between CHECK_MULTISIG_VERIFY and this opcode should be pretty straightforward to document/understand.

infrastructure/tari_script/src/op_codes.rs
```diff
@@ -116,6 +116,7 @@ pub const OP_CHECK_MULTI_SIG_VERIFY: u8 = 0xaf;
 pub const OP_HASH_BLAKE256: u8 = 0xb0;
 pub const OP_HASH_SHA256: u8 = 0xb1;
 pub const OP_HASH_SHA3: u8 = 0xb2;
+pub const OP_CHECK_MULTI_SIG_VERIFY_AGGREGATE_PUB_KEY: u8 = 0xb3;
```
sdbondi: This byte code is now in use since #4749 
```suggestion
pub const OP_CHECK_MULTI_SIG_VERIFY_AGGREGATE_PUB_KEY: u8 = 0xb4;
```

hansieodendaal:
> An alternative is make CHECK_MULTISIG_VERIFY push an aggregate public key (breaking change, but does not matter at this stage) and the user/script author could discard this using drop if needed but IDK this may be better and the difference between CHECK_MULTISIG_VERIFY and this opcode should be pretty straightforward to document/understand.

@CjS77 and @stringhandler your opinions? Do we add this additional op-code or condense `CheckMultiSigVerifyAggregatePubKey` into `CheckMultiSigVerify`?

stringhandler (approved):

PR #4743, opened 2022-09-27
fix(clients): fix tari nodejs client proto paths

sdbondi:
Description
---
Uses correct path for proto files for nodejs wallet client

Motivation and Context
---
NodeJS clients were put in `nodejs` directory, however the relative path to the proto files were not updated

How Has This Been Tested?
---
Running integration tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ commands:
           command: node -v
       - run:
           name: npm ci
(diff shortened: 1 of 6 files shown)


Cifko:
You also need to fix the path in github actions `integration_tests.yml`

stringhandler (dismissed):

PR #4749, opened 2022-09-28
feat(tariscript): adds ToRistrettoPoint op-code

sdbondi:
Description
---
- Adds ToRistrettoPoint opcodes
- Added scalar stack item

Motivation and Context
---
As per RFC-202
Ref https://github.com/tari-project/rfcs/pull/15
Ref https://github.com/tari-project/tari/pull/4742

How Has This Been Tested?
---
Additional unit tests + tests updated

### DIFF
--- a/infrastructure/tari_script/src/op_codes.rs
+++ b/infrastructure/tari_script/src/op_codes.rs
@@ -24,6 +24,7 @@ use tari_utilities::{hex::Hex, ByteArray, ByteArrayError};
 use super::ScriptError;
 
 pub type HashValue = [u8; 32];
+pub type ScalarValue = [u8; 32];
 pub type Message = [u8; MESSAGE_LENGTH];
 
 const PUBLIC_KEY_LENGTH: usize = 32;
--- a/infrastructure/tari_script/src/op_codes.rs
+++ b/infrastructure/tari_script/src/op_codes.rs
@@ -452,14 +452,14 @@ impl Opcode {
             CheckMultiSig(m, n, public_keys, msg) => {
                 array.extend_from_slice(&[OP_CHECK_MULTI_SIG, *m, *n]);
                 for public_key in public_keys {
-                    array.extend(public_key.to_vec());
+                    array.extend(public_key.as_bytes());
                 }
                 array.extend_from_slice(msg.deref());
             },
             CheckMultiSigVerify(m, n, public_keys, msg) => {
                 array.extend_from_slice(&[OP_CHECK_MULTI_SIG_VERIFY, *m, *n]);
                 for public_key in public_keys {
-                    array.extend(public_key.to_vec());
+                    array.extend(public_key.as_bytes());
                 }
                 array.extend_from_slice(msg.deref());
             },
--- a/infrastructure/tari_script/src/op_codes.rs
+++ b/infrastructure/tari_script/src/op_codes.rs
@@ -116,6 +116,7 @@ pub const OP_CHECK_MULTI_SIG_VERIFY: u8 = 0xaf;
 pub const OP_HASH_BLAKE256: u8 = 0xb0;
 pub const OP_HASH_SHA256: u8 = 0xb1;
 pub const OP_HASH_SHA3: u8 = 0xb2;
+pub const OP_TO_RISTRETTO_POINT: u8 = 0xb3;
 
 // Opcode constants: Miscellaneous
 pub const OP_RETURN: u8 = 0x60;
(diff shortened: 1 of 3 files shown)


infrastructure/tari_script/src/op_codes.rs
```diff
@@ -116,6 +116,7 @@ pub const OP_CHECK_MULTI_SIG_VERIFY: u8 = 0xaf;
 pub const OP_HASH_BLAKE256: u8 = 0xb0;
 pub const OP_HASH_SHA256: u8 = 0xb1;
 pub const OP_HASH_SHA3: u8 = 0xb2;
+pub const OP_TO_RISTRETTO_POINT: u8 = 0xb3;
```
delta1, line 120: https://github.com/tari-project/tari/pull/4742/files#diff-307c63687b44168345b04870922da3ee590273f9c3cc8b57795d1992d3706386R119

sdbondi: Thanks :eyes:

hansieodendaal (dismissed):
Nice, just noted today this is missing.

LGTM

infrastructure/tari_script/src/script.rs
```diff
@@ -537,6 +539,19 @@ impl TariScript {
...
+
+    fn handle_to_ristretto_point(&self, stack: &mut ExecutionStack) -> Result<(), ScriptError> {
+        let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
+        let scalar = match &item {
+            StackItem::Hash(hash) => hash.as_slice(),
+            StackItem::Scalar(scalar) => scalar.as_slice(),
+            _ => return Err(ScriptError::IncompatibleTypes),
+        };
+        let scalar = RistrettoSecretKey::from_bytes(scalar).map_err(|_| ScriptError::InvalidData)?;
+        let ristretto_pk = RistrettoPublicKey::from_secret_key(&scalar);
```
willyrgf: May improve readability.

```suggestion
        let ristretto_sk = RistrettoSecretKey::from_bytes(scalar).map_err(|_| ScriptError::InvalidData)?;
        let ristretto_pk = RistrettoPublicKey::from_secret_key(&ristretto_sk);
```

infrastructure/tari_script/src/stack.rs
```diff
@@ -52,11 +52,13 @@ pub const TYPE_HASH: u8 = 2;
 pub const TYPE_COMMITMENT: u8 = 3;
 pub const TYPE_PUBKEY: u8 = 4;
 pub const TYPE_SIG: u8 = 5;
+pub const TYPE_SCALAR: u8 = 6;
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StackItem {
     Number(i64),
     Hash(HashValue),
+    Scalar([u8; 32]),
```
willyrgf: Should we have something like `HashValue` to scalar? Like a `pub type ScalarValue = [u8; 32];` in `infrastructure/tari_script/src/op_codes.rs`.

willyrgf:
What do you think about this kind of approach to `StackItem` values and similar cases? https://github.com/willyrgf/tari/commit/8e67de1e178f1dd465aee5f1785c9f0afc38de14

infrastructure/tari_script/src/stack.rs
sdbondi: Yeah agree with that, will change

sdbondi:
> What do you think about this kind of approach to `StackItem` values and similar cases? [willyrgf@8e67de1](https://github.com/willyrgf/tari/commit/8e67de1e178f1dd465aee5f1785c9f0afc38de14)

Yeah that is a little nicer, however I'd favour the current approach over introducing another dependency.

infrastructure/tari_script/src/script.rs
sdbondi: Will change :ok_hand: - the RistrettoSecret key part is more of an intermediate step so not technically a secret

willyrgf:
> > What do you think about this kind of approach to `StackItem` values and similar cases? [willyrgf@8e67de1](https://github.com/willyrgf/tari/commit/8e67de1e178f1dd465aee5f1785c9f0afc38de14)
> 
> 
> 
> Yeah that is a little nicer, however I'd favour the current approach over introducing another dependency.

Yes, it's make sense. Maybe in the future with [`core::mem::variant_count`](https://github.com/rust-lang/rust/issues/73662).

stringhandler (approved):

stringhandler:
In future we will need to hard fork these in, but since this will not be used immediately, there should be enough time for nodes to upgrade

PR #4751, opened 2022-09-28
feat: adds a create combo key cli command for wallet

jorgeantonio21:
Description
---
Adds a `create_combo_key` method to wallet. The main goal is to allow the user to create a new pair of (secret key, public key), by passing in a given string value.

Motivation and Context
---
Wallet needs a terminal command to create a new public key and secret key from the key manager service.
This key needs to be created based on a provided string seed and send back to the user. 

The key manager service derives a key based on the string and the master seed of the wallet. We need to add this seed string as a branch on the local instance of the key manager and then we can track if the user asks for the key twice.

The current command is needed in order to as a firs step to execute an m-of-n multi-party script payment protocol. See [here](https://demo.hedgedoc.org/rgLUGP_YRiGKjKkgehiG4w#Step-2) and [here](https://demo.hedgedoc.org/0Lb9sKovSb-ct3g3E4A-9A?view) for more details.

How Has This Been Tested?
---
Unit tests


base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +79,20 @@ where TBackend: KeyManagerBackend + 'static
...
+            .write()
+            .await
+            .add_key_manager_branch(key_seed.clone())?;
+
+        let next_key = self.get_next_key(key_seed).await?;
+        let sk = next_key.key.clone();
+        let pk = next_key.to_public_key();
+        // (*self.key_manager_inner).read().await.get_next_key(branch).await;
+
+        Ok(KeyComboPair { sk, pk })
```
SWvheerden: I would just return a tuple here (Secret,Public)

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +655,15 @@ pub async fn command_runner(
                     Err(e) => eprintln!("BurnTari error! {}", e),
                 }
             },
+            CreateKeyCombo(args) => match create_key_combo(key_manager_service.clone(), args.key_seed).await {
+                Ok(KeyComboPair { sk, pk }) => {
+                    debug!(
```
SWvheerden, line 660:
```suggestion
                    println!(
```

SWvheerden:
```suggestion
                    println!(
```

can change more than a single line with the suggestion.

base_layer/wallet/src/key_manager_service/handle.rs
jorgeantonio21: Refactored, thanks !

applications/tari_console_wallet/src/automation/commands.rs
jorgeantonio21: Great, slipped my fingers !

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +649,17 @@ pub async fn command_runner(
                     Err(e) => eprintln!("BurnTari error! {}", e),
                 }
             },
+            CreateKeyCombo(args) => match create_key_combo(key_manager_service.clone(), args.key_seed).await {
+                Ok((sk, pk)) => {
+                    println!(
+                        "create new key combo pair: 
+                                1. secret key: {:?}, 
+                                2. public key {:?}",
+                        sk, pk
```
SWvheerden:
```suggestion
                        sk.to_hex(),pk.to_hex()
```

jorgeantonio21: Definitely ! Thanks

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +73,20 @@ where TBackend: KeyManagerBackend + 'static
...
+    async fn create_key_combo(&self, key_seed: String) -> Result<(PrivateKey, PublicKey), KeyManagerServiceError> {
+        (*self.key_manager_inner)
+            .write()
+            .await
+            .add_key_manager_branch(key_seed.clone())?;
+
+        let next_key = self.get_next_key(key_seed).await?;
+        let sk = next_key.key.clone();
+        let pk = next_key.to_public_key();
+        // (*self.key_manager_inner).read().await.get_next_key(branch).await;
```
SWvheerden:
```suggestion
```

infrastructure/tari_script/Cargo.toml
```diff
@@ -21,6 +21,4 @@ serde = "1.0.136"
 sha2 = "0.9"
 sha3 = "0.9"
 thiserror = "1.0.30"
-
-[dev-dependencies]
```
SWvheerden: Why has this been removed?

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -139,6 +140,16 @@ pub async fn burn_tari(
         .map_err(CommandError::TransactionServiceError)
 }
 
+pub async fn create_key_combo<TBackend: KeyManagerBackend + 'static>(
```
sdbondi: A better name for this feature is "create key pair"

base_layer/wallet/src/key_manager_service/handle.rs
jorgeantonio21: Removed, thanks :)

infrastructure/tari_script/Cargo.toml
jorgeantonio21: We are now in need of the `rand` library on `[dependencies]`', but can add it to `[dev-dependencies]` as well.

applications/tari_console_wallet/src/automation/commands.rs
jorgeantonio21: Definitely agreed, I changed it accordingly.

infrastructure/tari_script/Cargo.toml
SWvheerden: cool, github does not show that only the rand lib is left.

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +649,18 @@ pub async fn command_runner(
                     Err(e) => eprintln!("BurnTari error! {}", e),
                 }
             },
+            CreateKeyPair(args) => match create_key_pair(key_manager_service.clone(), args.key_seed).await {
```
sdbondi: The extra function seems unnecessary. I would also name the arg `key_branch` or `branch_name`
```suggestion
            CreateKeyPair(args) => match key_manager_service.create_key_pair(args.key_branch).await {
```

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +73,19 @@ where TBackend: KeyManagerBackend + 'static
         (*self.key_manager_inner).write().await.apply_encryption(cipher)
     }
 
+    async fn create_key_pair(&self, key_seed: String) -> Result<(PrivateKey, PublicKey), KeyManagerServiceError> {
+        (*self.key_manager_inner)
```
sdbondi: nit: I know this is copied from other parts of the code, but there is no need to explicitly dereference an Arc when using the `.` operator
```suggestion
        self.key_manager_inner
```

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +73,19 @@ where TBackend: KeyManagerBackend + 'static
...
+            .write()
+            .await
+            .add_key_manager_branch(key_seed.clone())?;
```
sdbondi: I think using the key seed as a branch exposes the key seed in cleartext in the db
Oh I see, `key_seed` is the branch name. A key seed sounds like it is the seed words or some secret data. Perhaps change this to `branch: T where T: Into<String>` as used in other parts of the code.

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +649,18 @@ pub async fn command_runner(
...
+                Ok((sk, pk)) => {
+                    println!(
+                        "create new key combo pair: 
+                                1. secret key: {}, 
+                                2. public key {}",
+                        sk.to_hex(),
+                        pk.to_hex()
```
sdbondi: The secret string is left unzeroed after it is dropped.
I think the following works:
```suggestion
                    println!(
                        "create new key combo pair: 
                                1. secret key: {}, 
                                2. public key {}",
                        Zeroizing::new(sk.to_hex()),
                        pk.to_hex()
```

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +649,18 @@ pub async fn command_runner(
                     Err(e) => eprintln!("BurnTari error! {}", e),
                 }
             },
+            CreateKeyPair(args) => match create_key_pair(key_manager_service.clone(), args.key_seed).await {
+                Ok((sk, pk)) => {
+                    println!(
+                        "create new key combo pair: 
+                                1. secret key: {}, 
+                                2. public key {}",
```
sdbondi: nit:
```suggestion
                        "New key pair: 
                                secret key: {}, 
                                public key: {}",
```

sdbondi (commented):
LGTM, some minor comments

base_layer/wallet/src/key_manager_service/handle.rs
hansieodendaal: Agreed

hansieodendaal:
**Nit:** 

I fail to understand the use case here?

>The key manager service derives a key based on the string and the master seed of the wallet. We need to add this seed string as a branch on the local instance of the key manager and then we can track if the user asks for the key twice.

base_layer/wallet/src/key_manager_service/handle.rs
delta1: This seems like something clippy should be able to catch

jorgeantonio21: Definitely, I see some other parts of the code in which this is also the case. Should these be addressed @sdbondi ? Thanks for the catch :)

@delta1, maybe because the types of `self.key_manager_inner` and `(*self.key_manager_inner)` are distinct ?

applications/tari_console_wallet/src/automation/commands.rs
jorgeantonio21: Much better ! Thanks

base_layer/wallet/src/key_manager_service/handle.rs
jorgeantonio21: Refactor accordingly :)

applications/tari_console_wallet/src/automation/commands.rs
jorgeantonio21: Great ! Actually `Zeroizing` does not implement the `Display` trait, add to refactor line `656` to use `{:?}`.

applications/tari_console_wallet/src/automation/commands.rs
jorgeantonio21: Much better :)

base_layer/wallet/src/key_manager_service/handle.rs
sdbondi: yeah I would vote for clippy catching these, but until then, not important at all to change this (nit) feel free to leave as is and resolve.

applications/tari_console_wallet/src/automation/commands.rs
sdbondi: Bummer, cant use debug format as that wont print out anything useful to the user.

Actually, you can do this because Zeroizing implements `Deref<T>`
`*Zeroizing::new(sk.to_hex())`

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +73,22 @@ where TBackend: KeyManagerBackend + 'static
         (*self.key_manager_inner).write().await.apply_encryption(cipher)
     }
 
+    async fn create_key_pair<T: Clone + Into<String> + Send>(
```
sdbondi:
```suggestion
    async fn create_key_pair<T: Into<String> + Send>(
```

base_layer/wallet/src/key_manager_service/handle.rs
```diff
@@ -73,6 +73,22 @@ where TBackend: KeyManagerBackend + 'static
...
+        &self,
+        branch: T,
+    ) -> Result<(PrivateKey, PublicKey), KeyManagerServiceError> {
+        self.key_manager_inner
+            .write()
+            .await
+            .add_key_manager_branch(branch.clone().into())?;
```
sdbondi: You won't need to clone this, into returns an owned String
```suggestion
            .add_key_manager_branch(branch.into())?;
```

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -637,6 +640,18 @@ pub async fn command_runner(
                     Err(e) => eprintln!("BurnTari error! {}", e),
                 }
             },
+            CreateKeyPair(args) => match key_manager_service.create_key_pair(args.key_branch).await {
+                Ok((sk, pk)) => {
+                    println!(
+                        "New key pair: 
+                                1. secret key: {:?}, 
+                                2. public key: {}",
+                        Zeroizing::new(sk.to_hex()),
```
sdbondi:
```suggestion
                                1. secret key: {}, 
                                2. public key: {}",
                        *Zeroizing::new(sk.to_hex()),
```

infrastructure/tari_script/Cargo.toml
sdbondi: I don't see any code changes that require rand as a dependency, I think the dev-dependency can be re-added?

base_layer/wallet/src/key_manager_service/handle.rs
sdbondi: Scratch that, I didnt see that you use the branch name below :)

jorgeantonio21: We need to clone the `branch` value, as we use it in line `85`. Better to clone it as a `String` as you suggested !

sdbondi (approved):

PR #4752, opened 2022-09-29
fix: batch rewind operations

Cifko:
Description
---
Split rewind DbTx into smaller pieces.

How Has This Been Tested?
---
I did rewind on 20000+ (empty) blocks.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1647,10 +1647,7 @@ fn check_for_valid_height<T: BlockchainBackend>(db: &T, height: u64) -> Result<(
 
 /// Removes blocks from the db from current tip to specified height.
 /// Returns the blocks removed, ordered from tip to height.
-fn rewind_to_height<T: BlockchainBackend>(
-    db: &mut T,
-    mut height: u64,
-) -> Result<Vec<Arc<ChainBlock>>, ChainStorageError> {
+fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec<Arc<ChainBlock>>, ChainStorageError> {
     let last_header = db.fetch_last_header()?;
 
     // Delete headers
@@ -1713,7 +1710,6 @@ fn rewind_to_height<T: BlockchainBackend>(
             effective_pruning_horizon
         );
         steps_back = effective_pruning_horizon;
-        height = 0;
     }
 
     for h in 0..steps_back {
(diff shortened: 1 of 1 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1718,34 +1719,43 @@ fn rewind_to_height<T: BlockchainBackend>(
...
         // mempool add reorg'ed tx.
-        for h in 0..(last_block_height - steps_back) {
-            debug!(
-                target: LOG_TARGET,
-                "Deleting blocks and utxos {}",
-                last_block_height - h - steps_back,
-            );
-            let header = fetch_header(db, last_block_height - h - steps_back)?;
-            txn.delete_block(header.hash());
+        for h in (0..(last_block_height - steps_back)).step_by(batch_by) {
```
sdbondi: We were seeing issues rewinding with < 500 blocks, I used `rewind-blockchain` to rewind 116 blocks (482 operations) and got a transaction size error. 

I've seen the block accumulated data is 110kb at height 17521, so for 100 blocks that is 11mb of the 20mb limit, and that isn't including the outputs, inputs and kernels. Admittedly, it's strange that a delete operation would take up this space, so perhaps something else is going on. 

I think we should commit every block, and if that turns out to be too slow (I suspect it wont), we can be more clever about figuring out the transaction size and when to commit.

sdbondi (commented):
LGTM, I think we need to commit every block for the `delete_block` step since I've observed this issue when rewinding 116 blocks.

base_layer/core/src/chain_storage/blockchain_database.rs
Cifko: Done. I delete all the batching. Too keep it simpler.

sdbondi: Thanks

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1758,6 +1761,7 @@ fn rewind_to_height<T: BlockchainBackend>(
         chain_header.accumulated_data().total_accumulated_difficulty
     );
 
+    let mut txn = DbTransaction::new();
```
sdbondi: Hmm just thought of something, we should update the best block metadata in the same transaction as we delete a block, otherwise it can get out of sync

sdbondi: luckily we already load the header in that loop

Cifko: Now I do setblock for one header previously from the deleted. But I'm not sure about setting the best block to 0, for the pruned ones. Because the deletion of blocks past pruned horizon is there as well. Please take a look.

sdbondi: yeah, I think thats correct - we are saying to other nodes "I have no blocks to offer", and the chain metadata should be set to the genesis height (0), hash and accumulated data

PR #4753, opened 2022-09-29
fix(dht/encryption): greatly reduce heap allocations for encrypted messaging

sdbondi:
Description
---
- Encrypt, decrypt and message padding mutate a single buffer for encrypted messages

Motivation and Context
---
Encrypted message handling should be as efficient as possible. The previous implementation performed allocations of the full padded message size twice for encryption and twice for decryption. Increasing memory usage, and negating the performance benefits of using an encryption keystream.

This PR allocates a single buffer for the message to be de/encrypted and de/encrypts the contents in-place using the BytesMut type from the `bytes` crate.

How Has This Been Tested?
---
This change is backwards compatible, tested on current esme network and updated existing tests as required.
Discovery: OK
Memorynet: OK
PingPong: OK
InteractiveTransactions: OK
SafTransactions: OK

### DIFF
--- a/comms/dht/src/dedup/mod.rs
+++ b/comms/dht/src/dedup/mod.rs
@@ -197,7 +197,7 @@ mod test {
         assert!(dedup.poll_ready(&mut cx).is_ready());
         let node_identity = make_node_identity();
         let inbound_message =
-            make_dht_inbound_message(&node_identity, vec![], DhtMessageFlags::empty(), false, false).unwrap();
+            make_dht_inbound_message(&node_identity, &vec![], DhtMessageFlags::empty(), false, false).unwrap();
         let decrypted_msg = DecryptedDhtMessage::succeeded(wrap_in_envelope_body!(vec![]), None, inbound_message);
 
         rt.block_on(dedup.call(decrypted_msg.clone())).unwrap();
@@ -213,12 +213,12 @@ mod test {
     #[test]
     fn deterministic_hash() {
         const TEST_MSG: &[u8] = b"test123";
-        const EXPECTED_HASH: &str = "d6333668f259f677703fbe4e89152ee41c7c01f6dec502befc63120246523ffe";
+        const EXPECTED_HASH: &str = "1c2bb1bcff443af4441b789bd1d6984bb8d7bed2c9f85e8cf4f45615fdd9e47d";
 
         let node_identity = make_node_identity();
         let dht_message = make_dht_inbound_message(
             &node_identity,
-            TEST_MSG.to_vec(),
+            &TEST_MSG.to_vec(),
             DhtMessageFlags::empty(),
             false,
             false,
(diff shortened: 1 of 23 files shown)


comms/dht/src/test_utils/makers.rs
```diff
@@ -123,9 +124,9 @@ pub fn make_valid_message_signature(node_identity: &NodeIdentity, message: &[u8]
         .to_encoded_bytes()
 }
 
-pub fn make_dht_inbound_message(
+pub fn make_dht_inbound_message<T: prost::Message>(
     node_identity: &NodeIdentity,
-    body: Vec<u8>,
+    body: &T,
```
CjS77, line 129: You might be able to save a lot of `&[...].to_vec()` calls with this type of thing:
```rust
fn foo<R: Debug + ?Sized, T: AsRef<R>>(t: T)  {
    println!("{:?}", t.as_ref());
}

fn main() {
    let v = vec![1];
    foo::<Vec<i32>, _>(&v);
    foo::<[i32], _>(&[1,2,3]);
}
```

comms/dht/src/dedup/mod.rs
```diff
@@ -213,12 +213,12 @@ mod test {
     #[test]
     fn deterministic_hash() {
         const TEST_MSG: &[u8] = b"test123";
-        const EXPECTED_HASH: &str = "d6333668f259f677703fbe4e89152ee41c7c01f6dec502befc63120246523ffe";
+        const EXPECTED_HASH: &str = "1c2bb1bcff443af4441b789bd1d6984bb8d7bed2c9f85e8cf4f45615fdd9e47d";
```
CjS77, line 216: 😕 Why does the hash change here? It looks like this should be a drop-in replacement?

CjS77 (dismissed):
utACK

Looks great. 1 honest question, and one nit/suggestion

comms/dht/src/dedup/mod.rs
sdbondi: Yeah looks like a breaking change :) This has more to do with changing the test helper to take a prost::Message instead of raw bytes. Now the Vec is serialized into protobuf before being hashed, which matches thebehaviour of the real code.

comms/dht/src/test_utils/makers.rs
sdbondi: We want this to work for any impl prost::Message, [u8] nor &[u8] implement this but Vec<u8> does which is why, for tests only, we get the wrong looking &[..].to_vec(). Some tests need to pass in a DhtEnvelope type which is the case where `&T where T: prost::Message`  means the test doesn't have to serialize first (and we can use the non-test function `prepare_message` instead of reimplementing that for tests).

jorgeantonio21:
Really great ! Was unaware of this issue, and seeing how to fix it with `BytesMut` was very insightful. Thanks @sdbondi :)

stringhandler (approved):

PR #4755, opened 2022-09-29
feat: different default grpc ports for different networks

sdbondi:
Description
---
Use different _default_ GRPC ports for different networks

- BaseNode
    - Dibbler: 18122
    - Esmeralda: 18142
    - Igor: 18152
- ConsoleWallet
    - Dibbler: 18123
    - Esmeralda: 18143
    - Igor: 18153
    
Motivation and Context
---
More explicit separation of networks allowing GRPC to work when running two different networks on the same machine.

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_base_node/src/config.rs
+++ b/applications/tari_base_node/src/config.rs
@@ -144,7 +144,7 @@ impl Default for BaseNodeConfig {
         Self {
             override_from: None,
             network: Network::default(),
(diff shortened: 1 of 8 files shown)


CjS77 (dismissed):
utACK

stringhandler (commented):
Thanks. We need to add `grpc_enabled` back to the base node as there is currently no way of turning GRPC off

sdbondi:
:100:  - Done

PR #4759, opened 2022-09-30
feat: add wallet cli command for utxo creation with aggregate public key

jorgeantonio21:
Description
---
The goal of the current PR is to add a cli command such that a wallet can create a utxo with a spend `CheckMultiSigVerifyAggregatePubKey` script that requires the a m-of-n list of keys to spend.
It should be merged after #4742 and #4751.

Motivation and Context
---
It addresses `Step 2` in the following [MvP](https://demo.hedgedoc.org/rgLUGP_YRiGKjKkgehiG4w#Step-2).

How Has This Been Tested?
---
Unit tests


SWvheerden (approved):
LGTM

PR #4769, opened 2022-10-04
feat: remove dalek ng

### LINKED ISSUE #4736: Update dependency to curve25519-dalek from dalek-ng

SWvheerden:
Description
---
Removed and replaced the  `curve25519-dalek-ng` dependency and rather use `curve25519-dalek` as this is the more updated repo.
Updated the version as this is a breaking change with respect to dependencies.
Matched the version of `rand` to `rand_core` so that the traits can work.`

How Has This Been Tested?
---
All unit test pass. 

Fixes: #4736

### DIFF
--- a/applications/tari_app_grpc/Cargo.toml
+++ b/applications/tari_app_grpc/Cargo.toml
@@ -11,9 +11,9 @@ edition = "2018"
 tari_common_types = { version = "^0.38", path = "../../base_layer/common_types" }
 tari_comms = { path = "../../comms/core" }
 tari_core = { path = "../../base_layer/core" }
-tari_crypto = { git = "https://github.com/tari-project/tari-crypto.git", tag = "v0.15.5" }
+tari_crypto = { git = "https://github.com/SWvheerden/tari-crypto.git", rev = "348d9864574707af97bd33e759756917c618b307" }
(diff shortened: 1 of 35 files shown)


applications/tari_app_grpc/Cargo.toml
```diff
@@ -23,10 +23,10 @@ log = "0.4"
 num-traits = "0.2.15"
 prost = "0.9"
 prost-types = "0.9"
-rand = "0.8"
+rand = "0.8.5"
```
CjS77, line 26: What happens if this is changed to 0.7.3?

CjS77 (commented):
ut ACK

applications/tari_app_grpc/Cargo.toml
SWvheerden: Cargo cannot find matching dependencies then. 
Argon2 uses rand_core 0.6, which is used by rand 0.8. 
The rand_core version used by rand and other libraries needs to match up. 
We need to downgrade argon2 too far back to get it to use rand_core 0.5 which is used by rand 0.7. 

Luckily app_grpc does not use tari_crypto which now requires rand to be 0.7.

stringhandler (approved):

PR #4778, opened 2022-10-05
feat: move nonce to first in sha hash

### LINKED ISSUE #4767: Move nonce to front of sha3 hash
Currently the mining moved the nonce to the back of hash for the sha3, we need to move  it forward so that the complete hash needs to be calculated each time.

SWvheerden:
Description
---

This moves the nonce to the front of the hashing order when hashing for the sha3 difficulty. 
This is done so that mining cannot cache part most the header and only load the nonce in. This forces the miner to hash the complete header each time the nonce chances. 

Motivation and Context
---

Fixes: #4767 

How Has This Been Tested?
---
Unit tests all pass.


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -519,6 +518,27 @@ impl ConsensusConstants {
...
+            max_script_byte_size: 2048,
+            input_version_range: input_version_range.clone(),
+            output_version_range: output_version_range.clone(),
+            kernel_version_range: kernel_version_range.clone(),
+            permitted_output_types: Self::current_permitted_output_types(),
+        },ConsensusConstants {
+            effective_from_height: 19000,
+            coinbase_lock_height: 6,
+            blockchain_version: 1,
+            valid_blockchain_version_range: 0..=0,
```
delta1: Doesn’t this need to be updated to include version 1?

SWvheerden: good catch

SWvheerden:
We need to move the activation height before merging this

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -505,29 +505,52 @@ impl ConsensusConstants {
...
+                faucet_value: (10 * 4000) * T,
+                transaction_weight: TransactionWeight::v1(),
+                max_script_byte_size: 2048,
+                input_version_range: input_version_range.clone(),
+                output_version_range: output_version_range.clone(),
+                kernel_version_range: kernel_version_range.clone(),
+                permitted_output_types: Self::current_permitted_output_types(),
+            },
+            ConsensusConstants {
+                effective_from_height: 19000,
```
stringhandler:
```suggestion
                effective_from_height: 23000,
```

stringhandler (approved):

PR #4780, opened 2022-10-06
fix: fix config.toml bug

hansieodendaal:
Description
---
The base node errored when reading the `block_sync_trigger = 5` setting
```
ExitError { exit_code: ConfigError, details: Some("Invalid value for `base_node`: unknown field `block_sync_trigger`, expected one of `override_from`, `unconfirmed_pool`, `reorg_pool`, `service`") }
```

Motivation and Context
---
Reading default config settings should not cause an error

How Has This Been Tested?
---
System level testing


sdbondi (approved):
+1

I think we may be able to clean up some of the "silly" config keys using `#[serde(flatten)]`

PR #4792, opened 2022-10-11
fix(core): only resize db if migration is required

sdbondi:
Description
---
Adds conditional to only increase database size if migration is required

Motivation and Context
---
A new database (cucumber, functional tests) has no inputs and so migration is not required.
Ref #4791 

How Has This Been Tested?
---


stringhandler (approved):
Definitely better. The clippy failure is pre-existing in dev

sdbondi:
@stringhandler Thanks - fix for that is here: https://github.com/tari-project/tari/pull/4793

PR #4800, opened 2022-10-12
fix(core): increase sync timeouts

sdbondi:
Description
---
- Increases default sync rpc deadline from 10s to 30s

Motivation and Context
---
Large 2Mb+ blocks often take more than 10s to come through, often 11/12s. This results in longer sync times because the data that was 1/2s+ from coming through was thrown away and sync has to re-establish and continue.

How Has This Been Tested?
---
Sync works without timing out

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4594,7 +4594,7 @@ dependencies = [
 
 [[package]]
 name = "tari_app_grpc"
-version = "0.38.6"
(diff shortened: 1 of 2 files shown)


sdbondi:
Ack

CjS77 (approved):
LGTM

CjS77:
LGTM

PR #4802, opened 2022-10-13
fix(core): dont request full non-tip block if block is empty

sdbondi:
Description
---
- checks for edge-case which prevents an unnecessary full candidate block request when block is empty.

Motivation and Context
---
A full block request for empty block is not necessary as we already have all the information required to construct the candidate block. This check was missing from the branch where the candidate block is not the next tip block.

How Has This Been Tested?
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4594,7 +4594,7 @@ dependencies = [
 
 [[package]]
 name = "tari_app_grpc"
-version = "0.38.6"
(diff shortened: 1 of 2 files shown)


CjS77:
Cool.
👍

sdbondi:
ACK

stringhandler (approved):

PR #4810, opened 2022-10-17
fix: impl flush for yamux socket

sdbondi:
Description
---

Motivation and Context
---
This PR attempts to flush the socket as soon as a request is sent.

Flushing has no effect on yamux sockets, but I've implemented a flush command here to see what difference if any it makes
https://github.com/tari-project/yamux/tree/tari-flush-cmd

After looking into it, this is unlikely to make a difference as yamux flushes while reading.
If anything it may make performance worse, as flush calls are expensive, but I have not observed this and it seems to perform similarly with a small make-it-rain (120 tx). Difficult to draw conclusions without perf testing.

The idea though is to remove an unnecessary delay before a request message is actually sent

How Has This Been Tested?
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,15 +153,6 @@ checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
 
 [[package]]
 name = "arrayvec"
-version = "0.4.12"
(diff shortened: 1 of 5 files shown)


CjS77:
utACK

sdbondi:
Going to close this experiment as there was no improvement.

PR #4817, opened 2022-10-18
fix(comms/rpc): measures client-side latency to first message received

sdbondi:
Description
---
measures latency using the time the first message is received back after a request, including chunked messages

Motivation and Context
---
The client RPC implementation would include time to read all chunked messages instead of only up to the first message. 
Last request latency would not be affected for messages that are not chunked (< 256Kb) including get_tip_info

How Has This Been Tested?
---
Manually (observing wallet latency read-outs which are unaffected), existing tests.


CjS77:
utACK

CjS77 (approved):
LGTM

CjS77:
👍

PR #4818, opened 2022-10-18
refactor: split tari_base_node and tari_console_wallet into a lib component

mrnaveira:
**Still a work in progress**

Description
---
* Splitted `tari_base_node` and `tari_console_wallet` into `lib.rs` and `main.rs`
* TBD: split `tari_miner`

Motivation and Context
---
To enable the new cucumber-rs integration tests on [tari-dan#103](https://github.com/tari-project/tari-dan/pull/103), specifically to spawn base nodes and wallets, it's desirable to be able to treat them as native Rust functions exported in a `lib.rs` module.

How Has This Been Tested?
---


applications/tari_console_wallet/src/config.rs
```diff
@@ -20,12 +20,15 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#![allow(dead_code, unused)]
```
sdbondi, line 23: I think the problem here is that we have a lib and a bin in the same crate. 
The bin is including the modules directly e.g. `mod automation` and also uses the lib `use tari_console_wallet::***`. The former bin modules are built into the bin but are unused. 

I think we'll need to expose the modules via the `tari_console_wallet` lib and have the main/bin use that.

stringhandler (approved):

stringhandler:
I'm going to merge it for now. We can remove the clippy allows later

PR #4819, opened 2022-10-19
feat(comms): adds periodic socket-level liveness checks

sdbondi:
Description
---
- adds socket-level liveness checks
- adds configuration to enable liveness checks (currently enabled by default in base node, disabled in wallet)
- update status line to display liveness status

Motivation and Context
---
Allows us to gain visibility on the base latency of the transport without including overhead of the noise socket and yamux

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_base_node/src/bootstrap.rs
+++ b/applications/tari_base_node/src/bootstrap.rs
@@ -108,6 +108,8 @@ where B: BlockchainBackend + 'static
 
         // TODO: This should probably be disabled in future and have it optionally set/unset in the config - this check
         //       does allow MITM/ISP/tor router to connect this node's IP to a destination IP/onion address.
+        //       Specifically, "pingpong" text is periodically sent on an unencrypted socket allowing anyone observing
+        //       the traffic to recognise the sending IP address as almost certainly a tari node.
         p2p_config.listener_liveness_check_interval = Some(Duration::from_secs(15));
 
         let mut handles = StackBuilder::new(self.interrupt_signal)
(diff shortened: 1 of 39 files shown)


applications/tari_base_node/src/bootstrap.rs
```diff
@@ -106,6 +106,10 @@ where B: BlockchainBackend + 'static
             .map_err(|e| ExitError::new(ExitCode::ConfigError, e))?;
         p2p_config.transport.tor.identity = tor_identity;
 
+        // TODO: This should probably be disabled in future and have it optionally set/unset in the config - this check
```
stringhandler, line 109: Can you add more detail about how this leaks the IP address?

sdbondi: Sure, TL;DR is that the "pingpong" text is periodically sent on an unencrypted socket allowing anyone observing the traffic to recognise the sending IP address as almost certainly a tari node. With tor, a tor router can make this observation and over IP an ISP/MITM. In fairness, this observation can also be made by observing the first network byte and probably heuristically.

applications/tari_base_node/src/bootstrap.rs
```diff
@@ -106,6 +106,10 @@ where B: BlockchainBackend + 'static
...
+        //       does allow MITM/ISP/tor router to connect this node's IP to a destination IP/onion address.
```
stringhandler, line 110:
```suggestion
        //       does allow MITM/ISP/tor router to connect this node's IP to a destination IP/onion address.
        // TL;DR is that the "pingpong" text is periodically sent on an unencrypted socket allowing anyone observing the traffic to recognise the sending IP address as a tari node. With tor, a tor router can make this observation and over IP an ISP/MITM. In fairness, this observation can also be made by observing the first network byte
```

stringhandler (approved):
utACK

sdbondi:
ACK

PR #4825, opened 2022-10-19
chore: remove autoupdate script

stringhandler:
Description
---
Removes this autoupdate DNS publishing script. 

Motivation and Context
---
It was not being used and is in the git history if we need to find it. 
It also had an alert in https://github.com/tari-project/tari/security/dependabot/202

How Has This Been Tested?
---


sdbondi (approved):
ACK

delta1:
lol

PR #4827, opened 2022-10-19
feat: better FFI feedback from transaction validation

### LINKED ISSUE #4806: Wallet TMS validation service sends back bool to FFI
The Wallet FFI has two callbacks for validation:
```
callback_txo_validation_complete: unsafe extern "C" fn(u64, u64),
callback_transaction_validation_complete: unsafe extern "C" fn(u64, bool),
```

UTXO validation uses `callback_txo_validation_complete` which sends back:
0 -  Success
1 - Already busy
2 - Failed due to a communication issue
3 - Failed other

TX validation only sends back:
true - success
false - all other including already busy. 

Tx validation also needs to send back 0-4 depending on why it failed. 

This feedback is used to determine if the mobile wallet should select a new base node or not. This is causing the mobile wallets to keep connecting to new base_nodes

SWvheerden:
Description
---
This returns a u64 rather than a bool for the transaction validation callback.
 
Motivation and Context
---
Using a bool, the subscriber only knows if the validation runs through without error or failed. 
By returning a u64 we give the subscriber much more information. 
0: successfully
1: Already running
2:  Internal failure
3: Communication failure.

How Has This Been Tested?
---
unit tests

Fixes: https://github.com/tari-project/tari/issues/4806


base_layer/wallet/src/transaction_service/handle.rs
```diff
@@ -378,10 +378,10 @@ impl fmt::Display for TransactionEvent {
                 write!(f, "Transaction validation state changed: {}", operation_id)
             },
             TransactionEvent::TransactionValidationCompleted(operation_id) => {
-                write!(f, "Transaction validation completed: {}", operation_id)
+                write!(f, "Transaction validation(#{}) completed", operation_id)
```
CjS77: You know that you can do this now, right? Great for readability.
```suggestion
                write!(f, "Transaction validation(#{operation_id}) completed")
```

CjS77:
utACK

base_layer/wallet/src/transaction_service/handle.rs
SWvheerden: No I did not, but that's great

stringhandler (approved):

PR #4830, opened 2022-10-19
fix: reduce buffer allocation size

### LINKED ISSUE #4829: Message padding allocates buffer space unnecessarily
The recent [PR 4753](https://github.com/tari-project/tari/pull/4753) changes message encryption to use an in-place `BytesMut` construction to avoid unnecessary allocation. When applying padding prior to encryption, it [reserves extra space](https://github.com/sdbondi/tari/blob/56ae0552e130f6aefe213012fbf363bc05b3fd2e/comms/dht/src/crypt.rs#L102) in the buffer to accommodate the padding before adding it. However, it reserves too much space.

The [documentation](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.reserve) for `BytesMut` indicates that `reserve` should be provided the number of _additional_ bytes needed in the buffer, not the total number of bytes needed. As a result, the current code appears to allocate the buffer to accommodate twice the message length plus padding, which appears to negate the desired benefit of using the `BytesMut` approach.

Changing the linked line to `message.reserve(padding_length);` seems to be sufficient to avoid unnecessary reallocation, while accounting for the extra space needed for padding.

AaronFeickert:
Description
---
Reduces the size of a buffer allocation when padding messages for encryption.

Closes [issue 4829](https://github.com/tari-project/tari/issues/4829).

Motivation and Context
---
As noted in [issue 4829](https://github.com/tari-project/tari/issues/4829), the `BytesMut` used to hold messages during padding and encryption allocates more space than is needed, possibly due to a mistaken assumption of the behavior of `reserve(...)`. This PR ensures that the minimum number of extra bytes needed for padding is reserved, avoiding unnecessarily large allocation.

How Has This Been Tested?
---
Existing tests.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -99,7 +99,7 @@ fn pad_message_to_base_length_multiple(
         get_message_padding_length(message.len().checked_sub(additional_prefix_space).ok_or_else(|| {
             DhtEncryptError::PaddingError("Message length shorter than the additional_prefix_space".to_string())
         })?);
-    message.reserve(message.len() + padding_length);
+    message.reserve(padding_length);
     message.extend(iter::repeat(0u8).take(padding_length));
 
     Ok(())


AaronFeickert:
Addressed in [PR 4832](https://github.com/tari-project/tari/pull/4832).

PR #4831, opened 2022-10-20
refactor: split tari_miner into a library

mrnaveira:
Description
---
Splitted `tari_miner` into `lib.rs` and `main.rs`

Motivation and Context
---
To enable the new cucumber-rs integration tests on [tari-dan#103](https://github.com/tari-project/tari-dan/pull/103) specifically to run the miner, it's desirable to be able to treat it as native Rust functions exported in a `lib.rs` module.

How Has This Been Tested?
---
Manually ran the `tari_miner` application to check that it works as before the split


mrnaveira:
Not needed any more, as the `tari-dan` layer can mine inside cucumber tests without calling the real miner.

PR #4832, opened 2022-10-20
fix(dht): fix over allocation for encrypted messages

### LINKED ISSUE #4829: Message padding allocates buffer space unnecessarily

sdbondi:
Description
---
- Fix over-allocation of message capacity for encrypted messages.
- Minor optimisation using `resize` to improve 0 padding performance 

Motivation and Context
---
`reserve` argument is additional bytes not total.
Fixes #4829 

resize also allows the addition of padding bytes using raw pointer operations which are presumed to be much quicker than using rust iterators.

How Has This Been Tested?
---
Updated unit test to check for a correct capacity.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -100,9 +100,8 @@ fn pad_message_to_base_length_multiple(
         get_message_padding_length(message.len().checked_sub(additional_prefix_space).ok_or_else(|| {
             DhtEncryptError::PaddingError("Message length shorter than the additional_prefix_space".to_string())
         })?);
-    if message.capacity() < message.len() + padding_length {
-        message.resize(message.len() + padding_length, 0);
-    }
+
+    message.resize(message.len() + padding_length, 0);
 
     Ok(padding_length)
 }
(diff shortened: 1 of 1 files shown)


SWvheerden (commented):
utACK

jorgeantonio21:
LGTM

CjS77 (dismissed):
LGTM

CjS77 (approved):

PR #4835, opened 2022-10-20
docs: add a docstring to the check_sig method

CjS77:
Description
---

Helps clarify the algorithm and point out some edge cases that are being handled. Also the return type is not 100% intuitive, so this makes this clear too.

### DIFF
--- a/infrastructure/tari_script/src/script.rs
+++ b/infrastructure/tari_script/src/script.rs
@@ -520,7 +520,7 @@ impl TariScript {
     /// * The _m_ signatures are expected to be the top _m_ items on the stack.
     /// * Every public key can be used AT MOST once.
     /// * Every signature MUST be a valid signature using one of the public keys
-    /// * _n_ and _m_ must be positive AND n <= m AND n <= MAX_MULTISIG_LIMIT (32).
+    /// * _m_ and _n_ must be positive AND m <= n AND n <= MAX_MULTISIG_LIMIT (32).
     fn check_multisig(
         &self,
         stack: &mut ExecutionStack,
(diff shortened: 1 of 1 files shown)


CjS77:
ACK

infrastructure/tari_script/src/script.rs
```diff
@@ -505,6 +505,22 @@ impl TariScript {
...
+    /// signatures. If the validation was NOT successful, `check_multisig` returns `Ok(None)`. This is a private
+    /// function, and callers will interpret these results according to their use cases.
+    ///
+    /// Other problems, such as stack underflows, invalid parameters etc return an `Err` as usual.
+    ///
+    /// Notes:
+    /// * The _m_ signatures are expected to be the top _m_ items on the stack.
+    /// * Every public key can be used AT MOST once.
+    /// * Every signature MUST be a valid signature using one of the public keys
+    /// * _n_ and _m_ must be positive AND n <= m AND n <= MAX_MULTISIG_LIMIT (32).
```
delta1:
```suggestion
    /// * _m_ and _n_ must be positive AND m <= n AND n <= MAX_MULTISIG_LIMIT (32).
```

delta1: from what I can see (and recall) in the code anyway :P

delta1 (approved):
ACK 2693b01

infrastructure/tari_script/src/script.rs
CjS77: derp

delta1:
PS: title slug should be `docs:`

sdbondi (commented):
ACK

PR #4836, opened 2022-10-20
feat: add opcode versions

### LINKED ISSUE #4821: Add allowed list of opcodes on consensus
We need to add a list of allowed opcodes on the consensus that's allowed per height. Similar to how output, input, etc versions are handled

SWvheerden:
Description
---
This adds in an opcode versioning that is allowed.
Refactors the validation to ensure transaction versioning info is check by blocks as well. 

Motivation and Context
---
We might need to add in new opcodes as hard forks. This adds the opcode version as part of the consensus code so we can track allowed opcodes. 

How Has This Been Tested?
---
Unit tests

Fixes: #4821

### DIFF
--- a/infrastructure/tari_script/src/op_codes.rs
+++ b/infrastructure/tari_script/src/op_codes.rs
@@ -261,7 +261,8 @@ impl Opcode {
     pub fn get_version(&self) -> OpcodeVersion {
         match self {
             Opcode::CheckMultiSigVerify(..) => OpcodeVersion::V1,
-            _ => OpcodeVersion::V0,
+// ... expand existing,
+            _ => OpcodeVersion::V2,
         }
     }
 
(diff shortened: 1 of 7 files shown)


infrastructure/tari_script/src/op_codes.rs
```diff
@@ -258,6 +258,13 @@ pub enum Opcode {
 }
 
 impl Opcode {
+    pub fn get_version(&self) -> OpcodeVersion {
+        match self {
+            Opcode::CheckMultiSigVerify(..) => OpcodeVersion::V1,
+            _ => OpcodeVersion::V0,
```
stringhandler: This should be expanded, so that new opcodes are not accidentally added to an old version
```suggestion
// ... expand existing,
            _ => OpcodeVersion::V2,
```

stringhandler: Happy otherwise

stringhandler (approved):

CjS77:
BTW - assuming this gets merged, you should add covering of this in the TariScript RFC

PR #4838, opened 2022-10-21
fix(base-node): use Network::from_str to parse network in cli

sdbondi:
Description
---
- parses network in cli using Network::from_str
- fixes bug where valid network string 'esme' would break config overrides

Motivation and Context
---
Bug fix, using --network esme would result in different configuration from --network esmeralda

How Has This Been Tested?
---
Manually, `--network esme`


SWvheerden (approved):
utACK

stringhandler (approved):

PR #4840, opened 2022-10-21
feat: add more detailed error mapping for the ffi

SWvheerden:
Description
---
This adds more detailed error mapping for the FFI.

Motivation and Context
---
On sentry we get error 999, which is the general mapping for all error types without any other error logging with it. 
This is an attempt to get back more information to narrow down which error is causing the 999 without any logging.


stringhandler (approved):

stringhandler:
utACK

PR #4841, opened 2022-10-21
fix: list-connections

Cifko:
Description
---
List-connections command prints wallets only if there are base node connections.

How Has This Been Tested?
---
Manually.


sdbondi (approved):
:rofl: good catch

sdbondi:
ACK

PR #4843, opened 2022-10-21
fix(dht): zeroize AEAD keys on drop

### LINKED ISSUE #4842: Authenticated encryption keys are not zeroized on drop
Authenticated encryption keys (via the `AuthenticatedCipherKey` struct) are intended to be [zeroized on drop](https://github.com/tari-project/tari/blob/23b2c9a529f6ce7523de23dd46c5a1ff911abec8/comms/dht/src/crypt.rs#L62) as a matter of good key hygiene. However, the relevant [macros](https://github.com/tari-project/tari/blob/23b2c9a529f6ce7523de23dd46c5a1ff911abec8/comms/dht/src/crypt.rs#L59-L60) are only applied to the `CipherKey` struct.

AaronFeickert:
Description
---
Zeroizes authenticated encryption keys (via the `AuthenticatedCipherKey` struct) on drop. Fixes [issue 4842](https://github.com/tari-project/tari/issues/4842).

Motivation and Context
---
Authenticated encryption (AEAD) keys are intended to be zeroized on drop, but the relevant macros are not applied. This work adds the macros.

How Has This Been Tested?
---
Manually tested that `Zeroize` was not supported previously, and is supported now.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -59,6 +59,9 @@ use crate::{
 #[derive(Debug, Clone, Zeroize)]
 #[zeroize(drop)]
 pub struct CipherKey(chacha20::Key);
+
+#[derive(Debug, Clone, Zeroize)]
+#[zeroize(drop)]
 pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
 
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):
utACK

stringhandler (approved):

PR #4844, opened 2022-10-21
fix(dht): use new DHKE shared secret type

AaronFeickert:
Description
---
Ensures safer use of ECDH shared secrets by switching to the new `DiffieHellmanSharedSecret` type. Updates `tari-crypto` to v0.15.7 to accomplish this.

Motivation and Context
---
Currently, an ECDH secret used for message keys is produced as a `RistrettoPublicKey`, converted to bytes, and returned as a byte array. However, neither the `RistrettoPublicKey` nor the byte array are cleared when dropped. In conjunction with `tari-crypto` [PR 137](https://github.com/tari-project/tari-crypto/pull/137), this work ensures both the `RistrettoPublicKey` and byte array representations of the ECDH secret are zeroized on drop by using that PR's new `DiffieHellmanSharedSecret` type.

How Has This Been Tested?
---
Tested after applying `tari-crypto` [PR 137](https://github.com/tari-project/tari-crypto/pull/137), which adds the new `DiffieHellmanSharedSecret` generic type.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -38,14 +38,13 @@ use prost::bytes::BytesMut;
 use rand::{rngs::OsRng, RngCore};
 use tari_comms::{
     message::MessageExt,
-    types::{CommsPublicKey, CommsSecretKey},
+    types::{CommsDHKE, CommsPublicKey},
     BufMut,
 };
 use tari_crypto::{
-    keys::DiffieHellmanSharedSecret,
     tari_utilities::{epoch_time::EpochTime, ByteArray},
 };
-use zeroize::{Zeroize, Zeroizing};
+use zeroize::Zeroize;
 
 use crate::{
     comms_dht_hash_domain_challenge,
@@ -65,13 +64,6 @@ pub struct CipherKey(chacha20::Key);
 pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
 
 const MESSAGE_BASE_LENGTH: usize = 6000;
-
-/// Generates a Diffie-Hellman secret `kx.G` as a `chacha20::Key` given secret scalar `k` and public key `P = x.G`.
-pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPublicKey) -> Zeroizing<Vec<u8>> {
-    let k = CommsPublicKey::shared_secret(secret_key, public_key);
-
-    Zeroizing::new(k.to_vec())
-}
 
 fn get_message_padding_length(message_length: usize) -> usize {
     if message_length == 0 {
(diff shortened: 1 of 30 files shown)


comms/dht/src/crypt.rs
```diff
@@ -64,14 +64,10 @@ pub struct AuthenticatedCipherKey(chacha20poly1305::Key);
...
-pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPublicKey) -> [u8; 32] {
-    // TODO: PK will still leave the secret in released memory. Implementing Zerioze on RistrettoPublicKey is not
-    //       currently possible because (Compressed)RistrettoPoint does not implement it.
+pub fn generate_ecdh_secret(secret_key: &CommsSecretKey, public_key: &CommsPublicKey) -> Zeroizing<Vec<u8>> {
     let k = CommsPublicKey::shared_secret(secret_key, public_key);
-    let mut output = [0u8; 32];
 
-    output.copy_from_slice(k.as_bytes());
-    output
+    Zeroizing::new(k.to_vec())
```
sdbondi: Great, I think that the best choice we have is to heap-allocate the secret as you have, as stack types (like `[u8; 32]`) have a leaking problem as they are moved. [As stated by zeroize](https://docs.rs/zeroize/0.7.0/zeroize/index.html#stackheap-zeroing-notes).

I think we'll have to wait for  https://github.com/tari-project/tari-crypto/pull/137 to be merged so that we can call `k.zeroize()` before returning. Otherwise, `k` would still leak once dropped.

AaronFeickert: Yep, merging should be considered blocked until the `tari-crypto` PR is merged.

We shouldn't need to explicitly call `k.zeroize()` anymore, since a recent change to the `tari-crypto` PR returns the shared secret in a `Zeroize` wrapper as well.

AaronFeickert: Another thought I had was to build a more comprehensive generic shared secret type that zeroizes on drop and provides its own byte representation in `tari-crypto`. This would allow things like KDFs to do things like `shared_secret.as_bytes()` and mean we wouldn't need to handle the separate `Zeroizing` wrapper around the byte array, increasing safety.

sdbondi: Ah yup that's right, missed that the shard_secret function now returns the wrapper. 
I agree it's a bit leaky and because it is that it is now part of the public API of tari-crypto, in most cases, users would have to include zerioze explicitly as a dependency just to use tari-crypto, or you would have to re-export it from tari-crypto.
I don't think that is the end of the world, but a newtype feels more ergonomic and safer here.

AaronFeickert: Separate PRs, or part of this and the `tari-crypto` PRs?

sdbondi (dismissed):

sdbondi:
ACK

comms/dht/src/crypt.rs
sdbondi: Up to you, this can go in as is. If you'd like to continue working in this PR mark as draft and it wont auto-merge :)

AaronFeickert: Working on adding this now. It'll be a generic `DiffieHellmanSharedSecret<P>` for arbitrary `Zeroize` public key types, will zeroize on drop, and only provides `as_bytes(&self) -> &[u8]` as visibility into the shared secret. So the caller would need to go out of its way to leave a non-zeroizing copy in memory.

AaronFeickert: Done.

stringhandler (dismissed):

AaronFeickert:
Do not merge until `tari-crypto` is tagged and `tari` is updated to reflect the new dependency.

hansieodendaal:
utACK

AaronFeickert:
Now updates `tari-crypto` to v0.15.7 and is safe to merge once CI passes.

AaronFeickert:
This is currently blocked on `tari-crypto` [issue 142](https://github.com/tari-project/tari-crypto/issues/142), which addresses the underlying CI failure.

sdbondi:
@AaronFeickert Try adding the tari_crypto/wasm feature

`base_layer/key_manager/Cargo.toml`

```diff
- wasm = ["wasm-bindgen", "js", "tari_common_types", "console_error_panic_hook"]
+ wasm = ["tari_crypto/wasm", "wasm-bindgen", "js", "tari_common_types", "console_error_panic_hook"]
```

stringhandler:
I had confused this with another issue. This is not network breaking

sdbondi (approved):

stringhandler (approved):

PR #4845, opened 2022-10-24
feat: add missing fields to grpc consensus constants interface

jorgeantonio21:
Description
---
We need to add to gRPC interface `ConsensusConstants` the missing fields with respect to Rust's `ConsensusConstants` struct. 

Motivation and Context
---
To tackle [issue](https://github.com/tari-project/tari-dan/issues/115), we need a complete list of fields for gRPC `ConsensusConstants`, as in Rust original struct `ConsensusConstants`.

How Has This Been Tested?
---
Cargo check


applications/tari_app_grpc/proto/types.proto
```diff
@@ -76,4 +129,28 @@ message ConsensusConstants {
...
+    /// This is to keep track of the value inside of the genesis block
+    uint64 faucet_value = 17;
+    /// Maximum byte size of TariScript
+    uint64 max_script_byte_size = 18;
+    /// How long does it take to timeout validator node registration
+    uint64 validator_node_timeout = 19;
+    /// The height at which these constants become effective
+    uint64 effective_from_height = 20;
+    /// Current version of the blockchain
+    repeated uint64 valid_blockchain_version_range = 21;
```
sdbondi: I think a Range type is better here
```suggestion
    Range valid_blockchain_version_range = 21;
    // ...
    
    message Range { 
      uint64 min = 1;
      uint64 max = 2;
    }
```

jorgeantonio21: Definitely ! I was unaware of the `Range` type, but it is much better, thanks for pointing it out !

sdbondi: You'd have to create the type in proto :)

sdbondi (dismissed):
LGTM

sdbondi:
utACK

sdbondi (approved):

PR #4847, opened 2022-10-24
fix(wallet)!: use KDFs on ECDH shared secrets

### LINKED ISSUE #4717: Diffie-Hellman results are parsed as scalars during output recovery
When scanning for one-sided outputs, the client computes a Diffie-Hellman shared secret, serializes it to bytes, and then parses the bytes using modular reduction to a scalar. This scalar is then hashed and parsed as a scalar for use in commitment mask extraction. That scalar is then hashed again, parsed as a scalar, and used as input to a KDF for value decryption.

The relevant code:
- for [non-stealth addresses](https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/output_manager_service/service.rs#L2560-L2566)
- for [stealth addresses](https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/output_manager_service/service.rs#L2605-L2607)

While Diffie-Hellman exchange results are uniformly-distributed group elements, they need not be uniformly-distributed byte arrays. Performing scalar reduction prior to the other operations used for key derivation is nonstandard and could lead to collisions. A more standard and safe approach could be the following:
- Use the exchange result directly as input to the value encryption KDF
- If mask extraction is not guaranteed to provide its own KDF (with scalar output) for the rewind key, use the exchange result directly as input to a scalar-output KDF to produce the rewind key; otherwise, it may be safe to pass the exchange result directly to the extractor

Note that this should not apply to the method used for [detecting stealth address matching](https://github.com/tari-project/tari/blob/development/base_layer/wallet/src/output_manager_service/service.rs#L2592-L2598). That code hashes the shared secret prior to parsing as a scalar.

AaronFeickert:
Description
---
Uses KDFs on ECDH shared secrets for output generation. Closes [issue 4717](https://github.com/tari-project/tari/issues/4717).

Motivation and Context
---
Several uses of ECDH shared secrets in the output manager and transaction services parse an ECDH shared secret as a scalar spending key, and use this as input to a chain of hash functions for use in rewinding and value encryption. This is non-standard.

This work uses separate KDFs to independently produce a spending key, rewind key, and value encryption key from a `DiffieHellmanSharedSecret`-type ECDH shared secret.

How Has This Been Tested?
---
Existing tests pass.

BREAKING CHANGE: Changes the way output keys are derived.

### DIFF
--- a/base_layer/wallet/src/lib.rs
+++ b/base_layer/wallet/src/lib.rs
@@ -55,8 +55,22 @@ pub type WalletSqlite = Wallet<
 >;
 
 hash_domain!(
-    WalletSecretKeysDomain,
-    "com.tari.tari_project.base_layer.wallet.secret_keys",
+    WalletOutputRewindKeysDomain,
+    "com.tari.tari_project.base_layer.wallet.output_rewind_keys",
     1
 );
-type WalletSecretKeysDomainHasher = DomainSeparatedHasher<Blake256, WalletSecretKeysDomain>;
+type WalletOutputRewindKeysDomainHasher = DomainSeparatedHasher<Blake256, WalletOutputRewindKeysDomain>;
+
+hash_domain!(
+    WalletOutputEncryptionKeysDomain,
+    "com.tari.tari_project.base_layer.wallet.output_encryption_keys",
+    1
+);
+type WalletOutputEncryptionKeysDomainHasher = DomainSeparatedHasher<Blake256, WalletOutputEncryptionKeysDomain>;
+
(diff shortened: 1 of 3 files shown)


hansieodendaal:
I like this approach, also getting rid of repetitive key hashing.

stringhandler:
Thanks. This looks like it is a breaking change. Please add a `BREAKING CHANGE:` footer to the PR description

stringhandler (approved):

stringhandler:
utACK

PR #4851, opened 2022-10-25
feat: add static lifetime to emission amounts calculation

hansieodendaal:
Description
---
Added a static lifetime to `emission_amounts` calculation. This is needed if the `pub fn emission_amounts` will be called from an external client.

Motivation and Context
---
The `&'[u64]` return value was at odds with the `emission_decay: &'static [u64]` definition.

How Has This Been Tested?
---
Unit tests
External client

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -142,7 +142,7 @@ impl ConsensusConstants {
     }
 
     /// This gets the emission curve values as (initial, decay, tail)
-    pub fn emission_amounts(&self) -> (MicroTari, &[u64], MicroTari) {
+    pub fn emission_amounts(&self) -> (MicroTari, &'static [u64], MicroTari) {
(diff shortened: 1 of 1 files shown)


stringhandler:
this doesn't seem right

hansieodendaal:
> this doesn't seem right

The definition for `self.emission_decay` is `pub(in crate::consensus) emission_decay: &'static [u64],`, but `&[u64]` is returned for  `self.emission_decay`.
Trying to call `ConsensusConstants::emission_amounts` from an external client in a separate function causes the return value to be dropped prematurely.

stringhandler:
yeah I think the original struct is sus, but this method is fine (when you take the original struct into account)

stringhandler (approved):
utACK 

Not really happy, but this PR is not the problem

CjS77:
Yeah, bad code smell

CjS77:
Oh, this is a reference to the decay parameters, which _are_ in fact static

PR #4860, opened 2022-10-26
fix!: refactor `CipherSeed`, zeroize, and fix key derivation

### LINKED ISSUE #4859: Cipher seed MAC key is derived incorrectly
To generate an encrypted `CipherSeed`, a user-supplied passphrase is used to produce a 64-byte main key. This key is intended to be split into two keys: a 32-byte MAC key, and a 32-byte encryption key. This is implemented incorrectly, such that the last 32 bytes of the main key are [used as the MAC key](https://github.com/tari-project/tari/blob/5b0eb0459c7d29a25339c22a289153d27d57388e/base_layer/key_manager/src/cipher_seed.rs#L320-L332) instead of the first 32 bytes. This means the MAC and encryption keys are identical, which is dangerous in general.

Since the password key derivation function is already run twice, one fix is to run it (using 32-byte output) with a different salt for both keys, by appending a different byte flag to the original salt for each derived key. Another option is to run the key derivation function only once, and parse both derived keys at the same time; however, this might make proper zeroization more challenging (see below).

While we're at it, MAC verification should be done in constant time.

This is also a good opportunity to address the overarching topic in [issue 4846](https://github.com/tari-project/tari/issues/4846) and ensure that secret data produced or handled during `CipherSeed` processing is properly zeroized. While the `CipherSeed` struct itself zeroizes on drop, there are several other locations where passphrase, entropy, or other key data are not.

AaronFeickert:
Description
---
Significantly refactors `CipherSeed` for clarity, zeroizes internal secrets, fixes key derivation, and verifies MACs in constant time. Does _not_ zeroize external passphrases, which should be done separately. Fixes [issue 4859](https://github.com/tari-project/tari/issues/4859).

Motivation and Context
---
As noted in [issue 4859](https://github.com/tari-project/tari/issues/4859), `CipherSeed` MAC keys are derived incorrectly; in fact, the same main key is derived twice to obtain the MAC key and encryption key used for `CipherSeed` encryption. This work refactors for several goals:
- The password is only hashed a single time, so MAC and encryption keys are produced at the same time. This makes it easier to identify that they are being derived correctly, which they now are.
- Internal secrets are zeroized. Notably, passphrases from callers are _not_ yet zeroized, which should be done in subsequent work.
- MACs are verified in constant time using `subtle` functionality.
- The code is much clearer, making it easier to identify secret and non-secret data for review.

How Has This Been Tested?
---
Existing tests, as the external API does not change. Manual inspection that keys are now derived correctly.

BREAKING CHANGE: Existing (version 0) seeds no longer work. Only version 1 seeds are valid.

### DIFF
--- a/base_layer/key_manager/src/cipher_seed.rs
+++ b/base_layer/key_manager/src/cipher_seed.rs
@@ -34,7 +34,6 @@ use rand::{rngs::OsRng, RngCore};
 use serde::{Deserialize, Serialize};
 use subtle::ConstantTimeEq;
 use tari_crypto::hash::blake2::Blake256;
-use tari_utilities::ByteArray;
 use zeroize::{Zeroize, Zeroizing};
 
 use crate::{
@@ -46,10 +45,16 @@ use crate::{
     LABEL_MAC_GENERATION,
 };
 
-const CIPHER_SEED_VERSION: u8 = 0u8;
-// seconds elapsed from unix epoch until '2022-01-01' == 60 * 60 * 24 * 365 * 52
-pub const BIRTHDAY_GENESIS_FROM_UNIX_EPOCH: u64 = 1639872000;
-pub const DEFAULT_CIPHER_SEED_PASSPHRASE: &str = "TARI_CIPHER_SEED";
+// The version should be incremented for any breaking change to the format
+// History:
+// 0: initial version
+// 1: fixed incorrect key derivation and birthday genesis
+const CIPHER_SEED_VERSION: u8 = 1u8;
+
+pub const BIRTHDAY_GENESIS_FROM_UNIX_EPOCH: u64 = 1640995200; // seconds since 2022-01-01 00:00:00 UTC
+pub const DEFAULT_CIPHER_SEED_PASSPHRASE: &str = "TARI_CIPHER_SEED"; // the default passphrase if none is supplied
+
+// Fixed sizes (all in bytes)
 pub const CIPHER_SEED_BIRTHDAY_BYTES: usize = 2;
 pub const CIPHER_SEED_ENTROPY_BYTES: usize = 16;
 pub const CIPHER_SEED_MAIN_SALT_BYTES: usize = 5;
@@ -59,22 +64,23 @@ pub const CIPHER_SEED_ENCRYPTION_KEY_BYTES: usize = 32;
 pub const CIPHER_SEED_MAC_KEY_BYTES: usize = 32;
 pub const CIPHER_SEED_CHECKSUM_BYTES: usize = 4;
 
-/// This is an implementation of a Cipher Seed based on the `aezeed` encoding scheme (https://github.com/lightningnetwork/lnd/tree/master/aezeed)
+/// This is an implementation of a Cipher Seed based on the `aezeed` encoding scheme:
+/// https://github.com/lightningnetwork/lnd/tree/master/aezeed
 /// The goal of the scheme is produce a wallet seed that is versioned, contains the birthday of the wallet,
 /// starting entropy of the wallet to seed key generation, can be enciphered with a passphrase and has a checksum.
 /// The `aezeed` scheme uses a new AEZ AEAD scheme which allows for enciphering arbitrary length texts and choosing
 /// custom MAC sizes. AEZ is unfortunately not available in the RustCrypto implementations yet so we use a similar
 /// AEAD scheme using the primitives available in RustCrypto.
-/// Our scheme must be able to be represented with the 24 word seed phrase using the BIP-39 word lists. The world
+/// Our scheme must be able to be represented with the 24 word seed phrase using the BIP-39 word lists. The word
 /// lists contain 2048 words which are 11 bits of information giving us a total of 33 bytes to work with for the
 /// final encoding.
 /// In our scheme we will have the following data:
 /// version     1 byte
-/// birthday    2 bytes     Days since Unix Epoch
+/// birthday    2 bytes     Days since fixed genesis point
 /// entropy     16 bytes
 /// MAC         5 bytes     Hash(birthday||entropy||version||salt||passphrase)
 /// salt        5 bytes
-/// checksum    4 bytes
+/// checksum    4 bytes     CRC32
 ///
 /// In its enciphered form we will use the MAC-the-Encrypt pattern of AE so that the birthday and entropy will be
 /// encrypted.
@@ -82,8 +88,8 @@ pub const CIPHER_SEED_CHECKSUM_BYTES: usize = 4;
 /// It is important to note that we don't generate the MAC directly from the provided low entropy passphrase.
 /// Instead, the intent is to use a password-based key derivation function to generate a derived key of higher
 /// effective entropy through the use of a carefully-designed function like Argon2 that's built for this purpose.
-/// The corresponding derived key has output of length 64-bytes, and we use the first and last 32-bytes for MAC and
-/// ChaCha20 encryption, respectively. In such way, we follow the motto of not reusing the same derived keys more
+/// The corresponding derived key has output of length 64-bytes, and we use the first and last 32-bytes for
+/// the MAC and ChaCha20 encryption. In such way, we follow the motto of not reusing the same derived keys more
 /// than once. Another key ingredient in our approach is the use of domain separation, via the current hashing API.
 /// See https://github.com/tari-project/tari/issues/4182 for more information.
 ///
@@ -99,9 +105,9 @@ pub const CIPHER_SEED_CHECKSUM_BYTES: usize = 4;
 ///
 /// The checksum allows us to confirm that a given seed phrase decodes into an intact enciphered CipherSeed.
 /// The MAC allows us to confirm that a given passphrase correctly decrypts the CipherSeed and that the version and
-/// salt are not tampered with. If no passphrase is provided a default string will be used
+/// salt are not tampered with. If no passphrase is provided a default string will be used.
 ///
-/// The Birthday is included to enable more efficient recoveries. Knowing the birthday of the seed phrase means we
+/// The birthday is included to enable more efficient recoveries. Knowing the birthday of the seed phrase means we
 /// only have to scan the blocks in the chain since that day for full recovery, rather than scanning the entire
 /// blockchain.
 
@@ -162,7 +168,7 @@ impl CipherSeed {
         let (encryption_key, mac_key) = Self::derive_keys(&passphrase, &self.salt)?;
 
         // Generate the MAC
-        let mut mac = Self::generate_mac(
+        let mac = Self::generate_mac(
             &self.birthday.to_le_bytes(),
             self.entropy.as_ref(),
             CIPHER_SEED_VERSION,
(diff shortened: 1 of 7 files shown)


base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -141,10 +140,11 @@ impl CipherSeed {
         CipherSeed::new_with_birthday(birthday)
     }
 
+    /// Generate a new seed with a given birthday
     fn new_with_birthday(birthday: u16) -> Self {
-        let mut entropy = [0u8; CIPHER_SEED_ENTROPY_BYTES];
-        OsRng.fill_bytes(&mut entropy);
-        let mut salt = [0u8; CIPHER_SEED_SALT_BYTES];
+        let mut entropy = vec![0u8; CIPHER_SEED_ENTROPY_BYTES];
```
AaronFeickert: Does this need a `Zeroizing` wrapper since it's a vector and the eventual initialized `CipherSeed` struct zeroizes on drop?

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
-        let salt_bytes = salt_vec.into_inner().map_err(|_| KeyManagerError::InvalidData)?;
-
         Ok(Self {
-            version: received_version,
-            birthday: decrypted_birthday,
-            entropy: decrypted_entropy,
-            salt: salt_bytes,
+            version,
+            birthday,
+            entropy: (*entropy).clone(),
```
AaronFeickert: See other comment. Same question.

base_layer/key_manager/src/cipher_seed.rs
AaronFeickert: If so, it would need to be dereferenced and cloned (see other comment).

sdbondi: Zeroizing here is not needed, the vector (pointer + length) will be moved into the CipherSeed struct, and the same underlying buffer is zeroized on drop.

```rust
    #[test]
    fn zerioze() {
        let mut seed = CipherSeed::new();
        unsafe {
            let ptr = seed.entropy.as_ptr();
            seed.zeroize();
            for i in 0..(CIPHER_SEED_ENTROPY_BYTES as isize) {
                assert_eq!(*(ptr.offset(i)), 0, "not zero at index {}", i);
            }
        }
    }
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
-
-        let calculated_checksum = crc_hasher.finalize();
-
-        let mut checksum_bytes: [u8; 4] = [0u8; 4];
-        checksum_bytes.copy_from_slice(&checksum_vec[..4]);
-        let checksum = u32::from_le_bytes(checksum_bytes);
-
-        if calculated_checksum != checksum {
+        crc_hasher.update(encrypted_seed.as_slice());
+        let expected_checksum = crc_hasher.finalize().to_le_bytes().to_vec();
```
sdbondi: nit: Allocation not necessary for byte comparison
```suggestion
        let expected_checksum = crc_hasher.finalize().to_le_bytes();
```

base_layer/key_manager/src/cipher_seed.rs
sdbondi: This is all good because of `CipherSeed` zeroize.
If you wanted to save a clone you could 

```rust
let entropy = secret_data.split_off(CIPHER_SEED_BIRTHDAY_BYTES);
// ...
  // Verify the MAC in constant time to avoid leaking data
        if mac.ct_eq(&expected_mac).unwrap_u8() == 0 {
            let _drop = Zeroizing::new(entropy);
            return Err(KeyManagerError::DecryptionFailed);
        }

 Ok(Self {
            version,
            birthday,
            entropy,
            salt,
        })
```
but completely optional, and means we have to be more careful about early function exits.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
 
-        let mut final_seed = vec![CIPHER_SEED_VERSION];
-        final_seed.append(&mut plaintext.to_vec());
-        final_seed.append(&mut self.salt.to_vec());
+        // Assemble the final seed: version, main salt, secret data, checksum
+        let mut encrypted_seed =
+            Vec::<u8>::with_capacity(1 + CIPHER_SEED_MAIN_SALT_BYTES + secret_data.len() + CIPHER_SEED_CHECKSUM_BYTES);
+        encrypted_seed.push(CIPHER_SEED_VERSION);
+        encrypted_seed.append(&mut secret_data.to_vec());
+        encrypted_seed.append(&mut self.salt.to_vec());
```
sdbondi: nit: allocations
```suggestion
        encrypted_seed.extend(&self.salt);
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
+        Self::apply_stream_cipher(&mut secret_data, encryption_key.as_ref(), &self.salt)?;
 
-        let mut final_seed = vec![CIPHER_SEED_VERSION];
-        final_seed.append(&mut plaintext.to_vec());
-        final_seed.append(&mut self.salt.to_vec());
+        // Assemble the final seed: version, main salt, secret data, checksum
+        let mut encrypted_seed =
+            Vec::<u8>::with_capacity(1 + CIPHER_SEED_MAIN_SALT_BYTES + secret_data.len() + CIPHER_SEED_CHECKSUM_BYTES);
+        encrypted_seed.push(CIPHER_SEED_VERSION);
+        encrypted_seed.append(&mut secret_data.to_vec());
```
sdbondi: nit:
```suggestion
        encrypted_seed.append(&mut secret_data);
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
+        encrypted_seed.append(&mut self.salt.to_vec());
 
         let mut crc_hasher = CrcHasher::new();
-        crc_hasher.update(final_seed.as_slice());
-        let checksum = crc_hasher.finalize();
-        final_seed.append(&mut checksum.to_le_bytes().to_vec());
-        Ok(final_seed)
+        crc_hasher.update(encrypted_seed.as_slice());
+        let mut checksum = crc_hasher.finalize().to_le_bytes().to_vec();
+        encrypted_seed.append(&mut checksum);
```
sdbondi: nit:allocations
```suggestion
        let  checksum = crc_hasher.finalize().to_le_bytes();
        encrypted_seed.extend(&checksum);
```

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -155,223 +155,205 @@ impl CipherSeed {
...
+            mac_key.as_ref(),
         )?;
 
-        plaintext.append(&mut mac);
+        // Assemble the secret data to be encrypted: birthday, entropy, MAC
+        let mut secret_data = Zeroizing::new(Vec::<u8>::with_capacity(
+            CIPHER_SEED_BIRTHDAY_BYTES + CIPHER_SEED_ENTROPY_BYTES + CIPHER_SEED_MAC_BYTES,
+        ));
+        secret_data.append(&mut self.birthday.to_le_bytes().to_vec());
+        secret_data.append(&mut self.entropy.clone());
```
sdbondi: self.entropy.clone()  leaks (I believe the compiler won't optimise the clone out because cloned value is being mutated/emptied but not zeroed)
```suggestion
        secret_data.extend(&self.birthday.to_le_bytes());
        secret_data.extend(&self.entropy);
```

sdbondi (commented):
LGTM - some minor comments

hansieodendaal:
utACK

stringhandler:
Nice. Please add a note on how existing seed phrases and wallets are affected

base_layer/key_manager/src/cipher_seed.rs
AaronFeickert: Good point about early exists. I'd much rather ensure that secret data is wrapped as early as possible so we don't have to worry about it.

AaronFeickert:
> Nice. Please add a note on how existing seed phrases and wallets are affected

Done. If you wanted an interesting test of the versioning system, we could bump the `CipherSeed` version number. This would allow for immediate detection of incompatible seeds that can be reported as such. Otherwise, existing seeds will generically fail as manipulated. But that seems a bit much in this case.

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -113,12 +116,13 @@ pub const CIPHER_SEED_MAC_BYTES: usize = 5;
 pub struct CipherSeed {
     version: u8,
     birthday: u16,
-    entropy: [u8; CIPHER_SEED_ENTROPY_BYTES],
-    salt: [u8; CIPHER_SEED_SALT_BYTES],
+    entropy: Vec<u8>,
```
jorgeantonio21: Why should we replace fixed size arrays to `Vec`'s, in this case ? I personally think it is expressive to know in advance the size of both `entropy` and `salt`.

jorgeantonio21:
Very cool ! Thanks for the detailed explanations as well. 

utACK

base_layer/key_manager/src/cipher_seed.rs
AaronFeickert: Using a `Vec` made it easier to handle zeroizing of secret data (due to heap storage), and also should keep the structure more general in case of future updates, where data sizes could change based on version. I could easily imagine a future where all size constants are part of a version-dependent lookup.

AaronFeickert: I'm certainly open to changing back to arrays for now, but it seems like this requires much more careful zeroizing wrappers to account for copies/moves, which seems like a footgun to avoid.

sdbondi: `Box<[u8; CIPHER_SEED_SALT_BYTES]>` is also an option as that allocates a fixed size array on the heap and you move a pointer (Box) around, not the data. But happy to keep it as a vec

jorgeantonio21: Thanks for the explanations @AaronFeickert. I would say that @sdbondi suggestion is the most explicit. I am also happy to use `Vec` here, it is probably not worth spending time on refactoring it it, as of now.

AaronFeickert: The refactoring was pretty straightforward, so I went ahead and switched from `Vec<u8>` to `Box<[u8]>` for clarity on sizes. For some reason `Zeroizing` didn't like `Box<[u8]>`, so there's one spot where we do `Zeroizing<[u8]>` and then create the `Box` later (this is in case of MAC failure, as @sdbondi pointed out elsewhere); it didn't complain when we `#[derive(Zeroize)]` on `Box<[u8]>` though, which is odd...

stringhandler (approved):

PR #4862, opened 2022-10-27
feat!: impl final tari pow algorithm

CjS77:
Tari's independent proof-of-work algorithm is very straightforward.

Calculate the _triple hash_ of the following input data:
 - Nonce (8 bytes)
 - Tari mining hash (32 bytes)
 - PoW record (for Sha-3x, this is always a single byte of value 1)

That is, the nonce in little-endian format, mining hash and the PoW record are chained together and hashed by the Keccak Sha3-256 algorithm. The result is hashed again, and this result is hashed a third time. The result of the third hash is compared to the target value of the current block difficulty.

A triple hash is selected to keep the requirements on hardware miners (FPGAs, ASICs) fairly low. But we also want to avoid making the proof-of-work immediately "NiceHashable". There are several coins that already use a single or double SHA3 hash, and we'd like to avoid having that hashrate immediately deployable against Tari.

This PR also stabilises RFC-0131

BREAKING CHANGE: The number of iterations in the Sha3 PoW has changed


SWvheerden (dismissed):
utACK

stringhandler (dismissed):

stringhandler:
I have merged #4875, which implements this, but I think let's keep this PR open so that we can remove the `if` statement

stringhandler (approved):

Issue #4808, opened 2022-10-15
(RS-W1099) Found manual implementation of a `char::is_ascii_*` method


CjS77:
## Description
To check if a character falls under specific ascii category, consider using the builtin methods, such as, - `char::is_ascii_lowercase`, instead of checking if it belongs to the respective range with `(&#x27;a&#x27;..=&#x27;z&#x27;).contains(_)`, and, - `char::is_ascii_uppercase` - `char::is_ascii_digit` Respectively, for using the builtin method improves readability and cognitive complexity of code.

## Occurrences
There is 1 occurrence of this issue in the repository.

See all occurrences on DeepSource &rarr; [deepsource.io/gh/tari-project/tari/issue/RS-W1099/occurrences/](https://deepsource.io/gh/tari-project/tari/issue/RS-W1099/occurrences/)

PR #4864, opened 2022-10-27
chore: replace manual implementation of char methods

hnidoaht-101:
Description
--
This PR served for fixing the bug described in #4808 

Motivation and Context
--
Following the deepsource link in the issue above to fix.

How Has This Been Tested?
---

Also, its not related to this issue but I noticed some duplication of these crates while checking lints. Take a look on my screenshots below:
image
image

Is the different versions of these crates intentional or did we just not update?
I guess the compile time will be improved if we use the same version for crates.

### DIFF
--- a/base_layer/core/src/consensus/emission.rs
+++ b/base_layer/core/src/consensus/emission.rs
@@ -123,7 +123,7 @@ impl EmissionSchedule {
             if !n.starts_with("0.") {
                 return None;
             }
-            if !n.chars().skip(2).all(|i| ('0'..='9').contains(&i)) {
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
RE argon: Currently used for independent purposes (key manager password hash and wallet grpc password hash) - we should upgrade the keymanager

sdbondi:
ACK

AaronFeickert:
> RE argon: Currently used for independent purposes (key manager password hash and wallet grpc password hash) - we should upgrade the keymanager

Agreed. I updated the key manager version in [PR 4860](https://github.com/tari-project/tari/pull/4860) already.

PR #4865, opened 2022-10-28
feat: improve block_chain backend reliability

### LINKED ISSUE #4850: BaseNode does not reset to old tip after failed reorg sync
When a sync failed, the base node should reset itself back to the old valid tip, and not stay at the height just before the failed blog
[base-node.zip](https://github.com/tari-project/tari/files/9857666/base-node.zip)

SWvheerden:
Description
---
This PR tries to improve db transactions to make sure all changes are atomic and can be reset if need be, also improves logging and comments. 

Motivation and Context
---
We changed the rewind logic to use smaller multiple transactions after we encountered a few too-large transactions. 
Between the too-large transactions and using multiple transactions, if they fail, they might leave the database in an inconsistent state. This PR tries to ensure when adding information, we always use a single transaction and never leave the database in an inconsistent state. 

This PR changes the rewind to try make sure that if it fails, the rewind will try to reset it self back to the previous state. 

How Has This Been Tested?
---

- Manually sync node, and run node
- Unit tests

Fixes: #4850

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1726,8 +1726,7 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
                 // had.
                 error!(
                     target: LOG_TARGET,
-                    "Failed to rewind the block chain, trying to recover the blockchain to previous height: {} ",
-                    e
+                    "Failed to rewind the block chain, trying to recover the blockchain to previous height: {} ", e
                 );
                 let mut txn = DbTransaction::new();
 
@@ -1778,7 +1777,7 @@ fn rewind_block<T: BlockchainBackend>(
     let mut txn = DbTransaction::new();
     info!(target: LOG_TARGET, "Deleting block {}", height);
     let block = fetch_block(db, height, false)?;
-    debug!(target: LOG_TARGET, "Goint to delete block: {}", block);
+    debug!(target: LOG_TARGET, "Going to delete block: {}", block);
     let block = Arc::new(block.try_into_chain_block()?);
     txn.delete_block(*block.hash());
     txn.delete_header(height);
@@ -2096,6 +2095,12 @@ fn insert_orphan_and_find_new_tips<T: BlockchainBackend>(
 
     // There cannot be any _new_ tips if we've seen this orphan block before
     if db.contains(&DbKey::OrphanBlock(hash))? {
+        info!(
+            target: LOG_TARGET,
+            "Orphan #{} ({}) already found in orphan database",
+            block.header.height,
+            hash.to_hex()
+        );
         return Ok(vec![]);
     }
     let mut txn = DbTransaction::new();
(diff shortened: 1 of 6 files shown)


stringhandler:
I'd be happier if there was a unit test showing that this works

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1771,6 +1768,46 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+fn rewind_block<T: BlockchainBackend>(
+    db: &mut T,
+    height: u64,
+    prune_past_horizon: bool,
+    next_block: &ChainHeader,
+) -> Result<Arc<ChainBlock>, ChainStorageError> {
+    let mut txn = DbTransaction::new();
+    info!(target: LOG_TARGET, "Deleting block {}", height);
+    let block = fetch_block(db, height, false)?;
+    debug!(target: LOG_TARGET, "Goint to delete block: {}", block);
```
```suggestion
    debug!(target: LOG_TARGET, "Going to delete block: {}", block);
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1985,11 +2023,19 @@ fn reorganize_chain<T: BlockchainBackend>(
...
-
-        insert_best_block(&mut txn, block.clone())?;
+        txn.delete_orphan(block_hash);
         // Failed to store the block - this should typically never happen unless there is a bug in the validator
         // (e.g. does not catch a double spend). In any case, we still need to restore the chain to a
         // good state before returning.
+        if let Err(e) = insert_best_block(&mut txn, block.clone()) {
+            warn!(
+                target: LOG_TARGET,
+                "Failed extract Monero data: {:?}. Restoring last chain.", e
```
stringhandler, line 2033: This doesn't look like the right error. Is it correct?

stringhandler (commented):
I think these changes are good, but I can't merge this without tests proving the case

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: Yeah it is correct. 
Insert best block can only fail if it fails to deserialise the monero data.

base_layer/core/tests/chain_storage_tests/chain_storage.rs
```diff
@@ -238,6 +238,41 @@ fn test_rewind_to_height() {
     assert_eq!(db.get_height().unwrap(), 1);
 }
 
+#[test]
+fn test_rewind_to_height_failure() {
+    let _ = env_logger::builder().is_test(true).try_init();
+    let network = Network::LocalNet;
+
+    let mut db = TempDatabase::new();
+    db.set_fail_write_counter(12);
```
stringhandler, line 251: need a comment with the significance of 12 here

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: nit: That is an internal detail of insert_best_block though, so the warn log text should be agnostic to that. The `e` can/should describe the problem itself.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1711,46 +1713,41 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
...
+            Err(e) => {
+                // We encountered some error here so lets try to re_add the block we removed to restore the chain we
+                // had.
+                error!(
+                    target: LOG_TARGET,
+                    "Failed to rewind the block chain, trying to recover the blockchain to previous height: {} ", e
+                );
+                let mut txn = DbTransaction::new();
+
+                for block in removed_blocks.into_iter().rev() {
```
sdbondi, line 1733: Wouldn't you call `restore_reorged_chain` here?

sdbondi: Ah i see that function calls rewind_to_hash

SWvheerden:
I am going to close this for now, this is way outdated. This was an attempt at improving reliability when we had a bug in the rewind code. 
This makes many small improvements in the layout of the code. 
BUT this touches core critical code that at this moment works perfectly. I am not in favor of touching working code.

PR #4868, opened 2022-10-28
fix: delete orphans if they exist

### LINKED ISSUE #4867: Base_node sync, should remove blocks from orphan db it syncs and not leave them in the orphan db
[base-node.zip](https://github.com/tari-project/tari/files/9889114/base-node.zip)
Kinda linked to #4866 , but not fully. As the title says block sync should clear from orphan db if they exist. 

This is the example of it happening:
```
ERROR Failed to handle incoming block message: Comms interface error: `Chain storage error: The requested Orphan accumulated data was not found via hash:642b34efbcd590e08769219738e86cf95da2a1280be671a365ebc882c181e801 in the database
```

SWvheerden:
Description
---
Orphans should be deleted from the orphan database if they exist. 

Motivation and Context
---
If the block is added to the main chain, it should not exist in the orphan database as well. It will also remove the orphan tips if it removes the block from the pool. 

How Has This Been Tested?
---
Unit test and manually running

fixes: #4867

### DIFF
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -307,6 +307,7 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
                     if let Err(err) = self
                         .db
                         .write_transaction()
+                        .delete_orphan(header_hash)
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Nice

sdbondi:
ACK

Did a basic rewind and sync test

PR #4876, opened 2022-11-02
fix(ci): correct ARM64 builds

leet4tari:
Description:
Remove selective hack that reverted resolver from 1 to 2. Fix ARM64 builds by removing env overrides in Cross.toml breaking openssl-sys

Motivation and Context:
Make ARM64 builds standard

How Has This Been Tested?
Build test in locally and in local fork


stringhandler (approved):
Nice

PR #4877, opened 2022-11-02
fix(comms/peer_manager): fix possible panic in offline calc

sdbondi:
Description
---
- fixes possible panic in `Peer::offline_since` 
- offline_at has second resolution

Motivation and Context
---
Possible panic when converting chrono Duration to std Duration
second resolution is good enough for offline_since

How Has This Been Tested?
---
Updated unit test for this case


stringhandler (approved):

stringhandler:
utACK

SWvheerden:
utACK

PR #4881, opened 2022-11-03
feat!: add tari address for wallet to use

### LINKED ISSUE #4886: Wallet should use checkable address when sending transactions
Currently, the wallet only sends to a public key. 
Because in Tari we have mostly interactive transactions, wallets need to communicate directly to the other party before sending transactions. This means that wallets can verify the address of the recipient beforehand and know it's a valid address on the same network and is a wallet and not a base node.

But we do have one-sided transactions as well as stealth addresses. Here the address is not validated before sending. 

We need some addressing scheme to addresses so that wallets know they are talking to a wallet, as well as the correct network.

### LINKED ISSUE #4888: Wallet sql file still has old removed DANv1 fields in it
These fields need to be removed:
```
features_parent_public_key
features_unique_id
contract_id
```
These where removed in: #4463 
They were accidentally re-added

SWvheerden:
Description
---
This changes the address used by the wallet to not just use the pubkey of another wallet, to a new custom address scheme.

Changes:
- Includes new address scheme : [version][network][pub_key]
- Replaces wallet pubkey with new address all over
- Calculate what address should be is message is received from network public key. 
- Cleans up database to remove broken fields and defunct tables

Motivation and Context
---
We need the wallets to be able to verify that the addresses used by the wallets are for the correct network and are wallets. In interactive transactions, transactions using the wrong pubkeys, will result in transactions eventually being self-canceled by the wallet as they are stale. But one-sided and stealth these will be sent and never recoverable. 

Using the new scheme wallets can verify addresses are on the correct network, and because base_nodes still use pubkeys the addresses differ between base_nodes and wallets. 

How Has This Been Tested?
---
All unit tests passed

Fixes: #4888 
Fixes: #4886 

BREAKING: This changes how the wallet accepts and presents addresses

### DIFF
--- a/.license.ignore
+++ b/.license.ignore
@@ -12,6 +12,7 @@
 ./applications/tari_base_node/osx-pkg/scripts/preinstall
 ./applications/tari_console_wallet/linux/start_tari_console_wallet
 ./base_layer/key_manager/Makefile
+./base_layer/wallet/src/schema.rs
 ./base_layer/p2p/src/dns/roots/tls.rs
 ./buildtools/docker/torrc
 ./docs/src/theme/book.js
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2877,9 +2877,9 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-src"
-version = "111.22.0+1.1.1q"
+version = "111.24.0+1.1.1s"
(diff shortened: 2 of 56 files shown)


stringhandler (dismissed):

stringhandler:
Needs BREAKING CHANGE footer please

stringhandler:
utACK

stringhandler (approved):

Issue #4870, opened 2022-10-31
Check median timestamp validation is done on wrong section


SWvheerden:
Currently, the validation `check_median_timestamp` is done on orphan headers, which should only be on full main headers added to the main chain. 

On Tari, we use two types of validators, one for orphans, and one for the main chain. Orphan validators only check rudimentary data, like does the block balance equation work out, does the block have one coinbase etc. Main validation checks all validation steps that are dependent on previous chain blocks, like if each input is a valid out, is the PoW valid, etc. 

`check_median_timestamp` is dependent on previous chain headers. In order to ensure the blockchain moves forward in time, there is a median timestamp rule that specifies that the header's timestamp must be greater than the median timestamp of the previous 11 blocks.

This check has no meaning on the orphan blocks, we need to check this as a main check on the validation

CjS77:
Is it currently checked on both orphans _and_ the main chain, or orphans only?

SWvheerden:
both

CjS77:
Ok, so not a chronic trainsmash. The issue is to remove the unnecessary check on orphan blocks.

CjS77:
It's not even consensus-changing, iiuc.

SWvheerden:
Yeah. 
It can happen that an orphan block fails validation because it's looking at the wrong data. Thats the only issue here

stringhandler:
Adding the consensus breaking label because it could result in two nodes validating the block differently

PR #4887, opened 2022-11-04
feat: timestamp validation

Cifko:
Description
---
Disable timestamp validation for orphan.
Enable during orphan new tips creation. It should be checked during the new tips creation so we don't select invalid tip in favor of lower difficulty but valid tip.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4408,7 +4408,7 @@ dependencies = [
 
 [[package]]
 name = "tari_app_grpc"
-version = "0.40.1"
+version = "0.40.2"
 dependencies = [
  "argon2",
  "base64 0.13.1",
@@ -4433,7 +4433,7 @@ dependencies = [
 
 [[package]]
 name = "tari_app_utilities"
-version = "0.40.1"
+version = "0.40.2"
 dependencies = [
  "clap 3.2.23",
  "config",
@@ -4455,7 +4455,7 @@ dependencies = [
 
 [[package]]
 name = "tari_base_node"
-version = "0.40.1"
+version = "0.40.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4552,7 +4552,7 @@ dependencies = [
 
 [[package]]
 name = "tari_common"
-version = "0.40.1"
+version = "0.40.2"
 dependencies = [
  "anyhow",
  "blake2 0.9.2",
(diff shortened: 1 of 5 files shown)


stringhandler:
Related to #4870

stringhandler (commented):
NACK. I don't think this is what @SWvheerden had in mind when logging the issue

SWvheerden (commented):
utACK

I had an easier fix in mind with just not running this with the header validation. And only running this when we actually add the block, but this is fine. 

I do however want to test run this for a day to see if it works as intended.

SWvheerden:
I did manage to sync one and tan through the night without any issues so far.

Cifko:
I'm going through the code. And I don't see any validation of the median timestamp outside of the orphan tree construction. In the whole code we validate the median timestamp at two location, header sync and orphan tree construction. I can add the timestamp validation if the block added is not an orphan. But once we add anything from the stored orphan block there is no validation. Currently there is some validation that is not consistent across nodes.
When we decide that some orphan chain will be added we do only body validation. I can't find the header validation. Any tips where it is, or what direction should I take now?

SWvheerden:
I would just add the check here:
https://github.com/tari-project/tari/blob/55d133494270fe92ab7cf48d58f18d2a2bdecd17/base_layer/core/src/validation/block_validators/body_only.rs#L58

This should be the last validator that's gets called right before a block is added to the database.

SWvheerden (commented):
utACK

This is only done for block propagation and not syncing. 
But I think this should be fine as header sync will add and validate blocks in order. 

But perhaps just in terms of keeping header/block sync and block propagation validation in sync, we should make the the same changes there?

Cifko:
> utACK
> 
> This is only done for block propagation and not syncing. But I think this should be fine as header sync will add and validate blocks in order.
> 
> But perhaps just in terms of keeping header/block sync and block propagation validation in sync, we should make the the same changes there?

Added. But IMHO I would not go this way. The sync has the validation in the headersync where it belongs. Now we do it double.

SWvheerden (commented):
utACK

sdbondi:
Yeah, agree with @Cifko, the async block validator does not need to validate headers, and that is efficiently done in the header sync by knowing that it will receive sequential headers. 

That said, the perf hit is minor so this could go in as is

stringhandler (dismissed):
Can be merged once CI passes

sdbondi (dismissed):

sdbondi (approved):

sdbondi:
utACK

Issue #4846, opened 2022-10-24
Ensure that KDF data is zeroized


AaronFeickert:
The codebase includes several KDFs that handle secret data. Even though underlying input data (like secret keys or ECDH shared secrets) may be zeroized, it should be checked that subsequent KDF output data is also zeroized.

jorgeantonio21:
Would love to tackle this one :)

AaronFeickert:
For KDFs that don't already produce a zeroizing type, there are options:
- Wrap the result in a `Zeroizing` type to ensure it's zeroized on drop, and check for copies/moves
- Produce a custom array type for this purpose, ensure it zeroizes on drop (`zeroize` has a custom macro for this), and check for copies/moves

AaronFeickert:
After discussion with @jorgeantonio21, it was suggested that one approach might be to build a generic design similar to that of [SafePassword](https://github.com/tari-project/tari_utilities/blob/main/src/password.rs). This would wrap the key type `T` (either a `u8` array or some other type) in a `Hidden<Box<T>>` wrapper that ensures the secret data stays in one place on the heap and won't accidentally be displayed or tossed into a debug log. We can then zeroize on drop via an `impl Drop` that zeroizes via the mutable reference accessible in `Hidden` (and which will work even on clones). Finally, we can ensure that access to the secret data is limited by only exposing it by reference (or mutable reference, which may be necessary if we need to generate the output in place). If a caller needs ownership, we either refactor the caller (safer from a data perspective, but likely more engineering risk) or implement an `into_inner` that consumes the data and returns it (which requires that we modify the caller to ensure it zeroizes its new data).

Such a generic design should ensure that different KDF output types are properly enforced by the type system, even if the underlying key types are the same. Even though the KDFs themselves (should) enforce domain separation, enforcing by type should help to keep design intent more clear.

AaronFeickert:
Maybe the type differentiation could be done in a similar fashion to the hashing API, where a macro is used to generate a struct that's used during instantiation. So we could have a macro that generates an effectively empty struct implementing some `KDFOutputType` marker trait. Then, we build `KDFOutput<T: Zeroize, K: KDFOutputType>`, where `T` is the underlying key type and `K` is stored as `PhantomData`. The `SafePassword`-like hiding and zeroizing and reference access functionality then goes into `KDFOutput`.

AaronFeickert:
[...]

PR #4890, opened 2022-11-04
fix: use hidden to mask seed words

### LINKED ISSUE #4861: `Mnemonic::to_mnemonic` should wrap the seed words in Zeroizing
https://github.com/tari-project/tari/blob/5b0eb0459c7d29a25339c22a289153d27d57388e/base_layer/key_manager/src/mnemonic.rs#L254

The TariSeedWords ffi struct may need to be updated, but I don't think this will cause a breaking change as it will not change the underlying representation

brianp:
Description
---
Don't let seed words leak into logfiles or other places. Mask them and only reveal them when specifically required.
I didn't use zeroize because I know we've had some issues with it and `Hidden` seemed like it would do the job well enough. Happy to change it if there are any strong preferences.

Motivation and Context
---
Prevent leaked secrets.

How Has This Been Tested?
---
Manually checking to ensure seed words can be returned and written to disk without a problem.

Closes: #4861


stringhandler (dismissed):

stringhandler:
utACK

stringhandler (dismissed):

base_layer/key_manager/src/cipher_seed.rs
```diff
@@ -34,6 +34,7 @@ use rand::{rngs::OsRng, RngCore};
 use serde::{Deserialize, Serialize};
 use subtle::ConstantTimeEq;
 use tari_crypto::hash::blake2::Blake256;
+use tari_utilities::{ByteArray, Hidden};
```
stringhandler:
```suggestion
use tari_utilities::Hidden;
```

AaronFeickert:
What issues specifically are of concern with `zeroize`? It is true that [moves and copies can be problematic](https://docs.rs/zeroize/latest/zeroize/#stackheap-zeroing-notes), but `Hidden` appears to do nothing for leaving in-memory traces of sensitive data.

sdbondi:
Agree with @AaronFeickert that we should be using the Zeroizing wrapper. The problems with moving are only tricky for data that lives on the stack. With `Vec<String>` all sensitive data is heap-allocated Zeroizing will zeroize [each element in the vec](https://github.com/RustCrypto/utils/blob/b898588b13e8005cd8860089f108bed09f86bd87/zeroize/src/lib.rs#L541).

brianp:
> What issues specifically are of concern with zeroize?

I believe the problem wasn't how it did anything but how it's being loaded and places it caused problems across crates.

I'm happy to move it to zeroize right now as I hadn't realized that key difference. As that's an important feature for us, maybe we should expand `Hidden` at some point.

stringhandler:
Maybe `Hidden` should implement `Zeroize`?

sdbondi:
SGTM, Hidden makes sense for seed words, in that you don't want any derive Debug accidents and `Hidden<T>` by definition contains sensitive data so should zeroize on drop for `T: Zeroize`.

CjS77:
The convo is inferring a lot of composability. 

Seed words _are_ hidden, seed words _are_ zeroizable etc.

This suggests that there's value in defining a `SeedWord` type:

`#[derive(zeroize)] pub struct SeedWord(Hidden<String>)`

CjS77:
> SGTM, Hidden makes sense for seed words, in that you don't want any derive Debug accidents and `Hidden<T>` by definition contains sensitive data so should zeroize on drop for `T: Zeroize`.

Also good. Perhaps the Zeroize trait bound should be compulsory, i.e. defined on the struct rather than the impl

AaronFeickert:
So would the overall idea be to add a `Zeroize` trait bound to `Hidden`, and then create the new `SeedWord` type that zeroizes on drop? It looks like `Hidden` implements `Copy`, so as I understand it you couldn't zeroize it on drop.

brianp:
`Hidden` isn't widely used yet, so if the zeroized version of it becomes the standard I think that's fine if that's the expectation of the struct, and it's inability to be copied.

Looking closer everywhere we make use of the seed words we end up creating a new variable with the seed words mutated into a single `String` using `join`. Even if we get `Hidden` zeroized correctly, we're still leaving traces of them around every time we make use of them. 
This may be solved with a new seed words struct or could be easy to return the combined string from the source instead of a `Vec<String>`. This rolls into your point about Copy no longer being implemented but it's probably a step in the right direction considering the existing possibility of leaks. I'll also have to look a bit closer around the mnemonic functions and ensure we're not pulling out single words getting left around there as well.

AaronFeickert:
A wrapper design similar to `SafePassword` might do the trick here, similar to what @CjS77 wrote up. You can have it contain a `Hidden<Vec<String>>`, zeroize on drop with a custom `impl Drop` using the mutable reference provided by `Hidden`, expose individual words by reference, and expose the joined word by reference too.

Even though `Hidden` lets you extract the treasure within via `into_inner`, I like the `SafePassword` approach of not doing this. It means callers need to deal with references whose values are safely locked away in a single spot. If there are cases where the caller can't operate by reference, you could either refactor so they can, or implement `into_inner` and then carefully ensure that the caller zeroizes it properly.

AaronFeickert:
I was playing around with a similar idea for another issue dealing with [KDF output](https://github.com/tari-project/tari/issues/4846), and came up with [this fairly generic design](https://github.com/tari-project/tari_utilities/pull/52). A similar non-generic version might be useful for a `SeedWord` struct.

stringhandler:
I prefer making `Hidden` implement `zeroize` to the extra `clandestine` struct (although the name is great). It seems in line with the intention of the struct, and users may expect it to `zeroize`. I was surprised to find out it didn't require it already, for example. 

I do like the idea of a `SeedWords` struct as well, though. Strings should not be passed around containing these words

AaronFeickert:
Just adding `Zeroize` (and zeroize on drop) to `Hidden` still retains the copy and move issues present with any stack allocation, right? Ensuring we also always have a `Box` wrapper makes it easier to assert that zeroization does what it says on the tin.

I also favor the type enforcement present with something like `Clandestine`, which makes it impossible to use data with the same underlying type in the wrong context.

AaronFeickert:
> I prefer making `Hidden` implement `zeroize` to the extra `clandestine` struct (although the name is great). It seems in line with the intention of the struct, and users may expect it to `zeroize`. I was surprised to find out it didn't require it already, for example.
> 
> I do like the idea of a `SeedWords` struct as well, though. Strings should not be passed around containing these words

The [linked PR](https://github.com/tari-project/tari_utilities/pull/52) has been updated to do just this. Now `Hidden` directly supports zeroizing on drop, display/debug hiding, and optional type enforcement. It works seamlessly with `SafePassword` and could likely be used in a similar way for `SeedWords`.

stringhandler:
Closing in favour of #4925

PR #4897, opened 2022-11-07
feat: mempool sync wait for node initial sync

### LINKED ISSUE #4785: Remove mempool sync on startup
Currently, when the base node starts up, it syncs the mempool. This is mostly useless since the base node will probably not be synced and the transactions will be invalid.

Also, recently there has been a change to sync the mempool after a significant block sync. 

I suggest removing it on start up, or at least only running it after the first sync after start up.

Cifko:
Description
---
Wait for mempool sync to start after the intial sync is done.
Fixes #4785

How Has This Been Tested?
---
Manually.

### DIFF
--- a/base_layer/core/src/mempool/sync_protocol/initializer.rs
+++ b/base_layer/core/src/mempool/sync_protocol/initializer.rs
@@ -102,7 +102,7 @@ impl ServiceInitializer for MempoolSyncInitializer {
                         target: LOG_TARGET,
                         "Mempool sync still on hold, waiting for node to do initial sync",
                     );
-                    sleep(Duration::from_secs(1)).await;
+                    sleep(Duration::from_secs(30)).await;
                 }
                 log_mdc::extend(mdc.clone());
             }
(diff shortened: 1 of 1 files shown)


base_layer/core/src/mempool/sync_protocol/initializer.rs
```diff
@@ -86,17 +86,21 @@ impl ServiceInitializer for MempoolSyncInitializer {
...
+                            "Initial sync is done. Starting mempool sync protocol"
+                        );
                         break;
                     }
                     trace!(
                         target: LOG_TARGET,
-                        "Mempool sync still on hold, waiting for bootstrap to finish",
+                        "Mempool sync still on hold, waiting for node to do initial sync",
                     );
                     sleep(Duration::from_secs(1)).await;
```
stringhandler:
```suggestion
                    sleep(Duration::from_secs(30)).await;
```

stringhandler (approved):

sdbondi:
utACK

SWvheerden (commented):
While I dont think this fixes the issue completely, it will still run on startup, it will only wait till the node says it is synced. I think this is the better approach. 

It might be that the node was off, and there were no new blocks added or not enough to trigger a mempool sync. Coming back on again, the node will ask for new transactions. 

If we really wan to stop this from running on startup as per issue https://github.com/tari-project/tari/issues/4785 then we need to make the counter for the mempool the max amount on startup. 
Base_node sync will then reset that down to 0 when required

stringhandler:
I'm happy with this approach for now

SWvheerden:
utACK

PR #4898, opened 2022-11-07
feat: add consensus encoding for signed ints

Cifko:
Description
---
Add consensus encoding for signed integerss.

Motivation and Context
---
I'm doing consensus encoding in tari-dan for the VoteMessage. And one of the fields is i64


sdbondi:
You can use [Borsh](https://borsh.io/) on the DAN layer, it serves the same purpose as ConsensusEncoding and can be used in signature challenges

stringhandler:
I think let's hold off on this until it's clear that we need them

PR #4905, opened 2022-11-09
fix: force wallet sqlite to do checkpoint after db decryption

### LINKED ISSUE #4904: Wallet FFI remove_encryption does not block while busy
When the FFI calls the remove_encryption function, it does not block on the mobile side. 
Calling the function does not produce a decrypted function. 
Putting a sleep in the function does produce a decrypted function.

SWvheerden:
Description
---
Forces the database to do a checkpoint after the last step of the database decryption. 

Motivation and Context
---
The mobile applications decrypt the SQLite wallet database, after which they copy it to a safe location to restore later.  
If we run database decrypt, the actual database file might not yet have the changes applied to it, this forces a checkpoint which will ensure that the changes are committed to on file

How Has This Been Tested?
---
manual

Fixes: #4904

### DIFF
--- a/base_layer/wallet/src/storage/sqlite_db/wallet.rs
+++ b/base_layer/wallet/src/storage/sqlite_db/wallet.rs
@@ -559,6 +559,8 @@ impl WalletBackend for WalletSqliteDatabase {
                 .map_err(|e| WalletStorageError::ConversionError(e.to_string()))?;
             WalletSettingSql::new(DbKey::TorId.to_string(), tor_string).set(&conn)?;
         }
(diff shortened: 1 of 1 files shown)


sdbondi:
Ah great so that was the problem? 

What does (2) mean? I do not see it in the docs
https://www.sqlite.org/pragma.html#pragma_wal_checkpoint

SWvheerden:
https://www.sqlite.org/c3ref/c_checkpoint_full.html

stringhandler (approved):

Issue #4807, opened 2022-10-14
new Wallet starts scanning at height 0 and not birthday


SWvheerden:
When a new wallet is started the UTXO scanning service starts scanning at height 0, it should start scanning at the birthday. 

```
DEBUG Scanning UTXO's from height = 0 to current tip_height = 22593 (starting header_hash: 967c4051737b2148edeed2746dba443376738bfb3c6452942c24623943e54cd2)"
```

PR #4911, opened 2022-11-11
fix: set wallet start scan height to birthday and not 0 (see issue #4807)

jorgeantonio21:
Description
---
When wallet starts new UTXO scanning service it should start at the birthday. Currently, it starts at height 0. This PR addresses this issue

Motivation and Context
---
Addresses #4807.

How Has This Been Tested?
---


base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -325,7 +325,11 @@ where
         client: &mut BaseNodeWalletRpcClient,
     ) -> Result<BlockHeader, UtxoScannerError> {
         let tip_info = client.get_tip_info().await?;
-        let chain_height = tip_info.metadata.map(|m| m.height_of_longest_chain()).unwrap_or(0);
+        let wallet_birthday_height = self.get_birthday_header_height_hash(client).await?.height;
```
SWvheerden: The old code is correct. 
If the wallet cant get the network tip from the base_node its connected to it needs to assume its 0. 
It also very unlikely that wallet wont have a tip, but has random height header?

SWvheerden (changes requested):
We need to implement this logic when the node selects its starting height for a scan, not the ending scan. 
The end scan should always be the network tip.

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
jorgeantonio21: Great point ! You are right :)

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -66,6 +66,7 @@ use crate::{
     },
 };
 
+pub const BIRTHDAY_GENESIS_FROM_UNIX_EPOCH: u64 = 1640995200; // seconds since 2022-01-01 00:00:00 UTC
```
AaronFeickert: This shouldn't be a hard-coded value. It needs to match the [value](https://github.com/tari-project/tari/blob/43efc07a9713ebb498ca635603b97c34b561b1e2/base_layer/key_manager/src/cipher_seed.rs#L54) corresponding to the `CipherSeed` version in question (of which there is currently only one).

AaronFeickert (changes requested):

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
jorgeantonio21: I was actually playing around to test it. The best approach, imo, is to add a method to get the birthday date from unix epoch, like `get_birthday_time_from_unix()`.

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -699,9 +700,10 @@ where
         client: &mut BaseNodeWalletRpcClient,
     ) -> Result<HeightHash, UtxoScannerError> {
         let birthday = self.resources.db.get_wallet_birthday()?;
-        // Calculate the unix epoch time of two days before the wallet birthday. This is to avoid any weird time zone
-        // issues
-        let epoch_time = u64::from(birthday.saturating_sub(2)) * 60 * 60 * 24;
+        // Calculate the unix epoch time of two days before the wallet birthday.
+        // This is to avoid any weird time zone issues
+        let epoch_time = get_birthday_from_unix_epoch(birthday, 2u16);
```
SWvheerden: I would go for 2 weeks, rather than 2 days. 
This is to go around re-org issues, we should rather error on the safe side/

SWvheerden (dismissed):
Looks good to me, I would just up the time to be longer to error on the safe side

SWvheerden (dismissed):
utACK

base_layer/key_manager/src/lib.rs
```diff
@@ -30,3 +31,7 @@ pub(crate) fn mac_domain_hasher<D: Digest + LengthExtensionAttackResistant>(
 ) -> DomainSeparatedHasher<D, KeyManagerDomain> {
     DomainSeparatedHasher::<D, KeyManagerDomain>::new_with_label(label)
 }
+
+pub fn get_birthday_from_unix_epoch(birthday: u16, to_days: u16) -> u64 {
+    u64::from(birthday.saturating_sub(to_days)) * 24 * 60 * 60 + BIRTHDAY_GENESIS_FROM_UNIX_EPOCH
+}
```
AaronFeickert: Nit: recommend a function-level comment for this, to indicate its purpose. Might also be helpful to rename it in a way that indicates an extra offset is present and clearly describes the unit conversion.

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -697,9 +698,10 @@ where
         client: &mut BaseNodeWalletRpcClient,
     ) -> Result<HeightHash, UtxoScannerError> {
         let birthday = self.resources.db.get_wallet_birthday()?;
-        // Calculate the unix epoch time of two days before the wallet birthday. This is to avoid any weird time zone
-        // issues
-        let epoch_time = u64::from(birthday.saturating_sub(2)) * 60 * 60 * 24;
+        // Calculate the unix epoch time of two weeks (14 days) before the wallet birthday.
+        // This is to avoid any weird time zone issues
```
AaronFeickert: Time zone issues?

base_layer/wallet/tests/utxo_scanner.rs
```diff
@@ -288,7 +288,7 @@ async fn test_utxo_scanner_recovery() {
     let mut test_interface = setup(UtxoScannerMode::Recovery, None, None, None).await;
 
     let cipher_seed = CipherSeed::new();
-    let birthday_epoch_time = u64::from(cipher_seed.birthday() - 2) * 60 * 60 * 24;
+    let birthday_epoch_time = get_birthday_from_unix_epoch(cipher_seed.birthday(), 14u16);
```
AaronFeickert: Nit: recommend a comment clearly indicating the units, since we switch between seconds and days.

AaronFeickert (dismissed):
Just a few nits for clarity, particularly because of unit conversions between the `CipherSeed` birthday and Unix time.

Are there tests that exercise the expected behavior? Namely:
- a birthday corresponding to a block that is not the genesis block
- a birthday corresponding to the genesis block

If not, recommend adding them if feasible.

Otherwise, did not specifically test, but the changes to the math appear to be doing what you want.

AaronFeickert:
Any reason for the large changes to `Cargo.lock`? Seems like a lot.

AaronFeickert:
May be useful to include a couple of unit tests to hit the saturating subtraction cases.

base_layer/key_manager/src/lib.rs
jorgeantonio21: I added some documentation and add `_in_seconds` to the naming of the function

base_layer/wallet/tests/utxo_scanner.rs
jorgeantonio21: Definitely !

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
jorgeantonio21: Actually, the main reason is to avoid any possible reorgs. I updated the comment to convey this more precisely.

jorgeantonio21:
Added unit tests and documentation that convey, more precisely, the points made by @AaronFeickert. I tried to make unit tests to be as expressive as possible. In particular, we use hardcoded values for second duration from specific dates. Also added a test on a mocked chain to check consistency of birthday and block timestamps, altogether.

stringhandler (approved):

PR #4919, opened 2022-11-15
fix(comms): spawn liveness check after address is final

### LINKED ISSUE #4915: Comms liveness check has "Address contained no components." error after creating new base node identity
```
🔌️ Failed to dial public address for self check: Failure due to invalid target address: Address contained no components.
```

Steps to reproduce:
1. Create a new base node with a new identity
2. `Address contained no components.` appears in status line periodically
3. Restarting base node fixes it

Probably a new onion address is not set before being passed to the liveness check

sdbondi:
Description
---
- Spawns the liveness check after the public address has been finalised

Motivation and Context
---
Fixes #4915 and fixes an edge case where the old address is used for liveness checks if the user deletes their tor identity before starting the node.

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_base_node/src/commands/command/status.rs
+++ b/applications/tari_base_node/src/commands/command/status.rs
@@ -131,7 +131,7 @@ impl CommandContext {
             ),
         );
 
-        match self.comms.listening_info().liveness_status() {
+        match self.comms.liveness_status() {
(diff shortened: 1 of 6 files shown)


stringhandler (approved):

stringhandler:
ACK

PR #4920, opened 2022-11-15
feat: replace consensus with borsh

Cifko:
Description
---
Don't mind for now the Cargo.toml links to my repos. It's just for testing.

Motivation and Context
---

How Has This Been Tested?
---


base_layer/core/src/covenants/covenant.rs
```diff
@@ -39,42 +39,38 @@ use crate::{
...
+}
 
-    pub(super) fn write_to<W: io::Write>(&self, writer: &mut W) -> Result<(), io::Error> {
-        CovenantTokenEncoder::new(self.tokens.as_slice()).write_to(writer)
+impl BorshDeserialize for Covenant {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let len = u32::deserialize(buf)?;
+        let mut tokens = Vec::with_capacity(len as usize);
+        for _ in 0..len {
+            tokens.push(BorshDeserialize::deserialize(buf)?);
```
sdbondi: For covenants and tari script, I would rather use the encoding provided since they are defined in their respective RFCs.

The type `&[u8]` implements `std::io::Read` so nothing much needs to change from the ConsensusEncoding impls

```rust
 let len = u32::deserialize(buf)? as usize;
        if len == 0 {
            return Ok(Covenant::new());
        };
        // Check the length  - this may be maliciously misreported
        if len > MAX_COVENANT_BYTES {
            return Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                "covenants: Length varint exceeded maximum",
            ));
        }
        // Ensure that no more than the maximum bytes can be read
        let mut limited = LimitedBytesReader::new(MAX_COVENANT_BYTES, buf);
        CovenantTokenDecoder::new(&mut limited)
            .collect::<Result<_, CovenantDecodeError>>()
            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))
```

sdbondi: Serialize impl

```rust
 fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
         let len = u32::try_from(self.get_byte_length()).map_err(....)?; 
        writer.write(len)?;
        self.write_to(writer)?;
        Ok(())
}
```

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -61,13 +58,12 @@ where D: Digest<OutputSize = U32>
         self.writer.0.finalize().into()
     }
 
-    pub fn update_consensus_encode<T: ConsensusEncoding + ?Sized>(&mut self, data: &T) {
-        // UNWRAP: ConsensusEncode MUST only error if the writer errors, HashWriter::write is infallible
-        data.consensus_encode(&mut self.writer)
-            .expect("Incorrect implementation of ConsensusEncoding encountered. Implementations MUST be infallible.");
+    pub fn update_consensus_encode<T: BorshSerialize>(&mut self, data: &T) {
+        BorshSerialize::serialize(data, &mut self.writer)
+            .expect("Incorrect implementation of BorshSerialize encountered. Implementations MUST be infallible.");
```
sdbondi, line 63: Unfortunately, this is not guaranteed by Borsh e.g. https://github.com/near/borsh-rs/blob/master/borsh/src/ser/mod.rs#L230. I believe the only case where this can fail in practice is if you have a vec/list of length > u32::MAX but we no longer have the guarantees that ConsensusEncoding had.

A hasher that can fail is pretty inconvenient `block_header.hash().map_err(...)` but that might be the only way to do it. 

I think you can leave the expect in for now, and we'll need to make another refactor for the hash() methods

base_layer/core/src/covenants/covenant.rs
Cifko: The Covenant (I haven't checked the TariScript) uses the consensus encoding in some parts (e.g.CovenantArg). That should be replace by borsh, right? Just checking, because that will slightly change the outputs.

base_layer/core/src/covenants/arguments.rs
```diff
@@ -65,37 +70,42 @@ impl CovenantArg {
...
                 Ok(CovenantArg::Hash(hash.into()))
             },
             ARG_PUBLIC_KEY => {
-                let pk = PublicKey::consensus_decode(reader)?;
+                let pk = PublicKey::deserialize(reader)?;
                 Ok(CovenantArg::PublicKey(pk))
             },
-            ARG_COMMITMENT => Ok(CovenantArg::Commitment(Commitment::consensus_decode(reader)?)),
+            ARG_COMMITMENT => {
+                println!("here {:?}", reader);
```
sdbondi:
```suggestion

```

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -39,6 +38,25 @@ pub struct FixedByteArray {
...
+        Ok(())
+    }
+}
+
+impl BorshDeserialize for FixedByteArray {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let elems: Vec<u8> = BorshDeserialize::deserialize(buf)?;
+        let len = BorshDeserialize::deserialize(buf)?;
+        Ok(Self {
+            elems: elems.as_bytes().try_into().unwrap(),
```
sdbondi: unwrap

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -39,6 +38,25 @@ pub struct FixedByteArray {
...
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        BorshSerialize::serialize(&self.elems.to_vec(), writer)?;
+        BorshSerialize::serialize(&self.len, writer)?;
+        Ok(())
+    }
+}
+
+impl BorshDeserialize for FixedByteArray {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let elems: Vec<u8> = BorshDeserialize::deserialize(buf)?;
```
sdbondi: Unbounded allocation controlled by an untrusted source.
I think this works (you'll need to update the serialize to write the length first)

```rust
 fn deserialize(reader: &mut &[u8]) -> io::Result<Self> {
        let mut buf = [0u8; 1];
        reader.read_exact(&mut buf)?;
        let len = buf[0] as usize;
        if len > MAX_ARR_SIZE {
            return Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                format!("length exceeded maximum of 64-bytes for FixedByteArray: {}", len),
            ));
        }
        let mut ret = FixedByteArray::new();
        for _ in 0..len {
            let mut buf = [0u8; 1];
            reader.read_exact(&mut buf)?;
             // PANIC: push cannot happen because len has been checked
            ret.push(buf[0]);
        }
        Ok(ret)
    }
```

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
```diff
@@ -180,35 +147,9 @@ mod test {
         FixedByteArray::from_bytes(&[1u8; 64][..]).unwrap_err();
     }
 
-    #[test]
-    fn length_check() {
```
sdbondi, line 184: These checks must pass, except for the varint check

base_layer/core/src/proof_of_work/monero_rx/merkle_tree.rs
```diff
@@ -132,6 +131,38 @@ pub struct MerkleProof {
     path_bitmap: u32,
 }
 
+impl BorshSerialize for MerkleProof {
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        BorshSerialize::serialize(
+            &(u32::try_from(self.branch.len()).map_err(|_| ErrorKind::InvalidInput)?).to_le_bytes(),
```
sdbondi: Borsh should be able to serialize a u32 without first converting it to bytes
```suggestion
            &u32::try_from(self.branch.len()).map_err(|_| ErrorKind::InvalidInput)?
```

base_layer/core/src/proof_of_work/monero_rx/pow_data.rs
```diff
@@ -59,10 +56,41 @@ pub struct MoneroPowData {
...
+}
+
+impl BorshDeserialize for MoneroPowData {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let header = monero::BlockHeader::consensus_decode(buf).unwrap();
+        let randomx_key = BorshDeserialize::deserialize(buf)?;
+        let transaction_count = BorshDeserialize::deserialize(buf)?;
+        let merkle_root = monero::Hash::consensus_decode(buf).unwrap();
+        let coinbase_merkle_proof = BorshDeserialize::deserialize(buf)?;
+        let coinbase_tx = monero::Transaction::consensus_decode(buf).unwrap();
```
sdbondi: unwraps

base_layer/core/src/proto/transaction.rs
```diff
@@ -248,7 +254,8 @@ impl TryFrom<proto::types::TransactionOutput> for TransactionOutput {
             .try_into()
             .map_err(|_| "Metadata signature could not be converted".to_string())?;
 
-        let covenant = Covenant::from_bytes(&output.covenant).map_err(|err| err.to_string())?;
+        let mut buffer = output.covenant.as_bytes();
+        let covenant = BorshDeserialize::deserialize(&mut buffer).unwrap();
```
sdbondi: unwrap

base_layer/core/src/proto/transaction.rs
```diff
@@ -202,10 +211,7 @@ impl TryFrom<TransactionInput> for proto::types::TransactionInput {
                     .to_vec(),
                 // Output hash is only used in compact form
                 output_hash: Vec::new(),
-                covenant: input
-                    .covenant()
-                    .map_err(|_| "Non-compact Transaction input should contain covenant".to_string())?
-                    .to_bytes(),
+                covenant,
```
sdbondi, line 214:
```suggestion
                input
￼                    .covenant()
￼                    .map_err(|_| "Non-compact Transaction input should contain covenant".to_string())?
￼                    .serialize()?,
```

infrastructure/tari_script/src/script.rs
```diff
@@ -59,6 +60,22 @@ pub struct TariScript {
...
+        Ok(())
+    }
+}
+
+impl BorshDeserialize for TariScript {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let script =
+            TariScript::from_bytes(buf).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e.to_string()))?;
+        *buf = &buf[script.to_bytes().len()..];
+        Ok(script)
```
sdbondi: Converting to bytes just to get the length isnt great. From bytes should probably just take in a `&mut &[u8]`
But can leave it for now since it's just an optimisation

base_layer/core/src/covenants/covenant.rs
```diff
@@ -46,19 +49,32 @@ pub struct Covenant {
...
 
+impl BorshSerialize for Covenant {
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        self.write_to(writer)
+    }
+}
+
+impl<'a> BorshDeserialize for Covenant {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let covenant = Self::from_bytes(buf).map_err(|e| io::Error::new(io::ErrorKind::InvalidInput, e.to_string()))?;
```
sdbondi: The covenant is variable length, so the length needs to be encoded 
Does this not work?
```rust
 fn deserialize(reader: &mut &[u8]) -> io::Result<Self> {
       let len = reader.read_varint::<usize>()?;
        if len == 0 {
            return Ok(Covenant::new());
        };
        // Check the length varint - this may be maliciously misreported
        if len > MAX_COVENANT_BYTES {
            return Err(io::Error::new(
                io::ErrorKind::InvalidInput,
                "covenants: Length varint exceeded maximum",
            ));
        }
        // Ensure that no more than the maximum bytes can be read
        let mut limited = LimitedBytesReader::new(MAX_COVENANT_BYTES, reader);
        CovenantTokenDecoder::new(&mut limited)
            .collect::<Result<_, CovenantDecodeError>>()
}
```

sdbondi (commented):
Looks good, needs a fair amount of testing.

base_layer/core/src/proto/transaction.rs
Cifko: Output of the serialize is not the buffer, it's just `()`

infrastructure/tari_script/src/script.rs
Cifko: This was reworked. I'm not using `to_bytes`. I do save the size of serialized buffer in serialize function and in deserialize I read the size and the pass the slice to opcodes::parse

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
Cifko: Reworked to use u8 as length and serialize the data one by one.

base_layer/core/src/covenants/covenant.rs
Cifko: Apparently we don't have a test for this. It was broken. I reworked it. Added tests for all custom (de)serializations.

base_layer/core/src/proof_of_work/monero_rx/fixed_array.rs
Cifko: Done.

Cifko:
Thanks guys. I even cought a bug when I did additional testing.

stringhandler (approved):
Still a bunch of unwraps that need fixing, but going to be easier to see in a new PR. I'll create an issue

base_layer/core/src/covenants/covenant.rs
sdbondi: How was it broken?

sdbondi: In any case, happy with the changes for this :+1:

Cifko: > How was it broken?

Because the borsh will send you all data, so if the covenants were like couple bytes. The CovenantTokenDecoder received more than it shoulld. This is also why I added a tests where I send more data to the deserialization and I expect the buffer still contain the extra data.

Cifko: I should probably also refactor the test for extra data. Because now I have the same code in all 5-6 tests.

Issue #4909, opened 2022-11-11
Clean up conflicting base node configs


sdbondi:
There are a few duplicate config keys that exist in the base node config and the base node state machine config
meaning settings and behaviour could mismatch.

`base_node.max_randomx_vms` (this setting will not work)
`base_node.bypass_range_proof_verification` (this setting will not work)
`base_node.force_sync_peers` (this was fixed in #4647 but the fix is a bit hacky) 

https://github.com/tari-project/tari/blob/956b27954dda1f15f82bff0ba0ba0ee1f0880d2d/base_layer/core/src/base_node/state_machine_service/state_machine.rs#L52

https://github.com/tari-project/tari/blob/development/applications/tari_base_node/src/config.rs#L109

ref #4646

Issue #4646, opened 2022-09-08
[base node] Force sync peers doesnt work (duplicate config)


sdbondi:

stringhandler:
To confirm I understand:
 "If you add a force sync peer to the config, it is not enforced"

sdbondi:
This was hack fixed in the PR that references this (https://github.com/tari-project/tari/pull/4647/files#diff-6e9e3f32e84954314cbbdf459c53d1e61da84ab09d309e787fecb99b57970ee6R98) so we can close this one.

I'll look into it and make a proper issue because I think there were a few config fields that were duplicated in different structs that could lead to bugs and should be cleaned up

PR #4938, opened 2022-11-21
fix: config cleanup

agubarev:
Description
---
Removed `max_randomx_vms` from `BaseNodeStateMachineConfig` and the config preset, now passing the initialized RandomXFactory in via the initializer, same for `bypass_range_proof_verification` but passing only the setting

Motivation and Context
---
https://github.com/tari-project/tari/issues/4909

There are a few duplicate config keys that exist in the base node config and the base node state machine config
meaning settings and behaviour could mismatch.

`base_node.max_randomx_vms` (this setting will not work)
`base_node.bypass_range_proof_verification` (this setting will not work)
`base_node.force_sync_peers` (this was fixed in #4647 but the fix is a bit hacky) 

https://github.com/tari-project/tari/blob/956b27954dda1f15f82bff0ba0ba0ee1f0880d2d/base_layer/core/src/base_node/state_machine_service/state_machine.rs#L52

https://github.com/tari-project/tari/blob/development/applications/tari_base_node/src/config.rs#L109

ref #4646

How Has This Been Tested?
---
manually

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/state_machine.rs
+++ b/base_layer/core/src/base_node/state_machine_service/state_machine.rs
@@ -55,9 +55,6 @@ pub struct BaseNodeStateMachineConfig {
     /// The amount of blocks this node can be behind a peer before considered to be lagging (to test the block
     /// propagation by delaying lagging)
     pub blocks_behind_before_considered_lagging: u64,
-    /// Bypass range proof verification to speed up validation
-    // TODO: This is a potential conflict with 'BaseNodeConfig::bypass_range_proof_verification'
-    pub bypass_range_proof_verification: bool,
 }
 
 #[allow(clippy::derivable_impls)]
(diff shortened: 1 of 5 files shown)


base_layer/core/src/base_node/state_machine_service/state_machine.rs
```diff
@@ -52,8 +52,6 @@ const LOG_TARGET: &str = "c::bn::base_node";
 pub struct BaseNodeStateMachineConfig {
     pub blockchain_sync_config: BlockchainSyncConfig,
     /// The maximum amount of VMs that RandomX will be use
-    // TODO: This is a potential conflict with 'BaseNodeConfig::max_randomx_vms'
-    pub max_randomx_vms: usize,
```
sdbondi, line 56: Github wont let me comment lower but can we remove `bypass_range_proof_verification` below?

sdbondi (approved):

Issue #4734, opened 2022-09-23
Commitment signature does not prove knowledge of witness summands


AaronFeickert:
Commitment signatures are used in the protocol as part of transaction authorization. These use a Schnorr-type representation proof that binds a message into the Fiat-Shamir transcript hash to form a signature. However, use in the protocol treats the input commitment as a sum of two values; one is the Pedersen commitment for the input/output, and the other is either a sender offset public key or a script public key. In either case, both values are bound to the signature via the transcript hash, but only their sum is used in the proof statement. This means successful verification of the signature only asserts that the signer(s) knew the sum of each representation component. Any further conclusions about summand representation must be offloaded to other parts of the protocol.

Unless there is a compelling reason to structure the signature this way (in which case it should be carefully documented), it should be modified.

AaronFeickert:
When a sender [completes a commitment signature](https://github.com/tari-project/tari/blob/development/base_layer/core/src/transactions/transaction_protocol/sender.rs#L488-L519), it generates its own Schnorr signature, extracts a component from it, and adds this into the commitment signature. This means there is an assumption about the structure of commitment signatures and their "compatibility" with Schnorr signatures. This doesn't seem like a robust abstraction to make, and it would affect any changes to commitment signatures.

SWvheerden:
This is for the [metadata signature](https://rfc.tari.com/Glossary.html#metadata-signature), just clarifying this. We use the commitment signature in the input as well for the script signature. 

The whole reason for the existence of this signature is that the receiver should not be able to change the script on the utxo the sender sent to the receiver. It was originally designed to be only a schonrr signature signed with the `sender_offset_private_key`

Linking this issue here: #2997

AaronFeickert:
> This is for the [metadata signature](https://rfc.tari.com/Glossary.html#metadata-signature), just clarifying this. We use the commitment signature in the input as well for the script signature.
> 
> The whole reason for the existence of this signature is that the receiver should not be able to change the script on the utxo the sender sent to the receiver. It was originally designed to be only a schonrr signature signed with the `sender_offset_private_key`
> 
> Linking this issue here: #2997

It applies to the script signature as well, if the method in [RFC-0201](https://rfc.tari.com/RFC-0201_TariScript.html#transaction-input-changes) for script signatures also uses a sum of witness values.

AaronFeickert:
[...]

PR #4943, opened 2022-11-23
feat!: update commitment signature 

SWvheerden:
Description
---
Replaces PR https://github.com/tari-project/tari/pull/4733
This updates  CommitmentSignature to CommitmentAndPublicKeySignature for transaction authorization.
Includes a new Gen block

Motivation and Context
---
See issue: https://github.com/tari-project/tari/issues/4734

How Has This Been Tested?
---
Unit tests

BREAKING CHANGE: Commitment signature proof has changed


stringhandler (approved):

applications/tari_console_wallet/src/automation/commands.rs
```diff
@@ -1032,13 +1032,13 @@ fn write_utxos_to_csv_file(utxos: Vec<UnblindedOutput>, file_path: PathBuf) -> R
     let mut csv_file = LineWriter::new(file);
     writeln!(
         csv_file,
-        r##""index","value","spending_key","commitment","flags","maturity","script","input_data","script_private_key","sender_offset_public_key","public_nonce","signature_u","signature_v""##
+        r##""index","value","spending_key","commitment","flags","maturity","script","input_data","script_private_key","sender_offset_public_key","emperical_commitment","emperical_nonce","signature_u_x","signature_u_a","signature_u_y""##
```
AaronFeickert, line 1035: Misspelling... should be `ephemeral_commitment` and `ephemeral_nonce`.

base_layer/core/src/transactions/transaction_components/transaction_input.rs
```diff
@@ -170,7 +172,8 @@ impl TransactionInput {
         match version {
             TransactionInputVersion::V0 | TransactionInputVersion::V1 => {
                 DomainSeparatedConsensusHasher::<TransactionHashDomain>::new("script_challenge")
```
AaronFeickert, line 174: May want to use the reverse-domain standard for this label, in order to be more consistent with other uses of the hashing API elsewhere in the codebase.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -269,39 +272,22 @@ impl TransactionOutput {
...
         sender_offset_public_key: &PublicKey,
-        public_commitment_nonce: &Commitment,
+        ephemeral_commitment: &Commitment,
+        ephemeral_pubkey: &PublicKey,
         commitment: &Commitment,
         covenant: &Covenant,
         encrypted_value: &EncryptedValue,
         minimum_value_promise: MicroTari,
     ) -> [u8; 32] {
         let common = DomainSeparatedConsensusHasher::<TransactionHashDomain>::new("metadata_signature")
```
AaronFeickert, line 288: May want to use the reverse-domain standard for this label, in order to be more consistent with other uses of the hashing API elsewhere in the codebase.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -490,23 +491,42 @@ impl SenderTransactionProtocol {
...
+            output.minimum_value_promise,
+        )?;
         // Create aggregated metadata signature
-        let (r_pub, u, v) = output.metadata_signature.complete_signature_tuple();
-        let r_pub_aggregated = r_pub + &public_commitment_nonce;
-        let u_aggregated = u + sender_signature;
-        let aggregated_metadata_signature = ComSignature::new(r_pub_aggregated, u_aggregated, v.clone());
-
-        let sender_offset_public_key = PublicKey::from_secret_key(sender_offset_private_key);
+        let aggregated_metadata_signature = &sender_signature + &output.metadata_signature;
```
AaronFeickert, line 510: It's possible that the recipient maliciously provided its own values for `ephemeral_pubkey` or `u_y` before returning it to the sender. Provided the output's sender offset public key cannot be overwritten by the recipient in a way that forces the sender to include it in the challenge, this will be caught by the verification check below, but could also be short-circuited by asserting that the recipient's partial metadata signature has `ephemeral_pubkey == u_y == 0`.

AaronFeickert: It may also be a good idea to include this check anyway for defense in depth.

AaronFeickert:
An even safer approach to partial metadata signatures would be not to add them at all (which is where you could get recipient shenanigans), but instead to build a fresh metadata signature from the appropriate constituent parts provided by the sender and recipient. This would eliminate the possibility of unexpected values included in partial signatures.

base_layer/core/src/transactions/transaction_protocol/sender.rs
SWvheerden: When doing m-of-n this will need to be bypassed. I prefer a one fit for all check. 

But yeah if the recipient changes `ephemeral_pubkey` or `u_y` the check below will fail and the signature will not pass.

The recipient can at all times replace  `sender offset public key` and create a new signature at will, but this will break the `script_offset` of the transaction.

SWvheerden:
One thing to keep in mind with these signatures, is that the main thing that stop any recipient shenanigans is `script_offset`
Nothing stops the recipient or anybody who knows the blinding factor of the commitment being created to put a new `sender_offset_public_key` on the utxo and create a new fake signature that verifies. This is not stoppable on the utxo. 
The only thing that stops this is the `script_offset`

PR #4949, opened 2022-11-23
fix: node gets banned on reorg

### LINKED ISSUE #4799: Node gets banned during reorgs if mining
WARN  Could not provide requested block d54cd80ab95951fbc8cc605935ee11f9d8f9bdb2cd24ca319643a2bc16198241 to peer because not stored

This results in my node being banned (1 less connection each time) because we sent a block hash, but could not provide it later due to reorg (race condition). Full block requests happen a lot during reorgs because we check `if header.prev_hash != *current_meta.best_block() {` and then request the full block immediately if true.

The reason for the ban is to prevent spamming of block hashes in the previous scheme.

I think we should:
1. Validate the header in reconcile block
2. if valid, we know PoW was done so continue - if we request the block but the peer no longer has it, dont ban and discard the message
3. if invalid, ban the peer that sent it
4. Problem: A peer could send a non-tip header it mined that is not part of the chain. I think that we'd have to deal with this heuristically.

brianp:
Description
---
Remove the `FetchBlocksByHash` handler. It was only called from a single place and although designed to handle multiple blocks it was only ever sending a single hash at once making the multi-block functionality useless. 
Instead, opt to use the existing `GetBlockByHash` handler and expand that handler to accept a new `orphans` flag. Passing this flag means we'll accept found blocks from the orphan pool,

Motivation and Context
---
Previously if a node had re-orged after a sync had started it may result in not providing the complete block for a block it claimed it had. This results in a brief ban.

Make it also return blocks from the orphan pool and let the peer figure out what to do with it. 

How Has This Been Tested?
---
Tests, and running nodes.

Fixes: #4799


stringhandler (commented):
looking good so far

brianp:
Got the failing test all worked out but now it's segfault'ing during syncs. I've already got an idea why that might be happening. New fix is incoming.

brianp:
segfault fix pushed. I had gotten a little too overzealous with the code removal. Removing the grpc response formats for HistoricalBlocks results in segfault when syncing. I've re-added the hastily removed code and the segfault no longer occurs.

Good for review now.

stringhandler (approved):

PR #4953, opened 2022-11-24
fix: minimize potential memory leaks of sensitive data on the wallet code

jorgeantonio21:
Description
---
Sensitive data is transmitted from wallet and db (ideally encrypted). However, the code is not completely robust against memory leaks. This PR proposes a way to minimize this risk. This implies enforcing encryption immediately whenever we process data for Sqlite interaction (or when fetching it). We further remove any update of sensitive data on SQL outputs, as this would increase the potential risk of memory leaks across the code. 

Motivation and Context
---
Address security vulnerabilities in the wallet code. This PR is related to issue #4846.

How Has This Been Tested?
---
Refactor some of the existing unit tests.


base_layer/wallet/src/util/encryption.rs
```diff
@@ -77,7 +78,7 @@ pub fn decrypt_bytes_integral_nonce(
 pub fn encrypt_bytes_integral_nonce(
     cipher: &XChaCha20Poly1305,
     domain: Vec<u8>,
-    plaintext: Vec<u8>,
+    plaintext: Hidden<Vec<u8>>,
```
jorgeantonio21, line 80: Another possibility is to pass a plain `Vec<u8>` as input and then zeroize `plaintext` at the end of the function.

stringhandler (approved):

Issue #4866, opened 2022-10-28
Failed block sync might leave some sync data in the main chain database


SWvheerden:
Currently, syncing is handled by the BN state machine. If the node things its behind, the node will try header_sync to download headers. 
If it has enough VALID headers, it will commit them as headers on the main chain, and continue do to so for as long it is provided VALID headers. 
If this fails, the node goes to waiting state. 
It should not, it should either try and sync the VALID blocks from a node, or delete the headers and on. 
But most cases this will return `Ok`

The block sync has the same problem, after failing, we need to check and see if we need to reset back to our previous state.

SWvheerden:
Rewind will store all the re-orged blocks in the orphan database. 
We should just chose the highest VALID tip and make sure the main chain is that one

sdbondi:
Initial reaction is to disagree. The chain metadata keeps track of the valid full block height. We already handle having a chain of valid headers but no/partial blocks. If we reset on an invalid header, we'll have to download valid headers all over again. Same for blocks. If a synced header/block is invalid after syncing others, the blocks that led up to them are still valid so why should we throw them away?

~~I dont think the rewinding puts the removed blocks back in the orphan database, only updates the orphan tip hash after a reorg, which doesnt seem right.~~ Ah I was mistaken, this happens inside `rewind_to_height`

https://github.com/tari-project/tari/blob/09eda1b815c86ece8fb9b919a7c5cf5ab1cbbaf1/base_layer/core/src/chain_storage/blockchain_database.rs#L2004

SWvheerden:
> I dont think the rewinding puts the removed blocks back in the orphan database, only updates the orphan tip hash after a reorg, which doesnt seem right.

It happens in rewind, not reorganise
https://github.com/tari-project/tari/blob/09eda1b815c86ece8fb9b919a7c5cf5ab1cbbaf1/base_layer/core/src/chain_storage/blockchain_database.rs#L1725

SWvheerden:
> Initial reaction is to disagree. The chain metadata keeps track of the valid full block height. We already handle having a chain of valid headers but no/partial blocks. If we reset on an invalid header, we'll have to download valid headers all over again. Same for blocks. If a synced header/block is invalid after syncing others, the blocks that led up to them are still valid so why should we throw them away?

The header sync does this correctly, it only "swaps" if the syncing chain has higher PoW and is Valid. It will even do that for a partial. Which is correct. 

[...]

PR #4955, opened 2022-11-24
feat: reset broken sync

SWvheerden:
Description
---
If sync fails resets chain to the highest pow chain the node locally has the data to. 

Motivation and Context
---

See: https://github.com/tari-project/tari/issues/4866

How Has This Been Tested?
---
Unit tests


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2042,42 +2067,61 @@ fn reorganize_chain<T: BlockchainBackend>(
...
+    Ok(removed_blocks)
+}
+
+fn swap_to_highest_pow_chain<T: BlockchainBackend>(
+    db: &mut T,
+    block_validator: &dyn PostOrphanBodyValidation<T>,
+    chain_strength_comparer: &dyn ChainStrengthComparer,
+) -> Result<(), ChainStorageError> {
+    let metadata = db.fetch_chain_metadata()?;
+    // lets clear out all remaining headers that done have a matching block
```
```suggestion
    // lets clear out all remaining headers that don't have a matching block
```

stringhandler (changes requested):
I think the logic in add_block should then call this method as well, so that we don't have duplicate code

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1838,75 +1865,104 @@ fn handle_possible_reorg<T: BlockchainBackend>(
...
+    block_validator: &dyn PostOrphanBodyValidation<T>,
+    chain_strength_comparer: &dyn ChainStrengthComparer,
+) -> Result<BlockAddResult, ChainStorageError> {
+    let metadata = db.fetch_chain_metadata()?;
+    // lets clear out all remaining headers that dont have a matching block
+    // rewind to height will first delete the headers, then try delete from blocks, if we call this to the current
+    // height it will only trim the extra headers with no blocks
+    rewind_to_height(db, metadata.height_of_longest_chain())?;
+    let all_orphan_tips = db.fetch_all_orphan_chain_tips()?;
+    if all_orphan_tips.is_empty() {
```
stringhandler, line 1943: Maybe this check should be before we rewind

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1838,75 +1865,104 @@ fn handle_possible_reorg<T: BlockchainBackend>(
...
+        // we have no orphan chain tips, we have trimmed remaining headers, we are on the best tip we have, so lets
+        // return ok
+        return Ok(BlockAddResult::OrphanBlock);
```
stringhandler, line 1946: We should probably rename this to `ChainReorgResult`

stringhandler: And `OrphanBlock` to change to `NoChange`

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: I specifically did not do this as I use the rewind code here to remove the "extra" headers. 
I want to "main chain" to be "clean" after this run. 

The intended case is this. 

Header sync downloads 10 headers. 
Block sync fails after 5. 
That extra 5 headers will stay on top of the main chain till we remove them again. I want to remove them at this point regardless of whether the new tip is higher than our original one.

Issue #4908, opened 2022-11-11
Remove Metadata from Output features [for non coinbase outputs]


stringhandler:
Metadata is a remnant of the previous DAN implementation. It should be removed from output features

stringhandler:
it might be useful to allow this field only for coinbases, and store information here to display the mining pool that mined this block. Showing the mining pool will make it useful to see if a mining pool has more than 30 or 51%

stringhandler:
I think the rule should be: This field should be optional for coinbases (mining pools and other merge mined coins can use it), but it  should be empty for non-coinbases.

For coinbases, the maximum length should be 64 bytes (2x hashes), so that arbitrary data cannot be included

agubarev:
Although merged, I should make another PR to reflect Stan's comment, PR got merged before I managed to push an update. Will address to this ticket.

agubarev:
On second thought, satisfying the aforementioned performance nitpick (on the PR), at the moment, would create unnecessary indirection in the code. Perhaps this can be addressed more properly later, via larger optimization.

PR #4957, opened 2022-11-25
feat: only coinbase output features may have metadata set, and is of limited size; ref #4908

agubarev:
Description
---
Added validation to OutputFeatures which, for now, makes sure that only coinbase output features may have metadata set.

Motivation and Context
---
Remove Metadata from Output features: metadata is a remnant of the previous DAN implementation. It should be removed from output features

https://github.com/tari-project/tari/issues/4908

How Has This Been Tested?
---
unit tests


agubarev:
my solution is completely wrong

PR #4960, opened 2022-11-25
feat: only coinbase output features may have metadata set, and is of limited size; ref #4908

agubarev:
Description
---
Added validation to OutputFeatures which, for now, makes sure that only coinbase output features may have metadata set.

Motivation and Context
---
Remove Metadata from Output features: metadata is a remnant of the previous DAN implementation. It should be removed from output features

https://github.com/tari-project/tari/issues/4908

How Has This Been Tested?
---
unit tests

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -27,7 +27,6 @@ use std::{
 
 use chrono::{DateTime, Duration, Utc};
 use tari_common::configuration::Network;
-use tari_common_types::types::FixedHash;
 use tari_script::{script, OpcodeVersion};
 use tari_utilities::epoch_time::EpochTime;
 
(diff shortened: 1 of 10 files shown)


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -209,6 +210,10 @@ impl ConsensusConstants {
         self.transaction_weight.calculate(1, 0, 1, metadata_size)
     }
 
+    pub fn coinbase_output_features_metadata_max_length(&self) -> usize {
+        FixedHash::byte_size() * 2
```
stringhandler: this should be a field, not hardcoded.

stringhandler (approved):

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -325,8 +325,29 @@ impl AggregateBody {
         Ok(())
     }
 
-    /// This function will check all stxo to ensure that feature flags where followed
-    pub fn check_stxo_rules(&self, height: u64) -> Result<(), TransactionError> {
+    pub fn check_output_features(&self, max_coinbase_metadata_size: usize) -> Result<(), TransactionError> {
+        for output in self.outputs() {
```
sdbondi, line 329: nit(perf): Rather than looping each time in the validation helper, reuse the loop and call this with the output.

image

sdbondi (approved):

PR #4975, opened 2022-11-29
fix(wallet): invalid metadata sig when creating code template utxo

sdbondi:
Description
---
- fixes invalid metadata signature error when submitting transaction with custom outputs e.g. CodeTemplateRegistration

Motivation and Context
---
`create_send_to_self_with_output`  generates a transaction with an invalid metadata signature. This function is used by `create_template_registration` in wallet GRPC. 

In order to send a send-to-self transaction, the UnblindedOutputBuilder needs to create a complete signature for the new output. 

The UnblindedOutputBuilder is only currently used in create_pay_to_self_containing_outputs calls.

How Has This Been Tested?
---
Manually, able to submit a valid template reg utxo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4898,6 +4898,7 @@ dependencies = [
  "tari_key_manager",
  "tari_libtor",
  "tari_p2p",
+ "tari_script",
  "tari_shutdown",
  "tari_utilities",
  "tari_wallet",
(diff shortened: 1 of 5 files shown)


SWvheerden (approved):
utACK

stringhandler (approved):
Nice, we actually need to create some tests for this in this repo

PR #4976, opened 2022-11-29
fix(wallet): fix wallet_setting keys

sdbondi:
Description
---
- Fixes various typos in wallet settings keys
- Removes usage if Display impl for key to string conversion
- Add key name db migration

Motivation and Context
---
A number of key names had typos, this is likely because it is not clear that the string used comes from the Display impl.
This PR uses the DbKey type in the function signatures to ensure that keys are well formed and replaces the Display impl with `to_key_string` to make it clear that these keys are used as keys, and do not need to be human readable

How Has This Been Tested?
---
Manually, wallet updates keys and works


SWvheerden (approved):
utACK

stringhandler (approved):
Way better

Issue #4977, opened 2022-11-29
Wallet add export/import functionality of utxos for backups


SWvheerden:
Currently, the wallet backup process uses decrypts the whole wallet database and copies the SQL file future use. 
This is dangerous see: https://github.com/tari-project/tari/issues/4954

The wallet does support the export of utxos to CSV file. 
We need to ensure that this works, and works with the FFI so mobile wallets can use this backup only the utxos and leave sensitive data alone and encrypted.

jorgeantonio21:
Encryption is possible only when the user provides a password. So we should enforce this on the wallet API, as well.

jorgeantonio21:
Encryption is possible only when user provides a password. So we should enforce this on the wallet API, as well.

PR #4984, opened 2022-11-30
fix: remove optionals from wallet set up

jorgeantonio21:
Description
---
Currently, it is possible to set up a wallet (in creation/recovery/existing mode) without a password. This is not desirable, as in that case, sensitive data might be potentially stored in the database and leak in memory. We address this issue by enforcing passphrase creation.

Motivation and Context
---
For more details see issue #4977.

How Has This Been Tested?
---


base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -631,24 +592,71 @@ impl WalletBackend for WalletSqliteDatabase {
...
 
-    let cipher = match (passphrase, db_passphrase_hash, db_encryption_salt) {
-        (Some(_), None, _) => {
-            error!(
-                target: LOG_TARGET,
-                "Passphrase is provided but database is not encrypted"
-            );
-            return Err(WalletStorageError::InvalidPassphrase);
+    let cipher = match (db_passphrase_hash, db_encryption_salt) {
+        (None, None) => {
```
jorgeantonio21, line 479: We will clean up this code in a posterior PR, which goal is to change the `Backend` api. Doing so here, would require a lot of further refactoring.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -631,24 +591,71 @@ impl WalletBackend for WalletSqliteDatabase {
 /// Master Public Key that is stored in the db
 fn check_db_encryption_status(
```
SWvheerden: Why cant we remove/change this function?
We are enforcing encryption, thus this should always return the XChaCha20Poly130
Its going to make this code much smaller and simplier

SWvheerden (commented):
This looks good. 
But I would say we need to remove all the possible encryption stuff in this PR.
This PR is already big, but most of the changes are tests. 

We still have the ``apply_encryption` and `remove_encryption` functions. I would remove them. 
We are enforcing encryption, but in the same breath, we allow a public outward function to decrypt the db. 

When this PR gets merged, we should not have the backend be in two possible states, it should only be encrypted.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
jorgeantonio21: I agree, I think we can safely remove/refactor this function. Moreover, removing `apply_encryption` and `remove_encryption` might a good idea to do know.

jorgeantonio21: Done

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -67,19 +67,19 @@ impl KeyManagerSqliteDatabase {
 
     fn decrypt_if_necessary<T: Encryptable<XChaCha20Poly1305>>(&self, o: &mut T) -> Result<(), KeyManagerStorageError> {
         let cipher = acquire_read_lock!(self.cipher);
-        if let Some(cipher) = cipher.as_ref() {
-            o.decrypt(cipher)
-                .map_err(|_| KeyManagerStorageError::AeadError("Decryption Error".to_string()))?;
-        }
+
```
stringhandler, line 70: These methods should probably change, but not in this PR

stringhandler (approved):

PR #4992, opened 2022-12-02
fix: use find by id rather

stringhandler:
This code strangely finds by spending key, which is not unique....


SWvheerden (dismissed):
utACK
pending green CI

PR #4967, opened 2022-11-28
fix: hide sensitive data on tari repo (see issue #4846)

jorgeantonio21:
Description
---
The goal of this PR is to zeroize sensitive data content across the Tari repo.  We pay special attention to kdf key data and other occurrences of sensitive data. For more details, we refer to issue #4846. Moreover, this PR is a companion to #4953 and #4925

Motivation and Context
---
Tackle issue #4846.

How Has This Been Tested?
---

PR #4994, opened 2022-12-02
fix: improve key handling

AaronFeickert:
Description
---
Updates the use of `CoreTransactionAEADKey` and `CommsNoiseKey` to be populated in place. Switches `CommsNoiseKey` to use a `SafeArray` under the hood.

Extends the work of [PR 4967](https://github.com/tari-project/tari/pull/4967).

Motivation and Context
---
Earlier work updates the use of keys for value encryption and Noise. Value encryption keys use a `SafeArray` type in a `Hidden` wrapper, and Noise keys use a `Hidden` array. However, in both cases, a copy of the hash output used to populate the keys is left in memory.

This work mitigates the problem. Because the hashing API now supports in-place output finalization via `finalize_into` and `finalize_into_reset`, we can populate keys directly by mutable reference.

It also renames `CoreTransactionAEADKey` to `EncryptedValueKey` for clarity, and switches `CommsNoiseKey` to be a `SafeArray` under the hood. There is [discussion](https://github.com/tari-project/tari/pull/4967#discussion_r1033628944) on whether `CommsNoiseKey` needs to be a `SafeArray`; while the reasoning makes sense, I still think it's good practice to take advantage of the benefits of `SafeArray` for array-like key types unless there's a compelling performance reason otherwise. Discussion on this is welcome!

How Has This Been Tested?
---
Existing tests pass.

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_value.rs
@@ -30,7 +30,7 @@ use chacha20poly1305::{
     KeyInit,
     Nonce,
 };
-use digest::generic_array::GenericArray;
+use digest::{generic_array::GenericArray, FixedOutput};
 use serde::{Deserialize, Serialize};
 use tari_common_types::types::{Commitment, PrivateKey};
 use tari_crypto::{hash::blake2::Blake256, hashing::DomainSeparatedHasher};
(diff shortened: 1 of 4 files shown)


AaronFeickert:
Before merging, it should be decided whether or not the domain separation labels for the relevant hashers should be updated to reflect the reverse-domain notation used elsewhere in the codebase. I think they probably should, but this will be a breaking change.

stringhandler (approved):

Issue #4989, opened 2022-12-02
feat! - Add room for arbitrary miner-supplied info in coinbase


CjS77:
A small field for miners to add arbitrary data has a few main use cases:
* Helps miners / pools track blocks they have won.
* Permits Tari to be used as a parent chain in merge mining.

The first case is largely cosmetic. The 2nd is pretty important.

It should be a fairly minor (sic!) change. Allowing up to 64 bytes of arbitrary data should be sufficient.

stringhandler:
Not consensus breaking because we have the `metadata` field specifically to store this. The miner needs to be updated to store this

SWvheerden:
This has been done, field is called: coinbase_extra

stringhandler:
Reopening as the merge mining use case is not yet handled

SWvheerden:
Addressed documentation on this: https://github.com/tari-project/rfcs/pull/91

PR #4995, opened 2022-12-02
fix: coinbase extra info is not checked

CjS77:
This was initially a PR to change the field name of `metadata` in output features to `coinbase_extra`.

This is to reduce confusion, since at least one other field is labelled metadata and looking for this in the code is very confusing with this ambiguity.

However, in making the change, two tests were added:
* Check that *transactions* with coinbase_extra set are rejected as invalid. This turned out not to be the case, because output features are _only checked on orphan block validations_. This is a bug, because invalid transactions (from a stateless perspective) must be rejected as transactions, and not after they have been placed in a block.

* Check that the size of the coinbase_extra field is enforced. This does not appear to be done correctly, since adding a test in the general blockchain end-to-end test did not throw a block with an oversized extra field out.

This then uncovered a very odd definition in the tx weight calculation, which has been marked for clarification.

At this point, this PR is more of an issue, with breaking tests and comments. I'm looking for some clarification on those before making further changes.


CjS77:
Ultimately, should close #4989

base_layer/core/src/proto/transaction.proto
```diff
@@ -97,8 +97,10 @@ message OutputFeatures {
     // The maturity of the specific UTXO. This is the min lock height at which an UTXO can be spend. Coinbase UTXO
     // require a min maturity of the Coinbase_lock_height, this should be checked on receiving new blocks.
     uint64 maturity = 3;
-    bytes metadata = 4;
+    // Additional arbitrary info in coinbase transactions supplied by miners
+    bytes coinbase_extra = 4;
```
CjS77, line 101: renaming fields are cool in protobuf btw

base_layer/core/src/transactions/weight.rs
```diff
@@ -33,7 +33,7 @@ pub struct WeightParams {
     /// Weight in grams per output, excl. TariScript and OutputFeatures
     pub output_weight: u64,
     /// Metadata per byte weight
-    pub metadata_bytes_per_gram: Option<NonZeroU64>,
+    pub metadata_bytes_per_gram: Option<NonZeroU64>, // Why is this defined like this? Seems very hacky
```
CjS77, line 36: Why?

base_layer/core/src/validation/test.rs
```diff
@@ -422,16 +422,33 @@ fn chain_balance_validation_burned() {
...
+        let factories = CryptoFactories::default();
+        let validator = TxInternalConsistencyValidator::new(factories, true, db);
+        let mut features = OutputFeatures { ..Default::default() };
+        features.coinbase_extra = b"deadbeef".to_vec();
+        let (tx, _, _) = tx!(MicroTari(100_000), fee: MicroTari(5), inputs: 1, outputs: 1, features: features);
+        let err = validator.validate(&tx).unwrap_err();
+        assert!(matches!(
+            err,
+            ValidationError::TransactionError(TransactionError::NonCoinbaseHasOutputFeaturesCoinbaseExtra)
+        ));
```
CjS77, line 452: This test was failing before adding the change to base_layer/core/src/validation/transaction_validators.rs#76

base_layer/core/src/validation/transaction_validators.rs
```diff
@@ -71,6 +71,10 @@ impl<B: BlockchainBackend> MempoolTransactionValidation for TxInternalConsistenc
             return Err(ValidationError::ErroneousCoinbaseOutput);
         }
 
+        // We can call this function with a constant value, because we've just shown that this is NOT a coinbase, and
+        // only coinbases may have the extra field set (the only field that the fn argument affects).
+        tx.body.check_output_features(1)?;
```
CjS77, line 76: Not the cleanest check, tbh, But getting the actual max size into this method would mean adding `rules` as an arg to a whole bunch of methods

base_layer/core/tests/block_validation.rs
```diff
@@ -682,32 +691,63 @@ async fn test_block_sync_body_validator() {
...
+    let err = validator.validate_body(new_block).await.unwrap_err(); // Whoops! This should not pass validation
+    assert!(
+        matches!(
+            err,
+            ValidationError::TransactionError(TransactionError::InvalidOutputFeaturesCoinbaseExtraSize{len, max }) if
+            len == 65 && max == max_len
+        ),
+        "{}",
+        err
+    );
```
CjS77: So this currently passes as a valid block. Unless I'm mistaken, this is a full blockchain test, and so all validation rules should be checked.

base_layer/core/src/transactions/weight.rs
sdbondi: I believe the option was necessary in the past because dibbler did not have this metadata weighting, since this value is the denominator if it is defined it cannot be 0, so setting a u64 to 0 is not the same as disabling it. The option can be removed now.

stringhandler: This is moot anyway, since the coinbase does not charge fees and is effectively free. Since it has a max, there is no need weight it specially, and in fact we want to encourage miners to use this field

stringhandler: Added #4997

stringhandler (dismissed):

base_layer/core/src/transactions/weight.rs
sdbondi: Yup agree, no need to use up the total block weight for coinbase extra field. metadata weighting refers to all UTXO metadata (e.g. sidechain features etc) not just that single field.

PR #4999, opened 2022-12-05
feat: add utxo import and export for ffi

### LINKED ISSUE #4954: Confirm that RFC-0153 - Stage 1b is implemented correctly
In the RFC, stage 1b says:

> The seed phrase MUST NOT be stored on the cloud in Stage 1b. Doing so would result in all future funds of the user being lost if the backup were ever compromised. Since the backup is unencrypted in Stage 1b, we store the minimum amount of data needed to recover the funds and limit the potential loss of funds in case of a breach to just that found in the commitments in the backup, which should not be more than $50.

We need to confirm that the wallet code does this correctly

SWvheerden:
Description
---
Add the ability to export and import utxos to the faucet

Motivation and Context
---
We need to be able to backup the utxo's without exposing all private information from the wallet. 
This only exposes the limited selection of private keys that are exported

How Has This Been Tested?
---
Unit tests. 

Fixes: https://github.com/tari-project/tari/issues/4954

Open question:
Do we only export the unspent set, or do we export the spent and unspent?


stringhandler (dismissed):

stringhandler:
Just need to fix that failing test

stringhandler (approved):

PR #5023, opened 2022-12-08
chore: remove unused lock logging

SWvheerden:
Description
---
Remove unused logging of how long it takes to acquire a lock.

Motivation and Context
---
We use to have an RWlock on the wallet database. We added logging to measure how much of a performance issue the lock was. We removed the RWlock and introduced pooled connections. We have not removed the lock logging.

### DIFF
--- a/base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
+++ b/base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
@@ -27,8 +27,6 @@ use std::{
 
 use chacha20poly1305::XChaCha20Poly1305;
 pub use key_manager_state::{KeyManagerStateSql, NewKeyManagerStateSql};
-use log::*;
-use tokio::time::Instant;
 
 use crate::{
     key_manager_service::{
@@ -40,8 +38,6 @@ use crate::{
 };
 
 mod key_manager_state;
-
-const LOG_TARGET: &str = "wallet::key_manager_service::database::wallet";
 
 /// A Sqlite backend for the Output Manager Service. The Backend is accessed via a connection pool to the Sqlite file.
 #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -97,44 +91,22 @@ impl KeyManagerBackend for KeyManagerSqliteDatabase {
...
         };
-        if start.elapsed().as_millis() > 0 {
-            trace!(
-                target: LOG_TARGET,
-                "sqlite profile - fetch key_manager: lock {} + db_op {} = {} ms",
-                acquire_lock.as_millis(),
-                (start.elapsed() - acquire_lock).as_millis(),
-                start.elapsed().as_millis()
-            );
-        }
```
hansieodendaal, line 94:
```suggestion
        if start.elapsed().as_millis() > 0 {
            trace!(
                target: LOG_TARGET,
                "sqlite profile - fetch key_manager: db_op {} ms",
                start.elapsed().as_millis()
            );
        }
```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -86,9 +82,7 @@ impl KeyManagerSqliteDatabase {
 
 impl KeyManagerBackend for KeyManagerSqliteDatabase {
     fn get_key_manager(&self, branch: String) -> Result<Option<KeyManagerState>, KeyManagerStorageError> {
-        let start = Instant::now();
         let conn = self.database_connection.get_pooled_connection()?;
```
hansieodendaal, line 85:
```suggestion
        let start = Instant::now();
        let conn = self.database_connection.get_pooled_connection()?;
```

hansieodendaal (changes requested):
Hi there,

I agree that measuring the "db lock" time has no place anymore, by virtue of `let conn = self.database_connection.get_pooled_connection()?;` being nearly instantaneous in all stress tests I have ever done as _I have never saw `get_pooled_connection` >= 1ms_, however, measuring the db operation time is still very relevant. Those trace logs should not be removed.

I propose that we not measure the time it takes to get a pooled connection individually anymore but group it with the actual db operation time, as shown in the code snippet example below.

Please apply this to all instances where the trace logs have been removed.

As an alternative, to be more accurate in what is actually logged, change `sqlite profile - xxxxxxxxxx: lock {} + db_op {} = {} ms` to `sqlite profile - xxxxxxxxxx: connection {} + db_op {} = {} ms` and rename the variable `acquire_lock` to `acquire_connection` everywhere, but please do not remove the trace level logs.

PR #5035, opened 2022-12-09
chore: remove metadata references

### LINKED ISSUE #5031: Lots of references to metadata
There are a lot of references to metadata left over since we renamed the field to `coinbase_extra`in #4995

SWvheerden:
Description
---
Removes remaining namings of metadata
Removes coinbase_extra from covenants

Motivation and Context
---
We don't need a covenant for the coinbase_extra field. This field can only be used as per consensus on the coinbase field. 
As soon as any other value except empty is on the coinbase extra, consensus will through that tx out. 

How Has This Been Tested?
---
unit tests and manual

Fixes: https://github.com/tari-project/tari/issues/5031


applications/tari_app_grpc/proto/types.proto
```diff
@@ -67,7 +67,7 @@ message WeightParams {
     uint64 kernel_weight = 1;
     uint64 input_weight = 2;
     uint64 output_weight = 3;
-    uint64 metadata_bytes_per_gram = 4;
+    uint64 features_and_scripts_bytes_per_gram = 4;
```
CjS77, line 70: This can still be called metadata imo.

CjS77: But this is good too

CjS77 (commented):
LGTM

I don't think `wallet.zip` is supposed to be there though

SWvheerden:
> LGTM
> 
> I don't think `wallet.zip` is supposed to be there though

haha yeah good catch, removed

sdbondi (approved):

Issue #5030, opened 2022-12-09
[base node] Header sync gets stuck on banned peer


hansieodendaal:
This base node gets stuck trying to sync headers from banned peer `cdee5a37e4478755e532a32d09 `:

The ban incident:
``` rust
2022-12-09 04:41:26.368536000 [c::val::helpers] WARN  Validation failed on block:4d0a304eed563a82389dda2b27318d138fda3c131a69b218859a726d64d9d6de:InvalidSignatureError("Metadata signature not valid!")
2022-12-09 04:41:26.368560500 [c::cs::database] WARN  Block #5287 (4d0a304eed563a82389dda2b27318d138fda3c131a69b218859a726d64d9d6de) failed validation - The transaction is invalid: Signature is invalid: Metadata signature not valid!
2022-12-09 04:41:26.368624400 [c::bn::comms_interface::inbound_handler] WARN  Peer cdee5a37e4478755e532a32d09 sent an invalid block: Validation error: The transaction is invalid: Signature is invalid: Metadata signature not valid!
```
Consecutive header sync retries to `cdee5a37e4478755e532a32d09 `
``` rust
   36537  2022-12-09 04:41:56.424062900 [c::bn::header_sync] DEBUG Starting header sync.
   36538  2022-12-09 04:41:56.424070100 [c::bn::header_sync] INFO  Synchronizing headers (1 candidate peers selected)
   36539: 2022-12-09 04:41:56.424076700 [c::bn::header_sync] INFO  Attempting to sync headers (1 sync peers)
   36540  2022-12-09 04:41:56.424085400 [c::bn::header_sync] DEBUG Dialing cdee5a37e4478755e532a32d09 sync peer
   36541  2022-12-09 04:41:56.424157600 [c::bn::header_sync] DEBUG Header sync failed: Connectivity Error: ConnectionFailed: Peer is banned, denying connection

...
...
...

  116090  2022-12-09 09:22:11.320246000 [c::bn::header_sync] DEBUG Starting header sync.
  116091  2022-12-09 09:22:11.320252400 [c::bn::header_sync] INFO  Synchronizing headers (1 candidate peers selected)
  116092: 2022-12-09 09:22:11.320258800 [c::bn::header_sync] INFO  Attempting to sync headers (1 sync peers)
  116093  2022-12-09 09:22:11.320266300 [c::bn::header_sync] DEBUG Dialing cdee5a37e4478755e532a32d09 sync peer
[...]

PR #5039, opened 2022-12-12
fix(core): fixes stale chain metadata being sent to listening state

### LINKED ISSUE #5037: ChainMetadataService will not clear out stale metadata
The service relies on a vec called `peer_chain_metadata`.
This vec stores all the received metadata pings and pong replies from nodes. 
This vec is never cleared of stale information. A node will only be removed when it's disconnected or banned, if this does not happen, a node will keep and make logical decisions based on an old node's metadata forever. 

We need to try and implement this vec as some TTL cache so that old information gets removed automatically. 

I encountered the following behaviour on my node

- a Node sent my node chain metadata that was much higher than my current
- Before my node could sync to that node, that node banned my local node
- My node kept trying to sync to that node as it had the highest pow chain stored in the chain metadata
- Syncing kept failing with `Connectivity Error: ConnectionFailed: Identity protocol failed: IoError: Broken pipe (os error 32)`
- My node was stuck at a much lower height, as it could never ban that node or sync to that node

sdbondi:
Description
---
- Removes "caching" of peer metadata
- Send only one peer chain metadata at a time to the listening state
- increase chain metadata event channel to 20 (so that 20 of the last chain metadata received can be read by listening)

Motivation and Context
---
Fixes #5037 and #5030 (needs to be confirmed since this case is not easy to reproduce)

What I've observed is that the peer is banned twice. The first time, for 30 minutes (between 04:07:23.802123900 and 2022-12-09 04:41:26.365429300) then again permanently (bad metadata sig). It's clear in the logs that the pong is received before being banned, the node is banned (chain metadata service clears the peer's metadata) but the message is already in the domain pipeline so another one is received. The chain metadata service now keeps the peer's chain metadata and sends it to the listening state every time because it is not cleared.

TL;DR classic race condition.

Order of events:

- peer is banned for 30 minutes
- peer ban expires
- we receive a ping from the peer
- at almost the same time peer is banned (see logs in [this comment](https://github.com/tari-project/tari/issues/5030#issuecomment-1344335004))
- chain metadata service clears the peer from peer_chain_metadata
- the ping/pong is received (already in the pipeline from before the ban)
- the chain metadata is added to the vec, and is never cleared (because the peer is not banned again)
- the peer stays in the list despite being banned and NOT connected, so the header sync continues to try to connect to it

How Has This Been Tested?
---
Manually: rewind-blockchain and sync, enters sync mode timeously
Removed some tests that test removed functions
Added a new unit test for determine_sync_state

### DIFF
--- a/base_layer/core/src/base_node/chain_metadata_service/handle.rs
+++ b/base_layer/core/src/base_node/chain_metadata_service/handle.rs
@@ -80,8 +80,17 @@ impl Display for PeerChainMetadata {
 
 #[derive(Debug)]
 pub enum ChainMetadataEvent {
-    PeerChainMetadataReceived(Vec<PeerChainMetadata>),
+    PeerChainMetadataReceived(PeerChainMetadata),
     NetworkSilence,
+}
+
+impl ChainMetadataEvent {
+    pub fn peer_metadata(&self) -> Option<PeerChainMetadata> {
+        match self {
+            Self::PeerChainMetadataReceived(metadata) => Some(metadata.clone()),
+            _ => None,
+        }
+    }
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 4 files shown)


stringhandler (approved):

hansieodendaal:
Nice, I will test it later on with a stress test by unbanning all manually banned peers.

PR #5043, opened 2022-12-12
fix: functional wallet encryption (issue #5007)

jorgeantonio21:
Description
---
Wallet encryption and decryption is now done via pure methods, instead of mutating in place the required fields, see #5007.

Motivation and Context
---
Address issue #5007.

How Has This Been Tested?
---
Existing unit tests.


base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -115,8 +111,8 @@ impl KeyManagerBackend for KeyManagerSqliteDatabase {
         let conn = self.database_connection.get_pooled_connection()?;
         let acquire_lock = start.elapsed();
 
-        let mut km_sql = NewKeyManagerStateSql::from(key_manager);
-        self.encrypt_if_necessary(&mut km_sql)?;
+        let km_sql = NewKeyManagerStateSql::from(key_manager);
```
stringhandler, line 104: I think it's better here to have the SQL structs always encrypted, and to pass the cipher, or create the cipher in the from
```suggestion
        let km_sql = NewKeyManagerStateSql::from(key_manager, cipher);
```

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -183,17 +186,20 @@ impl Encryptable<XChaCha20Poly1305> for NewKeyManagerStateSql {
             .to_vec()
     }
 
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.primary_key_index = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut output = self.clone();
+        output.primary_key_index = encrypt_bytes_integral_nonce(
```
stringhandler: It's so strange to encrypt the key index. Is this really necessary? What's the danger in leaking this?

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -135,13 +131,13 @@ impl KeyManagerBackend for KeyManagerSqliteDatabase {
         let start = Instant::now();
         let conn = self.database_connection.get_pooled_connection()?;
         let acquire_lock = start.elapsed();
-        let mut km = KeyManagerStateSql::get_state(&branch, &conn)?;
-        self.decrypt_if_necessary(&mut km)?;
+        let km = KeyManagerStateSql::get_state(&branch, &conn)?;
+        let mut km = self.decrypt_value(km)?;
```
stringhandler: This probably doesn't need to be encrypted, the key index can be changed without decrypting

stringhandler (commented):
This is better, but I'd prefer the encryption and decryption to occur in the conversion to and from the `.....Sql` types, so that a `....Sql` type is always encrypted and the non-sql type is always decrypted, but zeroized

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
jorgeantonio21: I have no idea, tbh. I think this is legacy code

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
jorgeantonio21: I removed the decryption for `KeyManagerStateSql`.

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
jorgeantonio21: Refactored accordingly

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -47,41 +42,16 @@ const LOG_TARGET: &str = "wallet::key_manager_service::database::wallet";
 #[derive(Clone)]
 pub struct KeyManagerSqliteDatabase {
     database_connection: WalletDbConnection,
-    cipher: Arc<RwLock<XChaCha20Poly1305>>,
 }
 
 impl KeyManagerSqliteDatabase {
     /// Creates a new sql backend from provided wallet db connection
     /// * `cipher` is used to encrypt the sensitive fields in the database, if no cipher is provided, the database will
```
SWvheerden: These comments don't match the current workings and fields

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
```diff
@@ -150,50 +145,3 @@ pub struct KeyManagerStateUpdateSql {
     branch_seed: Option<String>,
     primary_key_index: Option<Vec<u8>>,
 }
-
-impl Encryptable<XChaCha20Poly1305> for KeyManagerStateSql {
```
SWvheerden, line 154: why have we removed this?
Are we not encrypting the index of the km states anymore?

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -47,41 +42,16 @@ const LOG_TARGET: &str = "wallet::key_manager_service::database::wallet";
...
     ///   not encrypt sensitive fields
-    pub fn new(
-        database_connection: WalletDbConnection,
-        cipher: XChaCha20Poly1305,
-    ) -> Result<Self, KeyManagerStorageError> {
-        let db = Self {
-            database_connection,
-            cipher: Arc::new(RwLock::new(cipher)),
-        };
+    pub fn new(database_connection: WalletDbConnection) -> Result<Self, KeyManagerStorageError> {
```
SWvheerden: is the key manager not going to encrypt its index's and data anymore?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -1332,18 +1332,19 @@ impl KnownOneSidedPaymentScriptSql {
         let script = known_script.script.to_bytes().to_vec();
```
SWvheerden, line 1332: I cant leave a comment on the line above, 
But `known_script.private_key` is not zeroized, and `known_script` is moved here and dropped after this function.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -807,28 +807,32 @@ impl Encryptable<XChaCha20Poly1305> for OutputSql {
             .to_vec()
     }
 
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.spending_key = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut output = self.clone();
```
SWvheerden: Why clone self here?
Its moved in here, not borrowed. 
We clone it, leave the original to drop at the end of the function, and we don't zeroize there data. 
Before we replaced the sensitive keys with encrypted values, so we did not have to zeroize.

SWvheerden: Is it because we changed the trait to not be `&mut`?
If so I undo that change, as its going to lead to a lot of cases like this.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -660,33 +662,37 @@ impl Encryptable<XChaCha20Poly1305> for ClientKeyValueSql {
     }
 
     #[allow(unused_assignments)]
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.value = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut client_kv = self.clone();
```
SWvheerden: same as above, this wont zeroize the encrypted values, it will leave them in memory after this function

base_layer/wallet/src/util/encryption.rs
```diff
@@ -43,8 +43,10 @@ pub trait Encryptable<C> {
     const CLIENT_KEY_VALUE: &'static [u8] = b"CLIENT_KEY_VALUE";
 
     fn domain(&self, field_name: &'static str) -> Vec<u8>;
-    fn encrypt(&mut self, cipher: &C) -> Result<(), String>;
-    fn decrypt(&mut self, cipher: &C) -> Result<(), String>;
+    fn encrypt(self, cipher: &C) -> Result<Self, String>
```
SWvheerden, line 46:
```suggestion
    fn encrypt(mut self, cipher: &C) -> Result<Self, String>
```
This leads to bad behavior as marked out in the other functions. 
Self is dropped after the function, and because its not mut, the values cannot be changed or zeroized. Thus after the function when self is dropped, the sensitive data is still in memory.

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1387,58 +1363,42 @@ impl Encryptable<XChaCha20Poly1305> for InboundTransactionSql {
         .to_vec()
     }
 
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.receiver_protocol = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut output = self.clone();
```
SWvheerden: same problem with zeroize

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1644,60 +1622,43 @@ impl Encryptable<XChaCha20Poly1305> for OutboundTransactionSql {
         .to_vec()
     }
 
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.sender_protocol = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut outbound_tx = self.clone();
```
SWvheerden: same problem with zeroize

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -2055,61 +2048,36 @@ impl Encryptable<XChaCha20Poly1305> for CompletedTransactionSql {
         .to_vec()
     }
 
-    fn encrypt(&mut self, cipher: &XChaCha20Poly1305) -> Result<(), String> {
-        self.transaction_protocol = encrypt_bytes_integral_nonce(
+    fn encrypt(self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        let mut output = self.clone();
```
SWvheerden: !

SWvheerden (changes requested):
We are going backwards with regards to zeroizing the data after use. 

I made a comment how to fix this in the PR

base_layer/wallet/src/key_manager_service/storage/sqlite_db/key_manager_state.rs
jorgeantonio21: I think there is no need for encrypting this data, but I might be wrong ?

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
jorgeantonio21: ? eheheh

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
jorgeantonio21: Addressed

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
jorgeantonio21: What is the purpose of encrypting the key index ? Is there any vulnerability while maintaining it non-encrypted ? I can revert the changes to encrypt it, as well.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
jorgeantonio21: Very good observation. I refactored it to zeroize the associated sensitive data.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
jorgeantonio21: Very good observation ! Let me add that the main purpose of the current PR was to change the design of 'encryption' and 'decryption' methods, to consume `self` instead of reusing it via `&mut self`. I have refactored this function to zeroize the sensitive data, that is encrypted, of `self`. Accordingly, 

```rs
fn encrypt(mut self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
        let mut output = self.clone();
        output.private_key =
            encrypt_bytes_integral_nonce(cipher, self.domain("private_key"), Hidden::hide(output.private_key))?;

        // zeroize sensitive data
        self.private_key.zeroize();

        Ok(output)
    }
```

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
jorgeantonio21: Addressed, see comment above.

base_layer/wallet/src/util/encryption.rs
jorgeantonio21: I refactored the instances of these methods to receive as input `mut self`, zeroize its sensitive contents and drop the values, after function returning. See comment above

base_layer/wallet/src/key_manager_service/storage/sqlite_db/mod.rs
SWvheerden: I remember the discussion we had and we decided to encrypt the index as the index can leak how many transactions the wallet has sent. 

Its not a security issue, but its a privacy one.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
SWvheerden: I am pretty sure you can change to those to this:
```rust
fn encrypt(mut self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
        self.private_key =
            encrypt_bytes_integral_nonce(cipher, self.domain("private_key"), Hidden::hide(self.private_key))?;

        Ok(self)
    }
```

Then we move it in and out, we don't have to clone, we don't have to zerozie

jorgeantonio21: Perfect, this is better, thanks :)

base_layer/wallet/tests/transaction_service_tests/service.rs
```diff
@@ -4368,11 +4368,17 @@ async fn test_resend_on_startup() {
...
         .is_ok());
-    assert!(alice_ts_interface
+
+    alice_ts_interface
         .transaction_service_handle
         .restart_transaction_protocols()
         .await
-        .is_ok());
+        .unwrap();
+    // assert!(alice_ts_interface
```
SWvheerden: ? delete?

SWvheerden (dismissed):
utACK

stringhandler (approved):
utACK

PR #5045, opened 2022-12-12
fix: wallet errors

Cifko:
Description
---
- Add proper error for `invalid network` for normal Tx
- increase the size of pubkey in contacts (the last two bytes were hidden)
- add error dialog to contacts

How Has This Been Tested?
---
Manually.


sdbondi (approved):
utACK

stringhandler (approved):

PR #5049, opened 2022-12-13
fix: support arbitrary range proof batching

### LINKED ISSUE #5048: [base node] Fix `pub fn batch_verify_range_proofs(..)` to not use powers of two, but batch all proofs

AaronFeickert:
Description
---
Removes an unnecessary power-of-two batch decomposition for range proof batch verification.

Closes [issue #5048](https://github.com/tari-project/tari/issues/5048).

Motivation and Context
---
Range proofs can be verified in batches for significant performance improvements. However, they are currently verified only in batches with power-of-two sizes. This may be due to a misunderstanding between aggregation (which requires such a restriction) and batching (which does not).

This work removes the restriction and provides the range proof verifier with all proofs in a batch.

Because of performance limitations, it may be useful to impose a size restriction in the verifier, where a very large batch is split into smaller batches.

How Has This Been Tested?
---
Existing tests pass.

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/transaction_output.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_output.rs
@@ -503,6 +503,11 @@ pub fn batch_verify_range_proofs(
     prover: &RangeProofService,
     outputs: &[&TransactionOutput],
 ) -> Result<(), RangeProofError> {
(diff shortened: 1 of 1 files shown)


CjS77 (approved):
LGTM

PR #5057, opened 2022-12-15
chore: fix clippy toolchain upgrade

brianp:
Description
---
Clean up clippy warnings and errors

Motivation and Context
---
You don't make many friends when you bring errors to the repo

How Has This Been Tested?
---
CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           command: lints
           args: clippy --all-targets --all-features
   build:
-    name: check nightly
+    name: cargo check
     runs-on: [ self-hosted, ubuntu18.04-high-cpu ]
     steps:
       - name: checkout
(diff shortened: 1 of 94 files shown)


delta1:
This is probably why you want "check nightly" CI to be the same as `rust-toolchain.toml`, you're going to get breaking fmt and clippy changes all the time

brianp:
@delta1 but if check-nightly is the same as rust-toolchain then you're not actually checking "nightly". Nightly failing is okay. It's like a warning, that something may break in the future. If it checks the same as rust-toolchain then there is literally no point to the job. It is checked on clippy, and test ci jobs.

Clippy failing here is valid.  The toolchain has changed, many warnings need to be addressed. Hence this PR :D

delta1:
Sure. I’m saying that previously it meant “run cargo check with the specified nightly toolchain in rust-toolchain” not “run cargo check with today’s nightly” 🤷‍♂️

brianp:
@delta1 Aye mate I'm picking up what yer puttin down now.

What needs to happen then is to revert that job back to the pinned version, and then change the name to "run cargo check" to remove any ambiguity about running it on nightly :D

PR #5058, opened 2022-12-15
feat: add cucumber utils for mining blocks and arrive at height

jorgeantonio21:
Description
---
To set up cucumber tests successfully, we need to be able to execute the following steps:

```
mining node {word} mines {int} blocks
all nodes are at height {int}
```

This PR implements the underlying logic in Rust.

Motivation and Context
---

How Has This Been Tested?
---


brianp (approved):
LGTM
Ack

Once we fix the base node bug we should start uncommenting implemented steps in from features. Those features will go green with the 1 or 2 steps they implement and slowly all features will come back to life.

PR #5059, opened 2022-12-15
fix: add further cucumber utils

jorgeantonio21:
Description
---
To set up cucumber tests successfully, we need to be able to execute the following steps:

```
node {word} is at height {int}
I have mining node {word} connected to base node {word} and wallet {word}
I wait for wallet {word} to have at least {int} uT
```

This PR implements the underlying logic in Rust.

Motivation and Context
---

How Has This Been Tested?
---


brianp (approved):
UtAck

PR #5061, opened 2022-12-16
fix: fixing broken base node in the feature-cucumber-rs branch

agubarev:
Description
---

Motivation and Context
---

How Has This Been Tested?
---


brianp (approved):
ACK

Tested and fixes the base node startup bug. Cheers @agubarev !

PR #5064, opened 2022-12-16
fix: update message and signature key types

AaronFeickert:
Description
---
Updates communications message and signature keys to use the new `Hidden` and `SafeArray` designs. Minor refactoring for clarity of design intent.

Motivation and Context
---
Communications messages are encrypted using the `ChaCha20` unauthenticated stream cipher, and signatures are encrypted using the `ChaCha20-Poly1305` authenticated stream cipher. However, the keys used for these purposes are defined using zeroize-on-drop `GenericArray` types; further, the hash derivation of the keys results in copies left in memory.

This work updates the key types to use the new `Hidden` and `SafeArray` designs. They are renamed to reflect their specific purposes, instantiated in place to reduce in-memory copies, and associated functions are also renamed to make their intent more clear.

How Has This Been Tested?
---
Existing tests pass.

### DIFF
--- a/comms/dht/src/crypt.rs
+++ b/comms/dht/src/crypt.rs
@@ -25,11 +25,10 @@ use std::{iter, mem::size_of};
 use chacha20::{
     cipher::{NewCipher, StreamCipher},
     ChaCha20,
-    Key,
     Nonce,
 };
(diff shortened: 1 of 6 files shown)


SWvheerden (approved):
utACK

stringhandler (approved):

PR #5065, opened 2022-12-20
feat: add shutdown clone

Cifko:
Description
---
Added shutdown clone. So now we can trigger it from multiple places.
This change will trigger only proper shutdown.

How Has This Been Tested?
---
Manually, running two tests.


brianp (approved):
Looks great. Awesome work @Cifko !

PR #5067, opened 2022-12-20
feat: fix miner

Cifko:
Description
---
Add working miner.

How Has This Been Tested?
---
Running a mine step a seeing the wallet DB with an output.


integration_tests/tests/utils/miner.rs
```diff
@@ -70,225 +41,42 @@ pub fn register_miner_process(world: &mut TariWorld, miner_name: String, base_no
         wallet_name,
         mine_until_height: 100_000,
     };
-    world.miners.insert(miner_name, miner);
-}
-
-pub async fn mine_blocks(world: &mut TariWorld, miner_name: String, num_blocks: u64) {
```
jorgeantonio21, line 76: I think that for the time being we will need this code

Cifko: I put everything back in. And I'm using the new approach for the general mining.

brianp (dismissed):
LGTM
UtAck

PR #5077, opened 2022-12-30
feat: add remaining steps for wallet transfer

jorgeantonio21:
Description
---
Add remaining steps in the `WalletTransfer.feature` file.

Motivation and Context
---

How Has This Been Tested?
---


integration_tests/tests/cucumber.rs
```diff
@@ -400,24 +416,44 @@ async fn prune_node_connected_to_base_node(
...
+        .address
+        .to_hex();
+    let tx_ids = world.wallet_tx_ids.get(&wallet_address).unwrap();
 
     let num_retries = 100;
 
     for tx_id in tx_ids {
         println!("waiting for tx with tx_id = {} to be mined_confirmed", tx_id);
-        'inner: for _ in 0..num_retries {
+        for retry in 0..=num_retries {
```
SWvheerden, line 438: nit: Technically this counts higher than the num of retries by 1

integration_tests/tests/cucumber.rs
```diff
@@ -641,14 +681,17 @@ async fn send_amount_from_source_wallet_to_dest_wallet_without_broadcast(
     let tx_id = tx_res.transaction_id;
 
     // insert tx_id's to the corresponding world mapping
-    let mut source_tx_ids = world.wallet_tx_ids.get(&source_wallet_pubkey).unwrap().clone();
-    let mut dest_tx_ids = world.wallet_tx_ids.get(&dest_wallet_pubkey).unwrap().clone();
+    // insert tx_id's to the corresponding world mapping
```
SWvheerden: double comment?

integration_tests/tests/cucumber.rs
```diff
@@ -754,14 +796,17 @@ async fn send_one_sided_transaction_from_source_wallet_to_dest_wallt(
     }
 
     // insert tx_id's to the corresponding world mapping
-    let mut source_tx_ids = world.wallet_tx_ids.get(&source_wallet_pubkey).unwrap().clone();
-    let mut dest_tx_ids = world.wallet_tx_ids.get(&dest_wallet_pubkey).unwrap().clone();
+    // insert tx_id's to the corresponding world mapping
```
SWvheerden: double comment

integration_tests/tests/cucumber.rs
```diff
@@ -68,6 +70,8 @@ use crate::utils::{
 
 pub const LOG_TARGET: &str = "cucumber";
 pub const LOG_TARGET_STDOUT: &str = "stdout";
+const BLOCK_REWARD: u64 = 5_000_000;
```
SWvheerden: this does not seem correct?
How did you calculate this?
This is supposed to be a decaying nonconstant number + varying fee

integration_tests/tests/cucumber.rs
```diff
@@ -1509,7 +1642,518 @@ async fn transfer_tari_from_wallet_to_receiver(world: &mut TariWorld, amount: u6
...
+    base_node: String,
+) {
+    let peer_seeds = world.base_nodes.get(&base_node).unwrap().seed_nodes.clone();
+    println!(
+        "Start a new wallet {} connected to base node {}",
+        wallet.as_str(),
+        base_node.as_str()
+    );
+    spawn_wallet(world, wallet.clone(), Some(base_node.clone()), peer_seeds, None).await;
+    let num_blocks = amount * 1_000_000 / BLOCK_REWARD;
```
SWvheerden: you can get the actual block reward from the consensus constants code directly?

integration_tests/tests/utils/mod.rs
```diff
@@ -32,6 +32,8 @@ pub mod wallet;
 pub mod wallet_client;
 pub mod wallet_process;
 
+use tokio::time::Instant;
```
SWvheerden: why this change?
Did it not compile before

integration_tests/tests/cucumber.rs
jorgeantonio21: The idea is to have a failure check `retry`, but can make it consistent with the other methods, as well.

integration_tests/tests/cucumber.rs
jorgeantonio21: Removed

integration_tests/tests/cucumber.rs
jorgeantonio21: I did get it from the JS version. Take a look at `integration_tests/helpers/constants.js`

integration_tests/tests/cucumber.rs
jorgeantonio21: Removed double comments

integration_tests/tests/utils/mod.rs
jorgeantonio21: Artifacts from a previous version

integration_tests/tests/cucumber.rs
SWvheerden: There is a very good chance this might only work under very strict circumstances. The JS code was old, and not always maintained. Perhaps see if you can calculate it some other way, perhaps using the constants directly. Using RUST we now have access to that code. 

Although this is still a strange way of calculating the number of blocks mined.

jorgeantonio21: Refactored code to use ConsensusConstants directly and compute block reward from height

SWvheerden:
LGTM

brianp (changes requested):
Only request would be to remove the changes from `integration_tests/features/WalletTransfer.feature`

They don't hurt anything really, but they're also unnecessary changes.

Issue #5042, opened 2022-12-12
Consolidate Validators


stringhandler:
# Problem
Currently, there is duplicate validation logic in multiple places. This makes it hard to reason about whether it is correct according to the RFC, but it also means that as new rules are added or fixed, it could be done in one place, but not the other.

# Proposed solution
Instead, let's change the structure so that a validation rule is found in at most one validator struct, and only those structs are used. 

In all cases, the validation rules should be implemented according to 

These traits should be created:

## Headers
```rust 
pub struct [Internally]ValidHeader(pub Arc<BlockHeader>);

pub trait InternalConsistencyHeaderValidator {
   /// Validates a header in isolation, i.e. without looking at previous headers
   fn validate(&self, header: &BlockHeader) -> Result<ValidHeader, ValidationError>;
}

pub trait ChainLinkedHeaderValidator {
  /// Takes a valid header and validates it in context of previous headers in the chain
  fn validate(&self, header: &ValidHeader, [... state from the db needed for validation...]) -> Result<ChainHeader, ValidationError>;
}
```

There are a few places in the code that a header is validated:
1. During initial sync. In this case it should be validated with the `InternalConsistencyHeaderValidator` impl and then the `ChainLinkedHeaderValidator`.
2. When adding a block. Both should be run
3. When receiving a header via ChainMetadata. The `InternalConsistencyValidator` should be run before saving.

Similarly, the block body should be separated:

```rust
pub trait InternalConsistencyOutputValidator {
  pub fn validate(&self, output: &TransactionOutput) -> Result<(), ValidationError>;
} 

pub trait InternalConsistencyKernelValidator {
  pub fn validate(&self, kernel: &TransactionKernel) -> Result<(), ValidationError>;
}

/// Replaces OrphanValidation and MempoolTransactionValidator
pub trait InternalConsistencyAggregateBodyValidator {
   fn validate(&self, body: AggregateBody, offset: BlindingFactor, script_offset: BlindingFactor) -> Result<(), ValidationError>;
}

pub trait ChainLinkedAggregateBodyValidator {
  fn validate(&self, body: AggregateBody, header: ChainHeader, [.... state or db needed to validate....]) -> Result<(), ValidationError>;
}

```

# Acceptance criteria
Once done, the above validators should be the only validations done. I would also aim to move the `validate_internal_consistency` method from the inputs, outputs and kernels so that all the logic is in the structs implementing these traits. 

# Other requirements
[...]

PR #5079, opened 2023-01-03
refactor(validations): unify validators and remove duplicated logic

mrnaveira:
Description
---
The main idea behind this refactor was to create reusable and composable validation components, so that a particular validation function is found in one and only one of the new components. I tried to preserve the existing logic as much as possible. Also, all the validations are now encapsulated inside the `base_layer/core/src/validation` crate.

The core validation components, reused by every other validators are:
* **Aggregated bodies** (`validation/aggregate_body` crate): used directly as well as part of transactions and block validations.
* **Block headers** (`validation/header` crate): only the `HeaderFullValidator` was created, as the existing logic didn't required splitted validators for internal consistency and chain link verification.

Additionally, as a convenience layer on top of aggregated bodies, we have both **transactions** (`validation/transaction` crate) and **block bodies** (`validation/block_body` crate) validators. Some of them add some extra validations on top of the aggregate body ones. Maybe in the future some of these could be pushed down into the aggregate body validations to make them more generic.

To facilitate code reviewing, this is the list with the correspondence between old validators and new ones:
* The `AggregateBody.validate_internal_consistency` method was removed from the struct into a new, separated, reusable `AggregateBodyInternalConsistencyValidator` component. This required to add the `base_node` feature to the `tari_core` dependency of the `wallet_ffi`, so if for whatever reason this is not acceptable (binary size?) we will need to move back the validation to the `AggregateBody` struct.
* `HeaderValidator` -> `HeaderFullValidator`
* `MempoolValidator` was a trait with 3 different implementations, all could be merged into a single one and implemented with `TransactionFullValidator`
* `BodyOnlyValidator` -> `BlockBodyFullValidator`
* `OrphanBlockValidator` -> `BlockBodyInternalConsistencyValidator`
* `BlockHeaderSyncValidator` (`base_layer/core/src/base_node/sync/header_sync/validator.rs`) still exists for API convenience, but delegates the actual validation rules into the new `HeaderFullValidator` (used by the mempool as well), so we avoid duplicities
* `async_validator.rs` is removed, we are now reusing a `BlockBodyFullValidator` wrapped in a `tokio::task`

Motivation and Context
---
Currently there is duplicate validation logic in multiple places. This makes it hard to reason about whether it is correct according to the RFC, but it also means that as new rules are added or fixed, it could be done in one place, but not the other.

Solves https://github.com/tari-project/tari/issues/5042

How Has This Been Tested?
---
Existing unit tests pass, with the exception of: 
* `block_event_and_reorg_event_handling`
* `test_retrieve`
* `test_base_node_wallet_rpc`

I ignored them as it seems that they will need to be rewritten taking into account that the mempool now performs a lot more validations on incoming transactions.


base_layer/core/src/validation/aggregated_body.rs
```diff
@@ -0,0 +1,217 @@
...
+    ///
+    /// This function does NOT check that inputs come from the UTXO set
+    /// The reward is the total amount of Tari rewarded for this block (block reward + total fees), this should be 0
+    /// for a transaction
+    pub fn validate_internal_consistency(
+        &self,
+        body: &AggregateBody,
+        tx_offset: &BlindingFactor,
+        script_offset: &BlindingFactor,
+        bypass_range_proof_verification: bool,
```
SWvheerden: NIT: Because its no a trait I would rather make this a property of the `InternalConsistencyAggregateBodyValidator` rather than the function `validate_internal_consistency`

base_layer/core/src/validation/new/aggregate_body/chain_aggregate_body_validator.rs
```diff
@@ -0,0 +1,327 @@
...
+    pub fn new(db: BlockchainDatabase<B>) -> Self {
+        Self {
+            db,
+            factories: CryptoFactories::default(),
+        }
+    }
+
+    pub async fn validate(
+        &self,
+        tx: &Transaction,
```
SWvheerden:
```suggestion
        body: &AggregateBody,
```
make it body
These tests will need to be done by the mempool and block validation. 
Block validation does not have the tx anymore.

SWvheerden: We need all the helper functions to be the same code/function

base_layer/core/src/validation/new/aggregate_body/internal_aggregate_body_validator.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// The reward is the total amount of Tari rewarded for this block (block reward + total fees), this should be 0
+    /// for a transaction
+    #[allow(dead_code)]
+    pub fn validate(
+        &self,
+        body: &AggregateBody,
+        tx_offset: &BlindingFactor,
+        total_reward: MicroTari,
+    ) -> Result<(), ValidationError> {
+        // TODO: include bypass_range_proof_verification parameter?
```
SWvheerden: I say yes, but make it property of `InternalConsistencyAggregateBodyValidator`

base_layer/core/src/validation/new/aggregate_body/internal_aggregate_body_validator.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        validate_is_not_coinbase(body)?;
```
SWvheerden: why do we want to check this?
Coinbases also have a transaction with just an output and kernel. 

We might want to check that it contains only 1 coinbase thou

SWvheerden (commented):
this looks good so far

base_layer/core/src/validation/transaction/transaction_chain_validator.rs
```diff
@@ -0,0 +1,80 @@
...
+        };
+        verify_timelocks(tx, tip_height)?;
+
+        Ok(())
+    }
+}
+
+// This function checks that all the timelocks in the provided transaction pass. It checks kernel lock heights and
+// input maturities
+fn verify_timelocks(tx: &Transaction, current_height: u64) -> Result<(), ValidationError> {
```
SWvheerden: This should be checked for blocks as well, 
This function should check on the body

SWvheerden (commented):
Looks good so far

base_layer/core/src/validation/aggregate_body/aggregate_body_internal_validator.rs
```diff
@@ -0,0 +1,522 @@
...
+fn verify_metadata_signatures(body: &AggregateBody) -> Result<(), ValidationError> {
+    trace!(target: LOG_TARGET, "Checking sender signatures");
+    for o in body.outputs() {
+        o.verify_metadata_signature()?;
+    }
+    Ok(())
+}
+
+/// this will validate the script offset of the aggregate body.
+fn validate_script_offset(
```
SWvheerden:
```suggestion
/// this will validate the script and script offset of the aggregate body.
fn validate_script_and_script_offset(
```
nit, this does both script validation and script offset validation. 
Script validation can be done without script offset validation, but script offset cannot be done without running the script. So validating the offset, means the script must also be checked, but this is detailed knowledge of TariScript, that should not be required for reading the code.

SWvheerden (commented):
Looks good

base_layer/core/src/base_node/state_machine_service/initializer.rs
```diff
@@ -110,11 +110,10 @@ where B: BlockchainBackend + 'static
             let peer_manager = handles.expect_handle::<Arc<PeerManager>>();
 
             let sync_validators = SyncValidators::full_consensus(
-                db.clone(),
                 rules.clone(),
                 factories,
                 bypass_range_proof_verification,
-                config.blockchain_sync_config.validation_concurrency,
+                // config.blockchain_sync_config.validation_concurrency,
```
SWvheerden: remove?

SWvheerden (dismissed):
utACK

SWvheerden:
It seems this fails general sync to tip
```
2023-01-18 09:40:00.111181000 [c::mp::mempool_storage] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] DEBUG Compaction took 41.00ns // base_layer/core/src/mempool/mempool_storage.rs:184
2023-01-18 09:40:00.111184000 [c::mp::mempool_storage] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] DEBUG Mempool stats: Unconfirmed: 0, In Reorg Pool: 0, Total Weight: 0g // base_layer/core/src/mempool/mempool_storage.rs:185
2023-01-18 09:40:00.111275000 [c::bn::block_sync] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] DEBUG Validating block body #103 (PoW = Monero, 2 input(s), 3 output(s), 2 kernel(s), latency: 4.58µs) // base_layer/core/src/base_node/sync/block_sync/synchronizer.rs:288
2023-01-18 09:40:00.111308000 [c::val::aggregate_body_internal_consistency_validator] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] TRACE Checking kernel signatures // base_layer/core/src/validation/aggregate_body/aggregate_body_internal_validator.rs:150
2023-01-18 09:40:00.111420000 [c::val::aggregate_body_internal_consistency_validator] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] TRACE Checking kernel total // base_layer/core/src/validation/aggregate_body/aggregate_body_internal_validator.rs:169
2023-01-18 09:40:00.111426000 [c::val::block_body_internal_consistency_validator] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] WARN  Validation failed on block:d430baddbcacfacef842c85ecd28b86ca0283aebf752feda6e70154357e18a1b:TransactionError(MissingTransactionInputData) // base_layer/core/src/validation/block_body/block_body_internal_validator.rs:126
2023-01-18 09:40:00.111449000 [c::bn::async_db] [1ce2b9246f627755636fd140a35404c0c1192bd3b1aaa46f9fff2165af33e95a,487d2486fac5a4c0045b9f32d5] TRACE [write] Entered blocking thread. trace_id: 2302244471 // base_layer/core/src/chain_storage/async_db.rs:73
```

base_layer/core/src/base_node/state_machine_service/initializer.rs
mrnaveira: yes, that's a parameter of the former `async_validator`, not used anymore so I just deleted the comment to avoid confusion.

base_layer/core/tests/base_node_rpc.rs
```diff
@@ -123,6 +123,8 @@ async fn setup() -> (
     )
 }
 
+// FIXME: this test fails due to a unexpected validation failure of input maturity
+#[ignore]
 #[tokio::test]
 #[allow(clippy::identity_op)]
 async fn test_base_node_wallet_rpc() {
```
SWvheerden, line 130: So this fails with: 
```
---- test_base_node_wallet_rpc stdout ----
Validation failed due to error: The transaction is invalid: Input maturity not reached
thread 'test_base_node_wallet_rpc' panicked at 'assertion failed: `(left == right)`
  left: `ValidationFailed`,
 right: `DoubleSpend`', 
```
Any reason why the reason changed? 
Input maturity not reached and double spend is two completely different failures.

SWvheerden: Just make sure the constants are created with the correct coinbase lock height, in this case, should be 1. Then the test should pass.

SWvheerden:
Managed to sync to tip with the new changes

stringhandler (approved):
Tested and synced to tip

PR #5084, opened 2023-01-05
fix: add const to FixedHash::zero

sdbondi:
Description
---
Make `FixedHash::zero` a const fn

Motivation and Context
---
A function returning a constant value should be set as const to allow for minor compiled code optimisations. A zero hash is used more often in the DAN layer to indicate a genesis node.

How Has This Been Tested?
---
Minor change, clippy passes


SWvheerden (approved):
LGTM

stringhandler (approved):

PR #5086, opened 2023-01-06
fix: header sync start info

SWvheerden:
Description
---
When starting header sync if the remote tip is less than the local tip but has a higher pow, the description is wrong when it starts. 
```
10:06 v0.42.0 esmeralda State: Syncing headers: (5c38651e1b7ec007) 5241/5180 (101%) Tip: 5153
```

Motivation and Context
---
Fixes it to show the starting block height on startup and not the local tip before reorg. 

How Has This Been Tested?
---
Manual


sdbondi (approved):
LGTM

stringhandler (approved):

PR #5090, opened 2023-01-09
feat: graceful wallet grpc shutdown

Cifko:
Description
---
Add support for gracefully shutting down the grpc server.

Motivation and Context
---
GRPC server is responsible for keeping some instance of WalletDBConnection, which means it holds the exclusive lock for the wallet db. In the real life and in the js tests it was not a problem. System took care of it. But in the rust cucumber we are not ending the whole process. So system is not going to kill the grpc for us. We have to end it.

How Has This Been Tested?
---
Manually, I run some test showing if the count on the Arc is going to zero after this change. Without it the GRPC was holding 3 clones of it.


brianp (approved):

jorgeantonio21 (approved):
This is great ! Thanks @Cifko

PR #5092, opened 2023-01-09
fix: wallet reuse existing tor address

### LINKED ISSUE #5089: [Wallet] on restart changes onion address
On every restart, the wallet gets a new network onion address. This causes all peers to have an old address to the wallet. 

Screenshot 2023-01-09 at 11 39 07 1
after restart
Screenshot 2023-01-09 at 11 39 30

SWvheerden:
Description
---
Fixes wallet to reuse existing tor address rather than create a new one each time it restarts

Motivation and Context
---
A wallet should not create a new tor address on startup, but rather reuse the existing ones.

How Has This Been Tested?
---
Manual

Fixes: https://github.com/tari-project/tari/issues/5089


stringhandler:
Labelling this as controversial. The fix is simple enough, but I have two concerns:
1. This is a very easy mistake to make. There is clearly an unintended consequence in cloning, so there are possibly other places where this "bug" could occur. There is no indication to the developer of the correct way of doing this. I would like to look further into the cause
2. More importantly, a change in address should not affect the network. In fact, there should be a cucumber test checking this. There seems to be something missing in the comms code that would handle this case

SWvheerden:
As to the second point. 
Yeah, we need to handle this better but the problem is if both clients have the wrong address, they have no way of knowing that the address is wrong. All they know is that the address they have does not respond to any dial request. 

If only 1 client has the wrong address, then the client should update it and fix it with the incoming dial. 

Wallet could still send transactions as all transactions went through SAF, and SAF does not use the tor address for direct communication.

stringhandler:
Added a bit more resilience in #5142 to allow multiple addresses. Tested with a with a wallet turning off and on, getting a new tor address and it came online in time

stringhandler:
I'd like to merge #5142 first, since having this bug makes it easier to test

stringhandler (approved):

PR #5096, opened 2023-01-11
feat: add graceful shutdown of base node

Cifko:
Description
---
The non-interactive node was not expecting shutdown to be triggered. Because in real life it will never happen. It was expecting ctrl-c to be pressed. But for the sake of tests the shutdown should be monitored.

Motivation and Context
---
Without it the exclusive locks will panic

How Has This Been Tested?
---
Simple test scenario, start node, stop node, start again.


sdbondi (approved):
LGTM

jorgeantonio21 (approved):
LGTM

brianp:
UtAck

PR #5097, opened 2023-01-11
test: sync feature updates

brianp:
Description
---
Implement the steps for the Sync.feature scenarios

Motivation and Context
---
The great cucumber port! Regression testing and all.

How Has This Been Tested?
---
Locally, CI

### DIFF
--- a/integration_tests/tests/cucumber.rs
+++ b/integration_tests/tests/cucumber.rs
@@ -99,8 +99,8 @@ use crate::utils::{
 pub const LOG_TARGET: &str = "cucumber";
 pub const LOG_TARGET_STDOUT: &str = "stdout";
 const CONFIRMATION_PERIOD: u64 = 4;
-const NUM_RETIRES: u64 = 240;
-const RETRY_TIME_IN_MS: u64 = 500;
+const TWO_MINUTES_WITH_HALF_SECOND_SLEEP: u64 = 240;
+const HALF_SECOND: u64 = 500;
 
 #[derive(Error, Debug)]
 pub enum TariWorldError {
@@ -216,6 +216,7 @@ async fn start_wallet(world: &mut TariWorld, wallet_name: String, node_name: Str
     spawn_wallet(world, wallet_name, Some(node_name), seeds, None, None).await;
 }
 
+#[given(expr = "I have a base node {word} connected to all seed nodes")]
 #[when(expr = "I have a base node {word} connected to all seed nodes")]
 async fn start_base_node_connected_to_all_seed_nodes(world: &mut TariWorld, name: String) {
     spawn_base_node(world, false, name, world.all_seed_nodes().to_vec(), None).await;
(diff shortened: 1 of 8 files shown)


integration_tests/tests/cucumber.rs
```diff
@@ -98,8 +99,8 @@ use crate::utils::{
 pub const LOG_TARGET: &str = "cucumber";
 pub const LOG_TARGET_STDOUT: &str = "stdout";
 const CONFIRMATION_PERIOD: u64 = 4;
-const NUM_RETIRES: u64 = 240;
-const RETRY_TIME_IN_MS: u64 = 500;
+const TWO_MINUTES_WITH_HALF_SECOND_SLEEP: u64 = 240;
```
Cifko, line 102: This name is confusing. It's not actually used in sleep. It's kind of a number of retries, but it's multiplied by some values in the code. I suppose it is a value when the number of retries runs only this many times with the half second sleep it will be two minutes. Right?

integration_tests/tests/cucumber.rs
```diff
@@ -3828,18 +3766,18 @@ async fn receive_an_error(_world: &mut TariWorld, _error: String) {
...
+    spawn_base_node_with_config(world, true, delayed_node, vec![node], base_node_config).await;
 }
 
 #[then(expr = "node {word} has reached initial sync")]
 async fn node_reached_sync(world: &mut TariWorld, node: String) {
     let mut client = world.get_node_client(&node).await.unwrap();
     let mut longest_chain = 0;
 
-    for _ in 0..(NUM_RETIRES * 2) {
+    for _ in 0..(TWO_MINUTES_WITH_HALF_SECOND_SLEEP * 11) {
```
Cifko, line 3780: I'm just wondering, was this 22 minutes timeout in the `js` cucumber? Feels like it's way too much. Anyway we will probably (somewhere in the future) fine tunes these arbitrary values.

integration_tests/tests/cucumber.rs
```diff
@@ -3857,6 +3795,104 @@ async fn node_reached_sync(world: &mut TariWorld, node: String) {
...
+        let result = client
+            .get_transaction_info(grpc::GetTransactionInfoRequest {
+                transaction_ids: vec![tx_id],
+            })
+            .await
+            .unwrap();
+
+        last_status = result.into_inner().transactions.last().unwrap().status;
+
+        if let 1 | 2 | 6 = last_status {
```
Cifko, line 3822: Can't we use enum values instead of 1,2,6?

Cifko (approved):
utACK

integration_tests/tests/cucumber.rs
brianp: We can, it just seemed faster. I'll put the enum in.

integration_tests/tests/cucumber.rs
brianp: Yup 22 minutes in the previous one too. 😓

integration_tests/tests/cucumber.rs
brianp: > I suppose it is a value when the number of retries runs only this many times with the half second sleep it will be two minutes. Right?

yeah, but if confusing at first, I'll revert it.

brianp: DM suggested it made more sense in the end so we'll leave it

Issue #5003, opened 2022-12-06
Add functionality for update encryption on wallet, when passphrase gets updated by user


jorgeantonio21:

SWvheerden:
Currently, the wallet does not have the ability to change its passphrase.
If the user wants to change it, the wallet needs to be recovered to a new wallet with a new passphrase.

AaronFeickert:
Now that [PR 5154](https://github.com/tari-project/tari/pull/5154) is merged, it's possible to support this. A few things to keep in mind:
- If the client detects that the database is encrypted using an outdated parameter set, it can update to the newest parameter set when the user enters their passphrase.
- As noted in [this comment](https://github.com/tari-project/tari/pull/5154#issuecomment-1415963906), there are persistent access attacks that could be possible under certain circumstances and threat models.

AaronFeickert:
Some other questions to consider relating to password changes:
- Does anything need to change for `gRPC` authentication?
- Should the `CipherSeed` password be set or updated? By default, it is not, but still considered "encrypted" due to the `CipherSeed` API.
- If it is decided to allow main key resampling to mitigate the effects of persistent-access attacks, can this be done atomically to avoid the case where only some database fields are updated? The idea of atomicity still applies when updating only the secondary key, but presumably this is less likely to go awry.

PR #5111, opened 2023-01-12
feat: provide password feedback

### LINKED ISSUE #5101: Add passphrase complexity feedback
Currently, the only check performed on a new interactively-supplied passphrase is that it not be empty. If the user enters a weak passphrase, there is no feedback indicating this. While it should ultimately be up to the user if they wish to use a weak passphrase, there should be actionable feedback to guide the user if they wish to use a stronger passphrase.

Estimating the practical entropy of a passphrase and proving good feedback is nontrivial, but libraries like [zxcvbn](https://crates.io/crates/zxcvbn) try their best to do so. It may be useful to score the passphrase entered by the user with such a library, use the library to present actionable feedback, and then give the user the option to choose a new passphrase if they wish.

AaronFeickert:
Description
---
Adds wallet password complexity feedback. Allows empty passwords. Adds a warning indicating that password changing functionality is [not yet implemented](https://github.com/tari-project/tari/issues/5003). Adds tests.

Closes [issue 5101](https://github.com/tari-project/tari/issues/5101).

Motivation and Context
---
The only check on a wallet password is that it not be empty. This introduces two issues:
- The user has no feedback on the practical strength of their password.
- The user may specifically wish not to set a password for fail-available backups.

This PR uses the [zxcvbn](https://crates.io/crates/zxcvbn) password complexity library to score a password and provide actionable feedback to the user. When the user enters a new password or changes their password, feedback is displayed if applicable. This is informational; the user is free to ignore the feedback if they wish.

Further, the user is now allowed to set an empty password, which may be desired for backups that must fail available. A warning is displayed if this happens.

Finally, a warning message is displayed during the password changing process to indicate that this functionality is incomplete.

How Has This Been Tested?
---
Existing CI passes. New tests pass. Tested manually for new wallets.

### DIFF
--- a/applications/tari_console_wallet/src/init/mod.rs
+++ b/applications/tari_console_wallet/src/init/mod.rs
@@ -59,6 +59,7 @@ use tari_wallet::{
     WalletConfig,
     WalletSqlite,
 };
+use zxcvbn::zxcvbn;
 
 use crate::{
     cli::Cli,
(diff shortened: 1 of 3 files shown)


AaronFeickert:
This presents a somewhat suboptimal user flow. If the user enters a weak password, sees the feedback, and then wishes to try a better password, they need to restart the password changing flow.

Another option is to provide feedback prior to the confirmation prompt and return to the initial prompt. If the user enters the same password, continue to the confirmation prompt without feedback. This adds no additional steps to the case where the password is strong (as there is no feedback), but adds an additional password entry step to the case where the password is weak (and feedback is presented). Not sure if this is preferable, or just annoying to the user.

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -77,6 +78,46 @@ pub enum WalletBoot {
...
+            Ok(scoring) => scoring
+                .feedback()
+                .to_owned()
+                .map(|feedback| feedback.suggestions().to_owned()),
+            _ => None,
+        }
+    } else {
+        // Fail empty if we can't convert the passphrase back to a string, which shouldn't happen
+        None
+    }
```
AaronFeickert: This `if`/`match` nesting feels non-idiomatic to me. Suggestions on how to improve it are welcome!

AaronFeickert: Updated.

stringhandler (approved):

CjS77 (approved):
+1

PR #5114, opened 2023-01-13
feat: add recovery.feature steps

jorgeantonio21:
Description
---
Add missing steps for file `Recovery.feature`.

Motivation and Context
---

How Has This Been Tested?
---


integration_tests/tests/cucumber.rs
```diff
@@ -3879,10 +3925,29 @@ async fn force_sync_node_with_an_army_of_pruned_nodes(
...
+    config_path.push("config.toml");
+
+    let mut cli = bn_default_cli(base_path, config_path.into_os_string().into_string().unwrap());
+
+    cli.rebuild_db = true;
+
+    let peers = base_node_ps.seed_nodes.clone();
+    let is_seed_node = base_node_ps.is_seed_node;
+
+    spawn_base_node(world, is_seed_node, node.clone(), peers, Some(cli)).await
```
brianp, line 3948: If only 1 step needs `Some(cli)` and the rest are `None` it could be better to use the `spawn_base_node_with_config` function directly and keep the `spawn_base_node` as the simple quick setup.

PR #5124, opened 2023-01-17
fix!: prune mode sync

### LINKED ISSUE #5099: Horizon sync fails with kernel MMR mismatch at height 206 on Esme
```

18:04 WARN  Error during sync:MMR did not match for Kernel at height 206. Expected 8314ddab1ae9258093b3726aebfc546ef024e8d04a2459cd7edddb28199d9a70 to equal 5067c47d9f49743aeecafe202a9251a0bbc4e4c53e40074bddc89f3f2227d8f6
18:04 WARN  Synchronizing horizon state has failed. MMR did not match for Kernel at height 206. Expected 8314ddab1ae9258093b3726aebfc546ef024e8d04a2459cd7edddb28199d9a70 to equal 5067c47d9f49743aeecafe202a9251a0bbc4e4c53e40074bddc89f3f2227d8f6
```

## Steps to reproduce

- Start a new base node with pruning_horizon set > 0 (mine was set to 1000)
- Sync headers
- Kernels sync starts and fails on 206

SWvheerden:
Description
---
Fixes the syncing of pruned nodes

Motivation and Context
---
The utxo and kernel keys are stored as LE bytes, but when reading with the order, they need to be BE bytes to return them in the same order as they were written. When we retrieve these for normal block sync, we call `.sort()` on them with the construction of the block so the order does not matter. 
But prune mode streams the utxo and kernels via MMR positions. For this, the order is important and needs to be returned in the same order as requested. 

This broke to do no cucumber tests testing prune mode sync. This should be fixed when enabling pruned mode. 

How Has This Been Tested?
---
Manual and new unit tests. 

Fixes: https://github.com/tari-project/tari/issues/5099

---
REQUIRES THAT BASE NODES RESYNC AND RECALCULATE DB KEYS

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -1537,7 +1537,7 @@ impl LMDBDatabase {
                 &u64::from(pos + 1).to_be_bytes(),
             )
             .or_not_found("BlockHeader", "mmr_position", pos.to_string())?;
-            let key = OutputKey::try_from_parts(&[hash.as_slice(), pos.to_le_bytes().as_slice()])?;
+            let key = OutputKey::try_from_parts(&[hash.as_slice(), pos.to_be_bytes().as_slice()])?;
             debug!(target: LOG_TARGET, "Pruning output: {}", key);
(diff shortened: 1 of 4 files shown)


stringhandler:
CI failing

stringhandler (approved):

Issue #4339, opened 2022-07-22
Examine peer banning logic


AaronFeickert:
Since #4336 introduces a new message decryption failure mode, this is a good opportunity to review peer banning logic more broadly. Here are some hypothetical scenarios that should not occur:

1. A node attempts to decrypt a message not intended for it, so the message signature authenticated decryption fails, which does not represent malicious behavior. The node bans the peer that forwarded the message.
2. A sender crafts a message maliciously in such a way that only the recipient can detect the manipulation. The recipient bans the forwarding peer, but not the sender.
3. A forwarding node manipulates a message it receives in a way that only the recipient can detect, and then passes it to an honest forwarding node before delivery to its recipient. The recipient bans the honest forwarding node, but not the malicious forwarding node.

The first scenario would presumably result in mass bans even in the absence of active adversaries. The second and third scenarios could be easily abused by an active adversary to result in a denial-of-service attack on the network.

In general, encryption-related failures should only result in a peer ban if the following are true:
- The failure mode represents malicious behavior.
- The banning node knows definitively that its forwarding peer performed the malicious behavior. Note that identifying a message as malicious and forwarding it anyway is itself malicious behavior.

SWvheerden:
| Method                                                                             | Link                                                                                                                                                                                                                                                                                                                       | Where? (context)                                           | Reason to ban                                                                               | Who?                                | Cause | Ban duration       | Is this correct?                 |
| ---------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------- | ------------------------------------------------------------------------------------------- | ----------------------------------- | ----- | ------------------ | -------------------------------- |
| base\_node::comms\_interface::inbounds\_handlers::request\_full\_block\_from\_peer | [https://github.com/tari-project/tari/blob/9aed188ccd7b2caffa10e061ba0d0c3253fb0b16/base\_layer/core/src/base\_node/comms\_interface/inbound\_handlers.rs#L612](https://github.com/tari-project/tari/blob/9aed188ccd7b2caffa10e061ba0d0c3253fb0b16/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs#L612) | Received block header with pow                             | Peer sent block header with valid pow, but when asked for the full block did not provide it | peer who could not privde the block |       | CUSTOM: 100 Secs   | Y                                |
[...]

PR #5129, opened 2023-01-19
fix: do not propagate unsigned encrypted messages

AaronFeickert:
Description
---
Updates message handling to detect and prevent propagation of unsigned encrypted messages. Minor refactoring for clarity and efficiency. Improves comments.

See [this issue](https://github.com/tari-project/tari/issues/4339) for more on banning logic.

Motivation and Context
---
Encrypted messages must include an encrypted signature that verifies the sender and prevents malleability. Currently, this is only checked after an encrypted message reaches its recipient. If the recipient sees that the message does not include a signature, it will ban its forwarding peer. This is technically correct behavior on the part of the recipient, since either the forwarding peer rendered the message invalid, or itself forwarded a known invalid message. However, the forwarding peer will propagate such an invalid message due to the location of this check; even if it is honest, it will be banned as malicious.

This PR moves the check for the presence of an encrypted signature to the initial verification checks, so it will be caught before being propagated and ensure a ban is performed correctly. It also does some minor refactoring to avoid unnecessary computations, and improves comments.

How Has This Been Tested?
---
Existing tests pass.

### DIFF
--- a/comms/dht/src/inbound/decryption.rs
+++ b/comms/dht/src/inbound/decryption.rs
@@ -224,8 +224,10 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
         node_identity: Arc<NodeIdentity>,
         message: DhtInboundMessage,
     ) -> Result<DecryptedDhtMessage, DecryptionError> {
+        // Perform initial checks on message validity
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
Thanks LGTM - tested with the memorynet example and everything passed.

PR #5130, opened 2023-01-20
fix(dht/test): ban peers who send empty encrypted messages

### LINKED ISSUE #5132: Peers aren't banned for forwarding empty encrypted messages
Empty encrypted messages sent by peers [are not allowed](https://github.com/tari-project/tari/pull/5123). However, sending an empty message is not a [banned offence](https://github.com/tari-project/tari/blob/5f7f04b599ab49765466fb7e0faa37e2dfec4afe/comms/dht/src/inbound/decryption.rs#L190-L207).

Because all nodes can detect this, forwarding such a message should result in a ban.

AaronFeickert:
Description
---
Bans peers who send empty encrypted messages. Significantly updates tests to check for more failure modes and assert ban status for each.

Closes [issue 5132](https://github.com/tari-project/tari/issues/5132).

Motivation and Context
---
An [earlier PR](https://github.com/tari-project/tari/pull/5123) introduces an error when a peer sends an empty encrypted message, which is not allowed. However, the peer was not banned.

Further, [another PR](https://github.com/tari-project/tari/pull/5129) updates the handling of unsigned encrypted messages to ensure that bans are done correctly, but does not update tests to check for the bug that led to it.

This PR updates the banning logic to ban a peer who forwards an empty encrypted message, which is always detectable.

It also significantly refactors and updates tests. For each relevant high-level message failure mode, we test for proper error detection. We also check for the proper ban status of the forwarding peer.

How Has This Been Tested?
---
[Who tests the testers?](https://en.wikipedia.org/wiki/Quis_custodiet_ipsos_custodes%3F)

### DIFF
--- a/comms/dht/src/inbound/decryption.rs
+++ b/comms/dht/src/inbound/decryption.rs
@@ -47,7 +47,7 @@ use crate::{
 
 const LOG_TARGET: &str = "comms::middleware::decryption";
 
-#[derive(Error, Debug)]
+#[derive(Error, Debug, PartialEq)]
 enum DecryptionError {
     #[error("Failed to validate ENCRYPTED message signature")]
     MessageSignatureInvalidEncryptedSignature,
@@ -193,6 +193,7 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
             Err(err @ MessageSignatureClearTextDecodeFailed) |
             Err(err @ MessageSignatureInvalidClearTextSignature) |
             Err(err @ EncryptedMessageNoDestination) |
+            Err(err @ EncryptedMessageEmptyBody) |
             Err(err @ MessageSignatureErrorClearText(_)) => {
                 warn!(
                     target: LOG_TARGET,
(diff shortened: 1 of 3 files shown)


AaronFeickert:
Interestingly, prior to the earlier [fix PR](https://en.wikipedia.org/wiki/Quis_custodiet_ipsos_custodes%3F), the `MessageSignatureNotProvidedForEncryptedMessage` decryption error would never be raised, as it would instead be propagated as a [different error](https://github.com/tari-project/tari/blob/ff9d4e5340b0378fb1cc6c0be90ef73b90fc8424/comms/dht/src/inbound/decryption.rs#L300).

comms/dht/src/inbound/decryption.rs
```diff
@@ -790,4 +790,119 @@ mod test {
         mock_state.await_call_count(1).await;
         assert_eq!(mock_state.count_calls_containing("BanPeer").await, 1);
     }
+
+    #[runtime::test]
+    /// Ban a peer who sends an unsigned encrypted message for which we are the recipient
+    async fn decrypt_inbound_fail_missing_signature_encrypted_recipient() {
```
hansieodendaal: I think these two tests can be DRYed up, i.e. combined.

``` rust
...
let node_identity = make_node_identity();
let recipient_identity = make_node_identity();
...
for identity in [ node_identity, recipient_identity ] {
    ...
    let shared_secret = CommsDHKE::new(&e_secret_key, identity.public_key());
    ...
}
```

hansieodendaal (commented):
Nice addition. Just one comment below.

AaronFeickert:
Marking as draft for now. I'd like to expand the existing tests to include ban status.

stringhandler (approved):

PR #5134, opened 2023-01-23
fix(test): broken address test

### LINKED ISSUE #5133: Bad-network address test can fail
One [address test](https://github.com/tari-project/tari/blob/5f7f04b599ab49765466fb7e0faa37e2dfec4afe/base_layer/common_types/src/tari_address.rs#L304-L335) uses an invalid network byte and additional (possible) mutation with the expectation that the resulting address exhibit a failed checksum. However, because the network byte is baked into the checksum via an `XOR` operation, it is possible for the checksum to pass. This has the effect of failing [this error assertion](https://github.com/tari-project/tari/blob/development/base_layer/common_types/src/tari_address.rs#L324-L328).

The test should be fixed.

AaronFeickert:
Description
---
Fixes a broken [address test](https://github.com/tari-project/tari/blob/5f7f04b599ab49765466fb7e0faa37e2dfec4afe/base_layer/common_types/src/tari_address.rs#L304-L335) that exhibits inconsistent failure.

Closes [issue 5133](https://github.com/tari-project/tari/issues/5133).

Motivation and Context
---
The test in question claims to detect failures caused by bad network bytes. However, it attempts to do this by mutating the checksum, setting the second address byte to a fixed value that happens to match the checksum mutation, and checking for a checksum failure when the result is decoded using the same network used when generating the address.

It's not clear why the test is structured this way. What it's really doing is simulating the effect of two mutations, which the checksum algorithm is not guaranteed to detect (and which has caused CI failures).

This PR restructures the test to perform two separate but related tests.

In the first test, we generate a valid address on one network and attempt to decode it for another network. The checksum algorithm guarantees this will always be detected. In the second test, we mutate the checksum and attempt to decode it for the same network. The checksum algorithm guarantees this will also always be detected.

How Has This Been Tested?
---
Updated test passes. Manually ran 100000 iterations to ensure no obvious spurious failures occur.

### DIFF
--- a/base_layer/common_types/src/tari_address.rs
+++ b/base_layer/common_types/src/tari_address.rs
@@ -307,29 +307,18 @@ mod test {
         let mut rng = rand::thread_rng();
         let public_key = PublicKey::from_secret_key(&PrivateKey::random(&mut rng));
 
-        let emoji_id_from_public_key = TariAddress::from_public_key(&public_key, Network::Esmeralda);
-
-        // let create byte string with bad network u8
-        let mut buf = [0u8; INTERNAL_SIZE];
-        buf[0..32].copy_from_slice(emoji_id_from_public_key.public_key.as_bytes());
-        let checksum = compute_checksum(&buf[0..32].to_vec());
-        // 0xb3 is a bad network
-        buf[32] = 0xb3 ^ checksum;
-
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):
utACK

stringhandler (approved):

PR #5137, opened 2023-01-24
fix: feature flag separation for validation

### LINKED ISSUE #4974: Enable FFI build for aarch64-apple-ios-sim
We cannot build the FFi for `aarch64-apple-ios-sim` so we cannot use the simulator on the M1.

We need to wait until OpenSSL supports aarch64-apple-ios-sim
https://github.com/sfackler/rust-openssl/issues/1622

Or we need to try and build OpenSSL without vendor support?

SWvheerden:
Description
---
Removes OpenSSL from the WalletFFI lib as its not required or used by it. 
Fixes validation feature flags. 

Motivation and Context
---
Recent refactor added `base_node` feature flag to the walletFFI. This should not be done as it adds in all the base node 
specific crates that not required by the wallet like randomX, lmdb etc. 
This splits the validation traits and implementations between the features `transactions` and `base_node`.

How Has This Been Tested?
---
Manual and unit tests

Fixes: https://github.com/tari-project/tari/issues/4974


base_layer/wallet_ffi/Cargo.toml
```diff
@@ -27,7 +27,6 @@ libc = "0.2.65"
 log = "0.4.6"
 log4rs = {version = "1.2.0", features = ["console_appender", "file_appender", "yaml_format"]}
 # Needs to be higher than 0.10.41 to address a security issue
```
leet4tari: Best to remove the comment too?

base_layer/core/src/validation/error.rs
```diff
@@ -39,11 +42,15 @@ use crate::{
 #[derive(Debug, Error)]
 pub enum ValidationError {
     #[error("Block header validation failed: {0}")]
+    #[cfg(feature = "base_node")]
     BlockHeaderError(#[from] BlockHeaderValidationError),
```
stringhandler: There are probably a number of ways to allow this error in multiple features. 
It doesn't feel good to have enum variants different in different builds. This isn't serialized, but it could cause all sorts of errors if it were. 
I'd suggest either using a `dyn Error` or a string or anyhow or something other than the error type if that's what's causing the feature here. 

Otherwise move the error to a crate that can be referenced

base_layer/core/src/validation/error.rs
```diff
@@ -151,6 +160,7 @@ pub enum ValidationError {
 
 // ChainStorageError has a ValidationError variant, so to prevent a cyclic dependency we use a string representation in
 // for storage errors that cause validation failures.
+#[cfg(feature = "base_node")]
```
stringhandler: Also feels like someone might miss this

base_layer/core/src/validation/helpers.rs
```diff
@@ -20,15 +20,18 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#[cfg(feature = "base_node")]
 use log::*;
-use tari_crypto::tari_utilities::{epoch_time::EpochTime, hex::Hex};
+#[cfg(feature = "base_node")]
+use tari_crypto::tari_utilities::epoch_time::EpochTime;
```
stringhandler: Rather reference tari_utilities directly. Why would this need to be feature flagged?

stringhandler (commented):
Agree with the idea, but there are too many cfg's here. It's a recipe for dev mistakes and it's not clear when it needs to be applied versus not. I can see someone just adding `#cfg[base_node]` to every method.

Is there not a way we can group it all of this together under a common crate or file?

stringhandler (approved):

stringhandler:
Just waiting for CI

PR #5142, opened 2023-01-26
feat(peer_db)!: more accurate peer stats per address

stringhandler:
Description
---

Removes a lot of duplicate stats and online/offline fields on peers. 

Motivation
---
Due to a recent bug, wallets would get a new address every time they restarted. There was some handling of peers with multiple addresses, but in general only the latest address signed by the peer was used. This prevents a node from being a bridge between tor and IP or even IP and DNS. This PR controversially (I'm sure) removes the signed address and instead allows a node to try an address to see if the correct public key is on the other end. If so, the peer can choose the best latency according to communicate with it.

IMO, this is no less secure, since the worst case is that a node can force another node to contact an address that is not correct, only to find out the node at the other end has a different public key, This already can be done by generating a new peer id, using the address and signing it with the new peer id (i.e a sybil address).

Previously a peer may have been banned for answering an address with the wrong public key, but this can also be manipulated. Now, only the address is marked as bad.

Also reworked the general sorting of peer addresses, although in most previous cases, there would only be one address. 

Also reworked the order in which the peer tries neighbours. In order to make it more predictable and prevent looping, the connectivity service will try to find it's neighbours in a strict distance away from it. Previously it took into account a last_attempted_at or last_seen_at, which I am not convinced was working.

You'll note that there are some tokio improvements as well.

During the testing, I noticed that the peer query was, and still is, extremely heavy. In  tokio console, queries took in the order of 10s to 100s. Moving this to a `block_in_place` frees up the tokio worker threads. In a future PR I will improve this even further by removing unnecessary calls.

Also:
- Removed some dead code OptionalBoundedExecutor and replaced the re-export of runtime from tokio
- Used standard tokio spawns for BoundedExecutor instead of holding a runtime reference.

BREAKING CHANGE: Note that because the Peer struct has changed so dramatically, a migration is not really pragmatic. The peer_db for wallets and base nodes **must** be deleted on before starting up


comms/core/src/connection_manager/dialer.rs
```diff
@@ -320,24 +332,45 @@ where
...
                         authenticated_public_key,
                         conn_man_notifier,
                         supported_protocols,
                         &config,
                         cancel_signal,
                     )
                     .await;
-
-                    (dial_state, result)
+                    warn!(
```
sdbondi: These WARNs, here and elsewhere, are for debugging?

comms/dht/src/connectivity/mod.rs
```diff
@@ -363,6 +364,10 @@ impl DhtConnectivity {
     }
 
     async fn refresh_neighbour_pool(&mut self) -> Result<(), DhtConnectivityError> {
+        if self.is_refreshing_neighbour_pool {
```
sdbondi: `is_refreshing_neighbour_pool` will always be false here. There is no concurrency in DhtConnectivity (it runs on a single thread at a time) so this function has to exit before it can be called again.

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
+                    // excluded_count += 1;
                     return false;
                 }
 
                 true
             })
-            .sort_by(PeerQuerySortBy::DistanceFromLastConnected(node_id))
-            // Fetch double here so that there is a bigger closest peer set that can be ordered by last seen
-            .limit(n * 2);
+            // TODO: Check if this logic is correct. At the moment I want them in the same order every time
```
sdbondi: This is fine for the purposes of evaluating performance. The previous method would include a larger number of peers so that there is a better chance of selecting a peer that was recently available. This call was notably heavy and not an adequate or good strategy, a symptom of not having a good way to maintain an optimised/cached list of highly available neighbours in the peer manager (or some similar solution e.g. k-buckets).

applications/tari_app_grpc/proto/network.proto
```diff
@@ -48,7 +48,7 @@ message Peer{
     /// Features supported by the peer
     uint64 features = 9;
     /// Connection statics for the peer
-    google.protobuf.Timestamp last_connected_at = 10;    /// Protocols supported by the peer. This should not be considered a definitive list of supported protocols and is
+//    google.protobuf.Timestamp last_connected_at = 10;    /// Protocols supported by the peer. This should not be considered a definitive list of supported protocols and is
```
SWvheerden: remove?

comms/core/src/connection_manager/common.rs
```diff
@@ -93,33 +92,31 @@ pub(super) async fn validate_and_add_peer_from_peer_identity(
...
                 target: LOG_TARGET,
                 "Peer '{}' already exists in peer list. Updating.",
                 peer.node_id.short_str()
             );
-            peer.connection_stats.set_connection_success();
-            peer.addresses = addresses.into();
-            peer.set_offline(false);
+            // peer.connection_stats.set_connection_success();
+            peer.addresses.update_addresses(addresses);
+            // peer.set_offline(false);
```
SWvheerden: remove comments?
Or put these lines back?

comms/core/src/connection_manager/dialer.rs
```diff
@@ -511,87 +545,121 @@ where
...
-        loop {
-            let result = match addr_iter.next() {
-                Some(address) => {
-                    debug!(
-                        target: LOG_TARGET,
-                        "Attempting address '{}' for peer '{}'",
-                        address,
-                        dial_state.peer().node_id.short_str()
-                    );
+        for address in addresses {
```
SWvheerden, line 545: Could it not be possible to make this a parallel loop to make it execute faster? This could get quite slow if the peer has multiple old addresses.

SWvheerden: Perhaps with something like: `rayon::par_iter`

comms/core/src/connectivity/manager.rs
```diff
@@ -472,17 +480,20 @@ impl ConnectivityManagerActor {
     fn mark_peer_succeeded(&mut self, node_id: NodeId) {
         let entry = self.get_connection_stat_mut(node_id);
         entry.set_connection_success();
+        // self.peer_manager.mark_last_seen(node_id.clone(), Some(addr_used));
```
SWvheerden: remove?

comms/core/src/connectivity/manager.rs
```diff
@@ -472,17 +480,20 @@ impl ConnectivityManagerActor {
...
     }
 
     fn mark_peer_failed(&mut self, node_id: NodeId) -> usize {
         let entry = self.get_connection_stat_mut(node_id);
         entry.set_connection_failed();
+        // self.peer_manager.mark_peer_failed(node_id, addr_tried);
```
SWvheerden: remove?

comms/dht/src/peer_validator.rs
```diff
@@ -67,55 +67,55 @@ impl<'a> PeerValidator<'a> {
...
-                if let Some(sig) = new_peer.identity_signature {
-                    current_peer.set_valid_identity_signature(sig);
-                }
+                // current_peer
+                //     .update_addresses(new_peer.addresses.into_vec())
+                //     .set_features(new_peer.features);
+                // if let Some(sig) = new_peer.identity_signature {
+                //     current_peer.set_valid_identity_signature(sig);
+                // }
                 self.peer_manager.add_peer(current_peer).await?;
```
sdbondi: Noting that add_peer is changed to only append new addresses to mitigate the fact that addresses are no longer signed. It also merges stats for addresses however that does not apply here because the peer sync message cannot and should not contain this data. 

Should a remote peer F be allowed to add any amount (1000s, 10_000, ?) of addresses for any peer? If not, perhaps we should make the maximum e.g. 10. If the remote peer F provides 10 new addresses for peer A which already has 5 in our peer manager, we have no choice but to discard 5 of the addresses (at random?) provided by peer F. If peer A already has 10, we cannot add any addresses as all of them could be false. 

Comparison of happy path case:

Setup: Peer A has an updated address set, peer B is made aware of the change via an existing authenticated mechanism (DHT join, DHT discovery, direct connection). There are no malicious peers.

This PR:
Peer X later connects to peer B and _appends_ new addresses given by peer B for peer A. Peer X dials Peer A using out-of-date addresses that have a higher score first, but eventually uses one of the updated addresses which have no score after some time.

Previous with signed peer record:
Peer X later connects to peer B and _sets_ the new addresses given by peer B for peer A. Peer X can be sure that Peer B is giving the correct address set for peer A and therefore can use that address set (discarding addresses that it knows that peer A is not using). Peer B dials Peer A using the correct address set without having to first discover the previous addresses are out of date.

comms/core/src/connectivity/requester.rs
```diff
@@ -59,8 +59,7 @@ pub enum ConnectivityEvent {
     PeerConnected(PeerConnection),
     PeerConnectFailed(NodeId),
     PeerBanned(NodeId),
-    PeerOffline(NodeId),
-
+    // PeerOffline(NodeId),
```
SWvheerden: remove?

comms/core/src/connectivity/requester.rs
```diff
@@ -76,7 +75,7 @@ impl fmt::Display for ConnectivityEvent {
             PeerConnected(node_id) => write!(f, "PeerConnected({})", node_id),
             PeerConnectFailed(node_id) => write!(f, "PeerConnectFailed({})", node_id),
             PeerBanned(node_id) => write!(f, "PeerBanned({})", node_id),
-            PeerOffline(node_id) => write!(f, "PeerOffline({})", node_id),
+            // PeerOffline(node_id) => write!(f, "PeerOffline({})", node_id),
```
SWvheerden: remove?

comms/core/src/peer_manager/migrations/v7.rs
```diff
@@ -40,38 +32,40 @@ impl super::Migration<LMDBDatabase> for Migration {
...
+        //     });
+        //
+        //     if let Err(err) = result {
+        //         error!(
+        //             target: LOG_TARGET,
+        //             "Failed to deserialize peer: {} ** Database may be corrupt **", err
+        //         );
+        //     }
+        //     IterationResult::Continue
+        // })?;
```
SWvheerden, line 68: remove?

comms/core/src/peer_manager/peer.rs
```diff
@@ -76,12 +76,10 @@ pub struct Peer {
...
     pub banned_until: Option<NaiveDateTime>,
     pub banned_reason: String,
-    pub offline_at: Option<NaiveDateTime>,
-    pub last_seen: Option<NaiveDateTime>,
     /// Features supported by the peer
     pub features: PeerFeatures,
-    /// Connection statics for the peer
-    pub connection_stats: PeerConnectionStats,
+    // Connection statics for the peer
+    // pub connection_stats: PeerConnectionStats,
```
SWvheerden: remove?

comms/core/src/peer_manager/peer.rs
```diff
@@ -117,14 +118,13 @@ impl Peer {
             features,
             banned_until: None,
             banned_reason: String::new(),
-            offline_at: None,
-            last_seen: None,
-            connection_stats: Default::default(),
+            // connection_stats: Default::default(),
```
SWvheerden: remove?

comms/core/src/peer_manager/peer.rs
```diff
@@ -133,11 +133,31 @@ impl Peer {
...
     }
 
+    /// Merges the data with another peer. This is usually used to update a peer before it is saved to the
+    /// database so that data is not overwritten
+    pub fn merge(&mut self, other: &Peer) {
+        self.addresses.merge(&other.addresses);
+        self.banned_reason = other.banned_reason.clone();
+        self.added_at = cmp::min(self.added_at, other.added_at);
+        self.banned_until = cmp::max(self.banned_until, other.banned_until);
+        // self.connection_stats.merge(&other.connection_stats);
```
SWvheerden: remove?

comms/core/src/peer_manager/peer.rs
```diff
@@ -257,15 +237,15 @@ impl Peer {
...
-    }
+    // /// Marks the peer as offline if true, or not offline if false
+    // pub fn set_offline(&mut self, is_offline: bool) -> &mut Self {
+    //     if is_offline {
+    //         self.offline_at = Some(Utc::now().naive_utc());
+    //     } else {
+    //         self.offline_at = None;
+    //     }
+    //     self
+    // }
```
SWvheerden: remove?

comms/core/src/peer_manager/peer.rs
```diff
@@ -355,24 +326,19 @@ impl Display for Peer {
...
+            self.addresses,
             status_str,
             match self.features {
                 PeerFeatures::COMMUNICATION_NODE => "BASE_NODE".to_string(),
                 PeerFeatures::COMMUNICATION_CLIENT => "WALLET".to_string(),
                 f => format!("{:?}", f),
             },
             user_agent,
-            self.connection_stats,
+            // self.connection_stats,
```
SWvheerden: remove?

comms/core/src/peer_manager/peer_storage.rs
```diff
@@ -568,17 +506,17 @@ where DS: KeyValueStore<PeerId, Peer>
...
+    //     let mut peer = self
+    //         .find_by_node_id(node_id)?
+    //         .ok_or(PeerManagerError::PeerNotFoundError)?;
+    //     peer.last_seen = Some(Utc::now().naive_utc());
+    //     peer.set_offline(false);
+    //     self.peer_db
+    //         .insert(peer.id(), peer)
+    //         .map_err(PeerManagerError::DatabaseError)?;
+    //     Ok(())
+    // }
```
SWvheerden: remove?

comms/core/src/transports/socks.rs
```diff
@@ -73,7 +73,8 @@ impl SocksTransport {
 
     pub fn create_socks_tcp_transport() -> TcpTransport {
         let mut tcp_transport = TcpTransport::new();
-        tcp_transport.set_nodelay(true);
+        // TODO: Confirm, but probably don't need nodelay
+        // tcp_transport.set_nodelay(true);
```
SWvheerden, line 77: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -580,32 +589,33 @@ impl DhtConnectivity {
...
+                //     .cooldown_in_effect
+                //     .map(|ts| ts.elapsed() >= self.config.connectivity.high_failure_rate_cooldown)
+                //     .unwrap_or(true)
+                // {
+                //     if self.cooldown_in_effect.is_some() {
+                //         self.cooldown_in_effect = None;
+                //         self.recent_connection_failure_count = 1;
+                //     }
+                //     self.replace_pool_peer(&node_id).await?;
+                // }
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
-        let mut connect_ineligable_count = 0;
-        let mut banned_count = 0;
-        let mut excluded_count = 0;
-        let mut filtered_out_node_count = 0;
-        let mut already_connected = 0;
+        // let mut connect_ineligable_count = 0;
+        // let mut banned_count = 0;
+        // let mut excluded_count = 0;
+        // let mut filtered_out_node_count = 0;
+        // let mut already_connected = 0;
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
+        let offline_cooldown = self.config.offline_peer_cooldown;
         let query = PeerQuery::new()
             .select_where(|peer| {
                 if peer.is_banned() {
-                    banned_count += 1;
+                    // banned_count += 1;
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
                     return false;
                 }
 
                 if peer.features.is_client() {
-                    filtered_out_node_count += 1;
+                    // filtered_out_node_count += 1;
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
                 {
-                    connect_ineligable_count += 1;
+                    // connect_ineligable_count += 1;
                     return false;
                 }
 
                 let is_excluded = excluded.contains(&peer.node_id);
                 if is_excluded {
-                    excluded_count += 1;
+                    // excluded_count += 1;
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
                 }
 
                 if peer
                     .offline_since()
-                    .map(|since| since <= self.config.offline_peer_cooldown)
+                    .map(|since| since <= offline_cooldown)
                     .unwrap_or(false)
                 {
-                    connect_ineligable_count += 1;
+                    // connect_ineligable_count += 1;
```
SWvheerden: remove?

comms/dht/src/connectivity/mod.rs
```diff
@@ -786,68 +796,52 @@ impl DhtConnectivity {
...
 
                 if peer.features.is_client() {
-                    filtered_out_node_count += 1;
+                    // filtered_out_node_count += 1;
                     return false;
                 }
 
                 if connected.contains(&&peer.node_id) {
-                    already_connected += 1;
+                    // already_connected += 1;
```
SWvheerden: remove?

comms/dht/src/peer_validator.rs
```diff
@@ -67,55 +67,55 @@ impl<'a> PeerValidator<'a> {
...
+                //         .unwrap_or(true)
+                // };
+
+                // if !can_update {
+                //     debug!(
+                //         target: LOG_TARGET,
+                //         "Peer `{}` already exists or is up to date and will not be updated", new_peer.node_id
+                //     );
+                //     return Ok(false);
+                // }
```
SWvheerden: remove?

comms/dht/src/peer_validator.rs
```diff
@@ -67,55 +67,55 @@ impl<'a> PeerValidator<'a> {
...
-                    .update_addresses(new_peer.addresses.into_vec())
-                    .set_features(new_peer.features)
-                    .set_offline(false);
-                if let Some(sig) = new_peer.identity_signature {
-                    current_peer.set_valid_identity_signature(sig);
-                }
+                // current_peer
+                //     .update_addresses(new_peer.addresses.into_vec())
+                //     .set_features(new_peer.features);
+                // if let Some(sig) = new_peer.identity_signature {
```
SWvheerden: I dont think we should remove this

SWvheerden (commented):
This looks good, brings much-needed address functionality. 
I tried to mark all commented out code there is. 

I do however thing we should not remove address singing. I think each address needs to be signed. The reason for this is, it will allow a peer to poison the address of some other peer by spamming unsigned addresses to the network obfuscating the peer's true address and make it harder to connect to it. 
If we force peers to sign each address, then peers can only poison their own addresses and not those of other peers.

comms/core/src/connection_manager/dialer.rs
stringhandler: Yeah, I did consider that. In this case tokio would be better than rayon, because it would wait on the network dial. Rayon would be a better choice if it were CPU bound. The dials already, for a few cases, are parallelized during a call to each peer, but you are correct in saying that if a peer had many addresses that it would take a while.

Some things to note though:
1. The addresses for a peer are always sorted by their quality score, so the best addresses will be at the front.
2. Addresses that have failed previously are moved to the back of the list, so will be tried after all others.

So from these, it makes sense to keep it trying addresses in synch and to exit on the first dialled address

comms/dht/src/peer_validator.rs
stringhandler: See the notes in the motivation about this

sdbondi (commented):
The goals of this PR are to:
1. improve currently incomplete support for multiple peer addresses (perhaps to allow support later for WebRTC addresses)
   - This has been greatly improved in the PR. Although more work needs to happen, I like the addition of an address score and the improvements in the collection of address stats.
2. improve the performance and selection strategy for maintaining connections to neighbouring peers
   - As far as I can see, there are no improvements made in this area but are TODO for another PR. 
   - IMO one way to improve this is by maintaining a list of neighbouring and available peers in the peer manager DB. This list can be used and maintained by the DhtConnectivity component.
3.  Improve the propagation of updated address information across the network
    - This may have been made worse by removing the ability to validate peer sync information.
    - Was there a confirmed bug that caused wallet addresses not to eventually update over the network over time?

Other non-trivial changes:
1. Removal of PeerIdentitySignature

> Previously a peer may have been banned for answering an address with the wrong public key, but this can also be manipulated. 

Is this not only true because the peer no longer signs its addresses?

> since the worst case is that a node can force another node to contact an address that is not correct

This removes the invariant that each assigned address was in fact provided by the peer, allowing another node to provide arbitrary addresses for a peer on their behalf.

As you said, a peer may provide _other_ self-generated identities. I'd argue this is a distinct case from allowing a peer to modify another peer's address set, even if that is only to add addresses to the set.

Perhaps the worst case is wasting network resources (memory, time, disk, bandwidth), perhaps there are other attacks/abuses not thought of (is a coordinated DDoS possible?). 

Could you provide more details on the specific reasons for removing this invariant? 

I took a look into what libp2p does
https://github.com/libp2p/rust-libp2p/blob/d344406235c779922e6ffec85f0a94181f3efecc/core/src/peer_record.rs#L62

>, but in general only the latest address signed by the peer was used

Minor correction, the address set is signed by the peer not individual addresses. The handling of multiple addresses was primitive and even incorrect, this PR greatly improves this. This improvement is not precluded by the removal of signed address sets.

comms/core/src/connection_manager/dialer.rs
stringhandler: Yeah, wanted some feedback first, but will remove them in the final PR

comms/dht/src/connectivity/mod.rs
stringhandler: Cool, I will remove this. It's easy to see in tokio::console whether this is in fact the case, but I believe you are correct and will confirm

comms/dht/src/connectivity/mod.rs
stringhandler: My problem with the previous approach was that it could end up in a stale loop, if the peers came online again before new peers were tried. This approach is more predictable and ensures that all the peers are tried. It may be a symptom of the current network state, but the majority of these nodes are actually offline. 

Given that the offline data is more accurate now, it may be much of a muchness. I think I would however lobby for selecting the same n (instead of n*2) peers from the DB and returning them directly.

stringhandler:
I see there is a lot of contention about removing the address signature. Probably not unfounded. I will look into adding the signed address back, but the structure will need to change. Currently, it is difficult to reassemble the challenge when multiple addresses are signed, so I will change the identity message to sign each address in a way that makes it easy to re-verify and store with each address and see how that looks

comms/dht/src/connectivity/mod.rs
sdbondi: `n == 8` by default. 
The goal was to fetch peers that we recently connected to, ordered by distance. When a peer comes back online, it will send a join message, which will be propagated to neighbours, and/or directly connect. This will update the last_seen which will put them back on the top of the list for dialing if ordered by last seen. 

This change will select the same 8 peers (unless a nearer one joins) which may never come online again. Currently "culling" of offline peers is a slow process, I think we have to attempt and fail to connect to them at least once and then we have to have not connected to them for 24 hours before we remove them from the peer list entirely. 

I think a good solution would be to maintain a list that can be loaded that has available peers. Essentially just optimising this peer query from an LMDB perspective.

stringhandler: In this PR, as soon as an address fails it is immediately marked offline. If all addresses are offline, the peer is offline

hansieodendaal:
This PR touches many things and is a bit big to comprehend what the actual system-wide effects would be without proper stress testing.

I agree that it is an improvement to mark an address as bad and not ban the peer straight away, although I think that eventually, a peer should be medium-term banned if none of its addresses is working, however, I do not agree that addresses should not be signed anymore.

Maybe we can implement just one thing from this PR at a time and evaluate it, leaving the changes with the biggest system impact for last?

comms/core/src/peer_manager/migrations/v7.rs
stringhandler: "// Kept here as an example..."

stringhandler (commented):
self reviewed and removed all commented code

sdbondi (approved):

hnidoaht-101:
hi team, I know this PR was already merged. Just curious how @stringhandler used `tokio-console` to troubleshoot the long-running queries. Is it possible to do a demo for it?

stringhandler:
Sure, will try put something together

hnidoaht-101:
> Sure, will try put something together

Thanks a lot. 💯 I'll be waiting for it.

hnidoaht-101:
> Sure, will try put something together

@stringhandler Can you just tag my name once you have something? Thank you.

stringhandler:
will do

PR #5144, opened 2023-01-27
feat: remove unused dependencies

### LINKED ISSUE #5136: Clean up cargo toml files
Ensure all cargo toml files only have the required libraries in them.

SWvheerden:
Description
---
Removed all unused dependencies. 
Includes `[package.metadata.cargo-machete]` for false positives

How Has This Been Tested?
---
Manual, unit tests

Fixes: https://github.com/tari-project/tari/issues/5136


base_layer/key_manager/Cargo.toml
```diff
@@ -42,5 +39,5 @@ wasm-bindgen-test = "0.3.28"
 
 [features]
 avx2 = ["tari_crypto/simd_backend"]
-js = ["getrandom/js", "js-sys"]
```
sdbondi, line 45: Might be needed to provide random to JS. Guessing machete might not take this into account

CjS77: The wasm tests _should_ fail if this is the case

SWvheerden: I manually tested this with 'cargo build - - all-features' if the wasm build with this then it is it should be fine. I picked up 2 issues with this

delta1: cargo test isn't even compiling on this branch. 

anyway, the `js` feature is needed by the `wasm` feature right beneath it. 

test locally in `base_layer/key_manager/` by running `make test` 

https://github.com/tari-project/tari/blob/development/base_layer/key_manager/Makefile#L4

SWvheerden: The Js feature is not removed, its still there. Like you say its used by the `wasm` feature branch. 
This PR only removes `"getrandom/js"`

SWvheerden: If you look at the CI, the key manager wasm tests all pass.

SWvheerden:
Its weird that the cargo test compiles and runs fine locally not on CI
Means we have some dependency that is not properly defined

CjS77:
I'm happy
👍

SWvheerden:
It all green now,
Just added in the last ignore for machete. 
This PR will still complain about `openssl` in the wallet_ff cargo. But it will get removed here: #5137

stringhandler (commented):
Happy to merge this PR, but I think it's worth adding a few CI jobs in this PR as well.

1. Add a cargo machete job 
2. Confirm the wasm build is being done

CjS77:
1 can def be a separate PR
2 can be a separate PR too, even if the wasm build is broken tbh

CjS77 (approved):
LGTM

Issue #5135, opened 2023-01-23
Feature gating


brianp:
# Problem
Once mainnet is live we want to ensure the safe development and testing of features occur on the appropriate networks, and that we prevent the use of these features on the wrong network.

# Solution 1

## Define features

In build.rs, we define our features.

Features have a 
* name - will also be the attribute name you use in code
* description - a short description of the features
* issue tracking number on github
* The current status.

#### Status

* New: New feature, may not even be working or compiling. Will be present on dibbler testnet
* Testing: Potentially active feature, but still gated. Potentially present on nextnet,
* Active: Feature is live and gate has been removed. Will be running on stagenet and mainnet
* Removed: Feature has been cancelled. Can not be invoked anywhere

In `build.rs`, we maintain the list of feature flags. For example:

```rust
const FEATURE_LIST: [Feature; 4] = [
    Feature::new("add_pair", "Allows you to add two numbers together", Some(123), Status::New),
    Feature::new("about_to_launch", "Live in NextNet. If we stabilise, will go to mainnet", Some(123), Status::Testing),
    Feature::new("live_feature", "This feature has been stabilised and is live!", Some(150), Status::Active),
    Feature::new("old_feature", "We decided not to go with this featuree", Some(145), Status::Removed),
];
```

## Demarcate feature flag code

In your code, you can now use any of the flags as attributes to mark code belonging to a feature.

Example

```rust
#[tari_feature(add_pair)]
use add_pair::add_pair;

fn main() {
    println!("Hello world!");
    #[tari_feature(add_pair)]
    println!("40 + 2 = {}", add_pair(40, 2));
    println!("foo={}", foo());
    println!("Bye, world!");
}

#[tari_feature(add_pair)]
fn foo() -> usize {
    1
}

#[tari_feature(!add_pair)]
fn foo() -> usize {
    2
}
```

## Specify the target network when building

This PoC uses the `TARI_NETWORK` envar to specify the target network, but in principle, we can also read the `Cargo.toml` 
file, or compiler flags.

`$ TARI_NETWORK=dibbler cargo run -vv`

Filtered output:

```text
Building for Dibbler
These features are ACTIVE on mainnet (no special code handling is done)
live_feature. This feature has been stabilised and is live!. Tracking issue: https://github.com/tari-project/tari/issues/150

These features are DEPRECATED and will never be compiled
old_feature. We decided not to go with this featuree. Tracking issue: https://github.com/tari-project/tari/issues/145

[...]

PR #5147, opened 2023-01-27
feat: feature gate

brianp:
Description
---
This PR is for reflective purposes regarding feature gating at compile time. See #5135 for more info. This PR as it stands is to be used as an example of a network _type_ based feature gate at compile time. 
It allows setting an ENVVAR for the network with conditional compilation for features. The feature sets can be imported across any crate in the project easily and then used with the `#[cfg(tari_feature_...)]` attribute macro. 

I played with a few styles of gating code. In the case of burning, and validator registration what I found was it became easier to focus on the entry and exit points of the code. These functions don't have much outside effect other than their distinct purpose so preventing someone from calling them is good enough conditional compilation. It's easier than trying to remove every aspect of their existence throughout. This wouldn't always be the case though- with a feature that changes an existing feature the style of code writing has the potential to change heavily with possible duplication of entire enums or structs. It could be useful to have more examples to help guide people with, but also is something we'll probably see patterns for emerge naturally. The nice part is we have the flexibility to support different needs.

There was a desire to use the network based feature gating to prevent a "mainnet" compiled bin from compiling with testnet configuration at all. To remove conensus constants, or even the knowledge of the testnets. This proved more difficult with the current code base and would require a lot of refactoring. I've made another PR with a branch I consider broken demonstrating the minimal amount of changes needed to perform this kind of task and it doesn't seem to bring us a major benefit for the time being.

Motivation and Context
---
Preventing the use of certain network features that may not yet be prepared.

How Has This Been Tested?
---
Compilation locally.

Co-authored-by: Cayle Sharrock <CjS77@users.noreply.github.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5049,6 +5049,7 @@ dependencies = [
  "tari_comms_dht",
  "tari_core",
  "tari_crypto",
+ "tari_features",
  "tari_libtor",
  "tari_metrics",
  "tari_p2p",
@@ -5299,6 +5300,7 @@ dependencies = [
  "tari_comms_dht",
  "tari_core",
  "tari_crypto",
+ "tari_features",
  "tari_key_manager",
  "tari_libtor",
  "tari_p2p",
@@ -5408,6 +5410,13 @@ dependencies = [
  "thiserror",
  "wasm-bindgen",
  "zeroize",
+]
+
+[[package]]
+name = "tari_features"
+version = "0.0.1"
+dependencies = [
+ "tari_common",
 ]
 
 [[package]]
@@ -5522,6 +5531,7 @@ dependencies = [
  "tari_common",
  "tari_comms",
  "tari_core",
+ "tari_features",
  "tari_utilities",
  "tari_wallet_grpc_client",
  "thiserror",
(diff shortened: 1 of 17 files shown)


CjS77:
Put in an attribution to the original repo and author would ya? ;)

brianp:
@CjS77 my bad, I meant no offense 😄 

I've made some updates I'd like to bring to your attention. I spent some time checking out and having discussions in community chats and it really looks like most people stick to `cfg` usage when feature-gating non-runtime changes. It just has so much extra built-in support and flexibility. I didn't find any additional tooling (crates) that helped out in any useful way either. As a result, the major change is the removal of the macro. No proc macro crate is needed. We still manage the features, and the builds exactly the same as the original proof of concept, and when gating code we use the standard `#[cfg(tari_feature_coolness)]` attribute.

I played with a few styles of gating code. In the case of burning, and validator registration what I found was it became easier to focus on the entry and exit points of the code. These functions don't have much outside effect other than their distinct purpose so preventing someone from calling them is good enough conditional compilation. It's easier than trying to remove every aspect of their existence throughout. This wouldn't always be the case though- with a feature that changes an existing feature the style of code writing has the potential to change heavily with possible duplication of entire enums or structs. It could be useful to have more examples to help guide people with, but also is something we'll probably see patterns for emerge naturally. The nice part is we have the flexibility to support different needs.

I've created a feature branch to merge these changes into. As I don't think they should hit development until the CI work #5165 is ready and merged so we can avoid creating a release problem for ourselves.

CjS77:
> I meant no offense 😄

None taken. I genuinely mean it. If the feature doesn't add value in terms of safety and DevX, then we can chuck it.

PR #5153, opened 2023-01-31
test: add coverage to mmr

stringhandler:
Description
---
Cover some uncovered lines in the mmr crate

EDIT: 

Motivation
---
I'm not particularly concerned with getting this merged. I was investigating the panic in the MMR crate generated found here: https://github.com/tari-project/tari-dan/pull/295
My approach was to go through the coverage of the MMR crate at https://coveralls.io/github/tari-project/tari and add tests for lines that were missing. I've since realised that the MerkleProofs have not been used much outside of unit tests and calling `find_peaks` with certain inputs is probably the cause. Fixing this requires a bit more work, so I'm pushing the tests I created to increase the coverage so long.


CjS77:
I'm not a fan of putting `test_*` in test functions usually. And they are superfluous changes.

stringhandler:
I find naming them `test_` is extremely useful when using `Find all references` etc. but not tied to it. I'll remove them and push

stringhandler:
@CjS77 updated without the `test_` prefix

CjS77 (approved):

Issue #5151, opened 2023-01-30
Add versioning to PBKDF applications


AaronFeickert:
The codebase uses `Argon2` as a password-based key derivation function (PBKDF). Specifically, it uses [OWASP recommendations](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) for its parameters, which are updated from time to time in order to reflect evolving threats from more advanced hardware. In fact, these parameters were [recently updated](https://github.com/tari-project/tari/pull/5140).

Based on the current implementation, some future codebase updates to reflect parameter changes will be brittle, as there is currently no straightforward way to detect which parameters were used to generate any particular keys in applications not using the [PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md). There should always be backwards compatibility for keys generated using older parameters, but users should (to the extent possible) always receive the safety and security benefits of the most recent parameters in use.

The best and most standard approach is to use versioning wherever a PBKDF is used, where a hardcoded version is stored or otherwise encoded such that the client uses the correct parameters for key derivation. It may important to bind this version to authenticated data in order to mitigate risks associated with version mismatches or downgrade attacks.

AaronFeickert:
The codebase uses `Argon2` in the following places:
- for [gRPC authentication](https://github.com/tari-project/tari/blob/development/applications/tari_app_grpc/src/authentication/salted_password.rs#L31-L55) using the standard [PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md)
- for [encrypted seeds](https://github.com/tari-project/tari/blob/4c4a056f1f6623f6566b691a96c850ff905c0587/base_layer/key_manager/src/cipher_seed.rs#L359-L398), which include a [hardcoded version number](https://github.com/tari-project/tari/blob/development/base_layer/key_manager/src/cipher_seed.rs#L56)
- for [wallet storage](https://github.com/tari-project/tari/blob/4c4a056f1f6623f6566b691a96c850ff905c0587/base_layer/wallet/src/storage/sqlite_db/wallet.rs#L473-L568)

AaronFeickert:
Relating to [this issue](https://github.com/tari-project/tari/issues/5003), we can simplify database encryption while supporting straightforward passphrase changes and upgrades.

[...]

PR #5154, opened 2023-01-31
feat!: refactor database encryption

AaronFeickert:
Description
---
Refactors database encryption to enable easier [password changes](https://github.com/tari-project/tari/issues/5003) and [PBKDF parameter updates](https://github.com/tari-project/tari/issues/5151).

BREAKING CHANGE: This significantly changes how database encryption is performed, so existing databases will be rendered permanently inaccessible.

Motivation and Context
---
Currently, the user's passphrase is used with a PBKDF (as of now, `Argon2` with [recommended parameters](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id)) to derive the `XChaCha20-Poly1305` key used for database encryption. While this is a safe way to handle database encryption (up to the complexity of the passphrase), it is brittle. Changing the passphrase or updating the PBKDF parameters would require that all database entries be re-encrypted with a new key, which is likely to be tedious and prone to corruption.

This can be [improved](https://github.com/tari-project/tari/issues/5151#issuecomment-1409433614) by refactoring how encryption is done.

When the user wishes to set up encryption, we generate a high-entropy random `XChaCha20-Poly1305` key (the _main key_) to do the actual authenticated encryption of database entries. We then run the user's passphrase and a random salt through the PBKDF to derive another `XChaCha20-Poly1305` key (the _secondary key_), and encrypt the main key with authentication. Finally, we store the secondary key salt, the encrypted main key, and a version identifier in the database.

When we need to decrypt database entries, we derive the secondary key using the salt (and parameters from the version identifier), decrypt and authenticate the main key, and use the main key for database operations.

When the user decides to change their passphrase, or if we wish to update the PBKDF parameters, we simply derive a new secondary key, re-encrypt the main key with it, and store the new secondary key salt, encrypted main key, and version identifier. This operation can be done quickly and with minimal risk.

This PR includes the refactoring needed to support this two-key design. However, it does not directly integrate password change or PBKDF parameter update functionality, which is deferred to future work.

How Has This Been Tested?
---
Existing tests pass. Manual testing is still needed.

### DIFF
--- a/base_layer/wallet/src/error.rs
+++ b/base_layer/wallet/src/error.rs
@@ -173,6 +173,8 @@ pub enum WalletStorageError {
     KeyManagerError(#[from] KeyManagerError),
     #[error("Recovery Seed Error: {0}")]
     RecoverySeedError(String),
+    #[error("Bad encryption version")]
+    BadEncryptionVersion,
 }
 
 impl From<WalletStorageError> for ExitError {
--- a/base_layer/wallet/src/error.rs
+++ b/base_layer/wallet/src/error.rs
@@ -173,8 +173,8 @@ pub enum WalletStorageError {
     KeyManagerError(#[from] KeyManagerError),
     #[error("Recovery Seed Error: {0}")]
     RecoverySeedError(String),
-    #[error("Bad encryption version")]
-    BadEncryptionVersion,
+    #[error("Bad encryption version: `{0}`")]
+    BadEncryptionVersion(String),
 }
 
 impl From<WalletStorageError> for ExitError {
--- a/base_layer/wallet/src/storage/sqlite_db/wallet.rs
+++ b/base_layer/wallet/src/storage/sqlite_db/wallet.rs
@@ -27,11 +27,11 @@ use std::{
     sync::{Arc, RwLock},
 };
 
-use argon2::{
-    password_hash::{rand_core::OsRng, Decimal, PasswordHash, PasswordHasher, PasswordVerifier, SaltString},
-    Argon2,
+use argon2::password_hash::{
+    rand_core::{OsRng, RngCore},
+    SaltString,
 };
-use chacha20poly1305::{Key, KeyInit, Tag, XChaCha20Poly1305, XNonce};
+use chacha20poly1305::{Key, KeyInit, XChaCha20Poly1305};
 use diesel::{prelude::*, SqliteConnection};
 use log::*;
 use tari_common_types::chain_metadata::ChainMetadata;
@@ -43,6 +43,7 @@ use tari_comms::{
 use tari_key_manager::cipher_seed::CipherSeed;
 use tari_utilities::{
     hex::{from_hex, Hex},
+    hidden_type,
     safe_array::SafeArray,
     Hidden,
     SafePassword,
@@ -63,6 +64,13 @@ use crate::{
 };
 
 const LOG_TARGET: &str = "wallet::storage::wallet";
+
+// The main `XChaCha20-Poly1305` key used for database encryption
+// This isn't a `SafeArray` because of how we populate it from an authenticated decryption
+hidden_type!(WalletMainEncryptionKey, Vec<u8>);
+
+// The secondary `XChaCha20-Poly1305` key used to encrypt the main key
+hidden_type!(WalletSecondaryEncryptionKey, SafeArray<u8, { size_of::<Key>() }>);
 
 /// A Sqlite backend for the Output Manager Service. The Backend is accessed via a connection pool to the Sqlite file.
 #[derive(Clone)]
@@ -292,8 +300,9 @@ impl WalletSqliteDatabase {
             DbKey::CommsFeatures |
             DbKey::CommsAddress |
             DbKey::BaseNodeChainMetadata |
-            DbKey::PassphraseHash |
-            DbKey::EncryptionSalt |
+            DbKey::EncryptedMainKey |
+            DbKey::SecondaryKeySalt |
+            DbKey::SecondaryKeyVersion |
             DbKey::WalletBirthday |
             DbKey::CommsIdentitySignature => {
                 return Err(WalletStorageError::OperationNotSupported);
@@ -337,8 +346,9 @@ impl WalletBackend for WalletSqliteDatabase {
             DbKey::TorId => self.get_tor_id(&conn)?,
             DbKey::CommsFeatures => self.get_comms_features(&conn)?.map(DbValue::CommsFeatures),
             DbKey::BaseNodeChainMetadata => self.get_chain_metadata(&conn)?.map(DbValue::BaseNodeChainMetadata),
-            DbKey::PassphraseHash => WalletSettingSql::get(key, &conn)?.map(DbValue::PassphraseHash),
-            DbKey::EncryptionSalt => WalletSettingSql::get(key, &conn)?.map(DbValue::EncryptionSalt),
+            DbKey::EncryptedMainKey => WalletSettingSql::get(key, &conn)?.map(DbValue::EncryptedMainKey),
+            DbKey::SecondaryKeySalt => WalletSettingSql::get(key, &conn)?.map(DbValue::SecondaryKeySalt),
+            DbKey::SecondaryKeyVersion => WalletSettingSql::get(key, &conn)?.map(DbValue::SecondaryKeyVersion),
             DbKey::WalletBirthday => WalletSettingSql::get(key, &conn)?.map(DbValue::WalletBirthday),
             DbKey::CommsIdentitySignature => WalletSettingSql::get(key, &conn)?
                 .and_then(|s| from_hex(&s).ok())
@@ -402,11 +412,7 @@ impl WalletBackend for WalletSqliteDatabase {
     }
 }
 
-/// Confirm if database is encrypted or not and if a cipher is provided confirm the cipher is correct.
-/// Unencrypted the database should contain a MasterSecretKey and associated MasterPublicKey
-/// Encrypted the data should contain a Master Public Key in the clear and an encrypted MasterSecretKey
-/// To confirm if the provided Cipher is correct we decrypt the Master PrivateSecretKey and see if it produces the same
-/// Master Public Key that is stored in the db
+/// If the database is encrypted, produce a cipher that can be used for this purpose
 #[allow(clippy::too_many_lines)]
 fn get_db_encryption(
     database_connection: &WalletDbConnection,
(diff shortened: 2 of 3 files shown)


base_layer/wallet/src/error.rs
```diff
@@ -173,6 +173,8 @@ pub enum WalletStorageError {
     KeyManagerError(#[from] KeyManagerError),
     #[error("Recovery Seed Error: {0}")]
     RecoverySeedError(String),
+    #[error("Bad encryption version")]
+    BadEncryptionVersion,
```
hansieodendaal: We could maybe include a string here to let the user know what the expected and actual versions are inside the error.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -472,99 +451,108 @@ fn get_db_encryption(
...
-                    "Cipher is provided but there is no Master Secret Key in DB to decrypt"
-                );
-                return Err(WalletStorageError::InvalidEncryptionCipher);
+        // Encryption has already been set up
+        (Some(secondary_key_version), Some(secondary_key_salt), Some(encrypted_main_key)) => {
+            // For now, only one encryption version is supported
+            let secondary_key_version_u8 =
+                u8::from_str(&secondary_key_version).map_err(|_| WalletStorageError::BadEncryptionVersion)?;
+            if secondary_key_version_u8 != pbkdf_params_version {
+                return Err(WalletStorageError::BadEncryptionVersion);
```
hansieodendaal: Suggest include `secondary_key_version_u8` and `pbkdf_params_version` in `WalletStorageError::BadEncryptionVersion("...")`

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -472,99 +451,108 @@ fn get_db_encryption(
...
+    let pbkdf_params_version = 1u8;
+    let pbkdf_algorithm = argon2::Algorithm::Argon2id;
+    let pbkdf_version = argon2::Version::V0x13;
+    let pbkdf_params = argon2::Params::new(
+        46 * 1024,
+        1,
+        1,
+        Some(size_of::<Key>()), // for `XChaCha20-Poly1305`
+    )
+    .map_err(|e| WalletStorageError::AeadError(e.to_string()))?;
```
hansieodendaal: Maybe these can go in a separate vector containing `pbkdf` struct items, similar to how `ConsensusConstants` is handled.

hansieodendaal (commented):
Nice. Please see some comments below.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -64,6 +65,13 @@ use crate::{
 
 const LOG_TARGET: &str = "wallet::storage::wallet";
 
+// The main `XChaCha20-Poly1305` key used for database encryption
+// This isn't a `SafeArray` because of how we populate it from an authenticated decryption
```
jorgeantonio21, line 69: Possibly add a brief explanation why we this is sufficiently secure

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -472,99 +451,108 @@ fn get_db_encryption(
...
+            // Generate a high-entropy main key
+            let mut main_key = WalletMainEncryptionKey::from(vec![0u8; size_of::<Key>()]);
+            let mut rng = OsRng;
+            rng.fill_bytes(main_key.reveal_mut());
+
+            // We'll be encrypting the main key shortly, so keep a clone around
+            let main_key_clone = main_key.clone();
+
+            // Derive the secondary key from the user's passphrase and a high-entropy salt
+            let secondary_key_salt = SaltString::generate(&mut rng);
```
jorgeantonio21, line 543: Very nitty, but should we use a different source of randomness ?

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Yeah, there will eventually need to be a more unified way to handle multiple versions anyway, since the client will need to be able to gracefully handle old versions and update to new ones. We can then include a version query that either returns the correct parameters, or raises a `BadEncryptionVersion` error.

AaronFeickert: Worth noting that an even more general approach may eventually be needed, since there's no guarantee that all future versions will use `Argon2` as the PBKDF. But for now I think restricting to `Argon2` parameters should be fine.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -472,99 +451,108 @@ fn get_db_encryption(
...
-            )
-            .map_err(|e| WalletStorageError::AeadError(e.to_string()))?;
-
-            // Hash the passphrase to a PHC string for later verification
-            let passphrase_salt = SaltString::generate(&mut OsRng);
-            let passphrase_hash = argon2::Argon2::default()
-                .hash_password_customized(
+    // Set up the PBKDF used for secondary key derivation
+    // These parameters will likely change in the future, so we version them
+    // https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id
```
jorgeantonio21: Changing these parameters in production, might be extremely sensitive, as it will impact all wallet data encryption, across the network. On which basis, apart from possibly security breaches on the protocols, will these parameters change in the future ?

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -472,99 +451,108 @@ fn get_db_encryption(
...
-            encryption_salt.to_string()
+                .map_err(|e| WalletStorageError::AeadError(e.to_string()))?;
+
+            // Use the secondary key to encrypt the main key, authenticating with the version to mitigate mismatch
+            // attacks
+            let main_key_cipher = XChaCha20Poly1305::new(Key::from_slice(secondary_key.reveal()));
+            let mut aad = main_key_domain;
+            aad.push(pbkdf_params_version);
+            let encrypted_main_key =
+                encrypt_bytes_integral_nonce(&main_key_cipher, aad, Hidden::hide(main_key.reveal().clone()))
```
jorgeantonio21, line 559: Encryption is made in-place on `main_key`, right ? If so, we would not need a `Hidden::hide` to hide its contents away.

jorgeantonio21 (commented):
Really nice work ! Seems to be on the right track. I left a few minor comments.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
jorgeantonio21: In one of my comments, I raise precisely this question. How can we minimize the impact on the user experience, given any need to update PBKDF parameters (including, version).

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Can do. May also be worth it to update `SafeArray` for a more standard constructor too, so we can use it for cases like this.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Why? The randomizer provides high-entropy data each time.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Sources like [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#upgrading-the-work-factor) update their parameter recommendations over time to account for things like hardware advances. Fortunately, we can account for this by always supporting decryption of older versions, and by seamlessly updating by re-encrypting the main key.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: No, `encrypt_bytes_integral_nonce` returns a separate vector, so we should try our best to keep the unencrypted main key hidden.

AaronFeickert: I considered writing a separate in-place function in the hopes that it would play nicely with `SafeArray`, but it proved to be more of a hassle.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Updates can be done seamlessly on startup, with no user interaction:
- The user enters their passphrase on startup as usual.
- We derive the secondary key using the out-of-date version parameters.
- The main key is authenticated and decrypted using the secondary key.
- We derive a new secondary key using the new version parameters (and the same passphrase).
- The main key is re-encrypted using the new secondary key.
- The new encrypted main key, secondary key salt, and version identifier are stored in the database.
- Life goes on!

hansieodendaal (approved):
utACK

stringhandler (approved):
ACK - tested locally.  

When using it on an existing db, it just came back as wrong password. It's fine for now, since this will be a breaking change and there's no data at stake, but in future, we'll want to distinguish between version mismatches and incorrect passwords. Not sure if that is catered for here.

AaronFeickert:
> ACK - tested locally.
> 
> When using it on an existing db, it just came back as wrong password. It's fine for now, since this will be a breaking change and there's no data at stake, but in future, we'll want to distinguish between version mismatches and incorrect passwords. Not sure if that is catered for here.

It is accounted for to the extent possible. We store the parameter version in the clear, and a good future parameter update design will detect and support older versions (and perhaps even update automatically to the newest version), so "this version isn't supported" should never be a problem unless the user were to downgrade their client to one that doesn't support a newer parameter set used for their database.

However, malleation of the stored parameter version to a supported but incorrect value will result in a generic "bad passphrase" error. The key derivation function is such that we can't detect a bad key resulting from a bad passphrase from one resulting from different derivation parameters. (If we could, this would probably violate the guarantees you'd want from such a key.)

AaronFeickert:
It was noted elsewhere that this design doesn't protect against the following scenario:
- The user chooses a weak passphrase for database encryption.
- An attacker obtains the database and is able to decrypt it using brute force.
- The user changes their passphrase to a better one.
- The attacker obtains the new database and is able to decrypt it using the same main key.

If this threat model is of concern, it is possible to include functionality that chooses a new main key and re-encrypts the database with it. Like password change functionality, this is supported by the new design but not implemented here. It may be useful to present this as an option to the user. It is the case that persistent access to the encrypted database is a different threat model than one-time access.

PR #5157, opened 2023-02-01
ci(fix): add concurrency build limit

leet4tari:
Description
GHA add concurrency build limit

Motivation and Context
Reduce concurrent builds, so that new branch pushes cancel previous running builds.

How Has This Been Tested?
Tested in local fork with multi pushes.


stringhandler (approved):
happy to try it

PR #5169, opened 2023-02-07
feat: add more burn details to burn command

stringhandler:
Description
---
Add burn fields to the GRPC method of burning

Motivation and Context
---
When claiming funds on the second layer, you will need a few proofs to claim them. This allows the wallet to provide them

How Has This Been Tested?
---
CI

### DIFF
--- a/base_layer/wallet/src/transaction_service/service.rs
+++ b/base_layer/wallet/src/transaction_service/service.rs
@@ -1425,7 +1425,7 @@ where
         let range_proof = recipient_reply.output.proof.clone();
         let (nonce_a, pub_nonce_a) = PublicKey::random_keypair(&mut OsRng);
         let (nonce_x, pub_nonce_x) = PublicKey::random_keypair(&mut OsRng);
-        let pub_nonce = pub_nonce_a + pub_nonce_x;
+        let pub_nonce = self.resources.factories.commitment.commit(&nonce_x, &nonce_a);
         let mut ownership_proof = None;
 
         if let Some(claim_public_key) = claim_public_key {
@@ -1435,6 +1435,13 @@ where
                 .chain(claim_public_key.as_bytes());
 
             let challenge: FixedHash = digest::Digest::finalize(hasher).into();
+
+            warn!(target: LOG_TARGET, "Pub nonce: {}", pub_nonce.to_vec().to_hex());
+            warn!(
+                target: LOG_TARGET,
+                "claim_public_key: {}",
+                claim_public_key.to_vec().to_hex()
+            );
             warn!(target: LOG_TARGET, "Challenge: {}", challenge.to_vec().to_hex());
             ownership_proof = Some(RistrettoComSig::sign(
                 &PrivateKey::from(amount),
--- a/base_layer/wallet/src/transaction_service/service.rs
+++ b/base_layer/wallet/src/transaction_service/service.rs
@@ -1368,7 +1368,11 @@ where
         .await
     }
 
-    /// Creates a transaction to burn some Tari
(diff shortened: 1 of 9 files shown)


hansieodendaal (commented):
The code itself looks fine, but I have some questions regarding the use case of the ownership proof for burned UTXOs.

I do not know the exact use case details of the ownership proof, so in lieu thereof:
- If the "static" ownership proof (a `CommitmentSignature `) is leaked anyone can claim ownership afterwards, even the challenging party.
- An ownership proof produced on the fly will be much more convincing. The challenging party can send a public nonce `Rc` they generated, and the owner can produce the `CommitmentSignature` including the challenge `e = Hash(C || Rc)`.

SWvheerden:
This looks good, but I agree with @hansieodendaal, here. For the proofs to be convincing we need to either make them interactive, or we need Fiat–Shamir heuristic that's derived from the context you are using it. 

For example in the context of the second layer, we need to include in the challenge the shard of the second layer funds that was created from the burn tx. 

As it stands here anyone can reuse this proof and claim ownership.

stringhandler:
I thought about what goes into the challenge on the second layer. At the time of burn, if it is something that ties it to the time of claiming on the second layer, it will not be possible to create the signature. The other option is to return the private key to the commitment, to allow the second layer claimant to create the signature themselves. It felt a bit weird to return the private key, but I suppose we could encrypt it somehow

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1357,19 +1369,17 @@ where
     }
 
     /// Creates a transaction to burn some Tari
```
hansieodendaal: I think we need to explain _claim public key_ a bit, maybe something like this:
```suggestion
    /// Creates a transaction to burn some Tari. The optional _claim public key_ parameter is used in the challenge of the 
    corresponding optional _ownership proof_ return value. Burn commitments and ownership proofs will exclusively be 
    used in the 2nd layer (DAN layer). When such an _ownership proof_ is presented later on as part of some transaction 
    metadata, the _claim public key_ can be revealed to enable verification of the _ownership proof_ and the transaction 
    can be signed with the private key corresponding to the claim public key.
```

hansieodendaal (commented):
Looking good. Please see the comment below.

base_layer/wallet/src/transaction_service/service.rs
stringhandler: thanks

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1401,16 +1416,52 @@ where
...
-            spend_key,
+            spend_key.clone(),
             &self.resources.factories,
         );
 
         let recipient_reply = rtp.get_signed_data()?.clone();
-
+        let commitment = recipient_reply.output.commitment.clone();
+        let range_proof = recipient_reply.output.proof.clone();
+        let (nonce_a, _pub_nonce_a) = PublicKey::random_keypair(&mut OsRng);
```
SWvheerden:
```suggestion
        let nonce_a, = PrivateKey::random(&mut OsRng);
```
You only care about the secret key, so just generate a secret key here

SWvheerden: same with nonce_x

sdbondi: nit(perf): +1 the signature sign method does the nonce -> public nonce work so we can just generate a private nonce.

sdbondi (commented):
LGTM - can merge after minor compile errors and clippys are fixed

sdbondi (approved):

PR #5170, opened 2023-02-07
chore: feature gate release ci

brianp:
Description
---
This PR expeands on #5147 to add additional testing via CI that all variations of feature configuration functions for a given PR. 

It shouldn't bother being reviewed or merged until after #5147 is merged. 

TODO: commit release pipeline changes.

Motivation and Context
---
We don't want people pushing PR's that work on testnet and break compilation of mainnet.

How Has This Been Tested?
---
Via this PR's CI

### DIFF
--- a/.github/workflows/base_node_binaries.json
+++ b/.github/workflows/base_node_binaries.json
@@ -16,7 +16,11 @@
     "cross": true,
     "target_bins": "--bin tari_base_node --bin tari_console_wallet --bin tari_merge_mining_proxy --bin tari_miner",
     "target_cpu": "generic",
-    "features": "safe"
+    "features": "safe",
+    "networks": [
+      {"name": "mainnet", "type": "mainnet"},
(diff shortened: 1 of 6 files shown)


.github/workflows/base_node_binaries.json
```diff
@@ -16,7 +16,11 @@
     "cross": true,
     "target_bins": "--bin tari_base_node --bin tari_console_wallet --bin tari_merge_mining_proxy --bin tari_miner",
     "target_cpu": "generic",
-    "features": "safe"
+    "features": "safe",
+    "networks": [
```
leet4tari, line 20: This is only going to build multiple net versions for linux-arm64. I am thinking if we have different nets for different builds, it's worth having in the json file, but if we will always be building for the same nets, then best add as an include in the matrix for the workflow. I will send an example.

brianp:
Closing this in favour of #5254 

Any feature lost from this PR we can cherry-pick later.

PR #5173, opened 2023-02-09
feat: stagenet network

brianp:
Description
---
This PR introduces the genesis block, and consensus for the upcoming StageNet network.

Apart from this PR I've generated seed node identities, which will be added to the seeds after this is merged, and the seeds are brought up.

Motivation and Context
---
Get into release cycle habits and hygiene.

How Has This Been Tested?
---
CI (but I'm expecting some test to fail on first pass due to the default network change).

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -83,25 +83,26 @@ fn get_stagenet_genesis_block_raw() -> Block {
     // Note: Use print_new_genesis_block_stagenet in core/tests/helpers/block_builders.rs to generate the required
     // fields below
     let excess_sig = Signature::new(
-        PublicKey::from_hex("fcd46ec2a4ad022319b903972c4a30e9ec08ed353e9245f67b474a3028aaf917").unwrap(),
-        PrivateKey::from_hex("74744c8ecaa03c18ab767090590bac45830c47f4cd2b0b40c358d9e6de73080e").unwrap(),
+        PublicKey::from_hex("e487276e01039473094f492f20b707d3805399f799fc9d4ad77c648f2a5c0d18").unwrap(),
+        PrivateKey::from_hex("544bea13bfa55f26f3cbdf986b0f7fff4a056316b252531896577d201f362704").unwrap(),
     );
     let coinbase_meta_sig = CommitmentAndPublicKeySignature::new(
-        Commitment::from_hex("7e236c28aeac44c79a65d855dc678ccc19f960bc0a9c50a2d5bc19ee1df20d23").unwrap(),
-        PublicKey::from_hex("241f5034fd1dda26b223d2b12e816a1f729da86e0faa5609f2258ee6dd2c8e6e").unwrap(),
-        PrivateKey::from_hex("5271e5e4d4c8c4f791a6268aa3b21823a1e56c1c557304853192a876764aa902").unwrap(),
-        PrivateKey::from_hex("75b8fba7ecb555785bca632d142d33d464669c6c6288fb17122a220ca8cceb01").unwrap(),
-        PrivateKey::from_hex("b3000fb10de4c100cc271b3cb5065a916904dc265b37c4c0b45f9cb9a0f7cc00").unwrap(),
+        Commitment::from_hex("b60caad21427512b85ccc38a455419481ff4f45253b611374c68d8e82533ee66").unwrap(),
+        PublicKey::from_hex("76dfffcfa2c67c1b140ee04cf04fc271df193bdb4914e0b163b2507504f28b0f").unwrap(),
+        PrivateKey::from_hex("81631e30518430fd864a010fdba2f08f63c63f974972de179e0c9d7bdfb0d000").unwrap(),
+        PrivateKey::from_hex("be45ce74a588437453c4dc27ed5cc368c422cd6d86f4f180d66a6fd20a458d07").unwrap(),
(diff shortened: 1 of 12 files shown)


base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -63,6 +64,99 @@ pub fn get_genesis_block(network: Network) -> ChainBlock {
     }
 }
 
+pub fn get_stagenet_genesis_block() -> ChainBlock {
```
sdbondi, line 67: Please add a test for this at the bottom of this file like we do for esme and igor.

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -63,6 +64,99 @@ pub fn get_genesis_block(network: Network) -> ChainBlock {
...
+    let block = get_stagenet_genesis_block_raw();
+
+    let accumulated_data = BlockHeaderAccumulatedData {
+        hash: block.hash(),
+        total_kernel_offset: block.header.total_kernel_offset.clone(),
+        achieved_difficulty: 1.into(),
```
SWvheerden, line 73: Is this the actual difficulty of the block?

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -63,6 +64,99 @@ pub fn get_genesis_block(network: Network) -> ChainBlock {
...
+    let coinbase_meta_sig = CommitmentAndPublicKeySignature::new(
+        Commitment::from_hex("7e236c28aeac44c79a65d855dc678ccc19f960bc0a9c50a2d5bc19ee1df20d23").unwrap(),
+        PublicKey::from_hex("241f5034fd1dda26b223d2b12e816a1f729da86e0faa5609f2258ee6dd2c8e6e").unwrap(),
+        PrivateKey::from_hex("5271e5e4d4c8c4f791a6268aa3b21823a1e56c1c557304853192a876764aa902").unwrap(),
+        PrivateKey::from_hex("75b8fba7ecb555785bca632d142d33d464669c6c6288fb17122a220ca8cceb01").unwrap(),
+        PrivateKey::from_hex("b3000fb10de4c100cc271b3cb5065a916904dc265b37c4c0b45f9cb9a0f7cc00").unwrap(),
+    );
+    let coinbase = TransactionOutput::new(
+        TransactionOutputVersion::get_current_version(),
+        OutputFeatures::create_coinbase(6, None),
```
SWvheerden: Does not match the coinbase lockheight. 
But a test unit test should point this out.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -615,6 +615,53 @@ impl ConsensusConstants {
...
+        algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+            max_target_time: 800,
+            min_difficulty: 70_000_000.into(),
+            max_difficulty: u64::MAX.into(),
+            target_time: 200,
+        });
+        let (input_version_range, output_version_range, kernel_version_range) = version_zero();
+        vec![ConsensusConstants {
+            effective_from_height: 0,
+            coinbase_lock_height: 1,
```
SWvheerden: We have been using small numbers to make tests easier, but I take it we want to have stagenet be a larger more realistic number?

SWvheerden (commented):
Looks good so far

base_layer/core/src/consensus/consensus_constants.rs
brianp: Yeah, I did find my test mining was going a little fast. Who needs to make the call on what these final numbers should be?

brianp: Moved the numbers to match current esme constatns + larger lock height.

base_layer/core/src/blocks/genesis_block.rs
brianp: Updated to new lock height and ensured they match the constants.

base_layer/core/src/blocks/genesis_block.rs
brianp: It's the difficulty emse has so I left it as is.

base_layer/core/src/blocks/genesis_block.rs
brianp: Added test, which found errors, that is now fixed. 👍🏻

base_layer/core/src/blocks/genesis_block.rs
SWvheerden: Just check if this is the stagenet difficulty as well.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -615,6 +615,58 @@ impl ConsensusConstants {
...
+        let mut algos = HashMap::new();
+        algos.insert(PowAlgorithm::Sha3, PowAlgorithmConstants {
+            max_target_time: 1800,
+            min_difficulty: 60_000_000.into(),
+            max_difficulty: u64::MAX.into(),
+            target_time: 300,
+        });
+        algos.insert(PowAlgorithm::Monero, PowAlgorithmConstants {
+            max_target_time: 1200,
+            min_difficulty: 70_000_000.into(),
```
stringhandler: I think this is too high and will prevent mining

stringhandler (commented):
Happy to merge once the monero difficulty is confirmed

base_layer/core/src/consensus/consensus_constants.rs
brianp: Yeah, looks like this was the placeholder value from `mainnet` and not the value from esme. Updated to match the esme value.

stringhandler (approved):

PR #5175, opened 2023-02-09
feat: wallet password change

### LINKED ISSUE #5003: Add functionality for update encryption on wallet, when passphrase gets updated by user

AaronFeickert:
Description
---
Adds functionality to change wallet passphrase, updating to the latest encryption version parameters at the same time. Minor refactoring of database main and secondary key operations.

Removes unneeded functionality that tried to assert any cipher seed was encrypted (and [associated tests](https://github.com/tari-project/tari/blob/b7747a29c7032278b3ed88e13823d6e4fe7de45e/base_layer/wallet/src/storage/sqlite_db/wallet.rs#L771-L823)).

Closes [issue 5003](https://github.com/tari-project/tari/issues/5003).

Motivation and Context
---
Currently, wallet passphrases cannot be changed; there is a CLI flow, but it is non-functional.

This PR continues the work started in [PR 5154](https://github.com/tari-project/tari/pull/5154), which refactors wallet database encryption. When the user wishes to change their passphrase, we do the following:
- Attempt to use the existing passphrase to authenticate and decrypt the encrypted database main key. If this fails, the passphrase is incorrect, so we abort.
- Get the latest encryption version parameters from the hardcoded list.
- Generate a fresh secondary key salt. Use it and the new passphrase to derive a new secondary key, using the fetched encryption version parameters.
- Encrypt the database main key using the new secondary key.
- Store the new version identifier, secondary key salt, and encrypted main key in the database, overwriting the previous values.

The update operations are done in a single transaction to mitigate the (unlikely) risk of database corruption. There isn't an included test for this, but you can simulate a failure by having the transaction closure return an error, after which the existing wallet passphrase should continue to work.

How Has This Been Tested?
---
Existing unit tests pass. A new unit test passes. Manually tested a successful passphrase change. Manually tested a failed passphrase change. Manually tested a simulated database transaction failure.

### DIFF
--- a/applications/tari_console_wallet/src/init/mod.rs
+++ b/applications/tari_console_wallet/src/init/mod.rs
@@ -148,39 +148,39 @@ fn prompt_password(prompt: &str) -> Result<SafePassword, ExitError> {
 /// Allows the user to change the password of the wallet.
 pub async fn change_password(
     config: &ApplicationConfig,
-    arg_password: SafePassword,
+    existing: SafePassword,
     shutdown_signal: ShutdownSignal,
     non_interactive_mode: bool,
 ) -> Result<(), ExitError> {
-    let mut wallet = init_wallet(config, arg_password, None, None, shutdown_signal, non_interactive_mode).await?;
+    let mut wallet = init_wallet(
+        config,
+        existing.clone(),
+        None,
+        None,
+        shutdown_signal,
+        non_interactive_mode,
+    )
+    .await?;
 
-    let passphrase = prompt_password("New wallet password: ")?;
+    let new = prompt_password("New wallet password: ")?;
     let confirmed = prompt_password("Confirm new password: ")?;
 
-    if passphrase.reveal() != confirmed.reveal() {
+    if new.reveal() != confirmed.reveal() {
         return Err(ExitError::new(ExitCode::InputError, "Passwords don't match!"));
     }
-
-    // wallet
-    //     .remove_encryption()
-    //     .await
-    //     .map_err(|e| ExitError::new(ExitCode::WalletError, e))?;
-
-    // wallet
-    //     .apply_encryption(passphrase)
-    //     .await
-    //     .map_err(|e| ExitError::new(ExitCode::WalletError, e))?;
 
     println!("Passwords match.");
 
     // If the passphrase is weak, let the user know
-    display_password_feedback(&passphrase);
+    display_password_feedback(&new);
 
-    // TODO: remove this warning when this functionality is added
-    println!();
-    println!("WARNING: Password change functionality is not yet completed, so continue to use your existing password!");
-
-    Ok(())
+    // Use the existing and new passphrases to attempt to change the wallet passphrase
+    wallet.db.change_passphrase(&existing, &new).map_err(|e| match e {
+        WalletStorageError::InvalidPassphrase => {
+            ExitError::new(ExitCode::IncorrectOrEmptyPassword, "Your password was not changed.")
+        },
+        _ => ExitError::new(ExitCode::DatabaseError, "Your password was not changed."),
+    })
 }
 
 /// Populates the PeerConfig struct from:
@@ -667,7 +667,7 @@ pub(crate) fn boot_with_password(
         },
         WalletBoot::Existing | WalletBoot::Recovery => {
             debug!(target: LOG_TARGET, "Prompting for password.");
-            prompt_password("Prompt wallet password: ")?
+            prompt_password("Enter wallet password: ")?
         },
     };
 
--- a/base_layer/wallet/src/storage/sqlite_db/wallet.rs
+++ b/base_layer/wallet/src/storage/sqlite_db/wallet.rs
@@ -31,8 +31,8 @@ use argon2::password_hash::{
     rand_core::{OsRng, RngCore},
     SaltString,
 };
-use chacha20poly1305::{Key, KeyInit, Tag, XChaCha20Poly1305, XNonce};
-use diesel::{prelude::*, SqliteConnection};
+use chacha20poly1305::{Key, KeyInit, XChaCha20Poly1305};
+use diesel::{prelude::*, result::Error, SqliteConnection};
 use log::*;
 use tari_common_types::chain_metadata::ChainMetadata;
 use tari_comms::{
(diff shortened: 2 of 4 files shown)


AaronFeickert:
Opening this PR to enable CI. It's incomplete and still a draft. There's plenty of opportunity for refactoring to avoid code duplication.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -438,6 +438,87 @@ impl WalletBackend for WalletSqliteDatabase {
...
+                    &new_main_key_cipher,
+                    new_aad,
+                    Hidden::hide(main_key.reveal().clone()),
+                )
+                .map_err(WalletStorageError::AeadError)?;
+
+                // Store the secondary key version, secondary key salt, and encrypted main key
+                WalletSettingSql::new(DbKey::SecondaryKeyVersion, argon2_params.id.to_string()).set(&conn)?;
+                WalletSettingSql::new(DbKey::SecondaryKeySalt, new_secondary_key_salt.to_string()).set(&conn)?;
+                WalletSettingSql::new(DbKey::EncryptedMainKey, new_encrypted_main_key.to_hex()).set(&conn)?;
```
SWvheerden: This makes 3 transactions. 
It should be 1 here, this is the crucial part we need to be atomic

AaronFeickert: Updated to use a single transaction. You can simulate a failure by having the transaction closure return an error.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -533,39 +581,24 @@ fn get_cipher_for_db_encryption(
...
+            // Use the secondary key to encrypt the main key
+            let encrypted_main_key = encrypt_main_key(&secondary_key, &main_key, argon2_params.id)?;
 
             // Store the secondary key version, secondary key salt, and encrypted main key
-            WalletSettingSql::new(DbKey::SecondaryKeyVersion, argon2_params.id.to_string()).set(conn)?;
-            WalletSettingSql::new(DbKey::SecondaryKeySalt, secondary_key_salt.to_string()).set(conn)?;
-            WalletSettingSql::new(DbKey::EncryptedMainKey, encrypted_main_key.to_hex()).set(conn)?;
+            WalletSettingSql::new(DbKey::SecondaryKeyVersion, argon2_params.id.to_string()).set(&conn)?;
+            WalletSettingSql::new(DbKey::SecondaryKeySalt, secondary_key_salt.to_string()).set(&conn)?;
+            WalletSettingSql::new(DbKey::EncryptedMainKey, encrypted_main_key.to_hex()).set(&conn)?;
```
hansieodendaal: Please wrap this in a diesel transaction as well, similar to what was done in `fn change_passphrase(`

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -438,92 +442,136 @@ impl WalletBackend for WalletSqliteDatabase {
...
+                    WalletSettingSql::new(DbKey::SecondaryKeyVersion, new_argon2_params.id.to_string())
+                        .set(&conn)
+                        .map_err(|_| Error::RollbackTransaction)?;
+                    WalletSettingSql::new(DbKey::SecondaryKeySalt, new_secondary_key_salt.to_string())
+                        .set(&conn)
+                        .map_err(|_| Error::RollbackTransaction)?;
+                    WalletSettingSql::new(DbKey::EncryptedMainKey, new_encrypted_main_key.to_hex())
+                        .set(&conn)
+                        .map_err(|_| Error::RollbackTransaction)?;
+
```
hansieodendaal, line 490: An additional _defence-in-depth_ approach here would be, before the transaction closure, to read the 3 values and verify that they were changed as instructed and if not trigger the rollback. (_Needless to say, for this reading the 3 values could be DRYed up._) This would be in line with the diesel examples.

hansieodendaal (commented):
Nice, I tested the password change functionality from the command line, and it works as expected.

I have some comments regarding the atomicity below.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Is this necessary? If any of the operations were to fail, no encryption would be set up, the function would return an error, and no data would be corrupted.

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Is it possible that such a write operation could apparently succeed without actually writing the expected data? Or would this be more of a test of the `WalletSettingSql` wrappers?

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: I don't think is particularly necessary, since a failure in one of these operations should not result in any data loss. But it doesn't hurt, so it's been added.

SWvheerden (approved):
utACK

base_layer/wallet/src/storage/sqlite_db/wallet.rs
SWvheerden: This should not be necessary, the database should throw an error if the set command did not do as expected.

hansieodendaal (approved):
ACK

applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -148,39 +148,39 @@ fn prompt_password(prompt: &str) -> Result<SafePassword, ExitError> {
...
+        existing.clone(),
+        None,
+        None,
+        shutdown_signal,
+        non_interactive_mode,
+    )
+    .await?;
 
-    let passphrase = prompt_password("New wallet password: ")?;
+    let new = prompt_password("New wallet password: ")?;
```
stringhandler, line 165: nit: while not a reserved word in rust, it does seem strange to see `new` as a variable

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -438,92 +442,136 @@ impl WalletBackend for WalletSqliteDatabase {
...
+fn get_db_cipher(
+    database_connection: &WalletDbConnection,
+    passphrase: &SafePassword,
 ) -> Result<XChaCha20Poly1305, WalletStorageError> {
-    // We'll bind the version identifier to this domain before it's used
-    let main_key_domain = b"wallet_main_key_encryption_v".to_vec();
+    let conn = database_connection.get_pooled_connection()?;
+
+    // Fetch the database fields used for encryption, if they exist
+    let secondary_key_version = WalletSettingSql::get(&DbKey::SecondaryKeyVersion, &conn)?;
```
stringhandler, line 572: Should technically also be in  a transaction since it could change half way through reading

stringhandler (approved):

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: Would using a transaction catch this edge case? Even if not, I don't think it's a practical problem here. In the event that the fields were to be only partially updated when the reads happen, the subsequent key derivation will fail and return an error. The user can then simply try again to load the wallet.

AaronFeickert: Will address in [this issue](https://github.com/tari-project/tari/issues/5177).

PR #5178, opened 2023-02-13
feat: refactor key-related field operations to be atomic

### LINKED ISSUE #5177: Database reads on key-related fields should be atomic
It was [pointed out](https://github.com/tari-project/tari/pull/5175#discussion_r1104085191) that database reads on multiple fields have no inherent guarantee of consistency, and may return unexpected results due to write operations. However, SQLite apparently does [guarantee read consistency](https://www.sqlite.org/isolation.html) within a read-only transaction.

While the [use case in question](https://github.com/tari-project/tari/blob/d732845b3d1b2c2f49287a56180011305e5c7cb0/base_layer/wallet/src/storage/sqlite_db/wallet.rs#L572-L574) will fail safe in the event this edge case happens, it's something of a footgun to rely on failed key derivation to catch this.

A more comprehensive and general approach to read transactions may be warranted, but for now, the read operations used for key-related database fields (version identifier, secondary key salt, and encrypted main key) should be done atomically. Even if these reads produce outdated but consistent data due to a write transaction, they will still produce a valid database cipher with the correct main key.

This is also a good opportunity to refactor the atomic write operations for these same fields to reduce code duplication and make intent more clear.

AaronFeickert:
Description
---
Refactors key-related database field operations to be atomic.

Closes [issue 5177](https://github.com/tari-project/tari/issues/5177). 

Motivation and Context
---
Key-related database fields always travel together. We need a consistent set of secondary key version identifier, secondary key salt, and encrypted main key in order to set up the `XChaCha20-Poly1305` cipher used for database encryption operations and passphrase changes.

While a [recent PR](https://github.com/tari-project/tari/pull/5175) ensures that write operations for these fields are done atomically via a write transaction, there is no corresponding read transaction. It's therefore possible that those fields are inconsistent. While this should only result in an error and require the user to load their wallet again, it seemed like a smart idea to ensure that reads are consistent for any future use cases.

This PR refactors the handling of those fields to reduce redundancy and ensure atomicity for reads and writes. It introduces a new `DatabaseKeyFields` struct that handles reads and writes, and additionally takes care of encoding and decoding of the underlying data.

It also makes the handling of the three fields more consistent. Previously, individual reads and writes required the use of a complex `match` to handle different states. This functionality has been mostly moved into `DatabaseKeyFields` to make these states more apparent.

How Has This Been Tested?
---
Existing unit tests pass. Manually tested the following operations:
- setting up a new wallet and successfully loading it with the correct passphrase
- setting up a new wallet and unsuccessfully loading it with an incorrect passphrase
- setting up a new wallet and unsuccessfully loading it due to a simulated read transaction failure
- failing to set up a new wallet due to a simulated write transaction failure
- failing to set up a new wallet due to a simulated read transaction failure
- a successful passphrase change via CLI
- an unsuccessful passphrase change via CLI due to an incorrect existing passphrase
- an unsuccessful passphrase change via CLI due to a mismatched new passphrase
- an unsuccessful passphrase change via CLI due to a simulated write transaction failure
- an unsuccessful passphrase change via CLI due to a simulated read transaction failure

It does not seem possible to directly test read operation inconsistency caused by a simultaneous write operation.

### DIFF
--- a/base_layer/wallet/src/storage/sqlite_db/wallet.rs
+++ b/base_layer/wallet/src/storage/sqlite_db/wallet.rs
@@ -104,13 +104,13 @@ impl Argon2Parameters {
     }
 }
 
-/// A structure to hold key-related database field data, to make atomic operations cleaner
-pub struct DatabaseKeyFields {
+/// A structure to hold encryption-related database field data, to make atomic operations cleaner
+pub struct DatabaseEncryptionFields {
     secondary_key_version: u8,   // the encryption parameter version
     secondary_key_salt: String,  // the high-entropy salt used to derive the secondary key
     encrypted_main_key: Vec<u8>, // the main key, encrypted with the secondary key
 }
-impl DatabaseKeyFields {
+impl DatabaseEncryptionFields {
     /// Read and parse field data from the database atomically
     pub fn read(connection: &SqliteConnection) -> Result<Option<Self>, WalletStorageError> {
         let mut secondary_key_version: Option<String> = None;
@@ -143,7 +143,7 @@ impl DatabaseKeyFields {
                 let encrypted_main_key =
                     from_hex(&encrypted_main_key).map_err(|e| WalletStorageError::ConversionError(e.to_string()))?;
 
-                Ok(Some(DatabaseKeyFields {
+                Ok(Some(DatabaseEncryptionFields {
                     secondary_key_version,
                     secondary_key_salt,
                     encrypted_main_key,
@@ -523,7 +523,7 @@ impl WalletBackend for WalletSqliteDatabase {
         let conn = self.database_connection.get_pooled_connection()?;
 
         // Get the existing key-related data so we can decrypt the main key
-        match DatabaseKeyFields::read(&conn) {
+        match DatabaseEncryptionFields::read(&conn) {
             // Key-related data was present and valid
             Ok(Some(data)) => {
                 // Use the given version if it is valid
@@ -546,7 +546,7 @@ impl WalletBackend for WalletSqliteDatabase {
                 let new_encrypted_main_key = encrypt_main_key(&new_secondary_key, &main_key, new_argon2_params.id)?;
 
                 // Store the new key-related fields
-                DatabaseKeyFields {
+                DatabaseEncryptionFields {
                     secondary_key_version: new_argon2_params.id,
                     secondary_key_salt: new_secondary_key_salt,
                     encrypted_main_key: new_encrypted_main_key,
@@ -626,7 +626,7 @@ fn get_db_cipher(
     let conn = database_connection.get_pooled_connection()?;
 
     // Either set up a new main key, or decrypt it using existing data
-    let main_key = match DatabaseKeyFields::read(&conn) {
+    let main_key = match DatabaseEncryptionFields::read(&conn) {
         // Encryption is not set up yet
         Ok(None) => {
             // Generate a high-entropy main key
@@ -645,7 +645,7 @@ fn get_db_cipher(
             let encrypted_main_key = encrypt_main_key(&secondary_key, &main_key, argon2_params.id)?;
 
             // Store the key-related fields
-            DatabaseKeyFields {
+            DatabaseEncryptionFields {
                 secondary_key_version: argon2_params.id,
                 secondary_key_salt,
                 encrypted_main_key,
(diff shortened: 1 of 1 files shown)


AaronFeickert:
Marked as draft while manual testing is in progress.

AaronFeickert:
As before, you can simulate a [read](https://github.com/tari-project/tari/blob/3a1ebf87f5b7b95c548713e1f787a147547bdaa1/base_layer/wallet/src/storage/sqlite_db/wallet.rs#L121-L132) or [write](https://github.com/tari-project/tari/blob/3a1ebf87f5b7b95c548713e1f787a147547bdaa1/base_layer/wallet/src/storage/sqlite_db/wallet.rs#L163-L177) transaction failure by having the corresponding closure return an `Error::RollbackTransaction` [error](https://docs.diesel.rs/master/diesel/connection/trait.Connection.html#method.transaction).

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -104,6 +104,82 @@ impl Argon2Parameters {
     }
 }
 
+/// A structure to hold key-related database field data, to make atomic operations cleaner
+pub struct DatabaseKeyFields {
```
SWvheerden:
```suggestion
pub struct DatabaseEncryptionFields {
```
NIT: Just reading `DatabaseKeyFields` does not tell me what it is, to me it could be anything from database Max_read/write settings, to a key value store.

SWvheerden (dismissed):
utACK

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -104,6 +104,82 @@ impl Argon2Parameters {
...
+    secondary_key_version: u8,   // the encryption parameter version
+    secondary_key_salt: String,  // the high-entropy salt used to derive the secondary key
+    encrypted_main_key: Vec<u8>, // the main key, encrypted with the secondary key
+}
+impl DatabaseKeyFields {
```
hansieodendaal: This struct with its implementation is really neat, I like it.

hansieodendaal (dismissed):
utaCK

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -104,6 +104,82 @@ impl Argon2Parameters {
...
+    pub fn read(connection: &SqliteConnection) -> Result<Option<Self>, WalletStorageError> {
+        let mut secondary_key_version: Option<String> = None;
+        let mut secondary_key_salt: Option<String> = None;
+        let mut encrypted_main_key: Option<String> = None;
+
+        // Read all fields atomically
+        connection
+            .transaction::<_, Error, _>(|| {
+                secondary_key_version = WalletSettingSql::get(&DbKey::SecondaryKeyVersion, connection)
+                    .map_err(|_| Error::RollbackTransaction)?;
```
stringhandler, line 124: This mapping is not actually needed. Any Err will rollback the transaction. The docs on diesel.rs don't reflect this in v1.4, but have been updated in the 2.0 docs on docs.rs. In either case, I checked the source and it rolls back on Err.

So it's better to just return the error with `?` and have the original error find their way to the logs

base_layer/wallet/src/storage/sqlite_db/wallet.rs
```diff
@@ -104,6 +104,82 @@ impl Argon2Parameters {
...
+    }
+
+    /// Encode and write field data to the database atomically
+    pub fn write(&self, connection: &SqliteConnection) -> Result<(), WalletStorageError> {
+        // Because the encoding can't fail, just do it inside the write transaction
+        connection
+            .transaction::<_, Error, _>(|| {
+                WalletSettingSql::new(DbKey::SecondaryKeyVersion, self.secondary_key_version.to_string())
+                    .set(connection)
+                    .map_err(|_| Error::RollbackTransaction)?;
```
stringhandler, line 167: Same comment here

stringhandler (commented):
Happy to merge after the map_errs are removed

base_layer/wallet/src/storage/sqlite_db/wallet.rs
AaronFeickert: I had tried this initially, but the compiler complained that the internal error couldn't be converted to a `diesel` error.

stringhandler (approved):

PR #5179, opened 2023-02-14
chore: wrap leaf index in struct

Cifko:
Description
---
Wrap leaf index in a struct

Motivation and Context
---
To prevent accidents that leaf index is used as a node index and vice versa.

### DIFF
--- a/base_layer/mmr/src/common.rs
+++ b/base_layer/mmr/src/common.rs
@@ -32,12 +32,15 @@ use crate::{error::MerkleMountainRangeError, Hash};
 
 const ALL_ONES: usize = std::usize::MAX;
 
+#[derive(Copy, Clone)]
+pub struct LeafIndex(pub usize);
+
 /// Returns the MMR node index derived from the leaf index.
-pub fn node_index(leaf_index: usize) -> usize {
-    if leaf_index == 0 {
+pub fn node_index(leaf_index: LeafIndex) -> usize {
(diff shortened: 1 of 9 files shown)


stringhandler (dismissed):
Good idea

base_layer/mmr/src/common.rs
```diff
@@ -214,14 +217,14 @@ mod test {
...
-        assert_eq!(node_index(7), 11);
-        assert_eq!(node_index(8), 15);
+        assert_eq!(node_index(LeafIndex(0)), 0);
+        assert_eq!(node_index(LeafIndex(1)), 1);
+        assert_eq!(node_index(LeafIndex(2)), 3);
+        assert_eq!(node_index(LeafIndex(3)), 4);
+        assert_eq!(node_index(LeafIndex(5)), 8);
+        assert_eq!(node_index(LeafIndex(6)), 10);
+        assert_eq!(node_index(LeafIndex(7)), 11);
+        assert_eq!(node_index(LeafIndex(8)), 15);
```
delta1, line 227: why not implement `From` and then you can just do 

```suggestion
        assert_eq!(node_index(8.into()), 15);
```

Cifko: I thought about it. But then again, someone will take a number and without thinking call an `into` on it. And we will have the same hard to spot bug mixing node index and leaf index.

stringhandler (approved):

PR #5181, opened 2023-02-14
fix(wallet-grpc): return correct available balance and add timelocked balance

sdbondi:
Description
---
get_balance call to returns timelocked_balance
get_balance call returns funds that are immediately spendable

Motivation and Context
---
Calls to get_balance seemingly do not yield the correct balance, e.g your wallet ui says available balance == 0 but the grpc call says available_balance > 0 (because it includes the timelocked funds). This PR changes the available_balance to reflect the amount of funds immediately spendable. This only happens on the grpc level to avoid a breaking change in other code (FFI etc)

How Has This Been Tested?
---
Existing cucumber. Tested with VN that queries the balance.

BREAKING CHANGE: get_balance grpc call has changed, available_balance no longer includes timelocked funds only immediately available funds. This probably only is used currently in cucumber tests.

### DIFF
--- a/applications/tari_app_grpc/proto/wallet.proto
+++ b/applications/tari_app_grpc/proto/wallet.proto
@@ -232,6 +232,7 @@ message GetBalanceResponse {
     uint64 available_balance = 1;
     uint64 pending_incoming_balance = 2;
     uint64 pending_outgoing_balance = 3;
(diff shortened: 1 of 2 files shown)


SWvheerden (approved):
utAck

stringhandler (approved):

PR #5182, opened 2023-02-14
fix: addresses mmr `find_peaks` bug

jorgeantonio21:
Description
---
Currently, the Tari MMR repo contains a bug on the `find_peaks` function, in `base_layer/mmr/src/common.rs`. The source of the problem arises from the fact that the current implementation does not consider possible Merkle Mountain Ranges, in which all nodes are filled to the possible top (that is, every possible spawned peak exists in the MMR). Moreover, the current function implementation handles this issue by returning an empty `vec![]`. This works as a 'silent error' being passed upstream. In order to handle this problem, we impose that the height to check that a new peak exists only decreases if we are guaranteed that every peak at the same height does in fact not exists. In this way, proofs can be created for every MMR size. 

PS: Want to add further unit tests.

Motivation and Context
---

How Has This Been Tested?
---
More extensive unit testing.

### DIFF
--- a/base_layer/mmr/src/common.rs
+++ b/base_layer/mmr/src/common.rs
@@ -84,10 +84,12 @@ pub fn find_peaks(size: usize) -> Option<Vec<usize>> {
         //   / \
         //  0   1   3   4
         // is invalid, as it can be completed to form
+        //       6
+        //     /    \
         //    2      5
         //  /  \    /  \
         // 0    1  3    4
-        // which is of size 6 (with peaks [2, 5])
+        // which is of size 7 (with single peak [6])
         return None;
     }
     Some(peaks)
--- a/base_layer/mmr/src/common.rs
+++ b/base_layer/mmr/src/common.rs
@@ -75,8 +75,12 @@ pub fn find_peaks(size: usize) -> Vec<usize> {
             sum_prev_peaks += peak_size;
             num_left -= peak_size;
         }
-        // need to verify that peaks can exist on the same height, in which case
-        // the size doesn't give rise to a complete mmr
+        // need to verify if other peaks exist on the same height, in which case
+        // the size doesn't generate a complete mmr, e.g.
+        //    2
+        //   / \
+        //  0   1   3   4
+        // where we have two peaks at height 0.
         if num_left < peak_size {
             peak_size >>= 1;
         }
--- a/base_layer/mmr/src/common.rs
+++ b/base_layer/mmr/src/common.rs
@@ -248,17 +248,34 @@ mod test {
     fn peak_vectors() {
         assert_eq!(find_peaks(0), Vec::<usize>::new());
         assert_eq!(find_peaks(1), vec![0]);
+        assert_eq!(find_peaks(2), vec![0, 1]);
         assert_eq!(find_peaks(3), vec![2]);
         assert_eq!(find_peaks(4), vec![2, 3]);
-        assert_eq!(find_peaks(15), vec![14]);
-        assert_eq!(find_peaks(23), vec![14, 21, 22]);
-        assert_eq!(find_peaks(123), vec![62, 93, 108, 115, 122]);
-        assert_eq!(find_peaks(130), vec![126, 129]);
-        assert_eq!(find_peaks(56), vec![30, 45, 52, 55]);
-        assert_eq!(find_peaks(60), vec![30, 45, 52, 59]);
-        assert_eq!(find_peaks(28), vec![14, 21, 24, 27]);
         assert_eq!(find_peaks(5), vec![2, 3, 4]);
         assert_eq!(find_peaks(6), vec![2, 5]);
+        assert_eq!(find_peaks(7), vec![6]);
+        assert_eq!(find_peaks(8), vec![6, 7]);
+        assert_eq!(find_peaks(9), vec![6, 7, 8]);
+        assert_eq!(find_peaks(10), vec![6, 9]);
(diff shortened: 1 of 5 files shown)


base_layer/mmr/src/common.rs
```diff
@@ -75,7 +75,15 @@ pub fn find_peaks(size: usize) -> Vec<usize> {
             sum_prev_peaks += peak_size;
             num_left -= peak_size;
         }
-        peak_size >>= 1;
+        // need to verify if other peaks exist on the same height, in which case
+        // the size doesn't generate a complete mmr, e.g.
+        //    2
+        //   / \
+        //  0   1   3   4
+        // where we have two peaks at height 0.
```
sdbondi: Ah! So the reason for the vec![] return is to say that the mmr size cannot exist.

Say we have an MMR of size 5 (nodes not leaf nodes), which would look like this
```
   2
 /   \
0    1  3    4
```

However that isnt a valid MMR size because it should look like this (7 nodes)

```text
        //        6
        //      /    \
        //    2      5
        //   / \    /  \
        //  0   1   3   4
```
As expected, `find_peaks(7) == [6]`

So perhaps we should return an Option here instead of an empty vec to force the caller to handle the case of an invalid size. And add a comment explaining what `None` means.

Then in `MerkleProof::generate_proof` and `MerkleProof::verify`, there was a missing check, you would do something like

```rust
let peaks = find_peaks(self.mmr_size).ok_or(MerkleProofError::InvalidMmrSize)?;
```

base_layer/mmr/src/common.rs
```diff
@@ -75,7 +75,15 @@ pub fn find_peaks(size: usize) -> Vec<usize> {
...
+        if num_left < peak_size {
```
sdbondi, line 84: This is equivalent to an else branch on the previous if statement. I think the previous code was correct, though returning an vec! for an invalid size parameter should perhaps be an Option rather

Then in the merkle proof generation (`MerkleProof::generate_proof`), there was a missing check, you would do something like

```rust
if peaks.is_none() {
            return Err(MerkleProofError::Unexpected);
        }
```

sdbondi (commented):
I think the problem was missing checks in `MerkleProof::generate_proof` and `MerkleProof::verify`, suggest we return an option from `find_peaks` instead of an empty vec! so that any caller of find_peaks is forced to do the check.

This is important because we will be verifying untrusted data, so a malicious entity can set `mmr_size` in the merkle proof to whatever they want.

base_layer/mmr/src/common.rs
stringhandler: I think it's more intuitive (and useful) to always return the peak that would be created.

jorgeantonio21: I think the first example could make sense and might well represent a Merkle Mountain Range of size 5, and for that reason, we should be able to generate proofs for it. The implementations I have seen so far deal with any size MMR. But happy to make the `find_peaks` return an option in case, the size doesn't correspond to a fully generated MMR.

sdbondi: A merkle mountain range cannot be size 5 because then it is malformed. In the above example, nodes 3 and 4 must be hashed to 5 and 2 and 5 must be hashed to 6 or the MMR is not a valid MMR.

SWvheerden: The fact that it might break on `find_peaks` on invalid numbers, might point out that we might have an issue with the mmr, 
I agree here with @sdbondi an MMR cannot be for example size 5, 
And MMR containing 5 elements would be of size 8. 
```
         6
      /     \
    2        5
   / \      /  \
  0   1    3   4   7
```

It might be that our MMR implementation does not cap the peaks correctly

jorgeantonio21: I agree with the above comments. Let me refactor the code to output a `None` in case the MMR is not fully generated/complete/valid and check its usage later on. @SWvheerden in the example above, the size of the MMR is 8, spanning a 5 element set, I think.

CjS77: > Ah! So the reason for the vec![] return is to say that the mmr size cannot exist.
> 
> However that isnt a valid MMR size because it should look like this (7 nodes)
> 
> ```
>         //        6
>         //      /    \
>         //    2      5
>         //   / \    /  \
>         //  0   1   3   4
> ```
> 
> As expected, `find_peaks(7) == [6]`
> 
> So perhaps we should return an Option here instead of an empty vec to force the caller to handle the case of an invalid size. And add a comment explaining what `None` means.
> 

Yes, this. `find_peaks` should really be private and documented such that it should not be called before "bagging" the MMR. Returning an Option is ok too. I'll bet this wasn't done on day 1 on performance grounds and the assumption that it gets called after bagging (but should have been made private to make that assumption a contract). But @stringhandler and I discussed this and found that `Option` overheads aren't that high.

base_layer/mmr/src/merkle_proof.rs
```diff
@@ -137,7 +137,7 @@ impl MerkleProof {
 
         // Get the peaks of the merkle trees, which are bagged together to form the root
         // For the proof, we must leave out the local root for the candidate node
-        let peaks = find_peaks(mmr_size);
+        let peaks = find_peaks(mmr_size).ok_or(MerkleMountainRangeError::InvalidMmrSize)?;
```
CjS77, line 140: FYI If the MMR has been bagged, `find_peaks` will not error

CjS77:
ACK

SWvheerden (approved):
utACK

stringhandler (approved):

PR #5187, opened 2023-02-16
fix: ignores peer seed test that requires external DNS

sdbondi:
Description
---
 ignores peer seed test that requires external DNS

Motivation and Context
---
This test is currently failing and was originally ignored because it depends on an external DNS to pass.

How Has This Been Tested?
---
Test ignored


stringhandler (approved):

jorgeantonio21 (approved):
Perfect, thanks @sdbondi !

PR #5189, opened 2023-02-17
feat(consensus)!: add balanced binary merkle tree

Cifko:
Description
---
Implement balanced binary merkle tree and use it instead of mmr for vns. It has no effect on the base layer functionality apart for being a breaking change (it changes the vn merkle root).  In the base layer only the generation is used. The proofs are used only in `tari-dan` (signing votes).
The advantages of MMR over binary tree is the time complexity when adding an element. But that's not our case. We regenerate the tree everytime. So it's much simpler to generate balanced binary tree. Which is the best option for this. The proof time/size for binary tree is always log n. For MMR the worst case scenario is 2*log n (when there is peak of every possible height). 
The merged proofs are part of another PR.

Some comparison between balanced binary merkle tree vs merkle mountain range.
The data set I used was number of VNs of 1000 to 1M (step by 1000), committee size from 10 to 200 (step by 10) (In reality the committee can be bigger, this is representing the number of proofs).
For generating the proofs the random subset of VNs were selected (same for BMT and MMR). 

- **Tree generation takes 62.03% of the time**
tree_generation
- **Generating proofs takes 42.37% of the time**
**Generating merged proofs takes 85.96% of the time** 
In the graph below the green represents only the merging time. But in reality we need to generate the proof anyway and then merge them (figure above takes this into account)
proof_generation
- **Verifying proofs takes 68.98% of the time**
**Verifying merged proofs takes 54.77% of the time**
verifying_proofs
- **Proof size is 76.12% of the size**
**Merged proof size is 9.47% of the size**
proof_size

How Has This Been Tested?
---
There are couple cargo tests.

This is a breaking change : **The Merkle root for VNs is different.** It's equal in the genesis block, but once there is a single vn registration it starts to differ.

### DIFF
--- a/base_layer/core/src/chain_storage/tests/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/tests/blockchain_database.rs
@@ -583,6 +583,7 @@ mod validator_node_merkle_root {
     use super::*;
     use crate::{
         transactions::transaction_components::{OutputFeatures, ValidatorNodeSignature},
+        ValidatorNodeBMT,
         ValidatorNodeMmr,
     };
 
@@ -618,18 +619,13 @@ mod validator_node_merkle_root {
             .unwrap()
             .unwrap();
 
-        let mut vn_mmr = ValidatorNodeMmr::new(Vec::new());
-        vn_mmr
-            .push(
-                Blake256::new()
-                    .chain(public_key.as_bytes())
-                    .chain(shard_key.as_slice())
-                    .finalize()
-                    .to_vec(),
-            )
-            .unwrap();
+        let vn_bmt = ValidatorNodeBMT::create(vec![Blake256::new()
+            .chain(public_key.as_bytes())
+            .chain(shard_key.as_slice())
+            .finalize()
+            .to_vec()]);
 
         let tip = db.fetch_tip_header().unwrap();
-        assert_eq!(tip.header().validator_node_mr, vn_mmr.get_merkle_root().unwrap());
+        assert_eq!(tip.header().validator_node_mr, vn_bmt.get_merkle_root());
     }
 }
--- a/base_layer/mmr/src/balanced_binary_merkle_proof.rs
+++ b/base_layer/mmr/src/balanced_binary_merkle_proof.rs
@@ -30,41 +30,44 @@ use crate::{common::hash_together, BalancedBinaryMerkleTree, Hash};
 #[derive(Debug)]
 pub struct BalancedBinaryMerkleProof<D> {
     pub path: Vec<Hash>,
-    pub leaf_index: usize,
+    pub node_index: usize,
     _phantom: PhantomData<D>,
 }
+
+// Since this is balanced tree, the index `2k+1` is always left child and `2k` is right child
 
 impl<D> BalancedBinaryMerkleProof<D>
 where D: Digest + DomainDigest
 {
-    pub fn verify_consume(mut self, root: &Hash, leaf_hash: Hash) -> bool {
+    pub fn verify(&self, root: &Hash, leaf_hash: Hash) -> bool {
         let mut computed_root = leaf_hash;
+        let mut node_index = self.node_index;
         for sibling in self.path.iter() {
-            if self.leaf_index & 1 == 1 {
+            if node_index & 1 == 1 {
                 computed_root = hash_together::<D>(&computed_root, sibling);
             } else {
                 computed_root = hash_together::<D>(sibling, &computed_root);
             }
-            self.leaf_index = (self.leaf_index - 1) >> 1;
+            node_index = (node_index - 1) >> 1;
         }
         &computed_root == root
     }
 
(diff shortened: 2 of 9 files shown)


SWvheerden (dismissed):
utACK
Pending Green CI

base_layer/mmr/src/balanced_binary_merkle_proof.rs
```diff
@@ -0,0 +1,88 @@
...
+{
+    pub fn verify_consume(mut self, root: &Hash, leaf_hash: Hash) -> bool {
+        let mut computed_root = leaf_hash;
+        for sibling in self.path.iter() {
+            if self.leaf_index & 1 == 1 {
+                computed_root = hash_together::<D>(&computed_root, sibling);
+            } else {
+                computed_root = hash_together::<D>(sibling, &computed_root);
+            }
+            self.leaf_index = (self.leaf_index - 1) >> 1;
```
jorgeantonio21: Why do we have to update the leaf index while traversing a path, to verify a Merkle proof ?

base_layer/mmr/src/balanced_binary_merkle_tree.rs
```diff
@@ -0,0 +1,147 @@
...
+            self.hashes[0].clone()
+        }
+    }
+
+    pub(crate) fn get_hash(&self, pos: usize) -> &Hash {
+        &self.hashes[pos]
+    }
+
+    pub fn get_leaf(&self, leaf_index: usize) -> &Hash {
+        &self.hashes[leaf_index + (self.hashes.len() >> 1)]
```
jorgeantonio21: Nitty suggestion

```rs
&self.hashes[self.get_leaf_index(leaf_index)]

File: base_layer/mmr/src/balanced_binary_merkle_tree.rs [hunk 3]
```diff
@@ -0,0 +1,147 @@
...
+        let bmt = BalancedBinaryMerkleTree::<DomainSeparatedHasher<Blake256, TestDomain>>::create(leaves);
+        let root = bmt.get_merkle_root();
+        assert_eq!(root, vec![
+            72, 54, 179, 2, 214, 45, 9, 89, 161, 132, 177, 251, 229, 46, 124, 233, 32, 186, 46, 87, 127, 247, 19, 36,
+            225, 191, 167, 189, 58, 58, 39, 74
+        ]);
+    }
+
+    #[test]
+    fn test_single_node_tree() {
```
jorgeantonio21, line 125: I would suggest add a few more tests with other leaf lengths (e.g. 2, 4, 7), if feasible.

jorgeantonio21 (dismissed):
Very nice ! Left a few minor comments

base_layer/mmr/src/balanced_binary_merkle_proof.rs
Cifko: The naming is wrong, I'm going to change it to node_index.

base_layer/mmr/src/balanced_binary_merkle_tree.rs
Cifko: yeah, I wrote them in different order :-)

base_layer/mmr/src/balanced_binary_merkle_tree.rs
Cifko: There is not much to test with the generation of the tree. In the `balanced_binary_merkle_proof` file there is a bigger tree. I can add some other tree sizes there.

base_layer/core/src/chain_storage/tests/blockchain_database.rs
```diff
@@ -618,18 +619,13 @@ mod validator_node_merkle_root {
...
-                    .chain(shard_key.as_slice())
-                    .finalize()
-                    .to_vec(),
-            )
-            .unwrap();
+        let vn_bmt = ValidatorNodeBMT::create(vec![Blake256::new()
+            .chain(public_key.as_bytes())
+            .chain(shard_key.as_slice())
+            .finalize()
+            .to_vec()]);
```
AaronFeickert, line 626: Unless there's a compelling reason not to, this should use a domain-separated hasher.

AaronFeickert (changes requested):

base_layer/core/src/chain_storage/tests/blockchain_database.rs
Cifko: This is just a test. The balanced tree is using domain-separated hasher. But as I'm looking at the code for the hash of VNs we are not using the domain-separated hasher there. But that's not part of this PR.

AaronFeickert: Whoops, made this comment in the wrong spot. Sorry for the confusion.

AaronFeickert: Opened an [issue](https://github.com/tari-project/tari/issues/5205) for this, so another PR can address it.

stringhandler (approved):
looking good

Issue #4538, opened 2022-08-25
bug: log dir is relative to binary instead of base-path


delta1:
**Describe the bug**

logs dir is relative to binary instead of base-path 

**To Reproduce**

`./tari_base_node --base-path /tmp/tari --network dibbler` 

`logs` dir will be in current directory instead of directory specified by `--base-path` 

**Expected behavior**

logs should go in the base-path dir 

**Desktop (please complete the following information):**

 - OS & Version: 
Ubuntu 22.04
0.32.12 dibbler

brianp:
hah cheers mate, came to close this but it's taken care of 😆

PR #5197, opened 2023-02-21
feat: log to base dir

brianp:
Description
---
- Change the logging initialization to account for the `-b` option on application startup, and dynamically replace the new `{{log_dir}}` placeholder value inside the log4rs files.
- Update logging directory naming for consistency from `base-node` to `base_node`
- Fix the log4rs config for the tari miner so it will log to stdout as well as the created log file
- Update tari miner log target for consistency

Motivation and Context
---
log4rs considers the cli working directory the root. From there it appends the `path` from the log4rs yaml file as a relative path (unless it's an absolute path). The problem with this is that it completely disregards the standard Tari cli option `-b` for assigning a base directory. This can result in the configuration in the expected assigned location, and the logging ending up in whatever working directory you happened to be in at the time. This has resulted in me finding lots of orphaned logging directories and files spread over my local system.

How Has This Been Tested?
---
Tested locally by using the `-b` option to a location different than the current working directory and ensuring _all_ config and logs end up in the directory provided. Whereas before the `log` dir still would have been created in the working directory.

BREAKING CHANGE: This PR changes the logging configuration which is only generated on the first run. Users may wish to remove log config files:
- `base_dir/config/base_node/log4rs.yml`
- `base_dir/config/miner/log4rs.yml`
- `base_dir/config/wallet/log4rs.yml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2918,7 +2918,7 @@ dependencies = [
  "serde",
  "serde-value",
  "serde_json",
- "serde_yaml",
+ "serde_yaml 0.8.26",
  "thiserror",
  "thread-id",
  "typemap-ors",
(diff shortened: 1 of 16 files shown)


stringhandler (approved):

delta1:
#4538

SWvheerden (approved):
utACK

stringhandler (approved):

PR #5201, opened 2023-02-22
Merge stagenet to development DO NOT SQUASH

stringhandler:


brianp:
I actually thing we should _not_ do this.

Thought about it this morning. The appropriate way to handle the hotfix is:
Merge fix into stagenet branch
Also merge fix into development branch

Hotfixes will always require two PR's. MainNet, StageNet, nor NextNet should ever get merged backwards into development.

brianp (changes requested):
I actually think we should not do this.

Thought about it this morning. The appropriate way to handle the hotfix is:
Merge fix into stagenet branch
Also merge fix into development branch

Hotfixes will always require two PR's. MainNet, StageNet, nor NextNet should ever get merged backwards into development.

PR #5209, opened 2023-02-27
fix: source coverage ci failure

brianp:
Description
---
Don't run cucumber during the coverage runner. It always returns a non 0 exit code when additional cucumber params aren't passed making CI think it's a failure.

Motivation and Context
---
After PR's get merged the coverage ci fails: https://github.com/tari-project/tari/actions/runs/4281676173/jobs/7454994490

How Has This Been Tested?
---
#yolo 🤞🏻

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -45,7 +45,7 @@ jobs:
           LLVM_PROFILE_FILE: "coverage_data-%p-%m.profraw"
         with:
           command: test
-          args: --all-features --no-fail-fast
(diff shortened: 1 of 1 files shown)


stringhandler:
Would be really useful to get the coverage from the cucumber as well. Is there a solution that can make this work?

brianp:
I will re-investigate and check it out. We're also not passing the `lib` flag here, which makes me wonder if we're getting the correct coverage. I'll look at that too.

stringhandler (approved):
Decided to merge this and make an issue

PR #5223, opened 2023-03-07
fix(comms): dial if connection is not connected

sdbondi:
Description
---
Fix bug where recently disconnected connections do not get redialled

Motivation and Context
---
The connectivity manager pool maintains a list of connections including recently disconnected. These are cleared out periodically. The connectivity manager should not return disconnected connections to the caller.

How Has This Been Tested?
---
Manually: Connection is dialled when pool says the connection is not connected.

### DIFF
--- a/comms/core/src/connectivity/manager.rs
+++ b/comms/core/src/connectivity/manager.rs
@@ -315,22 +315,18 @@ impl ConnectivityManagerActor {
             },
         }
         match self.pool.get(&node_id) {
-            Some(state) => {
-                if !state.is_connected() {
-                    warn!(
-                        target: LOG_TARGET,
-                        "Existing connection is present but is_connected is false for some reason...."
-                    );
-                }
-
+            Some(state) if state.is_connected() => {
                 if let Some(reply_tx) = reply_tx {
                     let _result = reply_tx.send(Ok(state.connection().cloned().expect("Already checked")));
                 }
             },
-            None => {
+            maybe_state => {
                 info!(
                     target: LOG_TARGET,
-                    "No existing connection found for peer `{}`. Dialing...",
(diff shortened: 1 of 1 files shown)


stringhandler (dismissed):
lol so that's how that works.

stringhandler:
I changed it because it seemed to be a mistake. Perhaps a comment on the reasoning will prevent someone doing the same

comms/core/src/connectivity/manager.rs
```diff
@@ -315,22 +315,18 @@ impl ConnectivityManagerActor {
...
                 if let Some(reply_tx) = reply_tx {
                     let _result = reply_tx.send(Ok(state.connection().cloned().expect("Already checked")));
                 }
             },
-            None => {
+            maybe_state => {
                 info!(
                     target: LOG_TARGET,
-                    "No existing connection found for peer `{}`. Dialing...",
+                    "No connection found for peer (status={}) `{}`. Dialing...",
```
stringhandler:
```suggestion
                    "No connection found for peer (status={}) `{}`. It might not have existed or did exist but disconnected at some point. Dialling...",
```

sdbondi: Used `"Connection is not connected (status={}) `{}`. Dialing...",

stringhandler: maybe a comment then?

sdbondi:
There is no reason really at the moment. We could just remove disconnected connections. The thinking all those years ago (which could still apply) is you could have a UI that refreshes, showing recently disconnected/failed to dial connections. Currently we only have list-connections which only shows connected connections.

stringhandler (dismissed):

stringhandler (approved):

PR #5232, opened 2023-03-08
feat: add necessary trait bounds to balanced merkle tree

jorgeantonio21:
Description
---
Introduces necessary trait bounds to the `BalancedBinaryMerkleProof` struct.

Motivation and Context
---
In order to update the Tari DAN layer repo to make use of BMTs, we need that the current struct `BalancedBinaryMerkleProof` derives certain trait bounds, namely `#[derive(Clone, PartialEq, Eq, Deserialize, Serialize)]`.

How Has This Been Tested?
---
Run `cargo build`.


base_layer/mmr/src/balanced_binary_merkle_proof.rs
```diff
@@ -23,13 +23,14 @@
...
 use tari_utilities::ByteArray;
 use thiserror::Error;
 
 use crate::{common::hash_together, BalancedBinaryMerkleTree, Hash};
 
-#[derive(Debug)]
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Default)]
 pub struct BalancedBinaryMerkleProof<D> {
     pub path: Vec<Hash>,
     pub node_index: usize,
```
stringhandler: Change this u32 otherwise it'll differ between implementations

base_layer/mmr/src/balanced_binary_merkle_proof.rs
```diff
@@ -23,13 +23,14 @@
...
 use digest::Digest;
+use serde::{Deserialize, Serialize};
 use tari_common::DomainDigest;
 use tari_utilities::ByteArray;
 use thiserror::Error;
 
 use crate::{common::hash_together, BalancedBinaryMerkleTree, Hash};
 
-#[derive(Debug)]
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Default)]
```
stringhandler: `PartialEq` and `Eq` are fine, but why `Ord` and `PartialOrd`? 

Also think we'll need Borsh encording

jorgeantonio21: To be consistent with the `MerkleProof` type. Glad to remove it, as well. Regarding Borsh encoding, makes sense, will add it

AaronFeickert (changes requested):
The `as` casts can overflow.

base_layer/mmr/src/balanced_binary_merkle_tree.rs
```diff
@@ -102,7 +102,7 @@ where D: Digest + DomainDigest
             // The hash provided was not for leaf, but for node.
             Err(BalancedBinaryMerkleTreeError::LeafNotFound)
         } else {
-            Ok(pos - (self.hashes.len() >> 1))
+            Ok(pos as u32 - (self.hashes.len() as u32 >> 1))
```
AaronFeickert: This can still truncate.

AaronFeickert (changes requested):

base_layer/mmr/src/balanced_binary_merkle_tree.rs
jorgeantonio21: Resolved

AaronFeickert (approved):

AaronFeickert:
Looks like it fails `check` CI.

AaronFeickert (changes requested):
Fails CI.

base_layer/mmr/src/balanced_binary_merkle_proof.rs
```diff
@@ -81,6 +94,8 @@ pub enum MergedBalancedBinaryMerkleProofError {
     CantMergeZeroProofs,
     #[error("Bad proof semantics")]
     BadProofSemantics,
+    #[error("Math overflow")]
+    MathOverflow,
```
AaronFeickert, line 98: Perhaps consider renaming the entire `enum` to something like `BalancedBinaryMerkleProofError`, since right now a non-merged proof can return a merged proof error.

AaronFeickert (dismissed):
One small nit, otherwise looks good if CI agrees.

AaronFeickert (approved):
LGTM

sdbondi (approved):

PR #5233, opened 2023-03-08
ci: fail linting on truncated and wrapping casts

### LINKED ISSUE #5478: Linting for cast truncation and wrapping is disabled
Currently, the linter does not check for possible [truncation](https://rust-lang.github.io/rust-clippy/master/index.html#/cast_possible_truncation) or [wrapping](https://rust-lang.github.io/rust-clippy/master/index.html#/cast_possible_wrap) during casts. This could result in unwanted behavior.

These lints should be included and any occurrences mitigated or specifically allowed if safe.

AaronFeickert:
Description
---
Fails linting on possible truncated or wrapping casts. Various updates to address failures.

Closes #5478.

Motivation and Context
---
It was observed in [another PR](https://github.com/tari-project/tari/pull/5232) that truncated casts are not picked up by the linter. The `clippy::cast_possible_truncation` lint will catch this, but was disabled. Some of these casts may be risky, so it seems prudent to fail on them unless specifically allowed and documented.

This PR enables the lint as a `deny`, as well as a related lint for wrapping. For each lint failure, it either uses an `allow` (if the truncation is impossible or safe) or makes an update to fix it.

Note that after [PR 5237](https://github.com/tari-project/tari/pull/5237) is merged, it can be assumed that the target `usize` is not 32 bits.

How Has This Been Tested?
---
It hasn't. This behavior occurs in multiple places (around a couple dozen) within the codebase; in each case, it needs to be decided if the cast is safe (in which case the lint can be selectively disabled) or risky (in which case it should be replaced). This should happen before the PR is merged.

### DIFF
--- a/base_layer/common_types/src/emoji.rs
+++ b/base_layer/common_types/src/emoji.rs
@@ -97,7 +97,8 @@ lazy_static! {
     pub static ref REVERSE_EMOJI: HashMap<char, u8> = {
         let mut m = HashMap::with_capacity(DICT_SIZE);
         EMOJI.iter().enumerate().for_each(|(i, c)| {
-            m.insert(*c, i as u8);
+            #[allow(clippy::cast_possible_truncation)]
+            m.insert(*c, i as u8); // this can't truncate if the table contains at most 2^8 elements
         });
         m
     };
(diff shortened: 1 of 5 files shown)


AaronFeickert:
This PR is a draft until each instance of the lint behavior is addressed in the codebase. Seems reasonable to include them all here.

AaronFeickert:
This has become wildly out of date, and rebasing seems like an exercise in futility. Closing so that a newer, shinier, better PR can take its place.

PR #5234, opened 2023-03-09
fix: windows path format in log4rs files

brianp:
Description
---
This commit does a naive character replacement on the string path just before we do the path injection.

Motivation and Context
---
log4rs expects the paths to be in a unix format regardless of the system it's running on. When we started doing the dynamic injecting of the base path windows systems injected using a windows format, which mixed with the unix format. Resulting in strings that looked liked this: `C:\\brianp\.tari\config/log/base-node/network.log`

when log4rs wants the final format to be like this: `C://brianp/.tari/config/log/base-node/network.log`

How Has This Been Tested?
---
Manually on Hansie's windows machine

Co-authored-by: hansieodendaal 39146854+hansieodendaal@users.noreply.github.com


sdbondi (approved):
hah weird - LGTM

stringhandler (approved):

PR #5235, opened 2023-03-09
fix(wallet): correct change checks in transaction builder

sdbondi:
Description
---
Fix mislabelled transaction builder checks for change params

Motivation and Context
---
Previous code just checked the private nonce 3 times
```rust
        Self::check_value("Change script", &self.private_nonce, &mut message);
        Self::check_value("Change input data", &self.private_nonce, &mut message);
        Self::check_value("Change script private key", &self.private_nonce, &mut message);
```

None of the change builder fields are required.

How Has This Been Tested?
---
N/A Change has no effect on builder behaviour


SWvheerden (dismissed):
utACK

PR #5237, opened 2023-03-09
fix: immediately fail to compile on 32-bit systems

sdbondi:
Description
---
Fail to compile on 32-bit architectures

Motivation and Context
---
32-bit architectures are untested. Depending on the application, it may not compile already or there could be various classes of bugs (overflows, crashes, etc). This PR explicitly fails compilation to 32-bit targets for all applications.

How Has This Been Tested?
---
CI

### DIFF
--- a/applications/tari_app_utilities/src/lib.rs
+++ b/applications/tari_app_utilities/src/lib.rs
@@ -35,6 +35,6 @@ pub mod consts {
 macro_rules! deny_non_64_bit_archs {
     () => {
         #[cfg(not(target_pointer_width = "64"))]
-        compile_error!("Only 64-bit architectures are supported");
+        compile_error!("Only 64-bit architectures are supported. Here's a medal for trying 🏅");
     };
 }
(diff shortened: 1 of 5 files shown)


AaronFeickert:
I tried building on a 32-bit Xubuntu 18.04 VM with this PR applied. The build failed, but on a build script apparently related to a Tor library. It appears that at least some of the application libraries did build, which is not expected behavior.

AaronFeickert (dismissed):
While testing a full build on a 32-bit Xubuntu 18.04 VM failed for different reasons, attempting to compile the individual libraries from this PR fails on each with the expected message.

AaronFeickert:
Would it be more prudent to fail builds on anything _not_ 64 bit, rather than only on 32 bit?

CjS77:
If you get rustc running on an Apple ][ You deserve a fricken medal, not an error message!

sdbondi:
> Would it be more prudent to fail builds on anything not 64 bit, rather than only on 32 bit?

Makes sense, I created a macro so that if we want to change the message/conditional we can do that in one place.
I think explicitly calling the macro is preferable to failing compilation in tari_app_utilities directly (can enable selectively, might be unexpected, explicitness).

Also added a medal 🤣

SWvheerden (approved):
utACK

SWvheerden:
I think for now this is the correct approach, limit the end-user applications. 
We have to be careful about just blocking 32-bit on all core architecture as for example the hardware wallets may need to access those and they are 32-bit only.

AaronFeickert (approved):

PR #5242, opened 2023-03-13
feat: add other code template types

stringhandler:
Description
---
Add flow and manifest code template types 

Motivation and Context
---
Composing templates into a single wasm is not always practical. This change allows users to wire together multiple templates using a flow or manifest

How Has This Been Tested?
---
locally


SWvheerden (approved):
LGTM

PR #5244, opened 2023-03-13
feat: chat scaffold

brianp:
Description
---
- Refactor the contacts crate for better organization.
- Expand contacts with types, and capabilities for managing messaging over the existing comms network
- Add cucumber tests to provide support for new functionality

Motivation and Context
---
We want to expand Tari's communication network to handle p2p chat services. This PR sets us off in the right direction for the core Tari processes to be able to handle message passing and storage. Currently chat clients will be able to send, and receive messages when online or offline. Utilizing store and forward for sending communications offline.

How Has This Been Tested?
---
Via cucumber only. The chat client within cucumber is currently the only interface for trials.

Caveats and known issues
---
- [ ] Encryption at rest: We need to change the message storage to support encryption at rest

### DIFF
--- a/base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
+++ b/base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
@@ -1,9 +1,9 @@
 CREATE TABLE messages (
-    address    BLOB                NOT NULL,
-    message_id INTEGER PRIMARY KEY NOT NULL,
-    body       BLOB                NOT NULL,
-    stored_at  TIMESTAMP           NOT NULL,
-    direction  INTEGER             NOT NULL
+    address    BLOB             NOT NULL,
+    message_id BLOB PRIMARY KEY NOT NULL,
+    body       BLOB             NOT NULL,
+    stored_at  TIMESTAMP        NOT NULL,
+    direction  INTEGER          NOT NULL
 );
 
 CREATE INDEX idx_messages_address ON messages (address);
--- a/base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
+++ b/base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
@@ -1,0 +1,9 @@
+CREATE TABLE messages (
+    address    BLOB                NOT NULL,
+    message_id INTEGER PRIMARY KEY NOT NULL,
+    body       BLOB                NOT NULL,
+    stored_at  TIMESTAMP           NOT NULL,
+    direction  INTEGER             NOT NULL
+);
+
+CREATE INDEX idx_messages_address ON messages (address);
--- a/base_layer/contacts/src/contacts_service/handle.rs
+++ b/base_layer/contacts/src/contacts_service/handle.rs
@@ -35,7 +35,7 @@ use tower::Service;
 use crate::contacts_service::{
     error::ContactsServiceError,
     service::{ContactMessageType, ContactOnlineStatus},
-    storage::database::Contact,
+    types::{Contact, Message},
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -130,6 +130,8 @@ pub enum ContactsServiceRequest {
     RemoveContact(TariAddress),
     GetContacts,
     GetContactOnlineStatus(Contact),
+    SendMessage(TariAddress, Message),
+    GetMessages(TariAddress, u64, u64),
 }
 
 #[derive(Debug)]
@@ -139,6 +141,8 @@ pub enum ContactsServiceResponse {
     Contact(Contact),
     Contacts(Vec<Contact>),
     OnlineStatus(ContactOnlineStatus),
+    Messages(Vec<Message>),
+    MessageSent,
 }
 
 #[derive(Clone)]
@@ -221,6 +225,33 @@ impl ContactsServiceHandle {
             .await??
         {
             ContactsServiceResponse::OnlineStatus(status) => Ok(status),
+            _ => Err(ContactsServiceError::UnexpectedApiResponse),
+        }
+    }
+
+    pub async fn get_messages(
+        &mut self,
+        pk: TariAddress,
+        x: u64,
+        y: u64,
+    ) -> Result<Vec<Message>, ContactsServiceError> {
+        match self
+            .request_response_service
+            .call(ContactsServiceRequest::GetMessages(pk, x, y))
+            .await??
+        {
+            ContactsServiceResponse::Messages(messages) => Ok(messages),
+            _ => Err(ContactsServiceError::UnexpectedApiResponse),
+        }
+    }
+
+    pub async fn send_message(&mut self, message: Message) -> Result<(), ContactsServiceError> {
+        match self
+            .request_response_service
+            .call(ContactsServiceRequest::SendMessage(message.address.clone(), message))
+            .await??
+        {
+            ContactsServiceResponse::MessageSent => Ok(()),
             _ => Err(ContactsServiceError::UnexpectedApiResponse),
         }
     }
--- a/base_layer/contacts/src/contacts_service/handle.rs
+++ b/base_layer/contacts/src/contacts_service/handle.rs
@@ -131,7 +131,7 @@ pub enum ContactsServiceRequest {
     GetContacts,
     GetContactOnlineStatus(Contact),
     SendMessage(TariAddress, Message),
-    GetMessages(TariAddress, u64, u64),
+    GetAllMessages(TariAddress),
 }
 
 #[derive(Debug)]
@@ -229,15 +229,10 @@ impl ContactsServiceHandle {
         }
     }
 
-    pub async fn get_messages(
-        &mut self,
-        pk: TariAddress,
-        x: u64,
-        y: u64,
-    ) -> Result<Vec<Message>, ContactsServiceError> {
+    pub async fn get_all_messages(&mut self, pk: TariAddress) -> Result<Vec<Message>, ContactsServiceError> {
         match self
             .request_response_service
-            .call(ContactsServiceRequest::GetMessages(pk, x, y))
+            .call(ContactsServiceRequest::GetAllMessages(pk))
             .await??
         {
             ContactsServiceResponse::Messages(messages) => Ok(messages),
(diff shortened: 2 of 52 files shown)


integration_tests/tests/features/Chat.feature
```diff
@@ -0,0 +1,43 @@
+# Copyright 2023 The Tari Project
+# SPDX-License-Identifier: BSD-3-Clause
+
+Feature: Chat messaging
+
+  Scenario: A message is propagated between nodes
```
hansieodendaal:
```suggestion
  Scenario: A message is propagated between nodes via a 3rd party base node
```

integration_tests/tests/features/Chat.feature
```diff
@@ -0,0 +1,43 @@
...
+    Given I have a seed node SEED_A
+    When I have a chat client CHAT_A connected to seed node SEED_A
+    When I have a chat client CHAT_B connected to seed node SEED_A
+    When I use CHAT_A to send a message 'Hey there' to CHAT_B
+    Then CHAT_B will have 1 message with CHAT_A
+
+  Scenario: A message is sent directly between nodes
+    Given I have a seed node SEED_A
+    When I have a chat client CHAT_A connected to seed node SEED_A
+    When I have a chat client CHAT_B connected to seed node SEED_A
```
hansieodendaal, line 16: Is it still necessary to have the 3rd party base node here?

integration_tests/tests/features/Chat.feature
```diff
@@ -0,0 +1,43 @@
...
+    When I add CHAT_B as a contact to CHAT_A
+    When CHAT_A waits for contact CHAT_B to be online
```
hansieodendaal, line 19: If the base node is required to establish comms between the chat clients, then it should be stopped after the chat clients' online status is confirmed.

integration_tests/tests/features/Chat.feature
```diff
@@ -0,0 +1,43 @@
...
+    When I use CHAT_A to send a message 'Hey there' to CHAT_B
+    Then CHAT_B will have 1 message with CHAT_A
+
+  Scenario: Message counts are distinct
+    Given I have a seed node SEED_A
+    When I have a chat client CHAT_A connected to seed node SEED_A
+    When I have a chat client CHAT_B connected to seed node SEED_A
+    When I have a chat client CHAT_C connected to seed node SEED_A
```
hansieodendaal, line 27: The preferred test mode should be without the base node running

hansieodendaal (commented):
Looking good, I just have some comments about the cucumber scenarios.

integration_tests/tests/features/Chat.feature
brianp: Without the seed node the two chat clients don't seem to find each other after being added as contacts. We would probably need to introduce a peer dialing step for new contacts if the client doesn't otherwise have connectivity.

integration_tests/tests/features/Chat.feature
brianp: Yes, stopping the base node and having the clients communicate does work, I'll update the test.

integration_tests/tests/features/Chat.feature
hansieodendaal: Maybe we can register an issue to monitor if this will be a problem for the mobile wallets, as they will most probably not connect to the same base node at startup.

applications/tari_base_node/log4rs_sample.yml
```diff
@@ -124,6 +124,11 @@ loggers:
     level: debug
     appenders:
       - network
+  # Route log events sent to the "contacts" logger to the "network" appender
+  contacts:
```
SWvheerden: we dont have contact events in the base_node right?

base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE messages (
+    address    BLOB                NOT NULL,
+    message_id INTEGER PRIMARY KEY NOT NULL,
```
SWvheerden: rather make this bytes, and create a uuid.

base_layer/contacts/src/contacts_service/handle.rs
```diff
@@ -224,4 +228,31 @@ impl ContactsServiceHandle {
             _ => Err(ContactsServiceError::UnexpectedApiResponse),
         }
     }
+
+    pub async fn get_messages(
+        &mut self,
+        pk: TariAddress,
+        x: u64,
+        y: u64,
```
SWvheerden: Can you add docs here, and or change the x or y variable names? It's unclear what they do. I am assuming it's paging?

base_layer/contacts/src/contacts_service/service.rs
```diff
@@ -232,6 +266,47 @@ where T: ContactsBackend + 'static
                 let result = self.get_online_status(&contact).await;
                 Ok(result.map(ContactsServiceResponse::OnlineStatus)?)
             },
+            ContactsServiceRequest::GetMessages(pk, _x, _y) => {
+                let result = self.db.get_messages(pk);
```
SWvheerden: I think either drop x and y now and make it get_all_messages
or we choose a type for x,y  and impl
I don't like us having some types un implemented here.

base_layer/contacts/src/contacts_service/storage/database.rs
```diff
@@ -30,39 +30,13 @@ use log::*;
...
-            address,
-            last_seen,
-            latency,
-            favourite,
-        }
-    }
-}
-
 /// This trait defines the functionality that a database backend need to provide for the Contacts Service
 pub trait ContactsBackend: Send + Sync + Clone {
```
SWvheerden, line 41: this is fine for now, but see: #5320 
This design has been copied over from our badly designed first backend on the base_node which uses a key-value store and not a relational db.

base_layer/contacts/src/contacts_service/storage/types/contacts.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /// Find a particular Contact by their node ID, and delete it if it exists, returning the affected record
+    pub fn find_by_node_id_and_delete(
+        conn: &mut SqliteConnection,
+        node_id: &[u8],
+    ) -> Result<ContactSql, ContactsServiceStorageError> {
+        // Note: `get_result` not implemented for SQLite
+        let contact = ContactSql::find_by_node_id(node_id, conn)?;
+        if diesel::delete(contacts::table.filter(contacts::node_id.eq(node_id))).execute(conn)? == 0 {
```
SWvheerden: You cant use `.num_rows_affected_or_not_found(1)?;` like above?

base_layer/contacts/src/contacts_service/storage/types/contacts.rs
```diff
@@ -0,0 +1,184 @@
...
+    contacts_service::{error::ContactsServiceStorageError, types::Contact},
+    schema::contacts,
+};
+
+/// A Sql version of the Contact struct
+#[derive(Clone, Debug, Queryable, Insertable, PartialEq, Eq)]
+#[diesel(table_name = contacts)]
+pub struct ContactSql {
+    pub address: Vec<u8>,
+    node_id: Vec<u8>,
```
SWvheerden, line 42: Whats the reason we have both here?
address = [node_id][CRC + network]
Asking the node id from address is a simple slice read

base_layer/contacts/src/contacts_service/types/contact.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use chrono::NaiveDateTime;
+use tari_common_types::tari_address::TariAddress;
+use tari_comms::peer_manager::NodeId;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Contact {
+    pub alias: String,
+    pub address: TariAddress,
+    pub node_id: NodeId,
```
SWvheerden, line 31: I asked above, but same here, why store both.

comms/dht/src/outbound/requester.rs
```diff
@@ -52,6 +52,33 @@ impl OutboundMessageRequester {
 
     /// Send directly to a peer. If the peer does not exist in the peer list, a discovery will be initiated.
     pub async fn send_direct<T>(
```
SWvheerden:
```suggestion
    pub async fn send_direct_encrypted<T>(
```
We are changing the behaviour of an already existing pub API, I would force an error here to force users to look which one to use.

integration_tests/log4rs/wallet.yml
```diff
@@ -93,6 +93,11 @@ loggers:
     appenders:
       - network
     additive: false
+  contacts:
```
SWvheerden, line 113: We need to add this to the default log of the wallet as well. 
But I would also use unique appended and not `- network`

SWvheerden (commented):
This is looking good

applications/tari_base_node/log4rs_sample.yml
brianp: We shouldn't no. I didn't know how specific each setup file was supposed to be. Shall I take this out?

base_layer/contacts/migrations/2023-03-14-101758_create_messages/up.sql
brianp: Yes. Totally forgot about that convo once I was in the weeds. UUID makes a lot more sense from a debugging and linking perspective.

base_layer/contacts/src/contacts_service/service.rs
brianp: i'll pull them out for now, we can implement paging later.

base_layer/contacts/src/contacts_service/storage/types/contacts.rs
brianp: This code was moved over from its previous location in the refactor not written new. I can try out the recommendation though.

base_layer/contacts/src/contacts_service/types/contact.rs
brianp: Ahh, not sure. This code was moved in the refactor, not written new for this PR.

comms/dht/src/outbound/requester.rs
brianp: Like have three functions:
- `send_direct_encrypted`
- `send_driect_unencrypted`

And then `send_direct` would now return an error?

SWvheerden (approved):

SWvheerden:
Ack

PR #5245, opened 2023-03-14
fix(wallet)!: ensure burn shared keys and hashes match dan layer

sdbondi:
Description
---
- derives encryption key using DH shared spend key
- use consensus hasher to ensure consistent hash preimages
- recovered blinding factor is the shared key, so logic updated not to calculate another Diffie-Helman from the claim key

Motivation and Context
---
Mask and encrypted value key derivation currently should match confidential claims and transfers on the DAN layer.

```
shared_key = DH(claim_pubkey, spend_key) 
shared_mask = dan_mask_kdf(shared_key)
commitment = commit(shared_mask, value)
shared_value_encryption_key = dan_encrypted_value_kdf(shared_mask, commitment)
encrypted_value = encrypt(shared_value_encryption_key, value)
```

How Has This Been Tested?
---
Tested that claims work on the DAN

BREAKING CHANGE: Burn with claim mask and encrypted value key derivation has changed. Existing burnt funds are not claimable on the DAN.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
+++ b/base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
@@ -52,7 +52,7 @@ use crate::{
             OutputSource,
         },
     },
-    util::burn_proof::derive_diffie_hellman_burn_claim_encryption_key,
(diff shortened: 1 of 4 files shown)


SWvheerden (approved):
utACK

Issue #5227, opened 2023-03-07
wallet shows rejected tx, but balance still reflects transaction


SWvheerden:
The detail tx view shows the transaction as rejected. But the balance still shows the available balance as including that transaction
<img width="876" alt="Screenshot 2023-03-07 at 19 17 12" src="https://user-images.githubusercontent.com/2579813/223498947-7e26c406-b339-44ef-9dcf-9b9d3dc6ddd2.png">

SWvheerden:
[wallet.zip](https://github.com/tari-project/tari/files/10912121/wallet.zip)

SWvheerden:
[wallet_log.zip](https://github.com/tari-project/tari/files/10912208/wallet_log.zip)

PR #5246, opened 2023-03-14
fix: added transaction revalidation to the wallet startup sequence #5227

agubarev:
Description
---
Now wallet revalidates all transactions included in the wallet on startup. 

https://github.com/tari-project/tari/issues/5227

Motivation and Context
---
Outputs fetched from a fully synced node could be invalidated by the transaction validator, e.g. wallet has normal outputs,
fully validated, but at some point become invalidated when the wallet is trying to negotiate with another, not yet fully synced node that doesn't recognise said output due to not reaching its height at the moment.

So, the code was already there but not yet used in the startup process.

How Has This Been Tested?
---
manually


SWvheerden (approved):
utACK

PR #5254, opened 2023-03-16
ci(fix): add TestNet for tagged releases

leet4tari:
Description
Add TestNet for tagged releases for binaries and package builds

Motivation and Context
Split out the different networks when build on GitHub

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -86,6 +86,24 @@ jobs:
         with:
           submodules: recursive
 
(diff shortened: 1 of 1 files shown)


brianp (approved):
Cool, so like we discussed yesterday, only when tags are pushed we assign the env var which will get used for the rest of the workflow, compiling the bins. 👍🏻

brianp:
UtAck

SWvheerden (approved):

sdbondi (approved):

stringhandler (approved):

Issue #5109, opened 2023-01-12
Wallet FFI has no indication if the base node its connected is healthy or not


SWvheerden:
Currently, the wallet FFI has no information about the base node its connected to apart from if its online/offline/connecting. 
We should add the base node status to a callback so we can track it for the wallet like we do for the console wallet. 
![Screenshot 2023-01-12 at 15 40 14](https://user-images.githubusercontent.com/2579813/212081764-5faa7335-3b34-4750-a996-6d67f7c4be6f.png)

PR #5257, opened 2023-03-17
feat: added auxiliary callback to push base node state changes #5109

agubarev:
Description
---
Extra function argument to `wallet_create` FFI.

https://github.com/tari-project/tari/issues/5109

Motivation and Context
---
Giving mobile app more information about the active base node connection.

How Has This Been Tested?
---

### DIFF
--- a/base_layer/wallet_ffi/src/callback_handler.rs
+++ b/base_layer/wallet_ffi/src/callback_handler.rs
@@ -42,7 +42,6 @@ use tari_common_types::{tari_address::TariAddress, transaction::TxId, types::Blo
 use tari_comms_dht::event::{DhtEvent, DhtEventReceiver};
 use tari_contacts::contacts_service::handle::{ContactsLivenessData, ContactsLivenessEvent};
 use tari_shutdown::ShutdownSignal;
-use tari_utilities::ByteArray;
 use tari_wallet::{
     base_node_service::{
         handle::{BaseNodeEvent, BaseNodeEventReceiver},
@@ -63,7 +62,7 @@ use tari_wallet::{
 };
 use tokio::sync::{broadcast, watch};
 
-use crate::{ByteVector, TariBaseNodeState};
+use crate::ffi_basenode_state::TariBaseNodeState;
 
 const LOG_TARGET: &str = "wallet::transaction_service::callback_handler";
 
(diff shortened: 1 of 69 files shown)


base_layer/wallet_ffi/src/callback_handler_tests.rs
```diff
@@ -466,14 +478,31 @@ mod test {
...
+        let start = Instant::now();
+        while start.elapsed().as_secs() < 10 {
+            let lock = CALLBACK_STATE.lock().unwrap();
+
+            if lock.base_node_state_changed_callback_invoked {
+                break;
+            }
+
+            thread::sleep(Duration::from_millis(100));
+        }
```
SWvheerden, line 522: You are never checking that you actually received one?
This test will still pass if this event is never reached.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -135,6 +135,7 @@ use tari_utilities::{
     SafePassword,
 };
 use tari_wallet::{
+    base_node_service::service::BaseNodeState,
```
SWvheerden: We need to create a `pub type TariBaseNodeState = base_node_service::service::BaseNodeState;` for the .h file
Then we need to create accessor function for the FFI layer to be able to access the data inside if the pointer. Look at the other types.

SWvheerden (commented):
This looks good so far. Just need to expand the pointer to include accessor functions for the data.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -244,6 +244,40 @@ pub struct TariWallet {
     shutdown: Shutdown,
 }
 
+#[derive(Debug)]
+#[repr(C)]
```
SWvheerden: Look at the c code in the .h file we need accessor functions for the node id handling the `Option<TariNodeId>`
And the u128 is not support in C, its only fully supported in GCC c23 which had its draft published in Jan 2023.

SWvheerden (commented):
looking on the right track here

base_layer/wallet_ffi/src/callback_handler.rs
```diff
@@ -613,4 +648,38 @@ where TBackend: TransactionBackend + 'static
...
+            Some(chain_metadata) => TariBaseNodeState {
+                node_id: state.node_id,
+                height_of_longest_chain: chain_metadata.height_of_longest_chain(),
+                best_block: *chain_metadata.best_block(),
+                best_block_timestamp: chain_metadata.timestamp(),
+                pruning_horizon: chain_metadata.pruning_horizon(),
+                pruned_height: chain_metadata.pruned_height(),
+                is_node_synced: state.is_synced.unwrap_or(false),
+                updated_at: state.updated.map(|ts| ts.timestamp_millis() as u64).unwrap_or(0),
+                latency: state.latency.map(|d| d.as_micros() as u64).unwrap_or(0),
```
SWvheerden:
```suggestion
                latency: state.latency.map(|d| d.as_millis() as u64).unwrap_or(0),
```

SWvheerden: returning one as `ms`, and one as `us `is not constant. We don't need `us` latency.

SWvheerden (approved):

Issue #5261, opened 2023-03-21
CI - Fix wasm pack install


SWvheerden:
Since commit https://github.com/tari-project/tari/pull/5258/commits/1c65bc7f098a0b6474f2df237772a1a281042e25
Wasm pack has been broken for CI
This needs to be fixed again and re-added. 

Fails with: 
```
found 0 vulnerabilities
info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
info: installing component 'rust-std' for 'wasm32-unknown-unknown'
rustup run nightly-[20](https://github.com/tari-project/tari/actions/runs/4467454060/jobs/7847682092#step:5:21)[22](https://github.com/tari-project/tari/actions/runs/4467454060/jobs/7847682092#step:5:23)-11-03 wasm-pack test --node --features wasm
error: command failed: 'wasm-pack': No such file or directory (os error 2)
make: *** [test] Error 1
Makefile:5: recipe for target 'test' failed
Error: Process completed with exit code 2.
```

AaronFeickert:
This looks to be an upstream issue with `wasm-pack@0.11.0`. Freezing at `wasm-pack@0.10.3` is at least a temporary fix.

Opened an [upstream issue](https://github.com/rustwasm/wasm-pack/issues/1248) for this already.

PR #5262, opened 2023-03-21
test: revert and freeze `wasm-pack` module

AaronFeickert:
Description
---
Reverts and freezes `wasm-pack@0.10.3` due to an apparent upstream issue.

Fixes (at least temporarily) [issue 5261](https://github.com/tari-project/tari/issues/5261).

Motivation and Context
---
There's an apparent issue with the `wasm-pack` Node.js module at the 0.11.0 version that breaks CI operations (and which [also affects](https://github.com/tari-project/tari-crypto/pull/174) `tari-crypto`).

While an [earlier PR](https://github.com/tari-project/tari/pull/5258) disabled this CI, it seems unnecessary and somewhat risky not to have these tests.

Reverting to the 0.10.3 version is at least a temporary fix.

How Has This Been Tested?
---
With the fix, CI should pass.

---
Run the equivalent commands locally, and ensure that CI passes with the change.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,12 +140,12 @@ jobs:
         run: |
           sudo apt-get update
           sudo bash scripts/install_ubuntu_dependencies.sh
-      #- name: test key manager wasm
-      #  run: |
-      #    npm install -g wasm-pack
-      #    cd base_layer/key_manager
-      #    rustup target add wasm32-unknown-unknown
-      #    make test
+      - name: test key manager wasm
+        run: |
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This is now an [upstream issue](https://github.com/rustwasm/wasm-pack/issues/1248).

SWvheerden (approved):
Thanks for taking the time to look into this and provide a fix

Issue #5252, opened 2023-03-16
[Wallet]Add last ran wallet version to wallet database. 


SWvheerden:
Add the last ran wallet version to the table `WalletSettingSql`. 
Add the network the wallet is run on on this as well and check it at startup that it matches. 

The version should be retrieved from the FFI without initializing the wallet via `walet_create`

It should not run any migration on the database etc. It should just try and retrieve the version+ network. If this is not found, it should return None/Null

stringhandler:
There should be a different database per network, so no need to include the network

stringhandler:
although I suppose it couldn't hurt

SWvheerden:
Yeah should be as per: #5250 
But I think it would better if we check this.

stringhandler:
You can already see this version from the `__diesel_schema_migrations` table

SWvheerden:
Fixed in #5263

PR #5263, opened 2023-03-22
feat: added FFI function `wallet_get_network_and_version` #5252

agubarev:
Description
---
- Network and client APP_VERSION_NUMBER is now stored when wallet is initialized.
- Added FFI function `wallet_get_network_and_version` that returns the `network` and `version` of the client that has last accessed the wallet database.

Motivation and Context
---
For debugging purposes on the client (e.g. Mobile)

How Has This Been Tested?
---
manually

### DIFF
--- a/base_layer/wallet/src/storage/sqlite_utilities/mod.rs
+++ b/base_layer/wallet/src/storage/sqlite_utilities/mod.rs
@@ -146,7 +146,7 @@ pub fn initialize_sqlite_database_backends<P: AsRef<Path>>(
     ))
 }
 
-pub fn get_last_network_and_version<P: AsRef<Path>>(db_path: P) -> Result<(String, String), WalletStorageError> {
+pub fn get_last_version<P: AsRef<Path>>(db_path: P) -> Result<Option<String>, WalletStorageError> {
     let path_str = db_path
         .as_ref()
         .to_str()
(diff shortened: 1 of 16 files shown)


base_layer/wallet/src/storage/sqlite_utilities/mod.rs
```diff
@@ -142,3 +145,25 @@ pub fn initialize_sqlite_database_backends<P: AsRef<Path>>(
         key_manager_backend,
     ))
 }
+
+pub fn get_last_network_and_version<P: AsRef<Path>>(db_path: P) -> Result<(String, String), WalletStorageError> {
```
SWvheerden: Split this into `get_last_network`, and `get_last_version `and rather than return a "n/a"string, return option<String>
By returning a "n/a"string we need to do string processing in order to find out it does not have a stored string.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5447,6 +5463,69 @@ pub unsafe extern "C" fn wallet_create(
...
+/// as an out parameter.
+/// ## Returns
+/// `*mut TariNetworkAndVersion` - Returns the pointer to the TariNetworkAndVersion that contains the network and
+/// version
+///
+/// # Safety
+/// The ```network_and_version_destroy``` method must be called when finished with a TariNetworkAndVersion to prevent a
+/// memory leak
+#[no_mangle]
+pub unsafe extern "C" fn wallet_get_network_and_version(
```
SWvheerden: This is making this way more complicated, as above, if we make it 2 calls, we can only make this, get version. 
And then return a *char. We need need to create and destroy a custom mapped struct.

base_layer/wallet/src/wallet.rs
```diff
@@ -156,6 +156,7 @@ where
         key_manager_backend: X,
         shutdown_signal: ShutdownSignal,
         master_seed: CipherSeed,
+        app_metadata: Option<AppMetadata>,
```
SWvheerden: Don't make this an option that's fed in from outside. 
Let the wallet ask its own version with `const::APP_VERSION_NUMBER`

SWvheerden (commented):
Looks good so far.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5447,6 +5447,90 @@ pub unsafe extern "C" fn wallet_create(
...
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+
+    let sql_database_path = (*config)
+        .datastore_path
+        .join((*config).peer_database_name.clone())
+        .with_extension("sqlite3");
+
+    debug!(target: LOG_TARGET, "Obtaining the version of the last client run");
```
SWvheerden:
```suggestion
```
logging wont be initialised yet at this point,  so we don't have to log this,

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5447,6 +5447,90 @@ pub unsafe extern "C" fn wallet_create(
...
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+
+    let sql_database_path = (*config)
+        .datastore_path
+        .join((*config).peer_database_name.clone())
+        .with_extension("sqlite3");
+
+    debug!(target: LOG_TARGET, "Obtaining the network of the last client run");
```
SWvheerden:
```suggestion
```
Same as above

SWvheerden (dismissed):
ACK

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5447,6 +5447,88 @@ pub unsafe extern "C" fn wallet_create(
...
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+
+    let sql_database_path = (*config)
+        .datastore_path
+        .join((*config).peer_database_name.clone())
+        .with_extension("sqlite3");
+
+
```
SWvheerden:
```suggestion

```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5447,6 +5447,88 @@ pub unsafe extern "C" fn wallet_create(
...
```
SWvheerden:
```suggestion

```

SWvheerden (approved):

PR #5265, opened 2023-03-23
docs: add burn docs warning

SWvheerden:
Description
---
Adds a warning to the burn outputs features to be wary of this. The primary goal of burned commitments is to be used to reclaim second-layer funds. Claiming on the second layer requires extra features attached to the kernel.  But we don't want to limit burns to only the second layer as there exist many use cases for burn outputs. 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


CjS77:
ACK

CjS77 (approved):

PR #5266, opened 2023-03-24
ci(fix): update miner filename

leet4tari:
Description
update miner filename so release has unique filename per arch

Motivation and Context
currently release has only two miner files, which is broken as we should have a file per arch.

How Has This Been Tested?
---
Built in local fork

---
Look at artifacts and see miner filenames match artifacts.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -376,10 +376,11 @@ jobs:
       - name: Prep miner for upload
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):
utACK

SWvheerden:
ACK

SWvheerden:
ACK

Issue #5268, opened 2023-03-24
[CI] Investigate CI test failure


SWvheerden:
Test failed on CI, but succeeded locally. 
```
running 4 tests
test pruned\_mmr\_empty ... ok
test calculate\_mmr\_roots ... ok
test calculate\_pruned\_mmr\_roots ... FAILED
)test pruned\_mmrs ... ok



failures:



\---- calculate\_pruned\_mmr\_roots stdout ----

thread 'calculate\_pruned\_mmr\_roots' panicked at 'assertion failed: \`(left == right)\`

 left: \`\[115, 144, 67, 185, 145, 204, 211, 15, 120, 144, 121, 221, 175, 142, 107, 36, 74, 19, 201, 139, 110, 18, 8, 196, 15, 81, 200, 31, 111, 171, 16, 224\]\`,

right: \`\[38, 169, 21, 216, 240, 207, 113, 15, 40, 126, 254, 144, 70, 84, 246, 242, 162, 255, 254, 8, 152, 108, 63, 27, 186, 27, 67, 76, 55, 173, 113, 200\]\`', base\_layer/mmr/tests/pruned\_mmr.rs:104:5

note: run with \`RUST\_BACKTRACE=1\` environment variable to display a backtrace
```

AaronFeickert:
I was able to trigger local test failures by running the test on a loop. It's not a deterministic failure, so it must occur only on certain randomized parameters.

AaronFeickert:
The cause appears to be specific to the deletion functionality of the underlying mutable MMR implementation. If you use an empty deletion index vector in the test, it will always succeed.

AaronFeickert:
Looks like this is in fact a test failure. [The test in question](https://github.com/tari-project/tari/blob/e334a404e432b0911bae3054a28d8e8ca5876e6c/base_layer/mmr/tests/pruned_mmr.rs#L86-L105) fails to compress the deletion bitmap before computing a Merkle root.

Depending on implementation details, it may be useful to move the compression directly into the `get_merkle_root` functionality if feasible, to avoid the case where a caller forgets to compress and results in a possibly different root.

AaronFeickert:
Addressed by [PR 5278](https://github.com/tari-project/tari/pull/5278).

PR #5278, opened 2023-03-28
fix: ensures mutable MMR bitmaps are compressed

### LINKED ISSUE #5277: Check that all mutable MMRs properly compress deletion bitmaps
A recent [test failure](https://github.com/tari-project/tari/issues/5268) uncovered that failure to compress the deletion bitmap of a mutable MMR can result in a different root than if compression is done. Further, this failure may not always be detected since compression isn't guaranteed to change the resulting bitmap hash.

There are several critical places in the codebase where mutable MMR roots are computed, and it's not clear if compression is being done properly in all cases. If no deletions are performed, it's not necessary to compress. If deletions are performed, it is necessary.

There seem to be two approaches that could be taken to address this:
- Manually check that deletion bitmap compression is done as necessary by all callers.
- Add a compression step to the MMR root computation functionality, so the caller doesn't need to do this.

The second approach is presumably less efficient, but is much safer.

AaronFeickert:
Description
---
Ensures that mutable MMR root computation first performs deletion bitmap compression.

Closes [issue 5277](https://github.com/tari-project/tari/issues/5277).

Motivation and Context
---
Currently, when mutable MMR roots [are computed](https://github.com/tari-project/tari/blob/e334a404e432b0911bae3054a28d8e8ca5876e6c/base_layer/mmr/src/mutable_mmr.rs#L119-L131), it's implicitly assumed that the underlying deletion bitmap has been compressed. If it has not, it's possible for the resulting root to be different than if the bitmap were compressed. This already resulted in an intermittent [test failure](https://github.com/tari-project/tari/issues/5268).

To reduce the risk that a caller does not perform this compression correctly, this PR adds compression to the Merkle root computation functionality directly. It also removes a few compression calls that become redundant with this change.

Note that this may constitute an efficiency regression. If a mutable MMR does not have any deletions since its last bitmap compression, subsequent compression is not necessary. Further, we perform the compression on a clone of the bitmap; this is necessary to avoid mutability and ensure that operations like equality (which rely on root computation) function properly. The efficiency consequences of this change should be examined to ensure they are acceptable.

How Has This Been Tested?
---
Existing CI passes.

---
Run existing CI. To further check for intermittent test failures, loop affected tests.

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

It may be the case that this change affects the way mutable MMR roots are computed in certain cases, which in turn may be a breaking change.

### DIFF
--- a/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
@@ -606,7 +606,6 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                     // in the output MMR
                     let bitmap = self.full_bitmap_mut();
                     bitmap.or_inplace(&diff_bitmap);
-                    bitmap.run_optimize();
 
                     let pruned_output_set = output_mmr.get_pruned_hash_set()?;
                     let output_mmr = MutablePrunedOutputMmr::new(pruned_output_set.clone(), bitmap.clone())?;
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1360,8 +1360,6 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
         }
     }
 
-    output_mmr.compress();
-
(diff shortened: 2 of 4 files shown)


stringhandler (approved):
Nice

PR #5279, opened 2023-03-28
feat: improved passphrase flow

### LINKED ISSUE #5127: Wallet bad password just continues on
When a user selects a weak password:
```
>> 1
Create wallet password: 
Confirm wallet password: 
Passwords match.
However, the password you chose is weak; a determined attacker with access to your device may be able to guess it.
You may want to consider changing it to a stronger one.
Here are some suggestions:
- Add another word or two. Uncommon words are better.
- Avoid sequences.


=========================
       IMPORTANT!        
=========================
These are your wallet seed words.
They can be used to recover your wallet and funds.
WRITE THEM DOWN OR COPY THEM NOW. THIS IS YOUR ONLY CHANCE TO DO SO.

=========================
then frozen garage dash file barely purity illness mother bridge example level short burst swing acid cry orient crawl garbage snap maze mad plate
=========================

I confirm that I will never see these seed words again.
Type the word "confirm" to continue.
>> 
```
The wallet only warns the user, then continues without prompt. 
The user should be prompted if h/she wants to continue or select a new password. 

Currently, the only way to get a better password is to delete the wallet and start new as per this: #5120, the wallet cannot change its passwords once started

AaronFeickert:
Description
---
Improves the flow for setting or changing a passphrase.

Closes [issue 5127](https://github.com/tari-project/tari/issues/5127).

Motivation and Context
---
When setting or [changing](https://github.com/tari-project/tari/pull/5175) a wallet passphrase, the console wallet provides [feedback](https://github.com/tari-project/tari/pull/5111) on the strength of the provided passphrase. In the case of a weak passphrase, it does not prompt the user to choose a better one.

This PR implements a better flow for this process, as shown in [this flowchart](https://github.com/tari-project/tari/issues/5127#issuecomment-1487027081).

How Has This Been Tested?
---
Tested manually.

---
Testing needs to be done manually to assert that the process represented by the linked flowchart is implemented. Manual testing should cover the entire flow for these two operations:
- Setting the passphrase for a new wallet
- Changing the passphrase for an existing wallet

Breaking Changes
---
None.

### DIFF
--- a/applications/tari_console_wallet/src/init/mod.rs
+++ b/applications/tari_console_wallet/src/init/mod.rs
@@ -78,6 +78,69 @@ pub enum WalletBoot {
     Recovery,
 }
 
+/// Get and confirm a passphrase from the user, with feedback
+/// This is intended to be used for new or changed passphrases
+///
+/// You must provide the initial and confirmation prompts to pass to the user
+///
+/// We do several things:
+/// - Prompt the user for a passphrase
+/// - Have the user confirm the passphrase
+/// - Score the passphrase
+/// - If the passphrase is weak (or empty), give feedback and ask the user what to do:
+///   - Proceed with the weak (or empty) passphrase
+///   - Choose a better passphrase
+///   - Cancel the operation
+///
+/// If the passphrase and confirmation don't match, or if the user cancels, returns an error
+/// Otherwise, returns the passphrase as a `SafePassword`
+fn get_new_passphrase(prompt: &str, confirm: &str) -> Result<SafePassword, ExitError> {
+    // We may need to prompt for a passphrase multiple times
+    loop {
+        // Prompt the user for a passphrase and confirm it
+        let passphrase = prompt_password(prompt)?;
+        let confirmed = prompt_password(confirm)?;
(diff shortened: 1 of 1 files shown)


applications/tari_console_wallet/src/init/mod.rs
```diff
@@ -78,6 +78,69 @@ pub enum WalletBoot {
...
+        }
+
+        // Score the passphrase and provide feedback
+        let weak = display_password_feedback(&passphrase);
+
+        // If the passphrase is weak, see if the user wishes to change it
+        if weak {
+            println!("What would you like to do?");
+            println!("  1. Choose a different passphrase");
+            println!("  2. Use this passphrase");
```
SWvheerden: I would make this `y` and `n` rather than `1` and `2`

SWvheerden (commented):
This looks good. 
Tested working

applications/tari_console_wallet/src/init/mod.rs
AaronFeickert: Updated to improve the prompt.

SWvheerden (approved):

SWvheerden:
ACK

PR #5282, opened 2023-03-30
feat!: change signature construction to allow better HW support

SWvheerden:
Description
---
Changes the challenges of all signatures to not be `Hash(Key||Nonce||Feat_1||Feat_2||..||Feat_x) `but rather `Hash(Key||Nonce||Hash(Feat_1||Feat_2||..||Feat_x))`

Adds the version to the kernel signature as this is currently not checked. 

Motivation and Context
---
This is done because hardware devices need to sign all the data they need to be sent over to the hardware device to be signed. If we have to send all the features over this becomes quite significant in bytes. This is now changed to only send the `message` field as a single 32-byte array and this still comments to all the features. 

The kernel version is technically malleable as the version is not committed to in the signature, this is now fixed. 
How Has This Been Tested?
---
All tests pass

---
Make sure all chain signatures are changed. 
Adds version to Kernel signature

Breaking Changes
---
Change all chain signature challenges: Kernel, Script, Metadata

- [x] None
- [x] Requires hard fork
- [x] Other - Please specify
Complete chain reset. 

BREAKING CHANGE: Complete chain reset


base_layer/core/src/transactions/transaction_components/transaction_input.rs
```diff
@@ -168,15 +168,31 @@ impl TransactionInput {
...
+                    .finalize()
+            },
+        }
+    }
+
+    pub fn build_script_signature_message(
+        version: TransactionInputVersion,
+        script: &TariScript,
+        input_data: &ExecutionStack,
+        commitment: &Commitment,
```
AaronFeickert: It makes sense why `script` and `input_data` are hashed here for complexity reasons, but why `commitment`?

base_layer/core/src/transactions/transaction_components/transaction_kernel.rs
```diff
@@ -178,22 +181,43 @@ impl TransactionKernel {
...
             .chain(sum_public_nonces)
             .chain(total_excess)
+            .chain(&message);
+        match version {
+            TransactionKernelVersion::V0 => common.finalize(),
+        }
+    }
+
+    /// Helper function to create the message of the challenge
+    pub fn build_kernel_signature_message(
```
AaronFeickert, line 208: Can you document why these particular fields are included in this hash, but not others?

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -291,14 +291,38 @@ impl TransactionOutput {
...
             .chain(sender_offset_public_key)
             .chain(commitment)
+            .chain(&message);
+        match version {
+            TransactionOutputVersion::V0 | TransactionOutputVersion::V1 => common.finalize(),
+        }
+    }
+
+    /// Convenience function that calculates the message for the metadata commitment signature
+    pub fn build_metadata_signature_message(
```
AaronFeickert, line 317: Can you document why these particular fields are included in this hash, but not others?

AaronFeickert (changes requested):

base_layer/core/src/transactions/transaction_components/transaction_input.rs
SWvheerden: commitment should move, its part of the signing

hansieodendaal (dismissed):
utACK

stringhandler (commented):
I'm surprised to see no tests being updated, which probably means we are not checking for maleablity in tests

SWvheerden:
The reason we do not see tests updated is that the signatures are generated using the `helper functions` 
So updating them, updates the tests as well. 
We test comprehensibly  for malleability here: https://github.com/tari-project/tari/blob/development/base_layer/core/tests/chain_storage_tests/chain_storage.rs#L1981

stringhandler (approved):
utACK

PR #5286, opened 2023-04-04
chore: update to use crates.io versions of tari

stringhandler:
We currently use a tagged version of `tari_crypto` and `tari_utilities`, this is useful when there are many changes to those crates and we want to use a specific version, but with the `tari_dan` crates also using them, it means the libraries must all be updated at the same time. Using `crates.io` versions means that the end applications will be able to choose a version that works, adhering to semver rules

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1479,20 +1479,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


hansieodendaal (approved):
ACK

PR #5287, opened 2023-04-04
feat: feature gates

brianp:
Description
---
This PR is for reflective purposes regarding feature gating at compile time. See https://github.com/tari-project/tari/issues/5135 for more info. This PR as it stands is to be used as an example of a network _type_ based feature gate at compile time.
It allows setting an ENVVAR for the network with conditional compilation for features. The feature sets can be imported across any crate in the project easily and then used with the `#[cfg(tari_feature_...)]` attribute macro.

I've also introduced a secondary commit for binaries to perform a quick network check against itself to ensure the intended network is supported by the binary. This check is made non-invasive (not conditionally compiling possible network) as that method of check ends up being much more tedious.

Motivation and Context
---
We want some features on some networks, but maybe not all.

How Has This Been Tested?
---
Manually

---
Compile a binary with a set network
`TARI_NETWORK=nextnet cargo build --release --bin tari_base_node`

then run the binary with a test network as a parameter:

`./tari_base_node --network esme` 

and receive an error:

```
The application exited because of an internal network error. The network esmeralda is invalid for this binary built for NextNet
```

*Please note*

That running the binary will cause the build to default to the test network _always_. Meaning this will not work:

```
$ TARI_NETWORK=nextnet cargo build --release --bin tari_base_node
$ cargo run --release --bin tari_base_node --network nextnet
```

The second command `cargo run` will re-build the binary, and without the `TARI_NETWORK` env set the binary will default to a test binary.

In development if you want to call run and connect to a non test network you need to also pass the TARI_NETWORK envvar instead of the `--network` flag. 
`TARI_NETWORK=nextnet cargo run --bin tari_base_node --release -- --network nextnet`

Breaking Changes
---

- [x] None

### DIFF
--- a/applications/tari_app_utilities/Cargo.toml
+++ b/applications/tari_app_utilities/Cargo.toml
@@ -6,9 +6,10 @@ edition = "2018"
 license = "BSD-3-Clause"
 
 [dependencies]
-tari_comms = { path = "../../comms/core" }
 tari_common = { path = "../../common" }
 tari_common_types = { path = "../../base_layer/common_types" }
+tari_comms = { path = "../../comms/core" }
+tari_features = { version = "0.0.1", path = "../../common/tari_features"}
 tari_utilities = { version = "0.4.10"}
 
 clap = { version = "3.2.0", features = ["derive", "env"] }
@@ -22,3 +23,4 @@ thiserror = "^1.0.26"
 
 [build-dependencies]
 tari_common = { path = "../../common", features = ["build", "static-application-info"] }
+tari_features = { version = "0.0.1", path = "../../common/tari_features"}
--- a/applications/tari_app_utilities/Cargo.toml
+++ b/applications/tari_app_utilities/Cargo.toml
@@ -9,7 +9,7 @@ license = "BSD-3-Clause"
 tari_common = { path = "../../common" }
 tari_common_types = { path = "../../base_layer/common_types" }
 tari_comms = { path = "../../comms/core" }
-tari_features = { version = "0.0.1", path = "../../common/tari_features"}
+tari_features = { version = "0.49.0-pre.6", path = "../../common/tari_features"}
 tari_utilities = { version = "0.4.10"}
 
 clap = { version = "3.2.0", features = ["derive", "env"] }
@@ -23,4 +23,4 @@ thiserror = "^1.0.26"
 
 [build-dependencies]
 tari_common = { path = "../../common", features = ["build", "static-application-info"] }
-tari_features = { version = "0.0.1", path = "../../common/tari_features"}
+tari_features = { version = "0.49.0-pre.6", path = "../../common/tari_features"}
--- a/applications/tari_app_utilities/src/network_check.rs
+++ b/applications/tari_app_utilities/src/network_check.rs
@@ -53,18 +53,14 @@ pub const TARGET_NETWORK: Target = Target::NextNet;
 pub const TARGET_NETWORK: Target = Target::TestNet;
 
 pub fn is_network_choice_valid(network: Network) -> Result<(), NetworkCheckError> {
-    match TARGET_NETWORK {
-        Target::MainNet => match network {
-            Network::MainNet | Network::StageNet => Ok(()),
-            _ => Err(NetworkCheckError::MainNetBinary(network)),
-        },
-        Target::NextNet => match network {
-            Network::NextNet => Ok(()),
-            _ => Err(NetworkCheckError::NextNetBinary(network)),
-        },
-        Target::TestNet => match network {
-            Network::LocalNet | Network::Igor | Network::Esmeralda => Ok(()),
-            _ => Err(NetworkCheckError::TestNetBinary(network)),
-        },
+    match (TARGET_NETWORK, network) {
+        (Target::MainNet, Network::MainNet | Network::StageNet) => Ok(()),
+        (Target::MainNet, _) => Err(NetworkCheckError::MainNetBinary(network)),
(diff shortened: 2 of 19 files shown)


applications/tari_app_utilities/src/network_check.rs
```diff
@@ -0,0 +1,70 @@
...
+        },
+        Target::NextNet => match network {
+            Network::NextNet => Ok(()),
+            _ => Err(NetworkCheckError::NextNetBinary(network)),
+        },
+        Target::TestNet => match network {
+            Network::LocalNet | Network::Igor | Network::Esmeralda => Ok(()),
+            _ => Err(NetworkCheckError::TestNetBinary(network)),
+        },
+    }
```
CjS77:
```suggestion
    match (TARGET_NETWORK, network) {
        (Target::MainNet, Network::MainNet | Network::StageNet) => Ok(()),
        (Target::MainNet,   _) => Err(NetworkCheckError::MainNetBinary(network)),
        (Target::NextNet,  Network::NextNet) => Ok(()),
        (Target::NextNet, _) => Err(NetworkCheckError::NextNetBinary(network)),
        (Target::TestNet,   Network::LocalNet | Network::Igor | Network::Esmeralda) => Ok(()),
        (Target::TestNet, _) => Err(NetworkCheckError::TestNetBinary(network)),
        },
    }
```

A lot more readable I think you'll find

common/tari_features/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+// USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+mod feature;
+pub mod resolver;
+mod status;
+
+pub use feature::Feature;
+pub use status::Status;
+
```
CjS77, line 29: Some docs here on when / how / why to maintain this list would be highly useful.

CjS77 (commented):
Looks good. Some minor readability and maintenance suggestions

common/tari_features/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "tari_features"
+description = "Compilable features for Tari applications"
+authors = ["The Tari Development Community"]
+repository = "https://github.com/tari-project/tari"
+homepage = "https://tari.com"
+readme = "README.md"
+license = "BSD-3-Clause"
+version = "0.0.1"
```
SWvheerden: This is probably up for discussion, but I think we should version match here with the rest of the Tari code base. 
at least for now.

applications/tari_app_utilities/Cargo.toml
```diff
@@ -22,3 +23,4 @@ thiserror = "^1.0.26"
 
 [build-dependencies]
 tari_common = { path = "../../common", features = ["build", "static-application-info"] }
+tari_features = { version = "0.0.1", path = "../../common/tari_features"}
```
SWvheerden: is there a reason for including it twice with the same feature set?

applications/tari_app_utilities/src/network_check.rs
```diff
@@ -0,0 +1,70 @@
...
+    TestNetBinary(Network),
+}
+
+impl From<NetworkCheckError> for ExitError {
+    fn from(err: NetworkCheckError) -> Self {
+        Self::new(ExitCode::NetworkError, err)
+    }
+}
+
+#[cfg(tari_network_mainnet)]
```
SWvheerden, line 46: Can we add a default to this?
So if the user does not specify a network, it defaults to mainnet

common/tari_features/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+pub const FEATURE_LIST: [Feature; 0] = [
+    // Feature::new(
+    //     "example_feature_name",
+    //     "example description",
+    //     None,
+    //     Status::New,
+    // ),
```
SWvheerden: remove?

SWvheerden (commented):
This looks good
The only thing I would add is a default network, so if the user does not select a network, it will default to mainnet, or perhaps development as building from dev that's the most likely? But this is open to discussion

SWvheerden (commented):
We need to update the readme with all this as well.

common/tari_features/Cargo.toml
brianp: That makes sense to me.

applications/tari_app_utilities/Cargo.toml
brianp: I was thinking it wasn't available in the build deps unless specifically defined, although I'll double check it's needed.

applications/tari_app_utilities/src/network_check.rs
brianp: There is a default defined as the Target::TestNet. Anytime we do compiled compilation we have to ensure something will be present. For all the feature gating TestNet is the default compilation path.

`#[cfg(all(not(tari_network_mainnet), not(tari_network_nextnet)))]`

applications/tari_app_utilities/Cargo.toml
brianp: Validated. For use in build.rs it _is_ required to be included in both categories.

SWvheerden (approved):

SWvheerden:
Ack

PR #5288, opened 2023-04-04
feat: improve message encryption

### LINKED ISSUE #5255: Consider the use of key-blinding signatures for encrypted messages
A [recent preprint](https://eprint.iacr.org/2023/380) formalizes key-blinding signatures, which can be useful in cases where a single key is used to generate unlinkable signatures. It builds on earlier work to describe and prove security for `EdDSA` and `ECDSA` variants.

In this design (modified for non-`EdDSA` Schnorr signatures), the signer blinds their public key and uses it to generate a standard Schnorr signature that any verifier can verify using the blinded public key. A designated entity in possession of the secret data used to construct the blinded public key can additionally unblind the original public key. These signatures have the benefit of being unlinkable in the sense that an adversary (who does not know the secret data) cannot distinguish signatures produced from the same underlying public key from those produced from different keys.

This could be useful for encrypted messages. Right now, signatures and public keys are encrypted to the recipient, who must decrypt them, verify the signature using the public key, and then decrypt the message. Non-recipient entities can only assert that the semantics of the encrypted message, like its length and the presence of expected fields.

With this new design, the sender could instead blind its public key, encrypt the secret data along with the message, and sign both (along with existing header-type data) using the blinded public key. Non-recipient entities can verify the signature, but are unable to identify the underlying public key. The recipient can decrypt the verified message and unblind the public key to identify the sender.

Overall, this seems like a cleaner design that is more in line with some version of the [cryptographic doom principle](https://moxie.org/2011/12/13/the-cryptographic-doom-principle.html).

AaronFeickert:
Description
---
Improves message encryption using key-blinding signatures. Updates `tari-crypto` to the [latest version](https://crates.io/crates/tari_crypto/0.16.12) for proper compatibility.

Closes [issue 5255](https://github.com/tari-project/tari/issues/5255).

Motivation and Context
---
Currently, message encryption uses an encrypted signature. A [recent formalization](https://eprint.iacr.org/2023/380) of key-blinding signatures offers an interesting new approach that allows any network node to verify an unencrypted signature on a masked key. This PR updates message encryption to use such a design.

How Has This Been Tested?
---
Existing, modified, and new tests pass.

---
Examine and run tests. Assert that message encryption follows [these flowcharts](https://github.com/tari-project/tari/issues/5255#issuecomment-1476875050).

Breaking Changes
---
None. While existing messages will not verify, this should not affect new messages that use the updated design.

### DIFF
--- a/applications/tari_app_grpc/Cargo.toml
+++ b/applications/tari_app_grpc/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 tari_common_types = {  path = "../../base_layer/common_types" }
 tari_comms = { path = "../../comms/core" }
 tari_core = { path = "../../base_layer/core" }
-tari_crypto = { version="0.16.11"}
+tari_crypto = { version = "0.16.12"}
 tari_script = { path = "../../infrastructure/tari_script" }
 tari_utilities = { version = "0.4.10"}
 
--- a/applications/tari_base_node/Cargo.toml
+++ b/applications/tari_base_node/Cargo.toml
@@ -15,7 +15,7 @@ tari_comms = { path = "../../comms/core", features = ["rpc"] }
 tari_common_types = { path = "../../base_layer/common_types" }
 tari_comms_dht = { path = "../../comms/dht" }
 tari_core = { path = "../../base_layer/core", default-features = false, features = ["transactions"] }
-tari_crypto = { version = "0.16.11"}
+tari_crypto = { version = "0.16.12"}
 tari_libtor = { path = "../../infrastructure/libtor", optional = true }
 tari_p2p = { path = "../../base_layer/p2p", features = ["auto-update"] }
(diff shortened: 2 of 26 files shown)


AaronFeickert:
This PR is a work in progress and should not yet be reviewed or merged. It exists in draft form for CI testing purposes.

stringhandler:
Just marking this with DNM for now until the tari crypto version has been published to crates.io

AaronFeickert:
> Just marking this with DNM for now until the tari crypto version has been published to crates.io

@stringhandler: Should be fine to remove this label now.

SWvheerden (dismissed):

SWvheerden:
utAck

SWvheerden (dismissed):

SWvheerden (approved):

PR #5298, opened 2023-04-09
feat: relax message size constraint

hansieodendaal:
Description
---
Relaxed the message size constraint from 4 MiB to 5 MiB as it prevented faucet outputs from being streamed in an RPC request when a wallet requested all outputs.

Motivation and Context
---
Faucet outputs could not be streamed to the console wallet:
```
2023-04-06 17:08:52.630699700 [wallet::utxo_scanning] WARN  Failed to scan UTXO's from base node 6b8a34873128fc5323a86fbe6e: RpcStatus: `MalformedResponse: The response size exceeded the maximum allowed payload size. Max = 4.0000 MiB, Got = 4.5924 MiB`
2023-04-06 17:08:52.631420600 [wallet::utxo_scanning] WARN  Failed to scan UTXO's from base node 6b8a34873128fc5323a86fbe6e: RpcError: `Remote peer unexpectedly closed the RPC connection`
2023-04-06 17:08:52.631441300 [wallet::utxo_scanning] ERROR Error scanning UTXOs: Utxo Scanning Error: 'Failed to scan UTXO's after 2 attempt(s) using sync peer(s). Aborting...
```

How Has This Been Tested?
---
System-level testing on `igor`

---
System-level testing on `igor` 
- Prior to this PR: create a new wallet and monitor the log file for the error
- With this PR: create a new wallet and verify in the log file that UTXO scanning succeeds

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/core/src/protocol/rpc/mod.rs
+++ b/comms/core/src/protocol/rpc/mod.rs
@@ -32,7 +32,7 @@ mod test;
 /// This can be thought of as the hard limit on message size.
 pub const RPC_MAX_FRAME_SIZE: usize = 3 * 1024 * 1024; // 3 MiB
 /// Maximum number of chunks into which a message can be broken up.
-const RPC_CHUNKING_MAX_CHUNKS: usize = 16; // 16 x 256 Kib = 4 MiB max combined message size
+const RPC_CHUNKING_MAX_CHUNKS: usize = 20; // 20 x 256 Kib = 5 MiB max combined message size
 const RPC_CHUNKING_THRESHOLD: usize = 256 * 1024;
 const RPC_CHUNKING_SIZE_LIMIT: usize = 384 * 1024;
 


SWvheerden (approved):

SWvheerden (commented):
on hind sight, perhaps better to it this way: https://github.com/tari-project/tari/issues/5299

hansieodendaal:
> on hind sight, perhaps better to it this way: #5299

Maybe we can merge this, and revert the size constraint once #5299 is done so that testing on `igor` can continue.

SWvheerden:
This is an alternative to PR #5302 

The issue we have is that the Genesis block has too many utxos to send over the wire with the 4mb size limit. 
This should only be an issue with the Genesis block as with any other block the whole block should be less than 4mb. 
We can increase the size limit, but then increase the size for everybody. The other PR slightly ever increases the send/receive code to handle streaming of sub block batches of utxos.

SWvheerden:
Closing this in favour of: #5302

PR #5303, opened 2023-04-11
fix: allow public addresses from command line

### LINKED ISSUE #5260: public_addresses can't be set via env var or command line
**Describe the bug**
`public_addresses` can't be set via env var or command line. There need to be a custom deserializer (similar to `peer_seeds`) to support `Vec`. Also in the `config.toml` are still places with public_address which needs to be changed to `public_addresses`. It's currently done only in the base node section.

**Expected behavior**
Probably a comma separated values as in `peer_seeds`.

Cifko:
Description
---
Add `from_str` to `public_addresses` so it can now be set via command line or envvar. 

How Has This Been Tested?
---
There are cargo test and also manually.

---
You can set `public_addresses` via commandline.

Fixes: #5260 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_base_node/src/main.rs
+++ b/applications/tari_base_node/src/main.rs
@@ -136,7 +136,7 @@ fn main_inner() -> Result<(), ExitError> {
     // Load or create the Node identity
     let node_identity = setup_node_identity(
         &config.base_node.identity_file,
-        config.base_node.p2p.public_addresses.clone(),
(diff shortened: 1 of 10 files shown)


stringhandler (dismissed):

SWvheerden (dismissed):
Ack

SWvheerden (approved):

Issue #5295, opened 2023-04-06
Range proof validation mismatch [igor]


hansieodendaal:
This issue was caused by manually submitting a transaction to a base node that spends an igor faucet UTXO without a range proof.

- Any base node that the transaction was directly submitted to accepts and inserts it into its mempool.
```rust
2023-04-06 08:26:55.871685100 [c::mp::service::inbound_handlers] DEBUG Handling remote request: SubmitTransaction (3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a)
2023-04-06 08:26:55.871700300 [c::mp::service::inbound_handlers] DEBUG Transaction (3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a) submitted using request.
2023-04-06 08:26:55.871925900 [c::mp::mempool_storage] DEBUG Inserting tx into mempool: 3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a
2023-04-06 08:26:55.879899200 [c::mp::mempool_storage] DEBUG Transaction 3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a is VALID (7.97ms), inserting in unconfirmed pool in
2023-04-06 08:26:55.880437500 [c::mp::unconfirmed_pool::unconfirmed_pool_storage] DEBUG Inserted transaction 3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a (weight: 250, internal key: 0) into unconfirmed pool:
2023-04-06 08:26:55.880455100 [c::mp::mempool_storage] DEBUG Transaction 3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a inserted in 519.60µs
2023-04-06 08:26:55.880626900 [c::mp::service::inbound_handlers] DEBUG Transaction inserted into mempool: 3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a, pool: Unconfirmed pool.
2023-04-06 08:26:55.880652300 [c::mp::service::inbound_handlers] DEBUG Propagate transaction (3dd48fbf1c9cb7129e5ec6b5b227765e13c8beeb86fe5462bdc054402bfa390a) to network.
``` 
[...]

PR #5304, opened 2023-04-11
feat: fix rpc transaction conversion

hansieodendaal:
Description
---
Fixed RPC transaction proto conversion

Motivation and Context
---
With the RPC conversion `minimum_value_promise` was zeroed instead of assigned
See #5295 

How Has This Been Tested?
---

---
Run a system-level test to submit a transaction via gRPC to a base node where an output has `minimum_value_promise != 0` and then mine the transaction

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/proto/transaction.rs
+++ b/base_layer/core/src/proto/transaction.rs
@@ -259,7 +259,7 @@ impl TryFrom<proto::types::TransactionOutput> for TransactionOutput {
 
         let encrypted_value = EncryptedValue::from_bytes(&output.encrypted_value).map_err(|err| err.to_string())?;
 
-        let minimum_value_promise = MicroTari::zero();
+        let minimum_value_promise = output.minimum_value_promise.into();
 
         Ok(Self::new(
             TransactionOutputVersion::try_from(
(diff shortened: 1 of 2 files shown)


base_layer/core/src/proto/transaction.rs
```diff
@@ -259,7 +259,7 @@ impl TryFrom<proto::types::TransactionOutput> for TransactionOutput {
 
         let encrypted_value = EncryptedValue::from_bytes(&output.encrypted_value).map_err(|err| err.to_string())?;
 
-        let minimum_value_promise = MicroTari::zero();
+        let minimum_value_promise = output.minimum_value_promise.into();
```
SWvheerden, line 262: man these defaults...

SWvheerden (dismissed):

base_layer/core/src/proto/transaction.rs
hansieodendaal: Yip

SWvheerden (approved):

PR #5307, opened 2023-04-12
feat: define different network defaults for bins

### LINKED ISSUE #5310: Use feature gates to set Network::default for different bins
**Problem**
Different bins compiled for different networks still use the same default network, which means the "out of the box" default could be fundamentally broken.

**Solution**
Use the newely merged feature gating and set different default Networks for different bin types.

The defaults should be:

TestNet -> Esme
NextNet -> NextNet
StageNet/MainNet -> StageNet

brianp:
Description
---
This commit make a simple switch to define Network::default differently for the different bins compiled for different networks.

Motivation and Context
---
Making us one step closer to smarter defaults, and one click, out of the box.

How Has This Been Tested?
---
Manually

---
Compile the bin with `TARI_NETWORK=...` then validate the bin uses the correct default

Fixes: #5310 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (dismissed):

SWvheerden:
utAck

SWvheerden (approved):
Ack

PR #5309, opened 2023-04-12
chore: update log to print display and not debug

SWvheerden:
Description
---
Changes a log message to print out the display and not debug of the struct

Motivation and Context
---
The Monero pow data does implement display and we don't need to print out debug. 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
I checked to make sure MoneroPowData has a display trait and it does.

LGTM utACK

PR #5322, opened 2023-04-13
feat: added burn feature to the console wallet

agubarev:
Description
---
Added basic functionality to enable Minotari burning on the baselayer.
Copied `serde_from` module from `tari-dan` to `tari_common_types`

Motivation and Context
---
To allow users burn Minotari to be reclaimed on the second layer (DAN).

How Has This Been Tested?
---
manually


applications/tari_console_wallet/src/ui/components/burn_tab.rs
```diff
@@ -0,0 +1,694 @@
...
+            BurnInputMode::Message => f.set_cursor(
+                // Put cursor past the end of the input text
+                vert_chunks[4].x + self.message_field.width() as u16 + 1,
+                // Move one line down, from the border to the input line
+                vert_chunks[4].y + 1,
+            ),
+        }
+    }
+
+    #[allow(dead_code)]
```
SWvheerden: Is this needed, or just a left over from copy paste?

applications/tari_console_wallet/src/ui/components/burn_tab.rs
```diff
@@ -0,0 +1,694 @@
...
+
+                return KeyHandled::Handled;
+            },
+
+            _ => {},
+        }
+
+        KeyHandled::NotHandled
+    }
+}
```
SWvheerden:
```suggestion
      if !self.show_proofs || (c != 'd'  || c != 'o'){
           return KeyHandled::NotHandled;
      }
                 if let Some(proof) = self
                    .proofs_list_state
                    .selected()
                    .and_then(|i| app_state.get_burnt_proof_by_index(i))
                    .cloned()
                {
                    if self.proofs_list_state.selected().is_none() {
                        return KeyHandled::NotHandled;
                    }
                   if c == 'd' {
                    self.confirmation_dialog = Some(BurnConfirmationDialogType::DeleteBurntProof(proof.id));
                     }
                     else
                     {
                       if let Err(e) = fs::write(format!("{}.json", proof.id), proof.payload) {
                        self.error_message = Some(format!(
                            "Failed to save burnt proof payload to file {}.json: {}, Press Enter to continue.",
                            proof.id, e
                        ));
                    }
                    }

                return KeyHandled::Handled;
            },

            return KeyHandled::Handled;
}
```

Doing it this way will remove all the code duplication and reduce the function lines

applications/tari_console_wallet/src/ui/components/burn_tab.rs
```diff
@@ -0,0 +1,694 @@
...
+            'f' => self.burn_input_mode = BurnInputMode::Fee,
+            'm' => self.burn_input_mode = BurnInputMode::Message,
+            'b' => {
+                self.show_proofs = !self.show_proofs;
+            },
+            's' => {
+                // if self.burnt_proof_filepath_field.is_empty() {
+                // self.error_message = Some("Burn proof filepath is empty\nPress Enter to continue.".to_string());
+                // return;
+                // }
```
SWvheerden:
```suggestion

```
delete?

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -405,6 +485,14 @@ impl AppState {
         &self.cached_data.contacts[start..end]
     }
 
+    pub fn get_burnt_proofs_slice(&self, start: usize, end: usize) -> &[UiBurntProof] {
+        if self.cached_data.burnt_proofs.is_empty() || start > end || end > self.cached_data.burnt_proofs.len() {
```
SWvheerden:
```suggestion
        if self.cached_data.burnt_proofs.is_empty() || start => end || end > self.cached_data.burnt_proofs.len() {
```
if start == end and end == len, this will panic

applications/tari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -215,3 +225,101 @@ pub async fn send_one_sided_to_stealth_address_transaction(
...
+    mut transaction_service_handle: TransactionServiceHandle,
+    db: WalletDatabase<WalletSqliteDatabase>,
+    result_tx: watch::Sender<UiTransactionBurnStatus>,
+) {
+    result_tx.send(UiTransactionBurnStatus::Initiated).unwrap();
+    let mut event_stream = transaction_service_handle.get_event_stream();
+
+    match transaction_service_handle
+        .burn_tari(amount, selection_criteria, fee_per_gram, message, claim_public_key)
+        .await
```
SWvheerden: This adds an unnecessary nest, 
```suggestion
 let (burn_tx_id, original_proof)= ransaction_service_handle
        .burn_tari(amount, selection_criteria, fee_per_gram, message, claim_public_key)
        .await.map_err(|e|{result_tx
                .send(UiTransactionBurnStatus::Error(UiError::from(e).to_string()))
                .unwrap();})?;
```

applications/tari_console_wallet/src/ui/app.rs
```diff
@@ -186,3 +198,31 @@ impl<B: Backend> App<B> {
         self.menu.draw(f, title_chunks[3], &self.app_state);
     }
 }
+
+#[derive(Error, Debug)]
+pub enum UiError {
```
SWvheerden: Why move this here?
Smaller files are easier to read than larger monolithic files.

applications/tari_console_wallet/src/ui/ui_error.rs
```diff
@@ -1,38 +0,0 @@
...
-use tari_utilities::hex::HexError;
-use tari_wallet::{
-    error::{WalletError, WalletStorageError},
-    output_manager_service::error::OutputManagerError,
-    transaction_service::error::TransactionServiceError,
-};
-use thiserror::Error;
-
-#[derive(Error, Debug)]
-pub enum UiError {
```
SWvheerden: The move of deleting this and moving it breaks the code structure away from the rest of the code base. 
And having more smaller files are easier to navigate than larger monolithic files

base_layer/wallet/migrations/2022-08-08-134037_initial/up.sql
```diff
@@ -39,75 +42,88 @@ CREATE TABLE inbound_transactions (
...
 );
 
-CREATE TABLE wallet_settings (
+CREATE TABLE wallet_settings
+(
     key   TEXT PRIMARY KEY NOT NULL,
     value TEXT             NOT NULL
 );
+
+CREATE TABLE burnt_proofs
```
SWvheerden: This will break all existing wallets, and create a proper migration here

Leave this file as is

SWvheerden (changes requested):
This looks good. 
But this needs a migration before we can merge, we cannot do breaking changes anymore. 
There is one bug in the code as well, and a few nits

applications/tari_console_wallet/src/ui/components/burn_tab.rs
agubarev: it's a leftover, thanks for pointing out, I'll sweep the code for similar cases

applications/tari_console_wallet/src/ui/app.rs
agubarev: these structs are very small and putting them into separate files seems like an overkill at that point, once something grows in body size, then sure. I can split them back if you want though.

base_layer/wallet/migrations/2022-08-08-134037_initial/up.sql
agubarev: you're right, I'll create a new migration for that

applications/tari_console_wallet/src/ui/app.rs
stringhandler: It should be in a separate file because smaller files are easier to navigate and more friendly to conflicts

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -405,6 +485,14 @@ impl AppState {
         &self.cached_data.contacts[start..end]
     }
 
+    pub fn get_burnt_proofs_slice(&self, start: usize, end: usize) -> &[UiBurntProof] {
+        if self.cached_data.burnt_proofs.is_empty() || start => end || end > self.cached_data.burnt_proofs.len() {
```
stringhandler: I prefer returning anything up to the end of the vec if end is greater than the length
```suggestion
        if self.cached_data.burnt_proofs.is_empty() || start >= end  {
```

applications/tari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -779,6 +867,30 @@ impl AppStateInner {
...
+                id: proof.0,
+                reciprocal_claim_public_key: proof.1,
+                payload: proof.2,
+                burned_at: proof.3,
+            });
+        }
+
+        ui_proofs.sort_by(|a, b| {
+            a.burned_at
+                .partial_cmp(&b.burned_at)
```
stringhandler: Why not just use cmp so that we can avoid the expect?
```suggestion
                .cmp(&b.burned_at)
```

applications/tari_console_wallet/src/ui/state/mod.rs
```diff
@@ -25,4 +25,31 @@ mod debouncer;
 mod tasks;
 mod wallet_event_monitor;
 
+use serde::{Deserialize, Serialize};
+use tari_common_types::serde_with;
+
 pub use self::app_state::*;
+
+// ----------------------------------------------------------------------------
+// TODO: re-implement in a clean way
```
stringhandler: "Clean way" is subjective. Maybe expand on what needs to be done here. I don't think the comment is needed. This is suitable for now

applications/tari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -215,3 +225,101 @@ pub async fn send_one_sided_to_stealth_address_transaction(
...
+                                        proof_id,
+                                        original_proof.reciprocal_claim_public_key.to_hex(),
+                                        serialized_proof,
+                                        ts,
+                                    )))
+                                    .unwrap();
+
+                                return;
+                            }
+                        }
```
stringhandler, line 298: I would add an `else` here because there could be other events that are getting swallowed, potentially introducing a bug in future

applications/tari_console_wallet/src/ui/types.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Debug, Clone)]
+pub struct UiBurntProof {
+    pub id: i32,
+    pub reciprocal_claim_public_key: String,
+    pub payload: String,
+    pub burned_at: NaiveDateTime,
+}
+
+#[derive(Debug, Clone)]
+pub struct UiContact {
```
stringhandler: I don't agree with this. Please put it back

base_layer/common_types/src/serde_with/hex.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Use a serde deserializer to serialize the hex string of the given object.
+pub fn deserialize<'de, D, T>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: TryFrom<Vec<u8>>,
+{
+    let bytes = if d.is_human_readable() {
+        let hex = <String as Deserialize>::deserialize(d)?;
+        from_hex(&hex).map_err(serde::de::Error::custom)?
```
stringhandler: It would be useful if you combine these two deserializers by trying to decode into base64 and on failure trying to decode into hex and finally erroring if neither work

base_layer/common_types/src/serde_with/string.rs
```diff
@@ -0,0 +1,28 @@
+//   Copyright 2022 The Tari Project
+//   SPDX-License-Identifier: BSD-3-Clause
+
+use std::str::FromStr;
+
+use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize, Serializer};
+
```
stringhandler: I'm not sure what the purpose of this serializer is.

base_layer/wallet/src/storage/database.rs
```diff
@@ -79,6 +80,8 @@ pub enum DbKey {
     WalletBirthday,
     LastAccessedNetwork,
     LastAccessedVersion,
+    BurntProofId(i32),
```
stringhandler: i32 is very specific to sqlite. It should be a different type here, and only converted to i32 in the sqlite lib

base_layer/wallet/src/storage/database.rs
```diff
@@ -351,6 +359,41 @@ where T: WalletBackend + 'static
...
+    // burnt proofs
+
+    pub fn insert_burnt_proof(
+        &self,
+        id: i32,
+        reciprocal_claim_public_key: String,
+        proof: String,
+        burned_at: NaiveDateTime,
+    ) -> Result<(), WalletStorageError> {
+        self.db.write(WriteOperation::Insert(DbKeyValuePair::BurntProof((
```
stringhandler: I know the current implementation uses this pattern a lot, but I actually don't like it and I'd prefer to not increase the tech debt. Rather implement the methods individually.

Look at `self.db.clear_scanned_blocks` for the preferred pattern

stringhandler: Same goes for all the other methods you've implemented in this file

SWvheerden: yeah, having an explicit trait interface that does exactly what you want and takes in exactly what you want without some wrapper is much cleaner, we have been removing the this `DbKeyValuePair` from the codebase as we add more and more code

applications/tari_console_wallet/src/ui/state/mod.rs
agubarev: "Clean way" was a poor word choice, cleanliness here is more like finding a logical place to put these transient types to

base_layer/wallet/src/storage/database.rs
agubarev: not using DbKeys anymore

base_layer/wallet/src/storage/database.rs
agubarev: done

applications/tari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -215,3 +223,102 @@ pub async fn send_one_sided_to_stealth_address_transaction(
...
+                                proof_id,
+                                original_proof.reciprocal_claim_public_key.to_hex(),
+                                serialized_proof,
+                            )))
+                            .unwrap();
+
+                        return;
+                    }
+                } else {
+                    //
```
stringhandler:
```suggestion
                    //
                    warn!(target = LOG_TARGET, "Encountered an unexpected event");
```

stringhandler: or even a todo!() would be better than swallowing the event

applications/tari_console_wallet/src/ui/ui_contact.rs
```diff
@@ -1,5 +1,24 @@
-// Copyright 2022 The Tari Project
-// SPDX-License-Identifier: BSD-3-Clause
+//  Copyright 2022. The Tari Project
```
stringhandler: No need, but ok

base_layer/common_types/src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub mod emoji;
 pub mod encryption;
 pub mod epoch;
 pub mod grpc_authentication;
+pub mod serde_with;
```
stringhandler:
```suggestion
pub mod serializers;
```
There is a popular crate called `serde_with` and reusing this name will be misleading

stringhandler (approved):

PR #5323, opened 2023-04-13
chore(ci): add cargo machete

SWvheerden:
Description
---
Adds cargo machete to check for dependencies

Motivation and Context
---
Helps keep the dependencies clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,15 @@ jobs:
         with:
           command: lints
           args: clippy --all-targets --all-features
+      - name: Install cargo-machete
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-machete
+      - name: cargo machete check
+        uses: actions-rs/cargo@v1
+        with:
+          command: machete
   build:
     name: cargo check
     runs-on: [self-hosted, ubuntu18.04-high-cpu]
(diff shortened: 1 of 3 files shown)


CjS77:
Does this action just report, or actually change Cargo.toml?

If former, will it cause a PR to fail checks if there are unused dependencies?

It adds ±2.5 minutes to CI; this is something we could also run nightly as long as it reports properly / creates an issue.

SWvheerden:
This just runs a check on the cargo.toml files in the repo. 
But it does fail the ci
Here is a run of me testing it: https://github.com/tari-project/tari/actions/runs/4689245804/jobs/8310763738

I prefer it running on the pr itself and making sure the offending PR fixes its cargo file. Machete can give false positives, and you need to fix those with exceptions in the cargo files. I prefer the users checking those first before just fixing them. 

As to the time, I can add it directly to the clippy fix, that might be much quicker run, but will fail as part of clippy and not on its own?

stringhandler (approved):

CjS77:
CON: Since unused dependencies aren't a major issue, it could become annoying to do a push-commit-test cycle to GH for every PR (I have had several complaints about this for clippy & cargo fmt already). Do we really want to add more friction, vs run nightly and (potentially even) automatically create a new issue / new PR (GHP could do this)

PRO: On the plus side, I guess we don't add deps that often, so maybe it doesn't come up that much

tl;dr - If you include it in the clippy test, I think we can go ahead

stringhandler:
IMO it actually could save on build time because we are not downloading and building unnecessary crates

Issue #5326, opened 2023-04-14
Default network for different compiled bins is still defaulting to Esme


brianp:
# Problem

With the introduction of network based feature gating I attempted to [apply a change](https://github.com/tari-project/tari/pull/5307) to set different default networks for the different bin compliations.

This change did not have the desired effect and the NextNet binary still defaults to Esme. This is very confusing because a user would be immediately met with an error message informing them Esme is not a valid network choice for the NextNet binary.

The appropriate default was not used because the `tari_common` crate is not utilizing a build.rs, and in turn not utilizing the tari-feature gating configuration.

# Solution

Introduce the tari-feature build settings to tari common. This will also introduce a cyclic dependency so it's more likely that we should move tari-features directly into tari-common.

brianp:
A little more thinking on this and moving tari-features into common doesn't work because we're saying common needs to be tari-feature aware. 

I think what we need to do instead, is free tari-features of any outside dependency. It's only dep is tari-common, and it only uses the Network. I think we could probably address this as the problem.

PR #5327, opened 2023-04-14
chore(release): nextnet hotfix default network selection

brianp:
Description
---
Features is now a frequently used build dep in most our crates. Common actually also needs to be feature aware during build time. This means tari-features should have little to no dependencies, and especially none from our own crates.

Motivation and Context
---
Related to #5326 
Development fix: #5333 

How Has This Been Tested?
---
Manually

---
Run cargo build with the desired network type (nextnet):
`TARI_NETWORK=nextnet cargo build --bin tari_base_node`

Run the bin directly without using cargo. It's important not to use cargo during the testing as the bin will likely rebuild when using `run` and change the previous `TARI_NETWORK` compilation settings:
`./target/tari_base_node` 

See that the default network is NextNet.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5271,6 +5271,7 @@ dependencies = [
  "sha2 0.9.9",
  "structopt",
  "tari_crypto",
+ "tari_features",
  "tari_test_utils",
  "tempfile",
  "thiserror",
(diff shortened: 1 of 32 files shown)


SWvheerden:
This looks good, just needs to be updated to include: #5329

SWvheerden (approved):

PR #5281, opened 2023-03-30
feat!: add igor faucet

hansieodendaal:
Description
---
- Updated the igor genesis block to contain a faucet with revealed values
- Added a new range-proof option for the genesis block sanity test

Motivation and Context
---
- Evaluating faucet with revealed values

How Has This Been Tested?
---
- Unit test demonstrating the range proof for each output in the faucet
- Performed system-level testing to demonstrate that the igor blockchain works properly - 2x base nodes, 2x wallets, 1x SHA3 miner, 1x merge mining proxy, 1x XMRig miner.

What process can a PR reviewer use to test or verify this change?
---
Perform system-level testing

Breaking Changes
---

- [ ] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [x] Other - New genesis block.

 BREAKING CHANGE: The chain has to be restarted - delete the blockchain database and resync the chain

PR #5328, opened 2023-04-14
feat!: update faucet value test and Igor genesis block

AaronFeickert:
Description
---
Updates and simplifies verification of faucet value-revealing signatures. Includes new Igor genesis block data.

Motivation and Context
---
A [recent PR](https://github.com/tari-project/tari/pull/5281) uses metadata signatures with deterministic nonces to allow for commitment value assertion for the Igor genesis block while removing the need for range proof verification.

While this technique works, a pending [RFC update](https://github.com/tari-project/rfcs/pull/95) describes a simpler design that is more efficient to verify.

This PR updates the relevant unit test to use this method. It also includes new Igor genesis block data corresponding to the change.

How Has This Been Tested?
---
Existing tests pass. Updated test passes with new Igor genesis block data.

---
Check that the updated unit test matches the method described in the linked RFC PR. Ensure that tests pass.

Breaking Changes
---
Yes.

BREAKING CHANGE: Includes a new Igor genesis block and faucet data, which necessitates deleting existing chain data and resyncing.

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -584,15 +584,11 @@ fn get_esmeralda_genesis_block_raw() -> Block {
 
 #[cfg(test)]
 mod test {
-    use std::convert::TryFrom;
-
     use croaring::Bitmap;
-    use curve25519_dalek::scalar::Scalar;
     use tari_common_types::{
         epoch::VnEpoch,
         types::{Commitment, PrivateKey},
     };
-    use tari_crypto::{hash::blake2::Blake256, hash_domain, hashing::DomainSeparatedHasher};
     use tari_utilities::ByteArray;
 
     use super::*;
@@ -727,24 +723,12 @@ mod test {
             .unwrap();
     }
 
-    fn receiver_commit_nonce_a(commitment: &Commitment) -> PrivateKey {
-        hash_domain!(
-            CommitNonceHashDomain,
-            "com.tari-genesis-tools.multi_party_utxo.receiver_commit_nonce_a",
-            0
-        );
-        let hasher_bytes = DomainSeparatedHasher::<Blake256, CommitNonceHashDomain>::new()
-            .chain(commitment.as_bytes())
-            .finalize();
-        PrivateKey::from_bytes(hasher_bytes.as_ref()).expect("blake256 hash will succeed")
-    }
-
-    // For the faucet, the value of the commitment can be bound into the metadata signature using a deterministic nonce
+    // For the faucet, the value of the commitment can be bound into a value-revealing metadata signature
     fn do_alternate_range_proof_check(factories: &CryptoFactories, outputs: &[&TransactionOutput]) {
         for output_ref in outputs {
             let output = (*output_ref).clone();
             if output.features.output_type == OutputType::Standard {
-                let commit_nonce_a = receiver_commit_nonce_a(&output.commitment);
+                // Reconstruct the signature challnge
                 let e_bytes = TransactionOutput::build_metadata_signature_challenge(
                     TransactionOutputVersion::get_current_version(),
                     &output.script,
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This PR doesn't yet include new Igor genesis block data.

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -758,24 +742,15 @@ mod test {
...
-                let metadata_signature_u_a_scalar =
-                    Scalar::from_bits(<[u8; 32]>::try_from(output.metadata_signature.u_a().as_bytes()).unwrap());
-                let commit_nonce_a_scalar = Scalar::from_bits(<[u8; 32]>::try_from(commit_nonce_a.as_bytes()).unwrap());
-                let e_scalar = Scalar::from_bits(<[u8; 32]>::try_from(e.as_bytes()).unwrap());
-                let value_scalar = (metadata_signature_u_a_scalar - commit_nonce_a_scalar) * e_scalar.invert();
-                assert_eq!(value_scalar, Scalar::from(output.minimum_value_promise.as_u64()));
-                let mut value_bytes = [0u8; 8];
-                value_bytes.copy_from_slice(&value_scalar.to_bytes()[0..8]);
-                let value = u64::from_le_bytes(value_bytes);
-                assert_eq!(value, output.minimum_value_promise.as_u64());
```
hansieodendaal, line 778: I think this is still valid, we just need to remove `commit_nonce_a` from the equations

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -758,24 +742,15 @@ mod test {
...
+
+                // Check that `u_a = e*v`
+                // This _must_ be in addition to a separate check that the signature is valid!
+                // This check is not done here!
                 let value_as_private_key = PrivateKey::from(output.minimum_value_promise.as_u64());
                 assert_eq!(
                     output.metadata_signature.u_a().to_hex(),
-                    (commit_nonce_a.clone() + e.clone() * value_as_private_key).to_hex()
+                    (e.clone() * value_as_private_key).to_hex()
                 );
```
hansieodendaal, line 753: Agreed

hansieodendaal:
> This PR doesn't yet include new Igor genesis block data.

I will be able to compute a new igor genesis block towards the end of the week perhaps

base_layer/core/src/blocks/genesis_block.rs
AaronFeickert: Would it actually be useful? To extract the value, you need the challenge, which requires knowledge of the value to compute.

hansieodendaal: ~~Well, it asserts that the calculated `value` is equal to the `minimum_value_promise`, the latter being part of the challenge.~~

AaronFeickert: Isn't that somewhat circular? The risk is that if you don't compute the challenge yourself, the proof could be invalid (including a falsified minimum value promise) and you could be fooled into accepting an incorrect value. And if you do compute the challenge yourself, the existing multiplicative assertion is sufficient to show the claimed minimum value promise is valid (if the proof itself is also valid).

hansieodendaal: Yip, agree this does not add value

AaronFeickert: Pun intended, I hope!

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -727,24 +723,12 @@ mod test {
...
-        );
-        let hasher_bytes = DomainSeparatedHasher::<Blake256, CommitNonceHashDomain>::new()
-            .chain(commitment.as_bytes())
-            .finalize();
-        PrivateKey::from_bytes(hasher_bytes.as_ref()).expect("blake256 hash will succeed")
-    }
-
-    // For the faucet, the value of the commitment can be bound into the metadata signature using a deterministic nonce
+    // For the faucet, the value of the commitment can be bound into a value-revealing metadata signature
     fn do_alternate_range_proof_check(factories: &CryptoFactories, outputs: &[&TransactionOutput]) {
```
hansieodendaal, line 727: There is a current discussion about making range-proofs optional for all outputs by adopting these alternate range-proof checks and metadata signature constructs, not just for the genesis block faucet outputs. Pending that outcome, I imagine this function or a derivative will be elevated from test code to core code, alongside the optional range-proof implementation.

AaronFeickert:
Superseded by #5372. Closing.

PR #5329, opened 2023-04-17
fix: wallet ffi header file

SWvheerden:
Description
---
Fixes the header file by removing the auto-generated repc command as c does not have an option. 

How Has This Been Tested?
---
manual


brianp (approved):
LGTM utAck

PR #5332, opened 2023-04-17
docs: update readme to show esmeralda

SWvheerden:
Description
---
Update readme to mention Esmeralda rather than development
Replaced mentions of dibbler with Esmeralda in contributing doc


brianp (approved):
utACK

PR #5338, opened 2023-04-19
fix: remove wallet panic

SWvheerden:
Description
---
Removes a potential panic in the wallet that can occur

How Has This Been Tested?
---
manual and unit tests


hansieodendaal (dismissed):
ACK

PR #5349, opened 2023-04-21
feat: chat ffi

brianp:
Description
---
Create an FFI wrapper for a simple chat client

Motivation and Context
---
So we can chat on mobile and other non rust platforms.

How Has This Been Tested?
---
Cucumber

---
Review the cucumber tests as well

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/chat_ffi/build.rs
```diff
@@ -0,0 +1,55 @@
...
+// /// Find the location of the `target/` directory. Note that this may be
+// /// overridden by `cmake`, so we also need to check the `CARGO_TARGET_DIR`
+// /// variable.
+// fn target_dir() -> PathBuf {
+//     if let Ok(target) = env::var("CARGO_TARGET_DIR") {
+//         PathBuf::from(target)
+//     } else {
+//         PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap()).join("target")
+//     }
+// }
```
SWvheerden: delete?

base_layer/chat_ffi/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+/// ## Returns
+/// `*mut ChatClient` - Returns a pointer to a ChatClient, note that it returns ptr::null_mut()
+/// if config is null, an error was encountered or if the runtime could not be created
+///
+/// # Safety
+/// The ```client_destroy``` method must be called when finished with a TariWallet to prevent a memory leak
+#[no_mangle]
+pub unsafe extern "C" fn send_message(
+    client: &mut ClientFFI,
+    receiver_c_char: *const c_char,
```
SWvheerden:
```suggestion
    receiver_c_char: *TariAdresses,
```
Rather make it take a tari address here, then we can be sure it should work here and we don't have to expect it. We can effectively copy it over from the wallet ffi. It takes tari address ppointers.

base_layer/chat_ffi/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    let identity_path = PathBuf::from(
+        CStr::from_ptr(identity_file_path)
+            .to_str()
+            .expect("A non-null identity path should be able to convert to a string"),
+    );
+    let mut buf = Vec::new();
+    File::open(identity_path)
+        .expect("Can't open the identity file")
+        .read_to_end(&mut buf)
+        .expect("Can't read the identity file into buffer");
```
SWvheerden: can you try and make then return some error code rather than crashing on an except?

SWvheerden (commented):
This is looking good. Only thing I think we can improve on is removal of the excepts inside of the ffi

SWvheerden (approved):

SWvheerden:
ACK

PR #5356, opened 2023-04-26
fix: added missing log4rs features

sdbondi:
Description
---
Added missing log4rs features to applications which use these features.

Motivation and Context
---
Base node panics because the sample log config uses a number of log4rs features that are not compiled in.
The dependency was not added to the common lib as this adds extra dependencies to a lot of libs and it breaks wasm build.

How Has This Been Tested?
---
Base node runs

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
ACK

PR #5358, opened 2023-04-27
chore: update location of compiled bins in s3

brianp:
Description
---
Separate the bins by network type to make it easier to distinguish on the tari website for download.

Motivation and Context
---
It makes it easier to parse on the website, allowing us to deploy the new design which should help people disambiguate which binary they should download.

How Has This Been Tested?
---
#yolo

---
Know bash better than I do and validate with 100% certainty the change is most excellent

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -94,15 +94,23 @@ jobs:
           echo $tagnet
           # case match is not RegEx, but wildcards/globs
           case "$tagnet" in
-            v*-pre.*) TARI_NETWORK=esme
+            v*-pre.*)
+              TARI_NETWORK=esme
+              TARI_NETWORK_DIR=testnet
               ;;
-            v*-rc.*) TARI_NETWORK=nextnet
+            v*-rc.*)
+              TARI_NETWORK=nextnet
+              TARI_NETWORK_DIR=nextnet
               ;;
-            *) TARI_NETWORK=mainnet
+            *)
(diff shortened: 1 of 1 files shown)


.github/workflows/base_node_binaries.yml
```diff
@@ -94,15 +94,23 @@ jobs:
           echo $tagnet
           # case match is not RegEx, but wildcards/globs
           case "$tagnet" in
-            v*-pre.*) TARI_NETWORK=esme
+            v*-pre.*)
+              TARI_NETWORK=esme
+              TARI_NETWORK_DIR=testnet
```
sdbondi, line 99: Intentionally not `TARI_NETWORK_DIR=esme` I take it?

sdbondi (approved):
LGTM

.github/workflows/base_node_binaries.yml
brianp: Yeah, the whole need for a different var.

It's a gotcha of the network based compilations. You currently pass a real network when compiling, and it builds a bin for any network compatible with it. So passing `esme` builds a testnet bin comptabile for all testnets, but we can't pass `testnet` because `testnet` isn't a valid network. 

But we don't want every testnet bin to sit in an `esme` directory so here we are.

SWvheerden (approved):
ACK

SWvheerden:
ACK

PR #5359, opened 2023-04-28
feat!: range proof API change

AaronFeickert:
Description
---
Updates the handling of range proofs. Adds a seed nonce for mask recovery.

Motivation and Context
---
Recent changes affect how range proofs are handled. A [pending PR](https://github.com/tari-project/bulletproofs-plus/pull/29) in `bulletproofs-plus` enables [proof offloading](https://github.com/tari-project/bulletproofs-plus/issues/27), but requires the use of two seed nonces for mask recovery. A [corresponding PR](https://github.com/tari-project/tari-crypto/pull/177) in `tari-crypto` makes changes to accommodate this.

This work incorporates these changes. Notably, it requires the use of two seed nonces for mask recovery; one is known only to the signer, while another can be known to a helper device. This separation prevents the helper from learning the mask, and can be useful in cases where the signer is a device with limited computational ability.

How Has This Been Tested?
---
Existing tests pass.

---
Check that the derivation of seed nonces meets the goal of separate signer and helper roles. Assert that tests pass.

Breaking Changes
---
Existing range proofs that use mask recovery will verify, but will not recover the correct mask. Storage of seed nonces in the database will require that existing databases be cleared.

BREAKING CHANGE: Changes how masks are recovered from range proofs. Changes how seed nonces are stored in the database.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ dependencies = [
 
 [[package]]
 name = "aead"
-version = "0.5.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c192eb8f11fc081b0fe4259ba5af04217d4e0faddd02417310a927911abd7c8"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common",
  "generic-array",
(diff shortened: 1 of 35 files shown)


AaronFeickert:
Do not merge. This currently references a `tari-crypto` feature branch dependency, and is incomplete.

AaronFeickert:
Closing since the corresponding [Bulletproofs+ PR](https://github.com/tari-project/bulletproofs-plus/pull/29) has been closed.

PR #5367, opened 2023-05-05
fix: fix panic when no public addresses

stringhandler:
Description
---
If there are no public addresses specified, there was a panic. This is solved by not starting a liveness service if there is no public address.

NOTE: I have changed the reference on tari-crypto to be the lowest version that it works with (i.e. 0.16, instead of 0.16.*). This is the best practice for libraries, according to Rust, and also helps with mismatches on the tari-dan repo.

Motivation and Context
---
When using TCP, a node might not have a public address, this may cause a panic with the current code. AFAIK a comms node (base node, wallet, dan indexer) can still work without a public address, but it can only initiate outbound connections.

How Has This Been Tested?
---
existing tests

---
TODO: I'm busy checking this on the DAN indexer, which has no public address during testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/p2p/src/initialization.rs
+++ b/base_layer/p2p/src/initialization.rs
@@ -155,7 +155,7 @@ pub async fn initialize_local_test_comms<P: AsRef<Path>>(
 
     let comms = CommsBuilder::new()
         .allow_test_addresses()
-        .with_listener_address(node_identity.first_public_address())
+        .with_listener_address(node_identity.first_public_address().unwrap())
         .with_listener_liveness_max_sessions(1)
         .with_node_identity(node_identity)
         .with_user_agent(&"/test/1.0")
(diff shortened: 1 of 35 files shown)


sdbondi (dismissed):
utACK

base_layer/p2p/src/initialization.rs
```diff
@@ -155,7 +155,7 @@ pub async fn initialize_local_test_comms<P: AsRef<Path>>(
 
     let comms = CommsBuilder::new()
         .allow_test_addresses()
-        .with_listener_address(node_identity.first_public_address())
+        .with_listener_address(node_identity.first_public_address().unwrap())
```
SWvheerden, line 158: Can we not return an error here?

SWvheerden (commented):
Looks good, I would just like that one unwrap rather return an error

base_layer/p2p/src/initialization.rs
stringhandler: Yeah, my bad

stringhandler: This is test code actually

SWvheerden: awesome, cool I missed that

SWvheerden (approved):

PR #5372, opened 2023-05-08
feat: add optional range proof types

hansieodendaal:
Description
---
1. Added optional range proof types where the Bulletproof+ is no longer the only acceptable proof. 
   The `RevealedValue ` proof hinges on the prover revealing the commitment value as equal to the `minimum_value_promise`  
   and creating a special version of the metadata signature where the deterministic `ephemeral_commitment` nonce `r_a` is 
   set equal to zero (see [RFC-0182](https://rfc.tari.com/RFC-0182_CommitmentSignatures.html)).

1. As a result of the optional range proof, we do not use Bulletproof+ range-proof rewinding anymore. Encrypted value was replaced with encrypted openings and updated scanning and recovery based on encrypted openings.

1. Updated the `igor` and `esmeralda` genesis block faucets to make use of the new optional range proofs.

Motivation and Context
---
This will allow faucet outputs in the genesis block without the data overhead of a Bulletproof+ range proof for each. Other use cases include special transactions with revealed value outputs, but subject to the consensus constants where `RevealedValue` range proofs need to be switched on explicitly.

How Has This Been Tested?
---
Unit tests
Cucumber tests
System-level tests:
- On `igor` - base nodes, SHA3 mining, merged mining, console wallets
- On `esmeralda`  - base nodes, SHA3 mining, merged mining, console wallets

---
Recommend a system-level test

Breaking Changes
---

- [ ] None
- [X] Requires hard fork
- [ ] Other - Please specify

BREAKING CHANGE: New genesis block for `igor` and `esmeralda`, thus restart the blockchains.


base_layer/core/src/transactions/transaction_components/encrypted_value.rs
AaronFeickert, line 1: Recommend that the magic number `SIZE` be defined programmatically. You can use `std::mem::size_of::<Tag>()` for the `Poly1305` tag, which is the same size for all `ChaCha` and `XChaCha` variants.

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -79,6 +79,8 @@ impl From<Error> for EncryptionError {
 impl EncryptedValue {
     const TAG: &'static [u8] = b"TARI_AAD_VALUE";
 
+    /// Encrypt the value (with fixed length) using ChaCha20-Poly1305 with a fixed zero nonce to save space -
+    /// this design assumes a unique `encryption_key` every time and therefore uses a fixed nonce
```
AaronFeickert: Recommend adding a comment that key-commitment input pairs _must_ be unique, or the value will leak.

base_layer/core/src/transactions/transaction_components/encrypted_value.rs
```diff
@@ -98,6 +100,7 @@ impl EncryptedValue {
         Ok(EncryptedValue(data))
     }
 
+    /// Authenticate and decrypt the value
```
AaronFeickert: Recommend adding a comment that the design (like basically all existing AEADs) is not key committing, so the caller must not rely on successful decryption to assert that the expected key was used.

applications/tari_app_grpc/proto/transaction.proto
```diff
@@ -122,10 +122,20 @@ message OutputFeatures {
...
     // require a min maturity of the Coinbase_lock_height, this should be checked on receiving new blocks.
     uint64 maturity = 3;
+    // Additional arbitrary info in coinbase transactions supplied by miners
     bytes coinbase_extra = 4;
+    // Features that are specific to a side chain
     SideChainFeature sidechain_feature = 5;
+    // Encrypted Pedersen commitment openings (value and mask) for the output
+    EncryptedOpenings encrypted_openings = 6;
+    // The type of range proof used in the output
+    uint32 range_proof_type = 7;
```
AaronFeickert: Would it make more sense to make this a proper `enum` for clarity and consistency?

base_layer/core/src/proto/transaction.proto
```diff
@@ -99,8 +99,17 @@ message OutputFeatures {
     uint64 maturity = 3;
     // Additional arbitrary info in coinbase transactions supplied by miners
     bytes coinbase_extra = 4;
+    // Features that are specific to a side chain
     SideChainFeature sidechain_feature = 5;
-    // Additional
+    // Encrypted Pedersen commitment openings (value and mask) for the output
+    EncryptedOpenings encrypted_openings = 6;
+    // The type of range proof used in the output
+    uint32 range_proof_type = 7;
```
AaronFeickert: Would it make more sense to make this a proper `enum` for clarity and consistency?

base_layer/core/src/transactions/transaction_components/encrypted_openings_s.rs
```diff
@@ -0,0 +1,211 @@
...
+use tari_common_types::types::{Commitment, PrivateKey};
+use tari_crypto::{hash::blake2::Blake256, hashing::DomainSeparatedHasher};
+use tari_utilities::{safe_array::SafeArray, ByteArray, ByteArrayError};
+use thiserror::Error;
+use zeroize::Zeroize;
+
+use super::EncryptedOpeningsKey;
+use crate::transactions::{tari_amount::MicroTari, TransactionKdfDomain};
+
+const VALUE_SIZE: usize = 8;
```
AaronFeickert: You can define this as `std::mem::size_of::<u64>()` to avoid a magic number here.

base_layer/core/src/transactions/transaction_components/encrypted_openings_s.rs
```diff
@@ -0,0 +1,211 @@
...
+const KEY_SIZE: usize = 32;
```
AaronFeickert: This depends on the `PrivateKey` type, so hardcoding it here seems inelegant.

base_layer/core/src/transactions/transaction_components/encrypted_openings_s.rs
```diff
@@ -0,0 +1,211 @@
...
+const TAG_SIZE: usize = 16;
```
AaronFeickert: You can get this via `std::mem::size_of::<chacha20poly1305::Tag>()` to avoid a magic number here.

base_layer/core/src/transactions/transaction_components/encrypted_openings_s.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn from(err: Error) -> Self {
+        Self::EncryptionFailed(err)
+    }
+}
+
+impl EncryptedOpeningsS {
+    const TAG: &'static [u8] = b"TARI_AAD_VALUE_AND_MASK";
+
+    /// Encrypt the value and mask (with fixed length) using ChaCha20-Poly1305 with a fixed zero nonce to save space -
+    /// this design assumes a unique `encryption_key` every time and therefore uses a fixed nonce
```
AaronFeickert: Strictly speaking, it assumes a unique combination of `encryption_key` and `commitment`, since both are used in the internal key derivation.

AaronFeickert: Would also recommend a specific warning that reuse of such a combination leaks the resulting ciphertext.

base_layer/core/src/transactions/transaction_components/encrypted_openings_s.rs
```diff
@@ -0,0 +1,211 @@
...
+            aad: Self::TAG,
+        };
+        let aead_key = kdf_aead(encryption_key, commitment);
+        let cipher = ChaCha20Poly1305::new(GenericArray::from_slice(aead_key.reveal()));
+        let ciphertext = cipher.encrypt(&Nonce::default(), aead_payload)?;
+
+        EncryptedOpeningsX::from_bytes(ciphertext.as_slice())
+    }
+
+    /// Authenticate and decrypt the value and mask
```
AaronFeickert: Recommend a comment that this design is not key committing, so the caller should not rely solely on successful decryption to assert the expected key was used.

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+use tari_utilities::{safe_array::SafeArray, ByteArray, ByteArrayError};
+use thiserror::Error;
+use zeroize::Zeroize;
+
+use super::EncryptedOpeningsKey;
+use crate::transactions::{tari_amount::MicroTari, TransactionKdfDomain};
+
+const VALUE_SIZE: usize = 8;
+const KEY_SIZE: usize = 32;
+const TAG_SIZE: usize = 16;
```
AaronFeickert: See other comment about removing these magic numbers.

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+
+// Chacha error is not StdError compatible
+impl From<Error> for EncryptedOpeningsErrorX {
+    fn from(err: Error) -> Self {
+        Self::EncryptionFailed(err)
+    }
+}
+
+impl EncryptedOpeningsX {
+    const TAG: &'static [u8] = b"TARI_AAD_VALUE_AND_MASK";
```
AaronFeickert: Recommend a distinct AAD value for this, since the fixed-nonce version is technically a different algorithm.

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+            u64::from_le_bytes(value_bytes).into(),
+            PrivateKey::from_bytes(&mask_bytes)?,
+        ))
+    }
+}
+
+// Generate a ChaCha20-Poly1305 key from a private key and commitment using Blake2b
+fn kdf_aead(encryption_key: &PrivateKey, commitment: &Commitment) -> EncryptedOpeningsKey {
+    let mut aead_key = EncryptedOpeningsKey::from(SafeArray::default());
+    DomainSeparatedHasher::<Blake256, TransactionKdfDomain>::new_with_label("encrypted_value_and_mask")
```
AaronFeickert: Recommend a distinct domain separator for this, since the fixed-nonce version is technically a different algorithm.

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+impl From<Error> for EncryptedOpeningsErrorX {
+    fn from(err: Error) -> Self {
+        Self::EncryptionFailed(err)
+    }
+}
+
+impl EncryptedOpeningsX {
+    const TAG: &'static [u8] = b"TARI_AAD_VALUE_AND_MASK";
+
+    /// Encrypt the value and mask (with fixed length) using XChaCha20-Poly1305 with a secure random nonce
```
AaronFeickert: Recommend a clarifying comment here that, unlike the fixed-nonce version, this does not require or assume any uniqueness for `encryption_key` or `commitment`.

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+        let aead_key = kdf_aead(encryption_key, commitment);
+        let cipher = XChaCha20Poly1305::new(GenericArray::from_slice(aead_key.reveal()));
+        let mut ciphertext = cipher.encrypt(nonce_ga, aead_payload)?;
+        let mut ciphertext_integral_nonce = nonce.to_vec();
+        ciphertext_integral_nonce.append(&mut ciphertext);
+
+        EncryptedOpeningsX::from_bytes(ciphertext_integral_nonce.as_slice())
+    }
+
+    /// Authenticate and decrypt the value and mask
```
AaronFeickert: Recommend a comment that this design is not key committing, so the caller should not rely solely on successful decryption to assert the expected key was used.

AaronFeickert (changes requested):

base_layer/core/src/transactions/transaction_components/encrypted_openings_x.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+impl EncryptedOpeningsX {
+    const TAG: &'static [u8] = b"TARI_AAD_VALUE_AND_MASK";
+
+    /// Encrypt the value and mask (with fixed length) using XChaCha20-Poly1305 with a secure random nonce
+    pub fn encrypt_openings(
+        encryption_key: &PrivateKey,
+        commitment: &Commitment,
```
AaronFeickert: With the use of a secure random nonce, there's no need to use the commitment in the internal key derivation. You can certainly keep it for consistency if you like, since it doesn't hurt.

applications/tari_app_grpc/proto/transaction.proto
hansieodendaal: I would keep it like this as it is a simple conversion and consistent with `uint32 output_type = 2;`

base_layer/core/src/proto/transaction.proto
hansieodendaal: I would keep it like this as it is a simple conversion and consistent with `uint32 output_type = 2;`

hansieodendaal:
As per discussion with @stringhandler and @SWvheerden, it seems like a good strategy to implement the following:
- Replace `pub encrypted_value: EncryptedValue,` with `pub encrypted_openings: EncryptedOpeningsX,` in `TransactionOutput` and remove `encrypted_openings: Option<EncryptedOpeningsX>` from `OutputFeatures` (_this will add 24 + 32 = 56 bytes to `TransactionOutput`_)
- Change wallet recovery and one-sided scanning to make exclusive use of decrypting the `encrypted_openings` and verifying the commitment openings afterwards
- Use standard Bulletproof+ range proofs only, thus without _supplying the optional seed nonce that enables extracting_ ~embedding~ the mask for _non-aggregated_ Bulletproof+ rewinding ~(_embedding a different private key into the range proof could have interesting use cases in future_)~
- Also remove `EncryptedOpeningsS` as a contender

Any comments?

AaronFeickert:
> * Replace `pub encrypted_value: EncryptedValue,` with `pub encrypted_openings: EncryptedOpeningsX,` in `TransactionOutput` and remove `encrypted_openings: Option<EncryptedOpeningsX>` from `OutputFeatures` (_this will add 24 + 32 = 56 bytes to `TransactionOutput`_)

Does this mean that encrypting masks would be done for all output types, and not just those with no range proofs?

> * Change wallet recovery and one-sided scanning to make exclusive use of decrypting the `encrypted_openings` and verifying the commitment openings afterwards

I assume that for any output type relying on a wallet key appearing in the script (like legacy one-sided output), it's faster to scan for that directly. But for any output type where you'd otherwise rely on range proof rewinding, it should be _much_ faster to use authenticated decryption as an initial check.

> * Use standard Bulletproof+ range proofs only, thus without embedding the mask for bulletproof+ rewinding (_embedding a different private key into the range proof could have interesting use cases in future_)

This is certainly an option. For cases where the mask is deterministically derived by the sender, you could even get away with doing neither mask rewinding nor mask encryption. In that case, the recipient scans by first attempting authenticated value decryption, then derives the expected mask (perhaps via Diffie-Hellman), and finally checks that the commitment is well formed.

AaronFeickert:
I'm not sure this does everything you might hope.

I'll use the signer/helper terminology that I've used elsewhere, where the helper is a semi-trusted device that must not gain spend authority, but which is tasked with scanning for outputs intended for the signer.

For standard interactive outputs, it was presumably already possible for the helper to receive the value encryption key at some point from the signer. At this point, it could attempt value decryption as an initial check. Without the rewind key, it would be unable to extract the mask, which could only be done by the signer when it wishes to authorize a spend. Note that the helper has no way to itself determine if a commitment is spendable without the signer.

For legacy one-sided outputs, it was presumably already possible for the helper to scan for these. It could quickly check scripts to watch for the signer's wallet public key to appear. However, it could not decrypt values or perform mask extraction without being able to complete the sender's ephemeral Diffie-Hellman exchange, which would require access to the signer's wallet secret key.

For stealth one-sided outputs, it was not already possible for the helper to perform any scanning itself, since it could not check scripts for one-time public keys without completing the sender's ephemeral Diffie-Hellman exchange, which again would require access to the signer's wallet secret key. Further, it could not decrypt values or perform mask extraction for the same reason.

The proposed change does unify the handling of masks, which previously would have been different in the case of revealed-value outputs that do not have range proofs to use for mask extraction.

In the case of legacy and stealth one-sided outputs, the ability (or lack thereof) of the helper to scan for outputs does not change, as the value-with-mask encryption key would still be derived from a Diffie-Hellman exchange that requires the signer's wallet secret key to complete.

In the case of standard interactive outputs, the scanning process seems to be made _worse_ by the new design, depending on the handling of scripts. The use of a single encrypted field for value and mask means that a helper who can authenticate the ciphertext obtains both the value and mask, which grants it spend authority in the absence of any additional script public key whose corresponding secret key is known only to the signer. This seems out of line with the security goals.

AaronFeickert:
Another comment worth noting for standard interactive outputs is that if an additional script key (known only to the signer) is always required for spend authority in addition to knowledge of the commitment mask, the use of [range proof offloading](https://github.com/tari-project/bulletproofs-plus/issues/27) (which requires nontrivial state-machine interaction between signer and helper) would be unnecessary.

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -260,68 +351,27 @@ pub fn get_igor_genesis_block() -> ChainBlock {
...
-            let utxo: TransactionOutput = serde_json::from_str(line).unwrap();
-            utxos.push(utxo);
-        } else {
-            block.body.add_kernel(serde_json::from_str(line).unwrap());
-            block.header.kernel_mmr_size += 1;
-        }
-        counter += 1;
+    // Add faucet utxos - enable/disable as required
+    let add_faucet_utxos = true;
+    if add_faucet_utxos {
```
SWvheerden, line 356: Why put the if here, its not like we will have this be without faucets again

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -416,79 +466,30 @@ fn get_igor_genesis_block_raw() -> Block {
...
-    // block.header.output_mr =
-    //     FixedHash::from_hex("163304b3fe0f9072170db341945854bf88c8e23e23ecaac3ed86b9231b20e16f").unwrap();
-    // block.header.validator_node_mr =
-    //     FixedHash::from_hex("e1d55f91ecc7e435080ac2641280516a355a5ecbe231158987da217b5af30047").unwrap();
+    // lets get the block
+    let mut block = get_esmeralda_genesis_block_raw();
+
+    // Add faucet utxos - enable/disable as required
+    let add_faucet_utxos = true;
+    if add_faucet_utxos {
```
SWvheerden, line 474: same here

base_layer/core/src/transactions/transaction_components/unblinded_output_builder.rs
```diff
@@ -143,8 +144,10 @@ impl UnblindedOutputBuilder {
             &self.features,
             sender_offset_private_key,
             &self.covenant,
-            &self.encrypted_value,
+            &self.encrypted_openings,
             self.minimum_value_promise,
+            // TODO: Provide user options to use `RangeProofType::RevealedValue`
+            RangeProofType::BulletProofPlus,
```
SWvheerden: Dont make this hard-coded, either take it in here as a parameter, or let it be assigned to self, aka the builder.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -412,8 +420,10 @@ impl SenderTransactionInitializer {
                             &output_features,
                             &change_sender_offset_private_key,
                             &self.change_covenant,
-                            &encrypted_value,
+                            &encrypted_openings,
                             minimum_value_promise,
+                            // TODO: Provide user options to use `RangeProofType::RevealedValue`
+                            RangeProofType::BulletProofPlus,
```
SWvheerden: dont hardcode this in

base_layer/wallet/migrations/2023-05-15-112514_encrypted_openings/up.sql
```diff
@@ -0,0 +1,133 @@
+-- Any old 'outputs' will not be valid due to the new genesis block and changing 'encrypted_value' to
```
SWvheerden, line 1: I think you can just edit the initial in place as we cannot migrate old wallets, this will simply this code a bit

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -2042,11 +1973,13 @@ where
                 &output_features,
                 &sender_offset_private_key,
                 &covenant,
-                &encrypted_value,
+                &encrypted_openings,
                 minimum_value_promise,
+                // TODO: Provide user options to use `RangeProofType::RevealedValue`
+                RangeProofType::BulletProofPlus,
```
SWvheerden: dont hardcode this in

SWvheerden (commented):
This looks good

base_layer/wallet/migrations/2023-05-15-112514_encrypted_openings/up.sql
hansieodendaal: Ok, the idea here is to provide a slightly better UX than to panic if an older wallet is re-used with the new code. This effectively deletes all outputs, pending/completed transactions, scanned blocks and bunt proofs.

base_layer/core/src/blocks/genesis_block.rs
hansieodendaal: Ah, this is to get a bit of consistency in the `get_xxxx_genesis block_raw()` code, to do away from commenting code in and out. When faucets are no longer an option for nextnet and stagenet as well, we can just remove the if.

base_layer/core/src/blocks/genesis_block.rs
hansieodendaal: ditto

base_layer/core/src/transactions/transaction_components/unblinded_output_builder.rs
hansieodendaal: Upon looking closer, the `range_proof_type` is actually part of the `output_features`, so I was able to remove this parameter altogether.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: ditto

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: ditto

base_layer/core/src/transactions/transaction_components/encrypted_openings.rs
```diff
@@ -0,0 +1,289 @@
...
+        data_2.copy_from_slice(
+            bytes
+                .get(BORSH_64..SIZE)
+                .ok_or_else(|| EncryptedOpeningsError::IncorrectLength("Out of bounds 'data_2'".to_string()))?,
+        );
+        Ok(Self { data_1, data_2 })
+    }
+
+    /// Custom convert `EncryptedOpenings` to byte vector
+    pub fn as_byte_vector(&self) -> Vec<u8> {
```
stringhandler: Should be `to_byte_vec`. `as_byte_vec` implies that you are returning a reference

base_layer/core/src/chain_storage/tests/blockchain_database.rs
```diff
@@ -478,7 +478,7 @@ mod fetch_header_containing_kernel_mmr {
     fn it_returns_genesis() {
         let db = setup();
         let genesis = db.fetch_block(0, true).unwrap();
-        assert_eq!(genesis.block().body.kernels().len(), 1);
+        assert_eq!(genesis.block().body.kernels().len(), 2);
```
stringhandler, line 481: What's the reason for this?

stringhandler (changes requested):
I might have missed it, but I don't see the block validation to make sure a transaction is correct when the value is revealed

stringhandler:
I prefer the name `encrypted_data` to `encrypted_openings`, but either way I think the value and mask should be encrypted separately, although I suppose that's not directly used

base_layer/core/src/chain_storage/tests/blockchain_database.rs
stringhandler: Esme did not have a faucet, so this makes sense

stringhandler:
I think we need some new tests for:
1.  receiving onesided payments with revealed values
2.  block and mempool validation for revealed values
3. recovery of interactive and one sided payments with revealed values

hansieodendaal:
> I might have missed it, but I don't see the block validation to make sure a transaction is correct when the value is revealed

Functions `pub fn verify_range_proof(` and `pub fn batch_verify_range_proofs(` differentiate between doing Bulletproof+ range proof or revealed value range proof check based on the range proof type field.

hansieodendaal:
> I prefer the name `encrypted_data` to `encrypted_openings`, but either way I think the value and mask should be encrypted separately, although I suppose that's not directly used

- The choice of _openings_ was that it relates to the Pedersen commitment openings.
- There is no advantage in encrypting value and mask separately, it will just use more data (the tag size) and be more complex in the code.

base_layer/core/src/transactions/transaction_components/encrypted_openings.rs
```diff
@@ -0,0 +1,289 @@
...
+            msg: openings.as_slice(),
+            aad: Self::TAG,
+        };
+
+        // Produce a secure random nonce
+        let mut nonce = [0u8; size_of::<XNonce>()];
+        OsRng.fill_bytes(&mut nonce);
+        let nonce_ga = XNonce::from_slice(&nonce);
+
+        let aead_key = kdf_aead(encryption_key, commitment);
```
AaronFeickert: Depending on how [this draft RFC](https://github.com/tari-project/rfcs/pull/98) evolves, it may be more efficient to nix this KDF, and offload key generation to the caller. This would be particularly relevant for cases where `encryption_key` is fixed across multiple outputs, and where the addition of an internal KDF would add a lot of overhead.

AaronFeickert (changes requested):

SWvheerden (dismissed):

base_layer/core/src/transactions/transaction_components/encrypted_openings.rs
SWvheerden: This sounds good, but I think that's better for a next PR

hansieodendaal: Agreed for a next PR

AaronFeickert: Noted in a new [issue](https://github.com/tari-project/tari/issues/5393).

AaronFeickert:
> * There is no advantage in encrypting value and mask separately, it will just use more data (the tag size) and be more complex in the code.

You'd also need separate extended nonces, and would therefore bloat the ciphertext by an extra 40 bytes. Agreed that there is no advantage _unless_ there's a use case where an entity would need to be able to authenticate the ciphertext without learning, say, the commitment mask. And it sounds like that is not expected to be the case.

hansieodendaal:
> I think we need some new tests for:
> 
> 1. receiving onesided payments with revealed values
> 2. block and mempool validation for revealed values
> 3. recovery of interactive and one sided payments with revealed values

Must do ^^

stringhandler (approved):

AaronFeickert:
@hansieodendaal: Is it always guaranteed that a value-revealing range proof check will also have a separate check for a valid metadata signature? This is required according to the verification step described in [RFC-0182](https://rfc.tari.com/RFC-0182_CommitmentSignatures.html#value-revealing-protocol).

Previously, range proof verification was sufficient on its own to assert a commitment value is valid. Now, doing the zero-nonce scalar check in a value-revealing case is insufficient to conclude anything unless the entire metadata signature is also verified, which is not done internally to your range proof verifier.

My concern would be that some code path might assume validity without signature verification.

AaronFeickert:
> My concern would be that some code path might assume validity without signature verification.

Addressed in #5411 with a signature verification check.

PR #5374, opened 2023-05-10
ci(fix): libwallet build fix - remove changelog reference

leet4tari:
Description
Remove changelog reference for libwallet builds.

Motivation and Context
There are multiple changelog.md file, one for each MultiNet, which broke the libwallet build - tried to  include the base changelog.md, but does not exist any more.

How Has This Been Tested?
Built in local fork

See if the GitHub Action CI libwallet build is successful

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden:
We can use the same logic as we do to find the network type based on the version and then change the changelog name and pick the correct one for the network we are building?

leet4tari:
It's possible to use the same logic, but it does not map directly, so it would need to be extended and seeing that only libwallet referenced the ```changelog.md``` file, did not seem like a good time investment.  
  
Do you wish to do this?

SWvheerden:
I think it will be better to still upload the change logs

stringhandler (approved):

PR #5384, opened 2023-05-15
chore(nextnet): fix cargo toml pointing toward correct tag

SWvheerden:
Description
---
Fixes the monero-rs to not point toward main branch, but rather the correct commit
Motivation and Context
---
We should not use main branch as the target for a github repo

How Has This Been Tested?
---
manual


brianp (approved):
utACK LGTM

PR #5388, opened 2023-05-17
build(feature): add diag-utils for binary archives

leet4tari:
Description
Add a diag-utils archive, that can be downloaded per platform.

Motivation and Context
Including ```randomx-tests``` and friends. Needed to test randomx on Apple M1 for corner case bug, without waiting for block that triggered the bug.

Might be worth adding other tools or utilities and/or ReadMe file, but that would be better in the repo, than in the workflow.

How Has This Been Tested?
Built in local fork and tested downloads

### DIFF
--- a/.github/workflows/base_node_binaries.yml
+++ b/.github/workflows/base_node_binaries.yml
@@ -148,7 +148,7 @@ jobs:
         run: brew install cmake coreutils automake autoconf
 
       - name: Install Windows dependencies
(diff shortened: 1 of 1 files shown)


leet4tari:
Meant to add a screenshot of extra download archives.
Screenshot 2023-05-16 at 13 39 17

SWvheerden (approved):
Ack

PR #5390, opened 2023-05-17
feat: improved base node monitoring

stringhandler:
Add the ability to run tokio_console on the console_wallet. 

I found one case where the UI would use a lot of CPU, but I couldn't find the cause.

Reduced the UI caching time to be more responsive.

Renamed base_node_monitor_refresh_interval to base_node_monitor_max_refresh_interval and implemented a backoff strategy. 
If a new block is found when querying the base node, immediately query it again, backing off each time it has the same block. This allows you to update quickly if the node is syncing. Otherwise, backoff to the max interval. The default is 90 seconds.
(was previously 3 seconds)

Changed output and transaction manager service to rely on the base node service to send them an event when a new block is encountered. Both were doing this check themselves in addition to the check the base node service was doing.

The base node service also only checked on the height, now the hash is also checked.


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -521,21 +521,19 @@ where
...
-                    _ => false,
-                };
-                if trigger_validation {
-                    let _id = self.validate_outputs().map_err(|e| {
-                        warn!(target: LOG_TARGET, "Error validating  txos: {:?}", e);
-                        e
-                    });
-                }
-                self.last_seen_tip_height = state.chain_metadata.map(|cm| cm.height_of_longest_chain());
+            BaseNodeEvent::BaseNodeStateChanged(_state) => {
```
SWvheerden, line 497: This will stop the wallet from refreshing on node sync done, which might not be ideal

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -874,25 +874,20 @@ where
         >,
     ) {
         match (*event).clone() {
-            BaseNodeEvent::BaseNodeStateChanged(state) => {
-                let trigger_validation = match (self.last_seen_tip_height, state.chain_metadata.clone()) {
-                    (Some(last_seen_tip_height), Some(cm)) => last_seen_tip_height != cm.height_of_longest_chain(),
-                    (None, _) => true,
-                    _ => false,
-                };
+            BaseNodeEvent::BaseNodeStateChanged(_state) => {
```
SWvheerden, line 875: same here

SWvheerden (commented):
Are we fine with not triggering on node sync and only when it received a block via propagation? 
But apart from that it looks good

base_layer/wallet/src/output_manager_service/service.rs
stringhandler: This state is published when only the latency changes, so it was triggering much more frequently than needed

base_layer/wallet/src/transaction_service/service.rs
stringhandler: This will trigger the first time it gets a block from the node, so I think that makes sense

base_layer/wallet/src/base_node_service/handle.rs
```diff
@@ -55,8 +55,8 @@ impl fmt::Display for BaseNodeEvent {
             BaseNodeEvent::BaseNodeStateChanged(state) => {
                 write!(f, "BaseNodeStateChanged: Synced:{:?}", state.is_synced)
             },
-            BaseNodeEvent::NewBlockDetected(s) => {
-                write!(f, "NewBlockDetected: {}", s)
+            BaseNodeEvent::NewBlockDetected(_hash, height) => {
+                write!(f, "NewBlockDetected: {}", height)
```
hansieodendaal:
```suggestion
            BaseNodeEvent::NewBlockDetected(hash, height) => {
                write!(f, "NewBlockDetected: {} ({})", height, hash.to_hex())
```

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -198,24 +216,27 @@ where
         Ok(())
     }
 
-    async fn update_state(&self, new_state: BaseNodeState) {
+    // returns true if a new block, otherwise false
+    async fn update_state(&self, new_state: BaseNodeState) -> bool {
```
hansieodendaal, line 220: I think we should monitor changes to the base `node_id` here as well (not all the rest of  `BaseNodeState`) - if it changed then this function should return true as well. That way, if a base node is swapped, we can trigger `validate_outputs` to run upon that change,

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: See related comment ^^. I think we should add a `ConnectedBaseNodeChanged` event so we can refresh on that specifically.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: See related comment ^^.

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -198,24 +216,27 @@ where
...
         let mut lock = self.state.write().await;
-        let (new_block_detected, height) = match (new_state.chain_metadata.clone(), lock.chain_metadata.clone()) {
+        let (new_block_detected, height, hash) = match (new_state.chain_metadata.clone(), lock.chain_metadata.clone()) {
```
hansieodendaal, line 222:
```suggestion
        let (new_block_detected, connected_base_node_change_detected , height, hash) = match (new_state.chain_metadata.clone(), lock.chain_metadata.clone()) {
```

base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -198,24 +216,27 @@ where
...
-            (Some(new_metadata), _) => (true, new_metadata.height_of_longest_chain()),
-            (None, _) => (false, 0),
+            (Some(new_metadata), _) => (true, new_metadata.height_of_longest_chain(), *new_metadata.best_block()),
+            (None, _) => (false, 0, BlockHashType::default()),
         };
 
         if new_block_detected {
-            self.publish_event(BaseNodeEvent::NewBlockDetected(height));
+            self.publish_event(BaseNodeEvent::NewBlockDetected(hash, height));
         }
```
hansieodendaal, line 235:
```suggestion
if connected_base_node_change_detected {
  self.publish_event(BaseNodeEvent::ConnectedBaseNodeChanged)
}
```

hansieodendaal (commented):
Overall this is a good change. Please see the comments noted below, and maybe the PR title should be _feat: improved base node monitoring_.

base_layer/wallet/src/base_node_service/monitor.rs
stringhandler: It would be better to open a ticket for this, so that we add the handling of this event in the correct places.

I don't think it's necessary, because I believe the logic of the code is to send a `Default` state when the base node changes. This will not publish the NewBlockEvent, but when we do get a chain metadata from the new base node, if it has the same block hash, there is no need to update, but if it changes, it will trigger the appropriate jobs

hansieodendaal: Yeah, evaluating the block hash will sort this out

base_layer/wallet/src/output_manager_service/service.rs
hansieodendaal: Resolved with comment below

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: Resolved with comment below

base_layer/wallet/src/base_node_service/monitor.rs
hansieodendaal: Resolved with comment below

hansieodendaal (dismissed):
utACK

PR #4891, opened 2022-11-04
chore: delete some unneed, unsafe functions

stringhandler:
Description
---
Remove unneed and unsafe pieces of code

I am not sure what the intention was for `ScriptCommitment`, but it seems to be unused.

PR #5394, opened 2023-05-18
fix: prepend message to prevent arbitrary signing

stringhandler:
Description
---
Add a message to prevent arbitrary signing

Motivation and Context
---
This method is exposed via ffi and could allow users to sign transactions or arbitrary data

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - If this method was used externally, those methods will break.

### DIFF
--- a/base_layer/wallet/src/wallet.rs
+++ b/base_layer/wallet/src/wallet.rs
@@ -498,12 +498,14 @@ where
         Ok(tx_id)
     }
 
+    /// Signs a message, but prepends the message with the prefix "tari_signed_message" to prevent unintentional
+    /// signing of transactions, etc.
     pub fn sign_message(
         &mut self,
         secret: &RistrettoSecretKey,
         message: &str,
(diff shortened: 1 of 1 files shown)


AaronFeickert:
Wasn't this functionality removed in [PR 4891](https://github.com/tari-project/tari/pull/4891) a while back due to a previous unsafe implementation?

AaronFeickert (changes requested):
This can still technically induce collisions, albeit likely not ones that could lead to things like unintentional transaction signing.

A safer and more comprehensive fix would be to define a signature type that uses a different domain-separated challenge hasher. Then it's not possible (up to the collision resistance of the hasher) to produce collisions. This is supported by `tari-crypto`.

SWvheerden (approved):

AaronFeickert:
[This PR](https://github.com/tari-project/tari/pull/5400) applies domain separation as [referenced above](https://github.com/tari-project/tari/pull/5394#pullrequestreview-1432810894), and supersedes this one. It uses the `tari-crypto` signature API, which was built to handle use cases like this.

SWvheerden:
replaced by PR: #5400

PR #5399, opened 2023-05-19
fix: don't use in memory datastores for chat client dht in integration tests

brianp:
Description
---
Use filestore instead of in memory store for DHT tables.

Motivation and Context
---
I think at some point the reference to the store is being dropped causing the tables to no longer exist.

How Has This Been Tested?
---
Locally

---
Run the tests, watch CI.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):

SWvheerden:
ACK

PR #5400, opened 2023-05-19
fix: use domain separation for wallet message signing

AaronFeickert:
Description
---
Uses domain separation for wallet message Schnorr signatures to prevent context misuse. Supersedes [PR 5394](https://github.com/tari-project/tari/pull/5394).

Motivation and Context
---
Wallets can sign and verify arbitrary messages using their secret keys. Because such messages are signed using Schnorr signatures with a default Fiat-Shamir challenge hash domain, it may be possible to replay them in different contexts, with possibly dangerous consequences.

Another [pending PR](https://github.com/tari-project/tari/pull/5394) suggests prepending fixed data to wallet messages, but this can still lead to collisions.

Fortunately, the `tari-crypto` signature API provides a [handy type alias](https://github.com/tari-project/tari-crypto/blob/8a0823ff690f35409b64ad85a064033706f17580/src/ristretto/ristretto_sig.rs#L116) that makes domain separation straightforward. Using this, it is not possible to produce signature collisions (up to the collision resistance of the hash function itself). This is a safer and more idiomatic design.

How Has This Been Tested?
---
An existing test passes.

---
Confirm that the `tari-crypto` signature API is being used correctly, and that the macro-derived domain separator is unique within the codebase.

Breaking Changes
---
Existing signatures will fail to verify, but this is unlikely to be problematic.

### DIFF
--- a/base_layer/common_types/src/types/mod.rs
+++ b/base_layer/common_types/src/types/mod.rs
@@ -32,6 +32,7 @@ use tari_crypto::{
         RistrettoComAndPubSig,
         RistrettoPublicKey,
         RistrettoSchnorr,
+        RistrettoSchnorrWithDomain,
         RistrettoSecretKey,
     },
 };
@@ -43,6 +44,8 @@ pub use fixed_hash::{FixedHash, FixedHashSizeError};
 /// Define the explicit Signature implementation for the Tari base layer. A different signature scheme can be
 /// employed by redefining this type.
 pub type Signature = RistrettoSchnorr;
+/// Define a generic signature type using a hash domain.
+pub type SignatureWithDomain<H> = RistrettoSchnorrWithDomain<H>;
 /// Define the explicit Commitment Signature implementation for the Tari base layer.
 pub type ComAndPubSignature = RistrettoComAndPubSig;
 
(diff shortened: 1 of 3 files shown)


stringhandler (approved):
Nice. It does potentially limit verifying in non-rust code,  but I don't think that is a usecase for this in particular, and can be resolved with a WASM build at worst

AaronFeickert:
> Nice. It does potentially limit verifying in non-rust code, but I don't think that is a usecase for this in particular, and can be resolved with a WASM build at worst

Well, and there isn't a common standard for Schnorr-like signatures over arbitrary curves. Domain separation is handled in a really haphazard (and sometimes nonexistent) way across standards including things like [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032). Writing a compatible verifier outside of Rust wouldn't be that tricky,

AaronFeickert:
While it's not directly applicable here, [this](https://hdevalence.ca/blog/2020-10-04-its-25519am) is a good read on how subtle even something as seemingly simple as `ed25519` verification can get, and a good lesson on the importance of careful standardization.

Issue #5361, opened 2023-05-01
Some unmaintained `actions-rs` actions now fail to run in CI


AaronFeickert:
Apparently the [`actions-rs`](https://github.com/actions-rs/) suite of GitHub Actions has been unmaintained for some time now. Coupled with the fact that GitHub [previously indicated](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) a change to Actions runtime defaults, this means that certain `actions-rs` actions now [fail to run](https://github.com/tari-project/tari/actions/runs/4834737583?pr=5359).

Since it does not appear that `actions-rs` will see any additional action (pun intended), it may be time to migrate to something better supported. I've seen some folks recommend [`rust-toolchain`](https://github.com/dtolnay/rust-toolchain), but don't know if it can meet all CI needs.

AaronFeickert:
This is fixed in [PR 5362](https://github.com/tari-project/tari/pull/5362), but it still may be prudent to move away from `actions-rs` due to it being unmaintained for quite a while.

SWvheerden:
I have taken a quick look, but this is going to involve a much bigger PR to actually remove actions-rs. 
We might get around to just using the build in rust from the host machine and calling rustup update. 
But for now we just need to get the CI running again.

leet4tari:
We need to replace ```toolchain```, ```cargo``` and ```audit```. 
- ```toolchain``` 10x over 4 files - should be pretty straight forward with  [```actions-rust-lang/setup-rust-toolchain```](https://github.com/actions-rust-lang/setup-rust-toolchain) or [```dtolnay/rust-toolchain```](https://github.com/dtolnay/rust-toolchain)
- ```cargo``` 20x over 5 files - cross-compile needs some work, maybe [```houseabsolute/actions-rust-cross```](https://github.com/houseabsolute/actions-rust-cross)
- ```audit``` only one file to change - [```actions-rust-lang/audit```](https://github.com/actions-rust-lang/audit)

Do we want to do small changes, a little at a time or one big go? Maybe after the next release.

stringhandler:
One big go is fine IMO, no reason to be overly cautious in this area

PR #5401, opened 2023-05-22
ci(fix): replace abandoned actions-rs

leet4tari:
Description
Replace ```actions-rs``` with actively developed actions.  

Motivation and Context
Fix warnings in GitHub Action workflows - https://github.com/tari-project/tari/issues/5361

How Has This Been Tested?
Some parts have been tested in local fork.

Check actions working as expected and no warnings produced. 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
           cargo install cargo-machete
           cargo machete
   build-nightly:
-    name: cargo check for nightly
+    name: cargo check with nightly
     runs-on: [self-hosted, ubuntu-high-cpu]
     steps:
       - name: checkout
@@ -85,15 +85,17 @@ jobs:
       - name: rustup show
         run: rustup show
       - name: cargo check
-        run: cargo check --release --all-targets
+        run: cargo check --release --all-targets --locked
       - name: cargo check ffi separately
-        run: cargo check --release --package tari_wallet_ffi
+        run: cargo check --release --package tari_wallet_ffi --locked
   build-stable:
-    name: cargo check for stable
+    name: cargo check with stable
     runs-on: [self-hosted, ubuntu-high-cpu]
     steps:
       - name: checkout
         uses: actions/checkout@v3
+      - name: rust-toolchain.toml override by removing
+        run: rm -f rust-toolchain.toml
       - name: toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
@@ -108,7 +110,7 @@ jobs:
       - name: cargo check
         run: cargo check --release --all-targets --workspace --exclude tari_integration_tests --locked
       - name: cargo check ffi separately
-        run: cargo check --release --package tari_wallet_ffi
+        run: cargo check --release --package tari_wallet_ffi --locked
   licenses:
     name: file licenses
     runs-on: [ubuntu-20.04]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 ---
 name: CI
 
-"on":
+'on':
   workflow_dispatch:
   push:
     branches:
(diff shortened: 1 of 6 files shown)


stringhandler (dismissed):

AaronFeickert:
Would [`rust-toolchain`](https://github.com/dtolnay/rust-toolchain) do the trick as well? It's much more broadly used and written by a developer on the Rust team, so perhaps it's a better long-term choice if feasible to use.

AaronFeickert:
Looks like you're requesting the `stable` toolchain when loading the `rust-toolchain` action, which won't match the toolchain you later specify. The [documentation](https://github.com/dtolnay/rust-toolchain#inputs) for the action has advice for how to handle this properly.

AaronFeickert (changes requested):
Looks like you're requesting the `stable` toolchain when loading the `rust-toolchain` action, which won't match the toolchain you later specify. The [documentation](https://github.com/dtolnay/rust-toolchain#inputs) for the action has advice for how to handle this properly.

leet4tari:
> Looks like you're requesting the `stable` toolchain when loading the `rust-toolchain` action
```
dtolnay/rust-toolchain@stable
```
is the stable tag/branch of the ```dtolnay/rust-toolchain@stable```
ie:
```
        uses: dtolnay/rust-toolchain@stable
        with:
          components: rustfmt, clippy
          toolchain: ${{ matrix.builds.rust }}
```
```toolchain``` would be ```nightly-2022-11-03```, thou we might be able to drop that too and just be using the ```rust-toolchain.toml``` default.

**Comment by AaronFeickert** [comment]
Isn't the `stable` tag intended to correspond with the `stable` toolchain? Or am I misunderstanding the nature of that correspondence?

At any rate, the [documentation](https://github.com/dtolnay/rust-toolchain#inputs) recommends:
> When passing an explicit toolchain as an input instead of `@rev`, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

**Comment by leet4tari** [comment]
Updated, let's see how it runs

**Comment by AaronFeickert** [comment]
Looks like a `cargo check` [error](https://github.com/tari-project/tari/actions/runs/5145512657/jobs/9263254077?pr=5401#step:8:28) is present.

**Comment by leet4tari** [comment]
Need to replace all the ```actions-rs``` org actions, including the ```actions-rs/cargo```, thou I think we have another problem to look into, which might me related to ```error: usage of sparse registries requires "-Z sparse-registry"```

**Comment by leet4tari** [comment]
Should be ready for merge, builds all work in my own fork.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -34,109 +35,80 @@ jobs:
...
         with:
           toolchain: ${{ env.toolchain }}
-          profile: minimal
-          override: true
       - name: ubuntu dependencies
         run: |
           sudo apt-get update
           sudo bash scripts/install_ubuntu_dependencies.sh
+      - name: rustup show
+        run: rustup show
```
AaronFeickert, line 86: Is this just for human-readable debugging output of some kind?

.github/workflows/ci.yml
```diff
@@ -34,109 +35,80 @@ jobs:
...
-          override: true
       - uses: Swatinem/rust-cache@v2
       - name: ubuntu dependencies
         run: |
           sudo apt-get update
           sudo bash scripts/install_ubuntu_dependencies.sh
       - name: rustup show
-        run: |
-          rustup show
+        run: rustup show
```
AaronFeickert, line 109: See other comment.

leet4tari: was there before, I think it's for confirming what rust deps are at build.

.github/workflows/ci.yml
leet4tari: Match a good idea just to confirm what rust deps are in place at build time.

SWvheerden (approved):

PR #5402, opened 2023-05-23
test: improvements for debugging integration tests

brianp:
Description
---
This PR does a small handful of refactoring to help when debugging the integration tests:
- Better use of temporary directories for base node data
- Print base node identity files
- Better use of temporary directories for wallet data
- Move logs from each test run into the last run directory for easier to search and more isolated logs
- Fix 2 features so they'll run now
- Stop using in memory databases for DHT, it's possible for their reference to be dropped and the DHT tables to be wiped causing weird behaviour in the tests
- Stop manual cleaning of the World after a test runs, there is no need to do this, and it produced undesired behavior when printing panic data
- Miner was attempting to connect to Esme, ensure it stays on the localnetwork

Motivation and Context
---
The suite can be difficult to debug, hopefully this makes it easier

How Has This Been Tested?
---
Locally

---
See if it passes CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (dismissed):
Nice

SWvheerden (dismissed):
ACK

PR #5407, opened 2023-05-24
[WIP]Key management refactor

SWvheerden:
Description
---
This is a WIP PR for the upcoming key handling/key manager refactors. This PR supersedes:  #5274 
Some changes in #5274 are not yet in here such as the removal of private key API functions from the key_manager

The code had been changed to move the handling of all crucial private keys into a single contained service, the key manager.

Currently, the code is implemented in the following:
```rust
let private_key = key_manager.get_key(path);
let commitment = commitment_factory.commit(value, private_key);
tx.signature = signature.create(private_key, value, features, ...)
```

The new design will look like this:
```rust
let commitment = key_manager.get_commit(value, path);
tx.signature = key_manager.get_signaturet(path, message_hash);
```

And where is the wallet currently stores the all the private keys as encrypted bytes in its table, this is now just stored as a `KeyID`
```rust
pub enum KeyId {
    Default { branch: String, index: u64 },
    Imported { key: PublicKey },
}
```

So the actual field will look something like private_key: `default.spending_key.65` for example for the key generated on the spending key branch, with the index of 65. If it's an imported key the key will look as follows: `imported.xxxx` where xxxx is the public key hex. 

Because the key_manager will now require dedicated methods, we have created a core_key_manager that implements a base_layer-specific trait. This will still keep the key_manager service generic and open to use by for example the dan layer. 
But in keeping with the security, all functions now only return the public key and not the private key, 

Motivation and Context
---
The motivation is split between two reasons:
1: Limit use of private key to only the key manager, making it easier to audit for security. 
2: Make it easier to implement hardware wallet limiting, the changes to only the key manager and not the rest of the code base. 

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden:
Closing this as it was repleaced by #5462

PR #5411, opened 2023-05-25
feat: add metadata signature check

hansieodendaal:
Description
---
Added metadata signature check when verifying the revealed value proofs, just to be sure that the basis of the balance calculation check will have integrity.

Motivation and Context
---
See above.

How Has This Been Tested?
---
Added a new unit test.

---
Inspection

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/transaction_output.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_output.rs
@@ -261,6 +261,19 @@ impl TransactionOutput {
             &self.encrypted_data,
             self.minimum_value_promise,
         );
+        // Let's first verify that the metadata signature is valid
+        if !self.metadata_signature.verify_challenge(
+            &self.commitment,
+            &self.sender_offset_public_key,
+            &e_bytes,
+            &CommitmentFactory::default(),
+            &mut OsRng,
+        ) {
+            return Err(RangeProofError::InvalidRangeProof(
+                "Metadata signature not valid!".to_string(),
+            ));
+        }
+        // Now we can perform the balance proof
         let e = PrivateKey::from_bytes(&e_bytes).unwrap();
         let value_as_private_key = PrivateKey::from(self.minimum_value_promise.as_u64());
         let commit_nonce_a = PrivateKey::default(); // This is the deterministic nonce `r_a` of zero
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This supersedes #5412, which merely adds scary warning comments about verification.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -261,6 +261,19 @@ impl TransactionOutput {
             &self.encrypted_data,
             self.minimum_value_promise,
         );
+        // Let's first verify that the metadata signature is valid
```
AaronFeickert: Might be worth noting that this is (hopefully) an extra check that should also be done elsewhere, and is being done here for added safety.

base_layer/core/src/transactions/transaction_components/transaction_output.rs
```diff
@@ -825,6 +840,32 @@ mod test {
...
+            &factories,
+            MicroTari(20),
+            MicroTari(20),
+            RangeProofType::RevealedValue,
+        )
+        .unwrap();
+        assert!(output.verify_metadata_signature().is_ok());
+        assert!(output.revealed_value_range_proof_check().is_ok());
+
+        output.features.maturity += 1;
```
AaronFeickert, line 849: Can this overflow and fail the test? I know we've seen this elsewhere in tests.

AaronFeickert (changes requested):

base_layer/core/src/transactions/transaction_components/transaction_output.rs
hansieodendaal: For the test, it does not matter if it overflows, as long as it is changed, so that the features committed to in the metadata signature is different.

AaronFeickert: Right, I had been concerned that an overflow could panic and intermittently fail the test, depending on how the value was originally set. Looks like it's originally set to zero, so this can't happen. I'd do `output.features.maturity = 1` instead anyway; seems more elegant!

AaronFeickert (approved):
utACK

SWvheerden (approved):
Ack

PR #5412, opened 2023-05-25
docs: add range proof verification warnings

AaronFeickert:
Description
---
Adds warning comments about revealed-value range proof verification to help guard against misuse.

Motivation and Context
---
As noted in a [comment](https://github.com/tari-project/tari/pull/5372#issuecomment-1561898855) on #5372, verification of a revealed-value range proof requires separate verification of the corresponding metadata signature. Without such a check, range proof verification is inconclusive.

Because this is a major but subtle change to the guarantees provided by the single and batch range proof verification functions, this PR adds scary warning comments .

How Has This Been Tested?
---
The comments look sufficiently frightening.

---
Ensure that there are no code paths that assume revealed-value range proof validity without metadata signature verification.

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/transaction_output.rs
+++ b/base_layer/core/src/transactions/transaction_components/transaction_output.rs
@@ -210,6 +210,10 @@ impl TransactionOutput {
     }
 
     /// Verify that range proof is valid
+    ///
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This is superseded by #5411, which adds an internal signature check to range proof verification just to be safe.

PR #5415, opened 2023-05-25
chore(ci): move multinet envs target into single script

leet4tari:
Description
move multiNet envs target setup into single script

Motivation and Context
make future envs and maintenance easier.

How Has This Been Tested?
Built in local fork targeting testNet with tags

On release, check that envs that are needed are in place for workflow jobs.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
LGTM utAck

SWvheerden (approved):
ACK

PR #5416, opened 2023-05-26
feat: tx backend refactor

SWvheerden:
This PR refactors the entire transaction flow away from `unblindedoutputs`, to a new output `keyManagerOutput`.
This changes the use of private keys to keep all private key in the key manager service. 
This change is in line with PR #5407 . 
It includes much more big picture details as to why this is done

This ignores the wallet for now, this will be updated after

---
Large part of this porting unit tests to the new outputs. 

The import code to review here is the transaction protocol.


base_layer/core/src/core_key_manager/interface.rs
```diff
@@ -37,13 +37,21 @@ use crate::transactions::{
...
+        RangeProofType,
         TransactionError,
         TransactionInputVersion,
         TransactionKernelVersion,
         TransactionOutputVersion,
     },
 };
 
+#[derive(Clone, Copy, PartialEq)]
+pub enum TxoType {
```
stringhandler: This is a bad name, the "o" in Txo is for output. I would suggest a better name, but I can't tell how this is used

base_layer/core/src/core_key_manager/handle.rs
```diff
@@ -300,6 +302,8 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         total_excess: &PublicKey,
         kernel_version: &TransactionKernelVersion,
         kernel_message: &[u8; 32],
+        kernel_features: &KernelFeatures,
+        txo_type: TxoType,
```
stringhandler: This feels weird here. Why does the kernel signature need this info?

base_layer/core/src/core_key_manager/handle.rs
```diff
@@ -380,35 +386,42 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
...
         &self,
-        value_as_private_key: &PrivateKey,
         spending_key_id: &KeyId<PublicKey>,
-        sender_offset_private_key: &PrivateKey,
-        nonce_a: &PrivateKey,
-        nonce_b: &PrivateKey,
-        nonce_x: &PrivateKey,
-        challenge_bytes: &[u8; 32],
+        value_as_private_key: &PrivateKey,
+        ephemeral_commitment_nonce_id: &KeyId<PublicKey>,
```
stringhandler: I'm generally not in favour of having type aliases, but since we already have a PublicKey type alias, I would suggest making a type alias for `KeyId<PublicKey>` to eliminate this generic, Perhaps `type StoredKey = KeyId<PublicKey>`

stringhandler: Or StoredKeyId, or TariKeyId

base_layer/core/src/core_key_manager/service.rs
```diff
@@ -626,11 +660,22 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
...
+            PrivateKey::default() - self.get_private_key(spending_key).await?
+        };
         let private_nonce = self.get_private_key(nonce_id).await?;
-        let signing_key =
-            spending_private_key - &self.get_partial_private_kernel_offset(spending_key, nonce_id).await?;
+        // we cannot use an offset with a coinbase tx as this will not allow us to check the coinbase commitment
+        let signing_key = if kernel_features.is_coinbase() {
+            spending_private_key
+        } else {
+            spending_private_key - &self.get_partial_private_kernel_offset(spending_key, nonce_id).await?
```
stringhandler: would this not be 0 anyway?

base_layer/core/src/core_key_manager/service.rs
```diff
@@ -626,11 +660,22 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         total_excess: &PublicKey,
         kernel_version: &TransactionKernelVersion,
         kernel_message: &[u8; 32],
+        kernel_features: &KernelFeatures,
+        txo_type: TxoType,
     ) -> Result<Signature, TransactionError> {
-        let spending_private_key = self.get_private_key(spending_key).await?;
+        let spending_private_key = if txo_type == TxoType::Output {
```
stringhandler: I think there should be a comment here why this is negative for inputs

base_layer/core/src/proto/transaction.proto
```diff
@@ -120,7 +120,7 @@ message AggregateBody {
 // A transaction which consists of a kernel offset and an aggregate body made up of inputs, outputs and kernels.
 // This struct is used to describe single transactions only.
 message Transaction {
-    BlindingFactor offset = 1;
+    PrivateKey offset = 1;
```
stringhandler, line 123: I prefer `Scalar` here

base_layer/core/src/proto/transaction.proto
```diff
@@ -120,7 +120,7 @@ message AggregateBody {
...
     AggregateBody body = 2;
-    BlindingFactor script_offset = 3;
+    PrivateKey script_offset = 3;
```
stringhandler, line 125: Same

base_layer/core/src/core_key_manager/interface.rs
```diff
@@ -37,13 +37,21 @@ use crate::transactions::{
...
+    Input,
+    Output,
+}
+
 #[derive(Clone, Copy, EnumIter)]
 pub enum CoreKeyManagerBranch {
     DataEncryption,
```
stringhandler, line 60: Not part of this PR, I know, but this field name is vague and seems like it is general purpose

stringhandler: Same goes for "Nonce"

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -173,113 +162,167 @@ impl CoinbaseBuilder {
...
-            PrivateKey::from_bytes(spending_key_hash_bytes).map_err(|_| CoinbaseBuildError::InvalidSenderOffsetKey)?;
-        let sender_offset_public_key = PublicKey::from_secret_key(&sender_offset_private_key);
-        let covenant = self.covenant;
-
+        let public_nonce_id = self
+            .key_manager
+            .get_next_key_id(CoreKeyManagerBranch::Nonce.get_branch_key())
+            .await?;
+
+        let public_nonce = self.key_manager.get_public_key_at_key_id(&public_nonce_id).await?;
```
stringhandler: nit: this method could just be named `get_public_key()` since the key_id is implied by the parameter name

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -173,113 +162,167 @@ impl CoinbaseBuilder {
...
+        let public_spend_key = self.key_manager.get_public_key_at_key_id(&spending_key_id).await?;
+        let public_script_key = self.key_manager.get_public_key_at_key_id(&script_key_id).await?;
+
+        let kernel_signature = self
+            .key_manager
+            .get_partial_kernel_signature(
```
stringhandler: why not just pass in the key_id's here? `public_nonce_id` and `public_nonce` are passed in, when `public_nonce_id` could just be used

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -173,113 +162,167 @@ impl CoinbaseBuilder {
...
             &output_features,
-            &sender_offset_private_key,
             &covenant,
             &encrypted_data,
             minimum_value_promise,
-        )
-        .map_err(|e| CoinbaseBuildError::BuildError(e.to_string()))?;
+        );
+
+        // because the coinbase does not have a spending input with a script key, we use the output spending key as
```
stringhandler: @AaronFeickert is this secure?

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -173,113 +162,167 @@ impl CoinbaseBuilder {
...
+        // replacement to calculate the script offset
+        let sender_offset_public_key_id = self
```
stringhandler: nit: I would suggest adding a method that returns a pair (key_id, PublicKey) to make this more succinct and save on multiple service calls

base_layer/core/src/transactions/transaction_components/key_manager_output_builder.rs
```diff
@@ -150,33 +149,55 @@ impl KeyManagerOutputBuilder {
         &self.covenant
     }
 
-    pub async fn sign_as_sender_and_receiver<KM: BaseLayerKeyManagerInterface>(
-        &mut self,
+    pub async fn sign_as_sender_and_receiver_using_key_id<KM: BaseLayerKeyManagerInterface>(
```
stringhandler: I think the name can stay the same here

base_layer/core/src/transactions/transaction_components/key_manager_output_builder.rs
```diff
@@ -150,33 +149,55 @@ impl KeyManagerOutputBuilder {
...
+            .await?;
+        let ephemeral_pubkey_nonce_id = key_manager
+            .get_next_key_id(CoreKeyManagerBranch::Nonce.get_branch_key())
+            .await?;
+        let ephemeral_pubkey = key_manager.get_public_key_at_key_id(&ephemeral_pubkey_nonce_id).await?;
+        let ephemeral_commitment = key_manager
+            .get_metadata_signature_ephemeral_commitment(&ephemeral_commitment_nonce_id, self.features.range_proof_type)
+            .await?;
+        let metadata_signature = key_manager
+            .get_metadata_signature(
```
stringhandler, line 171: Since you are calling `get_metadata_signature` here on the keymanager service, why not just pass in all the inputs and return all the ephemeral keys generated? I say this because there is similar code in the coin base builder and you seem to be asking for data that is used in this call and then thrown away

stringhandler (commented):
I have only reviewed up until (excluding) the sender protocols.

Some notes so far:

1. The core_key_manager feels very strange being in the `core` crate, which is common to the base node and wallet. The base node would never make use of it, so it would be better suited in a wallet specific crate. If this is for tests, I'm sure it can be still be achieved with a dev_dependencies
2. The name `KeyManagerOutput` should use a name that refers to its stage in the life cycle, rather than an obscure reference to key manager. Perhaps UnfinallizedOutput, or MutableOutput
3. There are lots of names suffixed with `with_key_id` which are unnecessary because the parameter is typed already with this information

base_layer/core/src/core_key_manager/handle.rs
SWvheerden: it needs to swap the polarity the `signing key`. 

When signing a kernel with the input, we need to sign with `-k`
when signing a kernel with the output, er need to sign with `k`

base_layer/core/src/transactions/coinbase_builder.rs
SWvheerden: it should be yes,
currently, we have duplicate functions, for all functions still using the old API which uses private/public keys. 
Those will be renamed and the other ones removed before this goes into dev.

base_layer/core/src/core_key_manager/interface.rs
SWvheerden: I think we should think about and use the best names we can for the branches. 
Nonce I like as its used for nonces. 
But I am open to any suggestions to better names where we can use them

base_layer/core/src/proto/transaction.proto
SWvheerden: yeah, but I think we can leave that for this change. 
We just removed the duplicate type. 

we had
```
Type Blindingfactor = RistrettoSecretKey
Type PrivateKey  = RistrettoSecretKey
```

So we just removed the one

base_layer/core/src/core_key_manager/service.rs
SWvheerden: For kernels it needs to be 0. 
But the function to calculate the offset is oblivious to the context that is used to calculate the offset. 
It just knows that given two id's, it will calculate an offset. 

So we need to not call it if we sign a coinbase kernel.

base_layer/core/src/transactions/coinbase_builder.rs
SWvheerden: it will be good to know as thats how its currently done:
https://github.com/tari-project/tari/blob/f3833c9fc46d77fddaa7a23ef1d53ba9d860182a/base_layer/wallet/src/output_manager_service/service.rs#LL1024C10-L1024C10

base_layer/core/src/transactions/key_manager/wrapper.rs
```diff
@@ -279,73 +271,77 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
...
+        (*self.transaction_key_manager_inner)
             .read()
             .await
-            .get_script_signature(script_key_id, spend_key_id, value, tx_version, script_message)
+            .get_script_signature(script_key_id, spend_key_id, value, txi_version, script_message)
             .await
     }
 
-    async fn get_partial_kernel_signature(
+    async fn get_txo_kernel_signature(
```
stringhandler: I prefer the name `get_partial_kernel_signature`

base_layer/core/src/core_key_manager/service.rs
stringhandler: alright, not 100% convinced, but ok

base_layer/core/src/proto/transaction.proto
stringhandler: Alright, let's just do it in another PR

PR #5421, opened 2023-05-26
feat: update script offset

SWvheerden:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/core_key_manager/handle.rs
+++ b/base_layer/core/src/core_key_manager/handle.rs
@@ -269,11 +269,15 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
             .await
     }
 
-    async fn get_script_offset(&self, script_key_ids: &Vec<KeyId>, sender_offset_key_ids: &Vec<KeyId>) -> Result<PrivateKey, TransactionError> {
+    async fn get_script_offset(
+        &self,
+        script_key_ids: &Vec<KeyId>,
+        sender_offset_key_ids: &Vec<KeyId>,
+    ) -> Result<PrivateKey, TransactionError> {
         (*self.core_key_manager_inner)
             .read()
             .await
-            .get_script_offset(script_key_ids,sender_offset_key_ids)
+            .get_script_offset(script_key_ids, sender_offset_key_ids)
             .await
     }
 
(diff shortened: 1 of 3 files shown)


hansieodendaal (dismissed):
utACK

Do we have an issue with the integration tests?

---------
**Note for discussion outside of this PR**: 

I think we should rethink the use of deterministic keys in
- `async fn get_metadata_signature_ephemeral_private_key_pair`
- `async fn get_metadata_signature_ephemeral_private_key`
- `async fn get_private_kernel_signature_nonce`
- `pub async fn get_partial_private_kernel_offset`
- `pub async fn get_kernel_signature_nonce`

In my mind, those functions should be clearly named `_deterministic` and we should also include functions using secure randomness.

--------

hansieodendaal:
Based on:
```
[matrix] <Aaron Feickert> If you derive r using something like r = H(k, m)

[matrix] <Aaron Feickert> where H is a domain-separated cryptographic hash function, k is the secret key, and m is the message

[matrix] <Aaron Feickert> then you're fine

[matrix] <Aaron Feickert> This asserts that a nonce cannot be reused with the same key but a different message

[matrix] <Aaron Feickert> Some designs use this approach in part because it means you don't need to worry about a bad RNG

[matrix] <Aaron Feickert> and in general you shouldn't assume a computationally-limited hardware device has a good RNG

[matrix] <Aaron Feickert> Even if you don't directly reuse a nonce, it's possible to do key extraction if the RNG is bad (like linear congruential, for example)

[matrix] <Aaron Feickert> There was a recent preprint on exactly this

[matrix] <Aaron Feickert> So r must either be safely deterministic (like we just discussed) or super-duper random
```
I think we need to add the challenge in all the deterministic nonce functions.

hansieodendaal (commented):
I think we need to pass a reference of the input/output and then where the message is required the challenge should be calculated calling upon the reference.

hansieodendaal:
> I think we need to pass a reference of the input/output and then where the message is required the challenge should be calculated calling upon the reference.

Understood that the references may not be available at the time these functions are called. Let us park this for now.

hansieodendaal (approved):
utACK

PR #5425, opened 2023-05-30
feat: core key manager updates

hansieodendaal:
Description
---
- Move `KeyId` to `base_layer/key_manager` so that `KeyId` can be returned and passed without circular references.
- Constistent use of &str instead of String in key_manager.
- Use the new `CoreKeyManagerHandle` for key manager
- Use the new `CoreKeyManagerInitializer` to initialize services
- Removed `KeyManagerMock` in favour of a memory mapped SQL database for tests with function `create_test_key_manager()`

Motivation and Context
---
Part of the key manager refactoring

How Has This Been Tested?
---
Unit tests pass and were updated where necessary to use the new core key manager code

---
Inspection

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


CjS77:
You don't need to change this PR, but it's somewhat more idiomatic to use

`fn foo<S: Into<String>>(s: S) {...}` or `fn foo<S: AsRef<str>>(s: S) {...}` (depending on how the parameters are used in the function) than using the explicit `&str` or `String` types in parameter lists.

SWvheerden (approved):

PR #5426, opened 2023-05-30
feat: refactor configuration for chat so ffi can create and accept a config file

brianp:
Description
---
This refactoring come on the heels of making testing easier via the chat configuration. As a result we also solve the problem about peer seeds in FFI. Previously the ffi required passing in a collection of peer seeds. This wasn't in any way ideal. Now we take a standard tari configuration setup, which will utilize the DNS seeds in the default configuration.

Motivation and Context
---
ChatFFI had a bad way to connect to peers. It needed fixing. Also cleanup of the integration tests was desired and making configuration simpler is helpful.

How Has This Been Tested?
---
Locally and CI

No new test was added for the configuration creation yet. As that config defaults to real config, not local config.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "base_layer/tari_mining_helper_ffi",
     "clients/rust/base_node_grpc_client",
     "clients/rust/wallet_grpc_client",
+    # "clients/rust/web_chat",
     "comms/core",
     "comms/dht",
     "comms/rpc_macros",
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
     "base_layer/tari_mining_helper_ffi",
     "clients/rust/base_node_grpc_client",
     "clients/rust/wallet_grpc_client",
+    # "clients/rust/web_chat",
```
SWvheerden, line 17: remove?

SWvheerden (approved):
Ack

PR #5429, opened 2023-05-31
fix: fix custom wallet startup logic for console wallet

hansieodendaal:
Description
---
Added logic so that the wallet will not ask to connect to a detected base node if a custom base node has already been set in the database.

Motivation and Context
---
The custom-configured base node was ignored.

How Has This Been Tested?
---
System-level testing

---
- Start console wallet without `config.wallet.custom_base_node` set where a base node is running locally and accept the prompt to connect to the detected base node.
- With _Network_ tab configure a custom base node that is different to the one that is running locally.
- Exit.
- Restart the console walle.
- Verify that the user is not prompted again to connect to the local base node.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
Ack

PR #5432, opened 2023-05-31
fix(balanced_mp): removes some panics, adds some checks and new tests

sdbondi:
Description
---
- number of minor optimisations (e.g MergedBalancedBinaryMerkleIndexOrHash type vs tuple with index uint to Vec<u8> conversion)
- removed a number of possible panics
- adds integer underflow check in BalancedMerkleProof::verify
- adds edge case check for if all proofs to merge are zero height
- adds some extra tests for faulty proofs and "real world" data test
- minor memory optimisations
- format cargo.toml and reduce minimum required versions for some dependencies

Motivation and Context
---
None of these changes are game changing however they do address some important bugs that could prevent a bad actor from causing panics in a remote node.

@Cifko please check that I didn't break anything.

I wanted to investigate this because of an issue in a test in the DAN repo that fails with an invalid BMT even though all unmerged proofs are valid. I imported this data and strangely, the proof verifies (both in the previous code and code from this PR), so I was unable to track down the precise reason for the failure. 

How Has This Been Tested?
---
New and existing unit tests

---
Run tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4686,9 +4686,9 @@ checksum = "bebd363326d05ec3e2f532ab7660680f3b02130d780c299bca73469d521bc0ed"
 
 [[package]]
 name = "serde"
-version = "1.0.163"
+version = "1.0.150"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2113ab51b87a539ae008b5c6c02dc020ffa39afd2d83cffcb3f4eb2722cebec2"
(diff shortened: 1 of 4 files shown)


Cifko (approved):
Perfect, thanks

stringhandler (approved):

PR #5433, opened 2023-06-01
fix: zeroize openings, expose EncryptedData inner arrays

sdbondi:
Description
---
Zeriozes `openings` before returning from `EncryptedData::encrypt_data`
Adds `EncryptedData::to_inner_arrays() -> ([u8; 64], [u8; 16])`
Reduced heap allocations (up to 3 removed)

Motivation and Context
---
Openings contain the mask so must be zeroized.

For claim burn in the DAN repo (and potentially in general), we want to extract the inner arrays without a heap allocation or byte mangling.

How Has This Been Tested?
---
N/A

---
N/A

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
@@ -47,7 +47,7 @@ use tari_utilities::{
     ByteArrayError,
 };
 use thiserror::Error;
-use zeroize::Zeroize;
+use zeroize::{Zeroize, Zeroizing};
 
 use super::EncryptedDataKey;
 use crate::transactions::{tari_amount::MicroTari, TransactionSecureNonceKdfDomain};
@@ -84,11 +84,11 @@ impl EncryptedData {
         value: MicroTari,
         mask: &PrivateKey,
     ) -> Result<EncryptedData, EncryptedDataError> {
-        let mut openings = Vec::with_capacity(size_of::<u64>() + PrivateKey::key_length());
+        let mut openings = Zeroizing::new(Vec::with_capacity(size_of::<u64>() + PrivateKey::key_length()));
         openings.extend(value.as_u64().to_le_bytes());
         openings.extend(mask.as_bytes());
         let aead_payload = Payload {
-            msg: openings.as_slice(),
+            msg: &openings,
             aad: Self::TAG,
         };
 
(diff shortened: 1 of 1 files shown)


base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -99,6 +100,10 @@ impl EncryptedData {
         let aead_key = kdf_aead(encryption_key, commitment);
         let cipher = XChaCha20Poly1305::new(GenericArray::from_slice(aead_key.reveal()));
         let mut ciphertext = cipher.encrypt(nonce_ga, aead_payload)?;
+
+        // Openings contains the mask so we need to zeroize it
+        openings.zeroize();
```
AaronFeickert: Is this sufficient, since the previous line can fail without zeroizing the opening?

sdbondi: Well spotted!

AaronFeickert:
Why not go a step further and take advantage of const generics to use a single byte array for the encrypted data? This would let you reduce additional heap allocations like [this one](https://github.com/tari-project/tari/blob/50fa6b6bf74c1d14f1500f8e4891a2bcf1933a5f/base_layer/core/src/transactions/transaction_components/encrypted_data.rs#L117) and simplify the entire implementation significantly. All it requires is the `const-generics` feature flag on `borsh`, which presumably shouldn't cause any issues since integer const generics are already supported all the way up to stable.

AaronFeickert (changes requested):
The byte vector containing the decrypted mask is not zeroized.

AaronFeickert:
It looks like the [mask size](https://github.com/tari-project/tari/blob/50fa6b6bf74c1d14f1500f8e4891a2bcf1933a5f/base_layer/core/src/transactions/transaction_components/encrypted_data.rs#L56) is determined using `size_of`, which assumes that the encoded byte array length matches the internal storage size. This [happens to be the case](https://github.com/dalek-cryptography/curve25519-dalek/blob/618c5081f1ad7dea7c1c899e2c66fdc1cd7bce9c/src/scalar.rs#L699-L701) for a `RistrettoSecretKey`, but might not hold in general. (The other uses of `size_of` don't rely on trait implementations, so they seem fine.)

A safer approach is to use the change in a [pending PR](https://github.com/tari-project/tari-crypto/pull/181) in `tari-crypto`, which exposes the encoded key length as a constant.

AaronFeickert:
I went ahead and made the changes from my comments in [PR 5441](https://github.com/tari-project/tari/pull/5441), which supersedes this PR.

sdbondi:
Closing in favor of #5441

PR #5435, opened 2023-06-01
feat: wallet ffi can initialize and start indepedently 

### LINKED ISSUE #5377: wallet FFI should not await tor.
Currently when `fn wallet_create` is called, the function will block and await a tor connection before returning back and allowing interaction with the wallet. 
This should not be the case and the wallet should be created without having to wait for a tor connection

brianp:
Description
---
This PR renames the `wallet_create` to `wallet_init` which doesn't differ semantically but as there was a change under the hood it felt better to change the name as well. It also aligns with the console wallet process more.

The `wallet_init` function no longer starts the wallet transactions processes, or callback monitoring as it previously had. This is now reserved for later when calling `wallet_start`. Holding off on the network communications. 

Motivation and Context
---
The wallet freezes up for a duration while connecting to tor and the base node if the user has low connectivity. This prevents all other local actions from functioning. Instead we can return an initialized wallet that doesn't have connectivity allowing for local manipulation and request connectivity when the app is ready.

## Caveats
1) I think the ffi client will also require *not* adding a base node to the wallet after initialization. The ffi client should wait until the same time they'd like to start the wallet. Only then should they add a base node, then start. If a base node is added prior to calling start, it may attempt to make outside connections causing the same stalling problem.

2) `wallet_start` can only be called on an instance of a wallet a single time. It consumes the callback functions and can't be called again. Any subsequent call requires another `wallet_init` although this sounds like it follows the pattern of usage ios is already following, and shouldn't create a problem.

Closes #5377 

How Has This Been Tested?
---
CI only

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
Wallet team hates it. 

Going back to the drawing board.

brianp:
Closing in favour of https://github.com/tari-project/tari/pull/5437

PR #5441, opened 2023-06-05
feat: encrypted data cleanup

AaronFeickert:
Description
---
Cleans up and improves `EncryptedData` for simpler operations. Supersedes [PR 5433](https://github.com/tari-project/tari/pull/5433).

Motivation and Context
---
A pending PR updates `EncryptedData` to zeroize the plaintext commitment mask and expose the data as a tuple of byte arrays. However, there are more improvements and simplifications we can make.

This PR does several things:
- Uses the `const-generics` feature in `borsh` to allow for the use of a single byte array to represent the entire `EncryptedData` struct
- Performs zeroizing of the plaintext mask during encryption and decryption
- Uses the work in [another PR](https://github.com/tari-project/tari-crypto/pull/181) to get the byte-encoded length of the mask at compile time (instead of relying on a brittle use of `size_of` that isn't guaranteed to hold)
- Removes heap allocations during encryption and decryption
- Provides `as_bytes` to expose the `EncryptedData` as a byte slice
- Provides `to_bytes` to expose the `EncryptedData` as a byte array
- Cleans up comments and naming for clarity
- Updates faucet output JSON encoding to account for the changes (this does not affect genesis block data otherwise)

How Has This Been Tested?
---
Existing tests pass.

---
Confirm that existing tests pass, and that the listed features are implemented correctly.

Breaking Changes
---
There are no breaking changes. The existing `EncryptedData` API is unchanged.

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
@@ -51,15 +51,22 @@ use zeroize::{Zeroize, Zeroizing};
 use super::EncryptedDataKey;
 use crate::transactions::{tari_amount::MicroTari, TransactionSecureNonceKdfDomain};
 
-/// Total encrypted data size: nonce, encrypted value, encrypted mask, tag
-const SIZE: usize = size_of::<XNonce>() + size_of::<u64>() + PrivateKey::KEY_LEN + size_of::<Tag>();
+// Useful size constants, each in bytes
+const SIZE_NONCE: usize = size_of::<XNonce>();
+const SIZE_VALUE: usize = size_of::<u64>();
+const SIZE_MASK: usize = PrivateKey::KEY_LEN;
+const SIZE_TAG: usize = size_of::<Tag>();
+const SIZE_TOTAL: usize = SIZE_NONCE + SIZE_VALUE + SIZE_MASK + SIZE_TAG;
+
+// Number of hex characters of encrypted data to display on each side of ellipsis when truncating
+const DISPLAY_CUTOFF: usize = 16;
 
 #[derive(
     Debug, Copy, Clone, Deserialize, Serialize, PartialEq, Eq, Hash, BorshSerialize, BorshDeserialize, Zeroize,
 )]
 pub struct EncryptedData {
     #[serde(with = "tari_utilities::serde::hex")]
-    data: [u8; SIZE],
+    data: [u8; SIZE_TOTAL], // nonce, encrypted value, encrypted mask, tag
 }
 
 /// AEAD associated data
@@ -78,9 +85,9 @@ impl EncryptedData {
         mask: &PrivateKey,
     ) -> Result<EncryptedData, EncryptedDataError> {
         // Encode the value and mask
-        let mut bytes = Zeroizing::new([0u8; size_of::<u64>() + PrivateKey::KEY_LEN]);
-        bytes[..size_of::<u64>()].clone_from_slice(value.as_u64().to_le_bytes().as_ref());
-        bytes[size_of::<u64>()..].clone_from_slice(mask.as_bytes());
+        let mut bytes = Zeroizing::new([0u8; SIZE_VALUE + SIZE_MASK]);
+        bytes[..SIZE_VALUE].clone_from_slice(value.as_u64().to_le_bytes().as_ref());
+        bytes[SIZE_VALUE..].clone_from_slice(mask.as_bytes());
 
         // Produce a secure random nonce
         let nonce = XChaCha20Poly1305::generate_nonce(&mut OsRng);
@@ -93,11 +100,10 @@ impl EncryptedData {
         let tag = cipher.encrypt_in_place_detached(&nonce, ENCRYPTED_DATA_AAD, bytes.as_mut_slice())?;
 
         // Put everything together: nonce, ciphertext, tag
-        let mut data = [0u8; SIZE];
-        data[..size_of::<XNonce>()].clone_from_slice(&nonce);
-        data[size_of::<XNonce>()..size_of::<XNonce>() + size_of::<u64>() + PrivateKey::KEY_LEN]
-            .clone_from_slice(bytes.as_slice());
-        data[size_of::<XNonce>() + size_of::<u64>() + PrivateKey::KEY_LEN..].clone_from_slice(&tag);
+        let mut data = [0u8; SIZE_TOTAL];
+        data[..SIZE_NONCE].clone_from_slice(&nonce);
+        data[SIZE_NONCE..SIZE_NONCE + SIZE_VALUE + SIZE_MASK].clone_from_slice(bytes.as_slice());
+        data[SIZE_NONCE + SIZE_VALUE + SIZE_MASK..].clone_from_slice(&tag);
 
         EncryptedData::from_bytes(data.as_slice())
     }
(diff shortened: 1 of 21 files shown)


AaronFeickert:
NOTE: This PR relies on a [merged PR](https://github.com/tari-project/tari-crypto/pull/181) in `tari-crypto` (included here by manually dependency change) to properly expose encoded key lengths. It must not be merged until a new tag is issued and the corresponding crate update is done.

sdbondi:
> There are no breaking changes. The existing EncryptedData API is unchanged.

:+1: Just confirmed that the borsh and bincode encodings would not be affected by the change. 
```rust
      #[derive(Debug, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
        struct A {
            a: [u8; 10],
            b: [u8; 22],
        }
        #[derive(Debug, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
        struct B {
            a: [u8; 32],
        }

        let mut ba = vec![];
        BorshSerialize::serialize(&A { a: [0; 10], b: [0; 22] }, &mut ba).unwrap();
        let mut bb = vec![];
        BorshSerialize::serialize(&B { a: [0; 32] }, &mut bb).unwrap();

        assert_eq!(ba, bb);

        let mut ba = bincode::serialize(&A { a: [0; 10], b: [0; 22] }).unwrap();
        let mut bb = bincode::serialize(&B { a: [0; 32] }).unwrap();

        assert_eq!(ba, bb);
```

hansieodendaal:
Nice improvement - LGTM

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -114,28 +121,33 @@ impl EncryptedData {
...
-            aad: Self::TAG,
-        };
-        let decrypted_bytes = cipher.decrypt(nonce_ga, aead_payload)?;
-        let mut value_bytes = [0u8; VALUE_SIZE];
-        value_bytes.clone_from_slice(&decrypted_bytes[0..VALUE_SIZE]);
-        let mut mask_bytes = [0u8; KEY_SIZE];
-        mask_bytes.clone_from_slice(&decrypted_bytes[VALUE_SIZE..VALUE_SIZE + KEY_SIZE]);
+
+        // Decrypt in place using the buffer
+        cipher.decrypt_in_place(nonce_ga, ENCRYPTED_DATA_AAD, &mut buffer)?;
```
hansieodendaal: Maybe worthwhile adding this
```suggestion
        // Decrypt in place using the buffer, being careful to zeroize the buffer on failure
        cipher.decrypt_in_place(nonce_ga, ENCRYPTED_DATA_AAD, &mut buffer)
            .map_err(|e| {
                buffer.zeroize();
                EncryptedDataError::DecryptionFailed(e)
            })?;
```

hansieodendaal: I also could not test `fn esmeralda_genesis_sanity_check()` and `fn igor_genesis_sanity_check()` due to `error: failed to get `tari_crypto` as a dependency of package `tari_common v0.50.0-pre.2`, but I assume it will pass - the manual manipulation of the json files looks correct

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -84,25 +78,38 @@ impl EncryptedData {
...
-            aad: Self::TAG,
-        };
+        // Encode the value and mask
+        let mut bytes = Zeroizing::new([0u8; size_of::<u64>() + PrivateKey::KEY_LEN]);
+        bytes[..size_of::<u64>()].clone_from_slice(value.as_u64().to_le_bytes().as_ref());
+        bytes[size_of::<u64>()..].clone_from_slice(mask.as_bytes());
+
+        // Set up a buffer that also has room for the AEAD tag
+        let mut buffer = HeaplessVec::<u8, { size_of::<u64>() + PrivateKey::KEY_LEN + size_of::<Tag>() }>::new();
+        buffer.extend_from_slice(bytes.as_slice()).map_err(|_| EncryptedDataError::BufferError)?;
```
sdbondi: Should be able to copy the data straight into `buffer`
```suggestion
        // Set up a buffer containing the value, mask and with enough room for the AEAD tag
        let mut buffer = HeaplessVec::<u8, { size_of::<u64>() + PrivateKey::KEY_LEN + size_of::<Tag>() }>::new();
        buffer.extend_from_slice(value.as_u64().to_le_bytes().as_ref()).map_err(|_| EncryptedDataError::BufferError)?;
        buffer.extend_from_slice(mask.as_bytes()).map_err(|_| EncryptedDataError::BufferError)?;;
        
```

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: > Maybe worthwhile adding this

I assumed that failed decryption would either result in no plaintext being present in the buffer, or would be caused by manipulated ciphertext, both of which would mean zeroizing isn't very helpful. But you're right that it would be worthwhile just in case, even if it's a minor performance hit. Will update.

AaronFeickert: > I also could not test

Not sure why this didn't work for you, but I suppose we can rely on CI for this?

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: Good catch! Will update.

AaronFeickert: Working on a change that uses an array instead, which should allow for `Zeroizing` wrappers.

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: > Not sure why this didn't work for you

Looks like this was caused by my foolish deletion of the referenced branch after the corresponding PR was merged. Will update accordingly.

AaronFeickert: > Maybe worthwhile adding this

Working on a change that uses an array instead, which should render the closure moot.

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -29,50 +29,43 @@ use std::mem::size_of;
...
 use super::EncryptedDataKey;
 use crate::transactions::{tari_amount::MicroTari, TransactionSecureNonceKdfDomain};
 
-const VALUE_SIZE: usize = size_of::<u64>(); // 8 bytes
-const KEY_SIZE: usize = size_of::<PrivateKey>(); // 32 bytes
-const SIZE: usize = VALUE_SIZE + KEY_SIZE + size_of::<Tag>() + size_of::<XNonce>(); // 80 bytes
-const BORSH_64: usize = 64;
-const BORSH_X: usize = SIZE - BORSH_64; // 16 bytes
+/// Total encrypted data size: nonce, encrypted value, encrypted mask, tag
+const SIZE: usize = size_of::<XNonce>() + size_of::<u64>() + PrivateKey::KEY_LEN + size_of::<Tag>();
```
hansieodendaal: **Nit**: I would still prefer these and use them throughout as I think they enhance the readability of the code:
```suggestion
const NONCE_SIZE: usize = size_of::<XNonce>();
const VALUE_SIZE: usize = size_of::<u64>();
const KEY_SIZE: usize = PrivateKey::KEY_LEN;
const SIZE: usize = NONCE_SIZE + VALUE_SIZE + KEY_SIZE + size_of::<Tag>();
```

hansieodendaal (approved):
ACK

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: Good idea! Done.

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -84,25 +84,28 @@ impl EncryptedData {
...
+        // Encrypt in place
+        let tag = cipher.encrypt_in_place_detached(&nonce, ENCRYPTED_DATA_AAD, bytes.as_mut_slice())?;
+
+        // Put everything together: nonce, ciphertext, tag
+        let mut data = [0u8; SIZE_TOTAL];
+        data[..SIZE_NONCE].clone_from_slice(&nonce);
+        data[SIZE_NONCE..SIZE_NONCE + SIZE_VALUE + SIZE_MASK].clone_from_slice(bytes.as_slice());
+        data[SIZE_NONCE + SIZE_VALUE + SIZE_MASK..].clone_from_slice(&tag);
+
+        EncryptedData::from_bytes(data.as_slice())
```
sdbondi: Why the additional byte copy?

```suggestion
        Ok(Self { data })
```

sdbondi (dismissed):
Much better - thanks

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -173,11 +173,11 @@ impl EncryptedData {
             self.to_hex()
         } else {
             let encrypted_data_hex = self.to_hex();
-            if encrypted_data_hex.len() > 32 {
+            if encrypted_data_hex.len() > DISPLAY_CUTOFF {
```
hansieodendaal:
```suggestion
            if encrypted_data_hex.len() > 2 * DISPLAY_CUTOFF {
```

hansieodendaal (commented):
Thanks, readability is much improved, just the one comment

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: Unnecessary holdover from the previous design. Will update!

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
AaronFeickert: Good catch. Will fix.

hansieodendaal (approved):
ACK

SWvheerden (approved):
Ack

PR #5443, opened 2023-06-07
fix: use mined at timestamp in fauxconfirmation

### LINKED ISSUE #5167: [Wallet] One sided mined at timestamp is incorrect
The detected mined at the timestamp should be the block time and not unix epoch. 
Screenshot 2023-02-07 at 07 34 58

brianp:
Description
---
We were passing 0 every time when we should have been using the mined at timestamp.

Motivation and Context
---
Once unconfirmed one sided transactions had been confirmed we were dropping the proper mined at timestamp.

How Has This Been Tested?
---
Locally with one sided, and one sided stealth transactions between local wallets on igor.

---
Receive a one sided TX, and validate the mined at timestamp is present once confirmed.

Closes: #5167

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
Ack

PR #5444, opened 2023-06-08
feat: ui for template registration in console wallet

Cifko:
Description
---
Added code template registration tab to the console wallet.
The UI shows proper errors. It will not erase the fields when something is wrong.
Autofill everything that it can from the binary url (template name, version, type, repo link).
Computes the hash for the template.

Motivation and Context
---
The autofill works like this:
If the link is e.g. `https://github.com/tari-project/tari/blob/development/template.12.wasm`.
It will fill the repo as `https://github.com/tari-project/tari`
Template name `template`, template version `12`, template type `WASM:1`.
The hash is always computed as of now, because there is no way for the user to generate it manually.

How Has This Been Tested?
---
Manually.

---
Register a code template, and look into the VN UI.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
Ack

PR #5445, opened 2023-06-08
feat: peer sync limiter

brianp:
Description
---
This adds more criteria to the peers that get sent during a discovery, and on connect. Nodes should only send non-banned peers, who have been active within the week. They should send a max of 1000 peers per request. More discovery can occur when connecting to those new peers.

Motivation and Context
---
This helps to prevent trustworthy nodes from sharing bad peers, or junk data over the network. It limits an attack surface to a reach of 1, where a bad actor may sync 1000 max junk records to direct connections, but these connections would never propagate the junk peers forward as they'll never see them online.

How Has This Been Tested?
---
Unit test added for the filtering of peers. Untested in network enviornment

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -810,7 +810,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         let peers = self
             .comms
             .peer_manager()
-            .discovery_syncing()
+            .all()
             .await
             .map_err(|e| obscure_error_if_true(report_error_flag, Status::internal(e.to_string())))?;
         let peers: Vec<tari_rpc::Peer> = peers.into_iter().map(|p| p.into()).collect();
--- a/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/tari_base_node/src/grpc/base_node_grpc_server.rs
@@ -810,7 +810,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         let peers = self
             .comms
             .peer_manager()
-            .all()
+            .discovery_syncing()
             .await
             .map_err(|e| obscure_error_if_true(report_error_flag, Status::internal(e.to_string())))?;
         let peers: Vec<tari_rpc::Peer> = peers.into_iter().map(|p| p.into()).collect();
--- a/comms/core/src/peer_manager/peer_storage.rs
+++ b/comms/core/src/peer_manager/peer_storage.rs
@@ -296,15 +296,7 @@ where DS: KeyValueStore<PeerId, Peer>
         }
 
         let query = PeerQuery::new()
-            .select_where(|peer| {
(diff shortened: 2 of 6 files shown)


SWvheerden (dismissed):
Ack

applications/tari_base_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -806,7 +806,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         let peers = self
             .comms
             .peer_manager()
-            .all()
+            .discovery_syncing()
```
sdbondi: Peer sync doesnt happen through GRPC, you'll need to change the peer to peer RPC. 

`comms/dht/src/rpc/service.rs`

sdbondi (commented):
Looks good - unfortunately this will not stop a would-be peer spammer.

The receiver still trusts that they are sending them "real" peers. 

These are the options I can come up with now:
-  further limiting the number of peers we will accept (currently 1000, could maybe be 100)
- heuristic: limiting how many peer syncs we do per peer per unit of time, exceeding that may result in a timeout (ban)
- check with 1 or more other nodes in the neighbourhood if new peers are legit

EDIT(after properly reading the description lol): 
You are correct, this would prevent others from propagating it further. Good point! That'll help a lot. My node still goes out and connects to a bunch of nodes, but you're right, it would be much less effective at filling up peer lists.

This can go in once  `comms/dht/src/rpc/service.rs` is changed

hansieodendaal (commented):
Nice. 

We should be able to verify the chosen `PEER_MANAGER_SYNC_PEERS` and `PEER_ACTIVE_WITHIN_DURATION`. Lessers numbers in both cases would be more conservative but may be more constrictive at the same time as well.

CjS77:
ACK

comms/core/src/peer_manager/peer_storage.rs
```diff
@@ -278,10 +279,41 @@ where DS: KeyValueStore<PeerId, Peer>
...
+
+        let query = PeerQuery::new()
+            .select_where(|peer| {
+                features.map(|f| peer.features == f).unwrap_or(true) &&
+                    !peer.is_banned() &&
+                    peer.deleted_at.is_none() &&
+                    peer.last_seen_since().is_some() &&
+                    peer.last_seen_since().expect("Last seen to exist") <=
+                        Duration::from_secs(PEER_ACTIVE_WITHIN_DURATION) &&
+                    !excluded_peers.contains(&peer.node_id)
```
hansieodendaal: Maybe this can be DRYed up into `fn select_if(peer: Peer) -> bool`?
Repetition below.

hansieodendaal (approved):
utACK

comms/core/src/peer_manager/peer_storage.rs
brianp: I like the idea 👍🏻

sdbondi (approved):
LGTM

PR #5455, opened 2023-06-12
fix(nextnet): updates the block to the in use on stagenet now

SWvheerden:
Description
---
Sets the genesis block time to now and not a future data.
Release a new version

Motivation and Context
---
We cannot mine on stagenet till the genesisblock timestamp has past.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
LGTM

PR #5457, opened 2023-06-13
feat: sparse merkle trees

CjS77:
# Sparse Merkle trees 

## Description

A sparse Merkle tree is a Merkle tree where the non-empty nodes are stored in a map with the key being the hash
of the node and the value being the node itself.

Merkle trees are a mutable Merkle-ish tree structure, that supports full CRUD operations.

### Todo in this PR

* [x] Insert new values
* [x] Update existing values
* [x] Delete values
* [x] Inclusion proofs 
* [x] exclusion proofs.
* [x] Benchmarks

This implementation assumed that there is a key-value store elsewhere. Only the tree hashes themselves are
handled in this data structure.

When constructing a new tree, a hashing algorithm is specified. This is used to hash the non-leaf nodes. The
"values" provided to the tree must already be a hash, and should have been generated from a different hashing
algorithm to the one driving the tree, in order to prevent second pre-image attacks.

## Motivation and Context

tl;dr:
* Bandwidth  savings
* Scalability benefits
* Privacy benefits
* Compact inclusion _and_ exclusion proofs!

### Details
The SMT implementation is faster than `MutableMMR` up to around 10,000
nodes, and then the average depth starts to affect it. By 1,000,000
nodes, it is significantly slower than MMR.

This makes sense since MMR is basically O(1) for inserts, while SMT is O(log(n)).

You can see some results here:

```text
Starting: SMT: Inserting 1000000 keys
Finished: SMT: Inserting 1000000 keys - 1.921310493s
Starting: SMT: Calculating root hash
Tree size: 1000000. Root hash: 3e42ca40df366db52464c19b6ba71428976a56d7b120bc3c882fc29bf05dc1d7
Finished: SMT: Calculating root hash - 644.226062ms
Starting: SMT: Deleting 500000 keys
Finished: SMT: Deleting 500000 keys - 863.873761ms
Starting: SMT: Calculating root hash
Tree size: 500000. Root hash: 2a7b51f114a17c229f1067feb4ba5b6aad975689160a5eab0d90f89a3bcf09f8
Finished: SMT: Calculating root hash - 207.30907ms
Starting: SMT: Deleting another 500000 keys
Finished: SMT: Deleting another 500000 keys - 850.606501ms
Starting: SMT: Calculating root hash
Tree size: 0. Root hash: 0000000000000000000000000000000000000000000000000000000000000000
Finished: SMT: Calculating root hash - 3.892µs
Starting: MMR: Inserting 1000000 keys
Finished: MMR: Inserting 1000000 keys - 741.641704ms
Starting: SMT: Calculating root hash
Tree size: 1000000. Root hash: da6135ccaabf146024cae1b0e7ad6ba7e9dad79724fb9199b721d4cd243ba999
Finished: SMT: Calculating root hash - 8.649µs
Starting: MMR: Deleting 500000 keys
Finished: MMR: Deleting 500000 keys - 6.525858ms
Starting: SMT: Calculating root hash
Tree size: 500000. Root hash: fd60e168f27acba374109de9b8231e7252f0cfdf385f87dbfd92873d4956c995
Finished: SMT: Calculating root hash - 50.276µs
Starting: MMR: Deleting another 500000 keys
Finished: MMR: Deleting another 500000 keys - 6.862469ms
Starting: SMT: Calculating root hash
Tree size: 0. Root hash: 5d70f3177a0b46ea1b853c58d5e3f7d6e78cbc4149d71592bb6cea63d50ed96c
Finished: SMT: Calculating root hash - 93.618µs
```

SMT is taking 1.92s to insert 1 mil nodes, or 1.9us per node on average.

Still pretty rapid. Of course this nice thing here is that when we _delete_ all the nodes, the SMT hash is `000...`, which is the best part. The MMMR will never have a hash that's the same after adding, and then deleting the same node because of its bitmap tracking deleted entries :(

I also compared my SMT implementation to one on crates.io, and while it's hard to compare apples to apples (sparse-merkle-tree is really a KV store), my implementation is much, much faster. sparse-merkle-tree adds
10,000 nodes in ~1500ms, while my impl takes ~9ms plus 6ms for calculating the merkle root.

In terms of **space**, the SMT is much more efficient that MMMRs. To calculate the root of the UTXO set, for example, all you need is the UTXO set itself.

Imagine that there are 1,000,000 UTXOs, with another 2,000,000 UTXOs having being spent over the lifetime of the project.

If each UTXO hash is 32 bytes, you need serialize 32MB to recreate the merkle root for the SMT.

For the MMMR, you need all 3,000,0000 hashes (96MB) plus approximately 1MB for every million hashes in a bitmap to indicate which hashes have been deleted (3 MB) for a total of 100MB.

This only gets worse with time.

And keeping STXO hashes around forever -- that's not super great for privacy.

So even though the SMT is slower, it is still fast enough. The bandwidth savings are substantial and the privacy benefits are significant.

How Has This Been Tested?
---

Tests and benchmarks

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

This does not break anything right now. The intention is to write an RFC and then eventually replace the Output merkle tree with this data structure. That would be a breaking change though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,7 +279,7 @@ dependencies = [
  "matchit",
  "memchr",
  "mime",
- "percent-encoding 2.3.0",
+ "percent-encoding",
  "pin-project-lite",
  "rustversion",
  "serde",
@@ -686,6 +686,9 @@ name = "cc"
 version = "1.0.79"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
+dependencies = [
+ "jobserver",
+]
 
 [[package]]
 name = "cexpr"
@@ -1170,7 +1173,7 @@ dependencies = [
  "anes",
  "cast",
  "ciborium",
- "clap 4.3.0",
+ "clap 4.3.4",
  "criterion-plot",
  "is-terminal",
  "itertools",
@@ -2036,7 +2039,7 @@ version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a62bc1cf6f830c2ec14a513a9fb124d0a213a629668a4186f329db21fe045652"
 dependencies = [
- "percent-encoding 2.3.0",
+ "percent-encoding",
 ]
 
 [[package]]
@@ -2244,15 +2247,15 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.8.0"
+version = "0.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c7339329bfa14a00223244311560d11f8f489b453fb90092af97f267a6090ab0"
+checksum = "2994bee4a3a6a51eb90c218523be382fd7ea09b16380b9312e9dbe955ff7c7d1"
 dependencies = [
  "bitflags 1.3.2",
  "libc",
  "libgit2-sys",
  "log",
- "url 1.7.2",
+ "url",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


CjS77:
PR looks long, but it's mostly tests and benchmarks.

SWvheerden (commented):
utACK

Issue #5463, opened 2023-06-19
Remove genesis block coinbase


hansieodendaal:
Remove kernel and coinbase

hansieodendaal:
See #5464

PR #5464, opened 2023-06-19
feat!: remove genesis coinbase

hansieodendaal:
Description
---
Removed genesis block coinbase for all networks, and removed unused networks (`ridcully`, `dibbler`, `weatherwax` and `stibbons`) as part of the cleanup. Emission was changed to follow suit.

Motivation and Context
---
The mainnet genesis block should not have a coinbase transaction, as the open question would be who owns the spending key.

See #5463 and #5469.

How Has This Been Tested?
---
Added and updated unit tests.

---
Code walkthrough for the affected changes.
Verify applicable unit tests.

Breaking Changes
---

- [ ] None
- [X] Requires hard fork
- [ ] Other - Please specify

BREAKING CHANGE: This change requires a blockchain reset for all networks.

### DIFF
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -170,6 +170,4 @@ message ConsensusConstants {
     uint64 validator_node_registration_shuffle_interval_epoch = 33;
     /// An allowlist of range proof types
     repeated RangeProofType permitted_range_proof_types = 34;
-    /// Indicate if the genesis block has a coinbase
-    bool have_genesis_coinbase = 35;
 }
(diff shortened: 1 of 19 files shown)


applications/tari_app_grpc/proto/types.proto
```diff
@@ -170,4 +170,6 @@ message ConsensusConstants {
     uint64 validator_node_registration_shuffle_interval_epoch = 33;
     /// An allowlist of range proof types
     repeated RangeProofType permitted_range_proof_types = 34;
+    /// Indicate if the genesis block has a coinbase
+    bool have_genesis_coinbase = 35;
```
stringhandler: I don't think we need this, let's remove the coinbase for all networks

stringhandler (commented):
What's the reason for allowing the coinbase in some scenarios? IMO it would be easier to have it the same for all networks: No coinbase

hansieodendaal:
> What's the reason for allowing the coinbase in some scenarios? IMO it would be easier to have it the same for all networks: No coinbase

Was to have backward compatibility with dibbler etc., but I will remove those as well from the consensus constants.

SWvheerden (approved):

stringhandler (approved):

PR #5476, opened 2023-06-20
feat!: remove witness mr

SWvheerden:
Description
---
This removes the witness mr from the header and use of the witness data hash. Combines it into a single output hash.

Motivation and Context
---
We use two hashes for each output, the output hash and the witness hash. 
The output hash excludes the metadata signature and range proof. 
This is stored separately as the witness data in a separate Merkle mountain range. 
This complicates matters and there is no use for having two separate hashes. 

This makes it much clearer in having a single hash for each output. Rather than requiring two hashes for mutability per output. 

How Has This Been Tested?
---
Passes all unit tests

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

New genesis block and new wallet db


hansieodendaal (approved):
utACK

stringhandler (approved):
utACK.

PR #5477, opened 2023-06-20
feat: proof of work audit part 1

hansieodendaal:
Description
---
Added checks and balances for POW, mainly within consensus constants and added some documentation

Motivation and Context
---
Preparation for code audit

How Has This Been Tested?
---
Added unit tests

---
Code walkthrough
Review the new unit test and related code

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -387,30 +387,33 @@ impl ConsensusConstants {
             vn_registration_shuffle_interval: VnEpoch(100),
             coinbase_output_features_extra_max_length: 64,
         }];
-        assert_hybrid_pow_constants(
-            &consensus_constants,
-            &[120],
-            &[60],
-            &[40],
-            CheckDifficultyRatio::No,
-            CheckLwmaRatios::Yes,
-        );
+        #[cfg(any(test, debug_assertions))]
+        assert_hybrid_pow_constants(&consensus_constants, &[120], &[60], &[40], CheckDifficultyRatio::No);
         consensus_constants
     }
 
     pub fn igor() -> Vec<Self> {
+        // `igor` is a test network, so calculating these constants are allowed rather than being hardcoded.
+        let monero_split: u64 = 60;
+        let sha3_split: u64 = 100 - monero_split;
+        let monero_target_time = 20;
+        let sha3_target_time = monero_target_time * (100 - sha3_split) / sha3_split;
+        let target_time: u64 = (monero_target_time * sha3_target_time) / (monero_target_time + sha3_target_time);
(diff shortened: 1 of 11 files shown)


base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -32,7 +38,25 @@ pub struct LinearWeightedMovingAverage {
...
+        );
+        assert_eq!(
+            target_time * LWMA_MAX_BLOCK_TIME_RATIO,
+            max_block_time,
+            "LinearWeightedMovingAverage::new(...) expected max_block_time to be {} times greater than target_time, \
+             off by {}s",
+            LWMA_MAX_BLOCK_TIME_RATIO,
+            max(target_time * LWMA_MAX_BLOCK_TIME_RATIO, max_block_time) -
+                min(target_time * LWMA_MAX_BLOCK_TIME_RATIO, max_block_time)
+        );
```
SWvheerden: Using assert! and assert_eq! in prodection code is not ideal. Rather return an error here than a panic. 
Assertions are fine for tests, but this is not for tests

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -675,6 +697,109 @@ impl ConsensusConstants {
...
+
+// Assert the hybrid POW constants.
+// Note: The math and constants in this function should not be changed without ample consideration that should include
+//       discussion with the Tari community, modelling and system level tests.
+// For SHA3/Monero to have a 40/60 split:
+//   > sha3_target_time = monero_target_time * (100 - 40) / 40
+//   > monero_target_time = sha3_target_time * (100 - 60) / 60
+//   > target_time = monero_target_time * sha3_target_time / (monero_target_time + sha3_target_time)
+// `CheckDifficultyRatio` and `CheckLwmaRatios` are optional for internal testing (Network::LocalNet and Network::Igor).
+fn assert_hybrid_pow_constants(
```
SWvheerden, line 697: These asserts are not ideall in prodection code. 
I would say say move this to a unit test, that tests the constants of each network. 
We dont have to run assert and re check them every time on startup. 
A unit test will check them every time they are changed and checked into prodection.

SWvheerden (commented):
I dont like the asserts in production code. To rely on panics to enforce safety is the wrong way around as we have no control over how the panic will be handled and what thread it will shutdown. A panic does not guarentee a program shutdown, only a thread shutdown. 

For cases where we want to check input from the user rather make it an error forcing the user to implement some form of error handling and safely shutdown if required.

SWvheerden (approved):

PR #5479, opened 2023-06-21
ci(fix): build only bins and include in archive/packages

leet4tari:
Description
Limit build to only binaries that are included in the archives and packages

Motivation and Context
In the name of efficiency, which save about 30% of build time if we only build the binaries packaged

How Has This Been Tested?
Built in local fork and used archive to run a local node, wallet and miner.

---

Breaking Changes
---
Removed tari_mining_helper_ffi from archives and packages

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify


CjS77 (approved):
LGTM

PR #5480, opened 2023-06-21
feat: use caching across entire ci workflow

CjS77:
Github actions allows caching, which can dramatically speed up CI runtimes by caching compiled versions of our dependencies.

Caches were not being used in every step of the `ci` workflow. This PR adds this everywhere, so expect to see significant speedups.

The caching works well, so no need to run the "build_dependencies" step, which was a poor man's attempt to perform caching by create a docker container with just the compiled dependencies. At best this wasn't really working, and at worse was busting caches.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo bash scripts/install_ubuntu_dependencies.sh
+      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


leet4tari:
LGTM

SWvheerden (approved):

AaronFeickert:
Any idea what the speedups look like in practice for this?

PR #5481, opened 2023-06-21
perf: speed up slow test

CjS77:
`too_many_inputs` just tests the number of inputs in a tx, not that they're valid. Moving the expensive input generation call out of the hot loop reduces the test speed from 5-7minutes to 0.6s

You're welcome :)

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (approved):
Nice one

PR #5483, opened 2023-06-21
chore: small updates to consensus

SWvheerden:
Description
---
Changes on builder method to through an error on wrong input parameters
Adds some comments

Motivation and Context
---
This is done as part of internal audit and code cleanup

How Has This Been Tested?
---
Unit tests

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -107,12 +107,14 @@ pub struct ConsensusConstants {
     vn_epoch_length: u64,
     /// The number of Epochs that a validator node registration is valid
     vn_validity_period_epochs: VnEpoch,
+    /// The min amount of microTari to deposit for a registration transaction to be allowed onto the blockchain
     vn_registration_min_deposit_amount: MicroTari,
+    /// The period that the registration funds are required to be locked up.
     vn_registration_lock_height: u64,
+    /// The period after which the VNs will be reshuffled.
     vn_registration_shuffle_interval: VnEpoch,
 }
 
-// todo: remove this once OutputFeaturesVersion is removed in favor of just TransactionOutputVersion
 #[derive(Debug, Clone)]
 pub struct OutputVersionRange {
     pub outputs: RangeInclusive<TransactionOutputVersion>,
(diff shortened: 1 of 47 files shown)


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -211,12 +213,12 @@ impl ConsensusConstants {
...
     /// Maximum transaction weight used for the construction of new blocks. It leaves place for 1 kernel and 1 output
+    /// with default features
     pub fn get_max_block_weight_excluding_coinbase(&self) -> u64 {
-        self.max_block_transaction_weight - self.coinbase_weight()
+        self.max_block_transaction_weight - self.calculate_1_output_kernel_weight()
     }
 
-    pub fn coinbase_weight(&self) -> u64 {
-        // TODO: We do not know what script, features etc a coinbase has - this should be max coinbase size?
+    fn calculate_1_output_kernel_weight(&self) -> u64 {
```
sdbondi, line 218: should we not just pass in the OutputFeatures and script?

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -202,11 +206,15 @@ impl ConsensusManagerBuilder {
     }
 
     /// Builds a consensus manager
-    pub fn build(mut self) -> ConsensusManager {
+    pub fn build(mut self) -> Result<ConsensusManager, String> {
```
sdbondi: String does not implement Error so preferable not to use it as an error. I think create a thiserror type. You could use anyhow error in the GRPC at the application level if that is easier.

sdbondi (dismissed):
utACk

Just nitting on the use of String as an Err type

base_layer/core/src/consensus/consensus_constants.rs
SWvheerden: The wallet calculates thats details only after it gets a template from the miner. 
To get the template we need to know how large it will be beforehand. 
Its a potential bug atm I logged it in the audit findings. 
But for this PR I only changed to name of the function to correctly state what it does

sdbondi (dismissed):

sdbondi:
Looks like this was merged with a compile error in FFI - feature flags in core...

PR #5486, opened 2023-06-22
fix(core): minor audit improvements

sdbondi:
Description
---
fix(core): make max_input_maturity fallible
add details on the usage of core hash domains
include details on all fallible transaction input methods
minor comment update in coinbase_builder
fix(core/aggregate_body): remove unused buggy function min_input_maturity
tests(core/aggregate_body_validator): add test for unsorted aggregate body and comment on new_sorted_unchecked call
chore(core/transaction_builder): remove unnecessary mutation for more ergonomic api
chore(core): improve clarity of check_coinbase_output
fix(core): remove unused git link for monero

Motivation and Context
---
max_input_maturity: should be fallible rather than possibly return the incorrect result.
min_input_maturity: returns an incorrect result (u64::MAX) if there are no inputs. Removed because it was unused.
check_coinbase_output: had unwraps and a comment indicating something _should_ happen to prevent the panic. For clarity, updated the code so that it can easily be seen as correct.
transaction_builder: add components with a mutable reference, possibly unclear how the references are mutated. More ergonomic and useful to pass in an iterator.
monero dependency was using a git link and `version` ambiguously. Cargo seemed to use the crate and ignore the git field. 
Added an explicit test for validation failure on unordered tx components in an aggregate body.

How Has This Been Tested?
---
Existing tests and new unit test

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


sdbondi:
This PR is broken down into commits. Each one can be reviewed independently. Happy to revert any one of them as necessary.

base_layer/core/src/transactions/transaction_components/transaction.rs
```diff
@@ -97,19 +92,19 @@ impl Transaction {
...
-    pub fn min_spendable_height(&self) -> u64 {
+    pub fn min_spendable_height(&self) -> Result<u64, TransactionError> {
         self.body.min_spendable_height()
     }
 
-    /// This function adds two transactions together. It does not do cut-through. Calling Tx1 + Tx2 will result in
-    /// vut-through being applied.
+    /// This function adds two transactions together. It does not do cut-through but simply sums up the offsets and
+    /// extends all inputs, outputs and kernels.
     pub fn add_no_cut_through(mut self, other: Self) -> Self {
```
SWvheerden:
```suggestion
    pub fn add(mut self, other: Self) -> Self {
```
I dont see why we need to reference cut-through here as Tari does not use or support the use of cut-through anymore

SWvheerden: perhaps best to just delete this and use the impl add

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -72,38 +72,49 @@ pub(crate) struct OutputPair {
...
+    /// The TransactionOutput received from the recipient.
     pub recipient_output: Option<TransactionOutput>,
+    /// The partial kernel excess received from the recipient.
     pub recipient_partial_kernel_excess: PublicKey,
+    /// The partial kernel signature received from the recipient.
     pub recipient_partial_kernel_signature: Signature,
+    /// The partial kernel offset received from the recipient.
     pub recipient_partial_kernel_offset: PrivateKey,
-    // The sender's portion of the public commitment nonce
+    /// The change output details. This may be None if no change is required.
```
SWvheerden, line 87:
```suggestion
    /// The change output details. Always required
```

SWvheerden (commented):
Looks good, one nit, and one wrong comment

base_layer/core/src/transactions/transaction_protocol/sender.rs
sdbondi: In that case, I'll have to see how we handle none

SWvheerden: On build it will throw an error

sdbondi: That doesnt appear to be true. See `try_build_transaction` and `validate`

SWvheerden: You are correct, dont know why I got that idea in my head.

base_layer/core/src/transactions/transaction_components/wallet_output.rs
```diff
@@ -247,7 +247,7 @@ impl WalletOutput {
         ))
     }
 
-    pub async fn as_transaction_output<KM: TransactionKeyManagerInterface>(
+    pub async fn to_transaction_output<KM: TransactionKeyManagerInterface>(
```
SWvheerden, line 250: I dont agree with this change. 
`to_T` implies it takes `self`
`as_T` implies it takes `&self`

sdbondi: nope
https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

sdbondi: `into_` - takes self
`to_` - usually `&self` some greater cost to the conversion
`as_` - usually `&self` cheap/basically free conversion

SWvheerden: Thanks, I stand corrected.

SWvheerden (approved):

base_layer/core/src/transactions/transaction_components/wallet_output.rs
AaronFeickert: TIL!

PR #5487, opened 2023-06-22
fix: coinbase recovery

SWvheerden:
Description
---
Fixes the coinbase recovery. 
Speeds up recovery searching by looking on the private key.

Motivation and Context
---
The current code can recover the private key of the coinbase, but it searches in the `CommitmentMask` till index 1_000_000 not finding a match. Then it imports the key. For coinbase keys, the keys are derived not from `CommitmentMask` but from `Coinbase` so it much searches that index.

This bug will also slow down the wallet spending transactions by blocking the `CommitmentMask` branch via the mutex for that massive index search. This is now removed and will speed up the wallet also. 

How Has This Been Tested?
---
Unit tests, manual

---

### DIFF
--- a/base_layer/core/src/transactions/key_manager/inner.rs
+++ b/base_layer/core/src/transactions/key_manager/inner.rs
@@ -236,6 +236,28 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         Err(KeyManagerServiceError::KeyNotFoundInKeyChain)
     }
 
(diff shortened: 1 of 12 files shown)


sdbondi (dismissed):

sdbondi:
ACK

Tested and didn't get the mining slow down

hansieodendaal (approved):
Nice catch

utACK

PR #5488, opened 2023-06-22
fix(core): remove implicit change in protocol for partial/full signatures

sdbondi:
Description
---
Removes check for an invalid signature, presumed to be a partial signature, which results in partial signature aggregation being skipped if a valid signature is passed in.

Motivation and Context
---
This check validates the receiver signature to determine if the signature is sent is partial or full. If partial we aggregate the sender and receiver partial signatures. A partial signature and an invalid signature run the same code path (`  if received_output.verify_metadata_signature().is_err() {`) This code branch is executed as part of every normal Tari transaction.

It is not possible for the (non-local) receiver to unilaterally produce a correct signature and "bypass" aggregation as they do not possess the required secrets. The primary reason for this check is to allow the protocol to be shortcut when the sender and receiver are the same party and therefore can produce a valid signature (sign_as_sender_and_receiver). This PR implements this shortcut explicitly by adding a separate call for this case `add_presigned_recipient_info`.

How Has This Been Tested?
---
Existing transaction tests.

---
Send a normal and send-to-self transaction

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/transactions/transaction_protocol/sender.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/sender.rs
@@ -25,7 +25,7 @@ use std::fmt;
 use serde::{Deserialize, Serialize};
 use tari_common_types::{
     transaction::TxId,
-    types::{PrivateKey, PublicKey, Signature},
+    types::{ComAndPubSignature, PrivateKey, PublicKey, Signature},
 };
 use tari_crypto::{ristretto::pedersen::PedersenCommitment, tari_utilities::ByteArray};
 pub use tari_key_manager::key_manager_service::KeyId;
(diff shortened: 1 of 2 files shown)


base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -434,86 +448,100 @@ impl SenderTransactionProtocol {
...
+    async fn add_sender_partial_signature<KM: TransactionKeyManagerInterface>(
+        &self,
+        rec: &RecipientSignedMessage,
+        info: &RawTransactionInfo,
+        key_manager: &KM,
+    ) -> Result<ComAndPubSignature, TPE> {
+        let received_output = &rec.output;
+        // we need to make sure we use our values here and not the received values.
+        let metadata_message = TransactionOutput::metadata_signature_message_from_parts(
+            &received_output.version,
```
AaronFeickert, line 497: Is it possible for a downgrade or other version mismatch to arise if the received output's version is not what the sender expected?

AaronFeickert: This is now an [issue](https://github.com/tari-project/tari/issues/5529).

sdbondi: Thanks - completely agree there needs to be an explicit check for supported versions on both sides. FWIW currently, we have 1 enumerated version so I believe this can't be exploited.

PR #5489, opened 2023-06-23
feat: improve recovery speed

SWvheerden:
Description
---
Improves the recovery speed of utxo recovery. 

Motivation and Context
---
Utxo recovery will recognize that a utxo is its, do all the recovery and custom struct processes. Only at the end will the recovery "fail" by it not being inserted du to the commitment already existing the OMS db.  This moves this check forward before we attempt recovery and if we already know of the output, we continue with the next one. 

How Has This Been Tested?
---
Unit tests

---


sdbondi (approved):
utACK

PR #5492, opened 2023-06-23
chore: refactor out the shrink_hashmap

Cifko:
Description
---
Refactor shrink map to a separate file. And also shrink it to the size that it will not be shrink again on next call.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/mempool/shrink_hashmap.rs
+++ b/base_layer/core/src/mempool/shrink_hashmap.rs
@@ -1,0 +1,9 @@
+pub fn shrink_hashmap<K: Eq + Hash, V>(map: &mut HashMap<K, V>) -> (usize, usize) {
+    let cap = map.capacity();
+    let extra_cap = cap - map.len();
+    if extra_cap > 100 {
+        map.shrink_to(map.len() + 100);
+    }
+
+    (cap, map.capacity())
+}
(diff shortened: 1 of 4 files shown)


sdbondi (dismissed):
utACK

SWvheerden (dismissed):

sdbondi (approved):

base_layer/core/src/mempool/shrink_hashmap.rs
```diff
@@ -0,0 +1,33 @@
...
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+use std::{collections::HashMap, hash::Hash};
+
+pub fn shrink_hashmap<K: Eq + Hash, V>(map: &mut HashMap<K, V>) -> (usize, usize) {
+    let cap = map.capacity();
+    let extra_cap = cap - map.len();
+    if extra_cap > 100 {
+        map.shrink_to(map.len() + 100);
```
AaronFeickert, line 29: It might be useful to include a comment here on why this particular magic number was chosen for the capacity buffer, in addition to refactoring it to a constant (since both uses of it should match).

PR #5493, opened 2023-06-23
fix: overflow of target difficulty

Cifko:
Description
---
The target difficulty overflow over u64 could happen and it would produce only module by u64. Now it returns u64::MAX if it's over.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/proof_of_work/lwma_diff.rs
+++ b/base_layer/core/src/proof_of_work/lwma_diff.rs
@@ -15,6 +15,7 @@ use std::{
     cmp,
     cmp::{max, min},
     collections::VecDeque,
+    convert::TryFrom,
 };
 
 use log::*;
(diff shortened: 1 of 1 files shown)


base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -109,7 +110,7 @@ impl LinearWeightedMovingAverage {
         // k is the sum of weights (1+2+..+n) * target_time
         let k = n * (n + 1) * self.target_time / 2;
         #[allow(clippy::cast_possible_truncation)]
-        let target = (ave_difficulty * k / weighted_times) as u64;
+        let target = (ave_difficulty * k / weighted_times).to_u64().unwrap_or(u64::MAX);
```
sdbondi: You can remove the `#[allow(clippy::cast_possible_truncation)]`

sdbondi: We can use stdlibs conversion without using the num-traits dependency
```suggestion
        let target = u64::try_from(ave_difficulty * k / weighted_times).unwrap_or(u64::MAX);
```

sdbondi (approved):

SWvheerden (approved):

PR #5494, opened 2023-06-23
fix(core/keymanager): use tokio rwlock for keymanagers

sdbondi:
Description
---
Replaces futures-rs Mutex with tokio's RwLock for key manager locks

Motivation and Context
---
Allow multiple concurrent reads for keymanagers. In general, we prefer to use tokio's locks with the tokio runtime (no investigation done into performance differences).

If there was a particular reason to use futures-rs Mutex we can close this.

How Has This Been Tested?
---
Existing tests, this change is not benchmarked.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
utAck

PR #5495, opened 2023-06-23
feat: proof of work audit part 2

hansieodendaal:
Description
---
- Implemented creation, underflow and overflow protection for `Difficulty` in `difficulty.rs`.
- Removed all methods not adhering to these patterns/traits.
- Fixed usage in the code to adhere to these.
- Consolidated and refactored repetitive code in `difficulty.rs`.

Motivation and Context
---
Preparation for code audit

How Has This Been Tested?
---
Added unit tests
All unit tests pass

---
Code walkthrough
Review the new unit test and related code

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/proof_of_work/difficulty.rs
+++ b/base_layer/core/src/proof_of_work/difficulty.rs
@@ -20,14 +20,13 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{fmt, ops::Div};
+use std::fmt;
 
-use newtype_ops::newtype_ops;
 use num_format::{Locale, ToFormattedString};
 use serde::{Deserialize, Serialize};
 use tari_utilities::epoch_time::EpochTime;
 
-use crate::proof_of_work::error::DifficultyAdjustmentError;
+use crate::proof_of_work::{error::DifficultyError, DifficultyAdjustmentError};
 
 /// Minimum difficulty, enforced in diff retargeting
 /// avoids getting stuck when trying to increase difficulty subject to dampening
@@ -39,8 +38,11 @@ pub struct Difficulty(u64);
 
 impl Difficulty {
     /// A const constructor for Difficulty
-    pub const fn from_u64(d: u64) -> Self {
-        Self(d)
+    pub const fn from_u64(d: u64) -> Result<Self, DifficultyError> {
+        if d < MIN_DIFFICULTY {
+            return Err(DifficultyError::InvalidDifficulty);
+        }
+        Ok(Self(d))
     }
 
     /// Difficulty of MIN_DIFFICULTY
@@ -48,14 +50,14 @@ impl Difficulty {
         Difficulty(MIN_DIFFICULTY)
     }
 
+    /// Maximum Difficulty
+    pub const fn max() -> Difficulty {
+        Difficulty(u64::MAX)
+    }
+
     /// Return the difficulty as a u64
     pub fn as_u64(self) -> u64 {
         self.0
-    }
-
-    /// Subtract difficulty without overflowing
-    pub fn checked_sub(self, other: Difficulty) -> Option<Difficulty> {
-        self.0.checked_sub(other.0).map(Difficulty)
     }
 }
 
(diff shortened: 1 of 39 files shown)


base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -65,20 +70,68 @@ impl Default for Difficulty {
...
+    }
+}
+
+impl CheckedSub<u64> for Difficulty {
+    fn checked_sub(&self, other: u64) -> Option<Self> {
+        self.checked_sub(Difficulty(other))
+    }
+}
+
+impl Sub<Self> for Difficulty {
```
SWvheerden: I dont think we should implement this. Just leave it unimplemented. 
Doing it this way will trick the user in believing they can use +-/* etc and it will only panic on runtime. 
If the functions don't exits aka not implemented, it will give a compile error. 

We should not rely on runtime panics to force users to use the correct functions.

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -89,15 +121,54 @@ impl fmt::Display for Difficulty {
...
+#[cfg(any(test))]
+use std::ops::{Add, Div, Mul, Sub};
+
+// This trait should not be implemented for runtime, here only for testing
+#[cfg(any(test))]
+impl Add<Self> for Difficulty {
+    type Output = Self;
+
+    fn add(self, _rhs: Self) -> Self {
+        unimplemented!("Sub::sub<rhs> must not be used, use `.checked_add(value)` instead")
```
SWvheerden: I am not the biggest fan of this. I understand the reason here is to stop people from implementing the methods. 
My personal opinion here is that, I don't think the slight chance of that happening in getting missed in a PR is worth the extra clutter to keep these non-functioning methods in the file.

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -59,7 +59,7 @@ pub fn monero_difficulty(header: &BlockHeader, randomx_factory: &RandomXFactory)
 fn get_random_x_difficulty(input: &[u8], vm: &RandomXVMInstance) -> Result<(Difficulty, Vec<u8>), MergeMineError> {
     let hash = vm.calculate_hash(input)?;
     debug!(target: LOG_TARGET, "RandomX Hash: {:?}", hash);
-    let difficulty = little_endian_difficulty(&hash);
+    let difficulty = little_endian_difficulty(&hash).map_err(|e| MergeMineError::DeserializeError(e.to_string()))?;
```
SWvheerden: This is the wrong error. 
I would just add the difficulty error to MergeMineError
and then just use`?`

base_layer/core/src/validation/mocks.rs
```diff
@@ -115,10 +115,10 @@ impl<B: BlockchainBackend> HeaderChainLinkedValidator<B> for MockValidator {
         _: Option<Difficulty>,
     ) -> Result<AchievedTargetDifficulty, ValidationError> {
         if self.is_valid.load(Ordering::SeqCst) {
-            let achieved = sha3x_difficulty(header);
+            let achieved = sha3x_difficulty(header)?;
 
             let achieved_target =
-                AchievedTargetDifficulty::try_construct(PowAlgorithm::Sha3, achieved - Difficulty::from(1), achieved)
+                AchievedTargetDifficulty::try_construct(PowAlgorithm::Sha3, achieved, achieved.checked_add(1).unwrap())
```
SWvheerden, line 121: These 2 lines are not equal:
We had` (sha3, achieved -1, achieved)`
Now we have: `(sha3, achieved, achieved + 1)`

SWvheerden (commented):
This looks good, I have one question on a test change

base_layer/core/src/validation/mocks.rs
hansieodendaal: The original line in the test helper now panicked because it resulted in a difficulty of zero as these tests use `localnet` with min and max difficulties set to 1.

The single test that still failed due to this change, `base_layer/core/src/chain_storage/blockchain_database.rs/async fn it_correctly_handles_duplicate_blocks()`, because it bargained on a runtime difficulty of zero, was changed to honour minimum difficulty.

SWvheerden (approved):

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -39,59 +41,103 @@ pub struct Difficulty(u64);
...
+    pub fn big_endian_difficulty(hash: &[u8]) -> Result<Difficulty, DifficultyError> {
+        let scalar = U256::from_big_endian(hash); // Big endian so the hash has leading zeroes
+        Difficulty::u256_scalar_to_difficulty(scalar)
+    }
+
+    /// Helper function to provide the difficulty of the hash assuming the hash is little_endian
+    pub fn little_endian_difficulty(hash: &[u8]) -> Result<Difficulty, DifficultyError> {
+        let scalar = U256::from_little_endian(hash); // Little endian so the hash has trailing zeroes
+        Difficulty::u256_scalar_to_difficulty(scalar)
     }
```
AaronFeickert, line 76: These will panic if the provided slice length does not match the expected `U256` size. This can't happen with the current callers, which are either test functions or use the output from 256-bit `SHA3` hashing; however, it may be prudent to document this behavior, or to check the input slice and return an error if needed.

hansieodendaal: Thanks! I will add some checks.

AaronFeickert: Great! Since the functions already return a `Result`, it seems straightforward to add a check, and presumably this would have no performance hit in practice.

PR #5497, opened 2023-06-23
fix(core): compile error in wallet/FFI

sdbondi:
Description
---
Fix wallet/FFI compile error due to feature gated field

Motivation and Context
---
FII/wallet fail to compile due to feature gated field being used in non gated function in ConsensusBuilder

Introduced in https://github.com/tari-project/tari/pull/5483

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/transactions/mod.rs
+++ b/base_layer/core/src/transactions/mod.rs
@@ -27,6 +27,7 @@ pub mod weight;
 pub mod key_manager;
 
 #[macro_use]
+#[cfg(feature = "base_node")]
 pub mod test_helpers;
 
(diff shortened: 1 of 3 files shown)


base_layer/core/src/transactions/mod.rs
```diff
@@ -27,6 +27,7 @@ pub mod weight;
 pub mod key_manager;
 
 #[macro_use]
+#[cfg(feature = "base_node")]
```
sdbondi, line 30: I think this is fine, waiting for CI to confirm

SWvheerden (approved):
Tested locally, worked 100%

PR #5499, opened 2023-06-23
chore: rustsec fixes or ignores

brianp:
Description
---
This removes all rustsec flags and warnings

- Upgrades for openssl
- log4rs fork with dependency removal ensuring our repo doesn't include a  problematic version of the `time` crate
- Borsh investigation done, no types should be effected
- The rest are unmainted repo warnings and for different reasons we can't/won't upgrade right now

Motivation and Context
---
Remove the rustsec audit warnings and clean things up

How Has This Been Tested?
---
N/A

---
N/A

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/tari_base_node/Cargo.toml
```diff
@@ -38,7 +38,7 @@ futures = { version = "^0.3.16", default-features = false, features = ["alloc"]
 qrcode = { version = "0.12" }
 log = { version = "0.4.8", features = ["std"] }
 log-mdc = "0.1.0"
-log4rs = { version = "1.2.0", default_features = false, features = ["config_parsing", "threshold_filter", "yaml_format", "console_appender", "rolling_file_appender", "compound_policy", "size_trigger", "fixed_window_roller"] }
+log4rs = { git = "https://github.com/brianp/log4rs.git", default_features = false, features = ["config_parsing", "threshold_filter", "yaml_format", "console_appender", "rolling_file_appender", "compound_policy", "size_trigger", "fixed_window_roller"] }
```
stringhandler: I've forked log4rs into tari-project, it may be better to use that one

stringhandler (dismissed):

applications/tari_base_node/Cargo.toml
brianp: Updated

SWvheerden (approved):

applications/tari_base_node/Cargo.toml
```diff
@@ -38,7 +38,7 @@ futures = { version = "^0.3.16", default-features = false, features = ["alloc"]
 qrcode = { version = "0.12" }
 log = { version = "0.4.8", features = ["std"] }
 log-mdc = "0.1.0"
-log4rs = { version = "1.2.0", default_features = false, features = ["config_parsing", "threshold_filter", "yaml_format", "console_appender", "rolling_file_appender", "compound_policy", "size_trigger", "fixed_window_roller"] }
+log4rs = { git = "https://github.com/tari-project/log4rs.git", default_features = false, features = ["config_parsing", "threshold_filter", "yaml_format", "console_appender", "rolling_file_appender", "compound_policy", "size_trigger", "fixed_window_roller"] }
```
AaronFeickert, line 41: This doesn't appear to freeze at any particular version. Does that introduce risk on any breaking changes to the forked repository?

AaronFeickert: See other instances of this elsewhere in the PR.

brianp: I'm not so worried, as the original repo is essentially dead, and we don't plan on making additional fixes so likely this will stay as is, until we go back to the original if/when they ever update.

AaronFeickert: I mean is there a particular reason to use the main branch for this, instead of including a specific commit reference or tag?

PR #5501, opened 2023-06-23
fix(core/consensus): include `coinbase_extra` max size into coinbase weight calculation

mrnaveira:
Description
---
Include the `coinbase_extra` maximum size into block weight calculation

Motivation and Context
---
In the consensus constants function `get_max_block_weight_excluding_coinbase` we are not taking into account the size that the `coinbase_extra` field may have. This PR includes the maximum size of that field (provided by `coinbase_output_features_extra_max_length`) in to the calculation.

How Has This Been Tested?
---
Unit tests

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

Makes sense to me. Before this PR, a miner could mine an invalid block after putting in the coinbase extra.

SWvheerden (dismissed):
Nice definitely a safer fix here. It goes with the idea here in that the miner should not generate a block that exceeds the max_block_weight. 

We still need to account for the script size.

mrnaveira:
The weight calculation was not correct in my initial PR, it's now fixed

stringhandler (approved):

PR #5502, opened 2023-06-23
feat: proof of work audit part 3

hansieodendaal:
Description
---
- Removed `max_block_time` for `struct LinearWeightedMovingAverage`:
  - Removed manual assignment of `max_block_time` for `struct LinearWeightedMovingAverage` as it is always a fixed multiple of `target_time` according to the **LWMA-1** algorithm specification. 
  - Previously, `max_block_time` was asserted when compiling tests and debug mode with function `fn assert_hybrid_pow_constants(`. 
  - This change resulted in the corresponding removal of  `max_target_time` from `struct PowAlgorithmConstants` (and consensus constants) as it is now being calculated by `LinearWeightedMovingAverage`.
- Added documentation to public methods.

Dependent on #5495 

Motivation and Context
---
Preparation for code audit

How Has This Been Tested?
---
All unit tests pass

---
Code walkthrough

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -64,6 +64,7 @@ message CommitmentSignature {
 
 /// PoW Algorithm constants
 message PowAlgorithmConstants {
+    uint64 max_target_time = 1;
     uint64 min_difficulty = 2;
     uint64 max_difficulty = 3;
     uint64 target_time = 4;
@@ -109,7 +110,6 @@ message ConsensusConstants {
     uint32 blockchain_version = 2;
     uint64 future_time_limit = 3;
     uint64 difficulty_block_window = 5;
-    uint64 difficulty_max_block_interval = 6;
     uint64 max_block_transaction_weight = 7;
     uint64 pow_algo_count = 8;
     uint64 median_timestamp_count = 9;
--- a/applications/tari_app_grpc/proto/types.proto
+++ b/applications/tari_app_grpc/proto/types.proto
@@ -64,7 +64,6 @@ message CommitmentSignature {
 
 /// PoW Algorithm constants
 message PowAlgorithmConstants {
-    uint64 max_target_time = 1;
     uint64 min_difficulty = 2;
     uint64 max_difficulty = 3;
     uint64 target_time = 4;
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -34,7 +34,7 @@ use tari_utilities::epoch_time::EpochTime;
 use crate::{
     borsh::SerializedSize,
     consensus::network::NetworkConsensus,
-    proof_of_work::{lwma_diff::LWMA_MAX_BLOCK_TIME_RATIO, Difficulty, PowAlgorithm},
+    proof_of_work::{lwma_diff::LinearWeightedMovingAverage, Difficulty, PowAlgorithm},
     transactions::{
         tari_amount::{uT, MicroTari, T},
         transaction_components::{
@@ -142,7 +142,6 @@ fn version_zero() -> (
 /// This is a convenience struct to put all the info into a hashmap for each algorithm
 #[derive(Clone, Debug)]
 pub struct PowAlgorithmConstants {
-    pub max_target_time: u64,
     pub min_difficulty: Difficulty,
     pub max_difficulty: Difficulty,
     pub target_time: u64,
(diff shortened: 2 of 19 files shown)


stringhandler:
Needs more description about the difficulty changes

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -243,11 +242,9 @@ impl ConsensusConstants {
 
     /// The maximum time a block is considered to take. Used by the difficulty adjustment algorithms
     /// Multiplied by the PoW algorithm block percentage.
-    pub fn pow_max_block_interval(&self, pow_algo: PowAlgorithm) -> u64 {
-        match self.proof_of_work.get(&pow_algo) {
-            Some(v) => v.max_target_time,
-            _ => 0,
-        }
+    pub fn pow_max_block_interval(&self, pow_algo: PowAlgorithm) -> Result<u64, String> {
+        LinearWeightedMovingAverage::max_block_time(self.pow_target_block_interval(pow_algo))
```
stringhandler: This seems like a big change, possibly leaking out the implementation that's abstracted by proof of work. Can you give some more reasons as to this change?

base_layer/core/src/proof_of_work/error.rs
```diff
@@ -48,7 +48,9 @@ pub enum DifficultyAdjustmentError {
 }
 
 #[derive(Debug, Error)]
-pub enum DifficultyError {
+pub enum LwmaError {
```
stringhandler: This does not seem like a good renaming. The variant "InvalidDifficulty" does not seem related to Lwma at all

base_layer/core/src/validation/error.rs
```diff
@@ -152,7 +152,7 @@ pub enum ValidationError {
     #[error("Not enough timestamps provided. Expected {expected}, got {actual}")]
     NotEnoughTimestamps { expected: usize, actual: usize },
     #[error("Invalid difficulty: {0}")]
-    DifficultyError(#[from] DifficultyError),
+    LwmaError(#[from] LwmaError),
```
stringhandler: I also don't really like this rename

stringhandler: Comment is wrong if nothing else

stringhandler (changes requested):

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: I propose updating the comment as:
```
    /// The maximum time/interval a block is considered to take. This is used by the difficulty adjustment algorithms, 
    /// and multiplied by the PoW algorithm block percentage. For LWMA-1, `max_block_time` is a fixed multiple of 
    /// `target_block_time`, determined by `LinearWeightedMovingAverage`.
```

stringhandler: it's more about duplicating code in a strange place that is unlikely to be found. I preferred the approach of asking `self.proof_of_work` for this answer

hansieodendaal: After discussion with @stringhandler, accepting the suggestion to remove this function as it is easily calculated with `LinearWeightedMovingAverage::max_block_time(target_time)` and passing in `ConsensusConstants::pow_target_block_interval(..)`, as `max_block_time` is not a consensus parameter.

applications/tari_app_grpc/proto/types.proto
```diff
@@ -64,6 +64,7 @@ message CommitmentSignature {
 
 /// PoW Algorithm constants
 message PowAlgorithmConstants {
+    uint64 max_target_time = 1;
```
stringhandler: This does not exist in the struct anymore, so can be removed

base_layer/core/src/mempool/shrink_hashmap.rs
```diff
@@ -20,8 +20,10 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#[cfg(feature = "base_node")]
```
SWvheerden: put this in the mod file rather

SWvheerden: the cfg flag, for entire file

SWvheerden (approved):

PR #5503, opened 2023-06-23
fix(core/base_node): safe `mmr_position` cast in horizon sync

mrnaveira:
Description
---
Perform safe numeric conversion of the `mmr_position` inside the horizon state sync

Motivation and Context
---
There was an unsafe cast from `u64` to `u32` of the `mmr_position` inside the horizon state sync. This PR does a safe cast, raising a typed error instead of truncating the value.

The existing TODO comment suggested to do it inside the database function, but in my opinion at this point this would overcomplicate the solution.

How Has This Been Tested?
---
Unit tests

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/sync/horizon_state_sync/error.rs
+++ b/base_layer/core/src/base_node/sync/horizon_state_sync/error.rs
@@ -65,7 +65,7 @@ pub enum HorizonSyncError {
         expected_hex: String,
         actual_hex: String,
     },
-    #[error("Invalid MRR position {mmr_position} at height {at_height}")]
+    #[error("Invalid MMR position {mmr_position} at height {at_height}")]
     InvalidMmrPosition { at_height: u64, mmr_position: u64 },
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/sync/horizon_state_sync/error.rs
```diff
@@ -65,6 +65,8 @@ pub enum HorizonSyncError {
         expected_hex: String,
         actual_hex: String,
     },
+    #[error("Invalid MRR position {mmr_position} at height {at_height}")]
```
stringhandler:
```suggestion
    #[error("Invalid MMR position {mmr_position} at height {at_height}")]
```

stringhandler (dismissed):
utACK, just a spelling error

SWvheerden (approved):

PR #5505, opened 2023-06-26
fix(core): disable covenants for all networks except igor and localnet

sdbondi:
Description
---
- Adds check for covenant token length in outputs
- Disables (max len == 0) covenants on all networks except igor and localnet
- Rearranged some of the static and cheap checks before more expansive checks e.g script and covenant execution

Motivation and Context
---
Allows limiting covenant length in consensus constants. We disable them for now in all non-test networks.

We check covenant length in outputs but not inputs, as this may render previous UTXOs unspendable if the max length was set lower at later heights (unlikely). This feels like a debatable choice, as always, open to comments.

How Has This Been Tested?
---
Manually, synced base node

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/consensus/consensus_constants.rs
+++ b/base_layer/core/src/consensus/consensus_constants.rs
@@ -103,6 +103,8 @@ pub struct ConsensusConstants {
     permitted_range_proof_types: &'static [RangeProofType],
     /// Coinbase outputs are allowed to have metadata, but it has the following length limit
     coinbase_output_features_extra_max_length: u32,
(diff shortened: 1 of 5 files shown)


stringhandler (approved):

stringhandler:
utACK

SWvheerden (approved):

hansieodendaal:
Thanks @sdbondi , this change was needed.

PR #5508, opened 2023-06-26
fix: nit lwma_diff division check

jorgeantonio21:
Description
---
Add check for a possible edge case in the `lwma_diff` division check.

Motivation and Context
---
Auditing code.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/proof_of_work/lwma_diff.rs
+++ b/base_layer/core/src/proof_of_work/lwma_diff.rs
@@ -87,7 +87,15 @@ impl LinearWeightedMovingAverage {
             .skip(1)
             .fold(0u128, |difficulty, (_, d)| difficulty + u128::from(d.as_u64()));
 
-        let ave_difficulty = difficulty / n;
+        let ave_difficulty = if let Some(ave) = difficulty.checked_div(n) {
+            ave
+        } else {
+            trace!(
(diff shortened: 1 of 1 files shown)


stringhandler (approved):
Not technically possible at the moment, but could be if the code was edited later

base_layer/core/src/proof_of_work/lwma_diff.rs
```diff
@@ -87,7 +87,15 @@ impl LinearWeightedMovingAverage {
             .skip(1)
             .fold(0u128, |difficulty, (_, d)| difficulty + u128::from(d.as_u64()));
 
-        let ave_difficulty = difficulty / n;
+        let ave_difficulty = if let Some(ave) = difficulty.checked_div(n) {
```
sdbondi, line 90: `n` has already been checked above and can never be 0. `n = self.target_difficulties.len() - 1` where  `self.target_difficulties.len()` must be at least 2.

jorgeantonio21: Right, I didn't realize that. So these changes are not needed actually :)

PR #5509, opened 2023-06-26
chore: speed up ci

CjS77:
- Revert to using actions/cache, since Swatinem.rust-cache disables partial cahce matches
- Remove redundant `build-nightly` step

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
Ack

CjS77:
Note: GHA hosted runs are a lot quicker now, but self-run aren't because the cache transfer is slow (10MB/s). 
Best route here is to improve caching on self-hosted jobs, and then drop the remote caching steps

Issue #5511, opened 2023-06-26
FTL accessor method ftl() has clippy allows


SWvheerden:
Function `ftl()` has clippy allows

PR #5522, opened 2023-06-27
chore(core/consensus): remove unnecessary clippy allows on FTL

mrnaveira:
Description
---
Removed clippy allows on `flt` accesor method in consensus constants. Turns out that the allows were unnecessary, those rules are allowed by default in clippy and we haven't disabled them in the `lints.toml` file.

Motivation and Context
---
Solves https://github.com/tari-project/tari/issues/5511

How Has This Been Tested?
---
Clippy passes

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


mrnaveira:
Even if clippy allows it by default, I think it's safer to handle it explicitly

SWvheerden (approved):

PR #5532, opened 2023-06-27
chore: update local coverage script

CjS77:
The source coverage scripts are incredibly outdated. This updates to work (though I'm not convinced it's 100% accurate - function % results seems very wrong). But that's for another PR

## Deleted scripts

A few of the utility scripts have been deleted since they don't appear to be used anywhere or have any real use cases.
If this is not the case, comment below!

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


CjS77:
ACK

SWvheerden (approved):

PR #5533, opened 2023-06-27
fix: push test coverage even if some tests fail

CjS77:
Description
---

Our test coverage shows 0% because the GA fails if even one test fails (and the integration tests typically have at least one failure).

So allow the coverage report to complete even without a clean bill of health.

Remote caching was also disabled for this GA, since the compiler flags are different for code coverage runs.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


CjS77:
ACK

Issue #5465, opened 2023-06-19
bug(console-wallet): subtract with underflow panic when contact list is empty


sdbondi:
**Describe the bug**
Send tab, attempt to send via the contact list. I have a single NeverSeen contact that is not listed in the dialog. I press the down arrow and the console wallet crashes

```
thread 'main' panicked at 'attempt to subtract with overflow', applications/tari_console_wallet/src/ui/widgets/list_state.rs:62:36
```

**To Reproduce**
1. Go to send tab
2. press 'c' to get contacts dialog
3. press up or down arrow
4. panic

**Expected behavior**
The console wallet should not panic when up/down arrow pressed

**Screenshots**
<img width="1166" alt="image" src="https://github.com/tari-project/tari/assets/1057902/9e9da88a-4959-424a-8134-342ef6d3f938">

**Desktop (please complete the following information):**
 - OS & Version: iOS 13.3.1

**Additional context**

AaronFeickert:
I wasn't able to reproduce this bug. I did the following:
- Create a new wallet
- Move to the contacts tab
- Create a new contact
- Move to the send tab
- Activate the contacts dialog
- Press either the up or down arrow key

Regardless of which arrow key I press, the contact is selected and the wallet does not crash.

![Screenshot from 2023-06-27 15-36-10](https://github.com/tari-project/tari/assets/66188213/f30fef3d-5429-44c8-b000-d2600d8e603c)

If I just create the new wallet and do _not_ create a new contact, such that the contact dialog is empty, the wallet also does not crash.

![Screenshot from 2023-06-27 15-38-19](https://github.com/tari-project/tari/assets/66188213/3b17cde3-2dc8-42dd-ab34-98e5dc1bb410)

Do you still get the issue using a new wallet?

AaronFeickert:
Also, why does the contact not appear in the dialog?

sdbondi:
I'll check again, in the code too since it's likely a trivial fix.

AaronFeickert:
Looks like this issue can be closed @sdbondi.

PR #5535, opened 2023-06-28
fix(console-wallet): fix possible subtract underflow panic in list

sdbondi:
Description
---
Fix possible subtract with underflow panic in console wallet list state

Motivation and Context
---
Ref #5465 

Unable to reproduce this exact panic manually however added a test case which causes the same panic. I was previously able to reproduce this every time, either this has been fixed in other changes or the reproduction steps did not include some mix of circumstances that were pertinent to the bug. In any case, this PR fixes a bug in select_first function which uses the binary compliment where a boolean inequality check was likely intended !self.num_items == 0 vs self.num_items != 0, though strangely the correct logic is 'if self.num_items == 0 {'. On an empty list this causes a non-existent first item at index 0 to be selected, a call to update_list_state would panic.

How Has This Been Tested?
---
Unit test that reproduces the same underflow panic in #5465, although the send to contact list does not use select_first 
Manually

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_console_wallet/src/ui/components/burn_tab.rs
+++ b/applications/tari_console_wallet/src/ui/components/burn_tab.rs
@@ -234,7 +234,7 @@ impl BurnTab {
 
         let mut list_state = self
             .proofs_list_state
-            .get_list_state((list_areas[1].height as usize).saturating_sub(3));
+            .update_list_state((list_areas[1].height as usize).saturating_sub(3));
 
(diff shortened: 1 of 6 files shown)


SWvheerden (approved):

CjS77:
ACK

PR #5537, opened 2023-06-28
feat: borsh sized serialization should be fallible  

jorgeantonio21:
Description
---
Adds fallibility to borsh serialized size method. 

Motivation and Context
---
Audit

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -359,7 +359,10 @@ impl AggregateBody {
     }
 
     pub fn sum_features_and_scripts_size(&self) -> usize {
-        self.outputs.iter().map(|o| o.get_features_and_scripts_size()).sum()
+        self.outputs
+            .iter()
+            .map(|o| o.get_features_and_scripts_size().expect("Failed to serialize size"))
```
SWvheerden: prefer making all of these errors.

base_layer/core/src/transactions/test_helpers.rs
```diff
@@ -189,7 +189,12 @@ impl TestParams {
     pub fn get_size_for_default_features_and_scripts(&self, num_outputs: usize) -> usize {
         let output_features = OutputFeatures { ..Default::default() };
         self.fee().weighting().round_up_features_and_scripts_size(
-            script![Nop].get_serialized_size() + output_features.get_serialized_size(),
+            script![Nop]
+                .get_serialized_size()
+                .expect("Failed to get serialized size") +
+                output_features
+                    .get_serialized_size()
+                    .expect("Failed to get serialized size"),
```
SWvheerden: same here

base_layer/core/src/transactions/test_helpers.rs
```diff
@@ -528,9 +533,15 @@ pub async fn create_wallet_outputs(
...
-            output_covenant.get_serialized_size(),
+        output_features
+            .get_serialized_size()
+            .expect("Failed to get serialized size") +
+            output_script
+                .get_serialized_size()
+                .expect("Failed to get serialized size") +
+            output_covenant
+                .get_serialized_size()
+                .expect("Failed to get serialized size"),
```
SWvheerden: same here

SWvheerden (commented):
This is going the correct way. 
But we should remove the expects as they are the same as the unwrap we replace. 
I would prefer to replace them with Result<>

base_layer/core/src/transactions/aggregated_body.rs
jorgeantonio21: Refactored :) !

SWvheerden (dismissed):
It is so nice to see all these unwraps disappear

Issue #5515, opened 2023-06-26
y: cast_possible_truncation and clippy:cast_possible_wrap


SWvheerden:
We should uncomment clippy: cast_possible_truncation and clippy:cast_possible_wrap in lints.toml and see what breaks

AaronFeickert:
Duplicate of #5478.

PR #5538, opened 2023-06-28
chore: handle casts indicated by clippy lint rules

mrnaveira:
Description
---
Manually handle casts indicated by clippy lint rules

Motivation and Context
---
Currently we allow` clippy: cast_possible_truncation` and `clippy:cast_possible_wrap` in `lints.toml`. This is the default behaviour in clippy, but ideally we would want to deny them and solve manually each case.

Denying those lint rules in `lints.toml` ends up in dozens of warnings on code that, upon further inspection, do not seem to be particularly problematic.

This PR handles only a few of the most relevant casts, leaving the rules as allowed.

Addresses https://github.com/tari-project/tari/issues/5515

How Has This Been Tested?
---
Clippy and unit tests

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
Ack

base_layer/common_types/src/emoji.rs
```diff
@@ -97,7 +98,7 @@ lazy_static! {
     pub static ref REVERSE_EMOJI: HashMap<char, u8> = {
         let mut m = HashMap::with_capacity(DICT_SIZE);
         EMOJI.iter().enumerate().for_each(|(i, c)| {
-            m.insert(*c, i as u8);
+            m.insert(*c, u8::try_from(i).expect("Invalid emoji"));
```
AaronFeickert, line 101: Since the current table size ensures this case is safe, it seems like you could remove the panic and instead use a `debug_assert` where the table constants are defined.

PR #5539, opened 2023-06-28
[wip - fixing unit tests] feat: add genesis block checks with not-before-proof

hansieodendaal:
Description
---
- Added genesis block checks to `ConsensusManager::get_genesis_block`, as these were never checked in the production code. This check will be executed whenever a new genesis block is requested, either when a base node starts fresh or with blockchain recovery.
- Added a 'not-before-proof' to 'ProofOfWork::pow_data' that is also checked.
- Added a specific `localnet` genesis block that uses the `esmeralda` genesis block but without the faucet UTXOs.
- The `genesis_block_sanity_check` unit test now consolidates all previous genesis block sanity checks and also uses the production code to check the blocks.

```
genesis_block_sanity_check for 'stagenet': 24.5049ms
genesis_block_sanity_check for 'nextnet': 13.2398ms
genesis_block_sanity_check for 'localnet': 9.3392ms
genesis_block_sanity_check for 'igor': 22.1562012s
genesis_block_sanity_check for 'esmeralda': 10.3047587s
```

Motivation and Context
---
Preparation for code audit.
The genesis block was never checked in the production code. 

How Has This Been Tested?
---
Added to and modified `genesis_block_sanity_check` unit test

---
Review code changes

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -34,9 +34,6 @@ use crate::{
     transactions::{aggregated_body::AggregateBody, transaction_components::TransactionOutput},
 };
 
-// This can be adjusted as required, but must be limited
-const NOT_BEFORE_PROOF_BYTES_SIZE: usize = u16::MAX as usize;
-
 /// Returns the genesis block for the selected network.
 pub fn get_genesis_block(network: Network) -> ChainBlock {
     use Network::{Esmeralda, Igor, LocalNet, MainNet, NextNet, StageNet};
@@ -46,7 +43,7 @@ pub fn get_genesis_block(network: Network) -> ChainBlock {
         NextNet => get_nextnet_genesis_block(),
         Igor => get_igor_genesis_block(),
         Esmeralda => get_esmeralda_genesis_block(),
-        LocalNet => get_esmeralda_genesis_block(),
+        LocalNet => get_localnet_genesis_block(),
     }
 }
 
@@ -99,6 +96,25 @@ fn print_mr_values(block: &mut Block, print: bool) {
     println!("kernel mr: {}", block.header.kernel_mr.to_hex());
     println!("output mr: {}", block.header.output_mr.to_hex());
     println!("vn mr: {}", block.header.validator_node_mr.to_hex());
+}
+
+pub fn get_localnet_genesis_block() -> ChainBlock {
+    // Lets get the block
+    let block = get_esmeralda_genesis_block_raw();
+
+    // For testing we do not want any faucet UTXOs, so we omit that step
+
+    // Finalize the block
+    let accumulated_data = BlockHeaderAccumulatedData {
+        hash: block.hash(),
+        total_kernel_offset: block.header.total_kernel_offset.clone(),
+        achieved_difficulty: Difficulty::min(),
+        total_accumulated_difficulty: 1,
+        accumulated_randomx_difficulty: Difficulty::min(),
+        accumulated_sha3x_difficulty: Difficulty::min(),
+        target_difficulty: Difficulty::min(),
+    };
+    ChainBlock::try_construct(Arc::new(block), accumulated_data).unwrap()
 }
 
 pub fn get_stagenet_genesis_block() -> ChainBlock {
@@ -316,10 +332,6 @@ fn get_esmeralda_genesis_block_raw() -> Block {
 fn get_raw_block(genesis_timestamp: &DateTime<FixedOffset>, not_before_proof: &[u8]) -> Block {
     // Note: Use 'print_new_genesis_block_values' in core/tests/helpers/block_builders.rs to generate the required
     // fields below
-
-    let mut not_before_proof = not_before_proof.to_vec();
-    not_before_proof.truncate(NOT_BEFORE_PROOF_BYTES_SIZE);
-
     #[allow(clippy::cast_sign_loss)]
     let timestamp = genesis_timestamp.timestamp() as u64;
     Block {
@@ -346,7 +358,7 @@ fn get_raw_block(genesis_timestamp: &DateTime<FixedOffset>, not_before_proof: &[
             nonce: 0,
             pow: ProofOfWork {
                 pow_algo: PowAlgorithm::Sha3x,
-                pow_data: not_before_proof,
+                pow_data: not_before_proof.to_vec(), // Limited to consensus_manager::NOT_BEFORE_PROOF_BYTES_SIZE
             },
         },
         body: AggregateBody::new(vec![], vec![], vec![]),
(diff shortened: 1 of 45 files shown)


base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -299,127 +367,26 @@ fn get_raw_block(genesis_timestamp: &DateTime<FixedOffset>) -> Block {
...
-            .any(|k| k.features.contains(KernelFeatures::COINBASE_KERNEL));
-        assert!(!some_kernel_contains_coinbase_features);
-
-        // Check MMR
-        let mut kernel_mmr = KernelMmr::new(Vec::new());
-        for k in block.block().body.kernels() {
-            kernel_mmr.push(k.hash().to_vec()).unwrap();
+    fn genesis_block_sanity_check() {
+        for network in &[
+            // TODO: mainnet not yet implemented, uncomment when ready
```
SWvheerden, line 378:
```suggestion
```
I have gone through effort to remove all todo's and create issues for them. 
The uncommented Mainnet is fine

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -132,7 +152,18 @@ pub fn get_stagenet_genesis_block() -> ChainBlock {
 fn get_stagenet_genesis_block_raw() -> Block {
     // Set genesis timestamp
     let genesis_timestamp = DateTime::parse_from_rfc2822("15 Jun 2023 14:00:00 +0200").expect("parse may not fail");
-    get_raw_block(&genesis_timestamp)
+    let not_before_proof = b"i am the stagenet genesis block, watch out, here i come \
```
SWvheerden: This is great.

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -32,16 +40,29 @@ use crate::{
...
     proof_of_work::DifficultyAdjustmentError,
-    transactions::{tari_amount::MicroTari, transaction_components::TransactionKernel},
+    transactions::{
+        tari_amount::MicroTari,
+        transaction_components::{TransactionError, TransactionKernel},
+    },
 };
 
+#[cfg(feature = "base_node")]
+pub const NOT_BEFORE_PROOF_BYTES_SIZE: usize = u16::MAX as usize;
```
SWvheerden: Its perhaps not too bad, but this is not really required. There is no free reign on this data vec. 
For randomx blocks, this data needs to deserialize to a monero block
and if its a sha3x block, its required to be empty. 

And for the genesis block, because they can only be created in code we can see how large they are, and its not open to manipulation by hostile agents

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -69,17 +98,159 @@ impl ConsensusManager {
...
-        match network {
+        let genesis_block = match network {
             Network::LocalNet => self
                 .inner
                 .gen_block
                 .clone()
                 .unwrap_or_else(|| get_genesis_block(network)),
             _ => get_genesis_block(network),
+        };
+        self.check_genesis_block(&genesis_block)?;
```
SWvheerden, line 113: Not going to block this, but not a fan of this. 
This according to me gives us no benefits, all it does is make the boot-up time a node slower by 10-20 seconds.

base_layer/core/src/consensus/mod.rs
```diff
@@ -33,5 +33,7 @@ mod consensus_encoding;
 pub use consensus_encoding::{DomainSeparatedConsensusHasher, MaxSizeBytes, MaxSizeString};
 mod network;
 pub use network::NetworkConsensus;
+pub use test_helpers::TestConsensusConstantsBuilder;
 
 pub mod emission;
+pub mod test_helpers;
```
SWvheerden, line 39: can we make this #[cfg(test)]

base_layer/core/tests/helpers/mod.rs
```diff
@@ -6,14 +6,25 @@
 //! There are macros, such as `txn_schema!` that help you to easily construct valid transactions in test blockchains,
 //! through to functions that bootstrap entire blockchains in `sample_blockchains`.
 
+#[cfg(any(test))]
```
SWvheerden, line 9: Why not put this in tests/core_integretion_tests.rs
It should require one?

base_layer/core/src/consensus/consensus_manager.rs
hansieodendaal: Hmmm... I had an interesting chat with ChatGPT about this ;)

How many bytes would a digitized grayscale headline of 1st January 2000 of The New York Times consist of if we need to display it on a 1280 x 768 pixel screen after reconstruction? Let us assume the  headline is 

_"as I sip my drink, thoughts of esmeralda consume my mind, like a refreshing nourishing draught_

_The New York Times , 2000/01/01_

_Lorem Ipsum_

_Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."_

Without thinking too much about it, the response seemed plausible, so as a thought experiment I chose `u16::MAX` as the prediction was `40,320` .

**Response**

To estimate the number of bytes a digitized grayscale headline of January 1, 2000, from The New York Times would occupy, considering the provided headline text and a screen resolution of 1280 x 768 pixels, we need to account for the following factors:

- Text Size and Font: The size and font used to display the headline will impact the number of pixels required for each character.

- Length of the Headline: The number of characters in the headline will contribute to its size since each character will require corresponding pixels in the grayscale representation.

- Resolution: The headline will be displayed on a 1280 x 768 pixel screen.

- Grayscale Encoding: Grayscale images typically use one byte (8 bits) per pixel to represent the intensity level of gray.

Considering the specific headline text and assuming an average font size, we can estimate the dimensions occupied by the headline on the screen.

Taking into account that the headline spans three lines, we can approximate the width to be around 1120 pixels (assuming an average of 8 pixels per character) and the height to be around 36 pixels (assuming 12 pixels per line). Therefore, the headline would require approximately 40,320 pixels (1120 pixels width * 36 pixels height).

Since grayscale images use one byte per pixel, we can estimate that the headline would consist of approximately 40,320 bytes (40,320 pixels * 1 byte/pixel).

Please note that this is a rough estimation, and the actual byte count may vary depending on the specific font, font size, text content, and compression techniques applied. For a more accurate estimation, it would be necessary to have specific details about the font, headline text, and rendering process, as well as consider any compression or encoding methods used.

base_layer/core/src/consensus/consensus_manager.rs
hansieodendaal: I hear you, and I don't. I am currently fixing many invalid tests, and release mode should be a lot faster. Personally, I like the formal check on the genesis block.

base_layer/core/src/consensus/consensus_manager.rs
CjS77: You don't want a digitised version. That's too hard to verify (download, convert save as a file, with right extension, view). 

KISS. Just place a suitable headline text in as a string.

hansieodendaal: Well, in that case, we could limit the vector size all the way down to something like 2000, but I guess the specific number can be agreed upon later.

PR #5544, opened 2023-06-28
chore: add relevant crates to source coverage

CjS77:
The source coverage GHA is working fine, so this adds all the core crates to the coverage tests.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/scripts/source_coverage.sh
+++ b/scripts/source_coverage.sh
@@ -16,11 +16,6 @@ tari_base_node
 tari_base_node_grpc_client
 tari_chat_client
 tari_chat_ffi
-tari_common_sqlite
-tari_common_types
-tari_comms
-tari_comms_dht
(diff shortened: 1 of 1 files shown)


scripts/source_coverage.sh
```diff
@@ -30,18 +25,25 @@ tari_merge_mining_proxy
 tari_metrics
 tari_miner
 tari_mining_helper_ffi
-tari_p2p
-tari_service_framework
 tari_test_utils
 tari_wallet_ffi
```
stringhandler, line 29: This actually has some good tests in it, and would be good to track the coverage

stringhandler: i.e. wallet_ffi

PR #5548, opened 2023-06-28
chore: hotfix stagenet 0.49.2

brianp:
Description
---
This includes the fix for a non blocking wallet-ffi initialization required by the mobile apps, which are now running on stagenet.

Motivation and Context
---
Without this change the wallets will block while waiting for a tor connection.

How Has This Been Tested?
---
The wallet team... may have tried it... maybe.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5114,7 +5114,7 @@ dependencies = [
 
 [[package]]
 name = "tari_app_grpc"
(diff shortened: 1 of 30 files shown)


SWvheerden (approved):
Ack

PR #5549, opened 2023-06-28
fix: mempool should use the correct version of the consensus constant

SWvheerden:
Description
---
Mempool uses height 0 when selecting consensus constant.

Motivation and Context
---
Mempool now caches the height from the blocks it processes and uses this height to get the correct consensus when processing transactions. 

How Has This Been Tested?
---
Passes all unit tests

### DIFF
--- a/base_layer/core/src/mempool/mempool_storage.rs
+++ b/base_layer/core/src/mempool/mempool_storage.rs
@@ -238,18 +238,14 @@ impl MempoolStorage {
         self.insert_txs(removed_txs)
             .map_err(|e| MempoolError::InternalError(e.to_string()))?;
         if !new_blocks.is_empty() {
-            self.last_seen_height = new_blocks
-                .last()
-                .and_then(|block| Some(block.header.height))
-                .unwrap_or(0 as u64);
+            self.last_seen_height = new_blocks.last().map(|block| block.header.height).unwrap_or(0 as u64);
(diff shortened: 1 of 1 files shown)


base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -225,6 +230,17 @@ impl MempoolStorage {
...
+        if !new_blocks.is_empty() {
+            self.last_seen_height = new_blocks.last().map(|block| block.header.height).unwrap_or(0 as u64);
+        } else if !removed_blocks.is_empty() {
+            self.last_seen_height = removed_blocks
+                .first()
+                .map(|block| block.header.height)
+                .unwrap_or(1 as u64) -
+                1;
+        } else {
+        }
```
brianp: I don't quite get the significance of the different default return values. If they are insignificant we could cleanup the block with something like:

```suggestion
        if let Some(height) = new_blocks
            .last()
            .or(removed_blocks.first())
            .map(|block| block.header.height)
        {
            self.last_seen_height = height;
        }
```

brianp (changes requested):
This mostly looks good. I found the `if/elseif/empty else` to be a little off putting. 

I'm not sure my suggestion handles all the intended cases. If it does it's a bit cleaner. If it doesn't let's add a comment about the default returns, and the reason for the no-op else.

PR #5551, opened 2023-06-29
fix: mempool fetch_highest_priority_txs

Cifko:
Description
---
Fix the `fetch_highest_priority_txs`.
The process of selection is as follows:
Assume that all transaction have the same weight for simplicity. A(20)->B(2) means A depends on B and A has
fee 20 and B has fee 2. A(20)->B(2)->C(14), D(12)
1) A will be selected first, but B and C will be piggybacked on A, because overall fee_per_byte is 12, so we
  store it temporarily.
2) We look at transaction C with fee per byte 14, it's good, nothing is better.
3) We come back to transaction A with fee per byte 12, but now that C is already in, we recompute it's fee
  per byte to 11, and again we store it temporarily.
4) Next we process transaction D, it's good, nothing is better.
5) And now we proceed finally to transaction A, because there is no other possible better option.

Motivation and Context
---
As it was it was easy to piggyback heavy transaction with low fee_per_byte on light transaction with high fee per byte.

How Has This Been Tested?
---
There are two tests for zero conf, one is for the piggybacking.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/mempool/priority/prioritized_transaction.rs
+++ b/base_layer/core/src/mempool/priority/prioritized_transaction.rs
@@ -76,6 +76,7 @@ pub struct PrioritizedTransaction {
     pub key: usize,
     pub transaction: Arc<Transaction>,
     pub priority: FeePriority,
+    pub fee_per_byte: u64,
     pub weight: u64,
     pub dependent_output_hashes: Vec<HashOutput>,
 }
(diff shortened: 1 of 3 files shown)


SWvheerden (dismissed):
UtAck

brianp (approved):
LGTM

utAck

PR #5552, opened 2023-06-29
fix: remove optional timestamp verification bypass

SWvheerden:
Description
---
In previous refactors we had to bypass timestamp checks during header sync as it we where asking the database for the header timestamps.  We now cache the timestamps when syncing negating this limitation. 
We can thus remove this optional check and always check the header timestamps pass the median rule

How Has This Been Tested?
---
Pass all unit tests
Manual sync of multiple networks.


base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -309,16 +309,16 @@ mod test {
...
-            next.height = 10;
+            next.height = 14;
             let err = validator.validate(next).unwrap_err();
             unpack_enum!(BlockHeaderSyncError::ValidationFailed(val_err) = err);
             unpack_enum!(ValidationError::BlockHeaderError(header_err) = val_err);
             unpack_enum!(BlockHeaderValidationError::InvalidHeight { actual, expected } = header_err);
-            assert_eq!(actual, 10);
-            assert_eq!(expected, 3);
+            assert_eq!(actual, 14);
+            assert_eq!(expected, 13);
```
brianp, line 321: Why did these values require changing? The offset is still the same we just brought the actual and expected closer together, but I don't get why.

brianp (approved):
Ack

> We now cache the timestamps when syncing

Did this also happen in a previous PR?

SWvheerden:
Caching of values happened in a previous refactor.

base_layer/core/src/base_node/sync/header_sync/validator.rs
SWvheerden: So the problem what happened here is that we had 2 blocks in the chain. 
Now the validator took in the latest height, aka 10. And then said, if the header height is ten, we need 9 previous header heights to check the median timestamp.  So it was failing on not enough header timestamps. 
This test we want it to fail on incorrect header height. 

The solution was to make the chain 12, which is the higher than max median timestamps required. Then we can bump of the next header height above 13 to fail on the correct error.

SWvheerden: The fact that they are now closer together is got nothing to do with the values, we just need the initial to be higher than 11,
and then the next header to be higher than the expected
aka its expecting 13 in this case, so we make it 14

Issue #5536, opened 2023-06-28
Sender protocol does not send over features for tx sending


SWvheerden:
When doing a new transaction, the sender protocol does not send over to the receiver all the matching features. 
We need to make sure that all information signed in the challenge of the metadata sand kernel signatures are sent over for the receiver to be used and checked. 

Currently, the version info used is `get_current_version()` which should match for a matching running version of a wallet. But if the wallet versions mismatch this will cause invalid sigantures to be generated. 

A sender should send all the signature challenge features over to the receiver. This should be used when generating signatures and if the receiver does not support that version, the tx should be failed

stringhandler:
is this kernel features or output features?

SWvheerden:
Everything in the Metadata signature and kernel signature challenges should be sent over

PR #5553, opened 2023-06-29
fix(core/transactions)!: consistent output/kernel versions between sender and receiver

mrnaveira:
Description
---
* Added new fields `kernel_version` and `output_version` in `SingleRoundSenderData`
* Updated `protobuf` messages and conversions for the new fields in `SingleRoundSenderData`
* Sender never uses the versions provided by the receiver (e.g. in signatures), only uses its own versions
* Receiver checks (and uses) the versions provided by the sender. Raises an error if the version is not in the allowed range of the consensus constants.
* Implemented the `Default` trait in kernel and output version structs for convenience

Motivation and Context
---
Ref https://github.com/tari-project/tari/issues/5536

When doing a new transaction, the sender protocol must enforce consistent output and kernel versions between senders  and receivers. This is needed to prevent invalid signatures if wallet versions mismatch.

We need to make sure that:
* All information signed in the challenge of the metadata and kernel signatures are sent over to the receiver to be used and checked
* If the receiver does not support the sender's versions, the transaction should be failed
* The sender never uses the receiver versions when building a challenge

How Has This Been Tested?
---
New unit test to check expected behaviour on mismatched versions between sender and receiver.

Manual testing is tricky here, would involve spawning a sender and receiver with mismatched versions (probably by hardcoding values) and try sending transactions between them.

---
Manual testing with mismatched versions between transaction sender and receiver

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] The transaction protocol messages (`protobuf`) formats have new extra fields

**⚠ BREAKING CHANGE: old wallets will be incompatible as the messages for new transactions now expect additional fields**

### DIFF
--- a/base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
@@ -124,6 +124,7 @@ mod test {
                 TransactionKernel,
                 TransactionKernelVersion,
                 TransactionOutput,
+                TransactionOutputVersion,
                 WalletOutput,
             },
             transaction_protocol::{
@@ -202,6 +203,8 @@ mod test {
             ephemeral_public_nonce: ephemeral_public_nonce.clone(),
             covenant: Default::default(),
             minimum_value_promise: MicroTari::zero(),
+            transaction_output_version: TransactionOutputVersion::get_current_version(),
+            transaction_kernel_version: TransactionKernelVersion::get_current_version(),
         };
         let bob_public_key = key_manager
             .get_public_key_at_key_id(&test_params.sender_offset_key_id)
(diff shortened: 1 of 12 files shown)


base_layer/core/src/transactions/transaction_protocol/single_receiver.rs
```diff
@@ -43,9 +46,10 @@ impl SingleReceiverTransactionProtocol {
         sender_info: &SingleRoundSenderData,
         output: WalletOutput,
         key_manager: &KM,
+        consensus_constants: &ConsensusConstants,
     ) -> Result<RecipientSignedMessage, TPE> {
         // output.fill in metadata
```
SWvheerden:
```suggestion
```

SWvheerden (dismissed):
utAck

SWvheerden (approved):

PR #5555, opened 2023-06-29
perf: switch to local-cache for self-hosted runners

CjS77:
We've switched out sccache for local-cache on the self-hosted runners. local-cache is a drop-in replacement for actions/cache, so this should be a very easy change to shave another 2-4 minutes off every PR CI run.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


CjS77:
ACK

AaronFeickert:
Should approval of this PR be contingent on comparing its CI times to earlier PRs? Or is there enough variability that this wouldn't necessarily be useful?

CjS77:
> Should approval of this PR be contingent on comparing its CI times to earlier PRs? Or is there enough variability that this wouldn't necessarily be useful?

It's basically guaranteed to be ±2-4min faster for self-hosted runs vs using actions/cache, since it does the same thing, but stores the cache on the local HDD rather than in the cloud

AaronFeickert:
Maybe I'm misunderstanding what to expect from reported CI times, but looking at the [most recent](https://github.com/tari-project/tari/pull/5553/checks) PR that passes, the test job increased from 14 minutes to 29 minutes?

CjS77:
Yes, because the 1st time you run with a new caching strategy you're definitely miss the cache, so it'll be a slow run.. and subsequent PRs will only use the local cache after the PR is merged

AaronFeickert:
> Yes, because the 1st time you run with a new caching strategy you're definitely miss the cache, so it'll be a slow run

Ah, of course. Questions withdrawn :D

SWvheerden (approved):
Ack

CjS77:
ACK

PR #5556, opened 2023-06-29
chore!: hash domain consistency

brianp:
Description
---
Update more hash domains. My search prowess failed me earlier and I've not found many more domains using inconsistent patterns.

Motivation and Context
---
Consistency.

PR for dan layer: https://github.com/tari-project/tari-dan/pull/580
PR for ios_push_notifications: https://github.com/tari-labs/ios_aurora_push/pull/12

How Has This Been Tested?
---
CI

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

Older nodes will no longer be able to communicate with each other as message signatures have changed.


SWvheerden (dismissed):

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -105,7 +105,7 @@ mod tests {
 
     use super::*;
 
-    hash_domain!(TestHashDomain, "tari.test", 0);
+    hash_domain!(TestHashDomain, "com.tari.test", 0);
```
AaronFeickert: Heh, I suppose even test domains should be consistent :)

base_layer/core/src/lib.rs
```diff
@@ -74,37 +74,25 @@ mod domain_hashing {
...
     pub type MutableOutputMmr = MutableMmr<OutputMmrHasherBlake256, Vec<Hash>>;
     pub type PrunedOutputMmr = MerkleMountainRange<OutputMmrHasherBlake256, PrunedHashSet>;
     pub type MutablePrunedOutputMmr = MutableMmr<OutputMmrHasherBlake256, PrunedHashSet>;
 
-    hash_domain!(
-        InputMmrHashDomain,
-        "com.tari.tari_project.base_layer.core.output_mmr",
-        1
-    );
+    hash_domain!(InputMmrHashDomain, "com.tari.base_layer.core.output_mmr", 1);
```
AaronFeickert: Is this intended to collide with the `OutputMmrHashDomain` string?

base_layer/mmr/tests/support/mod.rs
```diff
@@ -25,7 +25,7 @@ use croaring::Bitmap;
 use tari_crypto::{hash::blake2::Blake256, hash_domain, hashing::DomainSeparatedHasher};
 use tari_mmr::{Hash, HashSlice, MerkleMountainRange, MutableMmr};
 
-hash_domain!(MmrTestHashDomain, "com.tari.tari_project.base_layer.core.kernel_mmr", 1);
+hash_domain!(MmrTestHashDomain, "com.tari.base_layer.core.kernel_mmr", 1);
```
AaronFeickert: This domain string should clearly identify it as being for testing. It currently collides with the `KernelMmrHashDomain` string.

common/src/hashing.rs
```diff
@@ -51,7 +51,7 @@ mod test {
 
     use crate::{hashing::mac_domain_hasher, DomainDigest};
 
-    hash_domain!(HashDomain, "com.tari.tari_project.hash_domain", 1);
+    hash_domain!(HashDomain, "com.tari.hash_domain", 1);
```
AaronFeickert: If this is only intended for testing, the string should probably indicate this for clarity.

common/src/hashing.rs
```diff
@@ -93,7 +93,7 @@ mod test {
 
     #[test]
     fn test_mac_domain_digest() {
-        hash_domain!(MacDomain, "com.tari.tari_project.mac_domain.my_function", 1);
+        hash_domain!(MacDomain, "com.tari.mac_domain.my_function", 1);
```
AaronFeickert: If this is only intended for testing, the string should probably indicate this for clarity.

AaronFeickert (changes requested):

base_layer/core/src/lib.rs
brianp: Could very well have been a mistake originally, which I've carried over. I reckon this should become: 
```suggestion
    hash_domain!(InputMmrHashDomain, "com.tari.base_layer.core.input_mmr", 1);
```

base_layer/mmr/tests/support/mod.rs
brianp: I noticed this while making the updates. I agree I think I'll quickly change test doains to `com.tari.test.xx`

AaronFeickert: Pretty sure I've used something like `com.example.test` elsewhere. I don't think it's essential to have all test domains be particularly consistent or identical, since a production collision shouldn't (knock on wood) ever cause a practical exploit. But it's at least useful for clarity and a good belt-and-suspenders approach, so probably worth doing if not too much work.

AaronFeickert (approved):
~~LGTM provided CI passes, which should catch any issues with the MMR rename.~~

See comment below, which indicates test failure.

AaronFeickert:
Looks like test CI is borked due to a `wasm-pack` issue that is fixed in #5555. Would recommend not merging until that is merged, this PR is rebased, and CI passes.

brianp:
Yeah I agree, the run before this did have a failure, so there could be a mismatch right now.

AaronFeickert:
> Yeah I agree, the run before this did have a failure, so there could be a mismatch right now.

I'll try running locally to see.

AaronFeickert:
My local testing shows that the hash in [this test](https://github.com/tari-project/tari/blob/74f9c35f6a34c1cf731274b7febb245734ae7032/comms/dht/src/dedup/mod.rs#L216) needs to be replaced with `64f1db1039353cf2ff7e09889d14ffd9885766c3dc494a59823a499edfe6997c` in order to pass.

AaronFeickert:
The [genesis block sanity tests](https://github.com/tari-project/tari/blob/74f9c35f6a34c1cf731274b7febb245734ae7032/base_layer/core/src/blocks/genesis_block.rs#L320-L350) also fail.

AaronFeickert (changes requested):
Some tests fail.

The hash in [this test](https://github.com/tari-project/tari/blob/74f9c35f6a34c1cf731274b7febb245734ae7032/comms/dht/src/dedup/mod.rs#L216) needs to be replaced with `64f1db1039353cf2ff7e09889d14ffd9885766c3dc494a59823a499edfe6997c` in order to pass.

The [genesis block sanity tests](https://github.com/tari-project/tari/blob/74f9c35f6a34c1cf731274b7febb245734ae7032/base_layer/core/src/blocks/genesis_block.rs#L320-L350) also fail.

AaronFeickert:
I'm adding fixes to [this branch](https://github.com/AaronFeickert/tari/tree/hash-domain-fixes) in case it's useful.

base_layer/mmr/src/balanced_binary_merkle_proof.rs
```diff
@@ -400,7 +400,7 @@ mod test {
     fn test_real_world_example() {
         hash_domain!(
             ValidatorNodeBmtHashDomain,
-            "com.tari.tari_project.base_layer.core.validator_node_mmr",
+            "com.tari.test.base_layer.core.validator_node_mmr",
```
AaronFeickert: This change fails the test. It may be the case that the domain _should_ collide with the production domain, but the intent of the test and source of the test vectors is unclear. Would appreciate any insights from @sdbondi, who authored the test in #5432.

AaronFeickert: I'm told by @sdbondi that the test can be removed.

hansieodendaal (commented):
I would argue that having the original domain string prefix `"com.tari.tari_project."` is better than the change to `"com.tari."`, as it differentiates `tari_project` from  say `tari_crypto`.

In my mind the prefix should not change.

SWvheerden:
I very much prefer this matching the code structure. 
It makes it easy and consistent as to how the domain should look. This has a broken test or two, but its fixed in: #5561

brianp:
Could also suggest tari crypto be `com.tari_cyrpto`.

I'm more for consistency in any particular pattern we use. `tari_project` wasn't always being used where it should, and was sometimes in the form of `tari-project`. It seemed the most problematic part of the string which is why I opted to remove it. I wouldn't object to setting a more defined "standard" on how to build them though :D

PR #5557, opened 2023-06-29
chore: enable clippy::cast_possible_truncation

SWvheerden:
Description
---
Enables clippy::cast_possible_truncation
Enables clippy::cast_possible_wrap

Motivation and Context
---
This warns us of possible casts in the code base

How Has This Been Tested?
---
unit tests all pass

### DIFF
--- a/applications/tari_base_node/src/commands/command/period_stats.rs
+++ b/applications/tari_base_node/src/commands/command/period_stats.rs
@@ -109,14 +109,14 @@ impl CommandContext {
             period_tx_count += block.block().body.kernels().len() - 1;
             period_block_count += 1;
             let st = if prev_block.header().timestamp.as_u64() >= block.header().timestamp.as_u64() {
-                1.0
+                1
             } else {
-                (block.header().timestamp.as_u64() - prev_block.header().timestamp.as_u64()) as f64
+                block.header().timestamp.as_u64() - prev_block.header().timestamp.as_u64()
             };
             let diff = block.accumulated_data.target_difficulty.as_u64();
             period_difficulty += diff;
-            period_solvetime += st as u64;
-            period_hash += diff as f64 / st / 1_000_000.0;
+            period_solvetime += st;
+            period_hash += diff as f64 / st as f64 / 1_000_000.0;
             if period_ticker_end <= period_end {
                 break;
             }
--- a/base_layer/wallet/tests/utxo_scanner/mod.rs
+++ b/base_layer/wallet/tests/utxo_scanner/mod.rs
@@ -20,7 +20,12 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{collections::HashMap, convert::TryInto, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    convert::{TryFrom, TryInto},
+    sync::Arc,
+    time::Duration,
+};
 
 use chrono::{Duration as ChronoDuration, Utc};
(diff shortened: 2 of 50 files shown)


base_layer/wallet/tests/utxo_scanner/mod.rs
```diff
@@ -227,6 +232,9 @@ pub struct TestBlockData {
 /// Generates a set of block headers and key manager outputs for each header. The `birthday_offset` specifies at which
 /// block in the `num_block` the birthday timestamp will have passed i.e. it occured during the previous block period.
 /// e.g. with `num_blocks` = 10 and `birthday_offset` = 5 the birthday timestamp will occur between block 4 and 5
+
+// casting wrap around here is okay
```
brianp: Feels vague. Maybe a little insight into _why_ it's okay here for people who may not know.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2916,22 +2935,23 @@ pub unsafe extern "C" fn liveness_data_get_public_key(
...
     error_out: *mut c_int,
-) -> c_int {
+) -> c_uint {
     let mut error = 0;
     ptr::swap(error_out, &mut error as *mut c_int);
     if liveness_data.is_null() {
         error = LibWalletError::from(InterfaceError::NullError("liveness_data".to_string())).code;
         ptr::swap(error_out, &mut error as *mut c_int);
-        return -1;
+        return 0;
```
brianp, line 2949: The docs for the function state:
`value of '-1' if it is None. Note that it also returns '-1' if liveness_data is null.`

This may effect mobile code, and makes the docs out of date.

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -2916,22 +2935,23 @@ pub unsafe extern "C" fn liveness_data_get_public_key(
...
     }
     if let Some(latency) = (*liveness_data).latency() {
-        latency as c_int
+        latency as c_uint
     } else {
-        -1
+        0
```
brianp, line 2954: Same as above ^

brianp (approved):
Ack

- One improved comment would be nice
- Spot in wallet ffi I just want to acknowledge will be a change of public API that will effect mobile

base_layer/wallet/tests/utxo_scanner/mod.rs
SWvheerden: added more detail to the comment

base_layer/wallet_ffi/src/lib.rs
SWvheerden: updated the docs, felt that a latency of 0ms is more unlikely as a latency wrap-around. It also stop the check of checking for <0 rather than == -1, which should now by == 0

brianp (approved):

applications/tari_base_node/src/commands/command/period_stats.rs
```diff
@@ -109,14 +109,14 @@ impl CommandContext {
...
             } else {
-                (block.header().timestamp.as_u64() - prev_block.header().timestamp.as_u64()) as f64
+                block.header().timestamp.as_u64() - prev_block.header().timestamp.as_u64()
             };
             let diff = block.accumulated_data.target_difficulty.as_u64();
             period_difficulty += diff;
-            period_solvetime += st as u64;
-            period_hash += diff as f64 / st / 1_000_000.0;
+            period_solvetime += st;
+            period_hash += diff as f64 / st as f64 / 1_000_000.0;
```
AaronFeickert, line 119: This can still cause issues since `f64` is more restricted than `u64`, right?

SWvheerden: if the u64 is really big, then yes converting to f64 might introduce some inaccuracies, but not going to be completely wrong. But here it should still kinda give a near enough answer. 
See this example of the lossy translation to f64: https://stackoverflow.com/questions/75518263/convert-u64-to-f64

PR #5559, opened 2023-06-30
chore: make sure mainnet values are set as placeholder

SWvheerden:
Description
---
Added a comment to ensure people know that the mainnet values are placeholder


brianp (approved):
Ack

PR #5560, opened 2023-06-30
feat: add a not before proof

hansieodendaal:
Description
---
Added a `not before proof` to the genesis block. The data to be added is just a simple Vec<u8>, which can represent anything, but must be limited in size.

_**Note:** This is to implement what some members of the Tari community have been discussing. The `not before proof` or how it is implemented is not my idea; I merely executed it. That being said, I had some fun with choosing the examples._

Motivation and Context
---
The basic idea behind adding a `not before proof` to the genesis block is to provide some level of authenticity. If we would, for example, insert a digitized headline of the _New York Times_  just before mainnet launches, it will provide a timestamped reference to a widely recognized and verifiable event and prove that the blockchain has not been in existence before that time. 

How Has This Been Tested?
---
Genesis block sanity test

---
Review changes

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -26,12 +26,16 @@ use chrono::{DateTime, FixedOffset};
 use tari_common::configuration::Network;
 use tari_common_types::types::{FixedHash, PrivateKey};
 use tari_crypto::tari_utilities::hex::*;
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):
Ack
This is a good proof of concept for a not_before proof

PR #5564, opened 2023-06-30
fix: remove statement from sparse Merkle tree proofs

### LINKED ISSUE #5527: Sparse Merkle tree proofs include statement
Sparse Merkle trees and related proofs were added in #5457, though they aren't yet used in any consensus-related applications.

The [proof structure](https://github.com/tari-project/tari/blob/2e400836111c14fdfc52d84e5a64bbbb2d438956/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L20-L26) for both inclusion and exclusion proofs includes the key and value hashes for the node in question, which means it effectively includes part of the proof statement (it does not include the root hash).

This seems like a bit of bad code smell. For both [inclusion](https://github.com/tari-project/tari/blob/2e400836111c14fdfc52d84e5a64bbbb2d438956/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L69-L78) and [exclusion](https://github.com/tari-project/tari/blob/2e400836111c14fdfc52d84e5a64bbbb2d438956/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L80-L84) proofs, the verifier provides the root hash and node key hash. For inclusion proofs, the verifier additionally provides the expected value hash; this is implicit for exclusion proofs. This requires that the verifier check the consistency of the provided data with that of the proof, which seems unnecessary. Additionally, one could imagine cases where an implementation mistakenly assumes the validity of the statement data within a proof provided by a malicious prover.

A more idiomatic approach would be to remove the key and value hashes from the proof and have them be provided solely by the verifier.

AaronFeickert:
Description
---
Removes the key and value hashes from sparse Merkle tree proofs. Adds a path check to inclusion proof verification. Closes #5527.

Motivation and Context
---
Currently, sparse Merkle tree proofs [include](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L23-L24) the key and value hashes for the proof. As these are effectively the proof statement, it doesn't really make sense to include them in the proof structure.

Including them means additional work for the verifier for both [inclusion](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L75-L76) and [exclusion](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L83) proofs that isn't needed. This design also seems like a footgun that could lead an implementer to unintentionally trust the statement from a malicious prover.

This PR removes the key and value hashes from the proof structure altogether, and updates relevant functions accordingly. It also adds a path check to the inclusion proof verifier for consistency with the exclusion proof verifier, which may also help with faster detection of invalid proofs.

How Has This Been Tested?
---
Existing tests pass.

---
Check that the new verification logic is correct.

Breaking Changes
---
None. While this changes the structure of sparse Merkle tree proofs, they are not currently used and do not have serialization defined.


AaronFeickert:
Marking as draft while investigating some unexpected behavior with the existing exclusion proof code, as well as an [issue](https://github.com/tari-project/tari/issues/5565) relating to keys.

base_layer/mmr/src/lib.rs
```diff
@@ -19,7 +19,6 @@
 // SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
```
AaronFeickert: Did the linter complain about this?

base_layer/mmr/src/sparse_merkle_tree/node.rs
```diff
@@ -110,8 +110,75 @@ hash_type!(NodeHash);
...
+    // position *after* next bit when going backwards
+    cursor_back: usize,
+    key: &'a NodeKey,
+}
+
+impl PathIterator<'_> {
+    pub fn new(key: &NodeKey) -> PathIterator {
+        PathIterator {
+            cursor_front: 0,
+            cursor_back: key.len() * 8,
```
AaronFeickert: Looks like this could overflow. Might be worth adding a check just in case, although this makes for a more awkward `Result`-based constructor...

base_layer/mmr/src/sparse_merkle_tree/node.rs
```diff
@@ -110,8 +110,75 @@ hash_type!(NodeHash);
...
+        Some(bit_to_dir(bit))
+    }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        if self.cursor_front >= self.cursor_back {
+            return None;
+        }
+        self.cursor_back -= n;
+        self.next_back()
+    }
```
AaronFeickert: I think the `None` condition here is incorrect, as the function should yield `None` if the full lookback would exceed the length (which in this case is the distance between the cursors). I think the call to `next_back` would correctly handle this, but that makes the initial `None` check in this function redundant anyway.

I'd say either rely on this function to do a proper `None` check against the full lookback, or offload the check entirely to `next_back`.

base_layer/mmr/src/sparse_merkle_tree/node.rs
```diff
@@ -526,4 +594,43 @@ mod test {
...
+
+    #[test]
+    fn path_iterator_connectors() {
+        let key = NodeKey::from(&[0; 32]);
+        let iter = key.as_directions();
+        assert_eq!(iter.len(), 256);
+        assert_eq!(iter.take(14).len(), 14);
+        let iter = key.as_directions();
+        assert_eq!(iter.rev().take(18).len(), 18);
+    }
```
AaronFeickert: I assume this is a sanity check against whatever issues arose with `size_hint`?

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
-    pub fn new(path: Vec<TraverseDirection>, siblings: Vec<NodeHash>, key: NodeKey, value: Option<ValueHash>) -> Self {
-        Self {
-            path,
-            siblings,
-            key,
-            value,
-            phantom: PhantomData::<H>::default(),
-        }
-    }
+/// An exclusion proof for a key in a sparse merkle tree.
```
AaronFeickert: `s/merkle/Merkle`

Gotta give [credit](https://en.wikipedia.org/wiki/Ralph_Merkle) where credit is due!

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
     NodeKey,
     SMTError,
     SparseMerkleTree,
     ValueHash,
-    EMPTY_NODE_HASH,
 };
 
-pub struct MerkleProof<H> {
-    path: Vec<TraverseDirection>,
+/// An inclusion proof for a key-value pair in a sparse merkle tree.
```
AaronFeickert: `s/merkle/Merkle`

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
+///
+/// Given a sparse merkle tree, `tree`, you can create a proof that a certain key-value pair exists by calling
```
AaronFeickert: `s/merkle/Merkle`

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
-            path,
-            siblings,
-            key,
-            value,
-            phantom: PhantomData::<H>::default(),
-        }
-    }
+/// An exclusion proof for a key in a sparse merkle tree.
+///
+/// Given a sparse merkle tree, `tree`, you can create a proof that a certain key does *not exist* in the tree by
```
AaronFeickert: `s/merkle/Merkle`

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
 
-    fn calculate_root_hash(&self) -> NodeHash {
-        let node_hash = match self.value.as_ref() {
-            Some(v) => LeafNode::<H>::hash_value(&self.key, v),
-            None => EMPTY_NODE_HASH,
-        };
-        let n = self.siblings.len();
-        let hash = self.siblings.iter().zip(self.path.iter()).rev().enumerate().fold(
-            node_hash,
+    /// Calculate the merkle tree root for this proof, given the key and value hash.
```
AaronFeickert: `s/merkle/Merkle`

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -271,9 +271,10 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
         Ok(node.map(|n| n.as_leaf().unwrap().value()))
     }
 
-    /// Constructs a Merkle proof for the value at location `key`.
-    pub fn build_proof(&self, key: &NodeKey) -> Result<MerkleProof<H>, SMTError> {
-        let mut path = Vec::new();
+    /// Construct the data structures needed to generate the merkle proofs. Although this function returns a struct
```
AaronFeickert: `s/merkle/Merkle`

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -6,57 +6,117 @@ use std::marker::PhantomData;
...
+    // The terminal node of the tree proof, or `None` if the the node is `Empty`.
+    leaf: Option<LeafNode<H>>,
+    phantom: std::marker::PhantomData<H>,
+}
 
-    pub fn from_tree(tree: &SparseMerkleTree<H>, key: &NodeKey) -> Result<Self, SMTError> {
-        tree.build_proof(key)
-    }
+trait MerkleProofDigest<H: Digest<OutputSize = U32>> {
+    /// Returns an array to the vector of sibling hashes along the path to the key's leaf node for this proof.
```
AaronFeickert, line 103: > array to the vector

Kinda confusing. Maybe just "slice of sibling hashes" instead?

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -65,30 +125,98 @@ impl<H: Digest<OutputSize = U32>> MerkleProof<H> {
...
+    pub fn new(siblings: Vec<NodeHash>) -> Self {
+        Self {
+            siblings,
+            phantom: PhantomData::<H>::default(),
+        }
+    }
+
+    /// Generates an inclusion proof for the given key and value hash from the given tree. If the key does not exist in
+    /// tree, or the key does exist, but the value hash does not match, then `from_tree` will return a
+    /// `NonViableProof` error.
```
AaronFeickert, line 142: It can also propagate a `StaleHash` error from `non_failing_exclusion_proof`.

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -271,9 +271,10 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
...
+    /// of type `ExclusionProof` it is not really a valid (exclusion) proof. The constructors do additional
+    /// validation before passing the structure on. For this reason, this method is `private` outside of the module.
+    pub(crate) fn non_failing_exclusion_proof(&self, key: &NodeKey) -> Result<ExclusionProof<H>, SMTError> {
```
AaronFeickert: Returning a wrapped `ExclusionProof` here feels wrong, even though any other approach (using a type alias for clarity, defining a custom return struct, or returning a wrapped tuple with the sibling and leaf data) seems like overkill. Maybe just rename the function to something like `get_siblings_and_leaf_for_proofs` as an added indicator that you're not really getting a proof out? But this is more of a nit.

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -65,30 +125,98 @@ impl<H: Digest<OutputSize = U32>> MerkleProof<H> {
...
+            phantom: PhantomData::<H>::default(),
+        }
     }
 
-    pub fn validate_exclusion_proof(&self, expected_key: &NodeKey, expected_root: &NodeHash) -> bool {
-        path_matches_key(expected_key, &self.path) &&
-            expected_root == &self.calculate_root_hash() &&
-            (self.value.is_none() || &self.key != expected_key)
+    /// Generates an exclusion proof for the given key from the given tree. If the key exists in the tree then
+    /// `from_tree` will return a `NonViableProof` error.
```
AaronFeickert, line 186: This can also return a `StaleHash` error from `non_failing_exclusion_proof`.

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
```diff
@@ -121,26 +249,41 @@ mod test {
...
+        // exclusion proof assertions
+        let ex_proof = ExclusionProof::from_tree(&tree, &key2).unwrap();
+        assert!(!ex_proof.validate(&key, &hash2));
+        assert!(!ex_proof.validate(&key, &hash));
+        assert!(!ex_proof.validate(&key2, &hash));
+        assert!(ex_proof.validate(&key2, &hash2));
     }
 
     #[test]
     fn merkle_proofs() {
```
AaronFeickert: I assume the PRNG seeds used for exclusion proofs happen to exercise both of the disjunctive verification criteria  (empty leaf node and non-matching leaf node), but it isn't clear from just reading the test that this in fact happens. Would recommend a modified or additional test that specifically checks for both cases.

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
AaronFeickert, line 1: There doesn't appear to be a test that exercises the `StaleHash` error condition that can be returned by the inclusion and exclusion proof constructors. If this error can reasonably arise in practice, it'd be a good idea to test for it; if it can't/won't arise, would be worth a comment.

base_layer/mmr/src/lib.rs
CjS77: 🤷 
Could have been `cargo fmt`. Could have been a fat finger

base_layer/mmr/src/sparse_merkle_tree/node.rs
CjS77: Yeah, I suppose. In practice len() is hard-coded to 32 (see `hash_type!`), but this is not coded as an invariant.

A more pragmatic approach might be to pull out the 32 as a const and reference it here.

base_layer/mmr/src/sparse_merkle_tree/node.rs
CjS77: Jip. This test failed prior to overriding size_hint.

base_layer/mmr/src/sparse_merkle_tree/tree.rs
CjS77: or `build_pre_proof`? `build_proof_candidate`?

CjS77 (commented):
Addressed some of the comments. Others will be tackled after some sleep

base_layer/mmr/src/sparse_merkle_tree/tree.rs
AaronFeickert: Even better! I think I like `build_proof_candidate` best. Seems like a good compromise between clarity and creating internal types unnecessarily.

base_layer/mmr/src/lib.rs
AaronFeickert: I'd put it back. Otherwise it'll probably bother me until I decide to PR it back in :D

CjS77: You're OCD about an empty line?

base_layer/mmr/src/sparse_merkle_tree/node.rs
CjS77: You're right. I find the documentation for this function to be ambiguous. I interpreted a failed nth_back to _not_ consume the entire iterator, but return the cursor to where it was. I've checked this now, and your interpretation is correct.

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
CjS77: Indeed. In fact there are a few cases still to be covered after examining the source code coverage

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
CjS77: I figured the data structure had been commoditised by now.. https://www.quora.com/What-common-non-proper-nouns-derive-from-the-names-of-individual-people

It's a great compliment, really :)

CjS77:
All cases should be covered now.

I've also elucidated on some properties of the exclusion proof in the tests, where a single exclusion proof actually covers all missing keys with a common prefix up until the first included key.

base_layer/mmr/src/sparse_merkle_tree/proofs.rs
CjS77: I've replaced the random tests with specific tests that should exercise every branch and failure mode

CjS77:
Covered in #5768

PR #5566, opened 2023-06-30
fix: sparse Merkle tree key querying

### LINKED ISSUE #5565: Sparse Merkle trees report keys incorrectly
Sparse Merkle trees may be [queried for keys](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/tree.rs#L256-L259) using the `contains` function.

This function will correctly return `true` if queried with a key that has been added to the tree via an `upsert` call. However, it will incorrectly return `true` if queried with a key that has not been added to the tree.

Aside from reporting keys incorrectly, it's unclear if there are broader implications for this.

AaronFeickert:
Description
---
Fixes key querying on sparse Merkle trees. Adds a test. Closes #5565.

Motivation and Context
---
It's possible to ask a sparse Merkle tree if it [contains](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/tree.rs#L256-L259) a particular key. However, the query returns `true` whether or not the key is in the tree.

This PR fixes the incorrect behavior. The query returns `true` if and only if the tree search yields a leaf node with the given key. If the tree is malformed such that the search returns an error, the query fails safe and returns `false`.

Interestingly, the existing unit tests did not detect the incorrect behavior, as no test checked for the case where a key did not exist in a tree. We add a test that checks this for a few cases of interest.

How Has This Been Tested?
---
Existing tests pass. A new test asserts the expected behavior (and would have caught the previous incorrect behavior).

---
Confirm that the detection logic in the `contains` function is correct.

Breaking Changes
---
None.

### DIFF
--- a/base_layer/mmr/src/sparse_merkle_tree/tree.rs
+++ b/base_layer/mmr/src/sparse_merkle_tree/tree.rs
@@ -255,7 +255,14 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
 
     /// Returns true if the tree contains the key `key`.
     pub fn contains(&self, key: &NodeKey) -> bool {
-        self.search_node(key).ok().is_some()
+        match self.search_node(key) {
+            // In the case of a malformed tree where the search fails unexpectedly, play it safe
+            Err(_) => false,
+            // The node is either empty or is a leaf with an unexpected key
+            Ok(None) => false,
+            // The node is a leaf with the expected key
+            Ok(Some(_)) => true,
+        }
     }
 
     /// Returns the value at location `key` if it exists, or `None` otherwise.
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):
utAck

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -255,7 +255,14 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
...
     pub fn contains(&self, key: &NodeKey) -> bool {
-        self.search_node(key).ok().is_some()
+        match self.search_node(key) {
+            // In the case of a malformed tree where the search fails unexpectedly, play it safe
+            Err(_) => false,
+            // The node is either empty or is a leaf with an unexpected key
+            Ok(None) => false,
+            // The node is a leaf with the expected key
+            Ok(Some(_)) => true,
+        }
```
CjS77, line 265: Weird. This is exactly what the original code is supposed to do.

`<Result>.ok()` takes `Result<T,E> -> Option<T>` returning `None` on an error. Then `is_some()` only returns true if the option was `Some<T>`

CjS77: Aha! See the mistake. There's a nested Option there. So this is what the function should have been:

```
/// Returns true if the tree contains the key `key`.
    pub fn contains(&self, key: &NodeKey) -> bool {
        self.search_node(key)
            .ok()
            .flatten()
            .is_some()
    }
 ```
 
 which passes the test

CjS77 (approved):
Great. Happy to take the fix as written. If there's a desire, we can switch to the functional form (as given in the comments) at another point.

PR #5572, opened 2023-07-04
fix: error out the stx protocol if the sender sends unsupported data

SWvheerden:
Description
---
Will error out the stx protocol if the sender sends unsupported script, features, or covenant

Motivation and Context
---
Its possible for the sender to block the spending of the transaction by using unsupported script, covenant or features. This will now make it that the receiver will not accept a transaction it does not know how to spend. 

How Has This Been Tested?
---
Passes all unit tests.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -233,7 +233,7 @@ where
                     .map(OutputManagerResponse::Balance)
             },
             OutputManagerRequest::GetRecipientTransaction(tsm) => self
-                .get_recipient_transaction(tsm)
+                .get_default_recipient_transaction(tsm)
                 .await
                 .map(OutputManagerResponse::RecipientTransactionGenerated),
             OutputManagerRequest::GetCoinbaseTransaction {
@@ -680,7 +680,7 @@ where
     }
 
     /// Request a receiver transaction be generated from the supplied Sender Message
-    async fn get_recipient_transaction(
+    async fn get_default_recipient_transaction(
         &mut self,
         sender_message: TransactionSenderMessage,
     ) -> Result<ReceiverTransactionProtocol, OutputManagerError> {
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -696,7 +696,7 @@ where
 
         // Confirm output features is default
         if single_round_sender_data.features != OutputFeatures::default() {
-            return Err(OutputManagerError::InvalidCovenant);
+            return Err(OutputManagerError::InvalidOutputFeatures);
         }
 
         let (spending_key_id, _, script_key_id, script_public_key) =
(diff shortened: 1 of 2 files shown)


AaronFeickert:
Do existing unit tests suffice to exercise the desired behavior and failure modes?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -690,14 +690,29 @@ where
...
-        if single_round_sender_data.script != script!(Nop) {
-            return Err(OutputManagerError::InvalidScriptHash);
+        // Confirm covenant is default
+        if single_round_sender_data.covenant != Covenant::default() {
+            return Err(OutputManagerError::InvalidCovenant);
+        }
+
+        // Confirm output features is default
+        if single_round_sender_data.features != OutputFeatures::default() {
+            return Err(OutputManagerError::InvalidCovenant);
```
hansieodendaal: Why would we want to force default features only?
- `maturity` > 0 can be accepted. Ideally, a user setting should determine how many blocks in the future is acceptable.
- `sidechain_feature` should be allowed, based on a user setting.
- `range_proof_type` of `BulletProofPlus` and `RevealedValue` can be accepted, based on a user setting.

_Note: The error should be `InvalidOutputFeatures` in this context._

hansieodendaal (commented):
Please see comments below

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: The default settings is based on current allowed features for the nextnet /stagenet networks. 
Which means all of those methods can be locked down. 

`maturity` while allowed should not be allowed for default transactions which are kinda for what the stx protocol is. Maturity is currently only used for coinbases. And left open to use in very custom use cases, which is beyond the scope of the stx protocol.

hansieodendaal: I would argue that the stx protocol should not be limited here for the 3x items I mentioned - that is enforced with the aggregate body internal validators. Those items do not fall into the category _the receiver will not accept a transaction it does not know how to spend_.

SWvheerden: So my expanded reasoning why we should lock it down and not include any of that is as follows. 
This piece of code for the stx protocol is for accepting default transactions. 
The following features are used for:
`maturity`: used by coinbases and custom transactions for limiting spending. So why would we be default allow a sender to block spending out our own utxos?
`sidechain_feature` : this is used by our own node creating special transactions like VN registration. Why would your node accept a utxo for spending that is flagged with such a feature. 
`rangeproof type` is currently not even allowed on most networks, and why would your node want to be limited to the utxo privacy type you want?


Look don't get me wrong, we should allow all weird combinations of transactions. But they do not belong in the stx protocol where the wallet auto accepts them. They are all custom transactions with custom rules.

brianp: If this is an issue of allowed features based on network type, why don't we use conditional compilation for the network types and adjust the behavior as needed?

brianp (commented):
The changes themselves look okay if it's what we really want, but I'm not sure this is the solution for the problem. Feels like an intermediary / stop gap, that we can also support via conditional compilation per network.

hansieodendaal (approved):
utACK

PR #5579, opened 2023-07-08
fix: peer connection to stale nodes

SWvheerden:
Description
---
This fixes nodes continuously selecting bad stale nodes as connections and attempting to connect to them. 

Motivation and Context
---
If nodes have bad nodes in their peer database they should not keep trying to connect to them. 
This will attempt to connect to a node once, and if the node was never online and it tried to connect to the node once, it will ignore the node till the node tries to connect to it first.

This stops the problem with if a node has too many stale connections in its peer database, it will keep selecting stale nodes and try to connect to them. If the amount of stale nodes are too much, it can cause a constant read/write on the peer database of the node. 

How Has This Been Tested?
---
Manually verified with database with mostly stale nodes. Node will try them once, then ignore them.

### DIFF
--- a/comms/core/src/peer_manager/peer.rs
+++ b/comms/core/src/peer_manager/peer.rs
@@ -158,6 +158,23 @@ impl Peer {
         &self.supported_protocols
     }
 
+    pub fn last_connect_attempt(&self) -> Option<NaiveDateTime> {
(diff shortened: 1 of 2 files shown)


hansieodendaal (approved):
utACK

PR #5582, opened 2023-07-10
fix: miner delay attack

SWvheerden:
Description
---
Stops the node blocking until it has a full block

Audit Finding Number
---
TARI-0001, TARI-0002

Motivation and Context
---
Its possible for a malicious node to block access to the full block by using first the mempool to say it does not have the missing transactions, then when asked for the full block to only then provide the full block. But these requests can be made slow to delay the node constructing the new block by almost 2 mins. This is the block time. 

By only accepting 1 block request at a time, a malicious node can lock down the local node for that entire 2 mins knowing they wont accept any other blocks. This changes it so that nodes can still process new requests. The first complete block will be served first. 

Also changes the display of block information to ensure block sorted is correctly used.

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -446,8 +446,9 @@ where B: BlockchainBackend + 'static
             );
             return Ok(());
         }
-        {
-            // we use a double lock to make sure we can only reconcile one block at a time
+
+       {
+            // we use a double lock to make sure we can only reconcile one unique block at a time. We may receive the same block from multiple peer near simultaneously. We should only reconcile each unique block once.
             let read_lock = self.list_of_reconciling_blocks.read().await;
             if read_lock.contains(&block_hash) {
                 debug!(
@@ -468,6 +469,14 @@ where B: BlockchainBackend + 'static
                 );
                 return Ok(());
             }
+            if write_lock.contains(&block_hash) {
+                debug!(
+                    target: LOG_TARGET,
+                    "Block with hash `{}` is already being reconciled",
+                    block_hash.to_hex()
+                );
+                return Ok(());
+            }
             write_lock.insert(block_hash);
         }
 
@@ -479,15 +488,14 @@ where B: BlockchainBackend + 'static
             source_peer
         );
 
-        // We need to make sure that we can ask for the block from multiple sources as a single source can maliciously
-        // delay the block blocking us from getting to true latest tip
-        let block = self.reconcile_block(source_peer.clone(), new_block).await?;
+
+        let block_result = self.reconcile_block(source_peer.clone(), new_block).await;
 
         {
             let mut write_lock = self.list_of_reconciling_blocks.write().await;
             write_lock.remove(&block_hash);
         }
-
+        let block = block_result?;
         self.handle_block(block, Some(source_peer)).await?;
 
         Ok(())
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -23,7 +23,6 @@
 use std::{
     collections::HashSet,
     convert::{TryFrom, TryInto},
-    ops::DerefMut,
     sync::Arc,
     time::{Duration, Instant},
 };
@@ -33,7 +32,7 @@ use strum_macros::Display;
 use tari_common_types::types::{BlockHash, FixedHash, HashOutput};
 use tari_comms::{connectivity::ConnectivityRequester, peer_manager::NodeId};
 use tari_utilities::hex::Hex;
-use tokio::sync::Mutex;
+use tokio::sync::RwLock;
 
 use crate::{
     base_node::{
@@ -83,7 +82,7 @@ pub struct InboundNodeCommsHandlers<B> {
     blockchain_db: AsyncBlockchainDb<B>,
     mempool: Mempool,
     consensus_manager: ConsensusManager,
-    list_of_reconciling_blocks: Arc<Mutex<HashSet<HashOutput>>>,
+    list_of_reconciling_blocks: Arc<RwLock<HashSet<HashOutput>>>,
     outbound_nci: OutboundNodeCommsInterface,
     connectivity: ConnectivityRequester,
 }
@@ -105,7 +104,7 @@ where B: BlockchainBackend + 'static
             blockchain_db,
             mempool,
             consensus_manager,
-            list_of_reconciling_blocks: Arc::new(Mutex::new(HashSet::new())),
+            list_of_reconciling_blocks: Arc::new(RwLock::new(HashSet::new())),
             outbound_nci,
             connectivity,
         }
@@ -448,21 +447,30 @@ where B: BlockchainBackend + 'static
             return Ok(());
         }
         {
-            // we have this in mutex to ensure we dont have any funnies with rwlock with multiples requests first
-            // passing the check then all wanting to reconcile. We should only have 1 request passing
-            let mut lock = self.list_of_reconciling_blocks.lock().await;
-            let list_of_reconciling_blocks = lock.deref_mut();
-            if list_of_reconciling_blocks.contains(&block_hash) {
+            // we use a double lock to make sure we can only reconcile one block at a time
+            let read_lock = self.list_of_reconciling_blocks.read().await;
+            if read_lock.contains(&block_hash) {
                 debug!(
                     target: LOG_TARGET,
                     "Block with hash `{}` is already being reconciled",
                     block_hash.to_hex()
                 );
                 return Ok(());
-            } else {
-                list_of_reconciling_blocks.insert(block_hash);
             }
         }
+        {
+            let mut write_lock = self.list_of_reconciling_blocks.write().await;
+            if self.blockchain_db.block_exists(block_hash).await? {
+                debug!(
+                    target: LOG_TARGET,
+                    "Block with hash `{}` already stored",
+                    block_hash.to_hex()
+                );
+                return Ok(());
+            }
+            write_lock.insert(block_hash);
+        }
+
         debug!(
             target: LOG_TARGET,
             "Block with hash `{}` is unknown. Constructing block from known mempool transactions / requesting missing \
@@ -476,9 +484,8 @@ where B: BlockchainBackend + 'static
         let block = self.reconcile_block(source_peer.clone(), new_block).await?;
 
         {
-            let mut lock = self.list_of_reconciling_blocks.lock().await;
-            let list_of_reconciling_blocks = lock.deref_mut();
-            list_of_reconciling_blocks.remove(&block_hash);
+            let mut write_lock = self.list_of_reconciling_blocks.write().await;
+            write_lock.remove(&block_hash);
         }
 
         self.handle_block(block, Some(source_peer)).await?;
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -488,19 +488,23 @@ where B: BlockchainBackend + 'static
             source_peer
         );
 
-        let block_result = self.reconcile_block(source_peer.clone(), new_block).await;
+        let result = self.reconcile_and_add_block(source_peer.clone(), new_block).await;
 
-        if block_result.is_err() {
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -436,14 +436,27 @@ where B: BlockchainBackend + 'static
             return Ok(());
         }
 
+        // Lets check if the block exists before we try and ask for a complete block
+        if self.blockchain_db.block_exists(block_hash).await? {
```
sdbondi, line 441: Problem is that it doesnt exist until after the block has been added (which is the reason for the lock). 

Many peers send the block to us so this will be the normal case. I suggest holding a hashset in memory of "in progress" blocks, the block hash is inserted where this check is and exits early($) for duplicates. The block is removed from the hashset when the critical area is left (i.e. once the block is added or if an error occurs). In this way competing blocks may be added at the same time in a race but the same block will not cause a bunch of reconciliations to many nodes at the same time for the same block. 

$ Exit early could be a problem as we may want to reconcile with the next peer (the lock worked as a queue). But I suspect these cases will be rare enough that it's fine if the base node can go into sync mode if it missed the highest block, or some late messages will arrive and the block can be reconciled if it wasnt added before e.g. due to network failure from another peer.

sdbondi: Oh I see the lock is still in below - I'm not sure this rectifies the problem then as a peer could still force a delay

EDIT: ah ok I missed that you moved reconcile up before the lock. I think there may be a way to avoid processing/reconciling the same block over and over a few times, the first `exists` more often than not won't catch them, but might get a little complex. This PR should prevent the delay.

SWvheerden: yeah the malicious peer can still force a delay on their block, but only their block. But an honest block will go forward and proceed through while the bad block is being delayed. 

And thats the thing we need. We cannot have a single peer lock up the block process for 2 minutes while they mine the next block and all the other miners have to wait.

AaronFeickert:
Is it possible to label this PR to reflect its status as an audit fix? It should be easy for interested researchers and developers to quickly find audit-related fixes.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -369,7 +369,8 @@ impl AggregateBody {
     }
 
     pub fn is_sorted(&self) -> bool {
-        self.sorted
+        // a block containing only a single kernel, single output and single input is sorted by default
+        self.sorted || (self.kernels.len() <= 1 && self.outputs.len() <= 1 && self.inputs.len() <= 1)
```
stringhandler: This is an unrelated fix and should be in a different PR

stringhandler (commented):
I think this is not a bad start.

1. I like doing the reconcile before the semaphore, but I think you should remove the semaphore entirely since it no longer serves a purpose.
2. This solution introduces another problem, as @sdbondi noted, we will be requesting many blocks. I suggest only doing one reconcile per block hash. This could be done using a hashmap/set in this file, or potentially even in `base_node_service.rs::spawn_handle_incoming_block` and only spawning a thread if there is no thread currently handling this blockhash

SWvheerden:
I disagree about removing the semaphore, after this point, we are making changes to the tip, and rely on the tip to stay constant during the process.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -452,7 +447,22 @@ where B: BlockchainBackend + 'static
             );
             return Ok(());
         }
-
+        {
+            // we have this in mutex to ensure we dont have any funnies with rwlock with multiples requests first
+            // passing the check then all wanting to reconcile. We should only have 1 request passing
+            let mut lock = self.list_of_reconciling_blocks.lock().await;
```
stringhandler: This could be made better by using a read/write lock and reading for the `contains` call, then obtaining a write for the insert call.

By adding the fast read, you avoid the attack of spamming a node with blocks and locking it up

stringhandler: There may also be concurrent hashmaps that do this for you

SWvheerden: I looked at using a RWLock, but decided to use a Mutex because its safer and does what we want. 

Take this scenario:
Block 1-A -> read lock
Block 1-B -> read lock

Both has the read lock, both see that Block 1 does not exist. 

Both block 1-A and Block 1-B now get reconciled. And this should not happen.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -461,7 +471,16 @@ where B: BlockchainBackend + 'static
             source_peer
         );
 
+        // We need to make sure that we can ask for the block from multiple sources as a single source can maliciously
+        // delay the block blocking us from getting to true latest tip
         let block = self.reconcile_block(source_peer.clone(), new_block).await?;
+
+        {
+            let mut lock = self.list_of_reconciling_blocks.lock().await;
```
sdbondi: This should also happen if there is an error

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
SWvheerden: We can just make it write locks, that solves this, but then again we might just use a mutex

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -461,7 +471,16 @@ where B: BlockchainBackend + 'static
             source_peer
         );
 
+        // We need to make sure that we can ask for the block from multiple sources as a single source can maliciously
```
sdbondi: Not sure I understand this comment. We "need to" or are we? This comment may make more sense in the place where we spawn the task (service.rs IIRC).

We dont ask multiple sources, only the sender of the message.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -452,7 +447,22 @@ where B: BlockchainBackend + 'static
             );
             return Ok(());
         }
-
+        {
+            // we have this in mutex to ensure we dont have any funnies with rwlock with multiples requests first
+            // passing the check then all wanting to reconcile. We should only have 1 request passing
```
sdbondi: We use a mutex because we have multiple threads interrogating the hashset at the same time. The hashset is the important part. We allow _different_ blocks to be reconciled concurrently. 
```suggestion
            //  This handler is called in its own task for each new block message. The same block message may
            // be received from multiple peers as a normal part of block propagation. We use a HashSet to prevent 
            // the same block from being reconciled at the same time before it has a chance to be added to the blockchain.
```

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -461,7 +471,16 @@ where B: BlockchainBackend + 'static
...
         );
 
+        // We need to make sure that we can ask for the block from multiple sources as a single source can maliciously
+        // delay the block blocking us from getting to true latest tip
         let block = self.reconcile_block(source_peer.clone(), new_block).await?;
+
+        {
+            let mut lock = self.list_of_reconciling_blocks.lock().await;
+            let list_of_reconciling_blocks = lock.deref_mut();
+            list_of_reconciling_blocks.remove(&block_hash);
```
sdbondi: nit: deref_mut is implicitly called
```suggestion
            lock.remove(&block_hash);
```

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -452,7 +447,22 @@ where B: BlockchainBackend + 'static
...
+            if list_of_reconciling_blocks.contains(&block_hash) {
+                debug!(
+                    target: LOG_TARGET,
+                    "Block with hash `{}` is already being reconciled",
+                    block_hash.to_hex()
+                );
+                return Ok(());
+            } else {
+                list_of_reconciling_blocks.insert(block_hash);
+            }
```
sdbondi: nit: there is no need to explicitly dereference a lock guard (you may want to rename lock to list_of_reconciling_blocks for clarity)
```suggestion
          
            if list_of_reconciling_blocks.contains(&block_hash) {
                debug!(
                    target: LOG_TARGET,
                    "Block with hash `{}` is already being reconciled",
                    block_hash.to_hex()
                );
                return Ok(());
            } else {
                list_of_reconciling_blocks.insert(block_hash);
            }
```

sdbondi (commented):
An attacker may still delay the reconciliation if it is the first message, in the worst case making the base node switch to BlockSync mode. This may be fine as an attacker would need to be the first message in, and we assume they have no control over that.

The hash removal also needs to be done in the error case.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -452,6 +446,30 @@ where B: BlockchainBackend + 'static
...
+            if self.blockchain_db.block_exists(block_hash).await? {
+                debug!(
+                    target: LOG_TARGET,
+                    "Block with hash `{}` already stored",
+                    block_hash.to_hex()
+                );
+                return Ok(());
+            }
+            write_lock.insert(block_hash);
+        }
```
sdbondi: You can't split these. 

thread 1 wakes, read doesnt contain block X, sleeps
thread 2 wakes, read doesnt contain block X, writes hash
thread 1 wakes, writes hash and continues

These can happen in any order but only one order is valid.

sdbondi: You can also use `write_lock.insert(block_hash);` to determine if the entry already exists, it returns true if the entry exists

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -453,6 +447,40 @@ where B: BlockchainBackend + 'static
...
+            }
+            if write_lock.contains(&block_hash) {
+                debug!(
+                    target: LOG_TARGET,
+                    "Block with hash `{}` is already being reconciled",
+                    block_hash.to_hex()
+                );
+                return Ok(());
+            }
+            write_lock.insert(block_hash);
```
sdbondi:
```suggestion
            if !write_lock.insert(block_hash) {
                debug!(
                    target: LOG_TARGET,
                    "Block with hash `{}` is already being reconciled",
                    block_hash.to_hex()
                );
                return Ok(());
            }
```

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -461,7 +488,13 @@ where B: BlockchainBackend + 'static
             source_peer
         );
 
-        let block = self.reconcile_block(source_peer.clone(), new_block).await?;
+        let block_result = self.reconcile_block(source_peer.clone(), new_block).await;
+
+        {
```
sdbondi, line 493: If we allow the block before adding it, the block could reconcile again. Any reason not to unlock after handle_block?

SWvheerden: I want to say I think there was a reason, cant think of one now, so lets do that

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -461,9 +488,19 @@ where B: BlockchainBackend + 'static
...
+            let mut write_lock = self.list_of_reconciling_blocks.write().await;
+            write_lock.remove(&block_hash);
+        }
+        let block = block_result?;
+        let add_result = self.handle_block(block, Some(source_peer)).await;
+        {
+            let mut write_lock = self.list_of_reconciling_blocks.write().await;
+            write_lock.remove(&block_hash);
+        }
+        add_result?;
```
sdbondi: Thanks, that's better - not critical but it would be cleaner to put reconcile and handle in a new function (something like `try_reconcile_and_add_block`) and handle the error case and locking in this outer function. Then you can use `?` for those calls

SWvheerden: I agree here, its cleaner to move to a single function

sdbondi (approved):

PR #5583, opened 2023-07-10
feat: chat ffi status callback

brianp:
Description
---
This PR adds a callback manager and the ability to pass a callback into the chat ffi for use when a contact's status changes.

Motivation and Context
---
Requested by the mobile team.
Pattern similar to wallet ffi callbacks, although purposefully not copied, it just seemed to be the natural way to go 😄 

How Has This Been Tested?
---
Added integration test.
Although we get 16 callback calls. Not _totally_ unexpected but more than I was... expecting.

---
Read some code.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/integration_tests/src/chat_ffi.rs
+++ b/integration_tests/src/chat_ffi.rs
@@ -25,7 +25,7 @@ use std::{
     ffi::{c_void, CString},
     path::PathBuf,
     str::FromStr,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Once},
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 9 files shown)


integration_tests/src/chat_ffi.rs
```diff
@@ -180,11 +186,37 @@ pub async fn spawn_ffi_chat_client(name: &str, seed_peers: Vec<Peer>, base_dir:
     let out_error = Box::into_raw(Box::new(0));
 
     unsafe {
-        client_ptr = create_chat_client(config_ptr, identity_path_c_char, out_error);
+        //*ChatCallback::instance().contact_status_change.lock().unwrap() = 0;
```
SWvheerden:
```suggestion
```

SWvheerden (dismissed):
utAck

integration_tests/src/chat_ffi.rs
brianp: Good catch. Although it's actually supposed to be enabled 🫣
It acts like a reset as this *might* actually persist across tests.

PR #5586, opened 2023-07-12
chore(ci): update nightly

SWvheerden:
Description
---
Update the nightly version

Motivation and Context
---
one of the used dependencies uses a feature that only went stable in April, so we need to update the used version

How Has This Been Tested?
---
ran all tests

### DIFF
--- a/applications/tari_miner/src/stratum/controller.rs
+++ b/applications/tari_miner/src/stratum/controller.rs
@@ -243,6 +243,7 @@ impl Controller {
             // problem with template
             let _result = self.send_message_get_job_template();
         } else {
+            // unhandled
         }
     }
 
--- a/base_layer/common_types/src/transaction.rs
+++ b/base_layer/common_types/src/transaction.rs
@@ -12,7 +12,7 @@ use thiserror::Error;
 
 pub use crate::tx_id::TxId;
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum TransactionStatus {
     /// This transaction has been completed between the parties but has not been broadcast to the base layer network.
     Completed,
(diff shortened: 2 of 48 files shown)


applications/tari_miner/src/stratum/controller.rs
```diff
@@ -243,6 +243,7 @@ impl Controller {
             // problem with template
             let _result = self.send_message_get_job_template();
         } else {
+            // unhandled
```
brianp, line 246: It only kinda bugs me that the comments aren't consistent 🙃 

What's the need for them all of a sudden though, all they failing a linter now or something?

base_layer/common_types/src/transaction.rs
```diff
@@ -24,6 +24,7 @@ pub enum TransactionStatus {
     /// This transaction was generated as part of importing a spendable unblinded UTXO
     Imported,
     /// This transaction is still being negotiated by the parties
+    #[default]
```
brianp, line 27: Neat! TIL.

comms/core/src/transports/memory.rs
```diff
@@ -137,7 +137,7 @@ mod test {
 
     #[tokio::test]
     async fn simple_listen_and_dial() -> Result<(), ::std::io::Error> {
-        let t = MemoryTransport::default();
+        let t = MemoryTransport;
```
brianp, line 140: is this a new style / default? Feels foreign and anti-intuitive.

brianp (approved):
LGTM

Just to clarify the purpose of the PR is to bump the rust version, and that required syntax fixing for linters?

SWvheerden:
Updates to clippy is what caused most of the format changes

comms/core/src/transports/memory.rs
SWvheerden: Believe it is,

base_layer/common_types/src/transaction.rs
SWvheerden: yeah same

applications/tari_miner/src/stratum/controller.rs
SWvheerden: yip cant have an empty else { }
tried to make them as far as I could see comment why its empty

AaronFeickert: Aren't there macros intended to handle these cases?

SWvheerden: You can, but having an actual comment in there is more useful than a macro bypassing this check.

PR #5587, opened 2023-07-12
feat: add validator node consensus bls public key to registration request

jorgeantonio21:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


AaronFeickert:
If you want to use "original" BLS signatures without the risk of rogue-key attacks, there are two approaches:
1. Require the signer to provide the verifier with a separate BLS signature on the verification key, which functions as a proof of possession of the corresponding signing key. The verifier must have verified this signature prior to verifying any later aggregated BLS signatures using the verification key.
2. Require that any BLS signature used in an aggregation be on a distinct message. This can be done by prepending the verification key to the message prior to hashing, an operation which the verifier does as well.

The first design simplifies the computational requirements for aggregated signature verification by reducing the number of hashing operations used in the verification process, but at the cost of a proof of possession for each key (and, depending on consensus requirements, base-layer BLS support).

The second design simplifies the base-layer requirements by not requiring a proof of possession, but adds computational complexity to aggregated signature verification.

It might be useful to get some more specific benchmarks on these operations.

AaronFeickert:
I ran some [initial benchmarks](https://github.com/AaronFeickert/blst-benchmarks) to compare the designs. All tests arbitrarily assume 32 signers for aggregated signatures, include point decompression as part of the timing, and assume messages are 256 bytes.

First is the proof-of-possession design. Verifying all 32 signers' proofs of possession takes about 33.5 ms, which includes internal checks for key validity. Thereafter, verification of an aggregated signature takes about 1.0 ms. If you use batch verification on the proof-of-possession verification, the marginal time per proof drops to around 0.2 ms.

Next is the key-prefixing design. Verification of an aggregated signature, which includes internal key validity checks, takes about 4.8 ms. This could also be batched.

Using the minimal-signature variant reduces all these times.

brianp:
What's the status here? Mostly green, but has no description. Is this still a WIP, and can we mark it as such if it is.

PR #5590, opened 2023-07-12
fix: sorted edge case

SWvheerden:
Description
---
fixes and edge case where aggregated bodies that are sorted by the fact that they don't have more than 1 kernel/output/input

Motivation and Context
---
These blocks are sorted by default and we do not have to go and sort them.

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -452,6 +452,7 @@ mod test {
     #[test]
     fn test_sorted() {
         let mut body = AggregateBody::empty();
+        assert!(body.is_sorted());
         let kernel = TransactionKernel::new_current_version(
             KernelFeatures::default(),
             0.into(),
@@ -487,6 +488,8 @@ mod test {
         body.add_output(output.clone());
         assert!(body.is_sorted());
         assert!(!body.sorted);
+        body.sort();
+        assert!(body.sorted);
 
         let mut body2 = body.clone();
         body2.add_kernel(kernel);
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -369,7 +369,8 @@ impl AggregateBody {
     }
 
     pub fn is_sorted(&self) -> bool {
-        self.sorted
+        // a block containing only a single kernel, single output and single input is sorted by default
+        self.sorted || (self.kernels.len() <= 1 && self.outputs.len() <= 1 && self.inputs.len() <= 1)
     }
 
     /// Lists the number of inputs, outputs, and kernels in the block
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -451,3 +451,66 @@ impl Display for AggregateBody {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    use tari_common_types::types::{ComAndPubSignature, Commitment, FixedHash, PublicKey, Signature};
+    use tari_script::{ExecutionStack, TariScript};
+
+    use super::*;
+    use crate::{
+        covenants::Covenant,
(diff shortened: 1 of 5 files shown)


AaronFeickert:
Is it straightforward to add a unit test for this?

base_layer/core/src/transactions/aggregated_body.rs
AaronFeickert, line 1: It looks like `add_kernels` unsets the sorted flag, but `add_kernel` does not. Is this intentional? It isn't in line with the corresponding input and output functions.

SWvheerden: good catch, thats incorrect.

SWvheerden: Setting the flag on the get mut ref for kernels, outputs and inputs as well.

AaronFeickert (approved):

base_layer/core/src/transactions/aggregated_body.rs
AaronFeickert: > Setting the flag on the get mut ref for kernels, outputs and inputs as well.

If these are intended for testing only, can you make them private? Or will that make them inaccessible for testing?

hansieodendaal (approved):
utACK

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -105,7 +105,9 @@ impl AggregateBody {
     }
 
     /// Should be used for tests only. Get a mutable reference to the inputs
+    /// This will set the sorted flag to false
     pub fn inputs_mut(&mut self) -> &mut Vec<TransactionInput> {
```
stringhandler: These `*_mut` methods violate encapsulation and make maintaining an internal flag like `sorted` impossible. I would rather remove these methods and implement a better pattern

stringhandler:
Marking it as controversial just so that it gets a few more eyes and doesn't get automerged

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: yip agree there, its used in some integration tests, which means we cannot hide these behind `#cfg[test]`.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -450,3 +437,66 @@ impl Display for AggregateBody {
...
+
+    use super::*;
+    use crate::{
+        covenants::Covenant,
+        transactions::transaction_components::{EncryptedData, OutputFeatures, TransactionInputVersion},
+    };
+
+    #[test]
+    fn test_sorted() {
+        let mut body = AggregateBody::empty();
```
AaronFeickert, line 454: Might be useful to add a check that the empty case is sorted.

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -450,3 +437,66 @@ impl Display for AggregateBody {
...
+    use tari_script::{ExecutionStack, TariScript};
+
+    use super::*;
+    use crate::{
+        covenants::Covenant,
+        transactions::transaction_components::{EncryptedData, OutputFeatures, TransactionInputVersion},
+    };
+
+    #[test]
+    fn test_sorted() {
```
AaronFeickert, line 453: This doesn't check the behavior of `AggregatedBody::sort`. Would be useful to include this, since the test includes several assertions about the relationship between the result of `AggregatedBody::is_sorted` and the `sorted` property.

AaronFeickert (changes requested):

AaronFeickert (approved):

brianp (commented):
If sorted is beneficial why would we ever want them unsorted? Instead of there being a chance that they may or may not be sorted why don't we change the implementation to always sort, and not maintain a flag, because they should always be sorted?

SWvheerden:
Its an optimization, 
You don't want to do: add -> sort -> add - > sort....
you only want to sort once.

stringhandler (approved):

PR #5594, opened 2023-07-14
feat: chat message fetching pagination

brianp:
Description
---
Previously the only function to get messages was `get_all_messages` which return all related records in the db. Now we've added pagination and you can select the amount of results per page, and what page you'd like returned. Results are now ordered by most recent received.

Motivation and Context
---
`get_all_messages` has the potential to get very slow and ddos the client. Limiting the amount returned and allowing to change the requested amounts makes things better to deal with.

How Has This Been Tested?
---
Test added for pagination.

---
Tests pass.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/chat_ffi/chat.h
+++ b/base_layer/chat_ffi/chat.h
@@ -161,9 +161,11 @@ int check_online_status(struct ClientFFI *client, struct TariAddress *receiver, 
  * The ```address``` should be destroyed after use
  * The returned pointer to ```*mut ChatMessages``` should be destroyed after use
  */
-struct ChatMessages *get_all_messages(struct ClientFFI *client,
-                                      struct TariAddress *address,
-                                      int *error_out);
+struct ChatMessages *get_messages(struct ClientFFI *client,
+                                  struct TariAddress *address,
+                                  int *limit,
+                                  int *page,
+                                  int *error_out);
 
 /**
  * Frees memory for messages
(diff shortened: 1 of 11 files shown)


AaronFeickert:
I could see the following case occurring:
- The user requests the first page of messages and receives them.
- New messages arrive while the user is processing/reading/handling the first page.
- The user requests the second page of messages and receives them.
- The two pages overlap, but the user does not see the new messages.

This seems like it would make for irritating UX. How about adding an optional cursor that ensures that requests within the same "session" always maintain _consistent_ pagination? Then the user doesn't need to deduplicate messages. You could even add a simple function that informs the user if there exist messages newer than the cursor.

brianp:
> This seems like it would make for irritating UX.

This happens to me on strava constantly

> How about adding an optional cursor that ensures that requests within the same "session" always maintain consistent pagination? Then the user doesn't need to deduplicate messages. 

I'll run it by the mobile team and see what they say, it's pretty much entirely for them.

AaronFeickert:
> I'll run it by the mobile team and see what they say, it's pretty much entirely for them.

Could add it anyway, in case the feature is useful for others! The cursor could be entirely optional.

SWvheerden (approved):

SWvheerden:
Cursor could be nice, but for a MvP which the whole chat FFI is, I think this is fine for now.

PR #5596, opened 2023-07-20
feat: add blake2 atomic swap functionality

jorgeantonio21:
Description
---
Allows to swap Tari <> MinoTari.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/tari_console_wallet/src/cli.rs
```diff
@@ -129,7 +129,10 @@ pub enum CliCommands {
     ClearCustomBaseNode,
     InitShaAtomicSwap(SendTariArgs),
     FinaliseShaAtomicSwap(FinaliseShaAtomicSwapArgs),
-    ClaimShaAtomicSwapRefund(ClaimShaAtomicSwapRefundArgs),
+    ClaimShaAtomicSwapRefund(ClaimAtomicSwapRefundArgs),
+    InitBlake2AtomicSwap(SendTariArgs),
+    FinaliseBlake2AtomicSwap(FinaliseBlake2AtomicSwapArgs),
+    ClaimBlake2AtomicSwapRefund(ClaimAtomicSwapRefundArgs),
```
SWvheerden: Is there any reason why we dont change `ClaimShaAtomicSwapRefund` -> `ClaimAtomicSwapRefund` and  the same for the other 2 functions?
Then we just add an enum for the hash algo?

We are duplicating a lot of code just for a simple hash change?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -2240,6 +2261,121 @@ where
...
+                    amount,
+                    spending_key_id,
+                    output.features,
+                    output.script,
+                    inputs!(pre_image),
+                    self.resources.wallet_identity.wallet_node_key_id.clone(),
+                    output.sender_offset_public_key,
+                    output.metadata_signature,
+                    // Although the technically the script does have a script lock higher than 0, this does not apply
+                    // to to us as we are claiming the Hashed part which has a 0 time lock
```
```suggestion
                    // to us as we are claiming the Hashed part which has a 0 time lock
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1273,6 +1293,233 @@ where
...
+        let pre_image = PublicKey::from_secret_key(&PrivateKey::random(&mut OsRng));
+        let hash: [u8; 32] = Blake256::digest(pre_image.as_bytes()).into();
+
+        // let's check that the timelock corresponds to some time in the future
+        let tip_height = self.last_seen_tip_height.unwrap_or(0);
+        if timelock <= tip_height {
+            return Err(TransactionServiceError::Blake2AtomicSwapTransactionError(format!(
+                "Invalid timelock, must be a value in the future"
+            )));
+        }
```
hansieodendaal: Maybe it is better to treat `timelock` as a delta so this can never fail:
- `0` means there is no timelock
- Any 'val > 0' means it is added to the `last_seen_tip_height`

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1273,6 +1293,233 @@ where
...
+            Else
+                CheckHeightVerify(timelock) PushPubKey(Box::new(self.resources.wallet_identity.node_identity.public_key().clone()))
+            EndIf
+        );
+
+        // Empty covenant
+        let covenant = Covenant::default();
+
+        // Default range proof
+        let minimum_value_promise = MicroTari::zero();
```
hansieodendaal: I would argue that these comments add no value and the comment above `minimum_value_promise` is not accurate
```suggestion
        let covenant = Covenant::default();
        let minimum_value_promise = MicroTari::zero();
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1273,6 +1293,233 @@ where
...
+            .get_public_key_at_key_id(&sender_offset_private_key)
+            .await?;
+
+        let spending_key_id = self
+            .resources
+            .transaction_key_manager_service
+            .import_key(spending_key)
+            .await?;
+
+        let minimum_value_promise = MicroTari::zero();
```
hansieodendaal: This looks like a repeat
```suggestion
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1273,6 +1293,233 @@ where
...
+            .map_err(|e| {
+                error!(
+                    target: LOG_TARGET,
+                    "Transaction (TxId: {}) could not be finalized. Failure error: {:?}", tx_id, e,
+                );
+                TransactionServiceProtocolError::new(tx_id, e.into())
+            })?;
+        info!(target: LOG_TARGET, "Finalized one-side transaction TxId: {}", tx_id);
+
+        // This event being setn is important, but not critical to the protocol being successful. Send only fails if
```
hansieodendaal:
```suggestion
        // This event being sent is important, but not critical to the protocol being successful. Send only fails if
```

integration_tests/tests/features/WalletTransfer.feature
```diff
@@ -90,3 +90,21 @@ Feature: Wallet Transfer
...
+    When I have mining node MINER connected to base node NODE and wallet WALLET_A
+    When I have mining node MINER_2 connected to base node NODE and wallet WALLET_C
+    When mining node MINER mines 10 blocks
+    Then I wait for wallet WALLET_A to have at least 10000000000 uT
+    When I broadcast a blake2 HTLC transaction with 5000000000 uT from wallet WALLET_A to wallet WALLET_B with timelock 12 at fee 20
+    When mining node MINER_2 mines 16 blocks
+    When I wait 5 seconds
+    When I claim an HTLC refund transaction with wallet WALLET_A at fee 20
+    When mining node MINER_2 mines 6 blocks
+    Then I wait for wallet WALLET_A to have at least 9000000000 uT
```
hansieodendaal: The numbers here seem off. I know the _Scenario: As a wallet I want to claim a HTLC refund transaction_ looks exactly the same, but that one is also wrong (please fix it as well).

If we mine 10 blocks we will have way more than 10 000 000 000 uT, so to test eventually after the refund has been claimed if we have at least 9 000 000 000 uT is a non-test. The 1st balance enquiry tells us we have enough funds for the _blake2 HTLC transaction_. After that transaction we should actually know the correct balance, or close to it, for a valid assert at the end. My suggestion:
- _When mining node MINER mines 10 blocks_
- **_Then I wait for wallet WALLET_A to have at least `BALANCE_A` uT_** (`BALANCE_A` must be a few Tari less than what is available to allow for fees and some change)
- _When I broadcast a blake2 HTLC transaction with `BALANCE_A` uT from wallet WALLET_A to wallet WALLET_B with timelock 12 at fee 20_
- _When mining node MINER_2 mines 16 blocks_
- **_I wait for wallet WALLET_A to have less than `BALANCE_A` uT_**
- ~_ When I wait 5 seconds_~
- _When I claim an HTLC refund transaction with wallet WALLET_A at fee 20_
- _When mining node MINER_2 mines 6 blocks_
- **_Then I wait for wallet WALLET_A to have at least `BALANCE_A` uT_**

hansieodendaal (commented):
Nice.

Please see some suggestions below.

applications/tari_console_wallet/src/cli.rs
jorgeantonio21: Yeah, I wanted to do this. But, I didn't want to change the already existing interfaces without further comments.

jorgeantonio21: Since, we currently have two functions for sending and claiming atomic swap txs, we need to distinguish these in the enum. A possibility is to integrate everything into a single function, with a hasher as a generic parameter. Not sure, what is preferable.

base_layer/wallet/src/transaction_service/service.rs
jorgeantonio21: Great !

base_layer/wallet/src/transaction_service/service.rs
jorgeantonio21: Great !

applications/tari_console_wallet/src/cli.rs
SWvheerden: I think an enum would be fine. 
and in code we do something like this
```
script = match enum{
sha => {sha_script}
blake =>{ blake_script}
}
```

PR #5597, opened 2023-07-20
chore: rename tari > minotari

SWvheerden:
Description
---
This renames all user interface names from tari to minotari

Motivation and Context
---
This is inline with the name changes of Dan -> Tari, and BaseLayer -> MinoTari

How Has This Been Tested?
---
All tests
Manual inspection of log files and binaries.


stringhandler (commented):
Happy in principle.

1. I believe the capitalization should be Minotari and not MinoTari
2. I think we should take this opportunity to rename minotari_base_node to just minotari_node

PR #5601, opened 2023-07-28
feat: remove orphan validation and only validate on insertion

SWvheerden:
Description
---
Currently, block validation is done in two steps orphan check and full check. 
An orphan check is done when a block is first inserted into the DB (all blocks are). But before this is done, the block must pass an internal integrity check (checking signatures, rangeproofs, etc)
The blocks are then stored in the orphan pool. 
When they are added to the main chain, a full check is done on the blocks to ensure the mmr's are correct, no double spends  + another integrity check. 

Bad blocks are also only tracked when syncing. 

This PR removes the orphan check and only does validation on the block when we add it to the main chain. But to ensure a node is not spammed by useless blocks, it makes sure the block received at least has a min difficulty. The orphan pool will clean out the oldest blocks and this stored amount is configurable via the config. 

As soon as we can track a block to the main chain, we will do a full Proof of work check. 

Then as when a chain/block proof of work is higher than the current chain the blocks will be added to the main chain. This is then when we do the full validation of the block. 

Changed up how the bad blocks are handled, and we now properly keep track of all bad blocks. When a block fails any part of the validation it will be added to the bad_block list. When we receive a new request for a block, this will be one of the first things we check, before we spent resources validation it again. 

Motivation and Context
---
TARI-0003
A base node waste resources validating a block more than once. From testing, the most expensive tests to run are(in order): Rangeproofs, Metadatasignature, Scriptsignature, kernel signature. These tests are all run on orphan blocks. Its very easy to put junk data in these blocks and spam a node to keep wasting resources and validated these fake blocks. 

By using the tests added in #5599, this PR improves `add_block` time by almost 100%. 

How Has This Been Tested?
---
Unit tests.

---
Checking that we have not changed or removed any required validation rules.


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -453,6 +463,40 @@ where B: BlockchainBackend + 'static
...
+                .add_bad_block(
+                    new_block.header.hash(),
+                    self.blockchain_db.get_chain_metadata().await?.height_of_longest_chain(),
+                )
+                .await?;
+            return Err(CommsInterfaceError::InvalidBlockHeader(
+                BlockHeaderValidationError::ProofOfWorkError(PowError::AchievedDifficultyBelowMin),
+            ));
+        }
+
```
hansieodendaal: I think this should be a proper difficulty check with `DifficultyCalculator::check_achieved_and_target_difficulty`, as it is too easy to mine a block with minimum difficulty.

If this check has passed for a specific block we do not need to check it again when we do full validation.

hansieodendaal (commented):
Please see comment below.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
SWvheerden: The problem with this is, we do not know what the actual target is at this point. The target difficulty is only available when we have a link to the main chain, and when we do, this is the first thing we check.  The blocks here might just be an orphan block. 

This is here to force the user to spend some time mining it. By making this take longer to create than validate, you make any attack here very expensive.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -453,6 +463,40 @@ where B: BlockchainBackend + 'static
...
+            self.blockchain_db
+                .add_bad_block(
+                    new_block.header.hash(),
+                    self.blockchain_db.get_chain_metadata().await?.height_of_longest_chain(),
+                )
+                .await?;
+            return Err(CommsInterfaceError::InvalidBlockHeader(
+                BlockHeaderValidationError::ProofOfWorkError(PowError::AchievedDifficultyBelowMin),
+            ));
+        }
```
sdbondi: Let's put this in a function e.g. `self.check_minimum_difficulty(&new_block).await?;`

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
hansieodendaal: Ok, let us make an issue to up the minimum difficulty for mainnet to be achievable when using 4x threads for mining on something like an _AMD Ryzen 7 2700X_. To add some context, for `esmeralda` when using a single thread on an _AMD Ryzen 7 2700X_ when that is the only hash rate the target difficulty is ~ 2x minimum difficulty.

hansieodendaal (approved):
utACK

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -498,6 +522,30 @@ where B: BlockchainBackend + 'static
         Ok(())
     }
 
+    async fn is_block_valid_and_exists(&self, block: FixedHash) -> Result<bool, CommsInterfaceError> {
```
sdbondi: nit: I'd probably separate validation and existence. Do the `block_exists` check in `handle_new_block_message` and then call `self.check_block_validity()?;` (or another name) and do the bad block and the min difficulty check

base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -118,17 +123,30 @@ impl<B: BlockchainBackend + 'static> BlockHeaderSyncValidator<B> {
...
+        if let Err(e) = &result {
+            match e {
+                // future timelimit validation can succeed at a later time. As the block is not yet valid, we discard it
+                // for now and ban the peer, but wont blacklist the block.
+                BlockHeaderError(BlockHeaderValidationError::InvalidTimestampFutureTimeLimit) => {},
+                _ => {
+                    let mut txn = self.db.write_transaction();
+                    txn.insert_bad_block(header.hash(), header.height);
+                    txn.commit().await?;
+                },
```
sdbondi: Think we should also handle storage failures `FatalStorageError`. In this case, the block shouldn't be marked as bad and the peer shouldn't be banned (this needs to be added in `synchonizer.rs`). There might be some other cases (`CustomError`?)

Risk: if the message was somehow able to cause a FatalStorageError due to a bug, then they could cause a good block to be marked as bad.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2172,7 +2183,20 @@ fn insert_orphan_and_find_new_tips<T: BlockchainBackend>(
...
         prev_timestamps.push(timestamp);
     }
-    validator.validate(db, &block.header, parent.header(), &prev_timestamps, None)?;
+    let result = validator.validate(db, &block.header, parent.header(), &prev_timestamps, None);
+    if let Err(e) = &result {
+        match e {
+            // future timelimit validation can succeed at a later time. As the block is not yet valid, we discard it
+            // for now and ban the peer, but wont blacklist the block.
+            BlockHeaderError(BlockHeaderValidationError::InvalidTimestampFutureTimeLimit) => {},
+            _ => {
```
sdbondi: Same here, handling of non-validation errors.

base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -27,12 +27,17 @@ use tari_utilities::{epoch_time::EpochTime, hex::Hex};
...
     chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend, ChainStorageError, TargetDifficulties},
     common::rolling_vec::RollingVec,
     consensus::ConsensusManager,
     proof_of_work::{randomx_factory::RandomXFactory, PowAlgorithm},
-    validation::{header::HeaderFullValidator, DifficultyCalculator, HeaderChainLinkedValidator},
+    validation::{
+        header::HeaderFullValidator,
+        DifficultyCalculator,
+        HeaderChainLinkedValidator,
+        ValidationError::BlockHeaderError,
```
sdbondi: nit: don't import enum variants at the module level

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2172,7 +2183,20 @@ fn insert_orphan_and_find_new_tips<T: BlockchainBackend>(
...
+                let mut txn = DbTransaction::new();
+                txn.insert_bad_block(block.header.hash(), block.header.height);
+                db.write(txn)?;
+            },
+        }
+    }
+    result?;
     let achieved_target_diff = difficulty_calculator.check_achieved_and_target_difficulty(db, &block.header)?;
```
sdbondi: nit: match on the result rather than `result?` which desugars into another match. It also makes it clearer IMO that we are still exiting the function early if FTL fails. 
Note that the validator returns the `achieved_target_diff` so it doesn't have to be done again below

```suggestion
    let achieved_target_diff = match result {
        Ok(achieved_diff) => achieved_diff,
        Err(e @ValidationError::BlockHeaderError(BlockHeaderValidationError::InvalidTimestampFutureTimeLimit)) => return Err(e.into()),
        Err(e @ValidationError::FatalStorageError(_)) => return Err(e.into()),
        Err(e) => {
            let mut txn = DbTransaction::new();
            txn.insert_bad_block(block.header.hash(), block.header.height);
            db.write(txn)?; 
            return Err(e.into());
        }
    };
```

sdbondi: Interestingly, if I add this code below, we get a mismatch in achieved difficulty in `it_correctly_handles_duplicate_blocks` test
```rust
    let d2 = difficulty_calculator.check_achieved_and_target_difficulty(db, &block.header)?;
    assert_eq!(achieved_target_diff.achieved(), d2.achieved());
```
 EDIT: ah it's a symptom of us using a mock validator in tests or use the real validator 
We'll need to change the mock to return valid difficulty
```rust
impl<B: BlockchainBackend> HeaderChainLinkedValidator<B> for MockValidator {
    fn validate(
        &self,
        db: &B,
        header: &BlockHeader,
        _: &BlockHeader,
        _: &[EpochTime],
        _: Option<Difficulty>,
    ) -> Result<AchievedTargetDifficulty, ValidationError> {
        if self.is_valid.load(Ordering::SeqCst) {
            let consensus = create_consensus_rules();
            let difficulty_calculator = DifficultyCalculator::new(consensus, Default::default());
            difficulty_calculator.check_achieved_and_target_difficulty(db, header)
        } else {
            Err(ValidationError::custom_error(
                "This mock validator always returns an error",
            ))
        }
    }
}
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -913,17 +916,16 @@ where B: BlockchainBackend
...
-            return Err(e.into());
+        if db.bad_block_exists(block_hash)? {
+            return Err(ChainStorageError::ValidationError {
+                source: ValidationError::BadBlockFound {
+                    hash: block_hash.to_hex(),
+                },
+            });
         }
+        // the only fast check we can perform that is slightly expensive to fake is a min difficulty check, this is done
+        // as soon as we receive the block before we do any processing on it. A proper proof of work is done as soon as
```
sdbondi, line 926:
```suggestion
        // as soon as we receive the block before we do any processing on it. Full block and header validation is done as soon as
```

sdbondi (commented):
Looks good, I think we need to exclude database errors and alike from causing a block to be marked as bad. Few other minor code clarity comments.

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -438,15 +448,33 @@ where B: BlockchainBackend + 'static
         }
 
         // Lets check if the block exists before we try and ask for a complete block
-        if self.blockchain_db.block_exists(block_hash).await? {
-            debug!(
-                target: LOG_TARGET,
-                "Block with hash `{}` already stored",
-                block_hash.to_hex()
-            );
+        if self.is_block_valid_and_exists(block_hash).await? {
```
brianp: I find this function name confusing because it doesn't actually check the validity of the block, which is what the PR is deferring in the first place. It sees if we already know about it, as a good or bad block, but not if the block is valid.

base_layer/core/tests/tests/node_comms_interface.rs
```diff
@@ -72,7 +72,7 @@ async fn inbound_get_metadata() {
     let (request_sender, _) = reply_channel::unbounded();
     let (block_sender, _) = mpsc::unbounded_channel();
     let outbound_nci = OutboundNodeCommsInterface::new(request_sender, block_sender.clone());
-
+    let randomx_factory = RandomXFactory::new(2);
```
brianp, line 75: Magic number 2?

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
sdbondi: I missed that this is called again below - we dont want to do the min difficulty check later. I rescind the nit :)

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: thats exactly not what I want to say. I want to say we first a just a pow check, 
the full block validation only happens later when we want to add the block to the main chain

base_layer/core/tests/tests/node_comms_interface.rs
SWvheerden: amount of vms that may be used

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: >  Full block and header validation is done as soon as we add it (previously: "can link it") to the main chain.

Is this not what the comment is saying?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -913,17 +916,16 @@ where B: BlockchainBackend
         if db.contains(&DbKey::BlockHash(block_hash))? {
             return Ok(BlockAddResult::BlockExists);
         }
-        // Perform orphan block validation.
-        if let Err(e) = self.validators.orphan.validate_internal_consistency(&block) {
```
sdbondi, line 917: Not sure if this trait and function are still used - perhaps should be removed in this PR

sdbondi: Checked that these validations are still done in the BlockBodyFullValidator::validate function, called in reorganize_chain when adding to the chain.

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: I expanded the comment some what. 
Is it now more clear?

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: Discussed offline, we are going to clean this up in another PR

base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -118,15 +118,38 @@ impl<B: BlockchainBackend + 'static> BlockHeaderSyncValidator<B> {
...
+            Ok(achieved_target) => achieved_target,
+            // future timelimit validation can succeed at a later time. As the block is not yet valid, we discard it
+            // for now and ban the peer, but wont blacklist the block.
+            Err(e @ ValidationError::BlockHeaderError(BlockHeaderValidationError::InvalidTimestampFutureTimeLimit)) => {
+                return Err(e.into())
+            },
+            // We dont want to mark a block as bad for internal failures
+            Err(
+                e @ ValidationError::FatalStorageError(_) |
+                e @ ValidationError::NotEnoughTimestamps { .. } |
```
sdbondi, line 141: confirmed: NotEnoughTimestamps only occurs when there is a bug in the code

sdbondi (approved):

PR #5603, opened 2023-08-01
fix: monero fork attack

SWvheerden:
Description
---
Forces the nonce of all Monero blocks to be 0. 

Motivation and Context
---
The Nonce field in the header is used when calculating the hash of the block. When merge mining monero blocks, the nonce used by the monero mining sits in pow->monero_header->nonce. It does not use the tari block header nonce. 

But this allows any node to change the nonce in the header and not break the mined pow, but change the hash of monero header. This will cause the network to fork of valid mined monero blocks. 

How Has This Been Tested?
---
unit tests

RFC update: https://github.com/tari-project/rfcs/pull/104


hansieodendaal (approved):
utACK

PR #5604, opened 2023-08-02
fix: duplicate tari header in monero coinbase

SWvheerden:
Description
---
Adds prefix to merge mining hash to detect duplicate tari block headers in monero coinbases.

Motivation and Context
---
Miners should not be able to mine multiple forks of tari at the same time with no cost. Addreses : TARI-005

How Has This Been Tested?
---
Unit tests

Audit Finding Number
---
TARI-005

### DIFF
--- a/base_layer/core/src/blocks/block_header.rs
+++ b/base_layer/core/src/blocks/block_header.rs
@@ -232,6 +232,15 @@ impl BlockHeader {
             .into()
     }
 
+    pub fn merge_mining_hash(&self) -> Vec<u8> {
+        let mining_prefix = (b"TARI").to_vec();
+        let mut mining_hash = self.mining_hash().to_vec();
+        for i in 0..4 {
+            mining_hash[i] = mining_prefix[i];
+        }
+        mining_hash
+    }
+
     #[inline]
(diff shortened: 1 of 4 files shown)


base_layer/core/src/blocks/block_header.rs
```diff
@@ -230,6 +230,15 @@ impl BlockHeader {
...
 
+    pub fn merge_mining_hash(&self) -> Vec<u8> {
+        let mining_prefix = (b"TARI").to_vec();
+        let mut mining_hash = self.mining_hash().to_vec();
+        for i in 0..4 {
+            mining_hash[i] = mining_prefix[i];
+        }
+        mining_hash
+    }
+
```
sdbondi: You can change the FixedHash byte content without allocations
```suggestion
    pub fn merge_mining_hash(&self) -> FixedHash {
     let mut mining_hash = self.mining_hash();
        mining_hash[0..4].copy_from_slice(b"TARI"); // Maybe put this in a `const`
        mining_hash
    }
```

sdbondi (commented):
Will think a bit more about the increase in chances of hash collision and if there's an exploit but otherwise LGTM. A minor suggestion to avoid some allocations.

PR #5607, opened 2023-08-03
ci: terminate slow tests

stringhandler:
Terminate slow running tests.


.config/nextest.toml
```diff
@@ -1,2 +1,5 @@
+[profile.ci]
+slow-timeout = { period = "20s", terminate-after=2}
```
SWvheerden:
```suggestion
slow-timeout = { period = "60s", terminate-after=2}
```

PR #5608, opened 2023-08-03
feat: ban peer unexpected response

SWvheerden:
Description
---
ban a peer if they sent an unexpected message

Motivation and Context
---
audit finding, TARI-002


sdbondi (approved):

brianp (commented):
Ack

I looked into the test failure. Just looks like a flaky test. Passes locally for me.

I do wonder if we should ban it so quickly though. If nodes are upgrading and a new message is being sent around old nodes could ignore it and run fine, but if they don't recognize it now they could ban perfectly good nodes. Maybe we need to have some kind of limiter, like repeated bad messages, or versioning of some kind?

PR #5611, opened 2023-08-04
fix(comms): check multiple addresses for inbound liveness check

sdbondi:
Description
---
fix(comms): check multiple addresses for inbound liveness check

Motivation and Context
---
Liveness check uses the first address only, when setting a node up for tcp when a tor address existed it breaks (🔌). It now checks all addresses until one of them works.

How Has This Been Tested?
---
Manually

---
Run a node with multiple addresses and check that the ⚡️ symbol still shows.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (dismissed):
utACK

brianp (commented):
utAck

The form to loop through the addresses with the incrementing `address_idx` took a minute to grock. Could we instead loop through the `Vec<Multiaddr>` more directly? 

`for address in addresses.iter()` ?

sdbondi:
> utAck
> 
> The form to loop through the addresses with the incrementing `address_idx` took a minute to grock. Could we instead loop through the `Vec<Multiaddr>` more directly?
> 
> `for address in addresses.iter()` ?

We aren't looping through the addresses really. We're looping to continuously dial an address until one works, and if we lose the connection, trying again. The address_idx is the index to the current address we're checking. I removed the last_working_address variable because it's unneccessary.

SWvheerden (approved):

PR #5612, opened 2023-08-04
chore: fixes monero build

SWvheerden:
Description
---
Fixes the monero-rs build by including the correct feature flag name 

How Has This Been Tested?
---
It builds

---
Delete the cargo.lock file, and monero-rs will not build anymore.


hansieodendaal (approved):
ACK

PR #5615, opened 2023-08-07
fix!: add validator mr to mining hash

SWvheerden:
Description
---
add validator mr to mining hash
Re-ordered structs/hashes to match each other

Motivation and Context
---
This field is not covered by mining, making it mutable and changeable by any node. 
Reordering was done to make auditing easier now and in future so that the two match each other

Audit Finding Number
---
TARI-012


sdbondi (approved):

brianp (approved):
utAck LGTM

PR #5617, opened 2023-08-07
fix: minor fixes for multiple address support

sdbondi:
Description
---
Fixed a few places in the code where multiple addresses were not completely supported, namely
- wallet base node detection via GRPC, and 
- whoami QR code

Fixed incorrect format for QR Code

Replaces replace_public_address with set_public_addresses.

Motivation and Context
---
Minor fixes to use all addresses rather than a single address.

With set_public_addresses, I wanted to set the addresses to only the public addresses. However, this is problematic because the tor address is added at a very late point in the comms initialization. This would mean that if public addresses were configured along with a late-configured tor address, the identity sig would change every time.

Other than in the comms builder, the tor hidden service code is completely decoupled from comms (that is, we could pass in the transport to comms and not the hidden service abstraction itself). In order to solve this we would setup the tor HS earlier before we initialize the node identity, and remove the code that mutates the node identity in the comms builder. We'd know the final tor address and would be able to add it to the public address list and have a final address set for node identity. Although I anticipate this refactor would be fairly straightforward, is not strictly necessary at this point so is not attempted. 

How Has This Been Tested?
---
Cucumbers

---
Add multiple addresses to the base node and check that the wallet is aware of those addresses.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
LGTM utAck

I checked out the test failure. Works locally for me, likely just an intermittent flaky test.

SWvheerden (approved):

PR #5619, opened 2023-08-08
feat!: add claim_public_key to sidechain features

sdbondi:
Description
---
Add claim public key to validator node registration

Motivation and Context
---
This claim key is informational and will be used in L2 to restrict fee claims to a particular signing key 

How Has This Been Tested?
---
Existing tests

---
Run register validator node wallet command

Breaking Changes
---

- [ ] None
- [x] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
LGTM utAck

I'm wondering how much stuff is being duplicated between the layers and if maybe things like constructing the challenge should come from an L2 lib, included into L1?

sdbondi:
Yup agree, but the other way round - L2 depends on tari base libs in this repo. In this case it's ok because we reuse the ValidatorNodeSignature abstraction in L2. I would like a library of shared signature types and hash domains, mainly to avoid depending on the whole tari_core crate just for 1/2 hash domains.

stringhandler (approved):

PR #5624, opened 2023-08-10
fix(core): always pass the correct timestamp window to header validatior

sdbondi:
Description
---
fix(core): sanity check for timestamp window length checks for the correct length of the timestamp window
fix(core): always include the correct timestamp window to header validation
tests(core): add a reorg test of greater size than median timestamp window size
chore: remove allocations when logging hashes in blockchain database  

Motivation and Context
---
`sanity_check_timestamp_count` (previously `check_timestamp_count`) allowed timestamp windows of  `min(median_window_size, header.height - 1)` or greater. However, the window size must always be the correct size for a given block. If it is not, the median timestamp calculation will be incorrect. This check also subtracted 1 from the header height BEFORE checking the height is correct. This would cause a panic/underflow if a header of height 0 is passed in.

In the reorg logic, we passed in the same timestamp window for the candidate block when validating orphan chains, this could cause correct reorgs to fail. This PR also ensures that timestamps are sorted as this is required for a correct median calculation. 

How Has This Been Tested?
---
Adds a new unit test `it_does_a_sanity_check_on_the_number_of_timestamps_provided`
Adds block chain unit test `it_links_many_orphan_branches_to_main_chain_with_greater_reorg_than_median_timestamp_window`
Existing test `it_links_many_orphan_branches_to_main_chain` failed after the fix to `check_timestamp_count`. Fixed in this PR
Manually caused a about 90 block reorg on localnet nodes

---
Create a reorg using localnet

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/validation/mocks.rs
```diff
@@ -108,22 +109,17 @@ impl InternalConsistencyValidator for MockValidator {
...
-            let achieved = sha3x_difficulty(header)?;
-
-            let achieved_target = AchievedTargetDifficulty::try_construct(
-                PowAlgorithm::Sha3x,
-                achieved,
-                achieved.checked_add(1).unwrap(),
-            )
-            .unwrap();
-            Ok(achieved_target)
+            // TODO: this assumes consensus rules are the same as the test rules which is a little brittle
```
SWvheerden:
```suggestion
            // this assumes consensus rules are the same as the test rules which is a little brittle
```

SWvheerden: I think its a valid assumption. Where this is not the case, we can update the test to reflect this

sdbondi: yeah, it'll probably never be an issue in practice but if you imagine a future developer who for some reason needs to change consensus rules in the tests and struggles to find why their difficulty doesn't make sense, this comment could help them.

base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -89,20 +89,24 @@ impl<B: BlockchainBackend> HeaderChainLinkedValidator<B> for HeaderFullValidator
...
+    if timestamps.is_empty() {
+        return Err(ValidationError::IncorrectNumberOfTimestampsProvided {
+            expected: expected_timestamp_count,
+            actual: 0,
+        });
+    }
+
+    if timestamps.len() as u64 != expected_timestamp_count {
+        return Err(ValidationError::IncorrectNumberOfTimestampsProvided {
+            actual: timestamps.len() as u64,
```
brianp, line 109: This code _looks_ fine.

but tbh I don't really understand it just yet. It looks like we're receiving a slice of timestamps and seeing if the number of timestamps is the same as the median timestamp count. But this slice could contain the same single timestamp repeated and this sanity check passes. 
So what I don't yet understand is why we're using this method of counting timestamps as a validation.

base_layer/core/src/validation/mocks.rs
sdbondi: Will remove the TODO

brianp (commented):
Code looks functional, but I what I don't understand reasoning behind the original, and new implementation. So I don't know if the code is _correct_. 😅

base_layer/core/src/validation/header/header_full_validator.rs
sdbondi: Yup the blockchain db code is super hard to reason about. It's not strictly a header validation since the timestamps come from headers previously inserted into the main/orphan chains so the contract we're making in the `HeaderChainLinkedValidator` trait is that the caller must provide the correct timestamps. This is why I renamed it to sanity_check because it's more about asserting an invariant we require when calling the validator. The other option is to take it out completely or to panic, but I went with the least risky approach.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2251,27 +2222,41 @@ fn find_orphan_descendant_tips_of<T: BlockchainBackend>(
...
             child.header.height,
-            child.hash().to_hex(),
+            child.hash(),
             prev_chain_header.height(),
-            prev_chain_header.hash().to_hex()
+            prev_chain_header.hash(),
         );
+        // Append the child timestamp - a RollingVec ensures that the number of timestamps can never be more than the
+        // median timestamp window size.
+        current_branch_prev_timestamps.push(child.header.timestamp);
```
SWvheerden: this should only happen after validation, not before

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2285,26 +2270,31 @@ fn find_orphan_descendant_tips_of<T: BlockchainBackend>(
...
+                let children = find_orphan_descendant_tips_of(
+                    db,
+                    chain_header,
+                    current_branch_prev_timestamps.clone(),
+                    validator,
+                )?;
                 res.extend(children);
+
+                // Restore the previous timestamps for this level of branching
+                current_branch_prev_timestamps = prev_timestamps.clone();
```
SWvheerden: why are you resetting completely here?

sdbondi: the function is recursive, so it's only resetting to the 'level' - but anyhow, you're correct about not including the timestamp from the current block which removes this line :) I've pushed the changes

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: thanks you are correct :)

SWvheerden (dismissed):

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2836,6 +2811,98 @@ mod test {
...
+            let result = test.handle_possible_reorg(block.to_arc_block()).unwrap();
+            result.assert_reorg(11, 12);
+
+            {
+                // Check 2b was added
+                let access = test.db_write_access();
+                let block = orphan_chain_b.get("2b").unwrap().clone();
+                assert!(access.contains(&DbKey::BlockHash(*block.hash())).unwrap());
+
+                // Check 7d is the tip
```
```suggestion
                // Check 12b is the tip
```

hansieodendaal (approved):
Definite improvement
One comment nit

utACK

PR #5636, opened 2023-08-13
feat!: update key parsing

AaronFeickert:
Description
---
Updates key parsing to differentiate between canonical- and reduction-based use cases. Makes several updates to account for `EpochTime` changes. Updates genesis data.

Motivation and Context
---
An [update](https://github.com/tari-project/tari-crypto/pull/194) to `tari-crypto` differentiates secret key parsing between cases that should use canonical byte arrays, and those that use random byte arrays requiring modular wide reduction.

This PR makes corresponding changes to account for this.

It also makes several updates required since the `tari_utilities` release used here also makes breaking `EpochTime` changes that can't be cleanly done separately.

How Has This Been Tested?
---
Existing tests pass. Uses of parsed secret keys have been manually inspected for correctness.

---
Assert that secret keys and scalars instantiated from byte arrays use the appropriate approach: parsing from canonical byte arrays, or modular wide reduction from uniform byte arrays.

Confirm that `EpochTime` changes reflect the intended logic, especially relating to timestamp and difficulty handling.

BREAKING CHANGE: This modifies how secret keys and scalars are constructed, and is therefore a breaking change.


AaronFeickert:
This PR currently references a `tari-crypto` fork for testing purposes. It should not be merged until:
- it includes all necessary changes
- all tests pass
- the linked `tari-crypto` PR is merged with a corresponding release/tag
- all `tari-crypto` dependencies reference this release/tag

stringhandler (dismissed):
happy with this so far

AaronFeickert:
This is being marked as ready to review. However, it must NOT BE MERGED until the `tari-crypto` fork dependency is switched.

AaronFeickert:
Special attention should also be paid to the opcode and script stack changes. It was not clear what the intended handling was for providing verification message data to the relevant opcodes, nor how array data should be parsed to scalars for computation of Ristretto points in scripts.

infrastructure/tari_script/src/script.rs
```diff
@@ -637,11 +640,11 @@ impl TariScript {
     fn handle_to_ristretto_point(&self, stack: &mut ExecutionStack) -> Result<(), ScriptError> {
         let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
         let scalar = match &item {
-            StackItem::Hash(hash) => hash.as_slice(),
+            StackItem::Hash(hash) => hash.as_slice(), // TODO: confirm this isn't problematic
```
sdbondi: I'm not too sure what makes sense here. For context, is conversion from a 256-bit hash using `from_bytes_mod_order` problematic? I assume that hashes (Blake Sha3 etc.) have a good chance of not being in the canonical form? 

Perhaps we would need to provide a 512-bit hash type and either add a new ToScalar op for reduction to canonical scalar bytes and then conversion to a Ristretto point, or only support scalar and 512-bit hash input for ToRistrettoPoint.

FWIW I can't find an instance of Hash to Ristretto point conversion being used in the Tari codebase.

AaronFeickert:
This PR now encompasses changes required from a `tari_utilities` [PR](https://github.com/tari-project/tari_utilities/pull/61) that removes certain arithmetic operations from `EpochTime`. Because these changes are included in the same release as the changes to `ByteArray`, it's not possible to do separate PRs that will pass testing.

Extra care should be taken to review the `EpochTime` changes in this PR, since they are used in timestamp and difficulty calculations.

infrastructure/tari_script/src/script.rs
AaronFeickert: There are no good use cases I can think of for doing reduction of 32-byte hash output to produce a scalar. If a byte slice was produced from an existing scalar (like in the case of serialization), then it should be parsed canonically. If it was produced as uniform byte output from a hash function, it should undergo wide reduction from 64 bytes in order to mitigate reduction bias.

An earlier discussion mentioned that cross-group scalar equality could be an intended use case for this kind of functionality. This would imply canonical parsing.

AaronFeickert: Separately, it was mentioned that the naming of the opcode may be confusing. Hashing to a group element is usually taken to mean something very different than what is done here, and confusion could have very bad results. Perhaps something like `ToPublicKey` or `ScalarToPublicKey` instead? I'm not sure the best naming for this.

AaronFeickert: See #5818.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -818,7 +820,9 @@ where B: BlockchainBackend
         // to increase the timestamp to be greater than the median timestamp otherwise the block wont be accepted by
         // nodes
         if median_timestamp > header.timestamp {
-            header.timestamp = median_timestamp.increase(1);
+            header.timestamp = median_timestamp
+                .checked_add(EpochTime::from(1))
+                .ok_or(ChainStorageError::OutOfRange)?;
```
```suggestion
                .ok_or(ChainStorageError::UnexpectedResult("Timestamp calculation overflowed".to_string()))?;
```
The OutofRange error is used when requests are out of range, this is not the case here and this is closer to the actual error.

infrastructure/tari_script/src/script.rs
```diff
@@ -637,11 +631,11 @@ impl TariScript {
...
```
SWvheerden:
```suggestion
            StackItem::Hash(hash) => hash.as_slice(),
```
There is an issue for this, so I am deleting the todo as these just get left in code and forgotten about.

SWvheerden (commented):
Looking good, just 1 nit about an error type

infrastructure/tari_script/src/script.rs
AaronFeickert: Sounds good! For future reference, the issue is #5818.

base_layer/core/src/chain_storage/blockchain_database.rs
AaronFeickert: Will update, thanks! I didn't know the best error to use.

AaronFeickert:
Marking as draft to make it clear that merging is blocked on dependency updates. Specifically, `tari-crypto` needs a new release version.

SWvheerden:
Replacing with PR: https://github.com/tari-project/tari/pull/5900 so that it includes the tari_crypto release tag 0.19.0 instead of the github main branch

PR #5643, opened 2023-08-15
refactor: change atomics to use SeqCst

stringhandler:
Description
---
Change all Atomic* to use SeqCst instead of a mishmash of Relaxed, Acquire and Release

Motivation and Context
---
From the nomicom:

> In practice, sequential consistency is rarely necessary for program correctness. However sequential consistency is definitely the right choice if you're not confident about the other memory orders. Having your program run a bit slower than it needs to is certainly better than it running incorrectly!

> Asking for guarantees that are too weak on strongly-ordered hardware is more likely to happen to work, even though your program is strictly incorrect. If possible, concurrent algorithms should be tested on weakly-ordered hardware.

So IMO, it's better to err on the safe side since we can't practically test every PR on ARM machines.

How Has This Been Tested?
---
As described in the nomicon, there will be little difference between these orderings on my machine. I suspect there will be no noticeable difference in CI either

---
See above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/proxy.rs
+++ b/applications/minotari_merge_mining_proxy/src/proxy.rs
@@ -195,9 +195,9 @@ impl InnerService {
             .as_ref()
             .map(|meta| meta.height_of_longest_chain)
             .ok_or(MmProxyError::GrpcResponseMissingField("base node metadata"))?;
-        if result.get_ref().initial_sync_achieved != self.initial_sync_achieved.load(Ordering::Relaxed) {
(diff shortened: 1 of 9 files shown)


sdbondi (approved):
Agree with this. We don't/won't test on architecture with weak ordering and none of the usages would impact perf even if they did apply.

CjS77:
ACK

SWvheerden (approved):

hansieodendaal (approved):
utACK

PR #5609, opened 2023-08-03
wip: ledger key manager

brianp:
Description
---
Last PR got closed? This is a WIP for tracking the progress of making the wallet work with the ledger

Motivation and Context
---
So people can use their most excellent hardware wallets with the MinoTari console wallet.

How Has This Been Tested?
---

What process can a PR reviewer use to test or verify this change?
---

Breaking Changes
---

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

PR #5644, opened 2023-08-16
feat: ledger key manager interface

brianp:
Description
---
This PR expands the console wallet and key manager interfaces to support hardware wallets. It *does not* introduce functionality to communicate with the hardware wallet, simply the interface to support it. Another PR will be opened swapping out the placeholder software key manager for the hardware key manager where appropriate.

Motivation and Context
---
So people can use their most excellent hardware wallets with the MinoTari console wallet.

How Has This Been Tested?
---
Manually on osx

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
This PR was previously tracked on #5609 but stopped updating when new commits were pushed. This is the same PR but updated against development and with fixes in the test suite which is now green.

hansieodendaal:
I added a fix for the failing cucumber test - see #5645

applications/minotari_console_wallet/Cargo.toml
```diff
@@ -29,14 +29,16 @@ console-subscriber = "0.1.8"
...
 bitflags = "1.2.1"
+blake2 = "0.10"
 chrono = { version = "0.4.19", default-features = false }
 clap = { version = "3.2", features = ["derive", "env"] }
 config = "0.13.0"
 crossterm = { version = "0.25.0" }
 digest = "0.10"
 futures = { version = "^0.3.16", default-features = false, features = ["alloc"] }
-log4rs = { git = "https://github.com/tari-project/log4rs.git", default_features = false, features = ["config_parsing", "threshold_filter", "yaml_format", "console_appender", "rolling_file_appender", "compound_policy", "size_trigger", "fixed_window_roller", "delete_roller"] }
+ledger-transport-hid = { git = "https://github.com/Zondax/ledger-rs" }
```
SWvheerden: lock to a commit

hansieodendaal (changes requested):
Hi @brianp, I can appreciate that this was a lot of work, however, I am not a fan of this approach.

In this PR we create a lot of scaffolding just to be able to create a ledger-specific or software-specific transaction key manager. My recommendations:

- Initial startup
  - Add a command line switch to force the use of the ledger hardware wallet - if the ledger is not attached and the app not started panic with the error message - if attached and app started ok.
  - If the ledger is attached but the wallet is started without the switch, ask the user if this should be a hardware console wallet, if yes follow the logic as if the command line switch was present.
  - ~~When started ok, earmark the console wallet as a ledger-only console wallet - consecutive startups without the ledger attached and the app started must panic.~~

- Key manager
  - We should not differentiate outside of the key manager if we have a ledger or not, so no `ledger_key_manager`.
  - The key manager should be aware if a hardware wallet should be used and if it is a ledger and later on perhaps a trezor if that is added.
  - Inside the key manager whenever we have to use a secret key, for example inside `inner/get_receiver_partial_metadata_signature(..)`, the appropriate software or hardware branch is called. There are not that many places where this branching will be necessary. All ledger-specific code can go into `key_manager/ledger.rs`, and later on we can add `key_manager/trezor.rs`.

- Private keys to be managed
  - ~~Both spend and script private keys should be managed by the ledger, but with the difference that the spend private key should be returned from the ledger to be used inside the key manager where needed and the script private key should never leave the ledger.~~
  - ~~This will mean that range-proof bulletproofs can be constructed outside the ledger but not without it, and script signatures will be wholly constructed inside the ledger.~~
  - The same index must be used to obtain the matching set of ~~spend and script~~ private keys, as is currently the case with the software key manager.

- ~~Spending scripts~~
  - ~~If both spend and script private keys are managed by the ledger, a hardware wallet will be able to receive an interactive transaction with any type of script, and will not be able to be sussed out as a hardware wallet by penny/poking transactions.~~

SWvheerden:
The ledger implementation very much follows this RFC: https://github.com/tari-project/rfcs/pull/98
This means spend keys should be managed by the software wallet, not the hardware wallet aka ledger in this first implementation.

As per the RFC, this allows us to receive transactions without the hardware wallet being online, allowing it to be a cold storage wallet.

leet4tari (commented):
utACK - scripts/install_ubuntu_dependencies.sh

applications/minotari_console_wallet/src/init/mod.rs
```diff
@@ -804,6 +810,53 @@ pub(crate) fn boot_with_password(
...
+    match boot_mode {
+        WalletBoot::New => {
+            if prompt("\r\nWould you like to use a connected hardware wallet? (Supported types: Ledger)") {
+                print!("Scanning for connected Ledger hardware device... ");
+                let err = "No connected device was found. Please make sure the device is plugged in before continuing.";
+                match TransportNativeHID::new(&HidApi::new().expect(err)) {
+                    Ok(_) => {
+                        println!("Device found.");
+                        let account = prompt_ledger_account().expect("An account value");
+                        Some(WalletType::Ledger(account))
```
hansieodendaal: The `TRANSPORT`  lazy static should be assigned a value
```suggestion
                    Ok(transport) => {
                        *TRANSPORT.lock().expect("lock exists") = Some(transport);
                        println!("Device found.");
                        let account = prompt_ledger_account().expect("An account value");
                        Some(WalletType::Ledger(account))
```

hansieodendaal (commented):
Nice so far, just one comment for now, apart form the CI/clippy errors.

hansieodendaal (dismissed):
utACK

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -429,6 +443,40 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
     // Transaction input section (transactions > transaction_components > transaction_input)
     // -----------------------------------------------------------------------------------------------------------------
 
+    pub async fn get_script_private_key(&self, script_key_id: &TariKeyId) -> Result<PrivateKey, TransactionError> {
```
SWvheerden, line 448: This is fine for now, but in the actual implementation we should ver get this script key out of the ledger, we should ask the ledger to sign the script signature for us and we should get that out

applications/minotari_console_wallet/src/init/mod.rs
brianp: This actually sent me on a wild goose chase trying to find where the code had previously been removed.

But, we don't actually need to assign this here. We check to make sure the ledger is present to see if we should create a ledger backed wallet. But the ledger could be unplugged anytime between this prompt, and performing a transaction. It doesn't need to be plugged in at all times. We need to check to see if it's present again, only during script and meta data sig generation for send transactions.

SWvheerden (approved):

PR #5653, opened 2023-08-22
feat: change default script to PushPubKey

brianp:
Description
---
Change the default script to `PushPubKey` to ensure the spending of outputs from standard transactions is only performed with the use of specific private keys.

Motivation and Context
---
This change to the default helps enforce that UTXOS can't be spent from a software wallet, when a hardware wallet is in use.

How Has This Been Tested?
---
Manually, locally, but pretty tests still require updating.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -802,7 +802,7 @@ where
             num_kernels,
             num_outputs
         );
-        // We assume that default OutputFeatures and Nop TariScript is used
+        // We assume that default OutputFeatures and PushPubKey TariScript is used
         let features_and_scripts_byte_size = self
             .resources
             .consensus_constants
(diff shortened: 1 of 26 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1243,7 +1242,7 @@ where
         let (change_spending_key_id, _spend_public_key, change_script_key_id, change_script_public_key) =
             self.resources.key_manager.get_next_spend_and_script_key_ids().await?;
         builder.with_change_data(
-            script!(Nop),
+            script!(PushPubKey(Box::new(change_script_public_key.clone()))),
             inputs!(change_script_public_key),
```
SWvheerden:
```suggestion
           ExecutionStack::default();
```

PR #5655, opened 2023-08-22
feat: cleanup errors

SWvheerden:
Description
---
This ensures that we do not have any unused error types left in the validation errors. 
Renames some old errors to more clearly represent the actual error. 

Motivation and Context
---
Some errors where unused while some where sending the wrong error type back

### DIFF
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -157,7 +157,6 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
                     self.db.cleanup_orphans().await?;
                     return Ok(());
                 },
-                Err(err @ BlockSyncError::ValidationError(ValidationError::AsyncTaskFailed(_))) => return Err(err),
                 Err(BlockSyncError::ValidationError(err)) => {
                     match &err {
                         ValidationError::BlockHeaderError(_) => {},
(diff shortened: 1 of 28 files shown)


base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -306,15 +305,12 @@ impl<B: BlockchainBackend + 'static> BlockSynchronizer<B> {
...
-            .map_err(|err| ValidationError::CustomError(err.to_string()))?;
+            .await?;
 
             let block = match res {
                 Ok(block) => block,
                 Err(err @ ValidationError::BadBlockFound { .. }) |
                 Err(err @ ValidationError::FatalStorageError(_)) |
-                Err(err @ ValidationError::AsyncTaskFailed(_)) |
-                Err(err @ ValidationError::CustomError(_)) => return Err(err.into()),
                 Err(err) => {
```
stringhandler, line 315: This changes the logic, are you sure it's correct?

base_layer/core/src/base_node/sync/header_sync/validator.rs
```diff
@@ -138,8 +138,7 @@ impl<B: BlockchainBackend + 'static> BlockHeaderSyncValidator<B> {
             // We dont want to mark a block as bad for internal failures
             Err(
                 e @ ValidationError::FatalStorageError(_) |
-                e @ ValidationError::IncorrectNumberOfTimestampsProvided { .. } |
-                e @ ValidationError::AsyncTaskFailed(_),
```
stringhandler, line 142: Another logic change, are you sure it's correct?

stringhandler (commented):
Nice work, but I'm worried that AsyncTaskError is now handled in a different match arm. Was that intended?

base_layer/core/src/base_node/sync/header_sync/validator.rs
SWvheerden: This is correct. 
This only suppresses internal errors to let them be the reason a block is flag as a bad block

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
SWvheerden: This is wrong, yes. BadBlockFound and FatalStorageError should still result in return err.

stringhandler (approved):

PR #5658, opened 2023-08-23
fix: memory overflow panic

SWvheerden:
Description
---
Limits the varint size that the merkle_tree, covenent, script and stack can be.

Motivation and Context
---
`fn deserialize_reader<R>(reader: &mut R) -> Result<Self, io::Error>` trusts that the len of the varint declaring the size of the vec<u8> it needs to read is honest. This can cause a node to comment up to u64::MAX memory and cause a memory overflow. This adds a hard limit to what the size can be and checks it before allocating the memory. 

How Has This Been Tested?
---
Unit tests

Audit Finding Number
--
TARI-020

### DIFF
--- a/base_layer/core/src/covenants/covenant.rs
+++ b/base_layer/core/src/covenants/covenant.rs
@@ -55,8 +55,7 @@ pub struct Covenant {
 impl BorshSerialize for Covenant {
     fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
         let bytes = self.to_bytes();
-        // writer.write_varint(bytes.len())?;
-        writer.write_varint(usize::MAX)?;
+        writer.write_varint(bytes.len())?;
         for b in &bytes {
             b.serialize(writer)?;
         }
--- a/base_layer/core/src/covenants/covenant.rs
+++ b/base_layer/core/src/covenants/covenant.rs
@@ -55,7 +55,8 @@ pub struct Covenant {
 impl BorshSerialize for Covenant {
     fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
         let bytes = self.to_bytes();
-        writer.write_varint(bytes.len())?;
+        // writer.write_varint(bytes.len())?;
+        writer.write_varint(usize::MAX)?;
         for b in &bytes {
             b.serialize(writer)?;
         }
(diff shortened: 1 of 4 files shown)


base_layer/core/src/covenants/covenant.rs
```diff
@@ -55,7 +55,8 @@ pub struct Covenant {
 impl BorshSerialize for Covenant {
     fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
         let bytes = self.to_bytes();
-        writer.write_varint(bytes.len())?;
+        // writer.write_varint(bytes.len())?;
+        writer.write_varint(usize::MAX)?;
```
sdbondi: Test code?

stringhandler (dismissed):

base_layer/core/src/proof_of_work/monero_rx/merkle_tree.rs
```diff
@@ -142,6 +144,12 @@ impl BorshDeserialize for MerkleProof {
     fn deserialize_reader<R>(reader: &mut R) -> Result<Self, io::Error>
     where R: io::Read {
         let len = reader.read_varint()?;
+        if len > MAX_MERKLE_TREE_BYTES {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Larger than max covenant bytes".to_string(),
```
stringhandler:
```suggestion
                "Larger than max merkle tree bytes".to_string(),
```

infrastructure/tari_script/src/script.rs
```diff
@@ -76,6 +77,12 @@ impl BorshDeserialize for TariScript {
     fn deserialize_reader<R>(reader: &mut R) -> Result<Self, io::Error>
     where R: io::Read {
         let len = reader.read_varint()?;
+        if len > MAX_SCRIPT_BYTES {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Larger than max covenant bytes".to_string(),
```
stringhandler:
```suggestion
                "Larger than max script bytes".to_string(),
```

stringhandler: In consensus constants this is set to 2048 but it doesn't necessarily need to be that here

infrastructure/tari_script/src/stack.rs
```diff
@@ -199,6 +199,12 @@ impl BorshDeserialize for ExecutionStack {
     fn deserialize_reader<R>(reader: &mut R) -> Result<Self, io::Error>
     where R: io::Read {
         let len = reader.read_varint()?;
+        if len > MAX_STACK_SIZE {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Larger than max covenant bytes".to_string(),
```
stringhandler: wrong name again

stringhandler (commented):
Some copy pasta errors

infrastructure/tari_script/src/script.rs
SWvheerden: I see it as still serving two purposes. 
The one is the max safety one for deserialization, the other one is the max allowed by the network for network consensus

stringhandler (approved):

PR #5661, opened 2023-08-23
fix: adds bans for horizon sync

SWvheerden:
Description
---
Adds bans for horizon sync bad behavior
Cleans up the horizonsync errors

Motivation and Context
---
We should ban bad behavior, which we don't do currently.  

Audit Finding Number
---
TARI-017

### DIFF
--- a/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
@@ -196,7 +196,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                 Ok(_) => match self.finalize_horizon_sync(sync_peer).await {
                     Ok(_) => return Ok(()),
                     Err(err) => {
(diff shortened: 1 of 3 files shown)


base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
```diff
@@ -217,6 +219,46 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
         Err(HorizonSyncError::FailedSyncAllPeers)
     }
 
+    async fn ban_peer_on_ban_able_error(
```
stringhandler:
```suggestion
    async fn ban_peer_on_bannable_error(
```

base_layer/core/src/base_node/sync/horizon_state_sync/synchronizer.rs
```diff
@@ -208,6 +209,7 @@ impl<'a, B: BlockchainBackend + 'static> HorizonStateSynchronization<'a, B> {
                     }
```
stringhandler, line 210: These errors should also result in a ban

stringhandler (approved):

PR #5663, opened 2023-08-25
fix: cases where bad incoming blocks could bypass a ban

SWvheerden:
Description
---
In some cases incoming blocks can escape a ban. 

Motivation and Context
---
Participating in bad behavior should result in a network ban.

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -512,17 +512,25 @@ where B: BlockchainBackend + 'static
         new_block: NewBlock,
         source_peer: NodeId,
     ) -> Result<(), CommsInterfaceError> {
-        if let Err(err @ CommsInterfaceError::ChainStorageError(ChainStorageError::ValidationError { .. })) =
-            self.handle_new_block_message_inner(new_block, source_peer.clone()).await
+        match self
+            .handle_new_block_message_inner(new_block, source_peer.clone())
+            .await
         {
-            if let Err(e) = self
-                .connectivity
-                .ban_peer(source_peer, format!("Peer propagated invalid block: {}", err))
-                .await
-            {
-                error!(target: LOG_TARGET, "Failed to ban peer: {}", e);
-            }
-            return Err(err);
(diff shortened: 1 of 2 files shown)


stringhandler (commented):
Not really a fan of this approach, it would be better to ban higher up. I don't know if it's possible, but this way introduces a lot of duplicate code

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -506,7 +503,35 @@ where B: BlockchainBackend + 'static
...
+    /// Handles a `NewBlock` message. Only a single `NewBlock` message can be handled at once to prevent extraneous
+    /// requests for the full block.
+    /// This may (asynchronously) block until the other request(s) complete or time out and so should typically be
+    /// executed in a dedicated task.
+    pub async fn handle_new_block_message(
+        &mut self,
+        new_block: NewBlock,
+        source_peer: NodeId,
+    ) -> Result<(), CommsInterfaceError> {
+        if let Err(err @ CommsInterfaceError::ChainStorageError(ChainStorageError::ValidationError { .. })) =
```
stringhandler: You are swallowing all the other error types here.... I think this is worse

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -520,13 +545,21 @@ where B: BlockchainBackend + 'static
                     self.blockchain_db.get_chain_metadata().await?.height_of_longest_chain(),
                 )
                 .await?;
-            return Err(CommsInterfaceError::InvalidBlockHeader(
-                BlockHeaderValidationError::ProofOfWorkError(PowError::AchievedDifficultyBelowMin),
-            ));
+            let err = BlockHeaderValidationError::ProofOfWorkError(PowError::AchievedDifficultyBelowMin);
```
stringhandler, line 552: Why not let this be caught higher up?

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
SWvheerden: your right,this is worse

SWvheerden:
closing this in favour of #5693

PR #5681, opened 2023-08-28
feat: chatffi simpler callbacks and managed identity and db

brianp:
Description
---
Offer a managed identity file, and db creation. 
Simplify callback return values to c-readable structs.

Motivation and Context
---
- Wallet can't provide it's identity file, and we probably shouldn't share it anyway. So lets write and use our own in the chatffi.
- Callbacks couldn't be read as rust structs
- db init can be managed on our end

How Has This Been Tested?
---
CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/chat_ffi/src/callback_handler.rs
+++ b/base_layer/chat_ffi/src/callback_handler.rs
@@ -35,6 +35,7 @@ const LOG_TARGET: &str = "chat_ffi::callback_handler";
 pub(crate) type CallbackContactStatusChange = unsafe extern "C" fn(*mut ChatFFIContactsLivenessData);
 pub(crate) type CallbackMessageReceived = unsafe extern "C" fn(*mut ChatFFIMessage);
 
+#[repr(C)]
 pub struct ChatFFIContactsLivenessData {
     pub address: *const c_char,
     pub last_seen: u64,
@@ -66,6 +67,7 @@ impl TryFrom<ContactsLivenessData> for ChatFFIContactsLivenessData {
     }
 }
 
+#[repr(C)]
 pub struct ChatFFIMessage {
     pub body: *const c_char,
     pub from_address: *const c_char,
--- a/base_layer/chat_ffi/src/callback_handler.rs
+++ b/base_layer/chat_ffi/src/callback_handler.rs
@@ -20,9 +20,10 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::ops::Deref;
+use std::{convert::TryFrom, ffi::CString, ops::Deref};
 
-use log::{debug, info, trace};
+use libc::c_char;
+use log::{debug, error, info, trace};
 use tari_contacts::contacts_service::{
     handle::{ContactsLivenessData, ContactsLivenessEvent, ContactsServiceHandle},
     types::Message,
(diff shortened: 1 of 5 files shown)


SWvheerden (dismissed):

stringhandler (dismissed):

base_layer/chat_ffi/src/callback_handler.rs
```diff
@@ -31,8 +32,74 @@ use tari_shutdown::ShutdownSignal;
 
 const LOG_TARGET: &str = "chat_ffi::callback_handler";
 
-pub(crate) type CallbackContactStatusChange = unsafe extern "C" fn(*mut ContactsLivenessData);
-pub(crate) type CallbackMessageReceived = unsafe extern "C" fn(*mut Message);
+pub(crate) type CallbackContactStatusChange = unsafe extern "C" fn(*mut ChatFFIContactsLivenessData);
+pub(crate) type CallbackMessageReceived = unsafe extern "C" fn(*mut ChatFFIMessage);
+
+pub struct ChatFFIContactsLivenessData {
```
SWvheerden, line 39:
```suggestion
#[repr(c)]
pub struct ChatFFIContactsLivenessData {
```
if you want to make this available over the FFI you need to add this? If not ignore this message

base_layer/chat_ffi/src/callback_handler.rs
```diff
@@ -31,8 +32,74 @@ use tari_shutdown::ShutdownSignal;
...
+
+        Ok(Self {
+            address: address.as_ptr(),
+            last_seen,
+            online_status: v.online_status().as_u8(),
+        })
+    }
+}
+
+pub struct ChatFFIMessage {
```
SWvheerden, line 71:
```suggestion
#[repr(c)]
pub struct ChatFFIMessage {
```
Same thing here?

base_layer/chat_ffi/src/callback_handler.rs
```diff
@@ -103,8 +170,14 @@ impl CallbackHandler {
             "Calling ContactStatusChanged callback function for contact {}",
             data.address(),
         );
-        unsafe {
-            (self.callback_contact_status_change)(Box::into_raw(Box::new(data)));
+
+        match ChatFFIContactsLivenessData::try_from(data) {
+            Ok(data) => unsafe {
```
SWvheerden, line 177: why is this an unsafe call?

base_layer/chat_ffi/src/callback_handler.rs
```diff
@@ -114,8 +187,12 @@ impl CallbackHandler {
             "Calling MessageReceived callback function for sender {}",
             message.address,
         );
-        unsafe {
-            (self.callback_message_received)(Box::into_raw(Box::new(message)));
+
+        match ChatFFIMessage::try_from(message) {
+            Ok(message) => unsafe {
```
SWvheerden, line 194: same here?

base_layer/chat_ffi/src/callback_handler.rs
```diff
@@ -103,8 +170,14 @@ impl CallbackHandler {
             "Calling ContactStatusChanged callback function for contact {}",
             data.address(),
         );
-        unsafe {
-            (self.callback_contact_status_change)(Box::into_raw(Box::new(data)));
+
+        match ChatFFIContactsLivenessData::try_from(data) {
+            Ok(data) => unsafe {
+                (self.callback_contact_status_change)(Box::into_raw(Box::new(data)));
```
SWvheerden, line 178: Why are we using a box here? Is the data to big tobe on the stack?
Box per say is not a problem, but using `Box::into_raw` means we have to ensure the safety of the data. Are we making sure we clean this up after we use it?

SWvheerden (commented):
Looking good, just have a few questions.

base_layer/chat_ffi/src/callback_handler.rs
brianp: Good catch.

base_layer/chat_ffi/src/callback_handler.rs
brianp: Good catch.

base_layer/chat_ffi/src/callback_handler.rs
brianp: We're using box because the data is accessed directly outside "our" scope, through the ffi boundaries. So we need to ensure it lives long enough. 

Currently, we are not ensuring it's cleaned up. So I'll make a destroy available.

base_layer/chat_ffi/src/callback_handler.rs
brianp: I double checked just to be sure, but this is because the value of `self.callback_contact_status_change` is of type `unsafe extern "C" fn(*mut ChatFFIContactsLivenessData);`

base_layer/chat_ffi/src/callback_handler.rs
brianp: destroy functions added.

base_layer/chat_ffi/src/callback_handler.rs
brianp: updated.

SWvheerden (approved):

PR #5688, opened 2023-08-29
fix: potential u64 overflow panic

SWvheerden:
Description
---
fork_hash_index comes from remote peer, if the remote peer sets this to u64::Max, the local node will panic. 
This data is sanity checked later on, but here a panic is possible

### DIFF
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -420,7 +420,11 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                     return Err(err.into());
                 },
             };
-
+            if resp.headers.len() > NUM_INITIAL_HEADERS_TO_REQUEST {
+                self.ban_peer_long(peer, BanReason::PeerSentTooManyHeaders(resp.headers.len()))
+                    .await?;
+                return Err(BlockHeaderSyncError::NotInSync);
+            }
             if resp.fork_hash_index >= block_hashes.len() as u64 {
                 let _result = self
                     .ban_peer_long(peer, BanReason::SplitHashGreaterThanHashes {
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -421,7 +421,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
                 },
             };
 
-            let steps_back = resp.fork_hash_index + offset as u64;
+            let steps_back = resp.fork_hash_index.saturating_add(offset as u64);
```
sdbondi, line 441: This should be a protocol violation if fork_hash_index is greater than `>` the number of block hashes sent - 1.

sdbondi: Then once that is checked, checked_add should be used - if that fails it can only be a bug in the code really (we've iterated `u64::MAX / NUM_CHAIN_SPLIT_HEADERS` times )

sdbondi (commented):
Good catch - have some suggestions on how to handle it

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: That is checked.
Just a few lines later in the code. This just ensures we don't crash before we sanity checks this
https://github.com/tari-project/tari/blob/385022ad32ea10ad6e73ac6eb972f4ac90d43a4e/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs#L467C5-L467C5

stringhandler: maybe we should use checked_add and ban if it overflows

SWvheerden: I moved the check thats its less than the headers, to here, and we do ban if that is out. It is order of magnitudes less than u64::max

SWvheerden: I moved both bans, to in front, so technically it can't overflow anymore, but going to leave that as is.

stringhandler (approved):

PR #5701, opened 2023-08-31
fix: handle out of sync errors when returning mempool transactions

brianp:
Description
---
Add a new error type and gracefully recover from possible runtime panic.

Motivation and Context
---
Currently, there is a minimal possibility that the collection of tx's is out sync. If this occurs the base node will likely crash. 

How Has This Been Tested?
---
CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/mempool/mempool_storage.rs
+++ b/base_layer/core/src/mempool/mempool_storage.rs
@@ -285,14 +285,22 @@ impl MempoolStorage {
         Ok(results.retrieved_transactions)
     }
 
-    pub fn retrieve_by_excess_sigs(&self, excess_sigs: &[PrivateKey]) -> (Vec<Arc<Transaction>>, Vec<PrivateKey>) {
-        let (found_txns, remaining) = self.unconfirmed_pool.retrieve_by_excess_sigs(excess_sigs);
-        let (found_published_transactions, remaining) = self.reorg_pool.retrieve_by_excess_sigs(&remaining);
+    pub fn retrieve_by_excess_sigs(
+        &self,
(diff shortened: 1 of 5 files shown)


base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -285,14 +285,22 @@ impl MempoolStorage {
...
     }
 
-    pub fn retrieve_by_excess_sigs(&self, excess_sigs: &[PrivateKey]) -> (Vec<Arc<Transaction>>, Vec<PrivateKey>) {
-        let (found_txns, remaining) = self.unconfirmed_pool.retrieve_by_excess_sigs(excess_sigs);
-        let (found_published_transactions, remaining) = self.reorg_pool.retrieve_by_excess_sigs(&remaining);
+    pub fn retrieve_by_excess_sigs(
+        &self,
+        excess_sigs: &[PrivateKey],
+    ) -> Result<(Vec<Arc<Transaction>>, Vec<PrivateKey>), MempoolError> {
+        let (found_txns, remaining) = match self.unconfirmed_pool.retrieve_by_excess_sigs(excess_sigs) {
```
SWvheerden:
```suggestion
        let (found_txns, remaining) = self.unconfirmed_pool.retrieve_by_excess_sigs(excess_sigs)?;
```
why not do this?

brianp: Because I overlooked the possibility 🙃 

updated.

SWvheerden (approved):

PR #5702, opened 2023-08-31
fix(comms/dht)!: limit number of peer claims and addresses for all sources

sdbondi:
Description
---
Implement validations and bans for peers that violate peer identity exchange protocol rules
Consolidate peer claim and identity validations
Removed a number of dead functions and error variants
Reduced memory footprint of peer connection handle
Removed unverified claim data, this data is not unverified and is only ever added for authenticated direct connections
Check that discovery response comes from the peer that directly connected to us
Reduce peer flags size from u64 to u32 (smaller would require many more changes)
Reduce memory footprint in dialer

Motivation and Context
---
TARI-021

All peer source validations use the same code. We previously allowed a peer to send any number of identity claims when doing discovery. Discovery RPC protocol now allows a limited number of claims to be sent, ordered by local quality score.

How Has This Been Tested?
---
Existing test, localnet network (2 bn 2 wallets)

---

Breaking Changes
---

- [ ] None
- [x] Requires _peer_ data directory on base node to be deleted
- [ ] Requires hard fork
- [x] Other - some network messages relating to peer sync are not compatible

### DIFF
--- a/comms/dht/src/discovery/service.rs
+++ b/comms/dht/src/discovery/service.rs
@@ -20,22 +20,15 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{
-    collections::HashMap,
-    convert::{TryFrom, TryInto},
-    sync::Arc,
-    time::Instant,
-};
+use std::{collections::HashMap, convert::TryFrom, sync::Arc, time::Instant};
 
 use log::*;
 use rand::{rngs::OsRng, RngCore};
 use tari_comms::{
+    connectivity::ConnectivityRequester,
     log_if_error,
-    multiaddr::Multiaddr,
-    net_address::PeerAddressSource,
-    peer_manager::{NodeId, NodeIdentity, Peer, PeerFeatures, PeerIdentityClaim, PeerManager},
+    peer_manager::{NodeId, NodeIdentity, Peer, PeerManager},
     types::CommsPublicKey,
-    validate_addresses,
 };
 use tari_shutdown::ShutdownSignal;
 use tari_utilities::{hex::Hex, ByteArray};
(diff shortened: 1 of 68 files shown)


comms/dht/src/discovery/service.rs
```diff
@@ -182,7 +174,22 @@ impl DhtDiscoveryService {
                     start_ts,
                 } = request;
 
-                let result = self.validate_then_add_peer(&public_key, discovery_msg).await;
+                // Make sure that the response is for the expected public key
+                if discovery_msg.public_key.as_bytes() != public_key.as_bytes() {
```
SWvheerden, line 189: this seems like a bannable offense

SWvheerden (commented):
Looks good, one question thou

comms/dht/src/discovery/service.rs
sdbondi: ~~Ah yeah this is actually checked before and banned in the dht message handler middleware, so shouldn't need to check it here again~~

Was mistaken while looking on my phone - this is a new ban because the peer responded from a discovery request with a different public key which should never happen for honest peers. I have a branch that I was adding ban infrastructure code in, I have merged this in this rather large PR to make this case work.

sdbondi: There are more bans, but i think lets save them for another PR

comms/dht/src/store_forward/saf_handler/task.rs
```diff
@@ -341,7 +341,7 @@ where S: Service<DecryptedDhtMessage, Response = (), Error = PipelineError>
 
                     // Every other error shouldn't happen if the sending node is behaving
                     Err(err) => {
-                        //  #banheuristics
+                        // TODO: #banheuristics
```
SWvheerden:
```suggestion
                        //  #banheuristics
```

Created an issue specifically for this: https://github.com/tari-project/tari/issues/5709

SWvheerden: todo's in code gets left out and we forget about them, a todo with doing is worth creating an issue for :)

sdbondi: Thanks working on SAF bans as we speak (new PR)

stringhandler (approved):

PR #5705, opened 2023-08-31
fix: better timeout for lagging

stringhandler:
Description
---
Removes some very specific off by one check, and instead use a configured timeout before considering lagging.

This should also prevent a bug where a node knows of a stronger chain, but it is at a lower height and this would cause it to think it was up to date

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/listening.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/listening.rs
@@ -209,6 +209,11 @@ impl Listening {
                                 sync_peers: sync_peers.clone(),
                             });
                         }
+                    } else {
+                        // We might have gotten up to date via propagation outside of this state, so reset the timer
+                        if sync_mode == SyncStatus::UpToDate {
+                            time_since_better_block = None;
+                        }
                     }
 
                     if sync_mode.is_lagging() {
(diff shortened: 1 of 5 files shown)


base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -213,6 +183,34 @@ impl Listening {
...
+                    if let SyncStatus::BehindButNotYetLagging {
+                        local,
+                        network,
+                        sync_peers,
+                    } = &sync_mode
+                    {
+                        if time_since_better_block.is_none() {
+                            time_since_better_block = Some(Instant::now());
+                        }
+                        if time_since_better_block
```
sdbondi, line 202: Where does this get set back to None when a block is received?

sdbondi: think we need to do that when sync status is up to date

stringhandler: yeah, I had assumed it would leave this state eventually, but you are right it could catch up outside of this

stringhandler: added a reset, lmk what you think

SWvheerden (approved):
On thing to note here, is that the way its written here is that fallen_behind_blocks take preference above time_since_better_block as the time wont be triggered if the block behind does not trigger.

base_layer/core/src/base_node/state_machine_service/states/listening.rs
sdbondi: Nice, LGTM

sdbondi (approved):

PR #5706, opened 2023-08-31
fix: addition overflow when coinbase + fees is too high

brianp:
Description
---
Use a `checked_addition` and catch errors when overflow occurs.

Motivation and Context
---
Block could be produced and shared that have coinbases and fees that are too high resulting in an overflow and node crash.

How Has This Been Tested?
---
CI

---
Mostly see if you like what's happening. In `base_node_grpc_server.rs` a fair bit of the new wrapping results in either getting the expected result or returning this specific problem and I'm not so sure it's the most generic way to recover.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
I copied a pattern from somewhere else in this file and SW has pointed out that pattern is _not doing what we want_. So I'll refactor this up and also fix where I copied the initial pattern from.

SWvheerden (approved):

PR #5708, opened 2023-08-31
fix: also ban on chainblock error

stringhandler:
Also ban if we can't construct a ChainBlock (hash does not match)


SWvheerden:
I disagree with this. 
If this is an error, it's something that's broken locally. 

We get a block from a remote peer. 
We use that to construct accumulated data from the block. But somehow the data we constructed from the block, does not match the block we got. So the hash we calculated, does not match the hash we calculated later on.

PR #5691, opened 2023-08-30
feat: improve block sync error handling

hansieodendaal:
Description
---
Improved block sync error handling
- Comms and PRC errors will not invalidate the block sync in itself, but permit block sync to be tried from another sync peer.
- Better and more consistent clean-up if block sync fails.
- More consistent retries from another sync peer in case of errors.

Motivation and Context
---
Error handling was not consistent and opened avenues for malicious peers to interfere.

How Has This Been Tested?
---
Existing tests pass

What process can a PR reviewer use to test or verify this change?
---
Code walkthrough

Breaking Changes
---

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

PR #5713, opened 2023-09-01
fix(sync): unify ban logic in all sync processes

mrnaveira:
Description
---
* New common `PeerManager` struct that handles peer bans. All synchronizers reuse this manager to perform the bans.
* Each synchornizer error definition (`HorizonSyncError`, `BlockHeaderSyncError`, etc.) now encapsulates the relationship (and duration) of the bans.
* Fixed logic of peer removal on long bans
* All synchronizers now handle single peer errors the same way
* All synchronizers now handle latency the same way

Motivation and Context
---
All sync processes (`header_sync`, `block_sync` and `horizon_sync`) handle bans differently., but the recommended approach is the recent block sync refactor (https://github.com/tari-project/tari/pull/5691).

The purpose of this PR is for all syncs to behave the same and ensure:
* All errors that should be banned do result in bans
* Peers are not banned for unrelated errors (database, async, etc.)
* Long bans remove the peers from the peer sync list

How Has This Been Tested?
---
Tests pass

---
Code review

Breaking Changes
---
- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (dismissed):
utACK

SWvheerden (dismissed):
pending clippy

PR #5722, opened 2023-09-01
fix!: remove tari prefix and only allow one mergemining tag

stringhandler:
Remove the merge mining tag. There can now only be one merge mining tag in a block

BREAKING CHANGE: Merge mining hash has changed


SWvheerden:
I am really in two minds on this one. 
On the one side, this makes it a slightly more secure in that we don't tamper with the hash. 
but on the other end this goes against Monero's whole move to support more than one merge mining coin at the same time. 
I do believe the change was made mostly for us and town forge.

stringhandler:
I was a bit hasty. I think the is_found check is not 100% correct

PR #5725, opened 2023-09-02
DNM feat: reset database after failed horizon sync

SWvheerden:
Description
---
This will clear out the entire database after horizon sync fails. 

Motivation and Context
---

When horizon sync fails, we know that the database is in an inconsistent state. The safest way to do this is clear out the entire database and start over from genesis block. 

Discussion issue: #5771

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync.rs
@@ -130,6 +130,12 @@ impl HorizonStateSync {
                     randomx_vm_cnt,
                     randomx_vm_flags,
                 });
+                if let Err(e) = shared.db.reset_blockchain_databases().await {
+                    error!(
+                        target: LOG_TARGET,
+                        "Failed to reset chain to 0: {}", e
+                    );
+                };
                 warn!(target: LOG_TARGET, "Synchronizing horizon state has failed. {}", err);
                 StateEvent::HorizonStateSyncFailure
             },
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1117,7 +1117,14 @@ where B: BlockchainBackend
     pub fn reset_blockchain_databases(&self) -> Result<(), ChainStorageError> {
         let mut db = self.db_write_access()?;
         let genesis_block = Arc::new(self.consensus_manager.get_genesis_block());
-        reset_blockchain_databases(&mut *db, self.config.pruning_horizon, genesis_block)
+        clear_blockchain_databases(&mut *db)?;
+        // lets add in the gen block again
+        info!(
+        target: LOG_TARGET,
+        "Cleared DB. Adding genesis block back in {}.",
+        genesis_block.block().body.to_counts_string()
+    );
+        insert_gen_block(&mut *db, self.config.pruning_horizon, genesis_block)
     }
 
     /// Rewind the blockchain state to the block hash making the block at that hash the new tip.
@@ -1875,19 +1882,19 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
     Ok(removed_blocks)
 }
 
-fn reset_blockchain_databases<T: BlockchainBackend>(
+fn clear_blockchain_databases<T: BlockchainBackend>(
+    db: &mut T,
+) -> Result<(), ChainStorageError> {
+    db.clear_blockchain_databases()?;
+    Ok(())
+}
+
+fn insert_gen_block<T: BlockchainBackend>(
     db: &mut T,
     pruning_horizon: u64,
     genesis_block: Arc<ChainBlock>,
 ) -> Result<(), ChainStorageError> {
-    db.reset_blockchain_databases()?;
-    // lets add in the gen block again
 
-    info!(
-        target: LOG_TARGET,
-        "Cleared DB. Adding genesis block back in {}.",
-        genesis_block.block().body.to_counts_string()
-    );
     let mut txn = DbTransaction::new();
     insert_best_block(&mut txn, genesis_block.clone())?;
     let body = &genesis_block.block().body;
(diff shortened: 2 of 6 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2491,6 +2491,38 @@ impl BlockchainBackend for LMDBDatabase {
...
+        lmdb_clear(&write_txn, &self.deleted_txo_mmr_position_to_height_index)?;
+        lmdb_clear(&write_txn, &self.orphans_db)?;
+        lmdb_clear(&write_txn, &self.monero_seed_height_db)?;
+        lmdb_clear(&write_txn, &self.orphan_header_accumulated_data_db)?;
+        lmdb_clear(&write_txn, &self.orphan_chain_tips_db)?;
+        lmdb_clear(&write_txn, &self.orphan_parent_map_index)?;
+        lmdb_clear(&write_txn, &self.reorgs)?;
+        lmdb_clear(&write_txn, &self.validator_nodes)?;
+        lmdb_clear(&write_txn, &self.validator_nodes_mapping)?;
+        lmdb_clear(&write_txn, &self.template_registrations)?;
```
sdbondi: Missing a commit

```suggestion
        for (name, db) in self.all_dbs() {
            debug!(target: LOG_TARGET, "Clearing database '{}'", name);
            lmdb_clear(&write_txn, &db)?;
        }
        
        write_txn.commit()?;
       
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1867,6 +1875,35 @@ fn rewind_to_height<T: BlockchainBackend>(db: &mut T, height: u64) -> Result<Vec
     Ok(removed_blocks)
 }
 
+fn reset_blockchain_databases<T: BlockchainBackend>(
```
sdbondi: I would name this create_genesis_block / state and call this in `BlockchainDatabase::new` (line 241). Remove the  db.reset_blockchain_databases()?; call

Then in reset_blockchain_database (line 1117) I would call reset_blockchain_databases and call create_genesis_block after.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2491,6 +2491,38 @@ impl BlockchainBackend for LMDBDatabase {
         }
         Ok(result)
     }
+
+    fn reset_blockchain_databases(&mut self) -> Result<(), ChainStorageError> {
```
sdbondi:
```suggestion
    fn clear_blockchain_databases(&mut self) -> Result<(), ChainStorageError> {
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1112,6 +1112,14 @@ where B: BlockchainBackend
         rewind_to_height(&mut *db, height)
     }
 
+    /// This will delete and clear out all databases except the list of bad_blocks. This is used when horizon sync
+    /// fails, and we need to reset the blockchain to tip 0 again
+    pub fn reset_blockchain_databases(&self) -> Result<(), ChainStorageError> {
+        let mut db = self.db_write_access()?;
+        let genesis_block = Arc::new(self.consensus_manager.get_genesis_block());
+        reset_blockchain_databases(&mut *db, self.config.pruning_horizon, genesis_block)
```
sdbondi:
```suggestion
        clear_blockchain_databases(&mut *db, self.config.pruning_horizon, genesis_block)
        create_genesis_state/block( ...)
```

base_layer/core/src/base_node/state_machine_service/states/horizon_state_sync.rs
```diff
@@ -130,6 +130,12 @@ impl HorizonStateSync {
                     randomx_vm_cnt,
                     randomx_vm_flags,
                 });
+                if let Err(e) = shared.db.reset_blockchain_databases().await {
```
sdbondi, line 133: I think we'll need to determine this from after header sync - so once we know the size of the reorg from the headers. We'll know if it's beyond the pruning horizon and can reset the db.

sdbondi (commented):
One option is to not handle this case. If the operator sets their pruning horizon close to the tip and their node cannot reorg that is what it is. The problem that we need to solve is that we get into an inconsistent state in the first place. Are there specific cases where this occurs?
 Suggestions:

- check in header sync if we'll reorg past the horizon. Once we find the chain split we can check this, even if the peer is lying we can abort header sync from them. We'll then probably never find a way to sync if the reorg is legit but that is fine.
- The above case, we should maybe add a bn command to reset the state so that the operator can choose to do it
- horizon sync should be continuing from where it left off if we get a spurious network error or something, if this isnt happening we should fix that.

SWvheerden:
This is not something easy to solve, but the problem here is that an attacker can get your node on broken chain, after which your node will be stuck, and potentially ban other nodes if we don't reset this, even if the other ban node has been banned.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2491,6 +2491,25 @@ impl BlockchainBackend for LMDBDatabase {
         }
         Ok(result)
     }
+
+    fn clear_blockchain_databases(&mut self) -> Result<(), ChainStorageError> {
+        let write_txn = self.write_transaction()?;
+
+        for (name, db) in self.all_dbs() {
+            debug!(target: LOG_TARGET, "Clearing database '{}'", name);
+            // we dont want to rest bad blocks only data
```
hansieodendaal:
```suggestion
            // we don't want to reset bad blocks only data
```

hansieodendaal (commented):
I think we need to set the pruning horizon to a large enough minimum value to circumvent some of the issues @sdbondi mentioned, like for instance a week's worth of blocks (i.e. 5040), then, if horizon sync fails we can wipe the databases and start fresh.

SWvheerden:
Yeah, one thing to keep in mind with a setting like this, we can advise "safe" defaults, or defaults we think are safe, but nothing stops anyone from overriding them.

stringhandler (commented):
I agree with the idea here. It's not urgent, so let's think of many options. 

We do already have code that resets the DB and tries to recreate the chain in `check_db`, can we not reuse that functionality?

PR #5726, opened 2023-09-02
fix: validate incoming block before reconcile

stringhandler:
Description
---
Validate an incoming header before trying to reconcile it. 

Note: We will only try to reconcile this block if it's stronger than the current chain.

Motivation and Context
---
todo

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -59,7 +59,7 @@ use crate::{
         PowError,
     },
     transactions::aggregated_body::AggregateBody,
-    validation::{helpers, ValidationError},
+    validation::{header::HeaderFullValidator, helpers, HeaderChainLinkedValidator, ValidationError},
 };
 
 const LOG_TARGET: &str = "c::bn::comms_interface::inbound_handler";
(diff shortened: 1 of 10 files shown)


SWvheerden (changes requested):
I am very much against this for one primary reason:
- We cannot receive blocks out of order anymore. If we receive blocks: 1,2,3,5,4, then our chain will be stuck at block 3, while the current code will adequately handle the case up to block 5 without having to go to sync. 

And as a miner, I can see this as a really big issue, I have also personally seen blocks coming in out of order on multiple occasions. 

This code also introduced a vulnerability, where an attacker can continually spam us with a new invalid block header, and we have to keep validating the blockheader. This code does not ban a bad block as it might be out of order. We must fix this by banning all except out-of-order.

As a secondary issue here, I don't see the attack that were are trying to stop here. If a peer is delaying blocks from them, all they are doing is dos'ing their own comms channel to you. Your node will continue to as normal process all other incoming blocks while only their blocks are being delayed.

hansieodendaal (commented):
Some observations:
- Validating the header is good, but I agree that we should be able to receive blocks out of order and then form a complete chain once all missing blocks are in - we can add a unit test for this
- To prevent spamming and tying up resources, we should not do away with `async fn check_min_block_difficulty` and having a list of bad blocks to check before using more resources
- Peers that send bad blocks should be banned for a long time

PR #5733, opened 2023-09-04
fix: syncing from prune node

SWvheerden:
Description
---
This filters out nodes when deciding if we need to sync or not.

Motivation and Context
---
We should not try and sync from nodes that cannot provide us with the blocks we require. 
This can happen in two cases:

1. We are a pruned node and their pruning horizon is less then ours. 
2. They are a pruning node and their effective pruned height is more than our current tip.  

This means that although we know we are behind, we should not enter syncing mode to those nodes as they will not be able to supply us the blocks we need, so we should not enter sync mode and destroy our current chain.

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/listening.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/listening.rs
@@ -220,7 +220,10 @@ impl Listening {
                         return StateEvent::FallenBehind(sync_mode);
                     }
 
-                    if !self.is_synced {
+                    // if the mode is SyncNotPossible, we know of peers that have a higher pow than us, but we cannot
+                    // sync from them due to their pruning mode settings as they cannot supply us with the required full
+                    // blocks we require, so we should not enter sync and await better peers who can provide us this
+                    if !self.is_synced && !sync_mode.is_sync_not_possible() {
                         self.is_synced = true;
                         shared.set_state_info(StateInfo::Listening(ListeningInfo::new(true)));
                         debug!(target: LOG_TARGET, "Initial sync achieved");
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -295,6 +298,42 @@ fn determine_sync_mode(
...
 
+        // If both the local and remote are pruned mode, we need to ensure that the remote pruning horizon is
+        // greater_equal to ours so that we can sync all the data from it. If the remote is a pruned mode, and
+        // we only require some data from it, we need to ensure that they can supply the data we need, as in their
+        // effective pruned horizon is greater than our local current chain tip.
+        let pruned_mode = local.pruning_horizon() > 0;
+        let pruning_horizon_check = network.claimed_chain_metadata().pruning_horizon() > 0 &&
+            network.claimed_chain_metadata().pruning_horizon() < local.pruning_horizon();
+        let pruning_height_check = network.claimed_chain_metadata().pruned_height() > local.height_of_longest_chain();
+        let sync_able_peer = match (pruned_mode, pruning_horizon_check, pruning_height_check) {
```
stringhandler, line 309: This is a really complicated way of specifying this. Is there not a simpler way?

SWvheerden: I think thats the simplest way. You are welcome to suggest a better one?
We only need to not sync to that peer on specific mode options on a truth table between remote pruning options, local pruning options.

stringhandler (commented):
Seems correct, but we need a simpler way to organize this logic

stringhandler (approved):

PR #5741, opened 2023-09-05
fix(comms+dht): mark peers as online inbound connection,join

sdbondi:
Description
---
- mark peers as online inbound connection and on join msg
- fix DHT peer query logic

Motivation and Context
---
Peers were not being marked as online whenever we have proof that they are online, specifically in the  inbound connection and join msg.

The peer query to exclude peers if we have attempted them but never succeeded was incorrect and filtered out most peers. This was changed to 

```rust
 // we have tried to connect to this peer, and we have never made a successful attempt at connection
if peer.last_connect_attempt().is_some() && peer.last_seen().is_none() {
    return false;
}
```

How Has This Been Tested?
---
Manually, node reconnects quickly to peers on startup

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/core/src/connectivity/manager.rs
+++ b/comms/core/src/connectivity/manager.rs
@@ -636,8 +636,6 @@ impl ConnectivityManagerActor {
             (_, Connected) => match self.pool.get_connection(&node_id).cloned() {
                 Some(conn) => {
                     self.mark_connection_success(conn.peer_node_id().clone());
-                    self.mark_peer_last_seen_now(&conn).await?;
-
                     self.publish_event(ConnectivityEvent::PeerConnected(conn.into()));
                 },
                 None => unreachable!(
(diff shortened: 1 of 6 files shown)


comms/core/src/connectivity/manager.rs
```diff
@@ -662,6 +664,19 @@ impl ConnectivityManagerActor {
         Ok(())
     }
 
+    async fn mark_peer_last_seen_now(&self, conn: &PeerConnection) -> Result<(), ConnectivityError> {
+        // Can't be None
+        if let Some(mut peer) = self.peer_manager.find_by_node_id(conn.peer_node_id()).await? {
+            if let Some(addr) = conn.known_address() {
+                peer.addresses.mark_last_seen_now(addr);
+            } else {
+                peer.addresses.mark_all_addresses_as_last_seen_now();
```
SWvheerden: I would prefer us only mark the address we have and we connected to as last seen true. 
If a peer has multiple addresses with only one if them true, we only need to mark that one as the address as last seen and not all.

sdbondi: Yup agree with that, let's update addresses given by the claims from these sources.

sdbondi: Pushed the change

SWvheerden (approved):

PR #5749, opened 2023-09-07
fix: correct epoch counting

Cifko:
Description
---
Fix the epoch counting.

Motivation and Context
---
When the epoch length changed it affected the history as well. E.g. when you have epoch length set to 10 and it's height 1000. then the epoch # is 100. But when there is new consensus constant from 1001 with epoch length 20. Suddenly the epoch is 50. 
Now it computes it like this:
When there is an constant change then it's effect is immediate but the epoch skip is prevented.
e.g.
1) Length is 15 and it's going to change on height 100 to 3. So on height 99 the epoch is 6, on 100 the epoch is 10 (because the current epoch is already of size 10).
2) Length is 15 and it's going to change on height 100 to 20. So on height 99 the epoch is 6, on 105 (without the change this would be the next epoch height, but with current new constant it's not changed). The epoch is changed on 110 (to 7).

Fixes https://github.com/tari-project/tari-dan/issues/832

How Has This Been Tested?
---
There is a new test `test_epoch_to_height_and_back`

---
`cargo.exe +stable test --package tari_core --lib -- consensus::consensus_manager::tests::test_epoch_to_height_and_back --exact --nocapture`

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify 
  This will break 2nd layer **ONLY** on chains that have new consensus constants.


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1400,24 +1400,21 @@ impl LMDBDatabase {
...
-        let prev_shard_key = store.get_shard_key(
-            current_epoch
-                .as_u64()
-                .saturating_sub(constants.validator_node_validity_period_epochs().as_u64()) *
-                constants.epoch_length(),
-            current_epoch.as_u64() * constants.epoch_length(),
-            vn_reg.public_key(),
-        )?;
+        let current_epoch = self.block_height_to_epoch(header.height);
+        // TODO: What if the validity period has changed?
```
```suggestion
```
If this is important enough create an issue, a todo randomly in code will just get deleted after a while anyway.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2450,7 +2455,7 @@ impl BlockchainBackend for LMDBDatabase {
         let constants = self.consensus_manager.consensus_constants(height);
 
         // Get the current epoch for the height
-        let end_epoch = constants.block_height_to_epoch(height);
+        let end_epoch = self.consensus_manager.block_height_to_epoch(height);
         // Subtract the registration validaty period to get the start epoch
         let start_epoch = end_epoch.saturating_sub(constants.validator_node_validity_period_epochs());
```
SWvheerden, line 2487: You have the constants above, why are you changing the one, but leaving the other one as be?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2466,7 +2471,7 @@ impl BlockchainBackend for LMDBDatabase {
         let constants = self.get_consensus_constants(height);
 
         // Get the epoch height boundaries for our query
-        let current_epoch = constants.block_height_to_epoch(height);
+        let current_epoch = self.consensus_manager.block_height_to_epoch(height);
         let start_epoch = current_epoch.saturating_sub(constants.validator_node_validity_period_epochs());
```
SWvheerden, line 2515: same here?

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -112,6 +113,51 @@ impl ConsensusManager {
         constants
     }
 
+    /// Returns the current epoch from the given height
```
SWvheerden:
```suggestion
    /// Returns the current epoch number as calculated from the given height
```

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -112,6 +113,51 @@ impl ConsensusManager {
...
+                (c.effective_from_height() - active_effective_height + leftover_height) % active_epoch_length,
+            );
+            active_effective_height = c.effective_from_height();
+            active_epoch_length = c.epoch_length();
+        }
+        epoch += (height - active_effective_height + leftover_height) / active_epoch_length;
+        VnEpoch(epoch)
+    }
+
+    /// Returns the block height of the start of the given epoch
```
SWvheerden:
```suggestion
    /// Returns the block height of the start of the given epoch number
```

SWvheerden (commented):
A few small nits

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
Cifko: Issue created.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
Cifko: Because `constants` represents the current consensus. But the end epoch can't be calculated from it. You need to know how the constants changed over time. `block_height_to_epoch` goes over all constants.
E.g. when we start the epoch will last 10 blocks, on height 100, we change it to 20 epoch. So on height 200 the epoch is 15. For this you need to know how it changed over time.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
Cifko: same as above

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -271,3 +317,64 @@ pub enum ConsensusBuilderError {
...
+        assert_eq!(manager.block_height_to_epoch(202), VnEpoch(23)); // But we have new length with size +2 so the epoch change will happen at 204
+        assert_eq!(manager.block_height_to_epoch(204), VnEpoch(24));
+        // Now test couple more back and forth
+        for epoch in 0..=100 {
+            assert_eq!(
+                manager.block_height_to_epoch(manager.epoch_to_block_height(VnEpoch(epoch))),
+                VnEpoch(epoch)
+            );
+        }
+    }
```
hansieodendaal, line 383: Maybe we can be more explicit in this test
```suggestion
        let mut epoch = manager.block_height_to_epoch(0).as_u64();
        for height in 0..600 {
            assert!(manager.block_height_to_epoch(height).as_u64() >= epoch);
            epoch = manager.block_height_to_epoch(height).as_u64();
        }
    }
```

base_layer/core/src/consensus/consensus_manager.rs
```diff
@@ -110,6 +111,51 @@ impl ConsensusManager {
         constants
     }
 
+    /// Returns the current epoch number as calculated from the given height
+    pub fn block_height_to_epoch(&self, height: u64) -> VnEpoch {
+        let mut epoch = 0;
+        let mut leftover_height = 0;
+        let mut active_effective_height = 0;
+        let mut active_epoch_length = self.inner.consensus_constants[0].epoch_length();
```
hansieodendaal, line 119: Maybe we should test if this value is `== 0` and return an error. It is possible to set `ConsensusConstants::vn_epoch_length = 0`

or 

add a test for it similar to 
```rust
#[cfg(any(test, debug_assertions))]
fn assert_hybrid_pow_constants(
```

hansieodendaal: This panics:
```rust
        let manager = ConsensusManager::builder(Network::LocalNet)
            .add_consensus_constants(
                ConsensusConstantsBuilder::new(Network::LocalNet)
                    .with_effective_height(0)
                    .with_vn_epoch_length(0)
                    .build(),
            )
            .build()
            .unwrap();
        println!("u64::MAX: {}", manager.block_height_to_epoch(1).as_u64());
```

hansieodendaal (commented):
Looks good (tested). Just some defence-in-depth comments.

base_layer/core/src/consensus/consensus_manager.rs
Cifko: I've created a new for the non decreasing `block_height_to_epoch`.

base_layer/core/src/consensus/consensus_manager.rs
Cifko: This should never happen. If we will be able to set it to zero then we can say good bye to the 2nd layer.

hansieodendaal: What if we update this function
```rust
    pub fn epoch_length(&self) -> u64 {
        self.vn_epoch_length
    }
```
as
```rust
    pub fn epoch_length(&self) -> u64 {
        #[cfg(any(test, debug_assertions))]
        assert_ne!(self.vn_epoch_length, 0, "Epoch length must be greater than 0");
        self.vn_epoch_length
    }
```
?

Cifko: added

hansieodendaal (approved):
utACK

SWvheerden:
This has gone stale, going to close it for now, wee can revisit this later on

PR #5752, opened 2023-09-07
feat(chatffi): tor configuration

brianp:
Description
---
Allow more configuration options for TOR

Motivation and Context
---
Mobile doesn't align with the tor transport config defaults that were being applied, so let's allow for a specific configuration.

How Has This Been Tested?
---
CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/chat_ffi/src/lib.rs
+++ b/base_layer/chat_ffi/src/lib.rs
@@ -895,3 +895,77 @@ pub unsafe extern "C" fn destroy_chat_ffi_liveness_data(address: *mut ChatFFICon
         drop(Box::from_raw(address))
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::ffi::CString;
+
+    use libc::c_char;
+
+    use crate::*;
+
+    #[test]
+    fn test_create_chat_tor_transport_config() {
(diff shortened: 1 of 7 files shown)


base_layer/chat_ffi/src/lib.rs
```diff
@@ -69,6 +71,9 @@ mod consts {
     include!(concat!(env!("OUT_DIR"), "/consts.rs"));
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct ByteVector(Vec<c_uchar>); // declared like this so that it can be exposed to external header
```
SWvheerden: if it needs exposing, then it needs rep_c as well?

SWvheerden (commented):
Just one question

base_layer/chat_ffi/src/lib.rs
brianp: I copied it verbatim from the wallet, which doesn't use rep_c for it.

Which is a roundabout way of saying I'm hesitant to "fix" something not broken :D

SWvheerden (approved):

PR #5753, opened 2023-09-07
docs: fix header sync diagrams

SWvheerden:
Description
---
updates and fixes the header sync diagrams


docs/src/diagrams/header_sync.md
```diff
@@ -87,71 +94,7 @@ flowchart TD
     S --> |No| U[Return Ok]
 
     N1:::note
+    META:::meta
    classDef note fill:#eee,stroke:#ccc
+   classDef meta fill:#b11,stroke:#ccc
 ```
-
-```mermaid
-flowchart TD
```
stringhandler, line 94: Did you mean to delete this?

SWvheerden: yes thats the old combined diagram, I added that to not loose it as the newer diagrams where done on dev at that point, and the combined was done using not yet merged PR's. 
Thats done now, and the newer diagrams are clearer and better, contains the same info, so no need for that diagram anymore.

stringhandler (approved):

PR #5756, opened 2023-09-08
fix: edge case about sync

SWvheerden:
Description
---
This fixes an edge case around sync where the local node has a higher chain header, but the remote node has a higher actual block chain height. 

Motivation and Context
---
When doing sync, and determining from where to sync, we need to look at the where out actual block chain is, especially when comparing chains. 
On the current code its possible that you node has a valid header chain of higher proof of work then the remote syncing peer, but the blocks are not valid for that chain, thus the remote chain has an actual higher proof of work block chain. 
The current code will ban the peer for lying about the pow, as the header sync uses the last chain header to determine pow, while the listing state will use the chain_meta_data to determine pow. 

We should look at out chain header tip when determining the fork header before starting header sync, but when we compare the chains, we need to look at what our actual pow chain is and not the tip chain header. 

How Has This Been Tested?
---
Manual. 

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -254,18 +254,19 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
             .fetch_chain_header(self.local_metadata.height_of_longest_chain())
             .await?;
         let local_total_accumulated_difficulty = local_blockchain_tip.accumulated_data().total_accumulated_difficulty;
-        let blockchain_tip_height = local_blockchain_tip.height();
+        let header_tip_height = local_tip_header.height();
+        let local_blockchain_tip_height = local_blockchain_tip.height();
         let sync_status = self
             .determine_sync_status(sync_peer, local_tip_header, local_blockchain_tip, &mut client)
             .await?;
         match sync_status {
             HeaderSyncStatus::InSync | HeaderSyncStatus::WereAhead => {
                 let metadata = self.db.get_chain_metadata().await?;
-                if metadata.height_of_longest_chain() < blockchain_tip_height {
+                if metadata.height_of_longest_chain() < header_tip_height {
                     debug!(
                         target: LOG_TARGET,
                         "Headers are in sync at height {} but tip is {}. Proceeding to archival/pruned block sync",
-                        blockchain_tip_height,
+                        header_tip_height,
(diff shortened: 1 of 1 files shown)


hansieodendaal (commented):
The revised code, although it may be correct, is really confusing. For example, there seems to be no difference in meaning between the words `local_tip_header` and `local_blockchain_tip_header`. When looking at the code, the significance of the change is also lost. For example, why would these be different?

`let local_tip_header = self.db.fetch_last_chain_header().await?;`
vs.
`let local_blockchain_tip_header = self.db.fetch_chain_header(self.local_metadata.height_of_longest_chain()).await?;`

`let header_tip_height = local_tip_header.height();`
vs.
`let header_tip_height = local_blockchain_tip_header.height();`

SWvheerden:
I am all ears if you have better names for them. 
but when syncing we have two important terms here: 
tip_header: the highest height header we have.
blockchain_tip_header: the height header we have a full block for, aka forms a blockchain. 

And the reason this is important is how the sync process works. 
1. We download headers only and add them to the main chain. 
2. We download the blocks for those headers and add them to the main chain. 

And it very much means that when entering header sync we have more headers than the current pee, but they have a larger block height than us. Which means we need to rewind our chain and sync to thiers.

stringhandler (changes requested):
I don't agree with this change. It feels incomplete. 

It is very confusing to pass both the highest header and the highest block into a method and then use one for some rules and one for the others.

I think it could lead to future bugs. 

Even if this does solve the problem, I would like a test to make sure that this is the case, so that this edge case is not forgotten about

SWvheerden:
The comments here point out the exact problem here, the sync code is stupidly complex. But it is all the edge cases vs. performance. 

We don't use one highest header for one rule, and the highest block for another,  we use both, always. 
The highest header we use to determine the split height. 
The highest block we use to determine pow. 


The much simpler approach would be to clear all the changes made after a failed header sync, but from previous discussions, the majority of the team is against that for performance reasons. And its valid, if you download most of the chain, and fail on connection, then you have to restart from scratch. So it is a valid objection, but for that to work, and be secure, we need complex code to solve this. As it stands, this opens up the node to a valid attack which will cause the node to be stuck forever, and ban the whole network.

SWvheerden:
closing this for: https://github.com/tari-project/tari/pull/5837
this code is contained in that PR + extra fixes

PR #5757, opened 2023-09-11
fix: remove peer

SWvheerden:
Description
---
Fixes remove peer in header sync
Fixes latency counter in block sync

Motivation and Context
---
Header sync should remove the peer for all errors
Block sync should only increase the latency once


hansieodendaal (approved):
utACK

PR #5758, opened 2023-09-11
feat: up the timeout for comms

SWvheerden:
Description
---
increase timeouts

Motivation and Context
---
We recently decreased timeouts across the communications. This increases it again somewhat after it was observed that the newer timeouts maybe too quick for slow tor connections


sdbondi (approved):
Maybe we should set these defaults based on the selected transport. But otherwise happy for this to go in.

PR #5759, opened 2023-09-11
fix: potential overflow

SWvheerden:
Description
---
Stops a potential overflow in a log

Motivation and Context
---
Overflow should not be possible


hansieodendaal (approved):
utACK

PR #5760, opened 2023-09-11
fix: potential sync stuck

SWvheerden:
Description
---
Fixes a potential case where sync process can get stuck

Motivation and Context
---
There exists a potential where latency keeps increasing forever and the local node will be stuck in sync forever. This adds in a max amount of times the local node will try again before exiting.

### DIFF
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -50,7 +50,7 @@ use crate::{
 
 const LOG_TARGET: &str = "c::bn::block_sync";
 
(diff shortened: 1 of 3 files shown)


base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -50,6 +50,8 @@ use crate::{
 
 const LOG_TARGET: &str = "c::bn::block_sync";
 
+const MAX_LATENCY_INCREASES: usize = 10;
```
hansieodendaal: **Nit**
Maybe these should be less (in all three cases), like 5 times?

hansieodendaal (approved):
utACK

PR #5763, opened 2023-09-11
feat!: improve block add where many orphan chain tips existed

hansieodendaal:
Description
---
Improved block add due to previous inefficient retrieval of the strongest orphan chain header.

The previous algo retrieved all orphan blocks for all orphan chain tips, without evaluating the total accumulated difficulty. This PR added total accumulated difficulty to the orphan chain tips db so that efficient retrieval of only the strongest orphan chain headers can be done.

Block add times improved drastically by multiple orders of magnitude in the case where many orphan chain tips existed as only one or very seldomly two orphan chain tips would have the same total accumulated difficulty, thereby minimizing the retrieval of orphan blocks to enable the construction of orphan chain headers.

These symptoms were aggravated when candidate blocks were solved quicker than the base node could process them, resulting in a negative spiral of adding even more orphan chain tips when corresponding slowing down of blocks that were added to the blockchain.

Motivation and Context
---
See above

How Has This Been Tested?
---
New unit test added
System level testing on Igor

---
Code walk through and system level testing

Breaking Changes
---

- [] None
- [ ] Requires hard fork
- [ ] Other - Please specify

BREAKING CHANGE: The blockchain database should be deleted and then re-synced from the network.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -892,113 +892,78 @@ where B: BlockchainBackend
             );
             return Err(ChainStorageError::AddBlockOperationLocked);
         }
-        {
-            let db = self.db_read_access()?;
-            if db.contains(&DbKey::BlockHash(block_hash))? {
-                return Ok(BlockAddResult::BlockExists);
-            }
-            if db.bad_block_exists(block_hash)? {
-                return Err(ChainStorageError::ValidationError {
-                    source: ValidationError::BadBlockFound {
-                        hash: block_hash.to_hex(),
-                    },
-                });
-            }
-        }
 
         let new_height = candidate_block.header.height;
-        let block_add_result = {
-            // This is important, we ask for a write lock to disable all read access to the db. The sync process sets
-            // the add_block disable flag,  but we can have a race condition between the two especially
-            // since the orphan validation can take some time during big blocks as it does Rangeproof and
-            // metadata signature validation. Because the sync process first acquires a read_lock then a
-            // write_lock, and the RWLock will be prioritised, the add_block write lock will be given out
-            // before the sync write_lock.
-            trace!(
-                target: LOG_TARGET,
-                "[add_block] waiting for write access to add block block #{} '{}'",
-                new_height,
-                block_hash.to_hex(),
-            );
+        // This is important, we ask for a write lock to disable all read access to the db. The sync process sets
+        // the add_block disable flag,  but we can have a race condition between the two especially
+        // since the orphan validation can take some time during big blocks as it does Rangeproof and
+        // metadata signature validation. Because the sync process first acquires a read_lock then a
+        // write_lock, and the RWLock will be prioritised, the add_block write lock will be given out
+        // before the sync write_lock.
+        trace!(
+            target: LOG_TARGET,
+            "[add_block] waiting for write access to add block block #{} '{}'",
+            new_height,
+            block_hash.to_hex(),
+        );
+        let before_lock = timer.elapsed();
+        let mut db = self.db_write_access()?;
+        let after_lock = timer.elapsed();
+        trace!(
+            target: LOG_TARGET,
+            "[add_block] acquired write access db lock for block #{} '{}' in {:.2?}",
+            new_height,
+            block_hash.to_hex(),
+            after_lock - before_lock,
+        );
 
-            let before_lock = timer.elapsed();
-            let mut db = self.db_write_access()?;
-            let after_lock = timer.elapsed();
+        if db.contains(&DbKey::BlockHash(block_hash))? {
+            return Ok(BlockAddResult::BlockExists);
+        }
+        if db.bad_block_exists(block_hash)? {
+            return Err(ChainStorageError::ValidationError {
+                source: ValidationError::BadBlockFound {
+                    hash: block_hash.to_hex(),
+                },
+            });
+        }
 
-            trace!(
-                target: LOG_TARGET,
-                "[add_block] acquired write access db lock for block #{} '{}' in {:.2?}",
-                new_height,
-                block_hash.to_hex(),
-                after_lock - before_lock,
-            );
-
-            // the only fast check we can perform that is slightly expensive to fake is a min difficulty check, this is
-            // done as soon as we receive the block before we do any processing on it. A proper proof of
-            // work is done as soon as we can link it to the main chain. Full block validation only happens
-            // when the proof of work is higher than the main chain and we want to add the block to the main
-            // chain.
-            let res = add_block(
-                &mut *db,
-                &self.config,
-                &self.consensus_manager,
-                &*self.validators.block,
-                &*self.validators.header,
-                self.consensus_manager.chain_strength_comparer(),
-                candidate_block,
-            )?;
-            debug!(
-                target: LOG_TARGET,
-                "[add_block] released write access db lock for block #{} in {:.2?}, `add_block` result: {}",
-                new_height, timer.elapsed() - after_lock, res
-            );
-            res
-        };
+        // the only fast check we can perform that is slightly expensive to fake is a min difficulty check, this is
+        // done as soon as we receive the block before we do any processing on it. A proper proof of
+        // work is done as soon as we can link it to the main chain. Full block validation only happens
+        // when the proof of work is higher than the main chain and we want to add the block to the main
+        // chain.
+        let block_add_result = add_block(
+            &mut *db,
+            &self.config,
+            &self.consensus_manager,
+            &*self.validators.block,
+            &*self.validators.header,
+            self.consensus_manager.chain_strength_comparer(),
+            candidate_block,
+        )?;
 
         // If blocks were added and the node is in pruned mode, perform pruning
         if block_add_result.was_chain_modified() {
-            let metadata = {
-                let db = self.db_read_access()?;
-                db.fetch_chain_metadata()?
-            };
             info!(
                 target: LOG_TARGET,
                 "Best chain is now at height: {}",
-                metadata.height_of_longest_chain()
+                db.fetch_chain_metadata()?.height_of_longest_chain()
             );
-
-            if metadata.is_pruned_node() {
-                let db_height = metadata.height_of_longest_chain();
-                let abs_pruning_horizon = db_height.saturating_sub(self.config.pruning_horizon);
-
-                debug!(
-                    target: LOG_TARGET,
-                    "Current pruned height is: {}, pruning horizon is: {}, while the pruning interval is: {}",
-                    metadata.pruned_height(),
-                    abs_pruning_horizon,
-                    self.config.pruning_interval,
-                );
-                if metadata.pruned_height() < abs_pruning_horizon.saturating_sub(self.config.pruning_interval) {
-                    let mut db = self.db_write_access()?;
-                    if metadata == db.fetch_chain_metadata()? {
-                        if let Err(e) = prune_to_height(&mut *db, abs_pruning_horizon) {
-                            warn!(target: LOG_TARGET, "Failed to prune database: {}", e);
-                        }
-                    } else {
-                        debug!(target: LOG_TARGET, "Database changed since we checked, not pruning this time round");
-                    }
-                }
-            }
+            // If blocks were added and the node is in pruned mode, perform pruning
+            prune_database_if_needed(&mut *db, self.config.pruning_horizon, self.config.pruning_interval)?;
         }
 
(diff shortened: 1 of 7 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -880,85 +880,125 @@ where B: BlockchainBackend
...
+                return Ok(BlockAddResult::BlockExists);
+            }
+            if db.bad_block_exists(block_hash)? {
+                return Err(ChainStorageError::ValidationError {
+                    source: ValidationError::BadBlockFound {
+                        hash: block_hash.to_hex(),
+                    },
+                });
+            }
+        }
```
```suggestion

```
We already check this before calling add_block. And if it somehow does not check it before, the validator will also check this, so this is just an unnecessary read lock.

SWvheerden: If we want to keep this check, move it till after the write lock

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2380,33 +2438,6 @@ fn cleanup_orphans<T: BlockchainBackend>(db: &mut T, orphan_storage_capacity: us
     db.delete_oldest_orphans(horizon_height, orphan_storage_capacity)
 }
 
-fn prune_database_if_needed<T: BlockchainBackend>(
```
SWvheerden: I prefer us keeping the logic here

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -880,85 +880,125 @@ where B: BlockchainBackend
...
-                db.fetch_chain_metadata()?.height_of_longest_chain()
+                metadata.height_of_longest_chain()
             );
-            // If blocks were added and the node is in pruned mode, perform pruning
-            prune_database_if_needed(&mut *db, self.config.pruning_horizon, self.config.pruning_interval)?;
-        }
 
-        if let Err(e) = cleanup_orphans(&mut *db, self.config.orphan_storage_capacity) {
-            warn!(target: LOG_TARGET, "Failed to clean up orphans: {}", e);
+            if metadata.is_pruned_node() {
```
SWvheerden: I prefer keeping the logic in a function, its easier to read

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1120,14 +1124,23 @@ impl LMDBDatabase {
...
-                    parent_hash.as_slice(),
-                    &parent_hash,
-                    "orphan_chain_tips_db",
-                )?;
+            if lmdb_exists(txn, &self.orphans_db, parent_hash.as_slice())? &&
+                lmdb_exists(txn, &self.orphan_header_accumulated_data_db, parent_hash.as_slice())?
+            {
+                let orphan_parent_accum: Option<BlockHeaderAccumulatedData> =
+                    lmdb_get(txn, &self.orphan_header_accumulated_data_db, parent_hash.as_slice())?;
+                if let Some(val) = orphan_parent_accum {
```
stringhandler, line 1132: Perhaps a warning for the else case

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1120,14 +1124,23 @@ impl LMDBDatabase {
...
             // Parent becomes a tip hash
-            if lmdb_exists(txn, &self.orphans_db, parent_hash.as_slice())? {
-                lmdb_insert(
-                    txn,
-                    &self.orphan_chain_tips_db,
-                    parent_hash.as_slice(),
-                    &parent_hash,
-                    "orphan_chain_tips_db",
-                )?;
+            if lmdb_exists(txn, &self.orphans_db, parent_hash.as_slice())? &&
```
stringhandler, line 1127: also maybe a warning in the else case

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1891,8 +1931,21 @@ fn handle_possible_reorg<T: BlockchainBackend>(
     chain_strength_comparer: &dyn ChainStrengthComparer,
     candidate_block: Arc<Block>,
 ) -> Result<BlockAddResult, ChainStorageError> {
```
stringhandler, line 1898: Maybe not related to this PR, but BlockAddREsult doesn't make sense here to return. For example, we return OrphanBlock when we are on the strongest chain

SWvheerden (changes requested):
Having read-write-read-write locks is going to be worse for performance than just having a single write lock. 

While you can have multiple read locks, as soon as you ask for a single write lock, you hold on all new locks and then continue one with that single write lock after all reads are done. The read lock instructions are soo fast its not a real problem to just use the write logic. 
The code is now much more complex for I would argue slower performance, as we specifically care about single block add performance, this will not be slower on average as you pause processing of the block to await locks.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1966,22 +2019,23 @@ fn swap_to_highest_pow_chain<T: BlockchainBackend>(
     // rewind to height will first delete the headers, then try delete from blocks, if we call this to the current
     // height it will only trim the extra headers with no blocks
     rewind_to_height(db, metadata.height_of_longest_chain())?;
-    let all_orphan_tips = db.fetch_all_orphan_chain_tips()?;
-    if all_orphan_tips.is_empty() {
+    let strongest_orphan_tips = db.fetch_strongest_orphan_chain_tips()?;
+    if strongest_orphan_tips.is_empty() {
         // we have no orphan chain tips, we have trimmed remaining headers, we are on the best tip we have, so lets
         // return ok
         return Ok(BlockAddResult::OrphanBlock);
```
stringhandler, line 1991: See previous comment, but this is a strange thing to return IMO

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -880,85 +880,125 @@ where B: BlockchainBackend
...
-            &mut *db,
-            &self.config,
-            &self.consensus_manager,
-            &*self.validators.block,
-            &*self.validators.header,
-            self.consensus_manager.chain_strength_comparer(),
-            candidate_block,
-        )?;
+            let before_lock = timer.elapsed();
+            let mut db = self.db_write_access()?;
```
stringhandler: You need to recheck for bad block and block exists here after getting the write lock

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -880,85 +880,125 @@ where B: BlockchainBackend
...
+                debug!(
+                    target: LOG_TARGET,
+                    "Current pruned height is: {}, pruning horizon is: {}, while the pruning interval is: {}",
+                    metadata.pruned_height(),
+                    abs_pruning_horizon,
+                    self.config.pruning_interval,
+                );
+                if metadata.pruned_height() < abs_pruning_horizon.saturating_sub(self.config.pruning_interval) {
+                    let mut db = self.db_write_access()?;
+                    if metadata == db.fetch_chain_metadata()? {
```
```suggestion
                   // Check that there haven't been other changes to the chain while we were getting the write lock
                    if metadata == db.fetch_chain_metadata()? {
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -880,85 +880,125 @@ where B: BlockchainBackend
...
             warn!(
                 target: LOG_TARGET,
                 "add_block is disabled, node busy syncing. Ignoring candidate block #{} ({})",
                 candidate_block.header.height,
-                candidate_block.hash(),
+                block_hash,
             );
             return Err(ChainStorageError::AddBlockOperationLocked);
         }
+        {
```
stringhandler: Perhaps we should set `is_add_block_disabled` at this point?

I know the intention here was to reduce the amount of time that we spend with the write lock, but the danger here is that we switch to header sync or block sync in between each of these write locks.

I would actually suggest putting the write locks back across all operations and keep the other improvements in this PR

stringhandler (changes requested):

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: I think its the correct thing to return. 
If all the orphans tips are empty. It means the block you added is an orphan, hence the result here. 
New blocks are first added as an orphan, and then from there, an orphan chain tip is created if you can link it to the main chain. Then do swap to the highest chain. So any new block will always be an orphan chain tip first.

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: If we want to do it, we should do it in another PR not this one. 
But this should not be required  because, header sync sets `is_add_block_disabled` and is the only one to set it, 
Block sync and header sync can disable it, But they will wait for the write lock here.

The only reason we have this here, and not just use the write lock, is because headersync and block sync has its own write lock each

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Reverting back to the previous version, logic is unchanged

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Reverting back to the previous version, logic is unchanged

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Reverting back to the previous version, logic is unchanged

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Reverting back to the previous version, logic is unchanged

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Reverting back to the previous version, logic is unchanged

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: I agree with @SWvheerden on this

SWvheerden (approved):

base_layer/core/src/chain_storage/blockchain_database.rs
stringhandler: I don't think you understand my point. The method name "swap_to_highest_chain" suggests it can be called at any time, but the value returned is tightly coupled to the caller context.... i.e. it is returning a critical result that the caller "handle_possible_reorg" requires. this is a private method, so it's probably fine, but it's a code smell that could potentially cause a bug in future.

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -552,6 +552,7 @@ mod test {
         };
         let hash = block_header.merge_mining_hash();
         append_merge_mining_tag(&mut block, hash).unwrap();
+        #[allow(clippy::redundant_clone)]
         let mut block_header2 = block_header.clone();
```
stringhandler, line 556:
```suggestion
        let mut block_header2 = block_header;
```

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -552,6 +552,7 @@ mod test {
         };
         let hash = block_header.merge_mining_hash();
         append_merge_mining_tag(&mut block, hash).unwrap();
+        #[allow(clippy::redundant_clone)]
```
stringhandler, line 555:
```suggestion
    
```

stringhandler (approved):

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: Issue created: #5765

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
hansieodendaal: The clone and `#allow` is to satisfy `cargo clippy --all-targets --all-features -- -D warnings`

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
hansieodendaal: ^^

PR #5768, opened 2023-09-14
fix: remove statement from sparse Merkle tree proofs

### LINKED ISSUE #5527: Sparse Merkle tree proofs include statement

SWvheerden:
Description
---
Removes the key and value hashes from sparse Merkle tree proofs. Adds a path check to inclusion proof verification. Closes #5527.

Motivation and Context
---
Currently, sparse Merkle tree proofs [include](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L23-L24) the key and value hashes for the proof. As these are effectively the proof statement, it doesn't really make sense to include them in the proof structure.

Including them means additional work for the verifier for both [inclusion](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L75-L76) and [exclusion](https://github.com/tari-project/tari/blob/87c070305951152c62a0179e13fadc55065cc318/base_layer/mmr/src/sparse_merkle_tree/proofs.rs#L83) proofs that isn't needed. This design also seems like a footgun that could lead an implementer to unintentionally trust the statement from a malicious prover.

This PR removes the key and value hashes from the proof structure altogether, and updates relevant functions accordingly. It also adds a path check to the inclusion proof verifier for consistency with the exclusion proof verifier, which may also help with faster detection of invalid proofs.

How Has This Been Tested?
---
Existing tests pass.

---
Check that the new verification logic is correct.

Breaking Changes
---
None. While this changes the structure of sparse Merkle tree proofs, they are not currently used and do not have serialization defined.


CjS77 (approved):

base_layer/mmr/src/sparse_merkle_tree/node.rs
```diff
@@ -110,7 +109,72 @@ hash_type!(NodeHash);
...
+        if self.cursor_front >= self.cursor_back {
+            return None;
+        }
+        self.cursor_back -= 1;
+        let bit = get_bit(self.key.as_slice(), self.cursor_back);
+        Some(bit_to_dir(bit))
+    }
+
+    fn nth_back(&mut self, n: usize) -> Option<Self::Item> {
+        self.cursor_back -= n;
```
sdbondi, line 168: Could underflow. Should check for this case and return None.

CjS77: Good catch

PR #5769, opened 2023-09-14
fix: handle possible underflow in smt

CjS77:
A few unchecked subtractions are replaced with checked alternatives.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
LGTM

PR #5770, opened 2023-09-14
fix: unwraps in rpc client

SWvheerden:
Description
---
Removes unwraps in the rpc client code

Motivation and Context
---
There should not be unwraps, and they should propagate errors on invalid data


sdbondi (dismissed):
Sheesh, introduced in PR #3961 - good catch

sdbondi (approved):

PR #5772, opened 2023-09-14
fix: update libs to fix cargo audit

stringhandler:
Description
---
Updates a few crates that are failing cargo audit. 
Notably: trust_dns, clap

Motivation and Context
---
This allows us to remove the exceptions in cargo audit, but also to prevent them from triggering cargo-audit failures downstream, for example in the tari dan project

How Has This Been Tested?
---
clippy and cargo audit

---
Run cargo audit manually

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_console_wallet/src/cli.rs
+++ b/applications/minotari_console_wallet/src/cli.rs
@@ -63,13 +63,13 @@ pub struct Cli {
     #[clap(long, alias = "seed-words")]
     pub seed_words: Option<SeedWords>,
     /// Supply the optional file name to save the wallet seed words into
-    #[clap(long, aliases = &["seed_words_file_name", "seed-words-file"], parse(from_os_str))]
+    #[clap(long, aliases = &["seed_words_file_name", "seed-words-file"])]
     pub seed_words_file_name: Option<PathBuf>,
     /// Run in non-interactive mode, with no UI.
     #[clap(short, long, alias = "non-interactive")]
     pub non_interactive_mode: bool,
     /// Path to input file of commands
-    #[clap(short, long, aliases = &["input", "script"], parse(from_os_str))]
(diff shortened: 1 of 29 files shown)


SWvheerden (dismissed):

AaronFeickert (changes requested):
Looks like several remaining issues that CI caught, and which I also saw locally:
- the `clap` API for `App` has been replaced with `Command`
- the `clap` API for `Arg` has changed
- the `clap` API for `ArgMatches` has changed
- `DnsResponse::from_message` should be used to parse a response
- `env_logger::from_env` is deprecated in favor of `env_logger::Builder::from_env`

AaronFeickert:
Fixes are available on [this commit](https://github.com/AaronFeickert/tari/commit/3f599e1b2128c9addcec26d110102741c8cc979b).

applications/minotari_console_wallet/src/cli.rs
```diff
@@ -163,11 +163,11 @@ pub struct MakeItRainArgs {
     pub start_amount: MicroMinotari,
     #[clap(short, long, alias = "tps", default_value_t = 25)]
     pub transactions_per_second: u32,
-    #[clap(short, long, parse(try_from_str = parse_duration), default_value="60")]
+    #[clap(short, long, value_parser= parse_duration, default_value = "60")]
     pub duration: Duration,
     #[clap(long, default_value_t=tari_amount::T)]
     pub increase_amount: MicroMinotari,
-    #[clap(long, parse(try_from_str=parse_start_time))]
```
AaronFeickert, line 170: Removing the parser fails testing, as `now` is no longer a valid value. Either the parser should be added back, or the offending tests should be modified accordingly.

AaronFeickert (changes requested):

PR #5773, opened 2023-09-14
fix: change truncate_from_bits to from_bits

SWvheerden:
Description
---
This removes all occurrences of `truncate_from_bits` to `from_bits`

Motivation and Context
---
`truncate_from_bits` will truncate all unknown bits and may cause bits to be interpreted as the wrong flag. This changes to the much more strict `from_bits` which forces peers to use the correct bits and only the correct bits.

### DIFF
--- a/comms/core/src/peer_manager/error.rs
+++ b/comms/core/src/peer_manager/error.rs
@@ -53,6 +53,8 @@ pub enum PeerManagerError {
     InvalidPeerFeatures { bits: u32 },
     #[error("Address {address} not found for peer {node_id}")]
     AddressNotFoundError { address: Multiaddr, node_id: NodeId },
+    #[error("Protocol error: {0}")]
+    ProtocolError(String),
 }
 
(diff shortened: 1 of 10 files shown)


sdbondi (commented):
utACK but since RpcFlags are not part of blockchain data this does limit some upgrade paths. Say we wanted to add a new optional flag to RPC that e.g. enables an optimisation for newer nodes, we could do that without breaking existing nodes. With this change we'll need to wait until sufficient have upgraded.

I can't think of a risk of truncating here given that this is a message flag for an ephemeral/non-persistent/not-hashed message.

stringhandler:
Yeah we may revert this when we add new flags, but I think for now let's err on the side of caution and use `from_bits`

hansieodendaal (approved):
utACK

AaronFeickert (approved):
utACK

sdbondi (approved):

PR #5774, opened 2023-09-15
refactor: reorganize chatffi

brianp:
Description
---
Just restructuring the chat ffi into separate files to make it easier to grok and find what you need.

Motivation and Context
---
1000+ line files get heavy on the brain.

How Has This Been Tested?
---
CI

---
- Does it build?
- Do you like the naming?

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/chat_ffi/Cargo.toml
```diff
@@ -11,7 +11,6 @@ minotari_app_utilities = { path = "../../applications/minotari_app_utilities" }
 tari_chat_client = { path = "../contacts/examples/chat_client" }
 tari_common = { path = "../../common" }
 tari_common_types = { path = "../common_types" }
-tari_comms = { path = "../../comms/core" }
```
SWvheerden, line 14: this correct?

brianp: I was surprised to see this as well. Machete caught it this morning (not yesterday), so it may have to do with the last PR. Everything compiles though. And nothing in comms was in the build.rs.

brianp: Could also be when I shuffled around the imports I selected an import from an existing lib that was re-exporting whatever was used from comms. The chat client lib does re-export a few things and I know I changed an import or two.

PR #5776, opened 2023-09-15
feat: add insert function to SMT

SWvheerden:
Description
---
Adds a new insert function to only add a leaf if it already exists. 

Motivation and Context
---
If this is used for an unspent UTXO set we want it to disallow duplicate UTXOs 

How Has This Been Tested?
---
unit tests

### DIFF
--- a/base_layer/mmr/src/sparse_merkle_tree/tree.rs
+++ b/base_layer/mmr/src/sparse_merkle_tree/tree.rs
@@ -22,6 +22,7 @@ use crate::sparse_merkle_tree::{
 pub enum UpdateResult {
     Updated(ValueHash),
     Inserted,
+    AlreadyExistingKey,
 }
 
 #[derive(Debug, PartialEq)]
@@ -251,6 +252,15 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
             self.size += 1
         }
         Ok(result)
+    }
+
+    /// This will only add new node when it does not exist
+    pub fn insert(&mut self, key: NodeKey, value: ValueHash) -> Result<UpdateResult, SMTError> {
+        if self.get(&key)?.is_some() {
+            return Ok(UpdateResult::AlreadyExistingKey);
+        }
+        // So we no know it does not exist, so lets add it.
+        self.upsert(key, value)
     }
 
     /// Returns true if the tree contains the key `key`.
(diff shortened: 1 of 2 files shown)


base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -859,4 +869,50 @@ mod test {
...
+        assert_eq!(
+            tree.insert(short_key(1), ValueHash::from([1u8; 32])).unwrap(),
+            UpdateResult::Inserted
+        );
+        assert!(!tree.contains(&short_key(0)));
+        assert!(tree.contains(&short_key(1)));
+        assert_eq!(
+            tree.insert(short_key(1), ValueHash::from([1u8; 32])).unwrap(),
+            UpdateResult::AlreadyExistingKey
+        );
```
AaronFeickert:
```suggestion
        assert_eq!(
            tree.insert(short_key(1), ValueHash::from([2u8; 32])).unwrap(),
            UpdateResult::AlreadyExistingKey
        );
        assert_eq!(tree.get(&short_key(1)).unwrap().unwrap(), &ValueHash::from([1u8; 32]));
```
This asserts that the value is in fact unchanged, and that an upsert did not occur.

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -859,4 +869,50 @@ mod test {
...
+        assert!(tree.contains(&short_key(0)));
+        assert!(tree.contains(&short_key(1)));
+        assert_eq!(
+            tree.insert(short_key(0), ValueHash::from([0u8; 32])).unwrap(),
+            UpdateResult::AlreadyExistingKey
+        );
+        assert_eq!(
+            tree.insert(short_key(1), ValueHash::from([1u8; 32])).unwrap(),
+            UpdateResult::AlreadyExistingKey
+        );
```
AaronFeickert:
```suggestion
        assert_eq!(
            tree.insert(short_key(0), ValueHash::from([1u8; 32])).unwrap(),
            UpdateResult::AlreadyExistingKey
        );
        assert_eq!(tree.get(&short_key(0)).unwrap().unwrap(), &ValueHash::from([0u8; 32]));
        assert_eq!(
            tree.insert(short_key(1), ValueHash::from([2u8; 32])).unwrap(),
            UpdateResult::AlreadyExistingKey
        );
        assert_eq!(tree.get(&short_key(1)).unwrap().unwrap(), &ValueHash::from([1u8; 32]));
```

AaronFeickert (changes requested):

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -253,6 +254,15 @@ impl<H: Digest<OutputSize = U32>> SparseMerkleTree<H> {
         Ok(result)
     }
 
+    /// This will only add new node when it does not exist
+    pub fn insert(&mut self, key: NodeKey, value: ValueHash) -> Result<UpdateResult, SMTError> {
+        if self.get(&key)?.is_some() {
+            return Ok(UpdateResult::AlreadyExistingKey);
```
CjS77: You want this to be an error, I feel.
As written, matching for success becomes more complicated:
```
  match tree.insert(foo) {
    Ok(AlreadyExistingKey) => fail(),
    Ok(Updated(_)) => success(),
    Err(e) => fail(),
}
```

vs

```
  match tree.insert(foo) {
    Ok(_) => successl(),
    Err(_) => fail(),
}
```

base_layer/mmr/src/sparse_merkle_tree/tree.rs
```diff
@@ -22,6 +22,7 @@ use crate::sparse_merkle_tree::{
 pub enum UpdateResult {
     Updated(ValueHash),
     Inserted,
+    AlreadyExistingKey,
```
CjS77: `KeyAlreadyExists` or just `KeyExists`; and I'd move this to SMTError as well (see next comment)

CjS77:
Concept ACK

base_layer/mmr/src/sparse_merkle_tree/tree.rs
SWvheerden: I was debating error vs UpdateResult. 
I do agree that it makes matching on insert more difficult. The reason that made me go this way was the fact that we already have an UpdateResult struct here. But yeah its probably better to make it a hard error

CjS77: Indeed, the enum is to differentiate between the different success modes. If you don't care why an (up)sert succeeded you should just be able to check against `.is_ok()`

base_layer/mmr/src/sparse_merkle_tree/tree.rs
SWvheerden: changed to error type

base_layer/mmr/src/sparse_merkle_tree/tree.rs
SWvheerden: agreed

CjS77 (approved):
LGTM

AaronFeickert (approved):

PR #5777, opened 2023-09-18
chore: move mempool fee check

SWvheerden:
Description
---
Moves the mempool fee check to before validation

Motivation and Context
---
This is a very cheap check compared to full validation. We should do this before we check if the tx is valid so that we don't do validation and then through away the tx because the fee is too low.

### DIFF
--- a/base_layer/core/src/mempool/mempool_storage.rs
+++ b/base_layer/core/src/mempool/mempool_storage.rs
@@ -77,6 +77,11 @@ impl MempoolStorage {
             .map(|k| k.excess_sig.get_signature().to_hex())
             .unwrap_or_else(|| "None?!".into());
         let timer = Instant::now();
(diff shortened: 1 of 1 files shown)


base_layer/core/src/mempool/mempool_storage.rs
```diff
@@ -88,10 +93,6 @@ impl MempoolStorage {
                 );
                 let timer = Instant::now();
```
sdbondi, line 94: Unrelated: It looks like this timer shouldnt be here - there is a timer above which includes validation time.

sdbondi (approved):
utACK

base_layer/core/src/mempool/mempool_storage.rs
SWvheerden: No this is correct, its a new timer. 
Just above this, we print the validation results. This timer is just for the db insertion.

PR #5778, opened 2023-09-18
fix: potential overflow

SWvheerden:
Description
---
Fixes potential overflow from fee calculation

Motivation and Context
---
The transaction comes from remove peers, if the fee is u64::Max this can overflow and panic the node.

### DIFF
--- a/base_layer/core/src/mempool/priority/prioritized_transaction.rs
+++ b/base_layer/core/src/mempool/priority/prioritized_transaction.rs
@@ -42,11 +42,7 @@ pub struct FeePriority(Vec<u8>);
 
 impl FeePriority {
     pub fn new(transaction: &Transaction, insert_epoch: u64, weight: u64) -> Result<Self, TransactionError> {
-        // The weights have been normalised, so the fee priority is now equal to the fee per gram ± a few pct points
-        // Include 3 decimal places before flooring
-        #[allow(clippy::cast_possible_truncation)]
-        #[allow(clippy::cast_sign_loss)]
-        let fee_per_byte = ((transaction.body.get_total_fee()?.as_u64() as f64 / weight as f64) * 1000.0) as u64;
+        let fee_per_byte = transaction.body.get_total_fee()?.as_u64().saturating_mul(1000) / weight;
         // Big-endian used here, the MSB is in the starting index. The ordering for Vec<u8> is taken from elements left
         // to right and the unconfirmed pool expects the lowest priority to be sorted lowest to highest in the
         // BTreeMap
@@ -99,7 +95,7 @@ impl PrioritizedTransaction {
         Ok(Self {
             key,
             priority: FeePriority::new(&transaction, insert_epoch, weight)?,
-            fee_per_byte: ((transaction.body.get_total_fee()? * 1000) / weight).as_u64(),
+            fee_per_byte: transaction.body.get_total_fee()?.as_u64().saturating_mul(1000) / weight,
             weight,
             transaction,
             dependent_output_hashes: dependent_outputs.unwrap_or_default(),
--- a/base_layer/core/src/mempool/priority/prioritized_transaction.rs
+++ b/base_layer/core/src/mempool/priority/prioritized_transaction.rs
@@ -29,7 +29,10 @@ use std::{
 use tari_common_types::types::{HashOutput, PrivateKey, PublicKey};
 use tari_utilities::{hex::Hex, ByteArray};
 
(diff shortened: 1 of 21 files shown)


base_layer/core/src/mempool/priority/prioritized_transaction.rs
```diff
@@ -38,12 +41,12 @@ use crate::transactions::{transaction_components::Transaction, weight::Transacti
...
 
 impl FeePriority {
-    pub fn new(transaction: &Transaction, insert_epoch: u64, weight: u64) -> Self {
+    pub fn new(transaction: &Transaction, insert_epoch: u64, weight: u64) -> Result<Self, TransactionError> {
         // The weights have been normalised, so the fee priority is now equal to the fee per gram ± a few pct points
         // Include 3 decimal places before flooring
         #[allow(clippy::cast_possible_truncation)]
         #[allow(clippy::cast_sign_loss)]
-        let fee_per_byte = ((transaction.body.get_total_fee().as_u64() as f64 / weight as f64) * 1000.0) as u64;
+        let fee_per_byte = ((transaction.body.get_total_fee()?.as_u64() as f64 / weight as f64) * 1000.0) as u64;
```
stringhandler: since we are multiplying by 10000 anyway, why can't we remove the conversion to f64?

stringhandler: also, can't the *1000 cause an overflow here?

base_layer/core/src/mempool/priority/prioritized_transaction.rs
```diff
@@ -87,16 +90,16 @@ impl PrioritizedTransaction {
...
         let insert_epoch = match SystemTime::now().duration_since(UNIX_EPOCH) {
             Ok(n) => n.as_secs(),
             Err(_) => 0,
         };
         Ok(Self {
             key,
-            priority: FeePriority::new(&transaction, insert_epoch, weight),
-            fee_per_byte: ((transaction.body.get_total_fee() * 1000) / weight).as_u64(),
+            priority: FeePriority::new(&transaction, insert_epoch, weight)?,
+            fee_per_byte: ((transaction.body.get_total_fee()? * 1000) / weight).as_u64(),
```
stringhandler: same here, `* 1000`

base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
```diff
@@ -393,13 +398,14 @@ impl UnconfirmedPool {
                 {
                     update_candidate_transactions_to_select.remove(&tx_key);
                     *update_total_transaction_weight -= prioritized_transaction.weight;
-                    *update_total_transaction_fees -= prioritized_transaction.transaction.body.get_total_fee().0;
+                    *update_total_transaction_fees -= prioritized_transaction.transaction.body.get_total_fee()?.0;
```
stringhandler: Needs an underflow check or a comment saying why it can't underflow. (lines 400 and 401)

base_layer/core/src/mempool/unconfirmed_pool/unconfirmed_pool.rs
```diff
@@ -732,7 +738,7 @@ impl UnconfirmedPool {
                     break;
                 }
 
-                let total_tx_fee = tx.transaction.body.get_total_fee();
+                let total_tx_fee = tx.transaction.body.get_total_fee()?;
```
stringhandler, line 798: Seems to still be a number of unchecked additions in this file (e.g. line 737)

stringhandler (commented):
While this PR is only changing a specific addition, there are a lot of the files that are touched that have similar overflow problems. Specifically, in the mempool there are a number of these unchecked additions. I would suggest rather fix all the additions/multiplies in the mempool in this PR, otherwise, it's difficult to approve when there are glaring potential overflows

base_layer/core/src/mempool/priority/prioritized_transaction.rs
SWvheerden: overflow is not possible as total fee is a u64 converted to f64
u64::max = 1.8 E^19
f64::max = 1.79 E^308

So there are 2 options are, keep as is with f64
```
((transaction.body.get_total_fee().as_u64() as f64 / weight as f64) * 1000.0) as u64
```
Or do
```
((transaction.body.get_total_fee().as_u64() as u128 *1000) /weight) as u64
```

stringhandler (approved):

PR #5784, opened 2023-09-19
fix: possible exception in request_context

hansieodendaal:
Description
---
Fixed possible exception in `pub struct RequestContext` that could occur when splitting the request and reply channel in `pub fn split`

Motivation and Context
---
The code should not be allowed to panic in use in the wrong context.

How Has This Been Tested?
---
Unit tests

---
Code walk through

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (dismissed):

sdbondi (dismissed):
utACK

PR #5790, opened 2023-09-20
feat: drop watch borrow after use

hansieodendaal:
Description
---
Ensured watch borrows are dropped immediately after use

Motivation and Context
---
From the tokio documentation:
```
The priority policy of the lock is dependent on the underlying lock implementation, and this type does not 
guarantee that any particular policy will be used. In particular, a producer which is waiting to acquire the 
lock in send might or might not block concurrent calls to borrow, e.g.:

  Potential deadlock example 

// Task 1 (on thread A)    |  // Task 2 (on thread B)
let _ref1 = rx.borrow();   |
                           |  // will block
                           |  let _ = tx.send(());
// may deadlock            |
let _ref2 = rx.borrow();   |
```

How Has This Been Tested?
---
Unt tests

---
Code walk through

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_console_wallet/src/ui/components/burn_tab.rs
+++ b/applications/minotari_console_wallet/src/ui/components/burn_tab.rs
@@ -497,29 +497,22 @@ impl<B: Backend> Component<B> for BurnTab {
 
         let rx_option = self.burn_result_watch.take();
         if let Some(rx) = rx_option {
-            {
-                // Ensure the watch borrow is dropped immediately after use
-                trace!(target: LOG_TARGET, "{:?}", (*rx.borrow()).clone());
-            }
-            let status = {
-                // Ensure the watch borrow is dropped immediately after use
-                match (*rx.borrow()).clone() {
-                    UiTransactionBurnStatus::Initiated => "Initiated",
-                    UiTransactionBurnStatus::Error(e) => {
-                        self.error_message =
-                            Some(format!("Error sending transaction: {}, Press Enter to continue.", e));
-                        return;
-                    },
-                    UiTransactionBurnStatus::TransactionComplete((
-                        _proof_id,
-                        _reciprocal_claim_public_key,
-                        _serialized_proof,
-                    )) => {
-                        self.success_message =
-                            Some("Transaction completed successfully!\nPlease press Enter to continue".to_string());
-                        return;
-                    },
-                }
+            trace!(target: LOG_TARGET, "{:?}", *rx.borrow());
+            let status = match (*rx.borrow()).clone() {
+                UiTransactionBurnStatus::Initiated => "Initiated",
+                UiTransactionBurnStatus::Error(e) => {
+                    self.error_message = Some(format!("Error sending transaction: {}, Press Enter to continue.", e));
+                    return;
+                },
+                UiTransactionBurnStatus::TransactionComplete((
+                    _proof_id,
+                    _reciprocal_claim_public_key,
+                    _serialized_proof,
+                )) => {
+                    self.success_message =
+                        Some("Transaction completed successfully!\nPlease press Enter to continue".to_string());
+                    return;
+                },
             };
             draw_dialog(
                 f,
--- a/applications/minotari_console_wallet/src/ui/components/burn_tab.rs
+++ b/applications/minotari_console_wallet/src/ui/components/burn_tab.rs
@@ -497,22 +497,29 @@ impl<B: Backend> Component<B> for BurnTab {
 
         let rx_option = self.burn_result_watch.take();
         if let Some(rx) = rx_option {
-            trace!(target: LOG_TARGET, "{:?}", (*rx.borrow()).clone());
-            let status = match (*rx.borrow()).clone() {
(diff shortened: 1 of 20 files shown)


applications/minotari_console_wallet/src/ui/components/burn_tab.rs
```diff
@@ -497,22 +497,29 @@ impl<B: Backend> Component<B> for BurnTab {
...
-                    _reciprocal_claim_public_key,
-                    _serialized_proof,
-                )) => {
-                    self.success_message =
-                        Some("Transaction completed successfully!\nPlease press Enter to continue".to_string());
-                    return;
-                },
+            {
+                // Ensure the watch borrow is dropped immediately after use
+                trace!(target: LOG_TARGET, "{:?}", (*rx.borrow()).clone());
```
sdbondi: This is dropping the borrow after this line ends (the watch::Ref isn't put a variable). Also if you look at the macro expansion of trace! debug! etc it already scopes everything.

```rust
{
        let lvl = ::log::Level::Trace;
        if lvl <= ::log::STATIC_MAX_LEVEL && lvl <= ::log::max_level() { // <---- start scope
            ::log::__private_api::log(
                ::log::__private_api::format_args!("{:?}",
                                                   (*rx.borrow()).clone()
                ),
                lvl,
                &(LOG_TARGET, ::log::__private_api::module_path!( ), ::log::__private_api::file!( )),
                ::log::__private_api::line!( ),
                ::log::__private_api::Option::None,
            );
        }
    } // <---- end scope
```

Also shouldn't need to clone the value since we're just logging the value
```suggestion
                trace!(target: LOG_TARGET, "{:?}", rx.borrow()));
```

hansieodendaal: Yeah, I did not look inside the macro...

applications/minotari_console_wallet/src/ui/components/register_template_tab.rs
```diff
@@ -638,29 +638,36 @@ impl<B: Backend> Component<B> for RegisterTemplateTab {
...
-                    warn!("unhandled transaction status {:?}", status);
-                    return;
-                },
+            {
+                // Ensure the watch borrow is dropped immediately after use
+                trace!(target: LOG_TARGET, "{:?}", (*rx.borrow()).clone());
+            }
+            let status = {
+                // Ensure the watch borrow is dropped immediately after use
+                match (*rx.borrow()).clone() {
```
sdbondi: The watch::Ref would go out of scope by the end of the match, so another outer scope is unnecessary.

Simple example here
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ad27efed3397c18f31312783969455d7

applications/minotari_console_wallet/src/ui/state/wallet_event_monitor.rs
```diff
@@ -221,7 +208,10 @@ impl WalletEventMonitor {
                     }
                 },
                 _ = base_node_changed.changed() => {
-                    let peer = base_node_changed.borrow().as_ref().cloned();
+                    let peer = {
+                        // Ensure the watch borrow is dropped immediately after use
+                        base_node_changed.borrow().as_ref().cloned()
+                    };
```
sdbondi: Borrowed reference is dropped at this line, no need for an outer scope

~~https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1b53c09b9b31090a303ec9580d1e1f39~~ (made a mistake and put the drop comments in the wrong place - should be after the println!)

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5312b92650300c1b001895fe49393b8

applications/minotari_console_wallet/src/ui/components/burn_tab.rs
sdbondi: The macro expansion isn't really relevant though, the scoping would not be needed in any case because the borrow (watch::Ref) would drop immediately after `format_args!`

applications/minotari_node/src/commands/command/version.rs
```diff
@@ -44,7 +44,11 @@ impl CommandContext {
         println!("Version: {}", consts::APP_VERSION);
         println!("Author: {}", consts::APP_AUTHOR);
 
-        if let Some(ref update) = *self.software_updater.update_notifier().borrow() {
+        let update_notifier = {
```
stringhandler: I think this is worse

stringhandler (commented):
I don't think these improve on the existing code, in some cases you have made it worse

hansieodendaal:
Ok, only two places found in the end where this could be an issue

hansieodendaal:
Thanks @sdbondi for the detailed investigation conducted:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5cb8c4f7b68c094b0b19d7a2b9c20221

Closing this PR because there actually were not any issues!

PR #5805, opened 2023-09-26
feat: failure of min difficulty should not add block to list of bad blocks

SWvheerden:
Description
---
Blocks that fail min difficulty should not be added to the list of bad blocks.

Motivation and Context
---
By adding the blocks to the list of bad blocks we open up the local node to be spammed to write bad block hashes into the database. These waste space and they require a write lock on the database. Checking the difficulty is a very quick and inexpensive check.


brianp (changes requested):
The code change itself is fine. But the PR is really easy to misinterpret:

> Blocks that fail min difficulty are not added to the list of bad blocks.

Is this a problem statement that the PR addresses Or the solution the PR proposes. Should I expect to see a codechange that adds the bad blocks to the list or doesn't?

Reading the motivation and context offers more information, but again it's not *immediately* apparent what direction the code change will go, or whether it will be added to the list or not.

PR #5812, opened 2023-09-26
fix: grpc request overflows

SWvheerden:
Description
---
Adds checked requests to potential overflows in grpc requests


brianp (approved):
utAck

TIL: Saturating addition is addition within bounds. Neat.

PR #5826, opened 2023-10-02
feat: add robustness to monero block extra field handling

hansieodendaal:
Description
---
Added robustness to editing and parsing the extra field in a Monero block. Use cases for the extra field are: 
- the extra field is changed to add a merge mining hash subfield (_performed by the merge mining proxy_) - **NOT applicable, verification is not compromised**
- the merge mining hash is extracted from the extra field (_performed by the merge mining proxy_) - **applicable**
- a Monero block header is verified - **NOT applicable, verification is not compromised**

As per Monero consensus rules, a parsing error with `ExtraField::try_parse(raw_extra_field)` will not represent a failure to de-serialize a block, as, upon error, it will just return the subfields that could be read. This means that when a Monero block template is requested by the merge mining proxy it can contain an "invalid" extra field, mainly due to the presence of a variable length `SubField::Padding(n)` with `n < 255` if not the last subfield in the extra field.

If we append the merge mining tag subfield to a valid existing extra field with a variable `SubField::Padding(n)` with `n < 255` as the last subfield, the new extra field cannot be parsed successfully. To circumvent this, we insert the merge mining subfield in front of the subfield array instead of appending it at the rear.

In the event that we have an invalid extra field to start with, we can now parse the extra field and clean it up in the same process before inserting the merge mining subfield. When extracting the merge mining hash we now also ignore an invalid extra field and return the merge mining hash if present.

Having more than one merge mining tag subfield in a block is currently not allowed in the Tari code base, so this is now checked prior to adding the merge mining tag subfield.

Motivation and Context
---
See above

How Has This Been Tested?
---
- Pass existing tests
- Extended unit test `test_duplicate_append_mm_tag` to verify that appending more than one merge mining tag will be blocked prior to validating the  Monero block header and that header validation will still fail if an additional merge mining tag is present.
- Added unit test `fn test_extra_field_with_parsing_error()`

---
Code walkthrough
Familiarize the monero.rs code base

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -556,7 +588,21 @@ mod test {
...
+        let mut extra_field = match ExtraField::try_parse(&block.miner_tx.prefix.extra) {
+            Ok(val) => val,
+            Err(val) => val,
+        };
+        let hash = monero::Hash::from_slice(hash.as_ref());
+        let mut fields = vec![SubField::MergeMining(Some(VarInt(0)), hash)];
+        fields.append(&mut extra_field.0);
+        extra_field.0 = fields;
+        block.miner_tx.prefix.extra = extra_field.into();
+
```
SWvheerden: can you add some more comments about this, like why is the API failing, and why are we now bypassing the API?
And then do we need to compare them or something?

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -75,11 +75,17 @@ fn get_random_x_difficulty(input: &[u8], vm: &RandomXVMInstance) -> Result<(Diff
...
     let expected_merge_mining_hash = header.merge_mining_hash();
-    let extra_field = ExtraField::try_parse(&monero_data.coinbase_tx.prefix.extra)
-        .map_err(|_| MergeMineError::DeserializeError("Invalid extra field".to_string()))?;
+
+    // Parsing an extra field from bytes will always return a valid extra field, even if it does not represent the
+    // original extra field. As per Monero consensus rules, an error here will not represent a failure to deserialize a
+    // block, so no need to error here.
+    let extra_field = match ExtraField::try_parse(&monero_data.coinbase_tx.prefix.extra) {
+        Ok(val) => val,
+        Err(val) => val,
```
SWvheerden: I know this is the Monero way of handling things, but we should at least include a log here, so we don't just suppress it.

brianp: It's such an antipattern. Why use `try_parse` if it always returns with a value. It should just be `parse`.

hansieodendaal: They do this for clients interested in the data irrespective. They also have a wrapper on top of `ExtraField::try_parse`, `RawExtraField::try_parse`, that will always return an `ExtraField`.

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -110,9 +127,9 @@ pub fn verify_header(header: &BlockHeader) -> Result<MoneroPowData, MergeMineErr
 }
 
 /// Extracts the Monero block hash from the coinbase transaction's extra field
-pub fn extract_tari_hash(monero: &monero::Block) -> Result<Option<monero::Hash>, MergeMineError> {
-    let extra_field = ExtraField::try_parse(&monero.miner_tx.prefix.extra)
-        .map_err(|_| MergeMineError::DeserializeError("Invalid extra field".to_string()))?;
+pub fn proxy_extract_tari_hash_from_block(monero: &monero::Block) -> Result<Option<monero::Hash>, MergeMineError> {
```
SWvheerden:
```suggestion
pub fn extract_tari_hash_from_block_truncate(monero: &monero::Block) -> Result<Option<monero::Hash>, MergeMineError> {
```
Using the proxy name here is adding confusion here. reading the name I cannot infer what proxy does, or what makes it unique. Having knowledge of the complete codebase and knowing exactly what makes it different in the monero methods, I can infer what makes it special. But we need to ensure that just reading this code we can infer the method meanings,

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -174,20 +191,37 @@ pub fn create_ordered_transaction_hashes_from_block(block: &monero::Block) -> Ve
         .collect()
 }
 
-/// Appends merge mining hash to a Monero block
-pub fn append_merge_mining_tag<T: AsRef<[u8]>>(block: &mut monero::Block, hash: T) -> Result<(), MergeMineError> {
+/// Inserts merge mining hash into a Monero block
+pub fn proxy_insert_merge_mining_tag_into_block<T: AsRef<[u8]>>(
```
SWvheerden:
```suggestion
pub fn insert_merge_mining_tag_into_block<T: AsRef<[u8]>>(
```
Same here, remove the proxy name, or something that tells the users what makes it special

SWvheerden (commented):
This looks good, I just have a nit about the new names. 
Adding the word proxy there is only useful and meaningful if you have knowledge of the complete code base. I would prefer adding the specific meaning in the function name or keep as is

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -110,9 +127,9 @@ pub fn verify_header(header: &BlockHeader) -> Result<MoneroPowData, MergeMineErr
 }
 
 /// Extracts the Monero block hash from the coinbase transaction's extra field
-pub fn extract_tari_hash(monero: &monero::Block) -> Result<Option<monero::Hash>, MergeMineError> {
-    let extra_field = ExtraField::try_parse(&monero.miner_tx.prefix.extra)
-        .map_err(|_| MergeMineError::DeserializeError("Invalid extra field".to_string()))?;
+pub fn proxy_extract_tari_hash_from_block(monero: &monero::Block) -> Result<Option<monero::Hash>, MergeMineError> {
+    // When we extract the merge mining hash, we do not care if the extra field can be parsed without error.
+    let extra_field = parse_extra_field_truncate_on_error(&monero.miner_tx.prefix.extra);
```
stringhandler, line 132: DNM! This will allow different proof of work data byte vecs to have the same difficulty

stringhandler (dismissed):
DNM

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
SWvheerden: Dont think this is possible: the function will only extract the tari hash from the already mine block. The Monero code keeps the extra field as a vec<u8> only only deserializes it if used by the code base. So all the bytes are mined, if they are correct and deserializable or not is a different question. But they are all locked by mining

hansieodendaal: Correct @SWvheerden

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -174,20 +209,37 @@ pub fn create_ordered_transaction_hashes_from_block(block: &monero::Block) -> Ve
...
+) -> Result<(), MergeMineError> {
     if hash.as_ref().len() != monero::Hash::len_bytes() {
         return Err(MergeMineError::HashingError(format!(
             "Expected source to be {} bytes, but it was {} bytes",
             monero::Hash::len_bytes(),
             hash.as_ref().len()
         )));
     }
+    // When we insert the merge mining tag, we need to make sure that the extra field is valid.
+    let mut extra_field = parse_extra_field_truncate_on_error(&block.miner_tx.prefix.extra);
```
stringhandler: I think it's better to error here. If there is an error, you will remove the remaining data from the extra field anyway, silently truncating the extra field

stringhandler (commented):
Just one more concern

base_layer/core/src/proof_of_work/monero_rx/helpers.rs
hansieodendaal: Yes, that would not hurt

stringhandler (approved):

PR #5829, opened 2023-10-04
refactor(chatffi): remove c rep types

brianp:
Description
---
- Remove the c rep types so clients don't have to deal with unsafe mutable pointers. It keeps the types used within chat consistent and creates far less duplication of the same data in different wrappers.
- Update documentation
- Update function and type names
- Try and use better aligned int types. In Rust if we want a u8 we should request a c_uchar. i64 should be a u_longlong. Previously most types were being casted down to c_int or c_uint (i32, u32). This presented a probability for casting problems later on.

Motivation and Context
---
Removing the creps is to help with overall library consistency. The rest of the changes are improvements for ease of use.

How Has This Been Tested?
---
- CI
- Cucumber
- New unit tests added

---
See if the changes make sense. Does the library architecture make sense to you, could you use it?

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (dismissed):

base_layer/chat_ffi/src/message.rs
```diff
@@ -128,68 +127,439 @@ pub unsafe extern "C" fn send_chat_message(client: *mut ChatClientFFI, message:
...
+    ptr::swap(error_out, &mut error as *mut c_int);
+
+    if message.is_null() {
+        error = LibChatError::from(InterfaceError::NullError("message".to_string())).code;
+        ptr::swap(error_out, &mut error as *mut c_int);
+        return ptr::null_mut();
+    }
+
+    let data_bytes = (*message).message_id.clone();
+    let len = u32::try_from(data_bytes.len()).expect("Can't cast from usize");
```
hansieodendaal: I think we should rather handle the error here

base_layer/contacts/examples/chat_client/src/networking.rs
```diff
@@ -22,6 +22,8 @@
```
hansieodendaal, line 22: Maybe not related to this PR:

This function
`\base_layer\contacts\examples\chat_client\src\networking.rs` :: `pub async fn start`
is called from
`\base_layer\contacts\examples\chat_client\src\client.rs` :: `pub async fn initialize`
is called from
`base_layer\chat_ffi\src\lib.rs` :: `pub unsafe extern "C" fn create_chat_client`

Why are those functions in an example if they appear to be library code?

hansieodendaal (commented):
Hi @brianp, I like the simplifications brought about by careful inspection of the data types.

Just some comments below.

base_layer/chat_ffi/src/message.rs
brianp: Good catch. I'll make that change.

base_layer/contacts/examples/chat_client/src/networking.rs
brianp: I've been waiting for someone to ask!

This is basically a refactoring waiting to happen. Initially, this client was an example and evolved into a library. It should absolutely be moved elsewhere. Did you have an opinion on where? From the example lib out one directory into contacts makes sense.

SWvheerden: Why dont we just rename the path to remove example, and if we have actual example only code, we can move that to `\base_layer\contacts\chat_client\examples\src\xx.rs`

So example, this `\base_layer\contacts\examples\chat_client\src\client.rs` becomes `\base_layer\contacts\chat_client\src\client.rs`

hansieodendaal: Yes, that could work.

SWvheerden (approved):

SWvheerden (approved):

PR #5830, opened 2023-10-04
fix: listing mode is synced

SWvheerden:
Description
---
Fixes the listening mode is_synced flag

Motivation and Context
---
The is synced flag should only be set if the node is actually up to date. This was the case, but recent updates changed this.


brianp (approved):
utAck

Had to pull the code to look everything over. What was the reason it had been changed before? This change does seem to make the logic more understandable.

Needs a cargo fmt though.

PR #5831, opened 2023-10-04
chore!: upgrade bitflags crate

### LINKED ISSUE #5669: Update bitflags to v2
bitflags has released a v2. We are currently using v1.

brianp:
Description
---
Upgrade the bitflags crate to 2.4.0. This made a few notable changes:
- serde serialization changed from a map `{"bits": 0}` to just outputting the compact bit `0`. It is possible to maintain the legacy format but we don't need to keep that.
- bitflags maintain an internal representation which Borsh can't serialize. So in the instance of KernerlFeatures the interface was migrated from the use of the standard bitflags macro to defining the struct ourself, and implementing a bitflags trait. This allows borsh to operate as normal on the unit struct wrapper of the u8 rep.
- The kernel feature string representation in the faucet file required updating, but the value is still the same once deserialized so it shouldn't require a change in the genesis block, or tx's.
  - It will still be a breaking change as old nodes and new nodes would have a different serialized representation over rpc, and grpc.
- the bitflags crate no longer derives on the structs it creates. I've manually added only the needed derivations to build, and pass tests.

Motivation and Context
---
We wanted to upgrade before release.
Closes: #5669 

[See v2 release changes for bitflags](https://github.com/bitflags/bitflags/releases/tag/2.0.0)

How Has This Been Tested?
---
CI

---
Review the changes

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - May not be able to communicate with older nodes.

BREAKING CHANGE: May not be able to communicate with older nodes. Will require users to upgrade their nodes


applications/minotari_console_wallet/Cargo.toml
```diff
@@ -28,7 +28,7 @@ console-subscriber = "0.1.8"
 # Uncomment for normal use (non tokio-console tracing)
 tokio = { version = "1.23", features = ["signal"] }
 
-bitflags = "1.3.2"
+bitflags = { version = "2.4.0", features = ["serde"] }
```
SWvheerden:
```suggestion
bitflags = { version = "2.4", features = ["serde"] }
```

base_layer/core/Cargo.toml
```diff
@@ -38,7 +38,7 @@ tari_common_sqlite = { path = "../../common_sqlite" }
 
 async-trait = {version = "0.1.50"}
 bincode = "1.1.4"
-bitflags = "1"
+bitflags = { version = "2.4.0", features = ["serde"] }
```
SWvheerden:
```suggestion
bitflags = { version = "2.4", features = ["serde"] }
```

comms/core/Cargo.toml
```diff
@@ -18,7 +18,7 @@ tari_utilities = { version = "0.5" }
 
 anyhow = "1.0.53"
 async-trait = "0.1.36"
-bitflags = "1"
+bitflags = { version = "2.4.0", features = ["serde"] }
```
SWvheerden:
```suggestion
bitflags = { version = "2.4", features = ["serde"] }
```

comms/dht/Cargo.toml
```diff
@@ -20,7 +20,7 @@ tari_storage = {  path = "../../infrastructure/storage" }
 tari_common_sqlite = { path = "../../common_sqlite" }
 
 anyhow = "1.0.53"
-bitflags = "1"
+bitflags = { version = "2.4.0", features = ["serde"] }
```
SWvheerden:
```suggestion
bitflags = { version = "2.4", features = ["serde"] }
```

hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #5834, opened 2023-10-09
feat: add overflow checks to change and fee calculations

hansieodendaal:
Description
---
- Added 3x overflow checks to the change calculation in the sender transaction protocol. 
- Added an additional unit test (`fn test_sender_transaction_protocol_for_overflow`) to verify three sender transaction protocol overflow errors are handled.
- **Edit:** Added overflow checks to kernel fee calculation in blocks.
- **Edit:** Added overflow checks to transaction fee calculation.
- **Edit:** Added an additional unit test (`fn test_fee_overflow`) to verify the two fee overflow errors are handled.
- **Edit:** Fixed `Minotari` and `MicroMinotari` string conversion issue (#5839) for big numbers (_with unit test_) to allow for the proper error message to be printed in the `fn test_fee_overflow` unit test.

Motivation and Context
---
It is possible to crash the system in the change calculation before the transaction is validated; this PR prevents it.
It is possible to crash the system with carefully crafted fees added to transactions and blocks; this PR prevents it.

How Has This Been Tested?
---
Added additional unit tests.

---
See unit tests and code changes.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/blocks/block.rs
+++ b/base_layer/core/src/blocks/block.rs
@@ -197,7 +197,8 @@ impl BlockBuilder {
     /// This function adds the provided transaction kernels to the block WITHOUT updating kernel_mmr_size in the header
     pub fn add_kernels(mut self, mut kernels: Vec<TransactionKernel>) -> Self {
         for kernel in &kernels {
-            self.total_fee += kernel.fee;
+            // Saturating add is used here to prevent overflow; invalid fees will be caught by block validation
+            self.total_fee = self.total_fee.saturating_add(kernel.fee);
         }
         self.kernels.append(&mut kernels);
         self
--- a/base_layer/core/src/transactions/tari_amount.rs
+++ b/base_layer/core/src/transactions/tari_amount.rs
@@ -113,30 +113,40 @@ impl MicroMinotari {
         Self(0)
     }
 
-    pub fn checked_add(self, v: MicroMinotari) -> Option<MicroMinotari> {
-        self.as_u64().checked_add(v.as_u64()).map(Into::into)
+    pub fn checked_add<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        self.as_u64().checked_add(v.as_ref().as_u64()).map(Into::into)
     }
 
-    pub fn checked_sub(self, v: MicroMinotari) -> Option<MicroMinotari> {
-        if self >= v {
-            return Some(self - v);
+    pub fn checked_sub<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        if self >= v.as_ref() {
+            return Some(self - v.as_ref());
         }
         None
     }
 
-    pub fn checked_mul(self, v: MicroMinotari) -> Option<MicroMinotari> {
-        self.as_u64().checked_mul(v.as_u64()).map(Into::into)
+    pub fn checked_mul<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        self.as_u64().checked_mul(v.as_ref().as_u64()).map(Into::into)
     }
 
-    pub fn checked_div(self, v: MicroMinotari) -> Option<MicroMinotari> {
-        self.as_u64().checked_div(v.as_u64()).map(Into::into)
+    pub fn checked_div<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        self.as_u64().checked_div(v.as_ref().as_u64()).map(Into::into)
     }
 
-    pub fn saturating_sub(self, v: MicroMinotari) -> MicroMinotari {
-        if self >= v {
-            return self - v;
+    pub fn saturating_sub<T>(&self, v: T) -> MicroMinotari
+    where T: AsRef<MicroMinotari> {
+        if self >= v.as_ref() {
+            return self - v.as_ref();
         }
         Self(0)
+    }
+
+    pub fn saturating_add<T>(&self, v: T) -> MicroMinotari
+    where T: AsRef<MicroMinotari> {
+        self.0.saturating_add(v.as_ref().0).into()
     }
 
     #[inline]
@@ -146,6 +156,12 @@ impl MicroMinotari {
 
     pub fn to_currency_string(&self, sep: char) -> String {
         format!("{} µT", format_currency(&self.as_u64().to_string(), sep))
+    }
+}
+
+impl AsRef<MicroMinotari> for MicroMinotari {
+    fn as_ref(&self) -> &MicroMinotari {
+        self
     }
 }
 
--- a/base_layer/core/src/transactions/tari_amount.rs
+++ b/base_layer/core/src/transactions/tari_amount.rs
@@ -120,10 +120,7 @@ impl MicroMinotari {
 
     pub fn checked_sub<T>(&self, v: T) -> Option<MicroMinotari>
     where T: AsRef<MicroMinotari> {
-        if self >= v.as_ref() {
-            return Some(self - v.as_ref());
-        }
-        None
+        self.as_u64().checked_sub(v.as_ref().as_u64()).map(Into::into)
     }
 
     pub fn checked_mul<T>(&self, v: T) -> Option<MicroMinotari>
@@ -138,15 +135,12 @@ impl MicroMinotari {
 
     pub fn saturating_sub<T>(&self, v: T) -> MicroMinotari
     where T: AsRef<MicroMinotari> {
-        if self >= v.as_ref() {
-            return self - v.as_ref();
-        }
-        Self(0)
+        self.as_u64().saturating_sub(v.as_ref().as_u64()).into()
     }
 
     pub fn saturating_add<T>(&self, v: T) -> MicroMinotari
     where T: AsRef<MicroMinotari> {
-        self.0.saturating_add(v.as_ref().0).into()
+        self.as_u64().saturating_add(v.as_ref().as_u64()).into()
     }
 
     #[inline]
@@ -182,7 +176,7 @@ impl From<MicroMinotari> for u64 {
     }
 }
 
-impl std::str::FromStr for MicroMinotari {
+impl FromStr for MicroMinotari {
     type Err = MicroMinotariError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
@@ -190,8 +184,10 @@ impl std::str::FromStr for MicroMinotari {
         // Is this Tari or MicroMinotari
         let is_micro_tari = if processed.ends_with("ut") || processed.ends_with("µt") {
             true
+        } else if processed.ends_with('t') {
+            false
         } else {
-            !processed.ends_with('t')
+            !processed.contains('.')
         };
 
         let processed = processed.replace("ut", "").replace("µt", "").replace('t', "");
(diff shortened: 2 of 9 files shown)


base_layer/core/src/blocks/block.rs
```diff
@@ -197,7 +197,7 @@ impl BlockBuilder {
     /// This function adds the provided transaction kernels to the block WITHOUT updating kernel_mmr_size in the header
     pub fn add_kernels(mut self, mut kernels: Vec<TransactionKernel>) -> Self {
         for kernel in &kernels {
-            self.total_fee += kernel.fee;
+            self.total_fee = self.total_fee.saturating_add(kernel.fee);
```
brianp, line 201: I'm curious what the threshold is for blocks so big the fee reaches a saturated max, but max block size hasn't been reached yet. Making blocks of that size and larger the same recommended fee. Or even if it's possible to end up with a recommended fee per gram large enough to reach the max.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -336,7 +344,11 @@ where KM: TransactionKeyManagerInterface
             .fee()
             .calculate(fee_per_gram, 0, 0, 1, change_features_and_scripts_size);
         // Subtract with a check on going negative
-        let total_input_value = total_to_self + total_amount + fee_without_change;
+        let total_input_value = [total_to_self, total_amount, fee_without_change]
+            .iter()
+            .fold(Ok(MicroMinotari::zero()), |acc, x| {
+                acc?.checked_add(*x).ok_or("Total input value overflow")
```
brianp: Strange it needs to be dereferenced here `checked_add(*x)`

brianp (approved):
LGTM
utACK

base_layer/core/src/blocks/block.rs
AaronFeickert, line 201: It would be helpful to include a comment about why using saturating addition is safe here (from a protocol perspective), as opposed to checked addition.

base_layer/core/tests/tests/node_comms_interface.rs
```diff
@@ -251,6 +267,165 @@ async fn inbound_fetch_blocks() {
...
+            .unwrap();
+
+        stx_builder.with_output(utxo, sender_offset_key_id).await.unwrap();
+    }
+
+    stx_builder
+}
+
+#[tokio::test]
+async fn test_sender_transaction_protocol_for_overflow() {
```
AaronFeickert, line 366: This test passes even with the fee `saturating_add` reverted. Is this the intended behavior? I assumed this test was intended to catch a regression, which it apparently does not.

AaronFeickert (changes requested):

base_layer/core/src/blocks/block.rs
hansieodendaal: I think that is the crux of the matter; it is possible to end up with such a large fee just with a few transactions

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: `.checked_add(` is not implemented for a reference value

base_layer/core/tests/tests/node_comms_interface.rs
hansieodendaal: I changed `pub fn add_kernels(mut self, mut kernels: Vec<TransactionKernel>)` without adding a regression test for it; this test is for the other three changes, will try to add a test for `add_kernels`.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
SWvheerden: is it possible to add that implementation?

SWvheerden (commented):
looks good, a small nit though

base_layer/core/src/blocks/block.rs
hansieodendaal: I added a test for this case

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: ~~We could have~~ 
``` rust
    pub fn checked_add(self, v: MicroMinotari) -> Option<MicroMinotari> {
        self.as_u64().checked_add(v.as_u64()).map(Into::into)
    }

    pub fn checked_add_ref(self, v: &MicroMinotari) -> Option<MicroMinotari> {
        self.as_u64().checked_add(v.as_u64()).map(Into::into)
    }
```
~~as~~
``` rust
new_val = val.checked_add_ref(x).ok_or("Total input value overflow");
```
~~vs.~~
``` rust
new_val = val.checked_add(*x).ok_or("Total input value overflow")
```
~~but I do not see the benefit of the additional function when we can just dereference when needed.~~

AaronFeickert (dismissed):
Manually tested that reverting any of the changes will correctly fail a corresponding unit test.

AaronFeickert:
ACK

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
hansieodendaal: Ok, please disregard my previous comment, I implemented `AsRef` for those methods to facilitate the overloading

base_layer/core/tests/tests/mempool.rs
```diff
@@ -330,6 +334,135 @@ async fn test_time_locked() {
     assert_eq!(mempool.insert(tx2).await.unwrap(), TxStorageResponse::UnconfirmedPool);
 }
 
+#[tokio::test]
+#[allow(clippy::too_many_lines)]
+async fn test_fee_overflow() {
```
SWvheerden: This test is in the wrong location, its in mempool.rs, it does not even use a mempool.

base_layer/core/tests/tests/mempool.rs
```diff
@@ -330,6 +334,135 @@ async fn test_time_locked() {
...
+        features: OutputFeatures::default()
+    );
+    let stx_builder = create_stx_protocol_internal(schema, &key_manager, &mut Vec::new()).await;
+
+    // This will call `Fee::calculate(...)`, which will overflow if regressed
+    let build_result = stx_builder.build().await;
+    assert!(build_result.is_err());
+    assert_eq!(
+        format!("{:?}", build_result.unwrap_err()),
+        "You are spending (18446744073709.550781 T) more than you're providing (10.000000 T)."
```
SWvheerden: This does not seem correct. How did it get to this amount?
u64 max = "18446744073709.551615"
So this test is seems if its not running into overflow

this should print out 18446744073709.551615 - 10 = 18446744073691.551615

base_layer/core/tests/tests/mempool.rs
hansieodendaal: Ok, I will move the test

base_layer/core/tests/tests/mempool.rs
hansieodendaal: The error message is misleading, I will change it:
```rust
        // Subtract with a check on going negative
        let total_input_value = [total_to_self, total_amount, fee_without_change]
            .iter()
            .fold(Ok(MicroMinotari::zero()), |acc, x| {
                acc?.checked_add(x).ok_or("Total input value overflow")
            })?;
        let change_amount = total_being_spent.checked_sub(total_input_value);
        match change_amount {
            None => Err(format!(
                "You are spending ({}) more than you're providing ({}).",
                total_input_value, total_being_spent
            )),
```

hansieodendaal: This is the part that will overflow if unbounded:
```rust
        let fee_without_change = self.fee().calculate(
            fee_per_gram,
            1,
            num_inputs,
            num_outputs,
            features_and_scripts_size_without_change,
        );
```

hansieodendaal: For this test the inputs are
```
total_to_self 0 µT, total_amount 0 µT, fee_without_change 18446744073709551615 µT
```
resulting in  error message
```
"You are spending more than you're providing: provided 10.000000 T, required 18446744073709.550781 T."
```

hansieodendaal: Ok, I found the error with the displayed value being off from `u64::MAX`; the issue is in `impl Display for MicroMinotari` as it does floating-point arithmetic.
u64 `18446744073709551615` is converted to f64 `18446744073709552000` then divided by `1_000_000f64` resulting in `18446744073709.550781`

hansieodendaal: Now the test values are:
```
total_to_self 0 µT, total_amount 0 µT, fee_without_change 18446744073709.551615 T
```
resulting in error message
```
"You are spending more than you're providing: provided 10.000000 T, required 18446744073709.551615 T."
```
which is consistent with `u64::MAX`

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -113,32 +113,42 @@ impl MicroMinotari {
...
-        self.as_u64().checked_add(v.as_u64()).map(Into::into)
+    pub fn checked_add<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        self.as_u64().checked_add(v.as_ref().as_u64()).map(Into::into)
     }
 
-    pub fn checked_sub(self, v: MicroMinotari) -> Option<MicroMinotari> {
-        if self >= v {
-            return Some(self - v);
+    pub fn checked_sub<T>(&self, v: T) -> Option<MicroMinotari>
```
AaronFeickert, line 121: Why not use the same approach as `checked_add` for consistency and simplicity?

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -113,32 +113,42 @@ impl MicroMinotari {
...
+    where T: AsRef<MicroMinotari> {
+        if self >= v.as_ref() {
+            return self - v.as_ref();
         }
         Self(0)
     }
 
+    pub fn saturating_add<T>(&self, v: T) -> MicroMinotari
+    where T: AsRef<MicroMinotari> {
+        self.0.saturating_add(v.as_ref().0).into()
```
AaronFeickert: You could use `self.as_u64()` and `v.as_ref().as_u64()` for consistency with the other operations.

base_layer/core/src/transactions/tari_amount.rs
```diff
@@ -113,32 +113,42 @@ impl MicroMinotari {
...
-        self.as_u64().checked_div(v.as_u64()).map(Into::into)
+    pub fn checked_div<T>(&self, v: T) -> Option<MicroMinotari>
+    where T: AsRef<MicroMinotari> {
+        self.as_u64().checked_div(v.as_ref().as_u64()).map(Into::into)
     }
 
-    pub fn saturating_sub(self, v: MicroMinotari) -> MicroMinotari {
-        if self >= v {
-            return self - v;
+    pub fn saturating_sub<T>(&self, v: T) -> MicroMinotari
```
AaronFeickert, line 136: This could use the same approach as `saturating_add` for consistency and simplicity.

AaronFeickert (changes requested):

base_layer/core/tests/tests/mempool.rs
SWvheerden: Good find, that far not ideal, especially when it's displaying amounts, we should remove that rounding. But not for this PR.
Opened a new issue: https://github.com/tari-project/tari/issues/5839

SWvheerden (approved):

PR #5835, opened 2023-10-10
refactor(proto)!: remove proto bytes for std bytes

brianp:
Description
---
This PR will remove the google proto bytes wrapper in favour of using a standard vec.

Motivation and Context
---
Use standard rust types, instead of library specific types.

Related to #5833

How Has This Been Tested?
---
CI

---
Previously used test values included `None`, `Some(vec![])` and `Some(FixedHash::zero())`. This PR removes the option wrappers and as a result the previous `Some(vec![])` values will be `vec![]` and considered the equivalent to `None`. 

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

Messages between new and old nodes are likely to break.


base_layer/core/src/base_node/rpc/service.rs
```diff
@@ -458,21 +458,20 @@ impl<B: BlockchainBackend + 'static> BaseNodeWalletService for BaseNodeWalletRpc
...
-                .fetch_header_by_block_hash(hash)
-                .await
-                .rpc_status_internal_error(LOG_TARGET)?
-                .is_none()
-            {
-                return Err(RpcStatus::not_found(
-                    "Chain does not include header. It might have been reorged out",
-                ));
-            }
+        let hash = message
```
sdbondi: To preserve previous functionality I think you need an empty check here

brianp: I suppose you're correct. If the value is empty then we're not really supposed to do anything here. I'll make that change.

SWvheerden (approved):

Issue #5816, opened 2023-09-26
Header sync should check find split blocks for duplicate headers.


SWvheerden:
A miner or attacker with a block with higher proof-of-work can lie about their
difficulty and height, forcing a peer to perform sync. This makes the victim not
accept new blocks and perform useless validations. The attacker is not penalized.
This issue is similar to TARI-013, but attackers need to provide a block with
higher proof of work. This makes the motivation similar to TARI-001: a miner
wants to delay other miners and gain advantage. In this case they can get up to 60
seconds advantage due to acquiring the add_block lock.
The attack leverages the fact that the initial header sync performed in
find_chain_split does not check for duplicated headers and makes requests to an
untrusted node.

All in all, an attacker has to:
1. Have a block with at least a slightly higher difficulty than the victim
2. Announce a much greater higher difficulty to the victim in order for them to
believe they have FallenBehind.
3. Wait 29s before responding to their find_chain_split request
4. Answer not_found
5. Victim will repeat request, wait 29s again
6. Answer with 998 known blocks and the one new block

Local node should ban peers who lie about their difficulty and provide known blocks.

PR #5837, opened 2023-10-11
feat: consistent handling of edge cases for header sync

hansieodendaal:
Description
---
This PR fixes an edge case for header sync where the local node has more headers, but the remote node has better proof-of-work (accumulated difficulty), thus higher actual blockchain height. It also simplifies the attempted header sync logic and error handling when a peer does not return headers and introduces consistent naming, variable re-use and use of information

Supersedes #5756

Motivation and Context
---
When doing header sync, and determining from which peer to sync, we need to consider our actual blockchain state when comparing chains. It is possible that our node has valid block headers that will translate into a  higher proof of work when fully synced than the remote syncing peer, but lacking the blocks to back the block headers, thus the remote chain has an actual higher proof of work blockchain.
With the current implementation, race conditions can exist when determining if the peer is misbehaving, i.e. lying about their proof-of-work or not wanting to supply block headers, due to a mismatch in tip height used for the check.

This PR fixes the race conditions by always using the latest local chain metadata and block headers, ignoring all sync peers that do not exceed our own accumulated difficulty and using consistent information in all the checks.

How Has This Been Tested?
---
Added integration-level unit tests:
- `test_header_sync_happy_path`
- `test_header_sync_with_fork_happy_path`
- `test_header_sync_uneven_headers_and_blocks_happy_path`
- `test_header_sync_uneven_headers_and_blocks_peer_lies_about_pow_no_ban`
- `test_header_sync_even_headers_and_blocks_peer_lies_about_pow_with_ban`
- `test_header_sync_even_headers_and_blocks_peer_metadata_improve_with_reorg`

---
- Code walk-through
- Review unit tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
+++ b/applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
@@ -388,7 +388,6 @@ impl TransactionsTab {
             let status = Span::styled(status_msg, Style::default().fg(Color::White));
             let message = Span::styled(tx.message.as_str(), Style::default().fg(Color::White));
 
-            // let mined_time = DateTime::<Local>::from_utc(tx.mined_timestamp, Local::now().offset().to_owned());
             let mined_timestamp = Span::styled(
                 match tx.mined_timestamp {
                     None => String::new(),
@@ -398,7 +397,6 @@ impl TransactionsTab {
                             .format("%Y-%m-%d %H:%M:%S")
                     ),
                 },
-                // format!("{}", mined_time.format("%Y-%m-%d %H:%M:%S")),
                 Style::default().fg(Color::White),
             );
 
--- a/base_layer/core/src/base_node/proto/rpc.proto
+++ b/base_layer/core/src/base_node/proto/rpc.proto
@@ -49,10 +49,6 @@ message FindChainSplitResponse {
   // `FindChainSplitRequest::block_hashes`. This value could also be used to
   // know how far back a split occurs.
   uint64 fork_hash_index = 2;
-  // The current header height of this node
-  uint64 tip_height = 3;
-  // The current geometric mean of the pow of the chain tip, or `None` if there is no chain
-  bytes accumulated_difficulty = 4;
 }
 
 message SyncKernelsRequest {
--- a/base_layer/core/src/base_node/proto/rpc.proto
+++ b/base_layer/core/src/base_node/proto/rpc.proto
@@ -51,6 +51,8 @@ message FindChainSplitResponse {
   uint64 fork_hash_index = 2;
   // The current header height of this node
   uint64 tip_height = 3;
+  // The current geometric mean of the pow of the chain tip, or `None` if there is no chain
+  bytes accumulated_difficulty = 4;
 }
 
 message SyncKernelsRequest {
--- a/base_layer/core/src/base_node/sync/header_sync/error.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/error.rs
@@ -56,6 +56,8 @@ pub enum BlockHeaderSyncError {
     ConnectivityError(#[from] ConnectivityError),
     #[error("Node is still not in sync. Sync will be retried with another peer if possible.")]
     NotInSync,
+    #[error("Internal data inconsistency found `{0}")]
+    InternalError(String),
     #[error("Unable to locate start hash `{0}`")]
     StartHashNotFound(String),
     #[error("Expected header height {expected} got {actual}")]
@@ -106,7 +108,8 @@ impl BlockHeaderSyncError {
             BlockHeaderSyncError::ConnectivityError(_) |
             BlockHeaderSyncError::NotInSync |
             BlockHeaderSyncError::PeerNotFound |
-            BlockHeaderSyncError::ChainStorageError(_) => None,
+            BlockHeaderSyncError::ChainStorageError(_) |
+            BlockHeaderSyncError::InternalError(_) => None,
 
             // short ban
             err @ BlockHeaderSyncError::MaxLatencyExceeded { .. } => Some(BanReason {
--- a/base_layer/core/src/base_node/sync/header_sync/error.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/error.rs
@@ -56,8 +56,6 @@ pub enum BlockHeaderSyncError {
     ConnectivityError(#[from] ConnectivityError),
     #[error("Node is still not in sync. Sync will be retried with another peer if possible.")]
     NotInSync,
-    #[error("Internal data inconsistency found `{0}")]
-    InternalError(String),
     #[error("Unable to locate start hash `{0}`")]
     StartHashNotFound(String),
     #[error("Expected header height {expected} got {actual}")]
@@ -108,8 +106,7 @@ impl BlockHeaderSyncError {
             BlockHeaderSyncError::ConnectivityError(_) |
             BlockHeaderSyncError::NotInSync |
             BlockHeaderSyncError::PeerNotFound |
-            BlockHeaderSyncError::ChainStorageError(_) |
-            BlockHeaderSyncError::InternalError(_) => None,
+            BlockHeaderSyncError::ChainStorageError(_) => None,
 
             // short ban
             err @ BlockHeaderSyncError::MaxLatencyExceeded { .. } => Some(BanReason {
--- a/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
@@ -40,13 +40,14 @@ use super::{validator::BlockHeaderSyncValidator, BlockHeaderSyncError};
 use crate::{
     base_node::sync::{ban::PeerBanManager, hooks::Hooks, rpc, BlockchainSyncConfig, SyncPeer},
     blocks::{BlockHeader, ChainBlock, ChainHeader},
-    chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend},
+    chain_storage::{async_db::AsyncBlockchainDb, BlockchainBackend, ChainStorageError},
     common::rolling_avg::RollingAverageTime,
     consensus::ConsensusManager,
     proof_of_work::randomx_factory::RandomXFactory,
     proto::{
         base_node as proto,
         base_node::{FindChainSplitRequest, SyncHeadersRequest},
+        core::BlockHeader as ProtoBlockHeader,
     },
 };
 
@@ -63,7 +64,7 @@ pub struct HeaderSynchronizer<'a, B> {
     connectivity: ConnectivityRequester,
     sync_peers: &'a mut Vec<SyncPeer>,
     hooks: Hooks,
-    local_metadata: &'a ChainMetadata,
+    local_cached_metadata: &'a ChainMetadata,
     peer_ban_manager: PeerBanManager,
 }
 
@@ -85,7 +86,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
             connectivity,
             sync_peers,
             hooks: Default::default(),
-            local_metadata,
+            local_cached_metadata: local_metadata,
             peer_ban_manager,
         }
     }
@@ -105,7 +106,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         self.hooks.add_on_rewind_hook(hook);
     }
 
-    pub async fn synchronize(&mut self) -> Result<SyncPeer, BlockHeaderSyncError> {
+    pub async fn synchronize(&mut self) -> Result<(SyncPeer, AttemptSyncResult), BlockHeaderSyncError> {
         debug!(target: LOG_TARGET, "Starting header sync.",);
 
         info!(
@@ -117,7 +118,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         let mut latency_increases_counter = 0;
         loop {
             match self.try_sync_from_all_peers(max_latency).await {
-                Ok(sync_peer) => break Ok(sync_peer),
+                Ok((peer, sync_result)) => break Ok((peer, sync_result)),
                 Err(err @ BlockHeaderSyncError::AllSyncPeersExceedLatency) => {
                     // If we have few sync peers, throw this out to be retried later
                     if self.sync_peers.len() < 2 {
@@ -135,7 +136,10 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
     }
 
     #[allow(clippy::too_many_lines)]
-    pub async fn try_sync_from_all_peers(&mut self, max_latency: Duration) -> Result<SyncPeer, BlockHeaderSyncError> {
+    pub async fn try_sync_from_all_peers(
+        &mut self,
+        max_latency: Duration,
+    ) -> Result<(SyncPeer, AttemptSyncResult), BlockHeaderSyncError> {
         let sync_peer_node_ids = self.sync_peers.iter().map(|p| p.node_id()).cloned().collect::<Vec<_>>();
         info!(
             target: LOG_TARGET,
@@ -145,7 +149,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         let mut latency_counter = 0usize;
         for node_id in sync_peer_node_ids {
             match self.connect_and_attempt_sync(&node_id, max_latency).await {
-                Ok(peer) => return Ok(peer),
+                Ok((peer, sync_result)) => return Ok((peer, sync_result)),
                 Err(err) => {
                     let ban_reason = BlockHeaderSyncError::get_ban_reason(
                         &err,
@@ -180,7 +184,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         &mut self,
         node_id: &NodeId,
         max_latency: Duration,
-    ) -> Result<SyncPeer, BlockHeaderSyncError> {
+    ) -> Result<(SyncPeer, AttemptSyncResult), BlockHeaderSyncError> {
         let peer_index = self
             .get_sync_peer_index(node_id)
             .ok_or(BlockHeaderSyncError::PeerNotFound)?;
@@ -214,8 +218,8 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
 
         debug!(target: LOG_TARGET, "Sync peer latency is {:.2?}", latency);
         let sync_peer = self.sync_peers[peer_index].clone();
-        self.attempt_sync(&sync_peer, client, max_latency).await?;
-        Ok(sync_peer)
+        let sync_result = self.attempt_sync(&sync_peer, client, max_latency).await?;
+        Ok((sync_peer, sync_result))
     }
 
     async fn dial_sync_peer(&self, node_id: &NodeId) -> Result<PeerConnection, BlockHeaderSyncError> {
@@ -236,7 +240,7 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         sync_peer: &SyncPeer,
         mut client: rpc::BaseNodeSyncRpcClient,
         max_latency: Duration,
-    ) -> Result<(), BlockHeaderSyncError> {
+    ) -> Result<AttemptSyncResult, BlockHeaderSyncError> {
         let latency = client.get_last_request_latency();
         debug!(
             target: LOG_TARGET,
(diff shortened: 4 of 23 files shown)


base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
 
-        let headers = headers
+        let headers = peer_response
+            .headers
             .into_iter()
             .map(BlockHeader::try_from)
             .collect::<Result<Vec<_>, _>>()
             .map_err(BlockHeaderSyncError::ReceivedInvalidHeader)?;
-        let num_new_headers = headers.len();
+        let num_new_headers = headers.len(); // Required fro later use, no 'Copy' trait on 'BlockHeader'
```
hansieodendaal:
```suggestion
        let num_new_headers = headers.len(); // Required for later use, no 'Copy' trait on 'BlockHeader'
```

sdbondi (dismissed):
Nice - utACK

applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -388,21 +388,23 @@ impl TransactionsTab {
             let status = Span::styled(status_msg, Style::default().fg(Color::White));
             let message = Span::styled(tx.message.as_str(), Style::default().fg(Color::White));
 
-            // let mined_time = DateTime::<Local>::from_utc(tx.mined_timestamp, Local::now().offset().to_owned());
+            // let mined_time = DateTime::<Local>::from_naive_utc_and_offset(tx.mined_timestamp,
+            // Local::now().offset().to_owned());
```
SWvheerden:
```suggestion
```

applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -388,21 +388,23 @@ impl TransactionsTab {
...
                 match tx.mined_timestamp {
                     None => String::new(),
                     Some(mined_timestamp) => format!(
                         "{}",
-                        DateTime::<Local>::from_utc(mined_timestamp, Local::now().offset().to_owned())
+                        DateTime::<Local>::from_naive_utc_and_offset(mined_timestamp, Local::now().offset().to_owned())
                             .format("%Y-%m-%d %H:%M:%S")
                     ),
                 },
                 // format!("{}", mined_time.format("%Y-%m-%d %H:%M:%S")),
```
SWvheerden:
```suggestion
```

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
-        if headers.is_empty() {
-            if fork_hash_index > 0 {
+        // If the peer returned no new headers, they have no headers, but may still have more blocks than we have,
+        // thus have a higher accumulated difficulty.
+        if peer_response.headers.is_empty() {
+            if peer_response.fork_hash_index > 0 {
                 debug!(
                     target: LOG_TARGET,
-                    "Peer `{}` has sent no headers but forked_hash_index is {}. The peer is behind our chain.",
+                    "Peer `{}` has sent no headers with forked_hash_index {}. The peer has less headers than we have.",
```
SWvheerden:
```suggestion
                    "Peer `{}` has sent no headers with some forked_hash_index {}. The peer has less headers than we have.",
```

base_layer/core/tests/tests/header_sync.rs
```diff
@@ -0,0 +1,294 @@
...
+    assert_eq!(bob_node.blockchain_db.get_height().unwrap(), 1);
+
+    // Alice attempts header sync, still on the genesys block, headers will be lagging
+    let event = sync_headers(&mut alice_state_machine, &alice_node, &bob_node).await;
+    // "Lagging"
+    match event.clone() {
+        StateEvent::HeadersSynchronized(val, sync_result) => {
+            assert_eq!(val.claimed_chain_metadata().height_of_longest_chain(), 1);
+            assert_eq!(val.claimed_chain_metadata().accumulated_difficulty(), 4);
+            assert_eq!(sync_result.headers_returned, 1);
```
SWvheerden, line 47: Nit: This is probably fine as is here. I would have preferred to test these values in unit tests, then we don't have to add code to allow testing of values. These values are only used by these tests.

SWvheerden (commented):
looks good, just needs a few commented out lines removed

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +249,69 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+
+        if best_header_height < best_block_height ||
+            best_block_height < self.local_cached_metadata.height_of_longest_chain()
+        {
+            return Err(BlockHeaderSyncError::ChainStorageError(
+                ChainStorageError::CorruptedDatabase("Inconsistent block and header data".to_string()),
+            ));
+        }
+
+        let peer_response = self
```
stringhandler: What is the purpose of moving this logic here? I would prefer to keep the logic in `determine_sync_status` but propagate `peer_response` as part of the Error or result

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +249,69 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
-        match sync_status {
-            SyncStatus::InSync | SyncStatus::WereAhead => {
-                let metadata = self.db.get_chain_metadata().await?;
-                if metadata.height_of_longest_chain() < header_tip_height {
+        let header_sync_status = self
+            .determine_sync_status(&sync_peer.to_string(), best_header, best_block, peer_response.clone())
+            .await?;
+
+        match header_sync_status.clone() {
+            HeaderSyncStatus::InSyncOrAhead => {
```
stringhandler, line 277:
```suggestion
            HeaderSyncStatus::InSyncOrAhead{ peer_response } => {
```
Rather propagate it here

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +249,69 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
-                    debug!(
+                    // We will only attempt sync if the our accumulated difficulty is less than the peer's claimed
+                    // accumulated difficulty, thus this is adverse behaviour form the peer.
+                    warn!(
                         target: LOG_TARGET,
                         "Headers and block state are already in-sync (Header Tip: {}, Block tip: {}, Peer's height: \
-                         {})",
-                        header_tip_height,
-                        metadata.height_of_longest_chain(),
+                         {}), peer has lied about chain metadata or did not want to provide headers",
```
stringhandler: This is not true. We only checked if our best_block < our best_headers

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +249,69 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+                        best_header_height,
+                        best_block_height,
                         sync_peer.claimed_chain_metadata().height_of_longest_chain(),
                     );
+
                     Err(BlockHeaderSyncError::PeerSentInaccurateChainMetadata {
                         claimed: sync_peer.claimed_chain_metadata().accumulated_difficulty(),
```
stringhandler: This seems like an assumption, the logic here is incorrect. See above comment

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +249,69 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+                        local: self.local_cached_metadata.accumulated_difficulty(),
                     })
                 }
             },
-            SyncStatus::Lagging(split_info) => {
+            HeaderSyncStatus::Lagging(split_info) => {
                 self.hooks.call_on_progress_header_hooks(
                     split_info
-                        .local_tip_header
+                        .best_block
```
stringhandler: Should this not be `best_header`?

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+                peer_response.headers.len(),
                 sync_peer
             );
         }
 
-        // If the peer returned no new headers, this means header sync is done.
-        if headers.is_empty() {
-            if fork_hash_index > 0 {
+        // If the peer returned no new headers, they have no headers, but may still have more blocks than we have,
+        // thus have a higher accumulated difficulty.
```
stringhandler: This comment makes no sense. The headers are what determines the accumulated difficulty.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
                     target: LOG_TARGET,
-                    "Peer `{}` has sent no headers but forked_hash_index is {}. The peer is behind our chain.",
+                    "Peer `{}` has sent no headers with forked_hash_index {}. The peer has less headers than we have.",
                     sync_peer,
-                    fork_hash_index
+                    peer_response.fork_hash_index
                 );
-
-                return Ok(SyncStatus::WereAhead);
+            } else {
```
stringhandler: This feels like the peer can abuse this. They return no headers but then return a fork index of 0.

How is it that we requested headers from them but didn't receive any? 

Feels like we should ban them

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
 
-        let headers = headers
+        let headers = peer_response
+            .headers
             .into_iter()
             .map(BlockHeader::try_from)
             .collect::<Result<Vec<_>, _>>()
             .map_err(BlockHeaderSyncError::ReceivedInvalidHeader)?;
-        let num_new_headers = headers.len();
+        let num_new_headers = headers.len(); // Required for later use, no 'Copy' trait on 'BlockHeader'
```
stringhandler:
```suggestion
        let num_new_headers = headers.len(); 
```

No need to explain

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -380,69 +398,76 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
 
         // NOTE: We can trust that the header associated with this hash exists because `block_hashes` was supplied by
         // this node. Bounds checking for fork_hash_index has been done above.
         #[allow(clippy::cast_possible_truncation)]
-        let chain_split_hash = block_hashes.get(fork_hash_index as usize).unwrap();
+        let chain_split_hash = peer_response
+            .block_hashes
+            .get(peer_response.fork_hash_index as usize)
```
stringhandler: This unwrap seems like the peer could cause a panic if `fork_hash_index > block_hashes.len()`

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -462,21 +487,21 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
         );
 
         // Basic sanity check that the peer sent tip height greater than the split.
-        let split_height = local_tip_header.height().saturating_sub(steps_back);
-        if remote_tip_height < split_height {
+        let split_height = best_header.height().saturating_sub(peer_response.reorg_steps_back);
+        if peer_response.remote_tip_height < split_height {
             return Err(BlockHeaderSyncError::InvalidProtocolResponse(format!(
```
stringhandler: Does this result in a ban? I think it should

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -462,21 +487,21 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
                 "Peer {} sent invalid remote tip height",
                 sync_peer
             )));
         }
 
         let chain_split_info = ChainSplitInfo {
-            local_tip_header,
-            remote_tip_height,
-            reorg_steps_back: steps_back,
+            best_block,
```
stringhandler: did you mean `best_header`?

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -781,18 +797,43 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+    headers: Vec<ProtoBlockHeader>,
+    fork_hash_index: u64,
+    remote_tip_height: u64,
+}
+
+/// Information about the chain split from the remote node.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ChainSplitInfo {
+    /// The best block on the local chain.
+    pub best_block: ChainHeader,
```
stringhandler: This is badly named. Is it the best header we have or the best block we have?

I made a few comments already, but looks like it could just be that this should have been `best_header`

base_layer/core/tests/helpers/nodes.rs
```diff
@@ -105,6 +108,7 @@ pub struct BaseNodeBuilder {
     mempool_config: Option<MempoolConfig>,
     mempool_service_config: Option<MempoolServiceConfig>,
     liveness_service_config: Option<LivenessConfig>,
+    p2p_config: Option<P2pConfig>,
```
stringhandler, line 111: Not related to this PR, but this struct should really be called "TestBaseNodeBuilder"

base_layer/p2p/src/initialization.rs
```diff
@@ -135,7 +135,7 @@ pub async fn initialize_local_test_comms<P: AsRef<Path>>(
     discovery_request_timeout: Duration,
     seed_peers: Vec<Peer>,
     shutdown_signal: ShutdownSignal,
-) -> Result<(CommsNode, Dht, MessagingEventSender), CommsInitializationError> {
+) -> Result<(UnspawnedCommsNode, Dht, MessagingEventSender), CommsInitializationError> {
```
stringhandler, line 138: Unrelated: This test method should also be moved into a file that shows it's a test method. Because the method is so long, it is difficult to see that it's test code

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: Its true by deduction. 
For us to have entered Header_sync the peer needed tot have a better pow(it claimed it did)
Then we check that out pow did not update and that they still have a claimed higher pow. 

We then asked the peer to provide us the chain which had a better pow, but it could not. So we have the same or more headers on the exact same chain as they have.  But we can have the headers, not the block, so if we check if we have the same blocks than headers, we must have the same or more blocks than they. Its not possible to have blocks and not the headers. 

So the only conclusion we can make is one of the following 3:
They lied about their pow
They did not want to provide us with a higher chain
The swapped to a lower-difficulty pow chain. 

So we are only left with 1 and 2 realisticly.

base_layer/core/tests/tests/header_sync.rs
hansieodendaal: Agreed that it is probably fine

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: It was for testability. Moving the `find_chain_split` back to `determine_sync_status` and returning `peer_response` alongside `header_sync_status`, as `AttemptSyncResult` pass back `header_sync_status` and that will become bloated.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: As above

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: This is exactly the fix we made, its the header of the best block. 
Best block and best header are not always the same

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: This comment exactly states the problem we need to address, the logic is divided into weird places. 

The reason this is, is that the PoW of the local node is determined by the headers, of the blocks we have the complete block for. 
But the node downloads the headers before it downloads the complete block. 

So in this case, the node could already have downloaded the headers, but not yet the blocks, so technically the pow of the local node is lower, although it might have the headers already.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: See the above comment, we first need to check if they have full blocks for those, as we might not have them yet. 
But if we have the full blocks for those, then yes, we do ban them, but only after that check

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: We can move the call back in, but move the decision logic out. 

```rust
fn determine_sync_status(...)->{
 let resp = self. find_chain_split(..)
 decide_logic(resp)
}

fn decide_logic(..)->
{
..
}
```

Something like that.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Ok, I provided better data for a better decision

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: As per above

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Fixed

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: See new logic test ^^

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Yes it does

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Fixed the name

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Fixed the name

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: I provided a better name

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
stringhandler: Please provide a link to the new code or more detail about what you did, I can't find it among the other changes

base_layer/core/src/base_node/proto/rpc.proto
```diff
@@ -51,6 +51,8 @@ message FindChainSplitResponse {
   uint64 fork_hash_index = 2;
   // The current header height of this node
   uint64 tip_height = 3;
+  // The current geometric mean of the pow of the chain tip, or `None` if there is no chain
+  bytes accumulated_difficulty = 4;
```
SWvheerden: this is meaning less, the peer can fill anything in here.  We already have the cl;aimed pow from the metadata ping/pong

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +248,72 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
+        // - If the peer lied about the start hash index or sent an invalid remote tip height, this method will return a
+        //   ban-able error.
+        let (header_sync_status, peer_response) = self
+            .determine_sync_status(sync_peer, best_header.clone(), best_block.clone(), &mut client)
+            .await?;
+
+        match header_sync_status.clone() {
+            HeaderSyncStatus::InSyncOrAhead => {
+                // Our POW is less than the peer's POW, as verified before attempted header sync, therefor, if the peer
+                // did not supply any headers and we know we are behind, then we can ban the peer.
```
SWvheerden:
```suggestion
                // Our POW is less than the peer's POW, as verified before the attempted header sync, therefor, if the peer
                // did not supply any headers and they claimed we are behind, then we can ban the peer.
```

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
```diff
@@ -245,59 +248,72 @@ impl<'a, B: BlockchainBackend + 'static> HeaderSynchronizer<'a, B> {
...
                         target: LOG_TARGET,
-                        "Headers and block state are already in-sync (Header Tip: {}, Block tip: {}, Peer's height: \
-                         {})",
-                        header_tip_height,
-                        metadata.height_of_longest_chain(),
-                        sync_peer.claimed_chain_metadata().height_of_longest_chain(),
+                        "Peer `{}` did not provide any headers although they have a better chain and more headers: their \
+                        difficulty: {}, our difficulty: {}. Peer will be banned.",
+                        sync_peer,
+                        peer_response.peer_accumulated_difficulty,
```
SWvheerden: just use the sync peer, no need to add new chain metadatata we cannot verify it anyway.

SWvheerden (changes requested):
adding in a new field to the proto for the peer to also send is meaningless.
They already sent us one that they claim is higher than ours. All we have to do now is verify it, we should not ask for a new one.

base_layer/core/src/base_node/proto/rpc.proto
hansieodendaal: If the peer is honest, it have may increased since the ping-pong. If the peer is dishonest, they can lie about `tip_height` as well.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: The conditions are now tested separately and both will be ban-able offences.
See
```rust
        // Basic sanity check about the peer's claimed chain metadata
        if chain_split_result.peer_accumulated_difficulty < sync_peer.claimed_chain_metadata().accumulated_difficulty()
```
and
```rust
        match header_sync_status.clone() {
            HeaderSyncStatus::InSyncOrAhead => {
                // Our POW is less than the peer's POW, as verified before the attempted header sync, therefore, if the
                // peer did not supply any headers and we know we are behind based on the peer's claimed metadata, then
                // we can ban the peer.
                if best_header.height() == best_block.height() {
```

base_layer/core/src/base_node/proto/rpc.proto
hansieodendaal: If the peer is honest, it have may increased since the pin-pong. If the peer is dishonest, they can lie about `tip_height` as well.

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
SWvheerden: That is not possible, it is checked above, with a comment git is not showing
https://github.com/tari-project/tari/blob/6e52f447a5f127af4a7523fb0f26684bc9e9173b/base_layer/core/src/base_node/sync/header_sync/synchronizer.rs#L442C11-L443C79

base_layer/core/src/base_node/proto/rpc.proto
SWvheerden: well then lets remove that as well, all this does, is its asking the same info twice, if the peer already lied, why lie again. 

If the difficulty went up, it has no effect on the rest of the code. If this data is not required, let's remove it, it adds extra comms data, extra code complexity.

hansieodendaal: Removed

base_layer/core/src/base_node/sync/header_sync/synchronizer.rs
hansieodendaal: Simplified - `find_chain_split` will return the `chain_split_hash `

SWvheerden:
Fixes: #5816

base_layer/core/src/base_node/sync/header_sync/error.rs
```diff
@@ -56,6 +56,8 @@ pub enum BlockHeaderSyncError {
     ConnectivityError(#[from] ConnectivityError),
     #[error("Node is still not in sync. Sync will be retried with another peer if possible.")]
     NotInSync,
+    #[error("Internal data inconsistency found `{0}")]
```
SWvheerden: I dont believe this is used anymore?

base_layer/core/tests/helpers/sync.rs
```diff
@@ -0,0 +1,261 @@
...
+            },
+        }
+        node.blockchain_db.write(txn).unwrap();
+        // Note: Something is funny here... the block is deleted but the block exists in the db
+        // match instruction {
+        //     WhatToDelete::Blocks => {
+        //         assert!(!node.blockchain_db.block_exists(*blocks[i].hash()).unwrap());
+        //     }
+        //     WhatToDelete::Headers => {}
+        // }
```
SWvheerden: delete?

base_layer/core/tests/tests/header_sync.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+    // All is good, Bob is not banned
+    assert!(!sync::wait_for_is_peer_banned(&alice_node, bob_node.node_identity.node_id(), 1).await);
+
+    // Bob adds more blocks and draws ahead of Alice
+    let _bob_blocks =
+        sync::create_and_add_some_blocks(&bob_node, &bob_blocks[1], 2, &consensus_manager, &key_manager, &[3; 2]).await;
+    assert_eq!(bob_node.blockchain_db.get_height().unwrap(), 4);
+
+    // Alice attempts header sync, on a higher chain with less POW, headers will be lagging with reorg steps
```
SWvheerden: wait, This is unclear
If Bob's chain has more blocks, but less pow, Alice should not be lagging, or try syncing?

base_layer/core/tests/tests/header_sync.rs
```diff
@@ -0,0 +1,373 @@
...
+                .unwrap()
+                .accumulated_difficulty()
+    );
+    let event = sync::sync_headers_execute(&mut alice_state_machine, &mut header_sync).await;
+    match event {
+        StateEvent::HeadersSynchronized(_val, sync_result) => {
+            assert_eq!(sync_result.headers_returned, 0);
+            assert_eq!(sync_result.peer_fork_hash_index, 3);
+            if let HeaderSyncStatus::InSyncOrAhead = sync_result.header_sync_status {
+                // Note: This behaviour is undetected! Bob cannot be banned.
```
SWvheerden:
```suggestion
                // Note: This behaviour is undetected! Bob cannot be banned here, and should be banned by block sync
```
This test and comment makes it seem like this is a bug and a viable attack. We should probably add a test for block sync where we test this.

SWvheerden (commented):
Looking good

base_layer/core/tests/helpers/sync.rs
hansieodendaal: No, I would like us to keep it here, would like to fix it soon, but it is not urgent right away.

base_layer/core/tests/tests/header_sync.rs
hansieodendaal: Agreed, it is a viable attack, and cannot be detected by header sync alone

base_layer/core/tests/helpers/sync.rs
SWvheerden: Can you add a comment exactly saying it there as well, so we know why its commented out and why we left it in

SWvheerden (approved):

PR #5843, opened 2023-10-18
fix: don't ban a peer for sending a banned peer

stringhandler:
Description
---
Removes a check that would ban a peer if they sent a peer that we have banned. 

Motivation and Context
---
The check would ban the source peer during peer sync if they sent a peer to us that we have banned. This is IMO exploitable, since a peer may not have connected to the banned peer in question and thus not know if it should be banned or not. 

You could also get a peer banned in this manner:
1. As an evil peer, I connect to a target node A. 
2. I then connect to a number of other nodes and get myself banned. 
3. The nodes I connect to then will ban node A when they send my node address during sync

How Has This Been Tested?
---
CI

---
Not sure it's quite a mission to replicate this.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/dht/src/discovery/service.rs
+++ b/comms/dht/src/discovery/service.rs
@@ -282,12 +282,6 @@ impl DhtDiscoveryService {
     ) -> Result<T, DhtPeerValidatorError> {
         match result {
             Ok(peer) => Ok(peer),
-            // Banned is an interesting case - if the peer is banned and we reban them, it will modify the original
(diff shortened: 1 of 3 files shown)


SWvheerden (approved):

sdbondi (approved):
Nice - so this was handled correctly discovery response but not in peer sync, think makes sense that the peer validator doesn't care that the peer is banned and that even a banned peer can have its peer record updated (as long as it stays banned)

PR #5844, opened 2023-10-18
fix: ban peer if it sends bad liveness data

stringhandler:
Description
---
Bans the peer if they sent invalid liveness data.

Motivation and Context
---
I think there was a change to the chainmetadata struct recently, and this resulted in a bunch of warnings on my igor node. This PR now bans the node if they sent data that doesn't deserialize.

How Has This Been Tested?
---
Tested with igor trying to connect to old nodes

---
Try to connect to an igor node running 0.52.0-pre.1

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/chain_metadata_service/service.rs
+++ b/base_layer/core/src/base_node/chain_metadata_service/service.rs
@@ -86,38 +86,32 @@ impl ChainMetadataService {
 
         loop {
             tokio::select! {
-                           Ok(block_event) = block_event_stream.recv() => {
+                Ok(block_event) = block_event_stream.recv() => {
+                    log_if_error!(
+                        level: warn,
+                        target: LOG_TARGET,
+                        "Failed to handle block event because '{}'",
+                        self.handle_block_event(&block_event).await
+                    );
+                },
+
+                Ok(event) = liveness_event_stream.recv() => {
+                    match
+                        self.handle_liveness_event(&event).await {
+                        Ok(_) => {}
+                        Err(e) => {
+                           error!( target: LOG_TARGET, "Failed to handle liveness event because '{}'", e);
+                           if let ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id,reason) = e {
                                log_if_error!(
-                                   level: warn,
-                                   target: LOG_TARGET,
-                                   "Failed to handle block event because '{}'",
-                                   self.handle_block_event(&block_event).await
-                               );
-                           },
+                                 level: warn,
+                                 target: LOG_TARGET, "Failed to ban node '{}'",
+                                 self.connectivity.ban_peer_until(node_id, Duration::from_secs(60), reason).await);                                           }
+                        }
+                    }
 
-                           Ok(event) = liveness_event_stream.recv() => {
-                               match
-                                   self.handle_liveness_event(&event).await {
-                                   Ok(_) => {}
-                                   Err(e) => {
-                                           error!( target: LOG_TARGET,
-                                   "Failed to handle liveness event because '{}'", e);
-                                       match e {
-            ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id,reason)  => {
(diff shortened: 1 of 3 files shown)


base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -85,19 +88,27 @@ impl ChainMetadataService {
             tokio::select! {
                 Ok(block_event) = block_event_stream.recv() => {
                     log_if_error!(
-                        level: debug,
+                        level: warn,
```
brianp: Maybe `info` ?

From an outsiders perspective there's a high likelihood at any time they could receive data with unserailizable information but showing them big red warnings in the console makes them feel like they need to do something, when in fact this is a perfectly fine event to happen, and there's not actually any user required interaction that needs to occur.

base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -85,19 +88,27 @@ impl ChainMetadataService {
...
-                        self.handle_liveness_event(&event).await
-                    );
+                    match
+                        self.handle_liveness_event(&event).await {
+                        Ok(_) => {}
+                        Err(e) => {
+                           error!( target: LOG_TARGET, "Failed to handle liveness event because '{}'", e);
+                           if let ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id,reason) = e {
+                               log_if_error!(
+                                 level: warn,
```
brianp: Same here. Banning a peer is informational not deserving of anything alerting to an end user.

base_layer/core/src/base_node/chain_metadata_service/service.rs
```diff
@@ -85,19 +88,27 @@ impl ChainMetadataService {
...
-                        target: LOG_TARGET,
-                        "Failed to handle liveness event because '{}'",
-                        self.handle_liveness_event(&event).await
-                    );
+                    match
+                        self.handle_liveness_event(&event).await {
+                        Ok(_) => {}
+                        Err(e) => {
+                           error!( target: LOG_TARGET, "Failed to handle liveness event because '{}'", e);
+                           if let ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id,reason) = e {
```
brianp, line 104: nit, i'm surprised rustfmt didn't fix this.
```suggestion
                           if let ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id, reason) = e {
```

brianp (changes requested):
Tested with expected results. My node banned existing older nodes.
image

Would like to change the log level before merge.

Ack

base_layer/core/src/base_node/chain_metadata_service/service.rs
stringhandler: rustfmt can't seem to format anything in a `tokio::select` I've tried to make it neat manually, but it's the wild west of fmtland

base_layer/core/src/base_node/chain_metadata_service/service.rs
stringhandler: I'll meet you halfway with `info`. I think it is an indication that you are using the wrong build or network

sdbondi (approved):

SWvheerden (approved):

hansieodendaal (commented):
I agree with the logic, but is it possible to have an integration unit test for the `if let ChainMetadataSyncError::ReceivedInvalidChainMetadata(node_id,reason) = e` branch in `ChainMetadataService::run`?

base_layer/core/src/base_node/chain_metadata_service/service.rs
AaronFeickert: [This issue](https://github.com/tokio-rs/tokio/issues/5441) on it was closed.

PR #5846, opened 2023-10-19
fix: fix erroneous warning message

hansieodendaal:
Description
---
Fixed erroneous warning message about 'orphans_db' and 'orphan_header_accumulated_data_db' being out of sync - the difference was assumed instead of verified.

Motivation and Context
---
See above

How Has This Been Tested?
---
System-level test

---
Code review

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (commented):
I'm not sure why this solves the problem. Stylistically, this `if` state was already complicated and this adds another level to it. Can we maybe clean up the method to have fewer nested `if`s?

stringhandler (approved):

PR #5849, opened 2023-10-20
feat(chatffi): get conversationalists

brianp:
Description
---
Add a feature to fetch a list of TariAddresses that a client currently has conversations with. This is any address a message has been sent to or received from.

Motivation and Context
---
Without being about to fetch a list of addresses a client has conversations with, the client can't actually fetch any independent conversations, as conversations are fetched with an address id.

How Has This Been Tested?
---
New test additions

---
Changes are pretty standard throughout. Handler/Service/DB updates.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/contacts/src/chat_client/src/client.rs
```diff
@@ -194,6 +195,18 @@ impl ChatClient for Client {
...
         message
     }
+
+    async fn get_conversationalists(&self) -> Vec<TariAddress> {
+        let mut addresses = vec![];
+        if let Some(mut contacts_service) = self.contacts.clone() {
+            addresses = contacts_service
+                .get_conversationalists()
+                .await
+                .expect("Addresses from conversations not fetched");
```
SWvheerden, line 205: Please change this to either return an error, or return an empty vec on error.

brianp: I didn't handle the error here specifically because every function in this client needs to have its errors managed. I can change just this one if you want, but the last task for chat is to clean up this class to remove panic potential, and anywhere else through the contacts service. Which I planned to do in it's own PR.

base_layer/chat_ffi/src/conversationalists.rs
```diff
@@ -0,0 +1,186 @@
...
+/// `client` - The ChatClient
+/// `error_out` - Pointer to an int which will be modified
+///
+/// ## Returns
+/// `*mut ptr ConversationalistsVector` - a pointer to a ConversationalistsVector
+///
+/// ## Safety
+/// The `ConversationalistsVector` should be destroyed after use
+#[no_mangle]
+pub unsafe extern "C" fn get_conversationalists(
```
hansieodendaal, line 49: I am missing the create and destroy test for this function in `fn test_retrieving_conversationalists_from_vector`

hansieodendaal (commented):
Looks good, just one comment ^^

base_layer/chat_ffi/src/conversationalists.rs
brianp: The destroy is used in the `test_retrieving_conversationalists_from_vector`.

As for a `get_conversationlists` test it gets harder because it requires a whole running client/db/service etc. Which I haven't setup for any of the unit testing in chat. It is covered in cucumber though.

PR #5863, opened 2023-10-24
fix: db resize issues

SWvheerden:
Description
---
Fixes a resize issue with horizon sync mode

Motivation and Context
---
Lmdb only resizes on transactions, and not pure table changes in our code, this changes the tip smt to always use the transactions, as this call might be large. 

How Has This Been Tested?
---
manual

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5868,7 +5868,6 @@ dependencies = [
  "log",
  "rand",
  "serde",
- "serde_derive",
  "serde_json",
  "tari_common",
  "tari_crypto",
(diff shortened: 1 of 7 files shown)


brianp (commented):
I'm not sure I follow the changes.

We removed the implementations of this functions from other databases backends (lmdb) and now only rely on the chain storage for this data?

Do we need to clean out anything from lmdb if we're no longer using it?

SWvheerden:
So the DB call only makes direct changes to the table. 
It now uses the transaction call, which intercepts the error `db resize required` and resizes the db, and tries again. The insert_tip_smt is a dangerous call since it does not do the resize, so I just removed it.

stringhandler (approved):

PR #5864, opened 2023-10-24
chore: few small code improvements

SWvheerden:
Description
---
Ensures that both header_sync and its rpc methods use the same constants. 
Improves the code readability 

Motivation and Context
---
These changes where discussed and brought up during the hazop code meeting


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -728,7 +728,7 @@ where B: BlockchainBackend
     ) -> Result<TargetDifficulties, ChainStorageError> {
         let db = self.db_read_access()?;
         let mut current_header = db.fetch_chain_header_in_all_chains(&current_block_hash)?;
-        let mut targets = TargetDifficulties::new(&self.consensus_manager, current_header.height() + 1)
+        let mut targets = TargetDifficulties::new(&self.consensus_manager, current_header.height().saturating_add(1))
```
brianp, line 731: Good catch

brianp (approved):
LGTM utAck

PR #5866, opened 2023-10-24
chore: fix spend test

### LINKED ISSUE #5815: Wrong test for double spend
The Tari testcase that checks for double spends is wrong, as it actually tests for
duplicated excess factors and does not consider that an input can be reused while
changing the excess factor of the transaction.
The test is named it_checks_double_spends:

```
async fn it_checks_double_spends() {
let (mut blockchain, validator) = setup(true);
let (_, coinbase_a) = blockchain.add_next_tip(block_spec!
("A")).await.unwrap();
let (txs, _) =
schema_to_transaction(&[txn_schema!(from: vec![coinbase_a], to:
vec![50 * T])], &blockchain.km).await;
blockchain
.add_next_tip(block_spec!("1", transactions: txs.iter().map(|t|
(**t).clone()).collect()))
.await
.unwrap();
let (block, _) = blockchain
.create_next_tip(
BlockSpec::new()
.with_transactions(txs.iter().map(|t|
(**t).clone()).collect())
.finish(),
)
.await;
let txn = blockchain.db().db_read_access().unwrap();
let err = validator.validate_body(&*txn,
block.block()).unwrap_err();
assert!(matches!(err, ValidationError::DuplicateKernelError(_)));
}
```

The issue is that the test uses the txs transactions twice, adding them to the
blockchain and then trying to validate a new block with the same transactions
again. This results in a DuplicateKernelError. But to correctly test for a double
spend, the inputs used are more important than the excess factor: an input can be
reused and added to other inputs to change the excess factor while still resulting
in a double spend.
No tests that check for correct error ContainsSTxO are in the project.

SWvheerden:
Description
---
Fixes spend test

Motivation and Context
---
double spend test should test for the correct error type

How Has This Been Tested?
---
unit tests

Fixes: #5815

### DIFF
--- a/base_layer/core/src/validation/block_body/test.rs
+++ b/base_layer/core/src/validation/block_body/test.rs
@@ -269,8 +269,11 @@ async fn it_checks_double_spends() {
     let (mut blockchain, validator) = setup(true);
 
     let (_, coinbase_a) = blockchain.add_next_tip(block_spec!("A")).await.unwrap();
-    let (txs, _) =
-        schema_to_transaction(&[txn_schema!(from: vec![coinbase_a.clone()], to: vec![50 * T])], &blockchain.km).await;
+    let (txs, _) = schema_to_transaction(
+        &[txn_schema!(from: vec![coinbase_a.clone()], to: vec![50 * T])],
+        &blockchain.km,
+    )
+    .await;
 
     blockchain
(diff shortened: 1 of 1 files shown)


brianp (approved):
LGTM utAck

PR #5871, opened 2023-10-25
feat: ban bad block-sync peers

hansieodendaal:
Description
---
- Added a check to ban a misbehaving peer after block sync when not supplying any or all of the blocks corresponding to the accumulated difficulty they claimed they had.
- Added a check in the RPC block-sync server method not to try and supply blocks if it does not have both blocks corresponding to the start and end hash in its chain.
- Moved all block sync RPC errors to the short ban category from the no=ban cetegrory.
- Added happy path and ban integration-level unit tests for block sync.

Motivation and Context
---
The new unit tests that were added highlighted some issues where sync peers are not banned for their bad behaviour.

How Has This Been Tested?
---
Added new integration-level unit tests.

---
- Code walk through.
- Review and run unit tests .

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/sync/rpc/service.rs
+++ b/base_layer/core/src/base_node/sync/rpc/service.rs
@@ -147,6 +147,11 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
         if db.fetch_block_by_hash(hash, true).await.is_err() {
             return Err(RpcStatus::not_found("Requested end block sync hash was not found"));
         }
+
+        if start_height > metadata.height_of_longest_chain() {
+            return Ok(Streaming::empty());
+        }
+
         let end_header = db
             .fetch_header_by_block_hash(hash)
             .await
--- a/base_layer/core/src/base_node/sync/rpc/service.rs
+++ b/base_layer/core/src/base_node/sync/rpc/service.rs
@@ -120,6 +120,9 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
             .start_hash
             .try_into()
             .map_err(|_| RpcStatus::bad_request(&"Malformed starting hash received".to_string()))?;
+        if db.fetch_block_by_hash(hash, true).await.is_err() {
+            return Err(RpcStatus::not_found("Requested start block sync hash was not found"));
+        }
         let start_header = db
             .fetch_header_by_block_hash(hash)
             .await
@@ -137,13 +140,13 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
             )));
         }
 
-        if start_height > metadata.height_of_longest_chain() {
-            return Ok(Streaming::empty());
-        }
         let hash = message
             .end_hash
             .try_into()
             .map_err(|_| RpcStatus::bad_request(&"Malformed end hash received".to_string()))?;
+        if db.fetch_block_by_hash(hash, true).await.is_err() {
+            return Err(RpcStatus::not_found("Requested end block sync hash was not found"));
+        }
         let end_header = db
             .fetch_header_by_block_hash(hash)
             .await
--- a/base_layer/core/src/base_node/sync/rpc/service.rs
+++ b/base_layer/core/src/base_node/sync/rpc/service.rs
@@ -148,10 +148,6 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
             return Err(RpcStatus::not_found("Requested end block sync hash was not found"));
         }
 
-        if start_height > metadata.height_of_longest_chain() {
-            return Ok(Streaming::empty());
-        }
-
         let end_header = db
             .fetch_header_by_block_hash(hash)
             .await
--- a/base_layer/core/src/base_node/sync/rpc/tests.rs
+++ b/base_layer/core/src/base_node/sync/rpc/tests.rs
@@ -75,7 +75,7 @@ mod sync_blocks {
     }
 
     #[tokio::test]
-    async fn it_sends_an_empty_response() {
+    async fn it_does_not_send_empty_responses_with_bad_requests() {
         let (service, db, rpc_request_mock, _tmp) = setup();
 
         let (_, chain) = create_main_chain(&db, block_specs!(["A->GB"])).await;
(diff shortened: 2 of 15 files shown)


base_layer/core/src/base_node/sync/rpc/service.rs
```diff
@@ -137,13 +140,18 @@ impl<B: BlockchainBackend + 'static> BaseNodeSyncService for BaseNodeSyncRpcServ
...
             .end_hash
             .try_into()
             .map_err(|_| RpcStatus::bad_request(&"Malformed end hash received".to_string()))?;
+        if db.fetch_block_by_hash(hash, true).await.is_err() {
+            return Err(RpcStatus::not_found("Requested end block sync hash was not found"));
+        }
+
+        if start_height > metadata.height_of_longest_chain() {
+            return Ok(Streaming::empty());
+        }
```
SWvheerden:
```suggestion
```
We can remove this, it wont ever get to this. 
We added checks for the start and end blocks, so we know we have them in our chain.

base_layer/core/tests/helpers/sync.rs
```diff
@@ -149,44 +165,48 @@ pub fn delete_some_blocks_and_headers(
...
+            for block in &blocks[1..] {
+                if node.blockchain_db.block_exists(*block.hash()).unwrap() {
+                    let mut txn = DbTransaction::new();
+                    txn.delete_orphan(*block.hash());
+                    node.blockchain_db.write(txn).unwrap();
+                }
+                assert!(!node.blockchain_db.block_exists(*block.hash()).unwrap());
+            }
+            node.blockchain_db.insert_valid_headers(headers[1..].to_vec()).unwrap();
+            // Note: This seems funny, as inserting the headers back into the db will cause this test to fail
```
SWvheerden:
```suggestion
            // Note: This seems funny, as inserting the headers back into the db will cause this test to fail
        //       This should be investigated and fixed as it will enhance the tests. If we uncomment the following assertion, the
        //       tests depending on this function will fail.
```

base_layer/core/tests/tests/header_sync.rs
```diff
@@ -267,7 +267,7 @@ async fn test_header_sync_uneven_headers_and_blocks_peer_lies_about_pow_no_ban()
     match event {
         StateEvent::HeadersSynchronized(_val, sync_result) => {
             assert_eq!(sync_result.headers_returned, 0);
-            assert_eq!(sync_result.peer_fork_hash_index, 3);
+            assert_eq!(sync_result.peer_fork_hash_index, 5);
```
SWvheerden: whats going here, why did this value change? To me 3 sounds like the correct number here?

base_layer/core/src/base_node/sync/rpc/service.rs
hansieodendaal: There is a specific unit test for it (`async fn it_sends_an_empty_response()`), that tests if empty is returned if end < start where both hashes are valid, so I opted to keep it in.

_After discussion, I agree we can remove this behaviour and just let the server return an error._

base_layer/core/src/base_node/sync/rpc/tests.rs
```diff
@@ -75,7 +75,7 @@ mod sync_blocks {
     }
 
     #[tokio::test]
-    async fn it_sends_an_empty_response() {
+    async fn it_does_not_send_empty_responses_with_bad_requests() {
```
SWvheerden:
```suggestion
    async fn is_sends_bad_request+on_bad_response() {
```
In a year we wont know that we sent and empty response and we will wonder why this test exists and what it tests

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1083,9 +1083,9 @@ where B: BlockchainBackend
     }
 
     /// Returns true if this block exists in the chain, or is orphaned.
-    pub fn block_exists(&self, hash: BlockHash) -> Result<bool, ChainStorageError> {
+    pub fn chain_block_or_orphan_block_exists(&self, hash: BlockHash) -> Result<bool, ChainStorageError> {
         let db = self.db_read_access()?;
-        Ok(db.contains(&DbKey::BlockHash(hash))? || db.contains(&DbKey::OrphanBlock(hash))?)
+        Ok(db.fetch_block_accumulated_data(&hash)?.is_some() || db.contains(&DbKey::OrphanBlock(hash))?)
```
SWvheerden, line 1089:
```suggestion
// we need to check if the block accumulated data exists, and the header might exist without a body
        Ok(db.fetch_block_accumulated_data(&hash)?.is_some() || db.contains(&DbKey::OrphanBlock(hash))?)
```

base_layer/core/tests/helpers/sync.rs
```diff
@@ -140,53 +155,77 @@ pub async fn create_network_with_local_and_peer_nodes() -> (
...
     node: &NodeInterfaces,
-    set_best_block: Option<bool>,
 ) {
     if blocks_with_anchor.is_empty() || blocks_with_anchor.len() < 2 {
         panic!("blocks must have at least 2 elements");
     }
-    let set_best_block = set_best_block.unwrap_or(false);
     let mut blocks: Vec<_> = blocks_with_anchor.to_vec();
     blocks.reverse();
     for i in 0..blocks.len() - 1 {
```
SWvheerden, line 188:
```suggestion
    for i in (0..blocks.len() - 1).rev() {
```

base_layer/core/tests/helpers/sync.rs
```diff
@@ -140,53 +155,77 @@ pub async fn create_network_with_local_and_peer_nodes() -> (
...
     instruction: WhatToDelete,
     node: &NodeInterfaces,
-    set_best_block: Option<bool>,
 ) {
     if blocks_with_anchor.is_empty() || blocks_with_anchor.len() < 2 {
         panic!("blocks must have at least 2 elements");
     }
-    let set_best_block = set_best_block.unwrap_or(false);
     let mut blocks: Vec<_> = blocks_with_anchor.to_vec();
     blocks.reverse();
```
SWvheerden, line 187:
```suggestion
```

base_layer/core/tests/helpers/sync.rs
SWvheerden: nit: if just do this, you don't have to swap around the entire array, and the code above is more direct what's going on. Because it seems strange that you delete a block, and then you set the next block as the best block

SWvheerden (commented):
Looking good, I want us to add a comment and rename a test, the other one is a nit

base_layer/core/src/base_node/sync/rpc/tests.rs
```diff
@@ -75,7 +75,7 @@ mod sync_blocks {
     }
 
     #[tokio::test]
-    async fn it_sends_an_empty_response() {
+    async fn is_sends_bad_request+on_bad_response() {
```
SWvheerden:
```suggestion
    async fn is_sends_bad_request_on_bad_response() {
```

SWvheerden (approved):

PR #5872, opened 2023-10-25
fix(tariscript): protect compare and check height from underflows

### LINKED ISSUE #5813: tariscript OP_COMPARE_VERIFY can crash with overflow
An attacker can craft a transaction with a malicious script that crashes any node
which attempts to execute it.
The crash is triggered by an underflow in the OP_COMPARE_VERIFY opcode:
```
fn handle_compare_height(stack: &mut ExecutionStack, block_height:
u64) -> Result<(), ScriptError> {
let target_height = stack.pop_into_number::<i64>()?;
let block_height = i64::try_from(block_height)?;
let item = StackItem::Number(block_height - target_height);
stack.push(item)
}
```
An attacker creates a script which puts into the stack the i64::MIN. Any
subtraction will then cause the underflow.
`let script = TariScript::new(vec![PushInt(i64::MIN), CompareHeight]);`

brianp:
Description
---
This PR corrects the operand position of the `handle_compare_height` function. This was identified as incorrect based on the op code description.
Additionally, protect it from underflows, and write test cases against all possible returns and errors.

I added underflow protection in the `handle_check_height` function as well but after some thought this function wouldn't currently be possible to underflow. The values passed into the function would never be lower than 0, and then converted to i64 which would handle subtraction from 0 fine. In opposition, the `handle_compare_height` uses a value from the stack that could be i64::MIN and then have i64::MAX subtracted from it. Which was the originally identified problem. This is all to say the functions aren't quite equal so it felt worth a comment for future readers.

Motivation and Context
---
Closes #5813 

How Has This Been Tested?
---
Added new tests.

---
Double check the [docs](https://github.com/tari-project/tari/blob/development/infrastructure/tari_script/src/op_codes.rs#L146) to validate the operand switch is a valid change.

See the new underflow protection. 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/infrastructure/tari_script/src/error.rs
+++ b/infrastructure/tari_script/src/error.rs
@@ -49,6 +49,8 @@ pub enum ScriptError {
     VerifyFailed,
     #[error("as_hash requires a Digest function that returns at least 32 bytes")]
     InvalidDigest,
+    #[error("A compare opcode failed, aborting the script immediately")]
+    CompareFailed,
 }
 
 impl From<TryFromIntError> for ScriptError {
--- a/infrastructure/tari_script/src/error.rs
+++ b/infrastructure/tari_script/src/error.rs
@@ -49,8 +49,8 @@ pub enum ScriptError {
     VerifyFailed,
     #[error("as_hash requires a Digest function that returns at least 32 bytes")]
     InvalidDigest,
-    #[error("A compare opcode failed, aborting the script immediately")]
-    CompareFailed,
+    #[error("A compare opcode failed, aborting the script immediately with reason: `{0}`")]
+    CompareFailed(String),
 }
 
 impl From<TryFromIntError> for ScriptError {
--- a/infrastructure/tari_script/src/script.rs
+++ b/infrastructure/tari_script/src/script.rs
@@ -346,7 +346,11 @@ impl TariScript {
         // This differs from compare_height due to a stack number being used, which can be lower than 0
         let item = match block_height.checked_sub(height) {
             Some(num) => StackItem::Number(num),
-            None => return Err(ScriptError::CompareFailed),
+            None => {
+                return Err(ScriptError::CompareFailed(
+                    "Subtraction of given height from current block height failed".to_string(),
+                ))
+            },
         };
 
         stack.push(item)
@@ -370,7 +374,11 @@ impl TariScript {
         // height does not use a stack number and it's minimum can't be lower than 0.
         let item = match block_height.checked_sub(target_height) {
             Some(num) => StackItem::Number(num),
-            None => return Err(ScriptError::CompareFailed),
+            None => {
+                return Err(ScriptError::CompareFailed(
+                    "Couldn't subtract the target height from the current block height".to_string(),
+                ))
+            },
         };
 
         stack.push(item)
@@ -1757,7 +1765,7 @@ mod test {
         let inputs = ExecutionStack::new(vec![Number(i64::MIN)]);
         let ctx = context_with_height(i64::MAX as u64);
         let stack_item = script.execute_with_context(&inputs, &ctx);
-        assert!(matches!(stack_item, Err(ScriptError::CompareFailed)));
+        assert!(matches!(stack_item, Err(ScriptError::CompareFailed(_))));
     }
 
     #[test]
--- a/infrastructure/tari_script/src/script.rs
+++ b/infrastructure/tari_script/src/script.rs
@@ -340,7 +340,14 @@ impl TariScript {
     fn handle_check_height(stack: &mut ExecutionStack, height: u64, block_height: u64) -> Result<(), ScriptError> {
         let height = i64::try_from(height)?;
         let block_height = i64::try_from(block_height)?;
-        let item = StackItem::Number(block_height - height);
+
+        // Due to the conversion of u64 into i64 which would fail above if they overflowed, these
+        // numbers should never enter a state where a `sub` could fail. As they'd both be within range and 0 or above.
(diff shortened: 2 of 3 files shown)


infrastructure/tari_script/src/script.rs
```diff
@@ -349,7 +349,14 @@ impl TariScript {
...
         let height = i64::try_from(height)?;
         let block_height = i64::try_from(block_height)?;
-        let item = StackItem::Number(block_height - height);
+
+        // Due to the conversion of u64 into i64 which would fail above if they overflowed, these
+        // numbers should never enter a state where a `sub` could fail. As they'd both be within range and 0 or above.
+        // This differs from compare_height due to a stack number being used, which can be lower than 0
+        let item = match block_height.checked_sub(height) {
+            Some(num) => StackItem::Number(num),
+            None => return Err(ScriptError::CompareFailed),
```
SWvheerden:
```suggestion
            None => return Err(ScriptError::StackUnderflow),
```
I would rather return the actual error here:

infrastructure/tari_script/src/script.rs
```diff
@@ -368,7 +375,12 @@ impl TariScript {
         let target_height = stack.pop_into_number::<i64>()?;
         let block_height = i64::try_from(block_height)?;
 
-        let item = StackItem::Number(block_height - target_height);
+        // Here it is possible to underflow because the stack can take lower numbers where check
+        // height does not use a stack number and it's minimum can't be lower than 0.
+        let item = match target_height.checked_sub(block_height) {
+            Some(num) => StackItem::Number(num),
+            None => return Err(ScriptError::CompareFailed),
```
SWvheerden: same here

SWvheerden (commented):
This looks good, just the error type

infrastructure/tari_script/src/script.rs
brianp: At one point I had the same thought but the `StackUnderflow` isn't a general `Underflow` error. I believe `StackUnderflow` is a specific error to the Input Stack underflowing during script execution. Which isn't the error here.

We could create an alternative error though. Change `CompareFailed` to `Underflow` or `CompareFailed(message)` with a specific message.

SWvheerden: `CompareFailed(message)` is probably the best way forward. Compare failed lets you believe the comparison failed, not that its an underflow.

infrastructure/tari_script/src/script.rs
```diff
@@ -368,7 +375,12 @@ impl TariScript {
         let target_height = stack.pop_into_number::<i64>()?;
         let block_height = i64::try_from(block_height)?;
 
-        let item = StackItem::Number(block_height - target_height);
+        // Here it is possible to underflow because the stack can take lower numbers where check
+        // height does not use a stack number and it's minimum can't be lower than 0.
+        let item = match target_height.checked_sub(block_height) {
```
hansieodendaal: I believe the order is the wrong way around here when compared to the original? Why did it change?
```suggestion
        let item = match block_height.checked_sub(target_height) {
```

infrastructure/tari_script/src/error.rs
```diff
@@ -49,6 +49,8 @@ pub enum ScriptError {
     VerifyFailed,
     #[error("as_hash requires a Digest function that returns at least 32 bytes")]
     InvalidDigest,
+    #[error("A compare opcode failed, aborting the script immediately")]
+    CompareFailed,
```
hansieodendaal:
```suggestion
    MathematicalUnderflow,
```

infrastructure/tari_script/src/script.rs
hansieodendaal: I suggest `MathematicalUnderflow` ^^ to differentaiate from `StackUnderflow`

hansieodendaal (commented):
Hi @brianp, please see the comments below, especially the one in `fn handle_compare_height`.

infrastructure/tari_script/src/script.rs
```diff
@@ -918,7 +930,7 @@ mod test {
             let ctx = context_with_height(u64::try_from(block_height).unwrap());
             assert_eq!(
                 script.execute_with_context(&inputs, &ctx).unwrap(),
-                Number(block_height - 5)
+                Number(5 - block_height)
```
hansieodendaal: I would rather retain the original behaviour

infrastructure/tari_script/src/script.rs
brianp: The previous behaviour is incorrect based on the documentation: 

> /// Pops the top of the stack as `height`, then pushes the value of (`height` - the current height) to the stack.
    /// In other words, this opcode replaces the top of the stack with the difference between `height` and the
    /// current height. Fails with `InvalidInput` if there is not a valid integer value on top of the stack. Fails
    /// with `StackUnderflow` if the stack is empty.

So the question becomes which one is correct, the behaviour, or the documentation. I decided to trust the docs, and correct the behaviour. Do you have a particular reasoning for wanting to do the inverse?

infrastructure/tari_script/src/script.rs
brianp: Updated to `CompareFailed(message)`

infrastructure/tari_script/src/script.rs
brianp: Spoke with @SWvheerden who agrees with you on this @hansieodendaal. So I've reverted the operands to the original positions and updated the op codes docs to match this behaviour.

SWvheerden (approved):

PR #5873, opened 2023-10-25
feat!: add validator mmr size

SWvheerden:
Description
---
Adds in the size of the validator balance tree to the header. 

Motivation and Context
---
The validator size needs to be committed too like the other Merkle roots as well. A peer syncing will have no way of verifying the validator merkle root is it does not know the size. 

How Has This Been Tested?
---
Unit tests

---
ensure its committed to and not malleable.


hansieodendaal (approved):
utACK

stringhandler (dismissed):
Not sure if it's necessary, but rather be safe than sorry

PR #5879, opened 2023-10-27
feat: enable multiple coinbase utxos

SWvheerden:
Description
---
This allows multiple coinbase utxos in a block

Motivation and Context
---
If we want to implement p2pool, implementing multiple coinbases is the easiest way to enable multiple parties to have their own verified UTXOs mined. There are other ways of doing this, but this approach is the simplest and requires the least overall utxos. 

How Has This Been Tested?
---
unit tests

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -29,6 +29,7 @@ use log::*;
 use serde::{Deserialize, Serialize};
 use tari_common_types::types::PrivateKey;
 use tari_crypto::commitment::HomomorphicCommitmentFactory;
+use tari_common_types::types::Commitment;
 
 use crate::transactions::{
     crypto_factories::CryptoFactories,
@@ -240,7 +241,7 @@ impl AggregateBody {
         factories: &CryptoFactories,
         height: u64,
     ) -> Result<(), TransactionError> {
-        let mut coinbase_utxo = None;
+        let mut coinbase_utxo_sum = Commitment::default();
         let mut coinbase_kernel = None;
         let mut coinbase_counter = 0; // there should be exactly 1 coinbase
         for utxo in self.outputs() {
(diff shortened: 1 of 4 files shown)


base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -625,16 +625,6 @@ mod test {
...
 
-        // test catches that coinbase count on the utxo is wrong
-        assert!(matches!(
-            tx.body.check_coinbase_output(
-                block_reward,
-                rules.consensus_constants(0).coinbase_min_maturity(),
-                &factories,
-                42
-            ),
-            Err(TransactionError::MoreThanOneCoinbase)
```
hansieodendaal, line 636: If we allow multiple coinbases, pleas remove this error:
```
TransactionError::MoreThanOneCoinbase
```

base_layer/core/src/validation/block_body/test.rs
```diff
@@ -212,20 +212,6 @@ async fn it_checks_exactly_one_coinbase() {
...
-    let err = {
-        // `MutexGuard` cannot be held across an `await` point
-        let txn = blockchain.db().db_read_access().unwrap();
-        let err = validator.validate_body(&*txn, block.block()).unwrap_err();
-        err
-    };
-    assert!(matches!(
-        err,
-        ValidationError::BlockError(BlockValidationError::TransactionError(
-            TransactionError::MoreThanOneCoinbase
```
hansieodendaal, line 226: The name for this test `async fn it_checks_exactly_one_coinbase()` should change

base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -250,22 +250,18 @@ impl AggregateBody {
                     warn!(target: LOG_TARGET, "Coinbase {} found with maturity set too low", utxo);
                     return Err(TransactionError::InvalidCoinbaseMaturity);
                 }
-                coinbase_utxo = Some(utxo);
+                coinbase_utxo_sum = &coinbase_utxo_sum + &utxo.commitment;
```
hansieodendaal, line 253: Lets add a unit test with say 50 coinbases in one block

hansieodendaal (commented):
Looks good, I would just like us to beef up on the test cases, please.

base_layer/core/src/transactions/aggregated_body.rs
SWvheerden: Existing tests are fine, we don't need to ber extravagant

SWvheerden:
We have enough test cases that already test for duplicate coinbase utxos and duplicate coinbase kernels, they all handle correctly. We don't have to add anything extravagant here atm

Issue #5860, opened 2023-10-23
Reduce timeout of find_chain_split 


SWvheerden:
The current timeout of 30 seconds is very long and can cause a peer to delay the node by a bit,.

PR #5882, opened 2023-10-27
feat: reduce timeouts and increase bans

### LINKED ISSUE #5867: Revise default ban period
We need to revise the default ban period for short and normal bans.

SWvheerden:
Description
---
Reduces the latency of timeouts from 30 seconds to 15 seconds
Increases the ban periods

Motivation and Context
---
Reduces the latency of the timeouts from 30 secs to 15 secs: #5860 

Increase the short ban to 4 mins up from 30 seconds. The short ban is used for timeouts etc. and is used to ensure the base node can select a new node to sync from etc. But if this is too short, or under the block time, timeouts can be used maliciously by nodes. For example a node can say it has a better pow, let your node try and sync to it, wait for the timeouts, get banned for 30 seconds. This means your node waits the 30 seconds -> 1 mins if needs more than 1 call, which is likely. This means you node rejects all new blocks for 1 min, and then you ban the peer for 30 secs. They can then submit their new block to you after the 30 secs. This makes it longer than the block time. This means the bad node is excluded from all new nodes for 2 blocks on avg. 

Fixes: : #5860 
Fixes: #5867

### DIFF
--- a/base_layer/core/src/base_node/sync/config.rs
+++ b/base_layer/core/src/base_node/sync/config.rs
@@ -56,13 +56,13 @@ pub struct BlockchainSyncConfig {
 impl Default for BlockchainSyncConfig {
     fn default() -> Self {
         Self {
-            initial_max_sync_latency: Duration::from_secs(30),
+            initial_max_sync_latency: Duration::from_secs(15),
             max_latency_increase: Duration::from_secs(2),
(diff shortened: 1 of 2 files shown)


SWvheerden:
Need to test and verify these changes on system level once merged in.

hansieodendaal (approved):
utACK

PR #5883, opened 2023-10-27
feat: print warning for wallets in direct send only

### LINKED ISSUE #5861: Set warning if wallet is set to only send via direct send
Its possible to only send transactions via direct send, while this is handy, we should print out a warning on startup as this is not guaranteed successful.

`Empirical testing determined that in some cases a direct send would be reported as true, even though the wallet was offline. Possibly due to the Tor connection remaining active for a few minutes after wallet shutdown.`

SWvheerden:
Description
---
Prints out a warning for wallets set to use direct send only

Motivation and Context
---
See: #5861 

How Has This Been Tested?
---
Manual

Fixes: #5861


hansieodendaal (approved):
ACK

Issue #5858, opened 2023-10-23
Not possible to disable gRPC methods


SWvheerden:
gRPC server is either enabled or disabled, with no way to filter which methods the
node or wallet operator wants to expose.
While this can be resolved by the use of a reverse-proxy, it is recommended that
the  configuration  exists  to  protect  unaware  operators.  In  particular,  certain
methods should be disabled by default unless the operator explicitly requests
them, because they provide information about a node that in most cases should be
private, such as:
```
check_for_updates
get_version
list_connected_peers
get_sync_progress
get_network_status
get_peers
```

Add  the  capability  to  disable  or  enable  gRPC  methods.  Disable  methods  that
provide internal information of the node by default.

hansieodendaal:
See #5886

PR #5886, opened 2023-10-30
feat: add config for grpc server methods

hansieodendaal:
Description
---
Added configuration options for the base node's gRPC methods whereby each method can be enabled or disabled with the startup-config settings.

Closing #5858

Some BloomRPC screenshots:

image

image

Motivation and Context
---
See #5858

How Has This Been Tested?
---
- Pass existing unit tests
- Added unit test `fn it_deserializes_enums`
- Pass cucumber tests on CI
- Manual gRPC queries to the base node (with BloomRPC)

---
- Code walkthrough
- Manual gRPC queries to the base node

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_node/src/config.rs
+++ b/applications/minotari_node/src/config.rs
@@ -87,6 +87,8 @@ pub struct BaseNodeConfig {
     pub grpc_enabled: bool,
     /// GRPC address of base node
     pub grpc_address: Option<Multiaddr>,
+    /// GRPC server config - which methods are active and which not
+    pub grpc_server_config: BaseNodeGrpcServerConfig,
     /// A path to the file that stores the base node identity and secret key
     pub identity_file: PathBuf,
     /// Spin up and use a built-in Tor instance. This only works on macos/linux - requires that the wallet was built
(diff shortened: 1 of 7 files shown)


applications/minotari_node/src/config.rs
```diff
@@ -143,6 +145,44 @@ impl Default for BaseNodeConfig {
...
+                enable_get_mempool_transactions: true,
+                enable_transaction_state: true,
+                enable_list_connected_peers: true,
+                enable_get_mempool_stats: true,
+                enable_get_active_validator_nodes: true,
+                enable_get_shard_key: true,
+                enable_get_template_registrations: true,
+                enable_get_side_chain_utxos: true,
+                // These gRPC server methods share sensitive information, thus disabled by default
+                enable_get_version: false,
```
sdbondi: Why not just have a list of denied methods?

```rust
deny_methods: Vec<GrpcMethod>
```

and implement FromStr for GrpcMethod

sdbondi (commented):
Think the idea is solid, I looked at tonic interceptors and middleware and there will be a way to do it, however tonic makes it very difficult (unless I missed something obvious), so happy to add a conditional to the start of every method.

The implementation adds more maintenance overhead than required. I think adding a list of methods that should be denied will make things easier to maintain.

applications/minotari_node/src/config.rs
hansieodendaal: Thanks for the discussion, implemented

SWvheerden (approved):

PR #5890, opened 2023-10-31
feat: allow multiple initial sync peers

### LINKED ISSUE #5852: Sync is always done with a single peer
Sync is always done with a single peer. 

While this is fine for syncing intermittently where your node is already in sync, it gets a notification from one of its closest connected peers that it's behind, it should sync to that peer. 

On start up this is different, the node should wait a bit and make a bit more informed decision on what peer to sync from and perhaps select a list.

SWvheerden:
Description
---
Allow syncing from multiple peers initially

Motivation and Context
---
See issue: #5852 
This allows a node to wait until it gets 5 ChainMetaData during initial sync until it goes to header sync. 

How Has This Been Tested?
---
Unit tests

Fixes: #5852

### DIFF
--- a/base_layer/core/src/base_node/state_machine_service/states/listening.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/listening.rs
@@ -241,6 +241,7 @@ impl Listening {
                                 if initial_sync_peer_list[i] == peer {
                                     found = true;
                                     initial_sync_peer_list[i] = peer.clone();
+                                    break;
                                 }
                             }
                             if !found {
--- a/base_layer/core/src/base_node/state_machine_service/states/listening.rs
+++ b/base_layer/core/src/base_node/state_machine_service/states/listening.rs
@@ -244,13 +244,13 @@ impl Listening {
                             let mut found = false;
                             // lets search the list list to ensure we only have unique peers in the list with the latest
                             // up-to-date information
-                            for i in 0..initial_sync_peer_list.len() {
+                            for initial_peer in &mut initial_sync_peer_list {
                                 // we compare the two peers via the comparison operator on syncpeer
-                                if initial_sync_peer_list[i] == peer {
+                                if *initial_peer == peer {
                                     found = true;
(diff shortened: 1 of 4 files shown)


base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -216,15 +219,44 @@ impl Listening {
...
+                            for i in 0..initial_sync_peer_list.len() {
+                                if initial_sync_peer_list[i] == peer {
+                                    found = true;
+                                    initial_sync_peer_list[i] = peer.clone();
+                                }
+                            }
+                            if !found {
+                                initial_sync_peer_list.push(peer.clone());
+                            }
+                        }
```
SWvheerden:
```suggestion
                        for peer in sync_peers {
                            let mut found = false;
                            for i in 0..initial_sync_peer_list.len() {
                                if initial_sync_peer_list[i] == peer {
                                    found = true;
                                    initial_sync_peer_list[i] = peer.clone();
                                    break;
                                }
                            }
                            if !found {
                                initial_sync_peer_list.push(peer.clone());
                            }
                        }
```

base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -216,15 +219,45 @@ impl Listening {
...
+                                    break;
+                                }
+                            }
+                            if !found {
+                                initial_sync_peer_list.push(peer.clone());
+                            }
+                        }
+                        // We use a list here to ensure that we dont wait for even for INITIAL_SYNC_PEER_COUNT different
+                        // peers
+                        if initial_sync_counter == INITIAL_SYNC_PEER_COUNT {
```
SWvheerden:
```suggestion
                        if initial_sync_counter >= INITIAL_SYNC_PEER_COUNT {
```

hansieodendaal (approved):
utACK

(jointly reviewed)

PR #5896, opened 2023-11-01
feat: adaptable min difficulty check

SWvheerden:
Description
---
This includes an adaptable min difficulty check for incoming blocks. 

Motivation and Context
---
See issue: #5620 for the reasoning behind why this needs to include an adaptable min difficulty check.

### DIFF
--- a/base_layer/core/src/proof_of_work/difficulty.rs
+++ b/base_layer/core/src/proof_of_work/difficulty.rs
@@ -84,12 +84,16 @@ impl Difficulty {
         Difficulty::from_u64(result.low_u64())
     }
 
-    pub fn min_difficulty_checked_div(&self, other: u64) -> Option<Difficulty> {
-        return match self.0.checked_div(other) {
+    pub fn checked_div_u64(&self, other: u64) -> Option<Difficulty> {
+        match self.0.checked_div(other) {
             None => None,
             Some(n) => {
-                n < MIN_DIFFICULTY ? None : Some(Difficulty(n))
-            }
+                if n < MIN_DIFFICULTY {
+                    None
+                } else {
+                    Some(Difficulty(n))
+                }
+            },
         }
     }
 }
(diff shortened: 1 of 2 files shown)


base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -83,6 +83,18 @@ impl Difficulty {
...
+        if other == 0 {
+            return None;
+        }
+        let result = Difficulty(self.0 / other);
+        if result.0 < MIN_DIFFICULTY {
+            None
+        } else {
+            Some(result)
+        }
+    }
```
brianp: There is a `checked_div` function for u64 already. It does divide by 0 checking. Is there a reason not to use it?
https://doc.rust-lang.org/stable/std/primitive.u64.html#method.checked_div

Additionally, this function is doing more than a standard `checked_*` function. We're comparing it against our own min difficulty. So we could go for a name that doesn't match a pre-existing usage pattern.

```suggestion
    pub fn min_difficulty_checked_div(&self, other: u64) -> Option<Difficulty> {
        return match self.0.checked_div(other) {
            None => None,
            Some(n) => {
                n < MIN_DIFFICULTY ? None : Some(Difficulty(n))
            }
        }
    }
```

brianp: Additionally, the `MIN_DIFFICULTY` constant is well... Constant. Shouldn't this be compared against the current minimum proof of work difficulty?

Is there a chance someone could use this to always achieve the minimum difficulty of 1? n could be the same as the constant which would result in a min difficulty lower than the pow difficulty.

brianp (changes requested):
Some suggestions for the `checked_div_u64` functions. I probably need a clarification on the way we're using the min difficulty we're calculating because it looks like it could be possible to keep the value at 1.

base_layer/core/src/proof_of_work/difficulty.rs
SWvheerden: Min_difficulty is a constant, and that why we need to keep checking it against that. And the context here keeps it to difficulty, and does not go towards having a min_chain difficulty that dependant on previous timestamps etc

brianp (approved):
LGTM, utAck

PR #5900, opened 2023-11-02
feat!: update key parsing

SWvheerden:
Description
---
Updates key parsing to differentiate between canonical- and reduction-based use cases. Makes several updates to account for `EpochTime` changes. Updates genesis data.

Motivation and Context
---
An [update](https://github.com/tari-project/tari-crypto/pull/194) to `tari-crypto` differentiates secret key parsing between cases that should use canonical byte arrays, and those that use random byte arrays requiring modular wide reduction.

This PR makes corresponding changes to account for this.

It also makes several updates required since the `tari_utilities` release used here also makes breaking `EpochTime` changes that can't be cleanly done separately.

How Has This Been Tested?
---
Existing tests pass. Uses of parsed secret keys have been manually inspected for correctness.

---
Assert that secret keys and scalars instantiated from byte arrays use the appropriate approach: parsing from canonical byte arrays, or modular wide reduction from uniform byte arrays.

Confirm that `EpochTime` changes reflect the intended logic, especially relating to timestamp and difficulty handling.

BREAKING CHANGE: This modifies how secret keys and scalars are constructed and is therefore a breaking change.


SWvheerden:
This is a replacement PR for : https://github.com/tari-project/tari/pull/5636

It now includes the official tari_crypto release tag 0.19.0 instead of github main branch

PR #5902, opened 2023-11-02
feat: add constant time comparison for grpc authentication

### LINKED ISSUE #5810: GRPC - Wallet username vulnerable to timing attacks
The validate method does not use a constant-time compare function to check
usernames. This makes it possible for an attacker to analyze the time it takes for
the method to return in order to optimize their bruteforce attack.

```
if self.user_name.as_bytes() != username.as_bytes() {
return Err(BasicAuthError::InvalidUsername);
}
```

The absence of constant-time comparison permits an attacker to infer details
about the correct username by observing how long the system takes to respond to
various inputs. Consequently, an attacker can exploit this vulnerability to perform a
side-channel timing attack, systematically guessing the username and analyzing
the time taken for the system to respond. Over repeated attempts, this could
allow an attacker to reconstruct the correct username.

Implement a constant-time comparison method for usernames to prevent sidechannel attacks.

hansieodendaal:
Description
---
- Added constant-time username comparison for gRPC authentication. This will largely mitigate side-channel attacks to uncover the gRPC username. (See `BasicAuthCredentials::constant_time_compare_username`)

- **Edit:** Credential validation for the combined username and password will now also run in constant time and not return if the username did not match as it did previously.

- **Edit:** Fixed an issue where the `BasicAuthCredentials` from header did not pass validation, impacted in `fn it_generates_a_valid_header()` and `it_decodes_from_well_formed_header`.

(Closes #5810)

Motivation and Context
---
See #5810

How Has This Been Tested?
---
- Added two unit tests to compare constant time performance with varying length username guesses. The unit test was performed in release mode with a no-load and fully loaded CPU.
- `fn it_compares_user_names_in_constant_time()` 
```
        // This unit test asserts that the minimum variance is less than 10% (chosen to be robust for running the unit
        // test with CI), indicating that the function behaves within acceptable constant-time constraints.
        //
        // Some consecutive results running in release mode on a Core i7-12700H (with no other processes running):
        //
        // Minimum variance:                          0.247 %
        // Average variance:                          4.65738 %
        // Average short username time:               1.17486 microseconds
        // Average long username time:                1.17344 microseconds
        // Average actual username time:              1.18388 microseconds
        //
        // Minimum variance:                          0.10214 %
        // Average variance:                          4.32226 %
        // Average short username time:               1.1619 microseconds
        // Average long username time:                1.16591 microseconds
        // Average actual username time:              1.18157 microseconds
        //
        // Minimum variance:                          0.17953 %
        // Average variance:                          5.51519 %
        // Average short username time:               1.17974 microseconds
        // Average long username time:                1.19232 microseconds
        // Average actual username time:              1.18709 microseconds
        //
        // Some consecutive results running in release mode on a Core i7-12700H (while entire CPU fully stressed):
        //
        // Minimum variance:                          0.60357 %
        // Average variance:                          6.30167 %
        // Average short username time:               1.81708 microseconds
        // Average long username time:                1.77562 microseconds
        // Average actual username time:              1.74824 microseconds
        //
        // Minimum variance:                          0.28176 %
        // Average variance:                          6.47136 %
        // Average short username time:               1.8317 microseconds
        // Average long username time:                1.8304 microseconds
        // Average actual username time:              1.80362 microseconds
        //
        // Minimum variance:                          0.53593 %
        // Average variance:                          6.99394 %
        // Average short username time:               1.82322 microseconds
        // Average long username time:                1.81431 microseconds
        // Average actual username time:              1.78002 microseconds
```
- `fn it_compares_credentials_in_constant_time()`
```
        // This unit test asserts that the minimum variance is less than 10% (chosen to be robust for running the unit
        // test with CI), indicating that the function behaves within acceptable constant-time constraints.
        //
        // Some consecutive results running in release mode on a Core i7-12700H (with no other processes running):
        //
        // Minimum variance:                          0.43478 %
        // Average variance:                          2.08995 %
        // Average short username time:               34.580 microseconds
        // Average long username time:                34.315 microseconds
        // Average actual username time:              34.260 microseconds
        //
        // Minimum variance:                          0.43731 %
        // Average variance:                          1.77209 %
        // Average short username time:               34.560 microseconds
        // Average long username time:                34.755 microseconds
        // Average actual username time:              34.690 microseconds
        //
        // Minimum variance:                          0.43988 %
        // Average variance:                          1.61299 %
        // Average short username time:               34.33999 microseconds
        // Average long username time:                34.38500 microseconds
        // Average actual username time:              34.28500 microseconds
        //
        // Some consecutive results running in release mode on a Core i7-12700H (while entire CPU fully stressed):
        //
        // Minimum variance:                          0.30326 %
        // Average variance:                          2.29341 %
        // Average short username time:               64.87500 microseconds
        // Average long username time:                65.55499 microseconds
        // Average actual username time:              65.81000 microseconds
        //
        // Minimum variance:                          1.18168 %
        // Average variance:                          2.99206 %
        // Average short username time:               67.970 microseconds
        // Average long username time:                68.000 microseconds
        // Average actual username time:              68.005 microseconds
        //
        // Minimum variance:                          1.01083 %
        // Average variance:                          2.31316 %
        // Average short username time:               68.715 microseconds
        // Average long username time:                69.675 microseconds
        // Average actual username time:              69.715 microseconds
```

---
Code walk through
Run the unit tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_app_grpc/Cargo.toml
+++ b/applications/minotari_app_grpc/Cargo.toml
@@ -26,6 +26,7 @@ rand = "0.8"
 thiserror = "1"
 tonic = "0.6.2"
 zeroize = "1"
+subtle = { version = "2.5.0", features = ["core_hint_black_box"] }
 
 [build-dependencies]
 tonic-build = "0.6.2"
--- a/applications/minotari_app_grpc/src/authentication/basic_auth.rs
+++ b/applications/minotari_app_grpc/src/authentication/basic_auth.rs
@@ -20,12 +20,15 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{borrow::Cow, ops::Deref, string::FromUtf8Error};
+use std::{borrow::Cow, cmp::max, ops::Deref, string::FromUtf8Error};
 
 use argon2::{password_hash::Encoding, Argon2, PasswordHash, PasswordVerifier};
+use rand::RngCore;
 use tari_utilities::SafePassword;
 use tonic::metadata::{errors::InvalidMetadataValue, Ascii, MetadataValue};
 use zeroize::{Zeroize, Zeroizing};
+
+const MAX_USERNAME_LEN: usize = 256;
 
 /// Implements [RFC 2617](https://www.ietf.org/rfc/rfc2617.txt#:~:text=The%20%22basic%22%20authentication%20scheme%20is,other%20realms%20on%20that%20server.)
 /// Represents the username and password contained within a Authenticate header.
@@ -36,8 +39,11 @@ pub struct BasicAuthCredentials {
 }
 
 impl BasicAuthCredentials {
-    pub fn new(user_name: String, password: SafePassword) -> Self {
-        Self { user_name, password }
+    pub fn new(user_name: String, password: SafePassword) -> Result<Self, BasicAuthError> {
+        if user_name.as_bytes().is_empty() || user_name.as_bytes().len() > MAX_USERNAME_LEN {
+            return Err(BasicAuthError::InvalidUsername);
+        }
+        Ok(Self { user_name, password })
     }
 
     /// Creates a `Credentials` instance from a base64 `String`
@@ -47,7 +53,7 @@ impl BasicAuthCredentials {
         let as_utf8 = Zeroizing::new(String::from_utf8(decoded)?);
 
         if let Some((user_name, password)) = as_utf8.split_once(':') {
-            let credentials = Self::new(user_name.into(), password.to_string().into());
+            let credentials = Self::new(user_name.into(), password.to_string().into())?;
             return Ok(credentials);
         }
 
@@ -77,8 +83,53 @@ impl BasicAuthCredentials {
         Ok(credentials)
     }
 
+    // This function provides an approximate constant time comparison of the given username with the registered
+    // username, and is not meant to be an optimally performant comparison. This function also returns a boolean
+    // instead of a Result as the 'Ok' and 'Err' variants of the Result would have different execution times,
+    // It is also acceptable for the function to be more or less performant on different platforms, as long as the
+    // variance is within acceptable limits.
+    fn constant_time_compare_username(&self, username: &str) -> bool {
+        const BUFFER_MULTIPLIER: usize = 16;
+        const OPERATE_LEN: usize = MAX_USERNAME_LEN * BUFFER_MULTIPLIER;
+
+        let a_bytes = self.user_name.as_bytes();
+        let b_bytes = username.as_bytes();
+
+        // We do not care if this check returns immediately, as the the maximum allowed value is sufficiently long
+        // and it will not leak any information about the correctness of the supplied username other than it is too
+        // long.
+        if b_bytes.is_empty() || b_bytes.len() > MAX_USERNAME_LEN {
+            return false;
+        }
+
+        // Comparison bytes for both usernames are initialized to a large array of equal random bytes.
+        let mut a_compare_bytes = [0u8; OPERATE_LEN];
+        let mut rng = rand::thread_rng();
+        rng.fill_bytes(&mut a_compare_bytes);
+        let mut b_compare_bytes = a_compare_bytes;
+
+        // Both username's bytes are interleaved with the respective comparison byte arrays.
+        let interleave_factor = max(BUFFER_MULTIPLIER / 3, 1);
+        let a_repeat = OPERATE_LEN / (a_bytes.len() * interleave_factor);
+        for (i, (j, byte)) in (0..a_repeat).zip(a_bytes.iter()).enumerate() {
+            a_compare_bytes[i * a_bytes.len() * interleave_factor + j * interleave_factor] = *byte;
+        }
+        let b_repeat = OPERATE_LEN / (b_bytes.len() * interleave_factor);
+        for (i, (j, byte)) in (0..b_repeat).zip(b_bytes.iter()).enumerate() {
+            b_compare_bytes[i * b_bytes.len() * interleave_factor + j * interleave_factor] = *byte;
+        }
+
+        // Perform a constant time bitwise comparison of the two byte arrays and accumulate the result.
+        let mut result = 0;
+        for (x, y) in a_compare_bytes.iter().zip(b_compare_bytes.iter()) {
+            result |= x ^ y;
+        }
+
+        result == 0
+    }
+
     pub fn validate(&self, username: &str, password: &[u8]) -> Result<(), BasicAuthError> {
-        if self.user_name.as_bytes() != username.as_bytes() {
+        if !self.constant_time_compare_username(username) {
             return Err(BasicAuthError::InvalidUsername);
         }
         // These bytes can leak if the password is not utf-8, but since argon encoding is utf-8 the given
@@ -126,6 +177,8 @@ pub enum BasicAuthError {
 
 #[cfg(test)]
 mod tests {
+    use std::time::Instant;
+
     use super::*;
 
     mod from_header {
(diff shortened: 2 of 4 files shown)


applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,8 +83,53 @@ impl BasicAuthCredentials {
...
+        let b_repeat = OPERATE_LEN / (b_bytes.len() * interleave_factor);
+        for (i, (j, byte)) in (0..b_repeat).zip(b_bytes.iter()).enumerate() {
+            b_compare_bytes[i * b_bytes.len() * interleave_factor + j * interleave_factor] = *byte;
+        }
+
+        // Perform a constant time bitwise comparison of the two byte arrays and accumulate the result.
+        let mut result = 0;
+        for (x, y) in a_compare_bytes.iter().zip(b_compare_bytes.iter()) {
+            result |= x ^ y;
+        }
```
AaronFeickert: I'd recommend using `subtle` for this, since it supports [constant-time equality checking](https://docs.rs/subtle/latest/subtle/trait.ConstantTimeEq.html#impl-ConstantTimeEq-for-%5BT%5D) on slices of equal length, and can do some tricks to avoid compiler optimization.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,8 +83,53 @@ impl BasicAuthCredentials {
         Ok(credentials)
     }
 
+    // This function provides an approximate constant time comparison of the given username with the registered
+    // username, and is not meant to be an optimally performant comparison. This function also returns a boolean
+    // instead of a Result as the 'Ok' and 'Err' variants of the Result would have different execution times,
+    // It is also acceptable for the function to be more or less performant on different platforms, as long as the
+    // variance is within acceptable limits.
+    fn constant_time_compare_username(&self, username: &str) -> bool {
```
AaronFeickert: Is the intent of the random-byte interleaving merely to pad out the overall operation runtime to obscure the length-dependent copying of username data into the byte arrays?

AaronFeickert (changes requested):

AaronFeickert:
It's possible to simplify this design significantly while making it likely more robust against unwanted compiler optimizations. As noted in the earlier review, using `subtle` can simplify constant-time equality testing on byte slices. Second, it's not necessary to use the randomized byte slice interleaving. Simply prepending fixed-length byte arrays with each username's length (to avoid padding collisions) and then padding with zeros results in low timing variance.

Also, even supposing that a leak occurs, security really shouldn't depend on the username being secret.

Here's a [commit](https://github.com/AaronFeickert/tari/commit/b9813cf135620e9b9ebd062104ba45d16670ea82) on a test branch that uses this design.

hansieodendaal:
@AaronFeickert, thanks for the comments, the code is now greatly simplified with additional unit tests.

applications/minotari_app_grpc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ rand = "0.8"
 thiserror = "1"
 tonic = "0.6.2"
 zeroize = "1"
+subtle = { version = "2.5.0", features = ["core_hint_black_box"] }
```
AaronFeickert, line 29: Good call using this feature, since it adds an additional layer of optimization protection!

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -36,8 +41,17 @@ pub struct BasicAuthCredentials {
 }
 
 impl BasicAuthCredentials {
-    pub fn new(user_name: String, password: SafePassword) -> Self {
-        Self { user_name, password }
+    pub fn new(user_name: String, password: SafePassword) -> Result<Self, BasicAuthError> {
+        // Validate the username is well formed
+        if user_name.as_bytes().is_empty() || user_name.as_bytes().len() > MAX_USERNAME_LEN {
```
AaronFeickert: Maybe a silly question, but is there a particular reason to forbid an empty username? It doesn't appear to be a requirement of RFC 2617, and security really shouldn't depend on username complexity.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -36,8 +41,17 @@ pub struct BasicAuthCredentials {
...
+            return Err(BasicAuthError::InvalidUsername);
+        }
+        // Validate the password is well formed
+        let bytes = password.reveal().to_vec();
+        let str_password = Zeroizing::new(String::from_utf8(bytes)?);
+        let _parse_result = PasswordHash::parse(&str_password, Encoding::B64)?;
```
AaronFeickert: The use of the `Zeroizing` wrapper is inconsistent here, as all three of these variables represent the same data.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -36,8 +41,17 @@ pub struct BasicAuthCredentials {
 }
 
 impl BasicAuthCredentials {
-    pub fn new(user_name: String, password: SafePassword) -> Self {
-        Self { user_name, password }
+    pub fn new(user_name: String, password: SafePassword) -> Result<Self, BasicAuthError> {
```
AaronFeickert: It might be helpful to rename `password` to something that reflects its status as a PHC string, not the "raw" password. This was a bit confusing when I first read this code.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -47,7 +61,8 @@ impl BasicAuthCredentials {
         let as_utf8 = Zeroizing::new(String::from_utf8(decoded)?);
 
         if let Some((user_name, password)) = as_utf8.split_once(':') {
```
AaronFeickert, line 91: This is also a somewhat inconsistent use of `Zeroizing`.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
         Ok(credentials)
     }
 
+    // This function provides a constant time comparison of the given username with the registered username.
+    fn constant_time_compare_username(&self, username: &str) -> bool {
+        let a_bytes = self.user_name.as_bytes();
+        let b_bytes = username.as_bytes();
+
+        let valid_username = !(b_bytes.is_empty() || b_bytes.len() > MAX_USERNAME_LEN);
```
AaronFeickert: Same comment as above; is there a good reason to forbid an empty username?

AaronFeickert: It's safe to return early if the supplied username is malformed, since doing so doesn't leak anything about the validity of the supplied username.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
         Ok(credentials)
     }
 
+    // This function provides a constant time comparison of the given username with the registered username.
+    fn constant_time_compare_username(&self, username: &str) -> bool {
```
AaronFeickert: I wonder if it's more idiomatic to have this function return a `Choice`, and have the caller decide what to do with it: unwrap the underlying `u8` or use the `Choice` in other constant-time operations.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
...
+        let a_bytes = self.user_name.as_bytes();
+        let b_bytes = username.as_bytes();
+
+        let valid_username = !(b_bytes.is_empty() || b_bytes.len() > MAX_USERNAME_LEN);
+        let b_bytes_len_clipped = min(b_bytes.len(), MAX_USERNAME_LEN);
+
+        // Comparison bytes for both usernames are initialized to a large array of equal bytes.
+        let mut a_compare_bytes = [7u8; MAX_USERNAME_LEN];
```
AaronFeickert: Because this uses padding, I think you need to be extra careful about cases where the padding collides with either the correct or supplied username. I can think of two ways to do this:
- add a length check
- prepend a fixed-size encoding of each username to its array

applications/minotari_app_grpc/src/authentication/basic_auth.rs
AaronFeickert: Doing so will also let you avoid having to use any clipping.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
...
+        let mut b_compare_bytes = a_compare_bytes;
+
+        // Add the comparison bytes for the actual username to the respective comparison byte arrays.
+        a_compare_bytes[0..a_bytes.len()].clone_from_slice(a_bytes);
+        b_compare_bytes[0..b_bytes_len_clipped].clone_from_slice(&b_bytes[..b_bytes_len_clipped]);
+
+        (a_compare_bytes.ct_eq(&b_compare_bytes).unwrap_u8() != 0) && valid_username
```
AaronFeickert: If you have the function return a `Choice`, there's no need to unwrap the `u8` here. Additionally, `Choice` supports `BitAnd`, which will let you do this kind of logic while retaining compiler optimization mitigations.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
...
         // These bytes can leak if the password is not utf-8, but since argon encoding is utf-8 the given
         // password must be incorrect if conversion to utf-8 fails.
         let bytes = self.password.reveal().to_vec();
         let str_password = Zeroizing::new(String::from_utf8(bytes)?);
         let header_password = PasswordHash::parse(&str_password, Encoding::B64)?;
-        Argon2::default().verify_password(password, &header_password)?;
-        Ok(())
+        let valid_password = Argon2::default().verify_password(password, &header_password).is_ok();
+
+        if valid_username && valid_password {
```
AaronFeickert: I worry that the use of boolean logic here could open the door to the compiler being too clever for its own good.

You could use the result of `verify_password` to set a `Choice`, have `constant_time_compare_username` also return a `Choice`, and then use `BitAnd` logic on those instead. This will keep the logic behind the `subtle` optimization barriers.

AaronFeickert (changes requested):

applications/minotari_app_grpc/src/authentication/basic_auth.rs
AaronFeickert: Not that it matters, but I think using zeros for the initial value is slightly more elegant.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,17 +92,40 @@ impl BasicAuthCredentials {
...
+        let mut b_compare_bytes = a_compare_bytes;
+
+        // Add the comparison bytes for the actual username to the respective comparison byte arrays.
+        a_compare_bytes[0..a_bytes.len()].clone_from_slice(a_bytes);
+        b_compare_bytes[0..b_bytes_len_clipped].clone_from_slice(&b_bytes[..b_bytes_len_clipped]);
+
+        (a_compare_bytes.ct_eq(&b_compare_bytes).unwrap_u8() != 0) && valid_username
+    }
+
     pub fn validate(&self, username: &str, password: &[u8]) -> Result<(), BasicAuthError> {
```
AaronFeickert: Recommend adding function documentation making the logic clear here, since it's very important that both the username and passphrase are checked before returning success or (generic) failure.

AaronFeickert:
It looks like the real kerfuffle is trying to mitigate the effects of padding the actual username. One alternate design that eliminates this entirely is to create a `PaddedString` type that stores the length of the original username and a padded byte array. You can then use `subtle` to cleanly implement `ConstantTimeEq` for this type. When a comparison username is supplied, you create a `PaddedString` from it and run `ct_eq`. Padding the supplied username on demand leaks nothing to an attacker, and all the constant-time logic is neatly handled by the `ConstantTimeEq` trait. Most importantly, it means that no length-dependent operations are performed on the actual username except at creation time.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
AaronFeickert: I also just noticed that `subtle` provides an [idiomatic way](https://docs.rs/subtle/latest/subtle/struct.Choice.html#impl-From%3CChoice%3E-for-bool) to get a `bool` from a `Choice` that would be useful for the final return value here.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
hansieodendaal: Ok, I will remove that restriction

applications/minotari_app_grpc/src/authentication/basic_auth.rs
hansieodendaal: Ok, I will remove that restriction

hansieodendaal: This is intended behaviour by design; I would like any result given by this function to perform in constant time.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
hansieodendaal: I have also changed the name of the function to be in line with the `Argon2` `verify_password` function; it is now called `constant_time_verify_username`.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
hansieodendaal: I added a length check and also opted to fill the buffer with random bytes, to ensure that the compiler does not do any funny optimizations (now or in future) and to ensure that every comparison for the same username will force a different bitwise comparison.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
hansieodendaal: I rather prefer this function to return a `bool`

AaronFeickert:
This closes #5904.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,19 +120,57 @@ impl BasicAuthCredentials {
...
-        if self.user_name.as_bytes() != username.as_bytes() {
-            return Err(BasicAuthError::InvalidUsername);
-        }
+    // This function provides a constant time comparison of the given username with the registered username.
+    fn constant_time_verify_username(&self, username: &str) -> Choice {
+        // The username is valid if the lengths are equal and the length is not greater than the maximum allowed length;
+        // any error here will only be factored in after the bitwise comparison has been done to force constant time.
+        let bytes = username.as_bytes();
+        let valid_username = Choice::from(u8::from(
+            !(self.user_name_bytes_length != bytes.len() || bytes.len() > MAX_USERNAME_LEN),
```
AaronFeickert: These checks should probably each use a separate `Choice` to avoid optimization of the boolean `OR`.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -20,24 +20,66 @@
...
+        // Validate the username is well formed
+        if user_name.as_bytes().len() > MAX_USERNAME_LEN {
+            return Err(BasicAuthError::InvalidUsername);
+        }
+        // Validate the password is a well formed byte representation of a PHC string
+        let bytes = phc_password_hash.reveal().to_vec();
+        let _parse_result = PasswordHash::parse(&String::from_utf8(bytes)?, Encoding::B64)?;
+        // Random bytes are used for constant time username comparison to ensure that the compiler does not do any
+        // funny optimizations and to ensure that comparison for the same username for every new credentials instance
+        // forces a different bitwise comparison.
```
AaronFeickert, line 67: I still disagree that the use of random bytes here is necessary, as it could only become relevant in the event that `subtle` fails, and I'm not a fan of requiring the RNG unless necessary. That being said, I get the defense-in-depth approach here and wouldn't block review on it.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
AaronFeickert: So the overall username check is effectively `length_matches_choice & supplied_length_valid_choice & bytes_match_choice`.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,19 +120,57 @@ impl BasicAuthCredentials {
...
+        compare_bytes[0..bytes_len_clipped].clone_from_slice(&bytes[..bytes_len_clipped]);
+
+        // The remaining bytes are padded afterwards (and not initialized at the start) to ensure that this function
+        // always does the same amount of work irrespective of the username length.
+        compare_bytes[bytes.len()..MAX_USERNAME_LEN]
+            .clone_from_slice(&self.random_bytes[bytes.len()..MAX_USERNAME_LEN]);
+
+        // Perform the bitwise comparison and combine the result with the valid username result.
+        // The use of `Choice` logic here is by design to hide the boolean logic from compiler optimizations.
+        self.user_name_bytes.ct_eq(&compare_bytes).bitand(valid_username)
```
AaronFeickert, line 146: Nit: You could use `&` directly here if it makes the intent more clear.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -77,19 +120,57 @@ impl BasicAuthCredentials {
...
         let str_password = Zeroizing::new(String::from_utf8(bytes)?);
         let header_password = PasswordHash::parse(&str_password, Encoding::B64)?;
-        Argon2::default().verify_password(password, &header_password)?;
-        Ok(())
+        let valid_password = Choice::from(u8::from(
+            Argon2::default().verify_password(password, &header_password).is_ok(),
+        ));
+
+        // The use of `Choice` logic here is by design to hide the boolean logic from compiler optimizations.
+        if valid_username.bitand(valid_password).into() {
```
AaronFeickert, line 166: Nit: You could use `&` directly here if it makes the intent more clear.

applications/minotari_app_grpc/src/authentication/basic_auth.rs
```diff
@@ -156,26 +247,322 @@ mod tests {
...
+            assert!(matches!(
+                err,
+                BasicAuthError::InvalidPassword(argon2::password_hash::Error::PhcStringInvalid)
+            ));
+        }
 
-            let credentials = BasicAuthCredentials::new("bruteforce".to_string(), "secret".to_string().into());
-            let err = credentials.validate("admin", b"secret").unwrap_err();
+        #[test]
+        fn it_rejects_validating_mismatching_credentials() {
```
AaronFeickert, line 283: For completeness, could you add a test with both incorrect username and passphrase? This will obviously (correctly) fail, but it seems prudent given the structure of the design (and could catch a future regression).

AaronFeickert (changes requested):

AaronFeickert (approved):
LGTM

PR #5905, opened 2023-11-04
fix: don't fail early on bad gRPC username

### LINKED ISSUE #5904: Early authentication timing failure could leak username correctness
When gRPC credentials are [validated](https://github.com/AaronFeickert/tari/blob/1864203c224611cdcac71adbae83e37161ce0a5c/applications/minotari_app_grpc/src/authentication/basic_auth.rs#L80-L91), an error is returned if either the username or passphrase are incorrect. However, a failed username check will fail the validation immediately without checking the passphrase. This could introduce a timing attack.

It should always be the case that both a supplied username and passphrase are validated before returning a generic error to the caller.

AaronFeickert:
Description
---
Modifies gPRC credential validation not to fail early on an incorrect username.

Closes #5904.

Motivation and Context
---
Concurrent work in #5902 modifies gPRC username comparison to be a (mostly) constant-time operation. However, overall credential validation will still fail early if the username is incorrect without running `Argon2` on the provided passphrase. This could leak timing information to an attacker, who could use it to determine if the provided username is correct.

This PR modifies the credential validation flow to run both username and password validation. If either fails, a more generic error is returned. It is still possible to fail early on data that can't be parsed correctly; however, this doesn't leak any information and is therefore safe.

It also updates credential failure tests, which were written in a way that didn't fully exercise the proper failure mode.

Note that this will introduce a small conflict with #5902 that will require a trivial modification to address.

How Has This Been Tested?
---
Existing tests pass. Modified tests pass.

---
Check that the new validation flow has the intended effect, and that the updated test exercises the correct failure mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3035,6 +3035,7 @@ dependencies = [
  "prost 0.9.0",
  "prost-types 0.9.0",
  "rand",
+ "subtle",
  "tari_common_types",
  "tari_comms",
  "tari_core",
--- a/applications/minotari_app_grpc/Cargo.toml
+++ b/applications/minotari_app_grpc/Cargo.toml
@@ -26,6 +26,7 @@ rand = "0.8"
 thiserror = "1"
 tonic = "0.6.2"
 zeroize = "1"
+subtle = "2.5.0"
 
 [build-dependencies]
 tonic-build = "0.6.2"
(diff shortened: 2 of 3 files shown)


brianp (approved):
LGTM, utAck

AaronFeickert:
This is now superseded by an update to #5902.

PR #5916, opened 2023-11-07
fix!: lmdb flag set wrong on database

SWvheerden:
Description
---
The database does not use an integer key, but rather a hash. 
Small other improvements. 

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
ACK

PR #5928, opened 2023-11-08
feat(node grpc): add grpc authentication to the node

brianp:
Description
---
Adds basic auth to the minotari node.
Adds support on the Miner, and MergeMiner Proxy for node basic auth credentials.

Motivation and Context
---
Make things just that much more secure.

How Has This Been Tested?
---
Locally running a node, wallet, and miner. (MergeMiner untested).

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3162,6 +3162,7 @@ dependencies = [
  "hyper",
  "jsonrpc",
  "log",
+ "minotari_app_grpc",
  "minotari_app_utilities",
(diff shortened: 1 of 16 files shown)


hansieodendaal (commented):
Looks good, although I think we should consider using the same approach defined in #5936?

SWvheerden (approved):

Issue #5933, opened 2023-11-09
Base node config ban periods are currently set under block sync options. 


SWvheerden:
They should move the base node settings as they are not specific to block sync.

SWvheerden:
We should probably split the bans into 3 as well:
short, med and long depending on what they are.

SWvheerden:
dht actor has its own ban configs. These should all be consolidated:
```
#[derive(Debug, Clone, Copy)]
pub enum OffenceSeverity {
    Low,
    Medium,
    High,
}
```

SWvheerden:
This would a nice to have, but not required for now, going to close for now

PR #5934, opened 2023-11-09
feat: bans for bad incoming blocks

### LINKED ISSUE #5795: Incoming block does not ban for all cases
Incoming blocks dont ban for all cases, such as already known bad blocks, difficulty below min etc

SWvheerden:
Description
---
Bans peers who send bad incoming blocks. 
Consolidates banning a bit more. 

Motivation and Context
---
Created an issue to fix the ban period in the config: https://github.com/tari-project/tari/issues/5933

Fixes: #5795

### DIFF
--- a/base_layer/core/src/base_node/service/error.rs
+++ b/base_layer/core/src/base_node/service/error.rs
@@ -45,61 +45,46 @@ pub enum BaseNodeServiceError {
 }
 
 impl BaseNodeServiceError {
-    pub fn get_ban_reason(&self) -> Option<BanReason> {
+    pub fn get_ban_reason(&self, short_ban: Duration, long_ban: Duration) -> Option<BanReason> {
         match self {
             BaseNodeServiceError::CommsInterfaceError(comms) => match comms {
-                CommsInterfaceError::UnexpectedApiResponse => Some(BanReason {
-                    reason: "Unexpected API response".to_string(),
-                    ban_duration: Duration::from_secs(60),
+                err @ CommsInterfaceError::UnexpectedApiResponse | err @ CommsInterfaceError::RequestTimedOut => {
+                    Some(BanReason {
+                        reason: err.to_string(),
+                        ban_duration: short_ban,
+                    })
+                },
+                err @ CommsInterfaceError::InvalidPeerResponse(_) |
+                err @ CommsInterfaceError::InvalidBlockHeader(_) |
+                err @ CommsInterfaceError::TransactionError(_) |
+                err @ CommsInterfaceError::InvalidFullBlock { .. } |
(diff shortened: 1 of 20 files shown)


base_layer/core/src/base_node/service/error.rs
```diff
@@ -45,61 +45,46 @@ pub enum BaseNodeServiceError {
...
-                CommsInterfaceError::UnexpectedApiResponse => Some(BanReason {
-                    reason: "Unexpected API response".to_string(),
-                    ban_duration: Duration::from_secs(60),
+                err @ CommsInterfaceError::UnexpectedApiResponse | err @ CommsInterfaceError::RequestTimedOut => {
+                    Some(BanReason {
+                        reason: err.to_string(),
+                        ban_duration: short_ban,
+                    })
+                },
+                err @ CommsInterfaceError::InvalidPeerResponse(_) |
```
brianp, line 55: I haven't seen this syntax before. What influenced the change?

brianp (commented):
I'll be honest I don't love this PR.

We're letting the ban periods be configurable per node which is okay, but then we're passing those periods around anytime we need to fetch _reasons_ for a ban. These reasons don't actually need to know the configuration details.

On the `BanReason` struct, we could use an enum to define the expected ban duration. `BanDuration::Short`, `BanDuration::Long` and `BanDuration::Permenant`.

This makes it easy to manage in different services without creating a heavy requirement of config details. Once the BanReason reaches the ConnectivityManager, it can read the enum, and apply the correct configuration. The ConnectivityManager is the only one that needs to be able to translate that a `BanDuration::Short` means 60 seconds.

base_layer/core/src/base_node/service/error.rs
SWvheerden: This so we can use `|` and only a single impl bracket with using the `err` variable in this case.

brianp (commented):
It's a lot cleaner

I would personally try to go one step closer, and get the timing config into the `ConnectivityManager` so `ban_peer_until` can take the `BanReason` and we still don't need to destructure the duration until inside `ban_peer_until`

`ConnectivityManager` is brought up with the `CommsNode` spawn.  So we have lots of config options at hand, and which we could pass directly to the `ConnectivityManager`. 

At the point the destructure is managed by the `ConnectivityManager` it can be a simple helper function. Pass in the `BanDuration` enum, and receive a `Duration` as the return.

SWvheerden:
I am going to leave the config to move to this issue:https://github.com/tari-project/tari/issues/5933
It requires a bit more refactoring that I don't want to add as part of this PR.

brianp (approved):
I accept pushing the last refactor to a new issue. 

LGTM, utAck.

PR #5936, opened 2023-11-09
fix!: standardize gRPC authentication and mitigate DoS

### LINKED ISSUE #5809: Attackers can crash the wallet via the gRPC service
The BasicAuthCredentials::validate method compares the username and
password for the service (specified in the service's configuration file) with the
BasicAuthCredentials parsed from the network. To do so it uses
Argon2::default().verify_password which takes a plain text password string and
a hash of the expected password. Unfortunately, the wallet implementation
inverts these parameters and instead of passing the client supplied password to
the function, it uses the expected password and performs PasswordHash::parse on
the attacker controlled input.

The way argon2 hashes are designed, allows users to store the hashing
parameters (salt, memory usage, CPU usage, etc) in the hash itself for future
updates and flexibility. This means that PasswordHash::parse does more than
parsing a hashed password; it sets the parameters (including algorithm, memory
used, and time spent) that will be used to verify the password, therefore the
parsed data should come from a trusted source, for instance a database or a local
configuration file, as is the case with the wallet.

Because the parameters are controlled by the attacker, it is possible for them to
crash the program or use an arbitrary amount of resources on the victim system.
This issue could also lead to complete authentication bypass due to the current
design: an attacker can supply a hash with parameters that are less secure than
the default ones defined by the argon2 crate. For example the argon2 specification
defines the minimum hash byte length to be 4 bytes, so an attacker could send a
hash with these parameters and they would have to only brute-force 2^32
passwords to find a collision with the actual password.
This situation is rendered several orders of magnitude harder due to the better
defaults that the argon2 crate defines for the minimum hash size, which is 10
bytes, making the attack unfeasible.
The likelihood of this issue is deemed low as it is expected that most
configurations will keep wallets and node running on the same host.

If the existing authentication method is retained, the client must provide a
password for the server to hash and compare with the stored hash of the expected
password.

### LINKED ISSUE #5927: Wallet gRPC authentication is nonstandard
The handling of wallet and node gRPC authentication is somewhat nonstandard:
- The server configuration contains plaintext credentials.
- The user is provided with a PHC string representing the `Argon2` hash of the password.
- The client (miner/proxy) configuration is updated with the PHC string.
- When authenticating, the client provides the username and PHC string in an HTTP basic authorization header.
- The server applies `Argon2` to its plaintext password as part of verification.

My understanding had been that this was an intended design choice, given the local threat model. However, it appears that this isn't the case. With ongoing work to [support TLS connections](https://github.com/tari-project/tari/pull/5897) for gRPC operations, and recent work adding [base node gRPC authorization](https://github.com/tari-project/tari/pull/5928), the design should be updated to a more standard approach.

The usual PBKDF threat model assumes compromise of the server's password store, and wants to limit the ability of the attacker to determine a password. This isn't really the case here. One design approach could simply be a static shared high-entropy token that the server and client both store, and which is simply checked in constant time during authorization. This is unwieldy and less in line with the idea of HTTP authorization, so a standard username/password approach could still be used to keep things simple for the user.

The design could be the following:
- Plaintext credentials are supplied during gRPC server setup.
- The password is processed with `Argon2`, and its PHC string is stored locally along with a username (which could be blank or a default if desired).
- The client specifies the password in an HTTP header, read either via configuration or environment variable or runtime options.
- The server verifies the supplied password against its stored PHC string as part of verification.

There are still other considerations and risks with this:
- If the HTTP connection isn't otherwise protected, a passive attacker can simply sniff the password on the wire.
- If the client is compromised, an attacker can read the plaintext password from configuration or environment variable or runtime logs.

AaronFeickert:
Description
---
Standardizes gRPC authentication by removing PHC strings from configuration and preventing a client DoS.

Closes #5809. Closes #5927.

Motivation and Context
---
As noted in #5927, gRPC authentication is nonstandard. In the current design, server credentials are processed against a client-supplied username and PHC string for verification. This can lead to server DoS, as described in #5809.

This PR fixes the DoS vector. When the gRPC server is started, it applies `Argon2` to produce a PHC string that is kept in memory. When a client supplies its (non-PHC) passphrase, it is processed against the stored PHC string. This ensures that the server completely controls the `Argon2` parameters that are used.

Note that this is still suboptimal, as the client and server passphrases are still stored in plaintext configuration. Deciding how to handle those is out of scope for this work.

How Has This Been Tested?
---
It should be tested manually.

---
It should be tested manually.

BREAKING CHANGE: Updates the public APIs for `BasicAuthCredentials` and `ServerAuthenticationInterceptor` to accommodate the new behavior. Additionally, existing configuration client/server credentials will stop working, and client credentials will need to use plaintext passwords.


AaronFeickert:
Merging should wait for #5902, which includes an API change that will conflict.

applications/minotari_app_grpc/src/authentication/server_interceptor.rs
```diff
@@ -34,23 +35,41 @@ pub struct ServerAuthenticationInterceptor {
...
+                // Parse the provided header into (temporary) credentials for easier access
+                let header = t.to_str().map_err(unauthenticated)?;
+                let header_credentials = BasicAuthCredentials::from_header(header).map_err(unauthenticated)?;
+
+                // Validate the header credentials against the valid credentials
+                let valid_credentials =
+                    BasicAuthCredentials::new(valid_username.to_owned(), valid_phc_password.clone());
+                valid_credentials
+                    .validate(&header_credentials.user_name, header_credentials.password.reveal())
                     .map_err(unauthenticated)?;
```
hansieodendaal: I like this approach, as the question is just asked from a different angle

applications/minotari_app_grpc/src/authentication/server_interceptor.rs
```diff
@@ -34,23 +35,41 @@ pub struct ServerAuthenticationInterceptor {
...
     ) -> Result<Request<()>, Status> {
         match req.metadata().get(AUTHORIZATION.as_str()) {
             Some(t) => {
-                let val = t.to_str().map_err(unauthenticated)?;
-                let credentials = BasicAuthCredentials::from_header(val).map_err(unauthenticated)?;
-                credentials
-                    .validate(valid_username, valid_password)
+                // Parse the provided header into (temporary) credentials for easier access
+                let header = t.to_str().map_err(unauthenticated)?;
+                let header_credentials = BasicAuthCredentials::from_header(header).map_err(unauthenticated)?;
```
hansieodendaal: Please note that #5902 fixed an issue with `BasicAuthCredentials::from_header` where the `generate_header` .. `from_header` roundtrip did not produce credentials that could be validated, as demonstrated in `fn it_generates_a_valid_header()` and also in `fn it_decodes_from_well_formed_header()`.

hansieodendaal (commented):
I like this approach, looks good. What about adding some unit tests to `ServerAuthenticationInterceptor`?

AaronFeickert:
> What about adding some unit tests

I looked into this, but it wasn't clear how to cleanly do such tests. Right now, it seems that all of the integration tests use `GrpcAuthentication::None` and are all part of much larger wallet test suites. Any insight on a simpler approach that would let us test all `GrpcAuthentication` modes without the extra baggage of the wallet?

AaronFeickert:
Marking as draft for some refactoring.

applications/minotari_app_grpc/src/authentication/server_interceptor.rs
AaronFeickert: Header parsing now returns the credentials as a tuple. This could be simplified even more by passing the authorization header directly to the validator, but that's for another day.

applications/minotari_app_grpc/src/authentication/server_interceptor.rs
AaronFeickert: Parsing the header credentials to `BasicAuthCredentials` is a kludge that's only there because of the parser; an upcoming refactor should make this unnecessary!

SWvheerden (approved):

brianp (approved):
Although I understand the reason for the move it feels a little weird to keep the password in cleartext, even though keeping the hashed password previously didn't change the level of security if someone could access it. 
And we are now moving to remove grpc from the wallet entirely so it looks like we're closing down vectors from multiple angles. So I'm good with this.
utAck.

AaronFeickert:
@brianp: I agree that it's not an ideal solution, but I think it meets a reasonable threat model and, improves the user experience, and removes the DoS vector. Even with removing wallet gRPC by default, that attack vector would still exist with node gRPC. Supporting TLS will be an additional layer of protection too. So I think overall it's a net win that doesn't sacrifice practical security.

PR #5943, opened 2023-11-10
feat: remove panics from applications

### LINKED ISSUE #5940: Attackers can crash whole node by panicking a single thread
A panic on a thread will cause the whole process to be aborted. This happens both
in the wallet and in the base node.
```rust
fn main() {
// Setup a panic hook which prints the default rust panic message
but also exits the process. This makes a panic in
// any thread "crash" the system instead of silently continuing.
let default_hook = panic::take_hook();
panic::set_hook(Box::new(move |info| {
default_hook(info);
process::exit(1);
}));
```
This design decision allowed several vulnerabilities in this report to be of critical
severity. For example, panics on RPC requests would have practically no impact
when their thread panics instead of being a tool for triggering a denial of service on
the whole network.

Panics on threads should be treated as bugs and fixed, but an attacker would have
a much harder time converting them to successful exploits if the program did not
crash immediately.

SWvheerden:
Description
---
Applications currently panic on any thread panic. Remove this so that applications will not panic. 

Motivation and Context
---
See: #5940 

This is somewhat debatable if this is desired or not behaviour. But currently, the favoured approach is to only panic the thread. 

Fixes: #5940

### DIFF
--- a/applications/minotari_console_wallet/src/main.rs
+++ b/applications/minotari_console_wallet/src/main.rs
@@ -20,7 +20,7 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{panic, process};
+use std::process;
 
 use clap::Parser;
 use log::*;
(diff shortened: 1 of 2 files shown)


brianp (approved):
LGTM.

As per the removed comment. Does this mean if some service does Panic the app may risk silently continuing while that particular service of the app has died?

SWvheerden:
Yip, not my preferred choice, I prefer global panic. But it seems if its not the preferred approach.

brianp:
Maybe we'll have to investigate some kind of timed-out respawn of lost services

PR #5945, opened 2023-11-10
test: relax constant time assert

hansieodendaal:
Description
---
Relaxed constant time variance asserts due to CI resource starving in release mode; this will still run properly in debug mode.

Motivation and Context
---
CI has varying performance

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_app_grpc/src/authentication/basic_auth.rs
+++ b/applications/minotari_app_grpc/src/authentication/basic_auth.rs
@@ -368,8 +368,8 @@ mod tests {
             // Notes:
             // - The `assert!(!do_performance_testing);` at the end of the test will cause a panic on CI if the flag is
             //   enabled, if it is enabled it will allow results to be printed when running in release mode.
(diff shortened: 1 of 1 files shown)


SWvheerden:
I dont feel like this gives us much. It might not fail much on CI, but if it runs on CI, it wont mean much. 

From the test point, we want to ensure that the username verification is constant time. But if it passes the test, we can only assert that it is within 25% which is not much at all.

AaronFeickert:
I'm still of the opinion that username obscurity shouldn't be part of the overall security model, to the extent reasonable; usernames should be for identification, not authentication. For this reason, while I like the idea of having a best-effort attempt at constant-time credential verification (and a corresponding test to examine this), detailed testing should be reserved for manual runs during algorithm testing, and not for CI.

hansieodendaal:
Functional only asserts remain for CI; variance measurement is only available for local performance and development testing.

SWvheerden (approved):

PR #5947, opened 2023-11-13
feat: create min dust fee setting

### LINKED ISSUE #4501: Add setting to ignore dust
The wallet needs a setting to ignore dust UTXOs

### LINKED ISSUE #5807: Default output selection criteria can be abused to increase the sender's fee
The output selection mechanism can be abused to increase the sender's
transaction fee in scenarios where the adversary can control the amount to be
transferred.
The default output selection mechanism can use two modes depending on the
amount to be transferred. If the amount is larger than the maximum output value,
the order will be descendant. Otherwise, it will be ascendant, meaning it will
select smaller outputs first.

Therefore, an adversary able to control the amount to be transferred can force the
sender to use dust values to increase the transaction fee. The dust input values to
be used by the sender could be previously sent by the adversary.
Even though the wallet admits configuring a flag to drop the transaction in case
the fees exceed the amount to be transferred, it can still be abused to spend an
amount of fees slightly lower than the value to be transferred.
Recommendation
Ignore dust amounts in the output selection mechanism.
Also, consider implementing algorithms such as Branch and Bound which
minimize the creation of change outputs which can be of dust amounts, by trying to
find the exact match for transaction amounts in the UTXO set.

SWvheerden:
Description
---
Create a setting to ignore dust below a certain value/ 

Motivation and Context
---

See https://github.com/tari-project/tari/issues/5807

Fixes: https://github.com/tari-project/tari/issues/5807
Fixes: #4501

### DIFF
--- a/base_layer/wallet/src/output_manager_service/input_selection.rs
+++ b/base_layer/wallet/src/output_manager_service/input_selection.rs
@@ -40,22 +40,25 @@ pub struct UtxoSelectionCriteria {
     pub filter: UtxoSelectionFilter,
     pub ordering: UtxoSelectionOrdering,
     pub excluding: Vec<Commitment>,
+    pub min_dust: u64,
     pub excluding_onesided: bool,
 }
 
 impl UtxoSelectionCriteria {
-    pub fn smallest_first() -> Self {
+    pub fn smallest_first(min_dust: u64) -> Self {
         Self {
             filter: UtxoSelectionFilter::Standard,
             ordering: UtxoSelectionOrdering::SmallestFirst,
+            min_dust,
             ..Default::default()
         }
     }
 
-    pub fn largest_first() -> Self {
+    pub fn largest_first(min_dust: u64) -> Self {
         Self {
             filter: UtxoSelectionFilter::Standard,
             ordering: UtxoSelectionOrdering::LargestFirst,
+            min_dust,
             ..Default::default()
         }
     }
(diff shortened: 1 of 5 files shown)


base_layer/wallet/src/output_manager_service/input_selection.rs
```diff
@@ -64,6 +67,7 @@ impl UtxoSelectionCriteria {
         Self {
             filter: UtxoSelectionFilter::SpecificOutputs { commitments },
             ordering: UtxoSelectionOrdering::Default,
+            min_dust: 0,
```
brianp, line 70: Should `specific` have other unknown qualifiers than those specified?

common/config/presets/d_console_wallet.toml
```diff
@@ -133,6 +133,8 @@ transaction_event_channel_size = 25000
 # transaction amount. Set this value to `false` to allow spending of "dust" UTXOs for small valued transactions
 # (default = true).
 prevent_fee_gt_amount = false
+# Ignores dust below this value, value in micro MinoTari
```
brianp:
```suggestion
# Ignores dust below this value, value in micro MinoTari, defaults to 100
```

Just to remove ambiguity around the fact this is happening whether this option is enabled or not.

brianp (commented):
Ack

Mostly looks good. Two comments.

base_layer/wallet/src/output_manager_service/input_selection.rs
SWvheerden: No just that, you are selecting very specific commitments, so there is no need for a dust filter

brianp: 🙃 

I clearly glossed over the actual value.

PR #5948, opened 2023-11-13
ci(fix): enable libtor for osx builds

leet4tari:
Description
Enable libtor for OSX builds in CI

Motivation and Context
Add libtor as internal dependancy, so that we don't require tor setup as a proxy.

How Has This Been Tested?
Testing locally build OSX base node.

Download OSX binary and run with libtor enabled.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):

stringhandler (approved):
LFG

PR #5949, opened 2023-11-13
feat: block wallets from sending if BN connection stale

### LINKED ISSUE #5855: Wallet has no warning for stale balance
The wallet returns stale balance results with no warning to the user if it cannot
get a live state from the base node. This will lead clients to use stale results.
There is no way to distinguish between stale and updated results.
This is because the GetBalance operation uses the get_chain_metadata function to
calculate the current tip. GetChainMetadata in turn will return an Ok result even
when it fetches possibly outdated data from its database.
```rust
            OutputManagerRequest::GetBalance => {
                let current_tip_for_time_lock_calculation = match 
self.base_node_service.get_chain_metadata().await {
                    Ok(metadata) => metadata.map(|m| 
m.height_of_longest_chain()),
                    Err(_) => None,
                };
                self.get_balance(current_tip_for_time_lock_calculation)
                    .map(OutputManagerResponse::Balance)
            },
            BaseNodeServiceRequest::GetChainMetadata => match 
self.get_state().await.chain_metadata {
                Some(metadata) => 
Ok(BaseNodeServiceResponse::ChainMetadata(Some(metadata))),
                None => {
                    // if we don't have live state, check if we've 
previously stored state in the wallet db
                    let metadata = self.db.get_chain_metadata()?;
                    
Ok(BaseNodeServiceResponse::ChainMetadata(metadata))
                },
            },
```

Provide a clear warning to users that they are getting outdated data

SWvheerden:
Description
---
block wallet from sending if metadata from base node is stale and older than 15 mins
The console wallet will print out a warning if the connection is stale. 

Motivation and Context
---
Sending via old metadata can cause issues with wallets using old data. The transactions might not be valid anymore 

Fixes: https://github.com/tari-project/tari/issues/5855

### DIFF
--- a/base_layer/wallet/src/base_node_service/service.rs
+++ b/base_layer/wallet/src/base_node_service/service.rs
@@ -22,7 +22,7 @@
 
 use std::{sync::Arc, time::Duration};
 
-use chrono::{NaiveDateTime, Utc};
+use chrono::NaiveDateTime;
 use futures::{future, StreamExt};
 use log::*;
 use tari_common_types::chain_metadata::ChainMetadata;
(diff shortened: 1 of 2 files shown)


base_layer/wallet/src/base_node_service/service.rs
```diff
@@ -152,13 +152,26 @@ where T: WalletBackend + 'static
...
+                // if the wallet has not gotten a ChainMetaData in the last 15 minutes, we dont return one as this can
+                // be an issue using old data to send transactions. Wallets should have up to date ChainMetaData as they
+                // need to have an active connection to a base node to send transactions.
+                let last_updated = self
+                    .get_state()
+                    .await
+                    .updated
+                    .ok_or(BaseNodeServiceError::NoChainMetadata)?;
+                if Utc::now().naive_utc().timestamp() - last_updated.timestamp() > 15 * 60 {
+                    return Err(BaseNodeServiceError::NoChainMetadata);
```
brianp: Should this error be more specific than `NoChainMetadata`. Assumedly we have _some metadata_ it's just that we consider it stale.

brianp (commented):
LGTM. Ack

One suggestion, but not vital.

base_layer/wallet/src/base_node_service/service.rs
SWvheerden: maybe.. I thought about it, but still not sure.

PR #5951, opened 2023-11-13
feat: block endless peer stream

### LINKED ISSUE #5811: Peer can hold the peer-stream open indefinitely
Malicious peers can hold a GetPeersRequest stream open indefinitely when an
innocent peers joins the network and requests peers from them. The malicious
peer can send the same peer over and over again.
The victim validates the shared peers before marking it as a duplicates and never
takes any action against the attacker as long as peers sent are valid.

As long as the attacker keeps sending peers, validate_and_add_peer will be
called.
An attacker could take advantage of this by setting up several malicious nodes on
the network which kept the discovery process busy in order to make it more
difficult for victims to find honest peers.

Set a maximum amount of peers the client will listen for as well as timeouts. Have
low tolerance for sending duplicated peers.

SWvheerden:
Description
---
Block endless peer stream from remote peer. 
Block duplicate peers.

Motivation and Context
---
See #5811 

Fixes #5811

### DIFF
--- a/comms/dht/src/network_discovery/discovering.rs
+++ b/comms/dht/src/network_discovery/discovering.rs
@@ -175,7 +175,7 @@ impl Discovering {
         while let Some(resp) = stream.next().await {
             counter += 1;
             if counter > self.params.num_peers_to_request {
-                error!(target: LOG_TARGET, "Remote peer sent more peers than we requested.");
+                warn!(target: LOG_TARGET, "Remote peer sent more peers than we requested.");
                 return Err(NetworkDiscoveryError::TooManyPeersReceived);
             }
             let GetPeersResponse { peer } = resp?;
@@ -185,7 +185,7 @@ impl Discovering {
                 .try_into()
                 .map_err(NetworkDiscoveryError::InvalidPeerDataReceived)?;
             if !peers_received.insert(new_peer.public_key.clone()) {
-                error!(target: LOG_TARGET, "Remote peer sent duplicate peer.");
+                warn!(target: LOG_TARGET, "Remote peer sent duplicate peer.");
                 return Err(NetworkDiscoveryError::DuplicatePeerReceived);
             }
             self.validate_and_add_peer(sync_peer, new_peer).await?;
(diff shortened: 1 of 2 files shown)


AaronFeickert (commented):
Is it straightforward to include unit tests exercising the two new failure modes?

SWvheerden:
I thought about it, but it would mean we need to a full malicious server client.

comms/dht/src/network_discovery/discovering.rs
```diff
@@ -170,14 +170,24 @@ impl Discovering {
...
+            }
             let GetPeersResponse { peer } = resp?;
 
             let peer = peer.ok_or_else(|| NetworkDiscoveryError::EmptyPeerMessageReceived)?;
-            let new_peer = peer
+            let new_peer: UnvalidatedPeerInfo = peer
                 .try_into()
                 .map_err(NetworkDiscoveryError::InvalidPeerDataReceived)?;
+            if !peers_received.insert(new_peer.public_key.clone()) {
+                error!(target: LOG_TARGET, "Remote peer sent duplicate peer.");
```
sdbondi: nit: I think WARN is more appropriate since a remote can cause this, `error!` implies a local error/malfunction. Same for the error! above

sdbondi (dismissed):
utACK

sdbondi:
There is a mock DHT RPC implementation that you could set to return duplicates or too many peers and then pass that into the discovery module. Will likely be a fair amount of work. 

**DHT Mock service**
https://github.com/tari-project/tari/blob/development/comms/dht/src/rpc/mock.rs
**Examples of making calls to a mocked service**
https://github.com/tari-project/tari/blob/development/comms/core/src/protocol/rpc/test/smoke.rs
https://github.com/tari-project/tari/blob/development/comms/core/src/protocol/rpc/client/tests.rs

comms/dht/src/network_discovery/discovering.rs
SWvheerden: agreed

PR #5952, opened 2023-11-13
chore: simplify metrics-only blocks

AaronFeickert:
Description
---
Does minor refactoring of metrics-only blocks to make the analyzer happy.

Motivation and Context
---
Recent work in #5944 makes metrics a non-default feature. This made the analyzer unhappy due to some unused variables, but it turns out this can be fixed while simplifying some of the handling of metrics-only blocks. This PR makes these changes and keeps the analyzer happy.

How Has This Been Tested?
---
Existing tests pass.

---
Check that the new metrics-only block handling is identical to that introduced by #5944.


SWvheerden (approved):
haha I added the exact same fixes here: https://github.com/tari-project/tari/pull/5947

AaronFeickert:
@SWvheerden: great minds think alike!

PR #5953, opened 2023-11-14
chore(ci): osx ARM64 build by removing envs

leet4tari:
Description
Remove build envs used for croaring with cross-compiling that was breaking OSX ARM64 builds.
Re-enabled OSX ARM64 builds with fix in place.
Fix some missing tari to minotari rename operations, also helps with OSX PKG assembly.

Motivation and Context
Bring back the OSX ARM64 builds and also fix all OSX PKG assembly.

How Has This Been Tested?
Tested build in local fork and pkg install locally.

Note that OSX ARM64 builds successful and pkg assembly without errors.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


.github/workflows/base_node_binaries.yml
```diff
@@ -192,13 +188,16 @@ jobs:
...
 
       - name: Build rust binaries
-        env:
-          CARGO_UNSTABLE_SPARSE_REGISTRY: true
-          RUSTFLAGS: "-C target_cpu=${{ matrix.builds.target_cpu }}"
-          ROARING_ARCH: "${{ matrix.builds.target_cpu }}"
+        # env:
+        #   CARGO_UNSTABLE_SPARSE_REGISTRY: true
+        #   RUSTFLAGS: "-C target_cpu=${{ matrix.builds.target_cpu }}"
+        #   ROARING_ARCH: "${{ matrix.builds.target_cpu }}"
```
brianp: We need to leave commented out code?

brianp (changes requested):
Mostly looks good. Lets try not to leave stuff that's commented out unless it's got an explicit reason to stay.

.github/workflows/base_node_binaries.yml
leet4tari: Until all the builds and binaries have been fully tested, I would prefer too.

SWvheerden: I prefer we remove them. 
We have the git history if we want to retrieve them again. 
Commented out code tends to live forever in the code.

SWvheerden (commented):
Looks good, I would just prefer we remove that commented code

SWvheerden (approved):

PR #5954, opened 2023-11-14
chore!: remove mutable mmr

SWvheerden:
Description
---
removes croaring and the mutable mmr struct

Motivation and Context
---
There are not used anymore


brianp (approved):
LGTM. utAck

CjS77:
It's a breaking change, so will need a major version change on the crate at next release.

Edit: though it looks like we haven't published it in a while: https://crates.io/crates/tari_mmr

base_layer/mmr/benches/smt_vs_mmr.rs
AaronFeickert, line 1: The remaining tests could be moved to the SMT bencher, since they're no longer used for comparison. Would also recommend moving them to `criterion` to get better and more consistent data.

AaronFeickert (changes requested):

base_layer/mmr/benches/smt_vs_mmr.rs
AaronFeickert: There's also a `clippy` failure in the same file.

SWvheerden:
> t's a breaking change, so will need a major version change on the crate at next release.

Yeah its synced to the tari version atm which is 0.53.0

base_layer/mmr/benches/smt.rs
```diff
@@ -38,5 +38,57 @@ pub fn benchmark_sparse_merkle_trees(c: &mut Criterion) {
...
+fn time_function(header: &str, f: impl FnOnce()) -> std::time::Duration {
+    println!("Starting: {header}");
+    let now = std::time::Instant::now();
+    f();
+    let t = now.elapsed();
+    println!("Finished: {header} - {t:?}");
+    t
+}
+
+pub fn root_hash(_c: &mut Criterion) {
```
AaronFeickert, line 62: Can these operations be refactored to use Criterion's `bench_function` design? It might get a little tricky because of how its iterations work, since you want to work on a fresh tree for each of its iterations to ensure the operations are being run as expected. But at least then you can probably get better timing data.

base_layer/mmr/benches/smt.rs
```diff
@@ -19,7 +19,7 @@ fn create_smt() -> SparseMerkleTree<Blake2b<U32>> {
     SparseMerkleTree::<Blake2b<U32>>::new()
 }
 
-pub fn benchmark_sparse_merkle_trees(c: &mut Criterion) {
+pub fn benchmark_smt_insert(c: &mut Criterion) {
```
AaronFeickert, line 22: Looking at this more closely, I think there's a subtle error in this benchmark. A fresh tree is created for each sample, but then each iteration of that sample applies the `upsert` operations to the same tree. This means the first iteration will perform inserts, and subsequent iterations will perform updates. The timing data will only be accurate if an insert and update take the same amount of time.

AaronFeickert:
Since rewriting the benchmarks to use a more standard Criterion design might take a bit, should this be merged as is? Benchmarks could be improved separately.

AaronFeickert:
Fixes and updates to the benchmarks are addressed in #5962.

PR #5959, opened 2023-11-15
chore: hazop code review changes

SWvheerden:
Description
---
Some changes from the hazop meeting

Motivation and Context
---
Clean secure code

### DIFF
--- a/base_layer/core/src/blocks/block.rs
+++ b/base_layer/core/src/blocks/block.rs
@@ -162,7 +162,6 @@ pub struct BlockBuilder {
     inputs: Vec<TransactionInput>,
     outputs: Vec<TransactionOutput>,
     kernels: Vec<TransactionKernel>,
-    total_fee: MicroMinotari,
 }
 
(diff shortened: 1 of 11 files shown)


base_layer/core/src/blocks/block.rs
```diff
@@ -196,10 +194,6 @@ impl BlockBuilder {
 
     /// This function adds the provided transaction kernels to the block WITHOUT updating kernel_mmr_size in the header
     pub fn add_kernels(mut self, mut kernels: Vec<TransactionKernel>) -> Self {
-        for kernel in &kernels {
```
stringhandler, line 199: Is this total fee not used?

stringhandler (commented):
Seems ok

base_layer/core/src/blocks/block.rs
SWvheerden: Nope

brianp (approved):
LGTM utAck

PR #5960, opened 2023-11-15
feat: limit max number of addresses

### LINKED ISSUE #5856: Attacker can flood victim with peer addresses
An  attacker  can  continuously  and  quickly  announce  new  addresses  to  a  victim,
flooding  a  victim  with  Join  messages  and  making  the  addresses  vector  of  the
victim grow indefinitely. The attacker can thus make the victim waste CPU time
and storage.
The storage use is negligible due to the limit on the amount of addresses sent on
each Join message imposed by the fixes on issue TARI-021. Nevertheless, the
computing time continues to be asymmetrical, as the attacker uses less computing
time to send messages than the victim to process them.
This  is  partially  due  to  the  fact  that  each  address  takes  longer  and  longer  to
process  due  to  the  increment  in  the  addresses  vector,  which  causes  a  linear
increase in the time to add each new address:
```rust
    pub fn merge(&mut self, other: &MultiaddressesWithStats) {
        for addr in &other.addresses {
            if let Some(existing) = 
self.find_address_mut(addr.address()) {
                existing.merge(addr);
            } else {
                self.addresses.push(addr.clone());
            }
        }
    }
    /// Finds the specified address in the set and allow updating of 
its variables such as its usage stats
    fn find_address_mut(&mut self, address: &Multiaddr) -> Option<&mut 
MultiaddrWithStats> {
        self.addresses.iter_mut().find(|a| a.address() == address)
    }
```
This process is protected from being multi-threaded by a lock, which further slows
down the processing of the flood of messages.

```rust
pub async fn add_peer(&self, peer: Peer) -> Result<PeerId, 
PeerManagerError> {
        let t = Instant::now();
        let mut lock = self.peer_storage.write().await;
        let peer_id = lock.add_peer(peer)?;
        #[cfg(feature = "metrics")]
        {
            let count = lock.count();
            metrics::peer_list_size().set(count as i64);
        }
        Ok(peer_id)
    }
```
Any node that connects to the victim can be automatically attacked by the attacker,
as the victim will share the attacker as a normal, benign peer.

Limit the amount of addresses a peer can have. Implement a rate limit mechanism
for all messages

SWvheerden:
Description
---
Limit the max number of addresses a peer can have

Motivation and Context
---
See #5856 

How Has This Been Tested?
---
Unit tests

Fixes: #5856

### DIFF
--- a/comms/core/src/net_address/mutliaddresses_with_stats.rs
+++ b/comms/core/src/net_address/mutliaddresses_with_stats.rs
@@ -13,6 +13,7 @@ use multiaddr::Multiaddr;
 use serde::{Deserialize, Serialize};
 
 use crate::net_address::{multiaddr_with_stats::PeerAddressSource, MultiaddrWithStats};
+const MAX_ADDRESSES: usize = 10;
 
 /// This struct is used to store a set of different net addresses such as IPv4, IPv6, Tor or I2P for a single peer.
 #[derive(Debug, Clone, Deserialize, Serialize, Default, Eq)]
@@ -29,9 +30,11 @@ impl MultiaddressesWithStats {
         for address in addresses {
             addresses_with_stats.push(MultiaddrWithStats::new(address, source.clone()));
         }
-        MultiaddressesWithStats {
+        let mut addresses = MultiaddressesWithStats {
             addresses: addresses_with_stats,
-        }
+        };
+        addresses.sort_addresses();
+        addresses
     }
 
     pub fn empty() -> Self {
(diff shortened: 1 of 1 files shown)


stringhandler (commented):
I don't agree with this approach. You should rather allow more than this in the Multiaddress struct and use consensus to ban a node that provides more than the ConsensusConstant.MaxAddresses

stringhandler (approved):
Not how I would do it, but I suppose it can work

Issue #5817, opened 2023-09-26
Signature opcode handling allows trivial forgery


AaronFeickert:
Signatures are handled within TariScript using several opcodes: `CheckSig`, `CheckMultiSig`, `CheckSigVerify`, `CheckMultiSigVerify`, and `CheckMultiSigVerifyAggregatePubkey`. In each case, one or more public keys and signatures are popped from the stack, and a signature challenge (called `message` in the code) is read from the opcodes' associated data. A "raw verifier" in `tari-crypto`, which assumes the challenge is correctly generated, is used to assess signature validity.

This allows anyone to trivially forge a signature. Because the challenge is supplied as part of the script, it is not possible for it to bind to the signer's initial transcript round. This allows an attacker to easily produce a signature that will pass verification, even if it does not possess the secret key corresponding to the expected public key.

It is also the case that the multisignature opcodes' handling uses a common challenge for all signatures provided on the stack, which means such a challenge cannot even individually bind to each signature's expected public key.

This can be easily fixed by treating the opcodes' associated challenge instead as a message to be signed and verified using proper Fiat-Shamir.

AaronFeickert:
It was noted elsewhere that it may be helpful to base the resulting design on Bitcoin's scripting functionality, since users may be more likely to be familiar with its design patterns, which could reduce risk and improve user and developer experience.

AaronFeickert:
On further thought, it isn't even possible for the current design to be done safely using Fiat-Shamir. Because the message (really the signature challenge) is part of the script via the relevant opcodes' associated data, it cannot bind the signer's initial transcript round. This makes forgery trivial.

Replacing the current design with one that uses a standard Fiat-Shamir technique on the message, public key, and initial transcript round fixes this. It also renders existing questions about scalar representation moot, as the message can be any arbitrary bytes that are semantically valid according to script rules.

PR #5961, opened 2023-11-16
fix(tariscript): multisig ordered signatures and pubkeys

### LINKED ISSUE #5806: Malicious script makes validators waste computations
An attacker can force the CHECK_MULTISIG family of opcodes to be always worst
case by ordering the public keys and their signatures in a way that maximizes the
amount of iterations the nested for-loop must do in order to verify the script.
By putting the public key of the signers and their signatures in reverse order (ie:
the signer for the first signature is the last public key, the second is the second to
last, and so on), an attacker makes all validators waste resources. This is not
reflected in the protocol fees. It is worth keeping in mind that verifying signatures
is an expensive process for validators.

brianp:
Description
---
Update the multisig op code checks to care about the order of keys and signatures. This is to assist with reducing the amount of iterations and processing a node is required to perform to validate a collection.

Motivation and Context
---
Low/Minor audit issue to reduce processing for nodes.

Closes: #5806 

How Has This Been Tested?
---
Test suite / CI

---
Look at the test additions. They make it more clear what the effects of the changes are.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/infrastructure/tari_script/src/script.rs
+++ b/infrastructure/tari_script/src/script.rs
@@ -591,6 +591,7 @@ impl TariScript {
     ///
     /// Notes:
     /// * The _m_ signatures are expected to be the top _m_ items on the stack.
+    /// * The public keys and signatures must both be ordered.
     /// * Every public key can be used AT MOST once.
     /// * Every signature MUST be a valid signature using one of the public keys
     /// * _m_ and _n_ must be positive AND m <= n AND n <= MAX_MULTISIG_LIMIT (32).
@@ -622,9 +623,13 @@ impl TariScript {
         let mut sig_set = HashSet::new();
 
         let mut agg_pub_key = RistrettoPublicKey::default();
-        // Check every signature against each public key looking for a valid signature
-        for s in &signatures {
-            for (i, pk) in public_keys.iter().enumerate() {
+
+        // Create an iterator that allows each pubkey to only be checked a single time
+        let mut pub_keys = public_keys.iter();
+
+        // Signatures and public keys must be ordered
+        for (i, s) in signatures.iter().enumerate() {
+            while let Some(pk) = pub_keys.next() {
                 if !sig_set.contains(s) && !key_signed[i] && s.verify_raw_canonical(pk, &message) {
                     // This prevents Alice creating 2 different sigs against her public key
                     key_signed[i] = true;
--- a/infrastructure/tari_script/src/script.rs
+++ b/infrastructure/tari_script/src/script.rs
@@ -624,16 +624,17 @@ impl TariScript {
 
         let mut agg_pub_key = RistrettoPublicKey::default();
 
-        // Create an iterator that allows each pubkey to only be checked a single time
+        // Create an iterator that allows each pubkey to only be checked a single time as they are
+        // removed from the collection when referenced
         let mut pub_keys = public_keys.iter();
 
         // Signatures and public keys must be ordered
         for (i, s) in signatures.iter().enumerate() {
-            if pub_keys.is_empty() {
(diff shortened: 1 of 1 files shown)


infrastructure/tari_script/src/script.rs
```diff
@@ -591,6 +591,7 @@ impl TariScript {
     ///
     /// Notes:
     /// * The _m_ signatures are expected to be the top _m_ items on the stack.
+    /// * The public keys and signatures must both be ordered.
     /// * Every public key can be used AT MOST once.
```
AaronFeickert:
```suggestion
    /// * The list may contain duplicate keys, but each occurrence of a public key may be used AT MOST once.
```

infrastructure/tari_script/src/script.rs
```diff
@@ -591,6 +591,7 @@ impl TariScript {
     ///
     /// Notes:
     /// * The _m_ signatures are expected to be the top _m_ items on the stack.
+    /// * The public keys and signatures must both be ordered.
```
AaronFeickert:
```suggestion
    /// * The ordering of signatures on the stack MUST match the relative ordering of the corresponding public keys.
```

infrastructure/tari_script/src/script.rs
AaronFeickert: Should also include a comment that signatures are required to be unique, a case which could legitimately arise on a repeated public key.

infrastructure/tari_script/src/script.rs
```diff
@@ -616,18 +617,24 @@ impl TariScript {
...
-            for (i, pk) in public_keys.iter().enumerate() {
-                if !sig_set.contains(s) && !key_signed[i] && s.verify_raw_canonical(pk, &message) {
-                    // This prevents Alice creating 2 different sigs against her public key
-                    key_signed[i] = true;
+            if pub_keys.len() == 0 {
+                return Ok(None);
+            }
+
+            for pk in pub_keys.by_ref() {
+                if !sig_set.contains(s) && s.verify_raw_canonical(pk, &message) {
```
AaronFeickert: I think you can move `!sig_set.contains(s)` outside of (and immediately prior to) the inner `for` loop to save on hash lookups. This means there is a single check for signature duplication. If the signature is verified and it is added to the hash set, the `break` will short-circuit subsequent lookups anyway in favor of the single check after the loop.

AaronFeickert: In fact, I think such a move improves clarity of intent as well. It means you assert that a signature initially hasn't been verified already (or else it's a duplicate and the operation fails), and then assert that the signature is eventually verified (or else the operation fails).

AaronFeickert (changes requested):

infrastructure/tari_script/src/script.rs
AaronFeickert: Tested this change locally, and all tests pass.

AaronFeickert: The use of `verify_raw_canonical` means that `message` _must_ safely bind the public key. In this case, it is identical across the full set of allowed public keys, which technically breaks Fiat-Shamir. In practice, it means that a forger would need to target any one of the public keys, the complexity of which depends on the number of public keys in the set.

AaronFeickert (changes requested):

infrastructure/tari_script/src/script.rs
AaronFeickert: See #5817 for a related issue.

SWvheerden (approved):

Issue #5908, opened 2023-11-06
Base node should ask for disabled grpc method is you want to mine


SWvheerden:
Mining on your base node required disabled by default options. The base node should ask on first startup when creating a config, if you want those enabled or not. Only initial. 

This is to enable double click on the base node to start mining.

SWvheerden:
this should be a cli option as well on startup like `-- enable_mining` or something

hansieodendaal:
See #6019

PR #5978, opened 2023-11-22
feat: validator registration timelock

Cifko:
Description
---
Add VN registration timelock. It has to be set to at least for the validity period (base on current consensus).
Fix the money lock value, add a test value of 1T to it.

Motivation and Context
---
Fixes https://github.com/tari-project/tari-dan/issues/831

How Has This Been Tested?
---
I did register a VN, and the I checked if it's timelocked in the wallet.

---
Register VN and see the timelock in the wallet.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


common/config/presets/c_base_node.toml
```diff
@@ -38,13 +38,13 @@ identity_file = "config/base_node_id_nextnet.json"
 
 # Uncomment all gRPC server methods that should be denied default (only active when `grpc_enabled = true`)
 grpc_server_deny_methods = [
-    "get_version",
```
SWvheerden: Dont change these, this was a security finding that these should not be enabled by default

SWvheerden: We have a tracking issue for the base node to ask you about these thou: https://github.com/tari-project/tari/issues/5908

This should make it easier to enable in future

SWvheerden (changes requested):

common/config/presets/c_base_node.toml
Cifko: Yeah, I forgot this here. Because I have to constantly comment them out, because I can't set them via command line.

SWvheerden: good point, we need to ensure its cli set-able as well.

SWvheerden (dismissed):

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -346,8 +344,11 @@ impl ConsensusConstants {
         self.vn_registration_min_deposit_amount
     }
 
-    pub fn validator_node_registration_min_lock_height(&self) -> u64 {
-        self.vn_registration_lock_height
+    pub fn validator_node_registration_min_lock_height(&self, height: u64) -> u64 {
+        // This calculates the height of the block in which the VN is not valid anymore.
+        self.epoch_to_block_height(VnEpoch(
+            self.block_height_to_epoch(height).as_u64() + 1 + self.vn_validity_period_epochs.as_u64(),
+        ))
```
hansieodendaal, line 356: The comment is confusing, maybe
```
// This calculates the minimum height of the block for which the VN registration maturity must be valid for.
```

hansieodendaal (dismissed):
utACK

Just one comment please.

base_layer/core/src/consensus/consensus_constants.rs
Cifko: done

hansieodendaal (dismissed):
utACK

hansieodendaal (approved):
utACK

SWvheerden:
this has gone stale, going to close for now

PR #5980, opened 2023-11-23
feat!: network specific domain hashers

### LINKED ISSUE #5652: Add network to domain seperated hashes
Currently, it's possible to replay duplicate utxos on separate networks. If we make the hashes require the network, this will enforce unique signatures per network.

brianp:
Description
---
Add a lazy static value that allows us to access what the current network is globally, so we can use the network byte to create network & domain-specific hashers.

Motivation and Context
---
This prevents the spending of duplicated UTXO's across networks as the hashes would compute differently based on the network byte.

Closes: #5652 

How Has This Been Tested?
---
CI

Breaking Changes
---

- [ ] None
- [x] Requires hard fork
- [x] Other - Please specify

This will invalidate all previous hashes on the network, and require both a network reset, with full data directory deletion.

### DIFF
--- a/applications/minotari_app_utilities/src/network_check.rs
+++ b/applications/minotari_app_utilities/src/network_check.rs
@@ -20,6 +20,8 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use std::sync::{MutexGuard, PoisonError};
+
 use tari_common::{
     configuration::{Network, CURRENT_NETWORK},
     exit_codes::{ExitCode, ExitError},
@@ -35,6 +37,8 @@ pub enum NetworkCheckError {
     NextNetBinary(Network),
     #[error("The network {0} is invalid for this binary built for TestNet")]
     TestNetBinary(Network),
+    #[error("Had a problem with the CURRENT_NETWORK guard: {0}")]
+    CurrentNetworkGuard(#[from] PoisonError<MutexGuard<'static, Network>>),
 }
 
 impl From<NetworkCheckError> for ExitError {
@@ -68,7 +72,7 @@ pub fn is_network_choice_valid(network: Network) -> Result<Network, NetworkCheck
 pub fn set_network_if_choice_valid(network: Network) -> Result<(), NetworkCheckError> {
     match is_network_choice_valid(network) {
         Ok(network) => {
-            let mut current_network = CURRENT_NETWORK.lock().unwrap();
+            let mut current_network = CURRENT_NETWORK.lock()?;
             *current_network = network;
             Ok(())
         },
--- a/applications/minotari_app_utilities/src/network_check.rs
+++ b/applications/minotari_app_utilities/src/network_check.rs
@@ -21,7 +21,7 @@
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use tari_common::{
-    configuration::Network,
+    configuration::{Network, CURRENT_NETWORK},
     exit_codes::{ExitCode, ExitError},
 };
 use tari_features::resolver::Target;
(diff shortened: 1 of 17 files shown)


base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -41,8 +42,9 @@ where D: Default
 {
     #[allow(clippy::new_ret_no_self)]
     pub fn new(label: &'static str) -> ConsensusHasher<D> {
+        let network = *CURRENT_NETWORK.lock().unwrap();
```
brianp, line 45: One thought I had was to make `CURRENT_NETWORK` an `Option<Network>` with `None` as the default. We can check to validate a `CURRENT_NETWORK` has been set, and if not fail to provide the hasher. It's a harder fault but would prevent the chance of using the wrong network in the hash and creating hard-to-track bugs.

Edit: An example of where this could go wrong otherwise is Stagenet & Mainnet. If the default network is Mainnet, but CURRENT_NETWORK isn't set, then on the stagenet network it would hash with the Mainnet network byte. This assumes once mainnet is out the default network is mainnet. Currently the default is Stagenet, so it would render mainnet hashes unusable.

SWvheerden: The only real danger comes with wallets, where you think you might be on stagenet, but actually mainnet. but when you spend to other wallets, the addresses are different so this should be blocked. 

We should either way clearly print out the current selected network on base nodes, console wallets already have a network label right bottom

SWvheerden (dismissed):

common/src/configuration/network.rs
```diff
@@ -25,12 +25,18 @@ use std::{
...
 };
 
 use serde::{Deserialize, Serialize};
+use structopt::lazy_static::lazy_static;
 
 use crate::ConfigurationError;
 
+lazy_static! {
+    pub static ref CURRENT_NETWORK: Mutex<Network> = Mutex::new(Network::default());
+}
```
AaronFeickert: Can you use `once_cell`'s `Lazy` or `OnceCell` here instead? Other uses of this pattern elsewhere are moving to them.

AaronFeickert (changes requested):

common/src/configuration/network.rs
AaronFeickert: See #5992, which migrates remaining uses of the `lazy_static!` macro.

brianp: Done and done my friend!

applications/minotari_app_utilities/src/network_check.rs
```diff
@@ -52,15 +52,26 @@ pub const TARGET_NETWORK: Target = Target::NextNet;
...
-        (Target::TestNet, Network::LocalNet | Network::Igor | Network::Esmeralda) => Ok(()),
+        (Target::TestNet, n @ Network::LocalNet | n @ Network::Igor | n @ Network::Esmeralda) => Ok(n),
         (Target::TestNet, _) => Err(NetworkCheckError::TestNetBinary(network)),
     }
 }
+
+pub fn set_network_if_choice_valid(network: Network) -> Result<(), NetworkCheckError> {
+    match is_network_choice_valid(network) {
+        Ok(network) => {
+            let mut current_network = CURRENT_NETWORK.lock().unwrap();
```
AaronFeickert: It seems [unlikely](https://doc.rust-lang.org/std/sync/struct.Mutex.html#panics), but since this could panic, it might be safer to propagate errors from `lock()` since you already return a `Result`.

base_layer/core/src/consensus/consensus_encoding/hashing.rs
AaronFeickert, line 45: Would there be any critical consequences if this were to [panic](https://doc.rust-lang.org/std/sync/struct.Mutex.html#panics)?

brianp: I think it would result in an unusable service. Which is actually a pretty good scenario. I'd rather the service crash than do something like mangle or produce bad transactions.

applications/minotari_app_utilities/src/network_check.rs
brianp: Fixed

AaronFeickert (approved):
utACK

AaronFeickert:
The changes look good, but on further thought, is it straightforward to write a sanity check that hashers for different networks are independent? That is, if you hash the same input data against two different networks, the output should _at least_ be distinct.

brianp:
Straight forward enough it shouldn't be skipped. I've written the test now and will push it once I see the CI results and what else needs changing.

AaronFeickert:
> Straight forward enough it shouldn't be skipped. I've written the test now and will push it once I see the CI results and what else needs changing.

Sounds good, thanks! I'd be surprised if the test uncovered any problems, but at least it's there in case of any future regressions.

SWvheerden (approved):
I am going to merge this as is, the test I will  mark as broken. 
I tested this locally and it works.

PR #5982, opened 2023-11-23
fix: display

SWvheerden:
Description
---
* Fix display panic in `base_layer/core/src/mempool/service/request.rs`. When a transaction without a kernel comes in, this will panic
* Fix display to no longer call FixedHash to hex, as fixedhash already impl display.


hansieodendaal (approved):
utACK

Issue #5968, opened 2023-11-20
create revealed value coinbases


SWvheerden:
Allow miners to create revealed value coinbases.

hansieodendaal:
See #5983

hansieodendaal:
See #5983

PR #5983, opened 2023-11-23
feat: enable revealed-value proofs

hansieodendaal:
Description
---
Enabled revealed-value proofs for all output types, which are be controlled via the consensus constants. Range-proof types are mapped to output types.

From the merge mining proxy/miner log, coinbase output showing `range_proof_type: RevealedValue` and `Proof: None`:

image

_**Note:** The faucets changed due to `RangeProofType` now being parsed with snake case - `"range_proof_type":"RevealedValue"` changed to `"range_proof_type":"revealed_value"` on every output line._

Motivation and Context
---
See #5968

How Has This Been Tested?
---
Added unit tests
Unit tests pass
Cucumber tests pass
System-level tests ~~[**TO BE COMPLETED**]~~

---
Code walkthrough
Review the additional unit tests
Run system-level tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -672,8 +672,33 @@ impl ConsensusConstants {
...
+                RangeProofType::RevealedValue,
+            ]),
+            (OutputType::ValidatorNodeRegistration, &[
+                RangeProofType::BulletProofPlus,
+                RangeProofType::RevealedValue,
+            ]),
+            (OutputType::CodeTemplateRegistration, &[RangeProofType::BulletProofPlus]),
+        ]
+    }
+
```
hansieodendaal: This must be reviewed and debated - this is an example. Easy to change the mapping that we want.

hansieodendaal: General consensus: For now, we will only allow `OutputType::Coinbase` to have `RangeProofType::RevealedValue` as well; all other output types will only be `RangeProofType::BulletProofPlus`.

applications/minotari_miner/src/config.rs
```diff
@@ -110,6 +113,7 @@ impl Default for MinerConfig {
             wait_timeout_on_error: 10,
             wallet_payment_address: TariAddress::default().to_hex(),
             stealth_payment: true,
+            range_proof_type: RangeProofType::BulletProofPlus,
```
SWvheerden: can we make the default revealed.

SWvheerden (commented):
This looks good, I think we just need to make the defaults revealed value so we can test them

applications/minotari_miner/src/config.rs
hansieodendaal: Let's just get @stringhandler's opinion as well

stringhandler: Default revealed please.

SWvheerden (approved):

Issue #5930, opened 2023-11-09
Remove GRPC via feature flag in wallet + 1-sided mining


SWvheerden:
We should hide and remove GRPC from the wallet only adding it in via a compile feature flag. 
The main reason for GRPC is the wallet is mining. 
This requires we must change mining to be one-sided.

hansieodendaal:
See #5967

hansieodendaal:
In the above PR the related gRPC methods and structs that supported mining have been removed from the wallet.

I do not think having a feature flag to hide the remaining gRPC methods is necessary, as we can implement `grpc_server_deny_methods` by default similar to what is done in the base node.

CjS77:
Note that launchpad uses gRPC to control the wallet via the docker API. 

So the builds for the wallet docker container should be updated to _include_ the gRPC feature when this change makes it into the code.

SWvheerden:
We should still feature flag this out. And make sure the dockers compile with the feature flag. 

The configs are not encrypted or protected in anyway, so enabling that is super easy

hansieodendaal:
See #5988 where the `features = ["grpc"]` feature flag is added.

PR #5988, opened 2023-11-27
feat!: disable console wallet grpc

hansieodendaal:
Description
---
Disabled the console wallet gRPC via a feature flag in the console wallet build configuration, where gRPC is not part of the default build.
Trying to run with the console wallet with the config option `grpc_enabled = true` where the gRPC feature is not enabled for the build will result in a runtime error.

Motivation and Context
---
See #5930

How Has This Been Tested?
---
Existing unit and cucumber tests pass

---
Code walkthrough

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - The console wallet must be build with the `features = ["grpc"]` to enable gRPC

### DIFF
--- a/applications/minotari_console_wallet/Cargo.toml
+++ b/applications/minotari_console_wallet/Cargo.toml
@@ -71,6 +71,7 @@ tari_features = { path = "../../common/tari_features"}
 
 [features]
 libtor = ["tari_libtor"]
+grpc = []
 
 [package.metadata.cargo-machete]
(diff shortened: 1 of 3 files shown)


SWvheerden (changes requested):
This looks good, but we need to add the grpc option to the docker workflow

SWvheerden (approved):
Will be fixed in another pr

PR #5991, opened 2023-11-28
feat!: dont store entire monero coinbase transaction

### LINKED ISSUE #5891: Dont store entire Monero coinbase transaction in the tari header
Monero coinbase transactions can be very large as they can have multiple outputs. 

We can store only the coinbase tx hash, using the coinbase tx hash and tx_extra data we can reconstruct the hashing blob.
There is a way to store a partial hash + coinbase tail data to reconstruct the full hash tx_extra is in the end of coinbase, so it can work
See how p2pool does it:
https://github.com/SChernykh/p2pool/blob/master/src/block_template.cpp#L1003

Storing only the 200-byte keccak state and whatever is in the tail of the coinbase tx

SWvheerden:
Description
---
only store the keccak state and not the entire monero coinbase transaction

Motivation and Context
---
The monero coinbase transaction can be very large as it may contain multiple coinbases. We only store the keccak inner hash state with half of the hashed transaction. We hash the last part to verify its working. 

How Has This Been Tested?
---
unit tests

fixes: #5891


hansieodendaal (approved):
System-level testing with Monero mainnet merged mining works fine

ACK

PR #5992, opened 2023-11-28
chore: migrate from `lazy_static` to `once_cell`

AaronFeickert:
Description
---
Migrates all remaining uses of `lazy_static` to `once_cell` for better consistency with the rest of the ecosystem.

Motivation and Context
---
Other repositories in the ecosystem are moving from `lazy_static` to `once_cell` for better maintenance and to remove the use of a macro. This repository had a mix of both still in use, so this PR finishes moving everything over. The API is very similar, so changes were minimal.

How Has This Been Tested?
---
Existing tests pass.

---
Check the correctness of each change by asserting that each use of `Lazy` reflects the same functionality as the previous `lazy_static!` macro.


AaronFeickert:
Apparently `std` now has [support](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) for the `OnceCell` and `OnceLock` APIs. It might be worth moving to this at some point, although they use a different API than `Lazy`.

SWvheerden (approved):

PR #5997, opened 2023-11-30
fix(shutdown): is_triggered returns up-to-date value without first polling

sdbondi:
Description
---

fix: always return correct value from is_triggered without first requiring the signal to be polled

Motivation and Context
---
OneshotSignal uses the `futures::Shared<Fut>` combinator to allow a single future to be cloned and awaited on in multiple places. Previously we used FuseFuture::is_terminated to determine if the signal is triggered. However the semantics of is_terminated and is_triggered are subtly different. is_terminated indicates whether a future should be polled again, which is true even if the trigger has fired. 

This PR uses an AtomicBool that is set to true at the time the trigger is fired so that polling first is not required.

Thanks to @hansieodendaal for discovering this issue

How Has This Been Tested?
---
Updated existing test to check is_triggered, previously failed and now passes.

---
N/A

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/infrastructure/shutdown/src/lib.rs
+++ b/infrastructure/shutdown/src/lib.rs
@@ -25,12 +25,11 @@ pub mod oneshot_trigger;
 use std::{
     future::Future,
     pin::Pin,
+    sync::{atomic, atomic::AtomicBool, Arc},
     task::{Context, Poll},
 };
 
 use futures::{future, future::FusedFuture};
-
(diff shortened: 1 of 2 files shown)


SWvheerden (approved):

hansieodendaal (approved):
ACK

Nice

PR #6004, opened 2023-12-02
chore: refactor current network handling and add test

### LINKED ISSUE #6003: Add test for network-specific hashing
Recent work in #5980 changes consensus hashing to bind to the network. It would be useful to add a sanity check that hashing the same input with different networks yields distinct hash outputs.

AaronFeickert:
Description
---
Refactors the handling of network-based hashing operations introduced in #5980 to better handle read operations. Adds a sanity test for hash independence.

Supersedes #6014.

Closes #6003.

Motivation and Context
---
Recent work in #5980 binds the current network into consensus hashing. It uses a `Mutex`, which has two subtle issues. First, it allows the network to be set multiple times, which should not occur. Second, it locks on reads, which is unnecessary and inefficient.

This PR updates how the current network is handled. It adds `Network::get_current_or_default` that will return either the current network (if it has been set) or the default network (if it has not). It adds `Network:set_current` that attempts to set the network; if it has been set before, this operation will fail. Note that calling `Network::get_current_or_default` does _not_ set the network for this purpose.

It modifies the API for consensus hashing to add a wrapper that uses the current network. This wraps functionality allowing for specification of a network, which is useful for testing.

On top of this new API, a new test is added that checks for distinct hashes using the same input but different networks. This is not comprehensive, but will detect obvious issues.

How Has This Been Tested?
---
Existing tests pass. A new test passes.

---
Check that the tests do what they claim. Check that the updates to consensus hashing properly introduce the expected wrapping functionality. Check that the updated network API does what it is supposed to.

### DIFF
--- a/base_layer/core/src/consensus/consensus_encoding/hashing.rs
+++ b/base_layer/core/src/consensus/consensus_encoding/hashing.rs
@@ -28,7 +28,7 @@ use digest::{
     consts::{U32, U64},
     Digest,
 };
-use tari_common::configuration::CURRENT_NETWORK;
+use tari_common::configuration::Network;
 use tari_crypto::{hash_domain, hashing::DomainSeparation};
 
 /// Domain separated consensus encoding hasher.
@@ -42,7 +42,10 @@ where D: Default
 {
     #[allow(clippy::new_ret_no_self)]
     pub fn new(label: &'static str) -> ConsensusHasher<D> {
-        let network = *CURRENT_NETWORK.lock().unwrap();
+        Self::new_with_network(label, Network::get_current_or_default())
+    }
+
+    pub fn new_with_network(label: &'static str, network: Network) -> ConsensusHasher<D> {
         let mut digest = D::default();
         M::add_domain_separation_tag(&mut digest, &format!("{}.n{}", label, network.as_byte()));
         ConsensusHasher::from_digest(digest)
@@ -140,6 +143,7 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
 mod tests {
     use blake2::Blake2b;
     use digest::consts::U32;
+    use tari_common::configuration::Network;
     use tari_crypto::hash_domain;
     use tari_script::script;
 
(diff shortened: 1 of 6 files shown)


base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -140,13 +140,40 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
...
+        let label = "test";
+        let input = [1u8; 32];
+
+        assert_eq!(*CURRENT_NETWORK.lock().unwrap(), Network::Esmeralda);
+
+        // Generate a mainnet hash
+        *CURRENT_NETWORK.lock().unwrap() = Network::MainNet;
+        let hash_mainnet = DomainSeparatedConsensusHasher::<TestHashDomain, Blake2b<U32>>::new(label)
+            .chain(&input)
+            .finalize();
```
brianp: These tests look pretty much like what I wrote, which didn't get pushed because I caught a stomach bug and died Thursday night.

I am surprised this works though. I thought this wouldn't release the mutex lock after update, making it impossible for the Domain hasher to read the change?

AaronFeickert: It definitely hangs if you keep the `Mutex` in scope via a `let mut`. But as written, I assume the compiler drops it after each use.

brianp: ohhhh a subtle difference. Big consequences. Makes sense 👍🏻

AaronFeickert: I do notice more failing integration tests than usual, but this seems to be the case for other recent PRs as well; likely unrelated?

AaronFeickert:
There is an issue that arises when hashing tests are run in parallel, where the `CURRENT_NETWORK` mutex is in an unexpected state that fails one or more of the existing unit tests. This doesn't seem to occur during CI, but I can reliably reproduce it locally: `cargo test --package tari_core --lib -- consensus::consensus_encoding::hashing::tests --nocapture`

AaronFeickert:
After discussing with @brianp, it sounds like changing `CURRENT_NETWORK` to be a `RwLock` is more idiomatic. Its type would be changed to `RwLock<Option<Network>>` and initially set to `None`. This ensures that the network is always in a known state that's intentionally set.

This doesn't solve the parallel test problem, unfortunately. Configuring those tests to run serially might be the best option, though I'm not sure how to do that.

SWvheerden (dismissed):

PR #6008, opened 2023-12-04
feat!: update status

SWvheerden:
Description
---
Changes statuses of coinbase transactions and one-sided transactions

Motivation and Context
---
Transactions now show as one-sided or coinbase, not Faux anymore. This is more inline of what they are and shows more information to the user.

How Has This Been Tested?
---
manual + unit tests

Breaking Changes
---
Wallet database changes, and requires recovery to keep existing database.

### DIFF
--- a/base_layer/common_types/src/transaction.rs
+++ b/base_layer/common_types/src/transaction.rs
@@ -15,37 +15,95 @@ pub use crate::tx_id::TxId;
 #[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum TransactionStatus {
     /// This transaction has been completed between the parties but has not been broadcast to the base layer network.
-    Completed,
+    Completed = 0,
     /// This transaction has been broadcast to the base layer network and is currently in one or more base node
     /// mempools.
-    Broadcast,
+    Broadcast = 1,
     /// This transaction has been mined and included in a block.
-    MinedUnconfirmed,
+    MinedUnconfirmed = 2,
     /// This transaction was generated as part of importing a spendable unblinded UTXO
-    Imported,
+    Imported = 3,
     /// This transaction is still being negotiated by the parties
     #[default]
-    Pending,
+    Pending = 4,
     /// This is a created Coinbase Transaction
-    Coinbase,
+    Coinbase = 5,
     /// This transaction is mined and confirmed at the current base node's height
-    MinedConfirmed,
+    MinedConfirmed = 6,
     /// This transaction was Rejected by the mempool
-    Rejected,
-    /// This is faux transaction mainly for one-sided transaction outputs or wallet recovery outputs have been found
-    FauxUnconfirmed,
-    /// All Imported and FauxUnconfirmed transactions will end up with this status when the outputs have been confirmed
-    FauxConfirmed,
+    Rejected = 7,
+    /// This transaction import status is used when a one-sided transaction has been scanned but is unconfirmed
+    OneSidedUnconfirmed = 8,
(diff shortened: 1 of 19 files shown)


base_layer/common_types/src/transaction.rs
```diff
@@ -15,38 +15,96 @@ pub use crate::tx_id::TxId;
...
+            TransactionStatus::Coinbase |
+            TransactionStatus::Rejected |
+            TransactionStatus::Queued |
+            TransactionStatus::MinedUnconfirmed |
+            TransactionStatus::MinedConfirmed => TransactionStatus::MinedConfirmed,
+            TransactionStatus::Imported |
+            TransactionStatus::OneSidedUnconfirmed |
+            TransactionStatus::OneSidedConfirmed => TransactionStatus::OneSidedConfirmed,
+            TransactionStatus::CoinbaseConfirmed | TransactionStatus::CoinbaseUnconfirmed => {
+                TransactionStatus::CoinbaseConfirmed
```
hansieodendaal, line 85: A coinbase can also be imported

base_layer/common_types/src/transaction.rs
```diff
@@ -15,38 +15,96 @@ pub use crate::tx_id::TxId;
...
+            TransactionStatus::Coinbase |
+            TransactionStatus::Rejected |
+            TransactionStatus::Queued |
+            TransactionStatus::MinedUnconfirmed |
+            TransactionStatus::MinedConfirmed => TransactionStatus::MinedUnconfirmed,
+            TransactionStatus::Imported |
+            TransactionStatus::OneSidedUnconfirmed |
+            TransactionStatus::OneSidedConfirmed => TransactionStatus::OneSidedUnconfirmed,
+            TransactionStatus::CoinbaseConfirmed | TransactionStatus::CoinbaseUnconfirmed => {
+                TransactionStatus::CoinbaseUnconfirmed
```
hansieodendaal, line 104: A coinbase can also be imported

hansieodendaal (commented):
Hi, some comments:

- The states a one-sided or coinbase scanned output goes through are exactly the same, and adding differentiation to statuses just adds complexity to the code.
- I think we should rather add an `is_scanned` column to the completed transactions (and potentially the outputs table as well) and track it that way - values could be `0 - not_scanned`, `1 - one_sided` and `2 - coinbase`. (_The outputs table already differentiates between the output types, but does not indicate if it was scanned or not._)
- When an output is imported or scanned, the `is_scanned` column can be populated based on the output features output type and associated script.

base_layer/common_types/src/transaction.rs
SWvheerden: When scanned, the coinbase state takes preference, and when important via import the import state takes preference. I think that's correct.

base_layer/common_types/src/transaction.rs
SWvheerden: When scanned, the coinbase state takes preference, and when important via import the import state takes preference. I think that's correct.

SWvheerden:
I disagree with the `is_scanned `column. The current wallet design works with those statuses to indicate what it is, where it came from, and in what state it is. Adding a new separate enum to indicate the status add complexity.

SWvheerden:
I have added a tech debt issue for addressing after this release: https://github.com/tari-project/tari/issues/6010

This is much more inline of what it should be.  But its a much bigger refactor.

hansieodendaal:
> Wallet database changes, and requires recovery to keep existing database.

This is not strictly true, as only the statuses of old scanned coinbase transactions will be reported incorrectly.

hansieodendaal (approved):
ACK

_(system-level tests performed)_

PR #6014, opened 2023-12-05
refactor: better network global

stringhandler:
Description
---
Removes the mutex and allows tests to run in parallel

Motivation and Context
---
Built on #6004 

The previous versions of these test relied on setting the Network globally. If other tests are running at the same time, these tests can fail because the network is switched. I investigated using ThreadLocalStorage and TaskLocalStorage from tokio, but these could result threads having different networks set and it would be hard to confirm that the entire process was using the correct network.

I opted instead to ignore the `set_network` test and add a parameter to the DomainSeparatedHasher to allow a different network to be tested.

How Has This Been Tested?
---
Ran cargo test

---
This fixes tests and provides a cleaner interface

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_app_utilities/src/network_check.rs
+++ b/applications/minotari_app_utilities/src/network_check.rs
@@ -79,3 +79,28 @@ pub fn set_network_if_choice_valid(network: Network) -> Result<(), NetworkCheckE
         Err(e) => Err(e),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn set_network() {
+        assert_eq!(*CURRENT_NETWORK.lock().unwrap(), Network::Esmeralda);
+
+        // Invalid networks
+        for network in [Network::MainNet, Network::StageNet, Network::NextNet] {
+            assert!(set_network_if_choice_valid(network).is_err());
+            assert_eq!(*CURRENT_NETWORK.lock().unwrap(), Network::Esmeralda);
+        }
+
+        // Valid networks (we can't test against other binary configurations)
+        for network in [Network::LocalNet, Network::Igor, Network::Esmeralda] {
+            assert!(set_network_if_choice_valid(network).is_ok());
+            assert_eq!(*CURRENT_NETWORK.lock().unwrap(), network);
+        }
+
+        // Ensure we've restored the network for other tests
+        assert_eq!(*CURRENT_NETWORK.lock().unwrap(), Network::Esmeralda);
+    }
+}
--- a/applications/minotari_app_utilities/src/network_check.rs
+++ b/applications/minotari_app_utilities/src/network_check.rs
@@ -20,10 +20,8 @@
 //  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::sync::{MutexGuard, PoisonError};
-
 use tari_common::{
-    configuration::{Network, CURRENT_NETWORK},
+    configuration::Network,
     exit_codes::{ExitCode, ExitError},
 };
 use tari_features::resolver::Target;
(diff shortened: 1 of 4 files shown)


base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -140,16 +144,44 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
...
+        let hash_stagenet =
+            DomainSeparatedConsensusHasher::<TestHashDomain, Blake2b<U32>>::new_with_network(label, Network::StageNet)
+                .chain(&input)
+                .finalize();
+
+        // They should be distinct
+        assert_ne!(hash_mainnet, hash_stagenet);
+
+        // Restore the network for other tests
+        //        *CURRENT_NETWORK.lock().unwrap() = Network::Esmeralda;
```
SWvheerden:
```suggestion
```

SWvheerden (dismissed):
this looks like it could solve the test problem

brianp (approved):
LGTM

applications/minotari_app_utilities/src/network_check.rs
```diff
@@ -72,10 +73,42 @@ pub fn is_network_choice_valid(network: Network) -> Result<Network, NetworkCheck
 pub fn set_network_if_choice_valid(network: Network) -> Result<(), NetworkCheckError> {
     match is_network_choice_valid(network) {
         Ok(network) => {
-            let mut current_network = CURRENT_NETWORK.lock()?;
-            *current_network = network;
-            Ok(())
+            // let mut current_network = Network::current();
```
AaronFeickert, line 76:
```suggestion
```

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -140,16 +144,42 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
...
 
     hash_domain!(TestHashDomain, "com.tari.test.test_hash", 0);
 
+    #[test]
+    fn network_yields_distinct_hash() {
+        let label = "test";
+        let input = [1u8; 32];
+
+        // assert_eq!(Network::current(), Network::Esmeralda);
+
```
AaronFeickert, line 161:
```suggestion
```

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -140,16 +144,42 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
...
+        // Generate a mainnet hash
+        // Network::set_current(Network::MainNet);
```
AaronFeickert, line 163:
```suggestion
```

base_layer/core/src/consensus/consensus_encoding/hashing.rs
```diff
@@ -140,16 +144,42 @@ impl<D: Digest> Write for WriteHashWrapper<D> {
...
+
+        // Generate a stagenet hash
+        let hash_stagenet =
+            DomainSeparatedConsensusHasher::<TestHashDomain, Blake2b<U32>>::new_with_network(label, Network::StageNet)
+                .chain(&input)
+                .finalize();
+
+        // They should be distinct
+        assert_ne!(hash_mainnet, hash_stagenet);
+
```
AaronFeickert, line 177:
```suggestion
```

applications/minotari_app_utilities/src/network_check.rs
```diff
@@ -72,10 +73,42 @@ pub fn is_network_choice_valid(network: Network) -> Result<Network, NetworkCheck
...
+
+        // Invalid networks
+        for network in [Network::MainNet, Network::StageNet, Network::NextNet] {
+            assert!(set_network_if_choice_valid(network).is_err());
+            assert_eq!(Network::current(), Network::Esmeralda);
+        }
+
+        // Valid networks (we can't test against other binary configurations)
+        for network in [Network::LocalNet, Network::Igor, Network::Esmeralda] {
+            assert!(set_network_if_choice_valid(network).is_ok());
```
AaronFeickert, line 106: This fails for me locally on the first trip through the loop. The issue is that `Network::set_current` returns an error once the value has been initialized. I don't think this behavior is correct.

AaronFeickert (changes requested):

SWvheerden:
Closing this for: https://github.com/tari-project/tari/pull/6004

PR #6032, opened 2023-12-11
chore: update dependencies due to security from dependabot

SWvheerden:
Description
---
update dependencies due to security from dependabot


brianp (approved):
LGTM utAck

PR #6034, opened 2023-12-12
feat: console wallet use dns seeds

SWvheerden:
Description
---
Console wallet use dns seeds

Motivation and Context
---
The console wallet should use the DNS seed, and not just the seeds in the config file 

How Has This Been Tested?
---
Manual

### DIFF
--- a/applications/minotari_console_wallet/src/init/mod.rs
+++ b/applications/minotari_console_wallet/src/init/mod.rs
@@ -50,7 +50,7 @@ use tari_common::{
 };
 use tari_comms::{
     multiaddr::Multiaddr,
-    peer_manager::{Peer, PeerFeatures},
+    peer_manager::{Peer, PeerFeatures, PeerQuery},
     types::CommsPublicKey,
     NodeIdentity,
 };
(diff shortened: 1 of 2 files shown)


stringhandler:
Nice find

brianp (commented):
I think I want to test this.

but a clarification if you will:

> should use the DNS seed, and not just the seeds in the config file

> *and not just*

So it should use both? Because the change swings the pendulum, to ignore the config seeds, and use whatever the PeerDB already has.
At what point did it fetch DNS seeds?

SWvheerden:
> > ld use the DNS seed, and not just the seeds in the config file
> 
> > _and not just_

it should use both:

https://github.com/tari-project/tari/blob/851cb0070f057399d128df37970916250a5cf1ec/base_layer/p2p/src/initialization.rs#L583

brianp (approved):
LGTM

Thanks for the response.

Tested locally:
Created new wallet on esme network. Removed the peers from the config. Deleted the dht, and peer db. Restarted the wallet and it started finding peers instantly, found and connected to a base node, and carried on.

👍🏻 

Ack (sans ut)

PR #6043, opened 2023-12-13
feat: fix windows installer

hansieodendaal:
Description
---
A recent PR changed how the default `config.toml` file is created based on user input, so that it cannot be created ahead of time anymore. This PR applies that behaviour to the Windows installer.

Motivation and Context
---
See above

How Has This Been Tested?
---
Generated the install file, tested the installation as well as the runtime

---
As above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (approved):
utAck

PR #6049, opened 2023-12-14
feat: make all apps non interactive

SWvheerden:
Description
---
make all applications hnour the` non-interactive` flag 
Make base node non mining config default, and disable grpc server

How Has This Been Tested?
---
manual

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/cli.rs
+++ b/applications/minotari_merge_mining_proxy/src/cli.rs
@@ -30,6 +30,8 @@ use tari_common::configuration::{ConfigOverrideProvider, Network};
 pub struct Cli {
     #[clap(flatten)]
     pub common: CommonCliArgs,
+    #[clap(short, long, alias = "non-interactive", env = "TARI_NON_INTERACTIVE")]
+    pub non_interactive_mode: bool,
 }
 
 impl ConfigOverrideProvider for Cli {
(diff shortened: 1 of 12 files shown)


applications/minotari_merge_mining_proxy/src/cli.rs
```diff
@@ -30,6 +30,8 @@ use tari_common::configuration::{ConfigOverrideProvider, Network};
 pub struct Cli {
     #[clap(flatten)]
     pub common: CommonCliArgs,
+    #[clap(short, long, alias = "non-interactive", env = "TARI_NON_INTERACTIVE")]
```
hansieodendaal, line 33:
```suggestion
    #[clap(short, long, alias = "non-interactive", env = "MINOTARI_NON_INTERACTIVE")]
```

applications/minotari_miner/src/cli.rs
```diff
@@ -37,6 +37,8 @@ pub struct Cli {
     pub miner_min_diff: Option<u64>,
     #[clap(long, alias = "max-difficulty")]
     pub miner_max_diff: Option<u64>,
+    #[clap(short, long, alias = "non-interactive", env = "TARI_NON_INTERACTIVE")]
```
hansieodendaal, line 40:
```suggestion
    #[clap(short, long, alias = "non-interactive", env = "MINOTARI_NON_INTERACTIVE")]
```

hansieodendaal (approved):
utACK

The comments are probably just nits

applications/minotari_merge_mining_proxy/src/cli.rs
SWvheerden: There is a good chance this will be used over both base layer apps and dan, prefer to keep it as is for now

PR #6058, opened 2023-12-19
fix: make tor startup async

stringhandler:
Description
---
Move tor connection to be lazy

Motivation and Context
---
Previously, the wallet_create method would wait for tor to connect before it would return. This makes for a very slow experience, especially if you just want to read values from the database.

How Has This Been Tested?
---
This is still in progress, but I'm testing with some adhoc timing methods in the wallet_create method.

---
Kind of difficult, it can be seen mostly in the ios and android apps

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/p2p/src/initialization.rs
+++ b/base_layer/p2p/src/initialization.rs
@@ -70,6 +70,7 @@ use tari_storage::{
 use thiserror::Error;
 use tokio::sync::{broadcast, mpsc};
 use tower::ServiceBuilder;
+use tari_comms::transports::HiddenServiceTransport;
 
 use crate::{
     comms_connector::{InboundDomainConnector, PubsubDomainConnector},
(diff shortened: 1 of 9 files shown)


base_layer/p2p/src/initialization.rs
```diff
@@ -251,20 +252,17 @@ pub async fn spawn_comms_using_transport(
             let listener_address_override = tor_config.listener_address_override.clone();
             let mut hidden_service_ctl = initialize_hidden_service(tor_config)?;
             // Set the listener address to be the address (usually local) to which tor will forward all traffic
-            let transport = hidden_service_ctl.initialize_transport().await?;
+            let instant = Instant::now();
+            let transport = HiddenServiceTransport::new(hidden_service_ctl);
+            error!(target: LOG_TARGET, "TOR transport initialized in {:.0?}", instant.elapsed());
```
hansieodendaal, line 257: Why would this be an error log and not an info or debug log?

SWvheerden:
replaced by : https://github.com/tari-project/tari/pull/6092

PR #6060, opened 2023-12-20
feat: make libtor on by default for nix builds

### LINKED ISSUE #6050: console wallet for all unix systems should have libtor compiled in
the console wallet should use lib_tor like the base node does

brianp:
Description
---
This changes the defaults to include libtor for the MinoTari Node and Console wallet by default. Removing the need to run tor independently.

Closes #6050 

Motivation and Context
---
Make setup a smidge easier

How Has This Been Tested?
---
Locally: Ran the node and wallet without having a tor session started on the machine. Node and wallet used the built in libtor no problem. 

Changed `use_libtor` to `false` in the config. Started the node and received a `ExitError { exit_code: TorOffline` error as expected.
Turned on local tor session, and restarted node with no issue. Node is now running on provided tor, and not built in.

Windows built without error in CI: https://github.com/brianp/tari/actions/runs/7274136545

@Cifko did the windows testing for me. He could compile natively as usual without any libtor related problems. 
He also ran the installed package from the test builds with no issue.

This confirms that us setting a libtor default for *nix systems is not currently interfering with windows compilation or builds.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_node/Cargo.toml
+++ b/applications/minotari_node/Cargo.toml
@@ -52,7 +52,7 @@ tonic = { version = "0.6.2", features = ["tls", "tls-roots" ] }
 tari_metrics = { path = "../../infrastructure/metrics", optional = true, features = ["server"] }
 
 [features]
-default = ["metrics"]
+default = ["metrics", "libtor"]
 metrics = ["tari_metrics", "tari_comms/metrics"]
 safe = []
 libtor = ["tari_libtor"]
--- a/.github/workflows/base_node_binaries.json
+++ b/.github/workflows/base_node_binaries.json
@@ -7,7 +7,7 @@
     "cross": false,
     "target_cpu": "x86-64",
     "target_bins": "--bin minotari_node --bin minotari_console_wallet --bin minotari_merge_mining_proxy --bin minotari_miner",
-    "features": "safe"
+    "features": "default, safe"
   },
   {
     "name": "linux-arm64",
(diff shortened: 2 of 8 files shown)


applications/minotari_node/Cargo.toml
```diff
@@ -52,7 +52,7 @@ tonic = { version = "0.6.2", features = ["tls", "tls-roots" ] }
 tari_metrics = { path = "../../infrastructure/metrics", optional = true, features = ["server"] }
 
 [features]
-default = ["metrics"]
+default = ["metrics", "libtor"]
```
delta1, line 55: this will break windows builds

brianp: This is what we're testing.
Including `tari_libtor` shouldn't cause any kind of error, and then inside `tari_libtor`, libtor is actually compiled out on windows. So in theory it shouldn't break because it never really gets included. We also specify not to include tari_libtor for our windows builds. So let's see what happens! 

Don't worry we're not merging until we validate.

delta1: nice, I don't actually care about windows - but I sweated a lot trying to get it to work on there :weary:  

(the tor project itself doesn't even compile natively on windows, they cross compile from linux with msys2/mingw)

brianp: > they cross compile from linux with msys2/mingw)

That's an interesting note. Not that I'm going to dive into it, but that our pre-compiled windows bin *could* actually contain libtor possibly, as long as we cross compile.

Compiling natively would never have it, but it's an interesting option.

delta1: oh that's definitely interesting, good point.

leet4tari (approved):
LGTM

stringhandler (approved):

PR #6067, opened 2024-01-03
chore: upgrade tonic and prost

brianp:
Description
---
We were hoping to get moved onto a more modern version of both but this will not be possible. Between versions 0.8.4 and 0.9.0 of tonic they dropped support for the rust 2018 version. We would also need to upgrade to 2021 to make further use of more recent versions of tonic.

This PR upgrades them to the most recent possible.

I additionally tried seeing if we could run older versions only _where nedded_ and run more modern version in other places but most of our tonic use is tied to the grpc and the grpc clients require the older version, in turn most everything requires the old version.

Motivation and Context
---
Upgrade the old to something newer.

How Has This Been Tested?
---
CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
I can't figure out how to silence clippy with proto files. I tried both:

```
#![allow(clippy::all)]
#[allow(clippy::all)]
```

AaronFeickert:
Looks like the lints are `pedantic`, which isn't included in `clippy::all`. [This source](https://stackoverflow.com/questions/74977003/how-to-prevent-cargo-clippy-from-analyizing-generated-prost-files) suggests using `#![allow(clippy::all, clippy::pedantic)]` should work.

brianp:
@AaronFeickert lol *all* but not everything 🙃

SWvheerden (approved):

AaronFeickert:
> @AaronFeickert lol _all_ but not everything 🙃

\* some restrictions apply

PR #6073, opened 2024-01-11
feat: turn off node metrics by default

### LINKED ISSUE #5787: Metrics in base_node should be opt-out by default
Metrics in the base node should be an optional compile-in. This should not be an always-enabled option and should be opt-in. 
The library used Prometheus is full of unwrap code and features no error handles.

brianp:
Description
---
This make the node metrics opt-in instead of opt-out at compile time.

It is strange we need to leave metrics on in the integration tests. This
is mostly because clippy is often run with `--all-features` which means
metrics would be on in the node, and then required in the integration
configuration.

Closes: #5787 

Motivation and Context
---
Mostly the metrics library has a chance of failure (panics) and it's generally better (and more private) if we just have metrics off by default.

How Has This Been Tested?
---
CI

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_node/Cargo.toml
+++ b/applications/minotari_node/Cargo.toml
@@ -52,7 +52,7 @@ tonic = { version = "0.6.2", features = ["tls", "tls-roots" ] }
 tari_metrics = { path = "../../infrastructure/metrics", optional = true, features = ["server"] }
 
 [features]
-default = ["metrics", "libtor"]
+default = ["libtor"]
 metrics = ["tari_metrics", "tari_comms/metrics"]
 safe = []
 libtor = ["tari_libtor"]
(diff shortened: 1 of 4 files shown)


applications/minotari_node/Cargo.toml
```diff
@@ -52,7 +52,7 @@ tonic = { version = "0.6.2", features = ["tls", "tls-roots" ] }
 tari_metrics = { path = "../../infrastructure/metrics", optional = true, features = ["server"] }
 
 [features]
-default = ["metrics", "libtor"]
+default = ["libtor"]
```
SWvheerden, line 55: With this PR, we need to ensure that the seed nodes turn this on

SWvheerden (commented):
this looks good, just one thing we need to decide is how the seed nodes gets compiled vs downloaded binaries

SWvheerden (approved):

PR #6074, opened 2024-01-11
fix: wallet coinbases not validated correctly

SWvheerden:
Description
---
Fixes adding inputs scanned from the block to be improperly marked on adding to the database.

Motivation and Context
---
Inputs scanned, such as coinbases are marked as `unspent` while this is technically true from the wallet's perspective as they are unspent and are detected from a mined block. We need to mark them as `unspentUnconfiremd`. We need to make sure they are properly confirmed by the validation task. The validation task will pick up these outputs and flag them as spent/reorged etc. 

But in the current case because they are flagged as `unspent` the wallet can select them when spending, although they might not have passed the confirmation time, or the might have been reorged out of the main chain.

### DIFF
--- a/base_layer/wallet/tests/output_manager_service_tests/service.rs
+++ b/base_layer/wallet/tests/output_manager_service_tests/service.rs
@@ -300,7 +300,7 @@ async fn generate_sender_transaction_message(
 async fn fee_estimate() {
     let (connection, _tempdir) = get_temp_sqlite_database_connection();
     let backend = OutputManagerSqliteDatabase::new(connection.clone());
-    let mut oms = setup_output_manager_service(backend, true).await;
+    let mut oms = setup_output_manager_service(backend.clone(), true).await;
 
     let uo = make_input(
         &mut OsRng.clone(),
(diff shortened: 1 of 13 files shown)


base_layer/wallet/tests/output_manager_service_tests/service.rs
```diff
@@ -309,7 +309,11 @@ async fn fee_estimate() {
         &oms.key_manager_handle,
     )
     .await;
-    oms.output_manager_handle.add_output(uo, None).await.unwrap();
+    oms.output_manager_handle.add_output(uo.clone(), None).await.unwrap();
+    backend
+        .mark_output_as_unspent(uo.hash(&oms.key_manager_handle).await.unwrap())
+        .unwrap();
```
hansieodendaal, line 315: Could these two calls not be combined into a helper function? Then used throughout.

hansieodendaal (commented):
Due to the failing cucumber tests, it seems something needs to be changed there so that the status is updated as unspent when outputs are created.

base_layer/wallet/tests/output_manager_service_tests/service.rs
SWvheerden: Well its a 4 lines trimmed by fmt, 2 commands. Not that big a gain. 
As an bonus, you can see what you need to set to do this. 
I don't see any benefit to one.

hansieodendaal (approved):
utACK

PR #6087, opened 2024-01-19
[wip]chore: fix cucumber test

SWvheerden:
Description
---
makes the last log, a best effort. 
If the mutex is held in a another thread which is sometimes is, the lock will fail, failing the entire test, and its just a logging issue.

### DIFF
--- a/base_layer/core/src/proof_of_work/difficulty.rs
+++ b/base_layer/core/src/proof_of_work/difficulty.rs
@@ -113,25 +113,6 @@ impl Default for Difficulty {
     }
 }
 
-/// This trait is used to add a type to `CheckedAdd`, which greatly simplifies usage in the code.
-/// It is implemented for `Difficulty` and `u64`.
-pub trait CheckedAdd<T> {
-    fn checked_add(&self, other: T) -> Option<Self>
-    where Self: Sized;
-}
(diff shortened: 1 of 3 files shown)


stringhandler (dismissed):

integration_tests/tests/cucumber.rs
```diff
@@ -117,17 +120,17 @@ fn main() {
...
-        fs::copy(file.path(), test_run_dir.join(file.file_name())).unwrap();
-        fs::remove_file(file.path()).unwrap();
-    }
+    // fs::create_dir_all(&test_run_dir).unwrap();
+    //
+    // for entry in fs::read_dir(log_dir).unwrap() {
+    //     let file = entry.unwrap();
+    //     fs::copy(file.path(), test_run_dir.join(file.file_name())).unwrap();
+    //     fs::remove_file(file.path()).unwrap();
+    // }
```
AaronFeickert: Why remove this? Is it because of `unwrap` failing the test?

hansieodendaal: Yeah, same question here.

SWvheerden: I believe it is related

PR #6095, opened 2024-01-23
chore: rename chain metadata fields to be consistent with its meaning

hansieodendaal:
Description
---
Renamed `best_block` and `height_of_longest_chain` in `ChainMetadata` to be technically more correct and assist with future maintenance tasks:

```rust
pub struct ChainMetadata {
    height_of_longest_chain: u64,     ->   best_block_height: u64,
    best_block: BlockHash,            ->   best_block_hash: BlockHash,
    pruning_horizon: u64,
    pruned_height: u64,
    accumulated_difficulty: U256,
    timestamp: u64,
}
```

Motivation and Context
---
- `best_block` can easily be confused with the actual best block and not the hash;
- `height_of_longest_chain` is technically not the same as `best_block_height`; in some re-org cases they may not be the same;
- using `best_block_hash` and `best_block_height` consistently also remove any confusion that may exist that `best_block` and `height_of_longest_chain` may not refer to metadata of the same entity on the blockchain.

How Has This Been Tested?
---
- `cargo clippy`

---
Code walk through - no technical changes were made

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_app_grpc/proto/base_node.proto
+++ b/applications/minotari_app_grpc/proto/base_node.proto
@@ -310,9 +310,9 @@ enum Sorting {
 
 message MetaData {
     // The current chain height, or the block number of the longest valid chain, or `None` if there is no chain
-    uint64 height_of_longest_chain = 1;
+    uint64 best_block_height = 1;
     // The block hash of the current tip of the longest valid chain, or `None` for an empty chain
-    bytes best_block = 2;
+    bytes best_block_hash = 2;
     // This is the min height this node can provide complete blocks for. A 0 here means this node is archival and can provide complete blocks for every height.
     uint64 pruned_height = 6;
     // The current geometric mean of the pow of the chain tip, or `None` if there is no chain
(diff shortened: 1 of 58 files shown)


SWvheerden:
I understand why you made this as our internal code can be confusing due to the way our code works. 
I am however not in favour of this rename because:
1: From a blockchain perspective, `height_of_longest_chain` is the correct wording, as we use PoW which uses the longest chain as the tip block. 
2: I would argue that `height_of_longest_chain` is the technically correct word here as its the longest chain, aka header + block. 
3: We should not reflect inside weirding naming and technicalities outside. Internally we use best_block and best_header. They are two distinctively different things,
4: We are changing the public interface here, requiring more than just Tari changes here. 


But this is only my opinion. I am open to more opinions and perhaps we can even find better names here for this.

hansieodendaal:
In my previous PR I spent quite a bit of time trying to match meaning with names, therefore this PR.

brianp:
> height_of_longest_chain is technically not the same as best_block_height; in some re-org cases they may not be the same;

Are we saying that...
Using psuedo code here as an example:

`assert!(find_block_by_hash(metadata.best_block_hash).height !== metadata.best_block_height)`

SWvheerden:
No that should be true always. 

what we have internally that started this route is this:
```
assert!(find_tip_header().hash() !== metadata.best_block_hash)
```

stringhandler:
I way prefer these names suggested by @hansieodendaal

SWvheerden (approved):
Cool, looks like this is the prefered name, so lets rename it

PR #6101, opened 2024-01-26
feat: update api

SWvheerden:
Description
---
Add FFI call to get the network of a TariAdress
Add Tari Unique ID for merge mining to mining calls

Motivation and Context
---
Small improvements to tari's outside API calls.

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -1237,6 +1237,38 @@ pub unsafe extern "C" fn tari_address_to_emoji_id(
     CString::into_raw(result)
 }
 
+/// Creates a char array from a TariWalletAddress's network
+///
+/// ## Arguments
+/// `address` - The pointer to a TariWalletAddress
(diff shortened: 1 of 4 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -1237,6 +1237,38 @@ pub unsafe extern "C" fn tari_address_to_emoji_id(
...
+/// as an out parameter.
+///
+/// ## Returns
+/// `*mut c_char` - Returns a pointer to a char array. Note that it returns empty
+/// if there was an error from TariWalletAddress
+///
+/// # Safety
+/// The ```string_destroy``` method must be called when finished with a string from rust to prevent a memory leak
+#[no_mangle]
+pub unsafe extern "C" fn tari_address_network(address: *mut TariWalletAddress, error_out: *mut c_int) -> *mut c_char {
```
hansieodendaal, line 1254: We probably need to include this method in a unit test

hansieodendaal (approved):
utACK

PR #6113, opened 2024-01-31
feat: wallet add restart validation to start

SWvheerden:
Description
---
Adds back the restart tx protocols on wallet start

Motivation and Context
---
The wallet should restart its transactions protocols on start


brianp (approved):
LGTM

Spoke with SW offline about adding this back in, as it was once removed. Validated it should have been put back and not left out.

PR #6114, opened 2024-01-31
feat: do validation after adding utxos and txs

SWvheerden:
Description
---
Trigger a validation task after recovery of outputs

Motivation and Context
---
During testing, new outputs are added slower than validation and an edge case occurs where outputs will only be validated after a new block is added again.

### DIFF
--- a/base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
+++ b/base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
@@ -606,6 +606,10 @@ where
                 })
                 .collect::<Result<Vec<_>, _>>()?,
         );
+        if !found_outputs.is_empty() {
+            // this is best effort, if this fails, its very likely that its already busy with a validation.
+            let _result = self.resources.output_manager_service.validate_txos().await;
+        }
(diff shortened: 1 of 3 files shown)


hansieodendaal (commented):
Maybe a better place to add this would be here if `num_outputs_recovered>0`?
```rust
            match self.get_next_peer() {
                Some(peer) => match self.attempt_sync(peer.clone()).await {
                    Ok((num_outputs_recovered, final_height, final_amount, elapsed)) => {
                        debug!(target: LOG_TARGET, "Scanned to height #{}", final_height);
                        self.finalize(num_outputs_recovered, final_height, final_amount, elapsed)?;
                        return Ok(());
                    },
                    Err(e) => {
                        warn!(
                            target: LOG_TARGET,
                            "Failed to scan UTXO's from base node {}: {}", peer, e
                        );
                        self.publish_event(UtxoScannerEvent::ScanningRoundFailed {
                            num_retries: self.num_retries,
                            retry_limit: self.retry_limit,
                            error: e.to_string(),
                        });
                        continue;
                    },
                },
```

I also see there are some failing unit tests.

hansieodendaal (approved):
utACK

PR #6117, opened 2024-02-02
fix: make monero extra data less strict

SWvheerden:
Description
---
It reverts a change to make the extra field acceptance less strict. 
Motivation and Context
---
Tari only cares about the Merge mining field and not the rest. As long as that deserializes correct we are secure.

### DIFF
--- a/base_layer/core/src/proof_of_work/monero_rx/helpers.rs
+++ b/base_layer/core/src/proof_of_work/monero_rx/helpers.rs
@@ -105,8 +105,11 @@ pub fn verify_header(
 ) -> Result<MoneroPowData, MergeMineError> {
     let monero_data = MoneroPowData::from_header(header, consensus)?;
     let expected_merge_mining_hash = header.merge_mining_hash();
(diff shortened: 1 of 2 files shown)


base_layer/core/src/proof_of_work/monero_rx/helpers.rs
```diff
@@ -105,8 +105,11 @@ pub fn verify_header(
 ) -> Result<MoneroPowData, MergeMineError> {
     let monero_data = MoneroPowData::from_header(header, consensus)?;
     let expected_merge_mining_hash = header.merge_mining_hash();
-    let extra_field = ExtraField::try_parse(&monero_data.coinbase_tx_extra)
-        .map_err(|_| MergeMineError::DeserializeError("Invalid extra field".to_string()))?;
+    let extra_field = ExtraField::try_parse(&monero_data.coinbase_tx_extra);
```
stringhandler, line 108: Is there no other data about the error here? 
Maybe create an issue to add more data about what went wrong

stringhandler (dismissed):
Happy for this to go in

PR #6092, opened 2024-01-22
fix: faster tor startup

SWvheerden:
Description
---
This makes so that the nodes/wallets don't have to wait for tor to startup

Motivation and Context
---
Mobile wallets can take very long to wait for tor on startup effectivity blocking the wallet from even doing background tasks. This allows the wallets to start without an active connection

How Has This Been Tested?
---
manual, unit tests

PR #6124, opened 2024-02-06
fix(comms): correctly initialize hidden service

sdbondi:
Description
---
Correctly sets up a tor hidden service in the HiddenServiceTransport
Tor control client will not error if a value in a key-value query is empty, as this can be valid
Disables signal handlers for libtor

Motivation and Context
---

This PR correctly initializes a tor hidden service in the HiddenServiceTransport. A task is spawned by `create_hidden_service` that monitors the tor control port connection and automatically tries to reestablish it if it disconnects. I also fix the incorrect setting of the proxied address to be the listening port of the Tari node rather than tor's SOCKS port.

The minor change to the control port client is minor and allows for an empty value to be returned, which is valid when querying key-value pairs. In practice, this never happens in our current usage but I encountered it when debugging and it prevented the real problem from coming through.

Ref PR #6092 

There is an additional existing problem where libtor handles interrupt signals and exits e.g. when pressing ctrl+c in the base node to type a command. I fixed this by [disabling signal handlers](https://github.com/torproject/torspec/blob/8961bb4d83fccb2b987f9899ca83aa430f84ab0c/control-spec.txt#L3946) in libtor.

How Has This Been Tested?
---
Manually

---

Nodes should receive inbound and outbound tor connections.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/core/src/tor/control_client/client.rs
+++ b/comms/core/src/tor/control_client/client.rs
@@ -125,9 +125,6 @@ impl TorControlPortClient {
     pub async fn get_info(&mut self, key_name: &'static str) -> Result<Vec<Cow<'_, str>>, TorClientError> {
         let command = commands::get_info(key_name);
         let response = self.request_response(command).await?;
-        if response.is_empty() {
-            return Err(TorClientError::ServerNoResponse);
-        }
         Ok(response)
     }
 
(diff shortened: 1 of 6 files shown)


SWvheerden (approved):
LGTM

PR #6125, opened 2024-02-06
fix(libtor): prevent metrics port conflict

sdbondi:
Description
---
Prevent conflict with metrics port when starting multiple nodes using libtor
Run tor in background without an unnecessary task spawn
Write the control port to a file (not used in the node implementation)
Support for tor auto assigning it's own ports

Motivation and Context
---

Set metrics port to 0 (os -assigned) so that multiple libtor instances do not conflict
Previously we'd wait on a shutdown signal in an async function, but tor runs in the background in an OS thread and waiting on a shutdown does nothing other than keep a task open in memory.

We could experience issues binding and dropping a listener quickly on some platforms. It would be better for tor to assign its own port however we could implement that if this is ever an issue, until then this code just makes us aware of that option.

How Has This Been Tested?
---

Running two base nodes and a wallet with libtor enabled on each

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5987,7 +5987,6 @@ dependencies = [
  "rand",
  "tari_common",
  "tari_p2p",
- "tari_shutdown",
  "tempfile",
  "tor-hash-passwd",
 ]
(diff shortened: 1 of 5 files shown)


SWvheerden (commented):
lgtm

SWvheerden (approved):

Issue #6122, opened 2024-02-03
Mainnet uses inflating tail emission


CjS77:
# The Minotari emission schedule

The Minotari emission schedule was designed to achieve multiple goals:

* Offer sufficient early liquidity, so that builders of digital assets have equitable access to Tari tokens. This is primarily achieved by implementing an exponential decay function into the primary emission curve.
* Provide support for the continued development and maintenance of the Tari ecosystem. This is achieved via the primary pre-mine.
* Ensure that there are enough tokens to fuel the ecosystem for perpetuity. This is achieved through the tail emission.

Coins that have a fixed supply with no tail emission, such as Bitcoin, are strictly _deflationary_. 

On the other hand, Monero has a constant tail emission. Its total supply is technically infinite, but its emission is _disinflationary_, since the annual inflation rate of the Monero supply is continuously approaching zero. The tail emission is broadly in place to account for coins that are lost due to the fact that people are involved in the system, and people tend to lose and/or break things.

Bitcoin and Monero have one job: To be money.

But Minotari has a very important second role: to be converted into the Tari tokens that fuel the Tari Digital Assets Network.

The word _fuel_ is not just metaphorical. A small portion of the transaction fee in every single one of Tari’s second layer instructions is _burnt_. This creates a small but constant driving force to convert Minotari into Tari. This simple and elegant design is called the [turbine model](https://rfc.tari.com/RFC-0320_TurbineModel.html). It maintains a soft 1:1 peg between Tari and Minotari without having to resort to complicated peg-out mechanisms like drive chains, or federated pegs like Liquid. In fact, there’s no peg-out at all. This offers the tremendous benefit of allowing the Tari network to operate at its native speed (fast!).

Because of the turbine model, in the absence of a tail emission, we would eventually run out of Tari. But even a disinflationary token may experience liquidity problems in the long run if the second layer experiences sustained growth.

Imagine if the Tari tail emission was a constant 800 XTR per block. If the Tari network was burning more than that every two minutes due to sustained high activity, then the supply of Tari globally would fall.

Therefore, Tari’s tail emission is mildly inflationary in nominal terms. Real GDP growth typically sits somewhere between 1% and 3% per annum. The genesis block has encoded parameters that initiate a tail emission inflation rate at the low end of this range.
[...]

PR #6131, opened 2024-02-06
feat: inflating tail

CjS77:
## Summary

Adds a feature `tari_feature_mainnet_emission` to allow for tail emission
inflation. See #6122 

This change necessitates the addition of 2 new consensus constants:
`inflation_bips` -- the annual inflation rate of the total supply.
and `tail_emission_epoch_length`, which controls the tail emission
inflation. These replace `tail_emission`.

We update the Protobuf definition for `ConsensusConstants` to account for
the new fields.

## Getting Feature flags working

When formally implementing compiler features, the setting of the `TARI_NETWORK`
and (from now) `TARI_TARGET_NETWORK` envar and the related compiler flags matter.

These changes require that when tests run, the correct network flags are set.

For example, `blocks::genesis_block::test::stagenet_genesis_sanity_check` should
only run when `TARI_NETWORK=stagenet`.

While writing this PR, I uncovered a problem:
There is not a 1:1 mapping between the build _target_ and the actual
network the binary may run on. E.g. the mainnet build can run on either
stagenet or mainnet.

Unfortunately the `TARI_NETWORK` envar was used to set both of these
variables. Without using feature flags, this discrepancy went unnoticed.
With feature flags, a lot of tests that implicitly assumed say a mainnet
target and a stagenet run would simply fail, with no way to resolve the
issue.

Since these are 2 different things, this commit introduced
`TARI_TARGET_NETWORK` which sets the **build target**. In most contexts,
this then decouples the configured network, which is specified by the
`TARI_NETWORK` envar  as usual.

The other changes in this commit revolve around updating merkle roots
and the correct faucet sets to make the target/network/faucet
combination consistent.
## CI workflows

The CI is configured to run tests for each of `{target} ({network})`
* Testnet (Esme)
* Nextnet (Nextnet)
* Mainnet (Stagenet)


applications/minotari_merge_mining_proxy/build.rs
```diff
@@ -5,7 +5,8 @@ use tari_features::resolver::build_features;
 
 #[cfg(windows)]
 fn main() {
-    build_features();
+    env_logger::init();
+    ();
```
SWvheerden: is this correct?

applications/minotari_console_wallet/build.rs
```diff
@@ -5,7 +5,9 @@ use tari_features::resolver::build_features;
 
 #[cfg(windows)]
 fn main() {
+    env_logger::init();
     build_features();
+    ();
```
SWvheerden, line 9: is this correct?

applications/minotari_node/build.rs
```diff
@@ -5,7 +5,9 @@ use tari_features::resolver::build_features;
 
 #[cfg(windows)]
 fn main() {
+    env_logger::init();
     build_features();
+    ();
```
SWvheerden: same here

base_layer/core/src/validation/chain_balance.rs
```diff
@@ -92,8 +92,14 @@ impl<B: BlockchainBackend> ChainBalanceValidator<B> {
     }
 
     fn get_emission_commitment_at(&self, height: u64) -> Commitment {
-        let total_supply =
-            self.rules.get_total_emission_at(height) + self.rules.consensus_constants(height).faucet_value();
+        // With inflating tail emission, we **must** know the value of the premine as part of the supply calc in order
+        // to determine the correct inflation curve. So we need to include the premine in the emission curve.
+        let total_supply = self.rules.get_total_emission_at(height) +
+            if cfg!(tari_feature_mainnet_emission) {
+                MicroMinotari::from(0)
```
SWvheerden, line 99: this 0 does not seem correct. 
Does mainnet emission not have a faucet value?

SWvheerden (commented):
I am not  fan of adding this feature: `tari_feature_mainnet_emission`
It complicates the code and makes a lot of the code duplicate. 
Why not let all run on the new tail emission ?

base_layer/core/src/validation/chain_balance.rs
CjS77: Read the comments

CjS77:
> I am not fan of adding this feature: `tari_feature_mainnet_emission` It complicates the code and makes a lot of the code duplicate. Why not let all run on the new tail emission ?

If you read the PR description, I explain the need and rationale for testing the feature gate before mainnet. This is as good a test as any.

applications/minotari_merge_mining_proxy/build.rs
CjS77: Copy pasta snafu. They've been removed

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -105,23 +105,23 @@ fn print_mr_values(block: &mut Block, print: bool) {
 
 pub fn get_stagenet_genesis_block() -> ChainBlock {
     let mut block = get_stagenet_genesis_block_raw();
-
+    eprintln!("Network: {:?}", Network::get_current_or_default());
```
stringhandler, line 108: missed?

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -891,181 +946,6 @@ mod test {
         ConsensusConstants::mainnet();
     }
 
-    // Comment out the feature flag to run this test
-    #[test]
-    #[cfg(feature = "schedule_get_constants")]
-    fn esmeralda_schedule_get_constants() {
```
hansieodendaal, line 897: Why is this test removed? It should be added back in.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -891,181 +946,6 @@ mod test {
...
-                println!("Total supply up to tail emission: {}", total_supply_up_to_tail_emission);
-                break;
-            }
-            esmeralda[0].emission_initial = esmeralda[0].emission_initial + MicroMinotari(1);
-        }
-        panic!("\n\nThis test may not pass in CI\n\n");
-    }
-
-    #[test]
-    fn esmeralda_schedule() {
```
hansieodendaal, line 938: Why is this test removed? It should be added back in.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -891,181 +946,6 @@ mod test {
...
-        assert_eq!(block_num, 3255553 + coinbase_offset);
-        assert_eq!(reward, 800_000_415 * uT);
-        let total_supply_up_to_tail_emission = supply + esmeralda[0].faucet_value;
-        assert_eq!(total_supply_up_to_tail_emission, 20_999_999_999_819_869 * uT);
-        let (_, reward, _) = rewards.next().unwrap();
-        assert_eq!(reward, esmeralda[0].emission_tail);
-    }
-
-    #[test]
-    fn nextnet_schedule() {
```
hansieodendaal, line 973: Why is this test removed? It should be added back in.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -891,181 +946,6 @@ mod test {
...
-        assert_eq!(block_num, 3255553 + coinbase_offset);
-        assert_eq!(reward, 800_000_415 * uT);
-        let total_supply_up_to_tail_emission = supply + nextnet[0].faucet_value;
-        assert_eq!(total_supply_up_to_tail_emission, 20_999_999_999_819_869 * uT);
-        let (_, reward, _) = rewards.next().unwrap();
-        assert_eq!(reward, nextnet[0].emission_tail);
-    }
-
-    #[test]
-    fn stagenet_schedule() {
```
hansieodendaal, line 1008: Why is this test removed? It should be added back in.

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -891,181 +946,6 @@ mod test {
...
-        assert_eq!(block_num, 3255553 + coinbase_offset);
-        assert_eq!(reward, 800_000_415 * uT);
-        let total_supply_up_to_tail_emission = supply + stagenet[0].faucet_value;
-        assert_eq!(total_supply_up_to_tail_emission, 20_999_999_999_819_869 * uT);
-        let (_, reward, _) = rewards.next().unwrap();
-        assert_eq!(reward, stagenet[0].emission_tail);
-    }
-
-    #[test]
-    fn igor_schedule() {
```
hansieodendaal, line 1043: Why is this test removed?

base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -78,7 +82,14 @@ pub struct ConsensusConstants {
     /// would be 1/4
     pub(in crate::consensus) emission_decay: &'static [u64],
     /// This is the emission curve tail amount
+    #[cfg(not(tari_feature_mainnet_emission))]
     pub(in crate::consensus) emission_tail: MicroMinotari,
+    /// The tail emission inflation rate in bips
```
hansieodendaal, line 87:
```suggestion
    /// The tail emission inflation rate in bips (basis points, otherwise known as bps or "bips")
```

.github/workflows/ci.yml
```diff
@@ -167,6 +167,17 @@ jobs:
     permissions:
       checks: write
       pull-requests: write
+    strategy:
+      matrix:
+        tari_target_network: [
```
hansieodendaal, line 172: Maybe a better name choice
```suggestion
        tari_network_map: [
```

applications/minotari_app_grpc/src/conversions/consensus_constants.rs
```diff
@@ -29,7 +29,10 @@ use crate::tari_rpc as grpc;
 impl From<ConsensusConstants> for grpc::ConsensusConstants {
     #[allow(clippy::too_many_lines)]
     fn from(cc: ConsensusConstants) -> Self {
+        #[cfg(not(tari_feature_mainnet_emission))]
         let (emission_initial, emission_decay, emission_tail) = cc.emission_amounts();
+        #[cfg(tari_feature_mainnet_emission)]
```
hansieodendaal, line 34: Why have this differentiation at all? It seems more prudent to choose if we will have inflating tail emission or not.

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -399,11 +393,13 @@ mod test {
     }
 
     #[test]
-    fn esmeralda_genesis_sanity_check() {
```
hansieodendaal, line 402: This esemeralda genesis block sanity test should not be removed, it should be added back in. We need to do the genesis block sanity check for every network.

hansieodendaal (commented):
I can appreciate that a lot was accomplished in this PR.

However, I think it would be better to have a tail emission and then inflate the tail emission if inflating tail emission is required. When inflating tail emission is not desired that value can be zero in the consensus constants. This would make the code much simpler and more consistent and remove the requirement to have the `tari_feature_mainnet_emission` feature flag.

base_layer/core/src/consensus/consensus_constants.rs
CjS77: It's covered by the new tail emission test. Esme's constants are changing with the tail inflation, so the test would fail anyway.

base_layer/core/src/consensus/consensus_constants.rs
CjS77: Same reason.

base_layer/core/src/consensus/consensus_constants.rs
CjS77: Same reason

applications/minotari_app_grpc/src/conversions/consensus_constants.rs
CjS77: I'm not sure I understand the question.

If you're asking why inflating tail emission is using a feature gate, I gave a rationale in other comments.

CjS77:
> I can appreciate that a lot was accomplished in this PR.
> 
> However, I think it would be better to have a tail emission and then inflate the tail emission if inflating tail emission is required. When inflating tail emission is not desired that value can be zero in the consensus constants. This would make the code much simpler and more consistent and remove the requirement to have the `tari_feature_mainnet_emission` feature flag.

I'm happy to hear alternatives.

What you are suggesting is not 100% clear. Are you suggesting inflating tail emission via hardforks "as we go" every year?

If not, and you're proposing a transparent, code-based approach as presented here, then please offer an alternative. Note that your constraints are:
* Inflation rate is not known _a priori_.
* The inflation rate _must_ be based on total supply, including faucets/premine
* Tail emission _must_ kick in the very block that the inflation rate falls below the target inflation rate.

applications/minotari_app_grpc/src/conversions/consensus_constants.rs
hansieodendaal: Why have both implementations in the code if we will only use one?

hansieodendaal:
What I mean is we can implement both schemes without feature flags and without duplicate code.

In the event that we want to have tail emission with tail inflation:
```rust
            emission_initial: INITIAL_EMISSION,
            emission_decay: &EMISSION_DECAY,
            emission_tail: 800 * T,
            inflation_bips: 1,
            tail_epoch_length: ANNUAL_BLOCKS,
```
In the event that we want to have tail inflation only:
```rust
            emission_initial: INITIAL_EMISSION,
            emission_decay: &EMISSION_DECAY,
            emission_tail: 0 * T,
            inflation_bips: 1,
            tail_epoch_length: ANNUAL_BLOCKS,
```
In the event that we want to have tail emission only:
```rust
            emission_initial: INITIAL_EMISSION,
            emission_decay: &EMISSION_DECAY,
            emission_tail: 800 * T,
            inflation_bips: 0,
            tail_epoch_length: 0,
```
In the event that we do not want either:
```rust
            emission_initial: INITIAL_EMISSION,
            emission_decay: &EMISSION_DECAY,
            emission_tail: 0 * T,
            inflation_bips: 0,
            tail_epoch_length: 0,
```

The `fn next_reward` function can change to this:
```rust
    fn next_reward(&mut self) {
        // Inflation phase
        if self.epoch > 0 && self.schedule.inflation_bips > 0 {
            self.epoch_counter += 1;
            if self.epoch_counter >= self.schedule.epoch_length {
                self.epoch_counter = 0;
                self.epoch += 1;
                self.reward = self.new_tail_emission();
            }
        } else {
            // Decay phase
            let cutoff = if self.schedule.tail > MicroMinotari::zero() {
                self.schedule.tail
            } else {
                self.new_tail_emission()
            };
            let next_decay_reward = self.next_decay_reward();
            if self.epoch == 0 && next_decay_reward > cutoff {
                self.reward = next_decay_reward;
            } else {
                self.epoch = 1;
                self.reward = cutoff;
            }
        }
    }
```

CjS77:
Replaced by #6160 and #6162

PR #6132, opened 2024-02-06
feat: add import tx method

SWvheerden:
Description
---
Allows users to import an exported transaction to their wallet. 

Motivation and Context
---
This allows true cold mining. A wallet can export any transaction, and it can then be imported to any transaction which will monitor and ensure the transaction is broadcasted to the network. 

How Has This Been Tested?
---
unit tests and manual

### DIFF
--- a/applications/minotari_console_wallet/src/wallet_modes.rs
+++ b/applications/minotari_console_wallet/src/wallet_modes.rs
@@ -475,6 +475,7 @@ async fn run_grpc(
 
 #[cfg(test)]
 mod test {
+    use std::path::Path;
 
     use crate::{cli::CliCommands, wallet_modes::parse_command_file};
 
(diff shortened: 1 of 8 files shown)


hansieodendaal (approved):
ACK

applications/minotari_console_wallet/src/wallet_modes.rs
```diff
@@ -501,6 +502,8 @@ mod test {
 
             export-tx 123456789 --output-file pie.txt
 
+            import-tx --input-file pie_this_message.txt
```
brianp, line 505: What's the significance of `pie`

brianp (approved):
Will a file only ever represent a single tx. Would it be trivial to expand the code to support a file with multiple tx's?

applications/minotari_console_wallet/src/wallet_modes.rs
SWvheerden: No idea...
Its a random word I use

SWvheerden:
So the export code only currently supports exporting a single tx. 
But the import of transactions can be multiple. 
Look the `load_tx_from_csv_file` function, it steps through the json file and tries to find all transactions it can

PR #6133, opened 2024-02-06
ci: bump nightly version

AaronFeickert:
Description
---
Consistent with other repositories in the ecosystem, this PR bumps the nightly toolchain to `nightly-2024-02-01`. It addresses many updated lints.

Motivation and Context
---
Other repositories are pinning their nightly toolchains at `nightly-2024-02-01` due to a curve library [issue](https://github.com/dalek-cryptography/curve25519-dalek/issues/618). It seems like a good time to bump the nightly version here as well.

How Has This Been Tested?
---
Tested via CI.

---
Ensure that CI and any other relevant build processes succeed.

### DIFF
--- a/.github/workflows/base_node_binaries.json
+++ b/.github/workflows/base_node_binaries.json
@@ -2,7 +2,7 @@
   {
     "name": "linux-x86_64",
     "runs-on": "ubuntu-20.04",
-    "rust": "nightly-2023-06-04",
+    "rust": "nightly-2024-02-01",
     "target": "x86_64-unknown-linux-gnu",
     "cross": false,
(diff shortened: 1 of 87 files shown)


AaronFeickert:
There are two warning lints that identify `Arc`s whose underlying types are not `Send + Sync`. It's unclear if it's safe to ignore these.

AaronFeickert:
After discussion, it sounds like while an eventual update to the nightly toolchain may be useful to avoid ongoing technical debt (and in case of a [future curve library update](https://github.com/tari-project/tari/issues/6138)), it isn't particularly crucial at this time.

Closing.

PR #6135, opened 2024-02-07
chore: fix versions to correct version

SWvheerden:
Description
---
Some libraries have still unused old versions. 
This fixes the versions of all libraries to use the correct new versions.


hansieodendaal (approved):
utACK

PR #6140, opened 2024-02-08
feat: allow ffi to see lock height

SWvheerden:
Description
---
Allows the FFI to see the lockheight of utxos

Motivation and Context
---
This will allow the FFI to not select coinbases that are still time locked.


hansieodendaal (approved):
utACK

PR #6144, opened 2024-02-12
fix: balanced binary merkle tree merged proof

Cifko:
Description
---
We now check if all the paths are correct. If they are not left dangling, if they merge into one and that computes the root hash.

How Has This Been Tested?
---
`cargo test --package tari_mmr --lib -- balanced_binary_merkle_proof::test --nocapture`

---
Same test, but it would be nice to confirm my approach as well.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (dismissed):

SWvheerden (dismissed):
Just need green CI

PR #6146, opened 2024-02-13
feat: check chain metadata

SWvheerden:
Description
---
Checks chain metadata for correctness

Motivation and Context
---
The base node makes certain logic desisions based on what the chain meta data is, but no where do we make checks to ensure that the data is valid inside of this. 

How Has This Been Tested?
---
unit tests

### DIFF
--- a/base_layer/common_types/src/chain_metadata.rs
+++ b/base_layer/common_types/src/chain_metadata.rs
@@ -47,6 +47,13 @@ pub struct ChainMetadata {
     /// Timestamp of the tip block in the longest valid chain
     timestamp: u64,
 }
+#[derive(Debug, thiserror::Error)]
+pub enum ChainMetaDataError {
+    #[error("Pruning Height is higher than the Best Block height")]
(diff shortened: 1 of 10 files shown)


brianp (approved):
utAck

base_layer/common_types/src/chain_metadata.rs
```diff
@@ -56,15 +63,22 @@ impl ChainMetadata {
...
             best_block_hash,
             pruning_horizon,
             pruned_height,
             accumulated_difficulty,
             timestamp,
-        }
+        };
+        if chain_meta_data.accumulated_difficulty == 0.into() {
+            return Err(ChainMetaDataError::AccumulatedDifficultyZero);
+        };
```
hansieodendaal, line 77: We could check `U256::MAX` as well?

SWvheerden: what do you want to check, u256::max is a valid difficulty, 0 is not

PR #6151, opened 2024-02-16
fix: wallet clear short term output

SWvheerden:
Description
---
This will clear short term encumbered outputs on startup

Motivation and Context
---
Its possible that the wallet if stopped at the right(wrong) moment it can leave encumbered outputs locked forever. 
This will cause the wallet to reset these outputs on startup, as they should only be encumbered for a short while. They will then be passed on to the validation task to revalidate them.

How Has This Been Tested?
---
manual

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -171,6 +171,8 @@ where
         let mut base_node_service_event_stream = self.base_node_service.get_event_stream();
 
         debug!(target: LOG_TARGET, "Output Manager Service started");
+        // Outputs marked as shorttermencumbered are not yet stored as transactions in the TMS, so lets clear them
+        self.resources.db.clear_short_term_encumberances()?;
         loop {
             tokio::select! {
                 event = base_node_service_event_stream.recv() => {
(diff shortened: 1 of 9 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -171,6 +171,8 @@ where
         let mut base_node_service_event_stream = self.base_node_service.get_event_stream();
 
         debug!(target: LOG_TARGET, "Output Manager Service started");
+        // Outputs marked as shorttermencumbered are not yet stored as transactions in the TMS, so lets clear them
+        self.resources.db.clear_short_term_encumberances()?;
```
hansieodendaal, line 175: It feels like this could create problems, for example, we create an interactive transaction, send the initial request to the receiver and quit immediately afterwards. Then we start up again and clear all those encumbered outputs, make a new transaction and use them again, and so on.

SWvheerden: as soon as we send them out, we flag them as encumbered, not short_term_encumbered. 

Even if we did not do that, the wallet would then clear them out, but it will never create a finalised transaction that can be broadcasted. 


On the otherside, if we don't clear out these, these funds will be locked forever, and your wallet wont be able to use them till you delete and recover your wallet

hansieodendaal: Perfect

PR #6152, opened 2024-02-16
feat: wallet ffi use dns

SWvheerden:
Description
---
Add the option for the wallet_ffi, to use the DNS build into it. 

Motivation and Context
---
Allow the fii to use DNS. 

How Has This Been Tested?
---
Manual


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -6365,6 +6379,45 @@ pub unsafe extern "C" fn wallet_set_base_node_peer(
...
+/// `error_out` - Pointer to an int which will be modified to an error code should one occur, may not be null. Functions
+/// as an out parameter.
+///
+/// ## Returns
+/// `TariPublicKeys` - Returns a list of all known public keys
+///
+/// # Safety
+/// None
+#[no_mangle]
+pub unsafe extern "C" fn wallet_get_seed_peers(wallet: *mut TariWallet, error_out: *mut c_int) -> *mut TariPublicKeys {
```
brianp, line 6391: What's the purpose of the function? So they can be listed as connectable base nodes?

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5259,6 +5259,7 @@ pub unsafe extern "C" fn wallet_create(
     passphrase: *const c_char,
     seed_words: *const TariSeedWords,
     network_str: *const c_char,
+    peer_seed_str: *const c_char,
```
brianp, line 5262: I don't think we need a whole parameter for this. It's based on the network, and we already have that:

```
dns_seeds: StringList::from(vec![format!("seeds.{}.tari.com", network.as_key_str())])
```

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5473,6 +5486,7 @@ pub unsafe extern "C" fn wallet_create(
     let peer_seeds = PeerSeedsConfig {
         dns_seeds_name_server: DEFAULT_DNS_NAME_SERVER.parse().unwrap(),
         dns_seeds_use_dnssec: true,
```
brianp: From my chat based experience- dnssec needs to be turned off or all our DNS validations will fail. 

This is the chat config for PeerSeeds which *does* use DNS (including the existing integrated wallet branch)

```
        peer_seeds: PeerSeedsConfig {
            dns_seeds_use_dnssec: false,
            dns_seeds_name_server: DEFAULT_DNS_NAME_SERVER.parse().unwrap(),
            dns_seeds: StringList::from(vec![format!("seeds.{}.tari.com", network.as_key_str())]),
            ..PeerSeedsConfig::default()
        },
        ```

brianp (dismissed):
Miner change could reduce overall changes and interface if we interpolate the DNS address based on the network instead of passing it in.

Also a note about dnssec, but if it's been tested #yolo.

base_layer/wallet_ffi/src/lib.rs
SWvheerden: yeah

base_layer/wallet_ffi/src/lib.rs
SWvheerden: thats only if we dont change it...

base_layer/wallet_ffi/src/lib.rs
SWvheerden: yip mapped that through

brianp (approved):

PR #6153, opened 2024-02-16
feat: change CLI  get_block to search orphans

SWvheerden:
Description
---
Allows the get_block command, to search for the block in the orphan db as well. 

Motivation and Context
---
From the CI we need to be able to see how reorged blocks look.


hansieodendaal (approved):
utACK

_**Note:** Pity we could not re-use the code from `NodeCommsRequest::GetBlockFromAllChains(hash)` easily_

Issue #6108, opened 2024-01-30
Completed too large transaction not being rejected but re-tried all the time


hansieodendaal:
- Wallets should not be able to negotiate and finalize transactions that will be rejected when broadcast due to size
- In the event that the above happens, those transactions should be rejected with reason, feedback to the user

```
┌Transaction Details────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│TxID:                   415804858639007959                                                                                         │
│Source Address:         🍞🍎🎂🌹🐺🎬🌻📈🎼🍰🚁🏆🐽🐭🐉👖💕🌠🐮🐴🌷💦🎹🏠🎣🐷🍪💺📣🎣🍗🐊🐑                                         │
│Destination address:    🍁👕🛁💄🚨🌴🎪🍀🏦😹🏦🌍💭🐜🎭🚧🎲🐴🔬🎸🚽🐚📈🍬🐉🐷🐭🔦👟💌👗🌽🎺                                         │
│Direction:              Outbound                                                                                                   │
│Amount:                 10.000000 T                                                                                                │
│Fee:                    65932 µT (weight: 65932g, #inputs: 8226, #outputs: 2)                                                      │
│Status:                 Completed                                                                                                  │
│Message:                                                                                                                           │
│Mined At (Local):                                                                                                                  │
│Imported At (Local):    2024-01-30 07:27:01                                                                                        │
│Excess:                 c55924af988864e6e961f6e4c3324b2a359260c1b6a919dc76d753aad565c800                                           │
│Confirmations:          N/A                                                                                                        │
│Mined Height:           N/A                                                                                                        │
│Maturity:               N/A                                                                                                        │
```

Broadcast protocol - submit
```rust
2024-01-30 07:27:03.094744900 [wallet::transaction_service::protocols::broadcast_protocol] INFO  Submitting Transaction (TxId: 415804858639007959) with signature 'c55924af988864e6e961f6e4c3324b2a359260c1b6a919dc76d753aad565c800' to Base Node
2024-01-30 07:27:03.197487500 [wallet::transaction_service::protocols::broadcast_protocol] INFO  Submitting Transaction (TxId: 415804858639007959) with signature 'c55924af988864e6e961f6e4c3324b2a359260c1b6a919dc76d753aad565c800' to Base Node
[...]

PR #6154, opened 2024-02-18
feat: limit transaction size

hansieodendaal:
Description
---
Added checks to ensure that transactions negotiated between parties, transactions to self and imported transactions do not exceed the RPC frame size limit when the transaction is broadcast to the base node. This can easily happen when lots of dust inputs are collected to be spent. If a too-large transaction is detected in any of the transaction service protocols it will be cancelled so that the user can try to create a new transaction with different parameters, for example, to reduce the recipient amount.

**Edit:**
- Added a wallet sqlite memory type connection to speed up tests with large amounts of db activity
- Used fake outputs without valid bulletproof range proofs in the `spend_dust` tests to speed it up even more as suggested by @SWvheerden 

Motivation and Context
---
See #6108

How Has This Been Tested?
---
Added unit tests:
- `test_spend_dust_to_self_in_oversized_transaction`
- `test_spend_dust_to_other_in_oversized_transaction`
- `test_spend_dust_happy_path`

---
Code walk-through
Review unit tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
+++ b/base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
@@ -63,6 +63,7 @@ use crate::{
         config::TransactionRoutingMechanism,
         error::{TransactionServiceError, TransactionServiceProtocolError},
         handle::{TransactionEvent, TransactionSendStatus, TransactionServiceResponse},
+        protocols::check_transaction_size,
         service::{TransactionSendResult, TransactionServiceResources},
         storage::{
             database::TransactionBackend,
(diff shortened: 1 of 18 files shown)


base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
```diff
@@ -281,17 +282,44 @@ where
...
+
         // Attempt to send the initial transaction
         let SendResult {
             direct_send_result,
             store_and_forward_send_result,
             transaction_status,
-        } = match self.send_transaction(msg).await {
-            Ok(val) => val,
+        } = match check_transaction_size(&outbound_tx_check, self.id) {
+            Ok(_) => match self.send_transaction(msg).await {
```
SWvheerden: Nit: Wont it be easier to read by just checking the size outize of this massive match?
Something akin to 
```
if let Err(e) = check_transaction_size(&outbound_tx_check, self.id){
handle error and exit
}
```

SWvheerden (dismissed):

base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
hansieodendaal: We need to return a `SendResult ` if the transaction is too large, similar to when the actual send has an issue, so it is combined with the other match.
```rust
                SendResult {
                    direct_send_result: false,
                    store_and_forward_send_result: false,
                    transaction_status: TransactionStatus::Queued,
                }
```

SWvheerden (approved):

PR #6156, opened 2024-02-19
feat!: change proof of work to be dependant on target difficulty

SWvheerden:
Description
---
Changes Tari to use Target difficulty rather than achieved difficulty when looking at the chain with the most proof of work done.

Motivation and Context
---
Achieved difficulty can make it so that you get a lot of reorgs based on blocks getting a lucky high achieved difficulty. This skews the results in favour of a lucky block and can cause that lucky block to easily reorg out many other blocks. It also allows an attacker to play games with the block after they have found such a lucky block. 

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/minotari_app_grpc/src/conversions/chain_metadata.rs
```diff
@@ -27,12 +27,12 @@ use crate::tari_rpc as grpc;
 impl From<ChainMetadata> for grpc::MetaData {
     fn from(meta: ChainMetadata) -> Self {
         let mut diff = [0u8; 32];
-        meta.accumulated_difficulty().to_big_endian(&mut diff);
+        meta.accumulated_target_difficulty().to_big_endian(&mut diff);
```
stringhandler: We can keep the name `accumulated_difficulty`

stringhandler (approved):

AaronFeickert:
Interestingly, a recent [preprint](https://eprint.iacr.org/2024/200) seems to come to the opposite conclusion, that computing chain weight using a form of achieved difficulty provides better performance without sacrificing security.

PR #6158, opened 2024-02-20
fix!: changes balance query

SWvheerden:
Description
---
changes balance query to be more in line with you would expect. 
FFI now matches grpc

BREAKING CHANGE
---
This changes the external API for the wallet_ffi and the wallet library


hansieodendaal (approved):
utACK

PR #6159, opened 2024-02-20
fix: hide unmined coinbase

SWvheerden:
Description
---
This reintroduced code to hide unmined coinbase transactions.

Motivation and Context
---
Will show and hide depending on the user preference all coinbases that are reorged out and not in the main chain anymore.


brianp (approved):
utAck

PR #6160, opened 2024-02-20
feat!: implement inflating tail emission

CjS77:
## Summary

Adds a feature `tari_feature_mainnet_emission` to allow for tail emission
inflation. See #6122 

This change necessitates the addition of 2 new consensus constants:
`inflation_bips` -- the annual inflation rate of the total supply.
and `tail_emission_epoch_length`, which controls the tail emission
inflation. These replace `tail_emission`.

We update the Protobuf definition for `ConsensusConstants` to account for
the new fields.

Note: Replaces part of #6131

### DIFF
--- a/applications/minotari_app_grpc/proto/types.proto
+++ b/applications/minotari_app_grpc/proto/types.proto
@@ -119,7 +119,7 @@ message ConsensusConstants {
     uint64 median_timestamp_count = 9;
     uint64 emission_initial = 10;
     repeated uint64 emission_decay = 11;
-    uint64 emission_tail = 12;
+    uint64 emission_tail = 12 [deprecated=true];
     uint64 min_sha3x_pow_difficulty = 13;
     uint64 block_weight_inputs = 14;
     uint64 block_weight_outputs = 15;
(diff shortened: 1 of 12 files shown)


applications/minotari_app_grpc/proto/types.proto
```diff
@@ -119,7 +119,7 @@ message ConsensusConstants {
     uint64 median_timestamp_count = 9;
     uint64 emission_initial = 10;
     repeated uint64 emission_decay = 11;
-    uint64 emission_tail = 12;
+    uint64 emission_tail = 12 [deprecated=true];
```
SWvheerden, line 122: Nit:
I think we can just delete this, it should default to 0 if you ask for it in a dated proto file. 
But I am 99% nothing uses this field

SWvheerden (dismissed):
This code reads and follows so much easier without the feature flag in here.

applications/minotari_app_grpc/proto/types.proto
CjS77: Let's nuke it in a separate PR so that it's easy to revert if you're wrong.

SWvheerden (approved):

PR #6161, opened 2024-02-21
fix: oms validation

SWvheerden:
Description
---
Fixes the OMS validation to invalidate unmined utxos and not keep them pending

Motivation and Context
---
Although these utxo would never reach spending, the wallet sees them as pending incoming or pending outgoing. We need to ensure their status is correctly set.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
+++ b/base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
@@ -321,6 +321,13 @@ where
             }
             for unmined_output in unmined {
                 if unmined_output.status == OutputStatus::UnspentMinedUnconfirmed {
+                    info!(
+                        target: LOG_TARGET,
+                            "Updating output comm:{}: hash {} as unmined(Operation ID: {})",
+                        unmined_output.output.commitment.to_hex(),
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
```diff
@@ -318,6 +319,13 @@ where
                 )
                 .await?;
             }
+            for unmined_output in unmined {
+                if unmined_output.status == OutputStatus::UnspentMinedUnconfirmed {
```
hansieodendaal, line 323: Maybe add an info log message here and check the `current height - mined height > some_value` before invalidating?

hansieodendaal (dismissed):
utACK

(I do not have a wallet with `OutputStatus::UnspentMinedUnconfirmed` coinbases to test this  with anymore, but it looks fine)

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
SWvheerden: If we add anything info is wrong here, perhaps trace
Its fixing the state of the db here

PR #6162, opened 2024-02-21
feat: multi-network ci

CjS77:
When formally implementing compiler features, the setting of the `TARI_NETWORK` and (from now) `TARI_TARGET_NETWORK` envar and the related compiler flags matter.

These changes require that when tests run, the correct network flags are set.

For example, `blocks::genesis_block::test::stagenet_genesis_sanity_check` should only run when `TARI_NETWORK=stagenet`.

The current setup reveals a problem:
There is not a 1:1 mapping between the build _target_ and the actual network the binary may run on. E.g. the mainnet build can run on either stagenet or mainnet.

Unfortunately the `TARI_NETWORK` envar was used to set both of these variables.

Since these are 2 different things, this commit introduced `TARI_TARGET_NETWORK` which sets the **build target**. In most contexts, this then decouples the configured network, which is specified by the `TARI_NETWORK` envar  as usual.

The other changes in this commit revolve around updating merkle roots and the correct faucet sets to make the target/network/faucet combination consistent.

The CI is configured to run tests for each of `{target} ({network})`
* Testnet (Esme)
* Nextnet (Nextnet)
* Mainnet (Stagenet)

Description
---

This replaces part of #6131 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -111,7 +111,7 @@ pub fn get_stagenet_genesis_block() -> ChainBlock {
     if add_faucet_utxos {
         // NB! Update 'consensus_constants.rs/pub fn igor()/ConsensusConstants {faucet_value: ?}' with total value
         // NB: `stagenet_genesis_sanity_check` must pass
-        let file_contents = include_str!("faucets/esmeralda_faucet.json");
+        let file_contents = include_str!("faucets/stagenet_faucet.json");
         add_faucet_utxos_to_genesis_block(file_contents, &mut block);
         // Enable print only if you need to generate new Merkle roots, then disable it again
         let print_values = false;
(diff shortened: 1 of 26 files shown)


leet4tari (commented):
Workflow looks good - utACK

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -399,11 +393,13 @@ mod test {
     }
 
     #[test]
-    fn esmeralda_genesis_sanity_check() {
+    #[cfg(tari_target_network_mainnet)]
+    fn stagenet_genesis_sanity_check() {
+        Network::set_current(Network::StageNet).unwrap();
         // Note: Generate new data for `pub fn get_esmeralda_genesis_block()` and `fn get_esmeralda_genesis_block_raw()`
         // if consensus values change, e.g. new faucet or other
```
SWvheerden:
```suggestion
```

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -399,11 +393,13 @@ mod test {
     }
 
     #[test]
-    fn esmeralda_genesis_sanity_check() {
```
SWvheerden, line 402: I dont see a test for esmeralda anymore?
We only have one for nextnet and stagenet

SWvheerden (commented):
This looks good, just one question about the tests, 
I cant seem to find any test for the esme gen block anymore?

SWvheerden (approved):

PR #6163, opened 2024-02-21
chore: removes panics from wallet burn task

SWvheerden:
Description
---
Removes panics from wallet burn task

### DIFF
--- a/applications/minotari_console_wallet/src/ui/state/tasks.rs
+++ b/applications/minotari_console_wallet/src/ui/state/tasks.rs
@@ -257,18 +257,19 @@ pub async fn send_burn_transaction_task(
     // burning minotari
     // ----------------------------------------------------------------------------
 
-    let (burn_tx_id, original_proof) = transaction_service_handle
+    let (burn_tx_id, original_proof) = match transaction_service_handle
         .burn_tari(amount, selection_criteria, fee_per_gram, message, claim_public_key)
         .await
-        .map_err(|err| {
-            log::error!("failed to burn minotari: {:?}", err);
-
+    {
+        Ok((burn_tx_id, original_proof)) => (burn_tx_id, original_proof),
+        Err(e) => {
+            error!(target: LOG_TARGET, "failed to burn minotari: {:?}", e);
             result_tx
-                .send(UiTransactionBurnStatus::Error(UiError::from(err).to_string()))
+                .send(UiTransactionBurnStatus::Error(format!("burn error: {}", e)))
                 .unwrap();
-        })
-        .unwrap();
-
+            return;
+        },
+    };
     // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


brianp (approved):
LGTM
utAck

applications/minotari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -257,18 +257,19 @@ pub async fn send_burn_transaction_task(
...
-            log::error!("failed to burn minotari: {:?}", err);
-
+    {
+        Ok((burn_tx_id, original_proof)) => (burn_tx_id, original_proof),
+        Err(e) => {
+            error!(target: LOG_TARGET, "failed to burn minotari: {:?}", e);
             result_tx
-                .send(UiTransactionBurnStatus::Error(UiError::from(err).to_string()))
+                .send(UiTransactionBurnStatus::Error(format!("burn error: {}", e)))
                 .unwrap();
```
hansieodendaal, line 269: We could get rid of unwrap as well

applications/minotari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -292,14 +293,38 @@ pub async fn send_burn_transaction_task(
...
+                        let serialized_proof = match serde_json::to_string_pretty(&wrapped_proof) {
+                            Ok(proof) => proof,
+                            Err(e) => {
+                                error!(target: LOG_TARGET, "failed to serialize burn proof: {:?}", e);
+                                result_tx
+                                    .send(UiTransactionBurnStatus::Error(format!(
+                                        "failure to create proof {:?}",
+                                        e
+                                    )))
+                                    .unwrap();
```
hansieodendaal, line 305: ditto unwrap

applications/minotari_console_wallet/src/ui/state/tasks.rs
```diff
@@ -292,14 +293,38 @@ pub async fn send_burn_transaction_task(
...
+                        match std::fs::write(filepath, serialized_proof.as_bytes()) {
+                            Ok(()) => {},
+                            Err(e) => {
+                                error!(target: LOG_TARGET, "failed to write burn proof: {:?}", e);
+                                result_tx
+                                    .send(UiTransactionBurnStatus::Error(format!(
+                                        "failure to write proof {:?}",
+                                        e
+                                    )))
+                                    .unwrap();
```
hansieodendaal, line 324: ditto unwrap

applications/minotari_console_wallet/src/ui/state/tasks.rs
SWvheerden: I think these are fine as a panic. 
The send will only fail if there is not listener on the channel. If there is no listener on the channel, there is a problem. There should be a listener on this channel.

PR #6166, opened 2024-02-23
fix: avoid cloning range proofs during verification

AaronFeickert:
Description
---
Removes unnecessary cloning of range proofs during batch verification.

Motivation and Context
---
Currently, range proofs exist in transaction outputs as byte vectors. When parsing outputs to collect these proofs for batch verification, the byte vectors are cloned. This is done because the verification API requires that the vector of proofs contain a vector reference to each proof, but the `BulletRangeProof` type wrapper provides a slice instead as part of its `ByteArray` implementation. Because proofs are several hundred bytes and batches can be large, this cloning is wasteful.

This PR adds `BulletRangeProof::as_vec`, which returns the underlying range proof as a vector reference. Doing so lets us avoid the clone while collecting the range proofs.

How Has This Been Tested?
---
Existing tests pass.

---
Confirm that the change removes the cloning.

### DIFF
--- a/base_layer/common_types/src/types/bullet_rangeproofs.rs
+++ b/base_layer/common_types/src/types/bullet_rangeproofs.rs
@@ -48,6 +48,11 @@ impl BulletRangeProof {
             .expect("This should be 32 bytes for a Blake 256 hash")
             .into()
     }
+
+    /// Get the range proof as a vector reference, which is useful to satisfy the verification API without cloning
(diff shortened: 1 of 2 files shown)


AaronFeickert:
While this approach seems to do the trick, it feels non-idiomatic to me. Is there a better design that wouldn't require the new `as_vec`, like some kind of refactoring?

SWvheerden (approved):

PR #6173, opened 2024-02-28
fix: wallet validation during reorgs

SWvheerden:
Description
---
Fixes wallet validation during reorgs.

Motivation and Context
---
spent outputs that have been reorged, should not be marked as unspent. They should be marked as unspen_mined_unconfirmed
If they are marked as unspent it means the wallet can immeadatly grab the output and use it in a new transaction. The validation service first needs to find out what the status of that output is, allthou the wallet knows the original block the transaction was mined in was reorged out, it needs to figure out if the transaction is still mined, unmined or mempool.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
+++ b/base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
@@ -236,7 +236,7 @@ where
                 if data.height_deleted_at == 0 && output.marked_deleted_at_height.is_some() {
                     // this is mined but not yet spent
                     self.db
-                        .mark_output_as_unspent(output.hash)
+                        .mark_output_as_unspent(output.hash, true)
                         .for_protocol(self.operation_id)?;
                     info!(
                         target: LOG_TARGET,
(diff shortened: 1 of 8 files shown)


base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
```diff
@@ -395,8 +395,10 @@ where
                     last_spent_output.commitment.to_hex(),
                     self.operation_id
                 );
+                // we mark the output as UnspentMinedUnconfirmed so it wont get picked it by the OMS to be spendable
+                // immediately as we first need to find out if this output is unspent, in a mempool, or spent.
                 self.db
-                    .mark_output_as_unspent(last_spent_output.hash)
+                    .mark_output_as_unspent(last_spent_output.hash, false)
```
hansieodendaal, line 401: I think we need a unit test covering this use case

hansieodendaal (commented):
I agree with this, but I think we need a unit test covering the new use case.

base_layer/wallet/src/output_manager_service/tasks/txo_validation_task.rs
SWvheerden: you want a unit test for the database call?

PR #6176, opened 2024-03-01
chore: add localnet gen block

SWvheerden:
Description
---
added localnet gen block so that cucumbers can use that block. Local net used esmeralda block before, but when rewinding back to 0, the ouput_mr will be different between esmeralda and localnet, so we need a unique block for local net as well.


brianp (approved):
LGTM
utAck

PR #6183, opened 2024-03-05
feat: mining ffi add coinbase add

SWvheerden:
Description
---
add functionality to add coinbase

### DIFF
--- a/base_layer/tari_mining_helper_ffi/src/lib.rs
+++ b/base_layer/tari_mining_helper_ffi/src/lib.rs
@@ -305,7 +305,7 @@ pub unsafe extern "C" fn inject_coinbase(
             return;
         },
     };
-    let network = match Network::try_from(network as u8) {
+    let network = match Network::try_from(network_u8) {
         Ok(v) => v,
         Err(e) => {
             error = MiningHelperError::from(InterfaceError::InvalidNetwork(e.to_string())).code;
--- a/base_layer/tari_mining_helper_ffi/src/lib.rs
+++ b/base_layer/tari_mining_helper_ffi/src/lib.rs
@@ -266,7 +266,7 @@ pub unsafe extern "C" fn inject_nonce(header: *mut ByteVector, nonce: c_ulonglon
 #[no_mangle]
 pub unsafe extern "C" fn inject_coinbase(
     block_template_bytes: *mut ByteVector,
-    value: c_ulonglong,
+    coibase_value: c_ulonglong,
     stealth_payment: bool,
     revealed_value_proof: bool,
     wallet_payment_address: *const c_char,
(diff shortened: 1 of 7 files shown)


brianp (dismissed):
utAck

base_layer/tari_mining_helper_ffi/src/lib.rs
```diff
@@ -233,6 +245,153 @@ pub unsafe extern "C" fn inject_nonce(header: *mut ByteVector, nonce: c_ulonglon
...
+    let key_manager = create_memory_db_key_manager();
+
+    let consensus_manager = match ConsensusManager::builder(network).build() {
+        Ok(v) => v,
+        Err(e) => {
+            error = MiningHelperError::from(InterfaceError::NullError(e.to_string())).code;
+            ptr::swap(error_out, &mut error as *mut c_int);
+            return;
+        },
+    };
```
hansieodendaal, line 351: I do not think this error type should be `NullError`

base_layer/tari_mining_helper_ffi/src/lib.rs
```diff
@@ -233,6 +245,153 @@ pub unsafe extern "C" fn inject_nonce(header: *mut ByteVector, nonce: c_ulonglon
...
+            value.into(),
+            height,
+            coinbase_extra_string.as_bytes(),
+            &key_manager,
+            &wallet_address,
+            stealth_payment,
+            consensus_manager.consensus_constants(height),
+            range_proof_type,
+        )
+        .await
```
hansieodendaal: Why are the fees (1st parameter) zero?

base_layer/tari_mining_helper_ffi/src/lib.rs
```diff
@@ -233,6 +245,153 @@ pub unsafe extern "C" fn inject_nonce(header: *mut ByteVector, nonce: c_ulonglon
...
+/// `block_template_bytes` - The updated block template
+/// `error_out` - Error code returned, 0 means no error
+///
+/// # Safety
+/// None
+#[allow(clippy::too_many_lines)]
+#[no_mangle]
+pub unsafe extern "C" fn inject_coinbase(
+    block_template_bytes: *mut ByteVector,
+    value: c_ulonglong,
```
hansieodendaal: Maybe `reward` is better?
```suggestion
    reward: c_ulonglong,
```

hansieodendaal (commented):
Looks good, but I think we are missing the fees.

base_layer/tari_mining_helper_ffi/src/lib.rs
SWvheerden: they only way this will fail is if the network is null

base_layer/tari_mining_helper_ffi/src/lib.rs
SWvheerden: No specifically moved to value as to get away from fee and reward

SWvheerden:
Nope fees are not missing, you need to include it in the value you want the coinbase to be

base_layer/tari_mining_helper_ffi/src/lib.rs
```diff
@@ -233,6 +245,153 @@ pub unsafe extern "C" fn inject_nonce(header: *mut ByteVector, nonce: c_ulonglon
     (*header).0 = buffer;
 }
 
+/// Injects a coinbase into a blocktemplate
+///
+/// ## Arguments
+/// `block_template_bytes` - The block template as bytes, serialized with borsh.io
+/// `value` - The value of the coinbase
```
hansieodendaal, line 252: If this includes the fee it should be like `reward_and_fee` or something and the comment should explain it

PR #6184, opened 2024-03-05
chore: change default network

SWvheerden:
Description
---
Changes default `Mainnet` type as `Stagenet`


hansieodendaal (dismissed):
utACK

PR #6188, opened 2024-03-06
feat!: change split to 50-50

### LINKED ISSUE #6187: Change coin split to 50-50
Change the code split between Sha3x and Randomx to be 50-50,

SWvheerden:
Description
---
Change the algo split to 50-50

Motivation and Context
---
Change the split not to favor any algorithm. Having the split not be exactly 50-50, it slightly favors the less one if they arrive within the same block period. 

Fixes: #6187


hansieodendaal (dismissed):
utACK

PR #6192, opened 2024-03-07
chore: nextnext build fixes

SWvheerden:
Description
---
Fixes the build of clean build of nextnet due to tari log4rs package rename


brianp (approved):
LGTM

utAck

Issue #6195, opened 2024-03-08
Console wallet with many outputs sql queries optimize


hansieodendaal:
Using a console wallet with a large amount of outputs (84,000+), it takes uo to 6s to construct a one-sided transaction spending one output to one receiver with change to self.

Some of the profiled db queries (there are other time-hungry operations apart form the db access as well):
```rust
2024-03-04 14:01:07.955113400 [wallet::output_manager_service::database::wallet] TRACE sqlite profile - fetch_unspent_outputs_for_spending: lock 0 + db_op 1047 = 1047 ms
2024-03-04 14:01:10.982744500 [wallet::output_manager_service::database::wallet] TRACE sqlite profile - short_term_encumber_outputs (TxId: 4048846998157462111): lock 0 + db_op 1 = 1 ms
2024-03-04 14:01:11.148263700 [wallet::output_manager_service::database::wallet] TRACE sqlite profile - confirm_encumbered_outputs (TxId: 4048846998157462111): lock 0 + db_op 163 = 163 ms
2024-03-04 14:01:11.156091400 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 5 + db_op 0 = 6 ms
2024-03-04 14:01:11.163138600 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 6 + db_op 0 = 6 ms
2024-03-04 14:01:11.170076200 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 0 + db_op 6 = 6 ms
2024-03-04 14:01:11.174403000 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 0 + db_op 4 = 4 ms
2024-03-04 14:01:11.178777200 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 0 + db_op 4 = 4 ms
2024-03-04 14:01:11.183380800 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 0 + db_op 4 = 4 ms
2024-03-04 14:01:11.187884400 [wallet::key_manager_service::database::wallet] TRACE sqlite profile - insert_imported_key: lock 0 + db_op 4 = 4 ms
[...]

PR #6196, opened 2024-03-08
feat: optimize transaction validation db queries

hansieodendaal:
Description
---
Optimized transaction validation SQLite queries to run in batch mode wherever possible to minimise db operations.

Motivation and Context
---
See #6195

How Has This Been Tested?
---
Updated unit tests
Added a raw query regression test 
System-level testing

---
Review code and tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/wallet/src/output_manager_service/storage/database/backend.rs
```diff
@@ -37,29 +38,32 @@ pub trait OutputManagerBackend: Send + Sync + Clone {
     /// Modify the state the of the backend with a write operation
     fn write(&self, op: WriteOperation) -> Result<Option<DbValue>, OutputManagerStorageError>;
     fn fetch_pending_incoming_outputs(&self) -> Result<Vec<DbWalletOutput>, OutputManagerStorageError>;
-
-    fn set_received_output_mined_height_and_status(
+    /// Perform a batch update of the received outputs' mined height and status
+    fn set_received_outputs_mined_height_and_status_batch_mode(
```
SWvheerden:
```suggestion
    fn set_received_outputs_mined_height_and_status(
```
NIT
I dont like the added `batch_mode` text to trait.
This leaks internal workings to the outer API. 
And the way you did in internally using a single or batched query, is the correct implementation that most people would assume its implemented if you read that the function takes in multiple fields.

SWvheerden (dismissed):
This looks good, I just dont like the new trait function names for the reasons stated

base_layer/wallet/src/output_manager_service/storage/database/backend.rs
hansieodendaal: Yeah, the `_batch_mode` suffix is just to support these changes for now, but will disappear when the PR is ready for final review.

SWvheerden (dismissed):

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -415,64 +416,98 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
+             value, output_type, maturity, hash, script, input_data, script_private_key, sender_offset_public_key, \
+             metadata_signature_ephemeral_commitment, metadata_signature_ephemeral_pubkey, metadata_signature_u_a, \
+             metadata_signature_u_x, metadata_signature_u_y, spending_priority, covenant, encrypted_data, \
+             minimum_value_promise
+            )
+             VALUES ",
+        );
+
+        for update in &updates {
+            query.push_str(&format!(
```
stringhandler: Use bound parameters instead

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -525,93 +562,155 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
         let start = Instant::now();
         let mut conn = self.database_connection.get_pooled_connection()?;
         let acquire_lock = start.elapsed();
-        let hash = hash.to_vec();
-        diesel::update(outputs::table.filter(outputs::hash.eq(hash)))
-            .set((outputs::last_validation_timestamp
-                .eq::<Option<NaiveDateTime>>(NaiveDateTime::from_timestamp_opt(Utc::now().timestamp(), 0)),))
+
+        diesel::update(outputs::table.filter(outputs::hash.eq_any(hashes.iter().map(|hash| hash.to_vec()))))
+            .set(outputs::last_validation_timestamp.eq(Some(Utc::now().naive_utc())))
```
stringhandler: Perhaps it's best to not update this if the `last_validation_timestamp` is newer than the current timestamp

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -525,93 +562,155 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
-            .execute(&mut conn)
-            .num_rows_affected_or_not_found(1)?;
+
+        debug!(
+            target: LOG_TARGET,
+            "`mark_outputs_as_spent` for {} outputs",
+            updates.len()
+        );
+
+        let mut query = String::from(
```
stringhandler, line 643: See above

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -415,64 +416,98 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
-                outputs::status.eq(status),
-                outputs::mined_timestamp.eq(timestamp),
-                outputs::marked_deleted_at_height.eq::<Option<i64>>(None),
-                outputs::marked_deleted_in_block.eq::<Option<Vec<u8>>>(None),
-                outputs::last_validation_timestamp.eq::<Option<NaiveDateTime>>(None),
-            ))
-            .execute(&mut conn)
-            .num_rows_affected_or_not_found(1)?;
+
+        let mut query = String::from(
```
stringhandler, line 454: This logic seems different. Why would we insert empty data? Is there a process to correct these amounts?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -525,93 +562,155 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
-        let hash = hash.to_vec();
-        let status = if confirmed {
-            OutputStatus::Unspent
-        } else {
-            OutputStatus::UnspentMinedUnconfirmed
-        };
-        debug!(target: LOG_TARGET, "mark_output_as_unspent({})", hash.to_hex());
-        diesel::update(outputs::table.filter(outputs::hash.eq(hash)))
+        // Split out the confirmed and unconfirmed outputs
+        let confirmed_hashes = hashes
```
stringhandler, line 698: Why split them? Perhaps a comment explaining the benefit here

stringhandler (changes requested):
While this SQL doesn't include a SQL Injection vulnerabilty, it is not using best practices for including parameters in sql. This should be changed now rather than later

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -415,64 +416,98 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
+                    val.to_string()
+                } else {
+                    "NULL".to_string()
+                },
+            ));
+        }
+
+        // Remove the trailing comma
+        query.pop();
+        query.pop();
```
brianp:
```suggestion
        // Remove the trailing comma, and space
        query.pop();
        query.pop();
```

I feel like the need for a `pop,pop` is a code smell. The normal way to handle this is a loop that generates a `Vec<String>` that you `join(', ')`, and then finally `query.push()`

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -415,64 +416,98 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
-            ))
-            .execute(&mut conn)
-            .num_rows_affected_or_not_found(1)?;
+
+        let mut query = String::from(
+            "INSERT INTO outputs ( commitment, mined_height, mined_in_block, status, mined_timestamp, spending_key, \
+             value, output_type, maturity, hash, script, input_data, script_private_key, sender_offset_public_key, \
+             metadata_signature_ephemeral_commitment, metadata_signature_ephemeral_pubkey, metadata_signature_u_a, \
+             metadata_signature_u_x, metadata_signature_u_y, spending_priority, covenant, encrypted_data, \
+             minimum_value_promise
```
brianp, line 459: How often do these values change, and does the ordering here matter.

For maintenance overtime how easy is this for a developer to come along, remove or add a field and break the query.

Are these fields managed anywhere as a list in constants or fields of an existing structure we could reference instead of strings in the query?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
hansieodendaal: The first part of the query only identifies the rows that would be affected
`INSERT INTO outputs (..) VALUES (...) ON CONFLICT (commitment)` and does not change anything.
The only values in that list that are important are the ones to be used in the second part of the query, so all other values can by `NULL / 0`.

The actual update comes after as 
```
DO UPDATE SET mined_height = excluded.mined_height, mined_in_block = \
    excluded.mined_in_block, status = excluded.status, mined_timestamp = excluded.mined_timestamp, \
    marked_deleted_at_height = NULL, marked_deleted_in_block = NULL, last_validation_timestamp = NULL
```
That is equivalent to
```
            .set((
                outputs::mined_height.eq(mined_height as i64),
                outputs::mined_in_block.eq(mined_in_block),
                outputs::status.eq(status),
                outputs::mined_timestamp.eq(timestamp),
                outputs::marked_deleted_at_height.eq::<Option<i64>>(None),
                outputs::marked_deleted_in_block.eq::<Option<Vec<u8>>>(None),
                outputs::last_validation_timestamp.eq::<Option<NaiveDateTime>>(None),
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
hansieodendaal: That list comprises field names in `diesel::table! {outputs (id) {...} }` that cannot be left empty when inserting a row and all fields that need to be updated in the second half of the query if not included already. The frequency of change would coincide with changes to that table.

For future maintenance, I have added unit test `pub async fn test_raw_custom_queries_regression()`. It specifically tests the raw queries in `set_received_outputs_mined_height_and_statuses`  and `mark_outputs_as_spent`.

As an example, executing the full query in **DB Browser for SQLite** for two rows:
```sql
INSERT INTO outputs 
(commitment, mined_height, mined_in_block, status, mined_timestamp, spending_key, value, output_type, maturity, hash, script, input_data, script_private_key, sender_offset_public_key,
metadata_signature_ephemeral_commitment, metadata_signature_ephemeral_pubkey, metadata_signature_u_a, metadata_signature_u_x, metadata_signature_u_y, spending_priority, covenant,
encrypted_data, minimum_value_promise) 
VALUES 
(x'be9a7ce5362db3882fdf138de9840a9329ae48e30c945d60a013758f76ce7239', 2, x'00', 6, '2024-03-01 11:52:54', 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0), 
(x'de0f779865d752563f609c07669afff260b946bb2c805b933c071c26e308c232', 2, x'00', 6, '2024-03-01 11:59:04', 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0) 
ON CONFLICT (commitment) DO UPDATE SET 
mined_height = excluded.mined_height, mined_in_block = excluded.mined_in_block, status = excluded.status, mined_timestamp = excluded.mined_timestamp, marked_deleted_at_height = NULL, marked_deleted_in_block = NULL, last_validation_timestamp = NULL
```
results in 
```
Execution finished without errors.
Result: query executed successfully. Took 0ms, 2 rows affected
```
As a test let's leave out `input_data` from the list of columns as well as the corresponding `0` in both value rows, then we get
```
Execution finished with errors.
Result: NOT NULL constraint failed: outputs.input_data
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
hansieodendaal: I do not follow. How could `last_validation_timestamp` be newer than the current timestamp?

base_layer/wallet/src/output_manager_service/storage/sqlite_db/mod.rs
```diff
@@ -415,64 +416,180 @@ impl OutputManagerBackend for OutputManagerSqliteDatabase {
...
-                outputs::mined_timestamp.eq(timestamp),
-                outputs::marked_deleted_at_height.eq::<Option<i64>>(None),
-                outputs::marked_deleted_in_block.eq::<Option<Vec<u8>>>(None),
-                outputs::last_validation_timestamp.eq::<Option<NaiveDateTime>>(None),
-            ))
-            .execute(&mut conn)
-            .num_rows_affected_or_not_found(1)?;
+
+        let mut query = String::from(
+            "INSERT INTO outputs ( commitment, mined_height, mined_in_block, status, mined_timestamp, spending_key, \
```
SWvheerden, line 455: One problem here is that we make a mistake here and give it something that does not exist, we will add it to the DB. 
And thus rather than fail on the update, its going to write junk into the database breaking the database. 

Something like a join union should work as well and not have this problem

SWvheerden (changes requested):

stringhandler (dismissed):

SWvheerden (dismissed):

PR #6197, opened 2024-03-11
fix: horizon sync

SWvheerden:
Description
---
Fixes horizon sync

Motivation and Context
---
Burned outputs are counted as spent, so we should not sync them when doing horizon sync. 

How Has This Been Tested?
---
manual 

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):

hansieodendaal (commented):
Nice, good catch

PR #6200, opened 2024-03-11
chore: new gen blocks

SWvheerden:
Description
---
creates new gen blocks for all networks


brianp (approved):
utAck

No faucets on esme?

PR #6210, opened 2024-03-13
fix: base node write tor address

SWvheerden:
Description
---
Base will now write tor address in the identity file

Motivation and Context
---
The base node needs to contain the tor address. 

How Has This Been Tested?
---
manual


hansieodendaal (approved):
ACK

PR #6213, opened 2024-03-13
chore: allow fixed tests

### LINKED ISSUE #5996: fix pruned mode tests in cucumber
There are broken disabled cucumber tests for pruned mode. 
These should be fixed. 
They are denoted by: `@broken_prune`

SWvheerden:
Description
---
Allow fixed tests

Fixes: #5996


hansieodendaal (approved):
utACK

PR #6177, opened 2024-03-01
fix!: change grpc deny to allow

stringhandler:
Description
---
Removes the antipattern of a deny list and puts in an allow list

Motivation and Context
---
A deny list is an antipattern because new methods are not automatically added to the list

How Has This Been Tested?
---
Tested with dan testing

What process can a PR reviewer use to test or verify this change?
---

Breaking Changes
---

- [ ] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [x] Other - Please specify

BREAKING CHANGE: Removed `grpc_server_deny_methods` and added `grpc_server_allow_methods`

PR #6218, opened 2024-03-14
fix!: change grpc deny to allow

SWvheerden:
Description
---
Removes the antipattern of a deny list and puts in an allow list

Motivation and Context
---
A deny list is an antipattern because new methods are not automatically added to the list

Replaces: #6177


common/config/presets/c_base_node_b_mining_allow_methods.toml
```diff
@@ -12,22 +13,22 @@ grpc_enabled = false
 # Use gRPC over TLS (default = false)
 #grpc_tls_enabled = false
 
-# Uncomment all gRPC server methods that should be denied default (only active when `grpc_enabled = true`)
-grpc_server_deny_methods = [
+# Uncomment all gRPC server methods that should be denied (only active when `grpc_enabled = true`)
```
hansieodendaal:
```suggestion
# Uncomment all gRPC server methods that should be allowed (only active when `grpc_enabled = true`)
```

common/config/presets/c_base_node_b_non_mining_allow_methods.toml
```diff
@@ -12,22 +13,22 @@ grpc_enabled = true
 # Use gRPC over TLS (default = false)
 #grpc_tls_enabled = false
 
-# Uncomment all gRPC server methods that should be denied default (only active when `grpc_enabled = true`)
-grpc_server_deny_methods = [
+# Uncomment all gRPC server methods that should be denied (only active when `grpc_enabled = true`)
```
hansieodendaal:
```suggestion
# Uncomment all gRPC server methods that should be allowed (only active when `grpc_enabled = true`)
```

common/src/configuration/utils.rs
```diff
@@ -119,18 +119,18 @@ pub fn prompt_default_config() -> [&'static str; 12] {
 /// Returns the default configuration file template in parts from the embedded presets. If use_mining_config is true,
 /// the base node configuration that enables mining is returned, otherwise the non-mining configuration is returned.
 pub fn get_default_config(use_mining_config: bool) -> [&'static str; 12] {
-    let base_node_deny_methods = if use_mining_config {
-        include_str!("../../config/presets/c_base_node_b_mining_deny_methods.toml")
+    let base_node_allow_methods = if use_mining_config {
```
hansieodendaal, line 122: What about these methods `second_layer_grpc_enabled` ? I assume it should be set from initial startup.
A node could also double as mining node and second layer comms node.

hansieodendaal (commented):
Some comments below, thnaks.

hansieodendaal (approved):
utACK

PR #6225, opened 2024-03-20
feat!: enable multiple second layers to be registered

stringhandler:
Description
---
Allows the registration of multiple second layers for validator nodes, templates and burns. The default second layer (Tari) is specified by leaving the `validator_network` field empty. For other networks, they can either be private, where the private key is kept secret or public in which case the `validator_network` private key can be exposed and published.

Motivation and Context
---
Specifically for testing it will be useful to have multiple second layer testnets run on the same base layer (I intend to use `igor`). For example we may want to run a second layer "ContractNet" specifically for users who wish to run templates and contracts but don't want to test running validator nodes. In contrast, users who are interested in testing validator nodes can run on the default second layer.

This way of doing it also allows users who wish to run their own private networks, with their own validator nodes, to run them. 

How Has This Been Tested?
---
Tested manually, and busy testing with changes to the second layer and dan-testing

---
You can register templates and burns in the console wallet. Registering validator nodes is a bit harder. 

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

BREAKING CHANGE: The merkle tree for validator nodes is calculated in a new manner


base_layer/core/src/validation/aggregate_body/aggregate_body_internal_validator.rs
```diff
@@ -143,6 +145,58 @@ impl AggregateBodyInternalConsistencyValidator {
...
+            let sig_pub_key = conf_output.network.as_ref().unwrap();
+            if !conf_output
+                .network_knowledge_proof
+                .as_ref()
+                .unwrap()
+                .verify(sig_pub_key, conf_output.claim_public_key.to_vec())
+            {
+                return Err(ValidationError::TemplateInvalidNetworkKnowledgeProof);
+            }
+        }
```
sdbondi: nit: think we should dry out the sidechain network proof code

```suggestion
        if conf_output.network.is_some() || conf_output.network_knowledge_proof.is_some() {
            // If one of these is set, both must be set
            if conf_output.network.is_none() || conf_output.network_knowledge_proof.is_none() {
                return Err(ValidationError::ConfidentialOutputNetworkNotSet);
            }
            
               check_sidechain_network_proof(
                 conf_output.network.as_ref().unwrap(),
                 conf_output.network_knowledge_proof.as_ref().unwrap()
               )?;   
       }
```

stringhandler: They return specific errors, but I suppose you could return the same error for each of them

sdbondi: ah yeah, not too fussed I see why you did that

sdbondi (approved):
utACK

PR #6226, opened 2024-03-20
feat: new template with coinbase call

SWvheerden:
Description
---
This adds a new grpc call to take a block template and add coinbases to it as desired. The callee is required to ensure that the coinbases amounts provided are correct and per consensus. 

Motivation and Context
---
This is to allow external miners to create a block with 1 sided coinbase utxo's to be added without having to create them as this necessitates the inclusion of all Tari crypto libraries. 

How Has This Been Tested?
---
New unit and cucumber tests

### DIFF
--- a/applications/minotari_app_grpc/proto/base_node.proto
+++ b/applications/minotari_app_grpc/proto/base_node.proto
@@ -59,6 +59,8 @@ service BaseNode {
     rpc GetNewBlock(NewBlockTemplate) returns (GetNewBlockResult);
     // Construct a new block from a provided template
     rpc GetNewBlockWithCoinbases(GetNewBlockWithCoinbasesRequest) returns (GetNewBlockResult);
+    // Construct a new block from a provided template
+    rpc GetNewBlockTemplateWithCoinbases(GetNewBlockTemplateWithCoinbasesRequest) returns (GetNewBlockResult);
     // Construct a new block and header blob from a provided template
     rpc GetNewBlockBlob(NewBlockTemplate) returns (GetNewBlockBlobResult);
     // Submit a new block for propagation
@@ -184,10 +186,17 @@ message NewBlockTemplateRequest{
     uint64 max_weight = 2;
 }
 
+/// return type of NewBlockTemplateRequest
+message GetNewBlockTemplateWithCoinbasesRequest{
+    PowAlgo algo = 1;
+    //This field should be moved to optional once optional keyword is standard
+    uint64 max_weight = 2;
+    repeated  NewBlockCoinbase coinbases = 3;
+}
+
 /// request  type of GetNewBlockWithCoinbasesRequest
 message GetNewBlockWithCoinbasesRequest{
     NewBlockTemplate new_template = 1;
-    //This field should be moved to optional once optional keyword is standard
     repeated  NewBlockCoinbase coinbases = 2;
 }
 
@@ -365,6 +374,7 @@ message GetNewBlockResult{
     Block block = 2;
     bytes merge_mining_hash =3;
     bytes tari_unique_id =4;
+    MinerData miner_data = 5;
 }
 
 // This is the message that is returned for a miner after it asks for a new block.
--- a/applications/minotari_app_grpc/proto/base_node.proto
+++ b/applications/minotari_app_grpc/proto/base_node.proto
@@ -57,6 +57,8 @@ service BaseNode {
     rpc GetNewBlockTemplate(NewBlockTemplateRequest) returns (NewBlockTemplateResponse);
     // Construct a new block from a provided template
     rpc GetNewBlock(NewBlockTemplate) returns (GetNewBlockResult);
+    // Construct a new block from a provided template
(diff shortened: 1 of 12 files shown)


applications/minotari_app_grpc/proto/base_node.proto
```diff
@@ -182,6 +184,21 @@ message NewBlockTemplateRequest{
     uint64 max_weight = 2;
 }
 
+/// request  type of GetNewBlockWithCoinbasesRequest
+message GetNewBlockWithCoinbasesRequest{
```
stringhandler, line 198: Would it not be better to just have one call instead of getting the Template and then calling this? The miner has no reason to make two calls anymore. 

The miner previously would add a coinbase but now it shouldn't do anything with the `NewBlockTemplate`

applications/minotari_app_grpc/proto/base_node.proto
```diff
@@ -182,6 +184,21 @@ message NewBlockTemplateRequest{
...
+    NewBlockTemplate new_template = 1;
+    //This field should be moved to optional once optional keyword is standard
```
stringhandler: Is this comment correct? AFIAK `optional` is being removed

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -690,6 +710,197 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+
+        let mut total_excess = Commitment::default();
+        let mut total_nonce = PublicKey::default();
+        let mut private_keys = Vec::new();
+        let mut kernel_message = [0; 32];
+        let mut last_kernel = Default::default();
+        for coinbase in coinbases {
+            let address = TariAddress::from_hex(&coinbase.address)
+                .map_err(|e| obscure_error_if_true(report_error_flag, Status::internal(e.to_string())))?;
+            let range_proof_type = if coinbase.revealed_value_proof {
```
stringhandler, line 1024: I think we should only allow revealed values

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -690,6 +710,197 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+                RangeProofType::RevealedValue
+            } else {
+                RangeProofType::BulletProofPlus
+            };
+            let (_, coinbase_output, coinbase_kernel, wo) = generate_coinbase_with_wallet_output(
```
stringhandler:
```suggestion
            let (_, coinbase_output, coinbase_kernel, wallet_output) = generate_coinbase_with_wallet_output(
```

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -690,6 +710,197 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+                &script_key_id,
+                &address,
+                coinbase.stealth_payment,
+                self.consensus_rules.consensus_constants(height),
+                range_proof_type,
+            )
+            .await
+            .map_err(|e| obscure_error_if_true(report_error_flag, Status::internal(e.to_string())))?;
+            block_template.body.add_output(coinbase_output);
+            let (new_p_nonce, pub_nonce) = key_manager
```
stringhandler:
```suggestion
            let (new_private_nonce, pub_nonce) = key_manager
```

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -690,6 +710,197 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+                return Err(obscure_error_if_true(report_error_flag, status));
+            },
+            Err(e) => {
+                return Err(obscure_error_if_true(
+                    report_error_flag,
+                    Status::internal(e.to_string()),
+                ))
+            },
+        };
+        let gen_hash = handler
```
stringhandler, line 1119: Where does this get used?

stringhandler (commented):
I think it is better to just use one call instead of getting the block template, and then adding the coinbases.

One problem is that the fees might be different. In this case I would suggest adding all the fees to the first address supplied. 

If we would prefer to do the fees fairly, then we can keep the 2 call approach, but I would suggest setting the timestamp at the same time then.

applications/minotari_node/src/grpc/base_node_grpc_server.rs
SWvheerden: In merge mining, its used as the unique identifier for tari

applications/minotari_node/src/grpc/base_node_grpc_server.rs
SWvheerden: If we do that, I think we need to make it a consensus rule with this

stringhandler (approved):

PR #6233, opened 2024-03-26
fix: potential in panic message_vector_get_at

SWvheerden:
Description
---
fixes potential in panic message_vector_get_at

Motivation and Context
---
creating the message vector and asking for index[0] will result in a panic

### DIFF
--- a/base_layer/chat_ffi/src/messages.rs
+++ b/base_layer/chat_ffi/src/messages.rs
@@ -147,15 +147,16 @@ pub unsafe extern "C" fn message_vector_get_at(
     }
 
     let messages = &(*messages);
+    let position = position as usize;
 
-    let len = messages.0.len() - 1;
-    if position as usize > len {
+    let len = messages.0.len();
+    if (position + 1 ) > len {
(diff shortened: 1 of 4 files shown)


base_layer/chat_ffi/src/messages.rs
```diff
@@ -147,15 +147,16 @@ pub unsafe extern "C" fn message_vector_get_at(
     }
 
     let messages = &(*messages);
+    let position = position as usize;
```
sdbondi, line 150: If `position` is set to usize::MAX (possible on 32 bit systems), `position + 1` will either panic or overflow to 0, in which case `messages.0[usize::MAX]` will probably panic.

Suggest: 
```rust
if messages.0.is_empty() {
// error
}
let len = messages.0.len() - 1;
if position as usize > len {
// error
}
```

`position as usize` should be ok as it _should_ be defined as a u32 with some rare exceptions according to the `c_uint` docs. To be safer you could use `usize::try_from`

sdbondi (approved):
:+1:

PR #6228, opened 2024-03-23
chore: update `tari-crypto` and curve library dependencies

AaronFeickert:
Description
---
Updates `tari-crypto` and associated dependencies. Bumps the nightly toolchain. Fixes lints. Removes unused code.

Closes #6138.

Motivation and Context
---
Updates to `tari-crypto` and related dependencies allow for a return to the upstream curve library. This also requires bumping the nightly toolchain, which brings fun new lints that need to be fixed. This PR addresses all of these, and removes unused code.

How Has This Been Tested?
---
Existing tests pass.

What process can a PR reviewer use to test or verify this change?
---
Watch that CI! Ensure that the removed code is unused.

PR #6238, opened 2024-03-26
chore: remove ahash as dependancy

SWvheerden:
Description
---
remove ahash as dependancy

ahash is not used directly but is a sub dependency,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,17 +53,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


AaronFeickert:
This will pave the way for #6228, which updates `tari-crypto` and the curve library, when it's ready to go. Specifically, it lets us move the nightly toolchain forward and support the most recent curve library version.

AaronFeickert (approved):
utACK

PR #6239, opened 2024-03-27
feat: prevent mempool panic

hansieodendaal:
Description
---
Under certain conditions, it is possible to let the mempool panic when is inserting an empty transaction. This PR prevents it.

Motivation and Context
---
HAZOP finding

How Has This Been Tested?
---
Unit tests added

---
Review code changes and unit tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/mempool/priority/prioritized_transaction.rs
+++ b/base_layer/core/src/mempool/priority/prioritized_transaction.rs
@@ -88,6 +88,9 @@ impl PrioritizedTransaction {
         dependent_outputs: Option<Vec<HashOutput>>,
     ) -> Result<PrioritizedTransaction, TransactionError> {
         let weight = transaction.calculate_weight(weighting)?;
+        if weight == 0 {
+            return Err(TransactionError::ZeroWeight);
+        }
         let insert_epoch = match SystemTime::now().duration_since(UNIX_EPOCH) {
             Ok(n) => n.as_secs(),
             Err(_) => 0,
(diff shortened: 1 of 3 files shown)


stringhandler (commented):
Where was the original panic? I can't see where it would happen from this pr

hansieodendaal:
> Where was the original panic? I can't see where it would happen from this pr

The panic is where we divide by zero weight in `PrioritizedTransaction::new`, here
```rust
priority: FeePriority::new(&transaction, insert_epoch, weight)?,
```
and here
```rust
fee_per_byte: transaction.body.get_total_fee()?.as_u64().saturating_mul(1000) / weight,
```

base_layer/core/src/mempool/priority/prioritized_transaction.rs
```diff
@@ -88,6 +88,9 @@ impl PrioritizedTransaction {
         dependent_outputs: Option<Vec<HashOutput>>,
     ) -> Result<PrioritizedTransaction, TransactionError> {
         let weight = transaction.calculate_weight(weighting)?;
+        if weight == 0 {
+            return Err(TransactionError::ZeroWeight);
+        }
```
AaronFeickert: It seems safer to replace the subsequent division by `weight` by a `checked_div`, which would make it much harder for a regression to sneak in. You could have failure of the `checked_div` return the same `TransactionError`  too, so the new test would still work the same.

AaronFeickert (changes requested):

base_layer/core/src/mempool/priority/prioritized_transaction.rs
hansieodendaal: Good observation, thanks

AaronFeickert (approved):
LGTM

SWvheerden (approved):

PR #6240, opened 2024-03-27
feat(wallet): add command to create owner proof

stringhandler:
Description
---
Create a wallet command to create a proof (commitment signature) that you own a commitment

Motivation and Context
---
In some scenarios you want to prove that you own or can spend a commitment on chain, without actually spending it. 

How Has This Been Tested?
---
Tested with `cargo run --release --bin minotari_console_wallet -- -b . create-owner-proof 6234f3569a9dac7acf690348468d1fd87f59db7273f3851641c12417d4efea2b hello`

---
as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -22,13 +22,15 @@
 
 use std::{convert::TryInto, fmt, sync::Arc};
 
+use blake2::Blake2b;
 use diesel::result::{DatabaseErrorKind, Error as DieselError};
+use digest::{consts::U64, Digest};
 use futures::{pin_mut, StreamExt};
 use log::*;
 use rand::{rngs::OsRng, RngCore};
 use tari_common_types::{
     transaction::TxId,
-    types::{BlockHash, Commitment, HashOutput, PrivateKey, PublicKey},
+    types::{BlockHash, Commitment, CommitmentFactory, HashOutput, PrivateKey, PublicKey},
 };
 use tari_comms::types::CommsDHKE;
 use tari_core::{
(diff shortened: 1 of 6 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -411,9 +427,48 @@ where
...
+            .await?;
+        let nonce_a = RistrettoSecretKey::random(&mut OsRng);
+        let nonce_x = RistrettoSecretKey::random(&mut OsRng);
+
+        let nonce_commitment = self.resources.factories.commitment.commit(&nonce_x, &nonce_a);
+        let challenge = Blake2b::<U64>::new()
+            .chain_update(commitment.as_bytes())
+            .chain_update(nonce_commitment.as_bytes())
+            .chain_update(format!("create_owner_proof::{}", message).as_bytes())
+            .finalize();
```
AaronFeickert, line 457: I realize this PR is still a draft, but I'd recommend using the existing domain-separated hash functionality instead of this message formatting approach if possible.

stringhandler: Yeah I got this working with what was available. This code is actually copied from the examples in taricrypto. I'm not really happy with it, but this is the API that ComSig exposes. Ideally the ComSig signing api should be updated to match the schnorr api, where the domain is part of the call

AaronFeickert: Right, and I know there's a longstanding [issue](https://github.com/tari-project/tari-crypto/issues/159) you filed to improve the API that was sidelined due to not using that signature until now. Even so, you should be able to generate the challenge hash using a domain-separated hasher anyway, no? Then you can ensure it'll take care of things like domain separation and input collisions automagically. I suspect it could even be done such that a future API update would retain full compatibility with existing signatures.

AaronFeickert:
After speaking with @stringhandler, it sounds like the use case of interest here is to prove two things: that you can open the commitment, and that the commitment binds to at least a specified value.

This means you don't need a commitment signature, but can get away with using a minimum-value range proof instead. The cryptographic plumbing already exists for this, and the Bulletproofs+ API (pending a new version tag) supports safe transcripting operations that allow for this to be done very safely. The proof is much larger (by several hundreds of bytes) than a commitment signature, but it handles both assertions simultaneously without the need to overhaul the commitment signature API.

I'll create a new PR that does this.

PR #6241, opened 2024-04-01
ci: bump nightly version

AaronFeickert:
Description
---
Bumps the nightly toolchain to `nightly-2024-02-04` in preparation for a curve library update.

Motivation and Context
---
An upcoming curve library update in #6228 will require moving the nightly toolchain. In preparation, this PR moves it as far as it can go without breaking. This lets us take care of lints now.

How Has This Been Tested?
---
Existing CI (but with the updated toolchain) passes.

---
Check that CI passes, that the nightly toolchain update is consistently applied, and that all code changes reflect only lint fixes.

### DIFF
--- a/.github/workflows/base_node_binaries.json
+++ b/.github/workflows/base_node_binaries.json
@@ -2,7 +2,7 @@
   {
     "name": "linux-x86_64",
     "runs-on": "ubuntu-20.04",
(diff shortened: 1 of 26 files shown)


hansieodendaal (approved):
ACK (tested on Windows)

SWvheerden (approved):

PR #6244, opened 2024-04-02
feat: show warning whrn GRPC method is disallowed

stringhandler:
Description
---
Adds a log and console warning if a GRPC method is called but not allowed

### DIFF
--- a/applications/minotari_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/minotari_node/src/grpc/base_node_grpc_server.rs
@@ -151,6 +151,17 @@ impl BaseNodeGrpcServer {
 
         self.config.grpc_server_allow_methods.contains(&grpc_method)
     }
+
+    fn check_method_enabled<T>(&self, method: GrpcMethod) -> Option<Result<Response<T>, Status>> {
+        if !self.is_method_enabled(method) {
+            warn!(target: LOG_TARGET, "`{}` method called but it is not allowed. Allow it in the config file or start the node with a different set of CLI options", method);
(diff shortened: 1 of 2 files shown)


hansieodendaal (approved):
utACK

brianp (approved):
I of course debated internally if `info` makes more sense then `warn` but I suppose someone is actively calling a method that has been disabled then a warn can be warranted.

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -151,6 +151,17 @@ impl BaseNodeGrpcServer {
...
+    fn check_method_enabled<T>(&self, method: GrpcMethod) -> Option<Result<Response<T>, Status>> {
+        if !self.is_method_enabled(method) {
+            warn!(target: LOG_TARGET, "`{}` method called but it is not allowed. Allow it in the config file or start the node with a different set of CLI options", method);
+            return Some(Err(Status::permission_denied(format!(
+                "`{}` method not made available",
+                method
+            ))));
+        }
+        None
+    }
```
delta1, line 164: why not return a `Result` so you can just do 

```
self.check_method_enabled(..)?
```

stringhandler: it has to return the Status to the client.

delta1: oh right

SWvheerden (approved):

PR #6245, opened 2024-04-02
feat: allow wallet type from db to have preference

hansieodendaal:
Description
---
- Allow wallet type from the wallet database to have preference. This will allow old wallets without wallet-type fields to be accepted as software wallets when added to the config file.
- Set `newline_style = "Auto"` as we do not use JavaScript cucumber anymore; this was an issue on Windows.

Motivation and Context
---
- Wallets prior to the wallet-type PR could not be used.
- Performing `cargo fmt` in Windows where the git preference has been set to _check out as is and commit unix style_ forced `CRLF`. `Auto` will give preference to the actual line-endings.

How Has This Been Tested?
---
System-level testing

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
utAck

Had to figure out why old wallets were a problem. I thought I had handled that situation. I've now realized I had initially made a change in the same function as you, but removed it and replaced it with a compile feature. Which means it didn't support old wallets when the feature was on. Which this would now fix.

PR #6249, opened 2024-04-05
refactor: fix template registration grpc

stringhandler:
Easier calling of template registration


base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1806,31 +1799,38 @@ where
...
-            OutputFeatures::for_template_registration(template_registration),
-            fee_per_gram,
-            message,
-            TransactionMetadata::default(),
-            join_handles,
-            transaction_broadcast_join_handles,
-            reply_channel,
-        )
-        .await
+        todo!()
```
hansieodendaal: Why is this function blocked out in this PR? I would expect some explanation in the PR description.

As an aside, I wonder what the 2nd layer will do with a lot of dust here, say 500,000 bogie registration UTXOs?

PR #6265, opened 2024-04-10
feat: keep smt memory

SWvheerden:
Description
---
This only keeps only a single copy of the smt in memory

Motivation and Context
---
Writing and load the smt from disc every time is inefficient and take long, this keeps in in memory. 
Keeping a single copy in an `Arc<RwLock<>>` alles the node to keep a single copy and use it. 

How Has This Been Tested?
---
unit tests and manual.

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -1332,6 +1332,7 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
     db: &T,
     rules: &ConsensusManager,
     block: &Block,
+    // we dont want to clone the SMT, so we rather change it and change it back after we are done.
     output_smt: &mut OutputSmt,
 ) -> Result<MmrRoots, ChainStorageError> {
     let header = &block.header;
@@ -1421,7 +1422,8 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
         validator_node_mr,
         validator_node_size: validator_node_size as u64,
     };
-    // lets rewind the smt back to tip again
+    // We have made changes to the SMT that we dont want, sp lets rewind the SMT back to tip again as we want to have
+    // the SMT at tip.
     for output in outputs_to_add {
         if output_smt.insert(output.0.clone(), output.1).is_err() {
             error!(
@@ -2013,7 +2015,7 @@ fn reorganize_chain<T: BlockchainBackend>(
             let mut write_smt = smt.write().map_err(|e| {
                 error!(
                     target: LOG_TARGET,
-                    "An attempt to get a write lock on the smt failed. {:?}", e
+                    "reorganize_chain could not get a write lock on the smt. {:?}", e
                 );
                 ChainStorageError::AccessError("write lock on smt".into())
             })?;
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -917,7 +917,7 @@ where B: BlockchainBackend
         let mmr_roots = match calculate_mmr_roots(&*db, self.rules(), &block, &mut smt) {
             Ok(v) => v,
             Err(e) => {
-                // some error happend, lets rewind the smt
+                // some error happend, lets reset the smt to its starting state
                 *smt = db.calculate_tip_smt()?;
                 return Err(e);
             },
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -208,6 +208,7 @@ pub struct BlockchainDatabase<B> {
     consensus_manager: ConsensusManager,
     difficulty_calculator: Arc<DifficultyCalculator>,
     disable_add_block_flag: Arc<AtomicBool>,
+    smt: Arc<RwLock<OutputSmt>>,
 }
 
 #[allow(clippy::ptr_arg)]
@@ -221,6 +222,7 @@ where B: BlockchainBackend
         validators: Validators<B>,
         config: BlockchainDatabaseConfig,
         difficulty_calculator: DifficultyCalculator,
+        smt: Arc<RwLock<OutputSmt>>,
     ) -> Result<Self, ChainStorageError> {
         debug!(target: LOG_TARGET, "BlockchainDatabase config: {:?}", config);
         let is_empty = db.is_empty()?;
@@ -231,6 +233,7 @@ where B: BlockchainBackend
             consensus_manager,
             difficulty_calculator: Arc::new(difficulty_calculator),
             disable_add_block_flag: Arc::new(AtomicBool::new(false)),
+            smt,
         };
         let genesis_block = Arc::new(blockchain_db.consensus_manager.get_genesis_block());
         if is_empty {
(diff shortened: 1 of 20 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2567,18 +2524,29 @@ impl BlockchainBackend for LMDBDatabase {
...
-        let k = MetadataKey::TipSmt;
-        let val: Option<OutputSmt> = lmdb_get(&txn, &self.tip_utxo_smt, &k.as_u32())?;
-        match val {
-            Some(smt) => Ok(smt),
-            _ => Err(ChainStorageError::ValueNotFound {
-                entity: "TipSmt",
-                field: "TipSmt",
-                value: "".to_string(),
-            }),
+    fn calculate_tip_smt(&self) -> Result<OutputSmt, ChainStorageError> {
```
hansieodendaal, line 2527: Please add a trace log that reports the time it takes to calculate the SMT, with some stat about the block height and SMT entry size etc.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1386,6 +1421,36 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
         validator_node_mr,
         validator_node_size: validator_node_size as u64,
     };
+    // lets rewind the smt back to tip again
```
hansieodendaal: Please explain a bit more in the comment, why this code is needed

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -891,7 +913,15 @@ where B: BlockchainBackend
                 "calculate_mmr_roots expected a sorted block body, however the block body was not sorted".to_string(),
             ));
         };
-        let mmr_roots = calculate_mmr_roots(&*db, self.rules(), &block)?;
+        let mut smt = self.smt_write_access()?;
+        let mmr_roots = match calculate_mmr_roots(&*db, self.rules(), &block, &mut smt) {
+            Ok(v) => v,
+            Err(e) => {
+                // some error happend, lets rewind the smt
```
```suggestion
                // some error happend, lets reset the smt to its starting state
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1299,6 +1332,7 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
     db: &T,
     rules: &ConsensusManager,
     block: &Block,
+    output_smt: &mut OutputSmt,
```
hansieodendaal, line 1336: Please add a comment to explain that the SMT will be temporarily mutated in order to calculate the MMR roots, and that it needs to be reset to its original state before the function exits

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1926,11 +2005,21 @@ fn reorganize_chain<T: BlockchainBackend>(
...
+            let mut write_smt = smt.write().map_err(|e| {
+                error!(
+                    target: LOG_TARGET,
+                    "An attempt to get a write lock on the smt failed. {:?}", e
+                );
+                ChainStorageError::AccessError("write lock on smt".into())
+            })?;
+            *write_smt = backend.calculate_tip_smt()?;
+            return Err(e);
+        }
```
hansieodendaal: I count 3x places in the code where this pattern is employed, where the SMT needs to be recalculated if some underlying error happens, If this function (`calculate_tip_smt`) errors we cannot recover. 

Maybe at this point the SMT should be wiped, and we should have a wrapper function that checks to see if the SMT is empty and try to calculate it whenever we want to use it.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -327,6 +330,30 @@ where B: BlockchainBackend
...
 
+    pub fn smt_write_access(&self) -> Result<RwLockWriteGuard<OutputSmt>, ChainStorageError> {
+        self.smt.write().map_err(|e| {
+            error!(
+                target: LOG_TARGET,
+                "An attempt to get a write lock on the smt failed. {:?}", e
+            );
+            ChainStorageError::AccessError("write lock on smt".into())
+        })
+    }
```
hansieodendaal, line 341: Could this function not be turned into a helper function instead of taking `&self`, taking `smt: Arc<RwLock<OutputSmt>>` instead? Then it could be used in all other places in the code? I count 4x other places where this pattern is employed.

base_layer/core/src/validation/block_body/block_body_full_validator.rs
```diff
@@ -78,8 +85,15 @@ impl BlockBodyFullValidator {
...
-        let mmr_roots = chain_storage::calculate_mmr_roots(backend, &self.consensus_manager, &block)?;
+        // validate the merkle mountain range roots+
+        let mut output_smt = smt.write().map_err(|e| {
+            error!(
+                target: LOG_TARGET,
+                "An attempt to get a write lock on the smt failed. {:?}", e
+            );
+            ChainStorageError::AccessError("write lock on smt".into())
+        })?;
+        let mmr_roots = chain_storage::calculate_mmr_roots(backend, &self.consensus_manager, &block, &mut output_smt)?;
```
hansieodendaal: See previous comment about a helper function.

hansieodendaal (changes requested):
I am really impressed by the effect of this PR. I ran system-level testing to sync a 927,181 UTXO blockchain from scratch and archival block sync took 30 min, whereas horizon sync took 2 hours and 7 minutes (still something off with our implementation there, but not due to this PR).

**_Calculating the SMT for this test blockchain took in excess of 1 minute._**
I propose we add an SMT "caching/snapshot" function to the code to save the SMT to file when the base node exits together with the block hash for that height. Whenever the base node starts up it can read the SMT from file, verify that it is for the best block it has, and continue. Whenever it exits again it can create a new snapshot at the best block and delete the previous one. Whenever the SMT needs to be re-calculated during runtime, it could also make use of the snapshot as a starting point.

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: if calculate_tip_smt fails, you entire db is stuffed, and you need to sync from scratch. Aka wipe the node. So I don't see the need to add a helper function to wipe it and only let it fail again. This replaces the SMT in memory.

SWvheerden:
If we want to save a min by caching the SMT, I created a ticket here: https://github.com/tari-project/tari/issues/6269
This PR is big enough as it is

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: It is better to have no SMT than to have one that is not in sync with the db. The calculate function could fail due to other reasons, not only due to bad lmdb data, for example the OS could be busy swapping memory to disk for a long time and impair other read operations.

SWvheerden: for that a restart of the node is better anyway

hansieodendaal: Not resolved

hansieodendaal (dismissed):
ACK

Notes: 
- Some comments were not resolved
- CI must pass

PR #6266, opened 2024-04-10
feat!: update new block proto to support multiple coinbases

### LINKED ISSUE #6189: fix new block proto message for multiple coinbases
```
// A new block message. This is the message that is propagated around the network. It contains the
// minimal information required to identify and optionally request the full block.
message NewBlock {
    // The block header.
    BlockHeader header = 1;
    // Coinbase kernel of the block.
    tari.types.TransactionKernel coinbase_kernel = 2;
    // Coinbase output of the block.
    tari.types.TransactionOutput coinbase_output = 3;
    // The scalar `s` component of the kernel excess signatures of the transactions contained in the block.
    repeated bytes kernel_excess_sigs = 4;
}
```
It can only handle 1 coinbase/kernel, we need multiple

SWvheerden:
Description
---
This changes the new block proto to allow multiple coinbases

Motivation and Context
---
Tari supports multiple coinbases and the new block proto only supports a single coinbase. 

Breaking Changes
---
Changes block propagation API 

Fixes: #6189

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -594,15 +594,16 @@ where B: BlockchainBackend + 'static
     ) -> Result<Block, CommsInterfaceError> {
         let NewBlock {
             header,
-            coinbase_kernel,
-            coinbase_output,
+            coinbase_kernels,
+            coinbase_outputs,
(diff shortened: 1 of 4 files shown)


hansieodendaal (commented):
Something is wrong with the cucumber tests, somehow this change causes many of them to fail.

Other than that the change seems fine.

PR #6284, opened 2024-04-17
feat: improve node and wallet connection times

hansieodendaal:
Description
---
- Improved peer db stats to only update if the specific public address was successfully connected to. This alleviates issues where the node tries to connect to a peer's public address with a quality score > 0 where it should be 0.
- Update stats with ping-pong.
- Added a peer sync delay config option to let a preferred connection get a head start in trying to connect.
- Discarded all peer identity claims without valid signatures immediately when they are received.
- Reduced the minimum number of sync peers a wallet wants to connect to from 50 to 8.

Motivation and Context
---
Peer db connection stats were not reflected correctly.
Wallet connection/re-connection times were slow.

How Has This Been Tested?
---
System-level testing

**Wallet re-connection times** with many pre-connected peers.

image

---
Code walk-through

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_app_grpc/proto/network.proto
+++ b/applications/minotari_app_grpc/proto/network.proto
@@ -70,7 +70,11 @@ message Address{
     bytes address =1;
     string last_seen = 2;
     uint32 connection_attempts = 3;
-    uint64 avg_latency = 5;
+    AverageLatency avg_latency = 5;
+}
+
+message AverageLatency {
+    uint64 latency = 1;
 }
 
 message ListConnectedPeersResponse {
--- a/common/config/presets/d_console_wallet.toml
+++ b/common/config/presets/d_console_wallet.toml
@@ -323,7 +323,7 @@ database_url = "data/wallet/dht.db"
 #network_discovery.enabled = true
 # A threshold for the minimum number of peers this node should ideally be aware of. If below this threshold a
 # more "aggressive" strategy is employed. Default: 50
-#network_discovery.min_desired_peers = 50
+network_discovery.min_desired_peers = 8
 # The period to wait once the number of rounds given by `idle_after_num_rounds` has completed. Default: 30 mins
 #network_discovery.idle_period = 1_800 # 30 * 60
 #  The minimum number of network discovery rounds to perform before idling (going to sleep). If there are less
@@ -334,6 +334,12 @@ database_url = "data/wallet/dht.db"
 # The maximum number of sync peer to select for each round. The selection strategy varies depending on the current state.
 # Default: 5
 #network_discovery.max_sync_peers = 5
+# The maximum number of peers we allow per round of sync. (Default: 500)
+#max_peers_to_sync_per_round = 500
+# Initial refresh sync peers delay period if more than one peer needs to be synced, handy when a wallet needs to
+# give preference to its base node peer connection. (Default: 15)
+#[serde(with = "serializers::seconds")]
+#initial_peer_sync_dalay = 15
 
 # Length of time to ban a peer if the peer misbehaves at the DHT-level. Default: 6 hrs
 #ban_duration = 21_600 # 6 * 60 * 60
--- a/common/config/presets/d_console_wallet.toml
+++ b/common/config/presets/d_console_wallet.toml
@@ -198,7 +198,7 @@ event_channel_size = 3500
 # CIDR for addresses allowed to enter into liveness check mode on the listener.
 #listener_liveness_allowlist_cidrs = []
 # Enables periodic socket-level liveness checks. Default: Disabled
-# listener_liveness_check_interval = 15
+# listener_self_liveness_check_interval = 15
 
 # User agent string for this node
 #user_agent = ""
@@ -206,6 +206,8 @@ event_channel_size = 3500
 # The maximum simultaneous comms RPC sessions allowed (default value = 100). Setting this to -1 will allow unlimited
 # sessions.
 #rpc_max_simultaneous_sessions = 100
+# The maximum comms RPC sessions allowed per peer (default value = 10).
+#rpc_max_sessions_per_peer = 10
 
 [wallet.p2p.transport]
 # -------------- Transport configuration --------------
@@ -335,11 +337,9 @@ network_discovery.min_desired_peers = 8
 # Default: 5
 #network_discovery.max_sync_peers = 5
 # The maximum number of peers we allow per round of sync. (Default: 500)
-#max_peers_to_sync_per_round = 500
-# Initial refresh sync peers delay period if more than one peer needs to be synced, handy when a wallet needs to
-# give preference to its base node peer connection. (Default: 15)
-#[serde(with = "serializers::seconds")]
-#initial_peer_sync_dalay = 15
+#network_discovery.max_peers_to_sync_per_round = 500
+# Initial refresh sync peers delay period, when a configured connection needs preference. (Default: Disabled)
+network_discovery.initial_peer_sync_delay = 15
 
 # Length of time to ban a peer if the peer misbehaves at the DHT-level. Default: 6 hrs
 #ban_duration = 21_600 # 6 * 60 * 60
--- a/comms/core/src/net_address/multiaddr_with_stats.rs
+++ b/comms/core/src/net_address/multiaddr_with_stats.rs
@@ -12,27 +12,29 @@ use std::{
 };
 
 use chrono::{NaiveDateTime, Utc};
+use log::trace;
 use multiaddr::Multiaddr;
 use serde::{Deserialize, Serialize};
 
 use crate::{peer_manager::PeerIdentityClaim, types::CommsPublicKey};
 
+const LOG_TARGET: &str = "comms::net_address::multiaddr_with_stats";
+
 const MAX_LATENCY_SAMPLE_COUNT: u32 = 100;
 const MAX_INITIAL_DIAL_TIME_SAMPLE_COUNT: u32 = 100;
-const HIGH_QUALITY_SCORE: i32 = 1000;
 
 #[derive(Debug, Eq, Clone, Deserialize, Serialize)]
 pub struct MultiaddrWithStats {
     address: Multiaddr,
     last_seen: Option<NaiveDateTime>,
     connection_attempts: u32,
-    avg_initial_dial_time: Duration,
+    avg_initial_dial_time: Option<Duration>,
     initial_dial_time_sample_count: u32,
-    avg_latency: Duration,
+    avg_latency: Option<Duration>,
     latency_sample_count: u32,
     last_attempted: Option<NaiveDateTime>,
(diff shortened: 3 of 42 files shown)


applications/minotari_app_grpc/proto/network.proto
```diff
@@ -70,7 +70,11 @@ message Address{
     bytes address =1;
     string last_seen = 2;
     uint32 connection_attempts = 3;
-    uint64 avg_latency = 5;
+    AverageLatency avg_latency = 5;
+}
+
+message AverageLatency {
+    uint64 latency = 1;
```
brianp, line 77: What's the benefit of the custom type here?

common/config/presets/d_console_wallet.toml
```diff
@@ -323,7 +323,7 @@ database_url = "data/wallet/dht.db"
 #network_discovery.enabled = true
 # A threshold for the minimum number of peers this node should ideally be aware of. If below this threshold a
 # more "aggressive" strategy is employed. Default: 50
```
brianp, line 327:
```suggestion
# more "aggressive" strategy is employed. Default: 8
```

comms/dht/src/connectivity/mod.rs
```diff
@@ -336,6 +339,27 @@ impl DhtConnectivity {
...
+        peers: I,
+    ) -> Result<(), DhtConnectivityError> {
+        self.handle_initial_delay().await;
+        self.connectivity.request_many_dials(peers).await?;
+        Ok(())
+    }
+
+    async fn handle_initial_delay(&mut self) {
+        if !self.first_round_completed {
+            self.first_round_completed = true;
```
brianp: Something about this feels like a code smell. The bool causes a delay in the first round. Can't we somehow apply that to the first round only, without requesting a function `request_many_dials_with_initial_delay` that implies a delay although we never actually expect the delay to occur again.

sdbondi: Yeah I agree, this delay should not be in this layer, the caller should call this when they mean to, e.g after a delay

comms/dht/src/connectivity/mod.rs
```diff
@@ -289,7 +291,7 @@ impl DhtConnectivity {
         match event {
             DhtEvent::NetworkDiscoveryPeersAdded(info) => {
```
sdbondi: We should just delay when this event is triggered. This event comes from peer sync, which is a state machine. The state machine should enter a "sleep" state for a configurable  period, then kick into peer discovery mode after that period, which will delay when the peer pools are refreshed

applications/minotari_app_grpc/proto/network.proto
hansieodendaal: It is to enable it as an option.

common/config/presets/d_console_wallet.toml
hansieodendaal: The default value in the struct is still set at 50; we change it here for the wallet.

comms/dht/src/connectivity/mod.rs
hansieodendaal: Let me try that

comms/dht/src/connectivity/mod.rs
hansieodendaal: > The bool causes a delay in the first round.

We do not know which code path is going to be in the first round. The other option is to call `handle_initial_delay` each time before `request_many_dials` or `refresh_neighbour_pool` is called.

However, I will investigate @sdbondi 's suggestion to add the delay in the state machine.

comms/core/src/net_address/multiaddr_with_stats.rs
```diff
@@ -184,27 +206,27 @@ impl MultiaddrWithStats {
         self.clone().address
     }
 
-    fn calculate_quality_score(&self) -> i32 {
+    fn calculate_quality_score(&self) -> Option<i32> {
         // If we have never seen or attempted the peer, we start with a high score to ensure that
         if self.last_seen.is_none() && self.last_attempted.is_none() {
-            return HIGH_QUALITY_SCORE;
+            return None;
```
SWvheerden, line 221: We had this discussion in the past. 
For overall network health, you need to try dial an address at least once.

hansieodendaal: Every node will still be tried and can easily be seen when running a base node and performing the `list-peers` command every minute or so; a high-quality score is not needed for that as now  `None` means it has not been tried and `Some(0)` means it is not reachable.

comms/core/src/net_address/mutliaddresses_with_stats.rs
```diff
@@ -354,15 +371,20 @@ mod test {
...
         );
-        net_addresses.merge(&other);
-        assert!(!net_addresses.contains(&net_address1));
-        assert!(net_addresses.contains(&net_address12));
+        net_addresses.merge(&address_12);
+        println!("net_address12: {:?}", address_12.addresses[0]);
+        println!(
+            "net_address1:  {:?}",
+            net_addresses.find_address_mut(&net_address1).unwrap()
+        );
```
sdbondi: Remove?

comms/dht/src/network_discovery/state_machine.rs
```diff
@@ -196,7 +196,12 @@ impl DhtNetworkDiscovery {
     async fn get_next_event(&mut self, state: &mut State) -> StateEvent {
         use State::{Discovering, Initializing, OnConnect, Ready, Waiting};
         match state {
-            Initializing => self::Initializing::new(&mut self.context).next_event().await,
+            Initializing => {
+                let initial_peer_sync_delay = self.config().network_discovery.initial_peer_sync_delay;
```
sdbondi: You dont need to pass in a separate delay. The state can access the config using context

comms/dht/src/network_discovery/initializing.rs
```diff
@@ -53,6 +57,13 @@ impl<'a> Initializing<'a> {
             }
         }
 
+        // Initial discovery and refresh sync peers delay period, when a configured connection needs preference,
+        // usually needed for the wallet to connect to its own base node first.
+        if let Some(delay) = self.initial_peer_sync_delay {
+            tokio::time::sleep(delay).await;
+            debug!(target: LOG_TARGET, "Discovery starting after delayed for {:.0?}", delay);
+        }
```
sdbondi:
```suggestion
        if let Some(delay) = self.context.config.network_discovery.initial_peer_sync_delay {
            tokio::time::sleep(delay).await;
            debug!(target: LOG_TARGET, "Discovery starting after delayed for {:.0?}", delay);
        }
```

comms/dht/src/peer_validator.rs
```diff
@@ -73,9 +76,23 @@ impl<'a> PeerValidator<'a> {
             });
         }
 
+        if let Some(existing) = &existing_peer {
```
sdbondi, line 79: This isn't something that can happen (bar some peerdb coding error). The caller must provide the correct existing peer and loads the peer by their authenticated public key. So this is checking that the developer has done things correctly rather than a legitimate peer validation error. I would say the check is not needed, alternatively, you could comment that if this occurs then there is a coding bug, and perhaps prefix the error text with `BUG: `.

comms/dht/src/peer_validator.rs
```diff
@@ -73,9 +76,23 @@ impl<'a> PeerValidator<'a> {
...
+            if existing.public_key != new_peer.public_key {
+                return Err(DhtPeerValidatorError::NewAndExistingMismatch {
+                    existing: format!("'{}' / '{}'", existing.node_id.to_hex(), existing.public_key.to_hex()),
```
sdbondi: To hex strings (2 allocations) are not required. Can do alloc free display with:

```suggestion
                    existing: format!("'{}' / '{}'", existing.node_id, existing.public_key),
```

comms/dht/src/peer_validator.rs
```diff
@@ -73,9 +76,23 @@ impl<'a> PeerValidator<'a> {
...
+                    new: format!(
+                        "'{}' / '{}'",
+                        NodeId::from_public_key(&new_peer.public_key),
+                        new_peer.public_key.to_hex()
```
sdbondi:
```suggestion
                        new_peer.public_key
```

comms/dht/src/peer_validator.rs
```diff
@@ -98,7 +115,13 @@ impl<'a> PeerValidator<'a> {
             peer.update_addresses(&claim.addresses, &PeerAddressSource::FromDiscovery {
                 peer_identity_claim: claim.clone(),
             });
-            peer.addresses.mark_all_addresses_as_last_seen_now(&claim.addresses);
+            trace!(
+                target: LOG_TARGET,
+                "Peer '{}' / '{}' added with address(es) from claim: {:?}",
+                node_id_hex,
```
sdbondi:
```suggestion
                node_id,
```

Cloning NodeId (if necessary) is cheap - it should implement the Copy trait but that refactor was never done

comms/dht/src/rpc/service.rs
```diff
@@ -68,9 +68,27 @@ impl DhtRpcServiceImpl {
             let iter = peers
                 .into_iter()
                 .filter_map(|peer| {
-                    let peer_info =
+                    let mut peer_info =
                         UnvalidatedPeerInfo::from_peer_limited_claims(peer, max_claims, max_addresses_per_claim);
 
+                    // Filter out all identity claims with invalid signatures
+                    let count = peer_info.claims.len();
```
sdbondi: How does this happen? Peers that you are streaming are, by definition, valid. 

If this is happening, then we should be removing invalidated addresses from the peer db.
If this is an issue, happy to keep this here but we should add a TODO comment saying that we should remove invalidated peer signatures.

comms/core/src/net_address/multiaddr_with_stats.rs
sdbondi: > Every node will still be tried and can easily be seen when running a base node and performing the list-peers command every minute or so; a high-quality score is not needed for that as now None means it has not been tried and Some(0) means it is not reachable.

Makes sense. Please make sure that there is a comment on the field that describes the meaning of None and Some.

sdbondi (commented):
Looks good, some nice changes. 

Few comments

SWvheerden (commented):
Looking good

comms/dht/src/rpc/service.rs
hansieodendaal: This is more _defence-in-depth_ than anything else, closing a potential attack vector. I will remove it, but we should actually ban the peer for offences in `UnvalidatedPeerInfo::from_peer_limited_claims` as well as providing invalid signatures. What is your opinion?

PR #6286, opened 2024-04-18
fix: improve wallet connection response time

SWvheerden:
Description
---
Improve the time a wallet will detect that the base node its connecting to is up and running

Motivation and Context
---
The time between tries here was 3 secs previously, but an unrelated change to the reuse of time variables bumped this up to 90 secs. This means that even thou the wallet is technically connected to the base node, the wallet service will wait for up to 90 secs before confirm this is the case. This bumps it down to 5 secs. 

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
utACK

PR #6287, opened 2024-04-18
feat: updates the emoji ID API to be more idiomatic

AaronFeickert:
Description
---
Updates the emoji ID public API to be more idiomatic.

Motivation and Context
---
The emoji ID public API includes functionality to convert to and from types like `String` and `PublicKey`, but not idiomatically. This PR updates the API to use `From` and `FromStr` trait implementations where possible. It also makes some minor changes to checksum function signatures to avoid unnecessary vector cloning, and improves a test.

How Has This Been Tested?
---
Existing tests pass.

---
Check that the new API is, in fact, more idiomatic :)

While this doesn't affect consensus or existing databases, this public API change is technically breaking. For example, existing [documentation](https://www.tari.com/lessons/05_emoji_id) needs to be updated.

### DIFF
--- a/applications/minotari_app_utilities/src/utilities.rs
+++ b/applications/minotari_app_utilities/src/utilities.rs
@@ -47,8 +47,8 @@ pub fn setup_runtime() -> Result<Runtime, ExitError> {
 
 /// Returns a CommsPublicKey from either a emoji id or a public key
 pub fn parse_emoji_id_or_public_key(key: &str) -> Option<CommsPublicKey> {
-    EmojiId::from_emoji_string(&key.trim().replace('|', ""))
-        .map(|emoji_id| emoji_id.to_public_key())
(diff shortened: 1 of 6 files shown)


sdbondi (commented):
Generally when something is constructed from a string the `FromStr` trait is more  idiomatic. 

```rust

fn do_something(emoji_id: EmojiId) { ... }

do_something(emoji_str.parse()?);
// equivalent to 
do_something(EmojiId::from_str(emoji_str)?);
```

It may make sense to implement `From<PublicKey> for EmojiId` if it isn't already

SWvheerden (dismissed):

SWvheerden:
`FromStr ` is perhaps the better option here and not try_from

https://doc.rust-lang.org/std/str/trait.FromStr.html

sdbondi (dismissed):

SWvheerden (approved):

PR #6290, opened 2024-04-19
fix: wallet ffi type incorrect

SWvheerden:
Description
---
Fixes FFI type to hex, and away from vec<u8>

Motivation and Context
---
c does not support vec<u8>


brianp (approved):

hansieodendaal (approved):
utACK

PR #6305, opened 2024-04-23
fix: hot for removing breaking change in wallet, new release for nexnet v1.0.0-rc.8

SWvheerden:
Description
---
Removed breaking change of wallet db type
Release new nextnet version

Motivation and Context
--
Release v1.0.0-rc.7 accidentally included an unmarked breaking change. This change is for future ledger support in with saving the wallet type as software or ledger in the wallet db. This is a breaking change. This forces a wallet to be software as ledger is currently in dev and not yet functioning. 

How Has This Been Tested?
---
Manual

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3206,7 +3206,7 @@ dependencies = [
 
 [[package]]
 name = "minotari_app_grpc"
-version = "1.0.0-rc.7"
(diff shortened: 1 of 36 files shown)


hansieodendaal (approved):
utACK

PR #6306, opened 2024-04-24
fix: potential overflow of coinbase calc

SWvheerden:
Description
---
fix potential overflow of multiple coinbase grpc call

Motivation and Context
---

How Has This Been Tested?
---
tests

### DIFF
--- a/applications/minotari_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/minotari_node/src/grpc/base_node_grpc_server.rs
@@ -807,13 +807,21 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         let mut prev_coinbase_value = 0u128;
         for coinbase in &mut coinbases {
             cur_share_sum += u128::from(coinbase.value);
-            coinbase.value =
-                u64::try_from((cur_share_sum * reward) / total_shares - prev_coinbase_value).map_err(|_| {
-                    obscure_error_if_true(
+            coinbase.value = u64::try_from(
+                (cur_share_sum.saturating_mul(reward))
+                    .checked_div(total_shares)
+                    .ok_or(obscure_error_if_true(
                         report_error_flag,
-                        Status::internal("Single coinbase fees exceeded u64".to_string()),
-                    )
-                })?;
+                        Status::internal("total shares are zero".to_string()),
+                    ))? -
+                    prev_coinbase_value,
+            )
+            .map_err(|_| {
+                obscure_error_if_true(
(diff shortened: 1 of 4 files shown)


applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -788,27 +788,33 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+        let mut prev_coinbase_value = 0u128;
         for coinbase in &mut coinbases {
-            coinbase.value *= reward / total_shares;
-            if remainder > 0 {
-                coinbase.value += 1;
-                remainder -= 1;
-            }
+            cur_share_sum += u128::from(coinbase.value);
+            coinbase.value =
+                u64::try_from((cur_share_sum * reward) / total_shares - prev_coinbase_value).map_err(|_| {
```
AaronFeickert: Any issue with possible overflow in the multiplication, division by zero, or underflow in the subtraction?

integration_tests/tests/steps/node_steps.rs
```diff
@@ -851,18 +851,19 @@ async fn generate_block_with_2_as_single_request_coinbases(world: &mut TariWorld
...
-            num_6154266700 += 1;
+        if output.minimum_value_promise.as_u64() == 6154272109 {
+            num_6154272109 += 1;
         }
-        if output.minimum_value_promise.as_u64() == 12308533398 {
-            num_12308533398 += 1;
+        if output.minimum_value_promise.as_u64() == 12308544218 {
+            num_12308544218 += 1;
         }
     }
```
hansieodendaal, line 863: These values should equal the total block reward, should it not? We could assert that.

hansieodendaal (commented):
Looks good, one comment below.

applications/minotari_node/src/grpc/base_node_grpc_server.rs
hansieodendaal: We could test for that, and subdue future comments.

integration_tests/tests/steps/node_steps.rs
SWvheerden: It is checked below by the fact that the block is accepted as valid

applications/minotari_node/src/grpc/base_node_grpc_server.rs
SWvheerden: over flow should not be possible as its all u64, but I made them checked and saturating anyway

PR #6308, opened 2024-04-25
fix: reload old wallets

SWvheerden:
Description
---
This allows older wallets that dont have a type stored in the DB to load up as software wallets.

Motivation and Context
---
If this type is not stored, either the entire DB is corrupted or its an older wallet without a type stored. 
If the DB is corrupted it would very likely fail earlier, or if not later. But older correct wallets should not be blocked from booting up

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
utACK

PR #6313, opened 2024-04-30
chore: show smt is loading

SWvheerden:
Description
---
Show smt is loading into memory

Motivation and Context
---
atm when you start the node, it looks like it crahsed or hanged. 

How Has This Been Tested?
---
manual


AaronFeickert (approved):
utACK

PR #6314, opened 2024-04-30
feat: add sent at timestamp to chat messages

SWvheerden:
Description
---
Add sent at timestamp to messages. 

Motivation and Context
---
Because of the lack of a central time server, its hard to get accurate timestamp, the chat services uses the received at as the timestamp. 
But when the receiver gets messages through saf, it can get multiple at the same time. This fudles up the order completely. So we add sent_at timestamp to help with ordering of messages.


base_layer/chat_ffi/src/message.rs
```diff
@@ -332,6 +332,31 @@ pub unsafe extern "C" fn read_chat_message_stored_at(message: *mut Message, erro
...
+/// `message` - A pointer to a Message
+/// `error_out` - Pointer to an int which will be modified
+///
+/// ## Returns
+/// `c_ulonglong` - The stored_at timestamp, seconds since epoch. Returns 0 if message is null.
+///
+/// ## Safety
+/// `message` should be destroyed eventually
+#[no_mangle]
+pub unsafe extern "C" fn read_chat_message_sent_at(message: *mut Message, error_out: *mut c_int) -> c_ulonglong {
```
hansieodendaal, line 347: We should add this ffi read to 
```rust
    #[test]
    fn test_reading_message_timestamps() {
```

hansieodendaal (approved):
One small comment

utACK

base_layer/chat_ffi/src/message.rs
SWvheerden: we could, but its just reading an accessor function?

PR #6320, opened 2024-05-06
feat!: remove user agent config option

SWvheerden:
Description
---
Removes the user-controlled user_agent field from the config and makes this a coded option only. 

Motivation and Context
---
This field is used by the application to report its name to the network. By default for the desktop and mobile apps this option is filled in by the application and reports the name and version of the application. This is very helpful for the network to decern whom they are connection to. 

Leaving this option to user change, makes it very easy to remove this option on an individual bases. Currently Launchpad already does this and only reports like follows:
```
d59a25f5a6855360364af9c2c9 | 02b0d0e37aaded4eac745608fdd6c08ceb06209a4ef5af807560f8fb14b36477 | /onion3/5uatfvtqxoo553k5x7iuel7ezkl5y7c7zy7vhyd3fo24s2yfzepexfid:18141 | Outbound  | 8h 45m 34s   | Launchpad base node        | height: 39379, hnd: 1, ss: 2, rpc: 0 
e9f94287e9893c4934f89edce7 | 826eda735986db5e57b3fd0342f0d9feebacae9329dc40ee74abf20c1d21f46d | /ip4/54.78.252.60/tcp/18189                                            | Outbound  | 10h 31m 52s  | tari/basenode/1.0.0-rc.8   | height: 39378, hnd: 1, ss: 2, rpc: 0 
```

Although this does not stop a user from changing this field, it lifts the barrier to do this by requiring code changes. 

Breaking Changes
---
Removes config option.

### DIFF
--- a/applications/minotari_console_wallet/src/config.rs
+++ b/applications/minotari_console_wallet/src/config.rs
@@ -45,8 +45,6 @@ impl ApplicationConfig {
             peer_seeds: PeerSeedsConfig::load_from(cfg)?,
         };
 
-        config.wallet.p2p.user_agent = format!("tari/wallet/{}", consts::APP_VERSION_NUMBER);
-
         config.wallet.set_base_path(config.common.base_path());
         Ok(config)
     }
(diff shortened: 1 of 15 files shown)


hansieodendaal (approved):
utACK

PR #6321, opened 2024-05-06
fix: change mmproxy to select new monerod on error

SWvheerden:
Description
---
Make sure the MMProxy does select a new node on error

Motivation and Context
---
Its possible for the monerod to error out on SSL, but not send back and error. The MMProxy then gets this error and keeps retrying the node:
```
ERROR Method "getblocktemplate" failed handling request: MonerodRequestFailed(reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("xmr.grub.net")), port: Some(18089), path: "/json_rpc", query: None, fragment: None }, source: hyper::Error(Connect, Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 167772294, library: "SSL routines", function: "tls_post_process_server_certificate", reason: "certificate verify failed", file: "ssl/statem/statem_clnt.c", line: 1889 }]))) }, X509VerifyResult { code: 18, error: "self-signed certificate" })) })
```

On Any error the MMProxy will now select a new monerod from the list it has.

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/proxy.rs
+++ b/applications/minotari_merge_mining_proxy/src/proxy.rs
@@ -858,15 +858,24 @@ impl InnerService {
                     return Ok(monerod_resp.map(|json| json.to_string().into()));
                 }
 
-                let response = self.get_proxy_response(request, monerod_resp).await?;
-                debug!(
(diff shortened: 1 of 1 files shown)


hansieodendaal (approved):
This should do it.

utACK

PR #6335, opened 2024-05-13
fix: new monero release for mr support in monero

SWvheerden:
Description
---
New monero rs release with mr support

Motivation and Context
---
The latest monero rs release now includes the correct support for merge mining by the field not being a simple hash, but rather an encoded u64. This simply bumps the version of monero rs so that we can support multiple merged mined coins.


hansieodendaal (approved):
utACK

PR #6337, opened 2024-05-15
chore: refactor checksum mask initialization

AaronFeickert:
Description
---
Makes checksums more efficient by using lazy initialization.

Motivation and Context
---
Checksum computation requires a [mask](https://github.com/tari-project/tari/blob/9920916f891fbd26759f7cb4912701189b7579a1/base_layer/common_types/src/dammsum.rs#L50-L55) whose value is fixed. Currently, this mask is computed each time a checksum is computed or validated. This is inefficient. Additionally, if the underlying coefficient parameters were later changed in an invalid way, this computation could overflow.

This PR refactors mask computation to use lazy initialization. It also checks for overflow during initialization, adds a test for this, and cleans up documentation and comments.

How Has This Been Tested?
---
Existing tests pass. A new test passes.

---
Check that mask computation is still correct, properly implemented using lazy initialization, and tested.


SWvheerden (dismissed):
pending green ci

SWvheerden (approved):

PR #6339, opened 2024-05-15
fix!: change position calculation in MTparams for mergemining

SWvheerden:
Description
---
Changes the way we calculate the position of the hash from the Merkle tree params

Motivation and Context
---
Its possible to have the same proof for different sized merkle trees. Verification needs to verify the total size from the decoded merkle tree params

---
Ensure it's the same as here: https://github.com/SChernykh/p2pool/blob/2c549f61a46a5f4f0be7c631dff6e928f73a52e2/src/merkle.cpp#L341-L369


hansieodendaal (dismissed):
utACK

PR #6344, opened 2024-05-16
feat: generate script challenge on the ledger

brianp:
Description
---
This generates the entire challenge for the script signature on the ledger.
Move signature signing functions into respective handler files.
Move hash domains into the hashing file.

Motivation and Context
---
The total payload size is smaller than before, and can now be done in a single message. With no batching needed I've removed the script signature context struct on the ledger side to reduce complexity, remove persisted state, and free up a bit of memory.

How Has This Been Tested?
---
Manually with new wallets

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
+++ b/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
@@ -20,8 +20,8 @@ use zeroize::Zeroizing;
 
 use crate::{
     alloc::string::ToString,
-    hashing::DomainSeparatedConsensusHasher,
-    utils::{alpha_hasher, derive_from_bip32_key, get_key_from_canonical_bytes, TransactionHashDomain},
+    hashing::{DomainSeparatedConsensusHasher, TransactionHashDomain},
+    utils::{alpha_hasher, derive_from_bip32_key, get_key_from_canonical_bytes},
     AppSW,
     KeyType,
     RESPONSE_VERSION,
--- a/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
+++ b/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
@@ -16,11 +16,12 @@ use tari_crypto::{
         RistrettoSecretKey,
     },
 };
+use tari_hashing::TransactionHashDomain;
 use zeroize::Zeroizing;
 
 use crate::{
     alloc::string::ToString,
-    hashing::{DomainSeparatedConsensusHasher, TransactionHashDomain},
+    hashing::DomainSeparatedConsensusHasher,
     utils::{alpha_hasher, derive_from_bip32_key, get_key_from_canonical_bytes},
     AppSW,
     KeyType,
--- a/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
+++ b/applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
@@ -3,97 +3,80 @@
 
 use alloc::format;
 
-use ledger_device_sdk::{io::Comm, ui::gadgets::SingleMessage};
-use tari_crypto::ristretto::{
-    pedersen::extended_commitment_factory::ExtendedPedersenCommitmentFactory,
-    RistrettoComAndPubSig,
-    RistrettoSecretKey,
+use blake2::Blake2b;
+use digest::consts::U64;
+use ledger_device_sdk::{io::Comm, random::Random, ui::gadgets::SingleMessage};
+use tari_crypto::{
+    commitment::HomomorphicCommitmentFactory,
+    keys::PublicKey,
+    ristretto::{
+        pedersen::{extended_commitment_factory::ExtendedPedersenCommitmentFactory, PedersenCommitment},
+        RistrettoComAndPubSig,
+        RistrettoPublicKey,
+        RistrettoSecretKey,
+    },
 };
 use zeroize::Zeroizing;
 
 use crate::{
     alloc::string::ToString,
-    utils::{alpha_hasher, derive_from_bip32_key, get_key_from_canonical_bytes},
+    hashing::DomainSeparatedConsensusHasher,
+    utils::{alpha_hasher, derive_from_bip32_key, get_key_from_canonical_bytes, TransactionHashDomain},
     AppSW,
     KeyType,
     RESPONSE_VERSION,
     STATIC_ALPHA_INDEX,
 };
 
-const MAX_TRANSACTION_LEN: usize = 312;
-pub struct ScriptSignatureCtx {
-    payload: [u8; MAX_TRANSACTION_LEN],
-    payload_len: usize,
-    account: u64,
-}
-
-// Implement constructor for TxInfo with default values
-impl ScriptSignatureCtx {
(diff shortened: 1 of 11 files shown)


applications/minotari_ledger_wallet/wallet/src/hashing.rs
```diff
@@ -6,7 +6,15 @@ use core::marker::PhantomData;
...
-use tari_crypto::hashing::DomainSeparation;
+use tari_crypto::{hash_domain, hashing::DomainSeparation};
+
+hash_domain!(LedgerHashDomain, "com.tari.minotari_ledger_wallet", 0);
+hash_domain!(
+    KeyManagerTransactionsHashDomain,
+    "com.tari.base_layer.core.transactions.key_manager",
+    1
+);
+hash_domain!(TransactionHashDomain, "com.tari.base_layer.core.transactions", 0);
```
hansieodendaal: Maybe we need to add a comment explaining that these hash domains are copies and must be the same as their counterpart in the larger tari project code base.

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
```diff
@@ -3,98 +3,81 @@
...
+        get_key_from_canonical_bytes::<RistrettoSecretKey>(&data[88..120])?.into();
+
+    let commitment: PedersenCommitment = get_key_from_canonical_bytes(&data[120..152])?;
+
+    let mut script_message = [0u8; 32];
+    script_message.clone_from_slice(&data[152..184]);
+
+    let r_a = derive_from_bip32_key(account, u32::random().into(), KeyType::Nonce)?;
+    let r_x = derive_from_bip32_key(account, u32::random().into(), KeyType::Nonce)?;
+    let r_y = derive_from_bip32_key(account, u32::random().into(), KeyType::Nonce)?;
```
hansieodendaal, line 65: Why not `u64::random()` or some other function that can provide a u64 random value?

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
```diff
@@ -3,98 +3,81 @@
...
-    }
+    let mut network_bytes = [0u8; 8];
+    network_bytes.clone_from_slice(&data[8..16]);
+    let network = u64::from_le_bytes(network_bytes);
 
-    // Set the account for the transaction
-    let mut account_bytes = [0u8; 8];
-    account_bytes.clone_from_slice(&signer_ctx.payload[0..8]);
-    signer_ctx.account = u64::from_le_bytes(account_bytes);
+    let mut txi_version_bytes = [0u8; 8];
```
hansieodendaal, line 43: NIT: What does `txi` stand for?

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
```diff
@@ -3,98 +3,81 @@
...
+    let mut account_bytes = [0u8; 8];
+    account_bytes.clone_from_slice(&data[0..8]);
+    let account = u64::from_le_bytes(account_bytes);
 
-    // If we expect more chunks, return
-    if more {
-        return Ok(());
-    }
+    let mut network_bytes = [0u8; 8];
+    network_bytes.clone_from_slice(&data[8..16]);
```
hansieodendaal, line 40: Why not send 1 byte accross, as network is only 1 byte?

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
```diff
@@ -3,98 +3,81 @@
...
+    let network = u64::from_le_bytes(network_bytes);
 
-    // Set the account for the transaction
-    let mut account_bytes = [0u8; 8];
-    account_bytes.clone_from_slice(&signer_ctx.payload[0..8]);
-    signer_ctx.account = u64::from_le_bytes(account_bytes);
+    let mut txi_version_bytes = [0u8; 8];
+    txi_version_bytes.clone_from_slice(&data[16..24]);
```
hansieodendaal, line 44: Ditto 1 byte

hansieodendaal (commented):
Looking nice, just a couple of comments

applications/minotari_ledger_wallet/wallet/src/hashing.rs
SWvheerden: I think we need to move those domains here: hashing/src/domains.rs
then we can point to it from both sides as a project

SWvheerden (commented):
This looks good, see if you can move the hashing to the hashing crate and that it still works. 

I don't know if you want to leave the optimizations of the data types to a later pr? 
We probably need to optimize all of them not just the script signature data

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
brianp: The ledger doesn't have a `u64::random()`. Only u8 and u32. This random is an "in house" random from their sdk. We don't have rand as a direct dependency right now, but rand_core is a derived dependent so we could use that if it's preferred. 

getrandom isn't supported on the ledger. I've just tried. Which is likely why they have their own implementation.

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
brianp: I'm guessing transaction input. It's the name of the var I copied from the key manager.

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
brianp: There is a ticket to fix this but it's because there's only helpers for changing u64's to strings at the moment. If the value had to become a string format then I just used functions that existed, and kept them as u64's.

brianp:
Moved the hash domains into hashing/src/. This does require making that create a `no_std` but that wasn't so difficult. 

I looked at using rand_core instead of ledger's random impl, but it's not no_std friendly. 

I'll save the byte optimizations for another PR.

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
hansieodendaal: You could pull two u32 random values and concatenate it to make up one u64 random value.

hansieodendaal:
@brianp, please see the comment about the u64 random^

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
SWvheerden: yeah, random is `std`, non std devices like the ledger need their own implementation of random

SWvheerden (approved):

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_signature.rs
SWvheerden: We can def add two u32 to make a u64, but deriving the u64 from a u32 via a hash, would also be fine. Both are secure

PR #6345, opened 2024-05-17
feat: remove chunking from rpc

hansieodendaal:
Description
---
Removed chunking from RPC as this does not improve latency or throughput and adds code complexity. 

Motivation and Context
---
See above.

How Has This Been Tested?
---
System-level tests were conducted syncing a fresh archival base node in a controlled environment over Gigabit wired LAN where only one other base node was connected. The blockchain had many full blocks and many UTXOs, as a result of a 24-hour stress test.

**Header sync**
No difference could be seen with header sync, as expected, due to the small message sizes.

image

**Block sync**
Blocks throughput increased during full block sync for both coins-split and normal transactions due to reduced latencies (data transfer time); up to 8% increased block throughput was measured when blocks were filled with normal transactions.

image

---
Review code changes and test data, or perform a similar comparative test.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


comms/core/src/protocol/rpc/server/mod.rs
```diff
@@ -749,12 +748,17 @@ where
...
                 if !message.status.is_ok() {
                     metrics::status_error_counter(&node_id, &protocol, message.status).inc();
                 }
-                stream::iter(ChunkedResponseIter::new(message))
+                proto::rpc::RpcResponse {
+                    request_id,
+                    status: message.status.as_u32(),
+                    flags: message.flags.bits().into(),
+                    payload: message.payload.to_vec(),
+                }
```
sdbondi: I didn't notice the `message.to_proto()` function before. That would be better to use than this

```suggestion
               message.to_proto()
```

comms/core/src/protocol/rpc/mod.rs
```diff
@@ -33,30 +33,12 @@ mod test;
 pub const RPC_MAX_FRAME_SIZE: usize = 3 * 1024 * 1024; // 3 MiB
 /// Maximum number of chunks into which a message can be broken up.
 const RPC_CHUNKING_MAX_CHUNKS: usize = 16; // 16 x 256 Kib = 4 MiB max combined message size
```
sdbondi: remove?

sdbondi (commented):
LGTM

I'm not sure if you intended to remove chunking from the client or not. It would be a breaking change to do so but that should be fine at this point.

SWvheerden (dismissed):

hansieodendaal:
Not sure what to do with the failing `response_too_big` unit test.  I get `ServerClosedRequest` instead of `RpcError::RequestFailed` at the first check.

sdbondi:
> Not sure what to do with the failing `response_too_big` unit test. I get `ServerClosedRequest` instead of `RpcError::RequestFailed` at the first check.

The problem was that the "payload too large" message wasn't being constructed properly and, without chunking, the max frame size should be >=  max response size.

I pushed fixes for these

sdbondi (dismissed):

hansieodendaal:
> The problem was that the "payload too large" message wasn't being constructed properly and, without chunking, the max frame size should be >= max response size.
> 
> I pushed fixes for these

Oops, did not see that. Working on it now.

sdbondi:
> > The problem was that the "payload too large" message wasn't being constructed properly and, without chunking, the max frame size should be >= max response size.
> > I pushed fixes for these
> 
> Oops, did not see that. Working on it now.

No problem :)  readded the fix

SWvheerden (approved):

PR #6348, opened 2024-05-20
feat: commitment proofs

### LINKED ISSUE #6282: Add commitment opening proofs
It was suggested by @stringhandler in #6240 to support an "ownership proof" that proves knowledge of the opening of a commitment. While this is what a [`CommitmentSignature`](https://github.com/tari-project/tari-crypto/blob/e755b26d4da6cef8eb00023a1fdcc923fc71870f/src/signatures/commitment_signature.rs#L53-L57) is for, its API is [outdated and not particularly friendly](https://github.com/tari-project/tari-crypto/issues/159). It was also suggested that such a proof also assert a minimum value, which could be useful in cases where the prover wishes to reveal a minimal (pun intended) amount of value information about the commitment.

One approach that handles both of these assertions is simply to use the Bulletproofs+ library to generate a minimum-value range proof. This will allow creation of such a proof using a transcript instantiated using a distinct label and arbitrary verifier message, which are important to avoid replay attacks.

It's very important to note that because of the structure of Tari transactions, neither the approach in #6240 nor this Bulletproofs+ suggestion actually prove control of value represented by a commitment. Depending on the structure of an output's script, it could be the case that both the sender and recipient in a transaction know the opening, but only the recipient knows the script private key required to authorize transfer of the corresponding value. Further, unless the verifier is careful to check for spend status, the commitment used in a proof may have already been spent, and is therefore no longer in the prover's control. Rather, such proofs only show _visibility_ into the value represented by the commitment, but not necessarily anything else.

AaronFeickert:
Description
---
Adds commitment proofs to the wallet.

Closes #6282.

Motivation and Context
---
It may be useful to assert knowledge of the opening of a commitment, and that it binds to at least a given value. Both can be accomplished through the use of a range proof, since the [Bulletproofs+](https://github.com/tari-project/bulletproofs-plus) range proving API supports the use of minimum value promises.

This PR adds a `CreateCommitmentProof` wallet command that produces such a proof. It accepts as arguments a commitment, message, and optional minimum value; and outputs a hex-encoded proof that binds to these arguments. The proof can be verified using the corresponding `VerifyCommitmentProof` wallet command. Both commands check that the commitment is unspent.

How Has This Been Tested?
---
**Tests are still in progress.**

---
Confirm that the proof is constructed and verified correctly using the range proving API.

Manually test:
- that a valid proof verifies against the correct commitment, message, and minimum value
- that a valid proof fails to verify against a different commitment
- that a valid proof fails to verify against a different message
- that a valid proof fails against a different minimum value
- that an invalid proof (for example, a truncated or random hex value) fails to verify
- that a proof cannot be created against a spent commitment
- that a proof fails to verify against a spent commitment


AaronFeickert:
This is very out of date. Closing in favor of a fresh PR to avoid rebase headaches.

PR #6350, opened 2024-05-22
chore: fix build by make ledger type optional

SWvheerden:
Description
---
Some packages are build for non std targets. The ledger libraries cannot be built for them. This PR makes sure that ledger stuff is optional for this.


leet4tari:
Run a build-all in local fork and all seems good.

hansieodendaal (dismissed):
utACK

PR #6353, opened 2024-05-23
feat!: new tari address scheme

SWvheerden:
Description
--- 
Changes the tari address scheme to support view keys and spend keys. 
Makes network and features human-readable

Motivation and Context
---
This allows Tari to support stealth one-sided transactions via Hardware devices. 
Makes features and network human readable. 

How Has This Been Tested?
---

PR #6358, opened 2024-05-27
feat: split message to have dedicated to and from fields

SWvheerden:
Description
---
split message to have dedicated to and from fields

Motivation and Context
---
This is needed for #6353 
When a chat client receives a message, it needs to know the complete address of the receiver, not just a part of it. This allows the user to get the complete address by saving what the sender said their address is

### DIFF
--- a/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
+++ b/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
@@ -2,8 +2,8 @@ DROP INDEX idx_messages_address;
 
 ALTER TABLE messages DROP address;
 
-ALTER TABLE messages ADD to_address BLOB NOT NULL;
-ALTER TABLE messages ADD from_address BLOB NOT NULL;
+ALTER TABLE messages ADD receiver_address BLOB NOT NULL;
+ALTER TABLE messages ADD sender_address BLOB NOT NULL;
 
-CREATE INDEX idx_messages_receiver_address ON messages (to_address);
-CREATE INDEX idx_messages_sender_address ON messages (from_address);
+CREATE INDEX idx_messages_receiver_address ON messages (receiver_address);
+CREATE INDEX idx_messages_sender_address ON messages (sender_address);
--- a/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
+++ b/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
@@ -5,5 +5,5 @@ ALTER TABLE messages DROP address;
 ALTER TABLE messages ADD to_address BLOB NOT NULL;
 ALTER TABLE messages ADD from_address BLOB NOT NULL;
 
-CREATE INDEX idx_messages_to_address ON messages (to_address);
-CREATE INDEX idx_messages_from_address ON messages (from_address);
+CREATE INDEX idx_messages_receiver_address ON messages (to_address);
+CREATE INDEX idx_messages_sender_address ON messages (from_address);
--- a/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
+++ b/base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
@@ -1,0 +1,9 @@
+DROP INDEX idx_messages_address;
+
+ALTER TABLE messages DROP address;
+
+ALTER TABLE messages ADD to_address BLOB NOT NULL;
+ALTER TABLE messages ADD from_address BLOB NOT NULL;
+
+CREATE INDEX idx_messages_to_address ON messages (to_address);
+CREATE INDEX idx_messages_from_address ON messages (from_address);
--- a/base_layer/contacts/proto/message.proto
+++ b/base_layer/contacts/proto/message.proto
@@ -7,9 +7,10 @@ package tari.contacts.chat;
 message Message {
   bytes body = 1;
   repeated MessageMetadata metadata = 2;
-  bytes address = 3;
-  DirectionEnum direction = 4;
-  bytes message_id = 5;
+  bytes to_address = 3;
+  bytes from_address = 4;
+  DirectionEnum direction = 5;
+  bytes message_id = 6;
 }
 
 enum DirectionEnum {
--- a/base_layer/contacts/proto/message.proto
+++ b/base_layer/contacts/proto/message.proto
@@ -7,8 +7,8 @@ package tari.contacts.chat;
 message Message {
   bytes body = 1;
   repeated MessageMetadata metadata = 2;
-  bytes to_address = 3;
-  bytes from_address = 4;
+  bytes receiver_address = 3;
+  bytes sender_address = 4;
   DirectionEnum direction = 5;
   bytes message_id = 6;
 }
--- a/base_layer/contacts/src/contacts_service/service.rs
+++ b/base_layer/contacts/src/contacts_service/service.rs
@@ -586,7 +586,7 @@ where T: ContactsBackend + 'static
         message: Message,
         source_public_key: CommsPublicKey,
     ) -> Result<(), ContactsServiceError> {
-        if source_public_key != *message.from_address.public_key() {
+        if source_public_key != *message.sender_address.public_key() {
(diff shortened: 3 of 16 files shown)


base_layer/contacts/src/contacts_service/service.rs
```diff
@@ -582,8 +582,10 @@ where T: ContactsBackend + 'static
         message: Message,
         source_public_key: CommsPublicKey,
     ) -> Result<(), ContactsServiceError> {
+        if source_public_key != *message.from_address.public_key() {
+            return Err(ContactsServiceError::MessageSourceDoesNotMatchOrigin);
+        }
```
brianp: Under the new address scheme, we can match the comms public key to the public view key. 
However, we can't validate the public spend key which makes up the second part of their complete address.

Is there any way a sender could use this as an attack vector by populating the `from_address` field with an address that contains a matching view key, and a non-matching spend key? I can't think of a way it would be useful, but mentioning it just to make sure.

base_layer/contacts/src/contacts_service/storage/types/messages.rs
```diff
@@ -83,14 +85,42 @@ impl MessagesSqlInsert {
...
+        page: i64,
+        conn: &mut SqliteConnection,
+    ) -> Result<Vec<MessagesSql>, ContactsServiceStorageError> {
+        Ok(messages::table
+            .filter(messages::from_address.eq(address))
+            .order(messages::stored_at.desc())
+            .offset(limit * page)
+            .limit(limit)
+            .load::<MessagesSql>(conn)?)
+    }
```
brianp: This interface makes it a bit difficult for ffi clients.

Previously you could get an entire message history of sent and received messages by querying the address of the contact (other party). Messages were distinguished by the `direction` to know if they were outbound or inbound. 

This new interface now requires two function calls, and a merging of the results, sorted by received dates to generate a timeline of messages.

I think the interface we're seeking isn't to find by `from` or `to` address independently, but to find by address where the address given is in *either* the `from` or `to` column. 
To put it plainly. It should still be a single function that checks if the address is in either field, but we don't actually care which one it belongs to.

We may also be able to remove the `direction` field as it becomes redundant if we're maintaining the to/from address and in all cases one of those fields will always be the current node. Direction becomes something derivable by which field our address is in.

brianp (commented):
It mostly looks good but I think some of the interface change goes in the wrong direction. It follows the pattern of providing readers for things, but then fails in practical usage.

base_layer/contacts/src/contacts_service/service.rs
SWvheerden: Comms_key == spend key, 
This key is super important to keep safe, as losing this you lose your utxos. 

But on the view key, its purely there for you to advertise your full address to your message contact as sender. From the contact service, the view key has no purpose or use. But if the sender where to send you a "fake" view key, they might not receive one-sided transactions that you where to send to him. So the sender has no incentive to lie about this and the receiver does not use the key here, so there is no attack vector here.

base_layer/contacts/src/contacts_service/storage/types/messages.rs
SWvheerden: I dont think these are even mapped to the FFI atm, 
Currently they still use `find_by_address`
This is now changed to do a `.filter(messages::from_address.eq(address).or(messages::to_address.eq(address)))`
So you can still get a complete history by just doing a simple lookup to an address.

brianp: ahhhh i thought the old method had been removed

brianp (dismissed):
LGTM

base_layer/contacts/src/contacts_service/types/message.rs
```diff
@@ -36,7 +36,8 @@ use crate::contacts_service::proto;
 pub struct Message {
     pub body: Vec<u8>,
     pub metadata: Vec<MessageMetadata>,
-    pub address: TariAddress,
+    pub to_address: TariAddress,
+    pub from_address: TariAddress,
```
hansieodendaal: I think this is more in line with the rest of the naming in the code base.
```suggestion
    pub receiver_address: TariAddress,
    pub sender_address: TariAddress,
```

base_layer/contacts/migrations/2024-05-27-145200_add_from_to_fields/up.sql
```diff
@@ -0,0 +1,9 @@
+DROP INDEX idx_messages_address;
+
+ALTER TABLE messages DROP address;
+
+ALTER TABLE messages ADD to_address BLOB NOT NULL;
+ALTER TABLE messages ADD from_address BLOB NOT NULL;
+
+CREATE INDEX idx_messages_to_address ON messages (to_address);
+CREATE INDEX idx_messages_from_address ON messages (from_address);
```
hansieodendaal:
```suggestion
CREATE INDEX idx_messages_receiver_address ON messages (to_address);
CREATE INDEX idx_messages_sender_address ON messages (from_address);
```

base_layer/contacts/proto/message.proto
```diff
@@ -7,9 +7,10 @@ package tari.contacts.chat;
 message Message {
   bytes body = 1;
   repeated MessageMetadata metadata = 2;
-  bytes address = 3;
-  DirectionEnum direction = 4;
-  bytes message_id = 5;
+  bytes to_address = 3;
+  bytes from_address = 4;
```
hansieodendaal:
```suggestion
  bytes receiver_address = 3;
  bytes sender_address = 4;
```

base_layer/contacts/src/contacts_service/storage/types/messages.rs
```diff
@@ -40,7 +40,8 @@ use crate::{
 #[diesel(table_name = messages)]
 #[diesel(primary_key(message_id))]
 pub struct MessagesSqlInsert {
-    pub address: Vec<u8>,
+    pub to_address: Vec<u8>,
+    pub from_address: Vec<u8>,
```
hansieodendaal: ditto `receiver` & `sender`

base_layer/contacts/src/contacts_service/storage/types/messages.rs
```diff
@@ -53,7 +54,8 @@ pub struct MessagesSqlInsert {
 #[diesel(table_name = messages)]
 #[diesel(primary_key(message_id))]
 pub struct MessagesSql {
-    pub address: Vec<u8>,
+    pub to_address: Vec<u8>,
+    pub from_address: Vec<u8>,
```
hansieodendaal: ditto `receiver` & `sender`

base_layer/contacts/src/contacts_service/storage/types/messages.rs
```diff
@@ -134,14 +164,20 @@ impl TryFrom<MessagesSql> for Message {
 
     #[allow(clippy::cast_sign_loss)]
     fn try_from(o: MessagesSql) -> Result<Self, Self::Error> {
-        let address = TariAddress::from_bytes(&o.address).map_err(|_| ContactsServiceStorageError::ConversionError)?;
+        let to_address =
+            TariAddress::from_bytes(&o.to_address).map_err(|_| ContactsServiceStorageError::ConversionError)?;
+        let from_address =
+            TariAddress::from_bytes(&o.from_address).map_err(|_| ContactsServiceStorageError::ConversionError)?;
```
hansieodendaal: ditto `receiver` & `sender`

base_layer/contacts/src/contacts_service/storage/types/messages.rs
```diff
@@ -164,9 +198,9 @@ impl TryFrom<Message> for MessagesSqlInsert {
 
     fn try_from(o: Message) -> Result<Self, Self::Error> {
         let metadata = serde_json::to_string(&o.metadata).map_err(|_| ContactsServiceStorageError::ConversionError)?;
-
         Ok(Self {
-            address: o.address.to_bytes().to_vec(),
+            to_address: o.to_address.to_bytes().to_vec(),
+            from_address: o.from_address.to_bytes().to_vec(),
```
hansieodendaal: ditto `receiver` & `sender`

hansieodendaal (commented):
Looks good, just some comments about the naming convention.

PR #6360, opened 2024-05-28
fix: atomic swap

SWvheerden:
Description
---
Fixes atomic swap

Motivation and Context
---
Atomic swap should use the existing on chain rangeproof and not generate a new one.


hansieodendaal (approved):
utACK

PR #6361, opened 2024-05-30
feat!: export unblinded outputs

hansieodendaal:
Description
---
Exported unblinded UTXOs. In a previous PR the ability to export unblinded UTXOs were removed; this PR reinstates that ability. Also updated the wallet FFI to include all related methods.

Motivation and Context
---
See above.

How Has This Been Tested?
---
System-level testing
Updated the wallet FFI unit test (`test_import_external_utxo`)

---
Export utxos

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: Wallet FFI interface change to import UTXOs.

### DIFF
--- a/applications/minotari_app_grpc/proto/wallet.proto
+++ b/applications/minotari_app_grpc/proto/wallet.proto
@@ -265,6 +265,7 @@ message CoinSplitResponse {
 
 message ImportUtxosRequest {
     repeated UnblindedOutput outputs = 1;
+    repeated RangeProof proofs = 2;
 }
 
 message ImportUtxosResponse {
--- a/applications/minotari_app_grpc/proto/wallet.proto
+++ b/applications/minotari_app_grpc/proto/wallet.proto
@@ -265,7 +265,6 @@ message CoinSplitResponse {
 
 message ImportUtxosRequest {
     repeated UnblindedOutput outputs = 1;
-    repeated RangeProof proofs = 2;
 }
 
 message ImportUtxosResponse {
--- a/applications/minotari_console_wallet/src/automation/commands.rs
+++ b/applications/minotari_console_wallet/src/automation/commands.rs
@@ -1189,6 +1189,12 @@ fn write_utxos_to_csv_file(
             },
         )
             .map_err(|e| CommandError::CSVFile(e.to_string()))?;
+        debug!(
+            target: LOG_TARGET,
+            "UTXO {} exported: {:?}",
+            i + 1,
+            utxo
+        );
     }
     Ok(())
 }
--- a/applications/minotari_console_wallet/src/automation/commands.rs
+++ b/applications/minotari_console_wallet/src/automation/commands.rs
@@ -54,7 +54,7 @@ use tari_common_types::{
     emoji::EmojiId,
     tari_address::TariAddress,
     transaction::TxId,
-    types::{Commitment, FixedHash, PublicKey, Signature},
+    types::{Commitment, FixedHash, PublicKey, RangeProof, Signature},
 };
 use tari_comms::{
     connectivity::{ConnectivityEvent, ConnectivityRequester},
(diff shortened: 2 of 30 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -113,7 +118,7 @@ impl<TBackend, TWalletConnectivity, TKeyManagerInterface>
 where
     TBackend: OutputManagerBackend + 'static,
     TWalletConnectivity: WalletConnectivityInterface,
-    TKeyManagerInterface: TransactionKeyManagerInterface,
+    TKeyManagerInterface: TransactionKeyManagerInterface + SecretTransactionKeyManagerInterface,
```
SWvheerden:
```suggestion
    TKeyManagerInterface: TransactionKeyManagerInterface ,
```

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -803,17 +803,29 @@ pub async fn command_runner(
             },
             ExportUtxos(args) => match output_service.get_unspent_outputs().await {
```
SWvheerden, line 818: I would prefer it if we add new methods, rather than changing the existing to expose secrets. 

Or make the secret key expose part optional.

applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -264,6 +264,7 @@ message CoinSplitResponse {
 
 message ImportUtxosRequest {
     repeated UnblindedOutput outputs = 1;
+    repeated RangeProof proofs = 2;
```
SWvheerden: Rather add the proof to the unblinded output. Makes more sense there
Its a vbec<u8> so it could be empty.

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -803,17 +803,31 @@ pub async fn command_runner(
...
                     } else {
-                        for (i, utxo) in utxos.iter().enumerate() {
-                            println!("{}. Value: {} {}", i + 1, utxo.0.value, utxo.0.features);
+                        for (i, utxo) in unblinded_utxos.iter().enumerate() {
+                            println!(
+                                "{}. Value: {}, Spending Key: {:?}, Script Key: {:?}, Features: {}",
+                                i + 1,
+                                utxo.0.value,
+                                utxo.0.spending_key.to_hex(),
+                                utxo.0.script_private_key.to_hex(),
```
SWvheerden: this should also be covered the `args.with_private_keys`

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -851,17 +865,31 @@ pub async fn command_runner(
...
```
SWvheerden: Same here only print if `args.with_private_keys`

base_layer/wallet/src/wallet.rs
```diff
@@ -521,8 +521,8 @@ where
         covenant: Covenant,
         encrypted_data: EncryptedData,
         minimum_value_promise: MicroMinotari,
+        range_proof: Option<RangeProof>,
     ) -> Result<TxId, WalletError> {
-        // lets import the private keys
         let unblinded_output = UnblindedOutput::new_current_version(
```
SWvheerden, line 526: Add the rangeproof to `UnblindedOutput`

SWvheerden: The fact that we need to carry this data outdie is going to cause problems. The RangeProof is a crucial part of the output can cannot change

base_layer/wallet_ffi/src/lib.rs
```diff
@@ -204,15 +204,18 @@ pub type TariPublicKey = tari_common_types::types::PublicKey;
 pub type TariWalletAddress = tari_common_types::tari_address::TariAddress;
 pub type TariNodeId = tari_comms::peer_manager::NodeId;
 pub type TariPrivateKey = tari_common_types::types::PrivateKey;
+pub type TariRangeProof = tari_common_types::types::RangeProof;
```
SWvheerden: nit: I would not even create this type, 
Its a wrapped vec<u8>, I would just just the bytevector class. 

But combine RP in unblinded output and see where that leads us

SWvheerden (commented):
Looking good, 
I think we just need to add RangeProof to UnblindedOutput. Its going to be dangerous to keep them decoupled if each output has a unique RP that we have to keep separate. 

and just make sure that private key bool is honoured in all branches of the CLI command.

hansieodendaal:
Yeah, I thought about adding RangeProof to UnblindedOutput, and maybe it is better that way. Will change.

SWvheerden (dismissed):

SWvheerden (approved):

PR #6362, opened 2024-05-30
feat: make console wallet ledger default

SWvheerden:
Description
---
Make ledger support enabled by default


hansieodendaal (approved):
utACK

PR #6364, opened 2024-05-31
feat: limit script max size

SWvheerden:
Description
---
Current script max size is 2kb, this reduces it down to 512 bytes.


hansieodendaal (approved):
utACK

PR #6367, opened 2024-06-03
feat: remove ablity to send 1-sided tx

SWvheerden:
Description
---
Removes the ability to send one-sided transactions from the CW, preferring to only send stealth one-sided

Motivation and Context
---
There is no benefit for sending normal one-sided, only negatives in regard to privacy


hansieodendaal (commented):
What about the one-sided code in the transaction service and scanning service? Maybe it should be removed in this PR as well.

SWvheerden:
I want to keep that in for now, until we know exactly how we want to work with exchanges, they want want to use it. So for now I am only removing the front end calling

hansieodendaal (dismissed):
utACK

PR #6369, opened 2024-06-04
[wip]fix: chat sending using wrong address

SWvheerden:
Description
---
Fixes chat using the wrong destination, and this getting the peer banned.

### DIFF
--- a/base_layer/contacts/src/chat_client/src/client.rs
+++ b/base_layer/contacts/src/chat_client/src/client.rs
@@ -228,7 +228,7 @@ impl ChatClient for Client {
     async fn send_read_receipt(&self, message: Message) -> Result<(), Error> {
         if let Some(mut contacts_service) = self.contacts.clone() {
             contacts_service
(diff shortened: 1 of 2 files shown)


base_layer/contacts/src/contacts_service/service.rs
```diff
@@ -620,7 +620,7 @@ where T: ContactsBackend + 'static
         &mut self,
         message: &Message,
     ) -> Result<(), ContactsServiceError> {
-        let address = &message.sender_address;
+        let address = &message.receiver_address;
```
brianp: This doesn't seem correct to me.

The message is a received message, so we should still be the receiver address in this case. Meaning the confirmation needs to be sent to the `sender`. If these addresses are mixed up, it was likely at the source of the message send.

brianp (changes requested):

brianp (approved):

SWvheerden:
Dont think this will solve the issue, closing for now

PR #6370, opened 2024-06-07
feat(miner): add SHA P2Pool mining related configuration and changes

ksrichard:
Description
---
There is a new decentralized P2Pool implementation in progress for SHA3 based mining on Tari.
To be able to connect and use this new implementation we need to have some new changes introduced in this PR.
Overall to make the usage of the new mining pool as easy as possible, I introduced a new configuration in the miner called `sha_p2pool_enabled` and if it's set to true and `base_node_grpc_address` config is set to the address of the `p2pool node` then the mining starts on the decentralized pool.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/minotari_miner/src/run_miner.rs
```diff
@@ -330,12 +416,71 @@ async fn mining_cycle(
...
+    Ok(GetNewBlockResponse{
+        block,
+        target_difficulty: block_result.target_difficulty,
+    })
+}
+
+async fn submit_block(
+    config: &MinerConfig,
+    base_node_client: &mut BaseNodeGrpcClient,
+    sha_p2pool_client: Arc<Mutex<Option<ShaP2PoolGrpcClient>>>,
```
sdbondi: I suspect you don't need a mutex.

```suggestion
    sha_p2pool_client: Option<&mut ShaP2PoolGrpcClient>,
```

Call with:
```rust
submit_block(base_node_client, sha_p2pool_client.as_mut(), ....)
```

ksrichard: okay makes sense, just pushed

SWvheerden (commented):
uACK

This looks good

sdbondi (approved):
utACK

PR #6372, opened 2024-06-18
feat!: change tar address to use base58 and not hex

SWvheerden:
Description
---
Remove hex display from tari address
Use Base 58 encoding instead. 
Keeps the network and features human readable

Motivation and Context
---
reduces the display size from 134 chars, down to 92

How Has This Been Tested?
---
unit tests and manual

### DIFF
--- a/base_layer/common_types/src/tari_address/dual_address.rs
+++ b/base_layer/common_types/src/tari_address/dual_address.rs
@@ -157,7 +157,9 @@ impl DualAddress {
 
     /// Construct Tari Address from Base58
     pub fn from_base58(hex_str: &str) -> Result<Self, TariAddressError> {
-        if hex_str.len() != 91 {
+        // Due to the byte length, it can be encoded as 90 or 91
+        if hex_str.len() != 90 && hex_str.len() != 91 {
+            dbg!(hex_str.len());
             return Err(TariAddressError::InvalidSize);
         }
         let (first, rest) = hex_str.split_at(2);
--- a/base_layer/common_types/src/tari_address/dual_address.rs
+++ b/base_layer/common_types/src/tari_address/dual_address.rs
@@ -159,7 +159,6 @@ impl DualAddress {
     pub fn from_base58(hex_str: &str) -> Result<Self, TariAddressError> {
         // Due to the byte length, it can be encoded as 90 or 91
         if hex_str.len() != 90 && hex_str.len() != 91 {
-            dbg!(hex_str.len());
             return Err(TariAddressError::InvalidSize);
         }
         let (first, rest) = hex_str.split_at(2);
--- a/base_layer/common_types/src/tari_address/dual_address.rs
+++ b/base_layer/common_types/src/tari_address/dual_address.rs
@@ -25,6 +25,7 @@ use std::convert::TryFrom;
 use serde::{Deserialize, Serialize};
 use tari_common::configuration::Network;
 use tari_crypto::tari_utilities::ByteArray;
+use tari_utilities::hex::{from_hex, Hex};
 
 use crate::{
     dammsum::{compute_checksum, validate_checksum},
@@ -187,6 +188,18 @@ impl DualAddress {
         network.push_str(&rest);
         network
     }
+
+    /// Convert Tari dual Address to hex
+    pub fn to_hex(&self) -> String {
+        let buf = self.to_bytes();
+        buf.to_hex()
+    }
+
+    /// Creates Tari dual Address from hex
+    pub fn from_hex(hex_str: &str) -> Result<DualAddress, TariAddressError> {
+        let buf = from_hex(hex_str).map_err(|_| TariAddressError::CannotRecoverPublicKey)?;
+        DualAddress::from_bytes(buf.as_slice())
+    }
 }
 
 #[cfg(test)]
@@ -296,6 +309,8 @@ mod test {
 
         let buff = address.to_bytes();
         let base58 = address.to_base58();
+        let hex = address.to_hex();
+        let emoji = address.to_emoji_string();
 
         let address_buff = DualAddress::from_bytes(&buff).unwrap();
         assert_eq!(address_buff.public_spend_key(), address.public_spend_key());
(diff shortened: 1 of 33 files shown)


base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,32 @@ impl DualAddress {
...
 
-    /// Construct Tari Address from hex
-    pub fn from_hex(hex_str: &str) -> Result<Self, TariAddressError> {
-        let buf = from_hex(hex_str).map_err(|_| TariAddressError::CannotRecoverPublicKey)?;
-        Self::from_bytes(buf.as_slice())
+    /// Construct Tari Address from Base58
+    pub fn from_base58(hex_str: &str) -> Result<Self, TariAddressError> {
+        if hex_str.len() < 3 {
+            return Err(TariAddressError::InvalidSize);
+        }
```
hansieodendaal: I think it would make sense to compare the actual length that `Self::from_bytes` expects here

base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,32 @@ impl DualAddress {
...
+        let (first, rest) = hex_str.split_at(2);
+        let (network, features) = first.split_at(1);
+        let mut network = bs58::decode(network).into_vec().map_err(|_| TariAddressError::CannotRecoverNetwork)?;
+        let mut features = bs58::decode(features).into_vec().map_err(|_| TariAddressError::CannotRecoverFeature)?;
+        if rest.is_empty(){
+            return Err(TariAddressError::CannotRecoverPublicKey);
+        }
```
hansieodendaal: This test would then be unnecessary

base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,32 @@ impl DualAddress {
...
+        let mut rest = bs58::decode(rest).into_vec().map_err(|_| TariAddressError::CannotRecoverPublicKey)?;
```
hansieodendaal: Maybe `public_key` instead or `rest`?

base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,32 @@ impl DualAddress {
...
+        network.append(&mut features);network.append(&mut rest);
+        Self::from_bytes(network.as_slice())
```
hansieodendaal: I think it would read better to create a new mutable buffer with a better name to append network, features and public_key to

base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,32 @@ impl DualAddress {
...
-        buf.to_hex()
+    /// Convert Tari Address to Base58 string
+    pub fn to_base58(&self) -> String {
+        let bytes = self.to_bytes();
+        let mut network = bs58::encode(&bytes[0..1]).into_string();
+        let features = bs58::encode(&bytes[1..2].to_vec()).into_string();
+        let rest = bs58::encode(&bytes[2..]).into_string();
+        network.push_str(&features);
+        network.push_str(&rest);
+        network
```
hansieodendaal, line 189: ditto new mutable buffer with a better name

base_layer/common_types/src/tari_address/mod.rs
```diff
@@ -214,15 +217,31 @@ impl TariAddress {
     }
 
     /// Construct Tari Address from hex
-    pub fn from_hex(hex_str: &str) -> Result<TariAddress, TariAddressError> {
-        let buf = from_hex(hex_str).map_err(|_| TariAddressError::CannotRecoverPublicKey)?;
-        TariAddress::from_bytes(buf.as_slice())
+    pub fn from_base58(hex_str: &str) -> Result<TariAddress, TariAddressError> {
+        if hex_str.len() < 3 {
+            return Err(TariAddressError::InvalidSize);
+        }
```
hansieodendaal: ditto length check

Same comments below as previous

base_layer/common_types/src/tari_address/single_address.rs
```diff
@@ -139,16 +138,32 @@ impl SingleAddress {
         buf
     }
 
-    /// Construct Tari Address from hex
-    pub fn from_hex(hex_str: &str) -> Result<Self, TariAddressError> {
-        let buf = from_hex(hex_str).map_err(|_| TariAddressError::CannotRecoverPublicKey)?;
-        Self::from_bytes(buf.as_slice())
+    /// Construct Tari Address from Base58
```
hansieodendaal, line 142: Same comments below as previous

base_layer/common_types/src/tari_address/dual_address.rs
SWvheerden: I will do something similar, but bytes expect a byte length, this is a string.

base_layer/common_types/src/tari_address/dual_address.rs
SWvheerden: I called it rest as its (public_key or public keys) + checksum. For the dual it is public keys only thou

hansieodendaal:
Some clippy and compile issues?

base_layer/common_types/src/tari_address/dual_address.rs
```diff
@@ -156,16 +155,38 @@ impl DualAddress {
...
```
hansieodendaal, line 189: **NIT**: I still think this reads better
```suggestion
        let bytes = self.to_bytes();
        let network = bs58::encode(&bytes[0..1]).into_string();
        let features = bs58::encode(&bytes[1..2].to_vec()).into_string();
        let rest = bs58::encode(&bytes[2..]).into_string();
        let mut address = "".into_string();
        address.push_str(&network);
        address.push_str(&features);
        address.push_str(&rest);
        address
```

hansieodendaal (dismissed):
utACK

hansieodendaal:
Some cucumber tests are still failing

hansieodendaal (dismissed):
utACK

PR #6374, opened 2024-06-19
fix: improve key scanning

SWvheerden:
Description
---
Improves utxo scanning with various fixes

Motivation and Context
---
Scanning for new outputs on the blockchain can be very slow.

How Has This Been Tested?
---
Manual

### DIFF
--- a/base_layer/core/src/transactions/key_manager/inner.rs
+++ b/base_layer/core/src/transactions/key_manager/inner.rs
@@ -65,7 +65,7 @@ use tari_utilities::{hex::Hex, ByteArray};
 use tokio::sync::RwLock;
 
 const LOG_TARGET: &str = "c::bn::key_manager::key_manager_service";
-const KEY_MANAGER_MAX_SEARCH_DEPTH: u64 = 1_000_000;
+const TRANSACTION_KEY_MANAGER_MAX_SEARCH_DEPTH: u64 = 1_000_000;
 
 use crate::{
     common::ConfidentialOutputHasher,
@@ -343,7 +343,7 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
 
         let current_index = km.key_index();
 
-        for i in 0u64..current_index + KEY_MANAGER_MAX_SEARCH_DEPTH {
+        for i in 0u64..current_index + TRANSACTION_KEY_MANAGER_MAX_SEARCH_DEPTH {
             let public_key = PublicKey::from_secret_key(&km.derive_key(i)?.key);
             if public_key == *key {
                 trace!(target: LOG_TARGET, "Key found in {} Key Chain at index {}", branch, i);
@@ -366,7 +366,7 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         let current_index = km.key_index();
 
         // its most likely that the key is close to the current index, so we start searching from the current index
-        for i in 0u64..KEY_MANAGER_MAX_SEARCH_DEPTH {
+        for i in 0u64..TRANSACTION_KEY_MANAGER_MAX_SEARCH_DEPTH {
             let index = current_index + i;
             let private_key = &km.derive_key(index)?.key;
             if private_key == key {
--- a/base_layer/core/src/transactions/key_manager/inner.rs
+++ b/base_layer/core/src/transactions/key_manager/inner.rs
@@ -242,9 +242,11 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
             KeyId::Derived { branch, label, index } => {
                 let public_alpha = match &self.wallet_type {
                     WalletType::Software(_k, pk) => pk,
-                    WalletType::Ledger(ledger) => ledger.pubkey.as_ref().ok_or(KeyManagerServiceError::LedgerError(
-                        "Key manager set to use ledger, ledger alpha public key missing".to_string(),
-                    ))?,
+                    WalletType::Ledger(ledger) => {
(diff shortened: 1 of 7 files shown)


base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -363,11 +365,21 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
 
         let current_index = km.key_index();
 
-        for i in 0u64..current_index + KEY_MANAGER_MAX_SEARCH_DEPTH {
-            let private_key = &km.derive_key(i)?.key;
+        // its most likely that the key is close to the current index, so we start searching from the current index
+        for i in 0u64..KEY_MANAGER_MAX_SEARCH_DEPTH {
```
hansieodendaal:
```suggestion
        for i in 0u64..KEY_MANAGER_MAX_SEARCH_DEPTH - current_index {
```

hansieodendaal: Why is `KEY_MANAGER_MAX_SEARCH_DEPTH` defined twice in the code? Can it not be defined once and included in the other location?

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -1218,12 +1230,7 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         self.crypto_factories
             .range_proof
             .verify_mask(output.commitment(), &private_key, value.into())?;
-        // Detect the branch we need to scan on for the key.
-        let branch = if output.is_coinbase() {
-            TransactionKeyManagerBranch::Coinbase.get_branch_key()
-        } else {
-            TransactionKeyManagerBranch::CommitmentMask.get_branch_key()
-        };
+        let branch = TransactionKeyManagerBranch::CommitmentMask.get_branch_key();
```
hansieodendaal, line 1242: Interesting, did we miss detected coinbase output keys? Otherwise, this is a breaking change to be reflected in the PR title and description as I see `TransactionKeyManagerBranch::Coinbase` has been removed.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -200,11 +204,16 @@ where
         known_scripts: &[KnownOneSidedPaymentScript],
     ) -> Result<Option<(ExecutionStack, TariKeyId)>, OutputManagerError> {
         let (input_data, script_key) = if script == &script!(Nop) {
-            // This is a nop, so we can just create a new key an create the input stack.
-            let key = KeyId::Derived {
-                branch: TransactionKeyManagerBranch::CommitmentMask.get_branch_key(),
-                label: TransactionKeyManagerLabel::ScriptKey.get_branch_key(),
-                index: spending_key.managed_index().unwrap(),
+            // This is a nop, so we can just create a new key for the input stack.
```
hansieodendaal, line 205:
```suggestion
            // This is a nop, so we can just create a new key for the input stack if the key does not exist.
```

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -265,37 +274,26 @@ where
     /// seen so far.
     async fn update_outputs_script_private_key_and_update_key_manager_index(
         &mut self,
-        output: &mut WalletOutput,
+        output: &WalletOutput,
```
hansieodendaal: The function has changed, so the comment and name must reflect what is happening inside.

hansieodendaal: Good catch; the wallet output already has the correct info.

hansieodendaal (commented):
Nice. Could you see any improvement with this PR? Some comments below.

base_layer/core/src/transactions/key_manager/inner.rs
SWvheerden: No this is incorrect. `i` us used as the counter, we should not subtract the current index from the counter

base_layer/core/src/transactions/key_manager/inner.rs
SWvheerden: `TransactionKeyManagerBranch::Coinbase` is never used, except in tests

base_layer/core/src/transactions/key_manager/inner.rs
SWvheerden: Different classes. They happen to be the same number. I renamed the one to remove this confusion.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
SWvheerden: The key does not exist, always. 
There is just an edge case depending on the type of key the blinding factor is. 
So we have to create a new key that is of type derived, or a new key that is of type import

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
SWvheerden: I removed the function, it has the potential to be very slow, and its not needed

SWvheerden: this function is anyway adding in the wrong info

hansieodendaal (approved):
utACK

PR #6375, opened 2024-06-20
fix: keep alpha key constant

SWvheerden:
Description
---
FFI now re-uses alpha from previous runs. 
Move DB read of the alpha key from console wallet to wallet lib

Motivation and Context
---
The alpha key should be constant across starts, the FFI creates a new alpha key on each startup

### DIFF
--- a/applications/minotari_console_wallet/src/init/mod.rs
+++ b/applications/minotari_console_wallet/src/init/mod.rs
@@ -430,7 +430,6 @@ pub async fn init_wallet(
     };
 
     let master_seed = read_or_create_master_seed(recovery_seed.clone(), &wallet_db)?;
-    let wallet_type = read_or_create_wallet_type(wallet_type, &wallet_db);
 
     let node_identity = match config.wallet.identity_file.as_ref() {
         Some(identity_file) => {
@@ -476,7 +475,7 @@ pub async fn init_wallet(
         key_manager_backend,
         shutdown_signal,
         master_seed,
-        wallet_type.unwrap(),
+        wallet_type,
         user_agent,
     )
     .await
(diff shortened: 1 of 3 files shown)


hansieodendaal:
Some failing cucumber integration tests

hansieodendaal (commented):
The name and description of the PR do not line up with the code changes. I think a better explanation is needed.

SWvheerden:
> The name and description of the PR do not line up with the code changes. I think a better explanation is needed.

The description is correct. Alpha is now always loaded in and not just from CW side.

applications/minotari_console_wallet/src/init/mod.rs
```diff
@@ -476,7 +475,7 @@ pub async fn init_wallet(
         key_manager_backend,
         shutdown_signal,
         master_seed,
-        wallet_type.unwrap(),
```
brianp, line 479: Dirty little unwrap snuck in there.

base_layer/wallet/src/wallet.rs
```diff
@@ -166,9 +166,10 @@ where
         key_manager_backend: TKeyManagerBackend,
         shutdown_signal: ShutdownSignal,
         master_seed: CipherSeed,
-        wallet_type: WalletType,
+        wallet_type: Option<WalletType>,
         user_agent: String,
     ) -> Result<Self, WalletError> {
+        let wallet_type = read_or_create_wallet_type(wallet_type, &wallet_database)?;
```
brianp, line 172: What's the benefit doing it on start instead of initialization? 

FFI wasn't calling the same initialization function?

brianp (approved):

base_layer/wallet/src/wallet.rs
SWvheerden: the init was in console wallet, not being called by the wallet_lib and wallet_ffi

PR #6376, opened 2024-06-20
feat!: make stealth one-sided tx's look the same as all other tx's

SWvheerden:
Description
---
Make stealth one-sided transactions look like interactive and normal 1-sided

Motivation and Context
---
This increases privacy for transactions leaking less detail.


hansieodendaal:
Some clippy errors

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -21,15 +21,13 @@
...
 use log::*;
 use tari_common_types::{
     tari_address::TariAddress,
-    types::{Commitment, PrivateKey, PublicKey},
+    types::{Commitment, PrivateKey},
 };
-use tari_crypto::keys::PublicKey as PK;
 use tari_key_manager::key_manager_service::{KeyManagerInterface, KeyManagerServiceError};
-use tari_script::{one_sided_payment_script, stealth_payment_script, ExecutionStack, TariScript};
+use tari_script::{push_pubkey_script, ExecutionStack, TariScript};
```
hansieodendaal, line 30: Let's rather call this `one_sided_payment_script` as it relays the intent, and it will be the same for stealth or non-stealth addresses.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1329,11 +1328,17 @@ where
...
             JoinHandle<Result<TxId, TransactionServiceProtocolError<TxId>>>,
         >,
-        script: TariScript,
+        recipient_script: Option<TariScript>,
         payment_id: PaymentId,
     ) -> Result<TxId, TransactionServiceError> {
         let tx_id = TxId::new_random();
 
+        // If the script is not provided, use the default script as we will replace this later on here with a stealth
+        // one with the correct public key, for now we only care that the script size is correct
```
hansieodendaal:
```suggestion
        // For a stealth transaction, the script is not provided because the destination address is not known at this stage, 
        // which will be calculated using the stealth public key a bit later on. For now, we use the default script to ensure
        // that the script size is correct.
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1329,11 +1328,17 @@ where
...
+        let (mut script, replace_script) = match recipient_script {
```
hansieodendaal:
```suggestion
        let (mut script, use_stealth_address) = match recipient_script {
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1377,6 +1383,25 @@ where
                 TransactionServiceError::InvalidKeyId("Missing sender offset keyid".to_string()),
             ))?;
 
+        if replace_script {
```
hansieodendaal:
```suggestion
        if use_stealth_address {
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1899,7 +1917,7 @@ where
             fee_per_gram,
             message,
             transaction_broadcast_join_handles,
-            stealth_payment_script(&nonce_public_key, &script_spending_key),
+            None,
```
hansieodendaal:
```suggestion
            None, // The stealth address for the script will be calculated in the next step
```

infrastructure/tari_script/src/lib.rs
```diff
@@ -38,15 +38,7 @@ hash_domain!(CheckSigHashDomain, "com.tari.script.check_sig", 1);
...
-pub fn stealth_payment_script(
-    nonce_public_key: &RistrettoPublicKey,
-    script_spending_key: &RistrettoPublicKey,
-) -> TariScript {
-    script!(PushPubKey(Box::new(nonce_public_key.clone())) Drop PushPubKey(Box::new(script_spending_key.clone())))
-}
-
 /// The standard payment script to be used for one-sided payment to public addresses
-pub fn one_sided_payment_script(destination_public_key: &RistrettoPublicKey) -> TariScript {
+pub fn push_pubkey_script(destination_public_key: &RistrettoPublicKey) -> TariScript {
```
hansieodendaal, line 42:
```suggestion
pub fn one_sided_payment_script(destination_public_key: &RistrettoPublicKey) -> TariScript {
```

hansieodendaal (commented):
Looks good, just some comments to improve readability about the intent.

base_layer/core/src/transactions/coinbase_builder.rs
SWvheerden: I specifically changed it away from that as I want to avoid confusion in the code base between the two. The same script is used for interactive, one-sided and one-sided stealth

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: We know the destination, we dont know the public key.

hansieodendaal: The comment is not clear. Suggest:
```
// For a stealth transaction, the script is not provided because the public key that should be included 
// is not known at this stage, which will be calculated using the stealth public key later on. For now, 
// we include a default public key to ensure that the script size is correct.
```

hansieodendaal:
Lots of integration tests failing

hansieodendaal:
```
github-actions / Test Results (Integration tests)
All 3 runs failed: Scenario: Syncing node while also mining before tip sync: tests/features/Sync.feature:142:5
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1353,7 +1359,8 @@ where
             .await?;
 
         // This call is needed to advance the state from `SingleRoundMessageReady` to `SingleRoundMessageReady`,
-        // but the returned value is not used
+        // but the returned value is not used. We have to wait until the stp creates a sender_offset_private_key for us,
```
brianp:
```suggestion
        // but the returned value is not used. We have to wait until the sender transaction protocol creates a sender_offset_private_key for us,
```

brianp (dismissed):
LGTM

PR #6378, opened 2024-06-24
chore: fix integration tests

SWvheerden:
Description
---
fixes integration tests due to regression bug

### DIFF
--- a/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
+++ b/base_layer/core/src/transactions/transaction_components/encrypted_data.rs
@@ -294,7 +294,7 @@ impl Hex for EncryptedData {
 impl Default for EncryptedData {
     fn default() -> Self {
         Self {
-            data: Vec::with_capacity(STATIC_ENCRYPTED_DATA_SIZE_TOTAL),
+            data: vec![0;STATIC_ENCRYPTED_DATA_SIZE_TOTAL],
         }
     }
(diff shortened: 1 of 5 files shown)


hansieodendaal:
This test still fails:
```
All 3 runs failed: Scenario: Verify gprc cna create block with more than 1 coinbase:
```
```
Step panicked. Captured output: called `Result::unwrap()` on an `Err` value: InvalidSize
  Scenario: Verify gprc cna create block with more than 1 coinbase
   ✔  Given I have a seed node SEED_A
   ✔  When I have 1 base nodes connected to all seed nodes
   ✘  Then generate a block with 2 coinbases from node SEED_A
```

SWvheerden:
> This test still fails:
> 
> ```
> All 3 runs failed: Scenario: Verify gprc cna create block with more than 1 coinbase:
> ```
> 
> ```
> Step panicked. Captured output: called `Result::unwrap()` on an `Err` value: InvalidSize
>   Scenario: Verify gprc cna create block with more than 1 coinbase
>    ✔  Given I have a seed node SEED_A
>    ✔  When I have 1 base nodes connected to all seed nodes
>    ✘  Then generate a block with 2 coinbases from node SEED_A
> ```

A bit jumping the gun here, but its fixed now😀

hansieodendaal (approved):
utACK

PR #6380, opened 2024-06-26
fix: make tx id random

SWvheerden:
Description
---
Force tx_id to be random

Motivation and Context
---
If some apparent reason a transaction fails to be submitted after creation using any method that does not supply a random tx_id, the tx_id is created using the first output. But this will create deterministically create a tx_id. This mean you can never rerun that transaction as it will always create the same tx_id which results in a db unique constraint failure.

### DIFF
--- a/base_layer/core/src/transactions/transaction_protocol/mod.rs
+++ b/base_layer/core/src/transactions/transaction_protocol/mod.rs
@@ -85,9 +85,7 @@
 
 // #![allow(clippy::op_ref)]
 
-use blake2::Blake2b;
 use derivative::Derivative;
-use digest::consts::U32;
 use serde::{Deserialize, Serialize};
 use tari_common_types::types::PrivateKey;
 use tari_crypto::{errors::RangeProofError, signatures::SchnorrSignatureError};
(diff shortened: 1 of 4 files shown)


hansieodendaal (commented):
We can also remove:

```rust
pub fn calculate_tx_id(pub_nonce: &PublicKey, index: usize) -> TxId {
    let hash = CalculateTxIdTransactionProtocolHasherBlake256::new()
        .chain(pub_nonce.as_bytes())
        .chain(index.to_le_bytes())
        .finalize();
    let mut bytes: [u8; 8] = [0u8; 8];
    bytes.copy_from_slice(&hash.as_ref()[..8]);
    u64::from_le_bytes(bytes).into()
}
```

```rust
// hash domain
hash_domain!(
    CalculateTxIdTransactionProtocolHashDomain,
    "com.tari.base_layer.core.transactions.transaction_protocol.calculate_tx_id",
    1
);
```

SWvheerden:
good catch

hansieodendaal (dismissed):
utACK

brianp (approved):
LGTM

PR #6381, opened 2024-06-26
fix: update curve library

AaronFeickert:
Description
---
Updates the curve library dependency.

Motivation and Context
---
The curve library has a [timing vulnerability](https://github.com/advisories/GHSA-x4gp-pqpj-f43q) that was recently fixed. This PR updates the main lock file to pull in the patched version.

How Has This Been Tested?
---
Existing tests pass.

---
Confirm that the updated version is consistent with the security advisory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1384,9 +1384,9 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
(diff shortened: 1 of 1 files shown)


AaronFeickert:
Note that #6373 is also needed to fully patch this.

PR #6382, opened 2024-06-27
feat: enlarge console wallet display

SWvheerden:
Description
---
enlarges the console wallet display to properly display the new addresses

How Has This Been Tested?
---
manual 

Wh


brianp (approved):
utAck

PR #6384, opened 2024-06-28
feat!: faucet and new esme gen block

SWvheerden:
Description
---
Adds in new faucet for esme block
Adds in a new gen block for esme
Adds new Wallet CLI functions to create/spend multiple party n-of-m utxos

How Has This Been Tested?
---
manual

Breaking Changes
---
New Gen block for esme


brianp (dismissed):
LGTM

PR #6389, opened 2024-07-04
fix: claim n of m faucet

SWvheerden:
Description
---
Fixes n-of-m spending of faucets. 

Motivation and Context
---
Allows spending of m-of-n

How Has This Been Tested?
---
Manual


hansieodendaal:
I found another issue. The encumbered transaction (`FaucetEncumberAggregateUtxo`) is broadcast if the wallet is opened before it is finalized with `FaucetSpendAggregateUtxo`.

SWvheerden:
Yeah lets keep that as a known issue, and you should use `--command-mode-auto-exit` for now

hansieodendaal (dismissed):
Nice work.

If https://github.com/SWvheerden/tari/pull/43 is merged I am happy.

PR #6396, opened 2024-07-10
chore: fix unwraps in memory key manager

SWvheerden:
Description
---
Fixes a bunch of unwraps in the memory key manager. 

Motivation and Context
---
We should not use unwraps outside fo tests

### DIFF
--- a/base_layer/core/src/transactions/key_manager/memory_db_key_manager.rs
+++ b/base_layer/core/src/transactions/key_manager/memory_db_key_manager.rs
@@ -28,7 +28,10 @@ use tari_common_sqlite::connection::{DbConnection, DbConnectionUrl};
 use tari_common_types::wallet_types::WalletType;
 use tari_key_manager::{
     cipher_seed::CipherSeed,
-    key_manager_service::storage::{database::KeyManagerDatabase, sqlite_db::KeyManagerSqliteDatabase},
+    key_manager_service::{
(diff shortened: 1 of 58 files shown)


base_layer/core/src/transactions/key_manager/memory_db_key_manager.rs
```diff
@@ -42,25 +46,26 @@ fn random_string(len: usize) -> String {
...
+    let connection = DbConnection::connect_url(&DbConnectionUrl::MemoryShared(random_string(8)))?;
     let cipher = CipherSeed::new();
 
     let mut key = [0u8; size_of::<Key>()];
     OsRng.fill_bytes(&mut key);
     let key_ga = Key::from_slice(&key);
     let db_cipher = XChaCha20Poly1305::new(key_ga);
     let factory = CryptoFactories::new(size);
-
+    key.zeroize();
```
AaronFeickert: Even though this code can't return early before zeroizing the key, I'd recommend using a `Zeroizing` wrapper for `key` to ensure zeroizing always occurs. This would be a hedge against any future updates that could change this behavior, so it's mostly a nit that I think is generally good practice.

Note that you'd need to replace the references with `key.as_mut()` and `key.as_ref()` as appropriate to access the wrapped array.

AaronFeickert (dismissed):
Good catch on zeroizing that key. One small nit suggested, but otherwise LGTM.

AaronFeickert (approved):
LGTM

PR #6398, opened 2024-07-11
fix: script dependance on party order

SWvheerden:
Description
---
fixes script dependence on party order

Motivation and Context
---
Currently, the m-of-n ceremony is dependant on the leader selection to succeed or not. This fixes the process so that the leader correctly constructs the input stack for the script in the correct order. 

How Has This Been Tested?
---
manual


hansieodendaal (approved):
utACK

PR #6399, opened 2024-07-11
feat: simplify leader sig generation

SWvheerden:
Description
---
Simplify the key generation of the leader in m-of-n party

Motivation and Context
---
This will allow us to only have to implement 1 function on the ledger for key generation of the m-n scripts

How Has This Been Tested?
---


hansieodendaal (dismissed):
utACK

Simpler, nice. We just need to get rid of the commented-out code.

PR #6403, opened 2024-07-12
ci(feature): check for signed commits

leet4tari:
Description
Workflow to check all commits are signed

Motivation and Context
Secure all git commits

How Has This Been Tested?
Tested in local fork with personal unsigned and signed commits

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/pr_signed_commits_check.yml
+++ b/.github/workflows/pr_signed_commits_check.yml
@@ -1,0 +1,24 @@
+---
+# Checks if the comments are signed or not
+name: PR - Signed commits check
+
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_signed_commits_check.yml
```diff
@@ -0,0 +1,23 @@
+---
```
SWvheerden, line 1:
```suggestion
# Checks if the comments are signed or not
---
```

PR #6407, opened 2024-07-16
feat!: fix key manager use of keys

SWvheerden:
Description
---
Fixes Key manager key use
Allows creating of scan only wallets that don't have access to the spend key
Allows more than one address

Motivation and Context
---
This allows us to display interactive and one-sided address for ledger type wallets, displaying the correct keys for use with each type. 
The new wallet type also allows users to create scan only wallets that have a private view key and not have access to the private key for spending. 

How Has This Been Tested?
---
Manual

Breaking change
---
Changes byte representation of keys which needs matching ledger wallet version to work correctly

### DIFF
--- a/applications/minotari_console_wallet/src/ui/components/receive_tab.rs
+++ b/applications/minotari_console_wallet/src/ui/components/receive_tab.rs
@@ -29,7 +29,7 @@ impl ReceiveTab {
 
         let chunks = Layout::default()
             .direction(Direction::Vertical)
-            .constraints([Constraint::Length(6), Constraint::Length(23)].as_ref())
+            .constraints([Constraint::Length(8), Constraint::Length(23)].as_ref())
             .margin(1)
             .split(area);
 
@@ -46,6 +46,8 @@ impl ReceiveTab {
                     Constraint::Length(1),
                     Constraint::Length(1),
                     Constraint::Length(1),
+                    Constraint::Length(1),
+                    Constraint::Length(1),
                 ]
                 .as_ref(),
             )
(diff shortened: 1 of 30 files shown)


applications/minotari_console_wallet/src/ui/components/receive_tab.rs
```diff
@@ -57,54 +59,76 @@ impl ReceiveTab {
...
-        const ITEM_01: &str = "Tari Address:     ";
-        const ITEM_02: &str = "Node ID:        ";
-        const ITEM_03: &str = "Network Address: ";
-        const ITEM_04: &str = "Emoji ID:       ";
+        const ITEM_01: &str = "Tari Address interactive:   ";
+        const ITEM_02: &str = "Tari Address one-sided:     ";
+        const ITEM_03: &str = "Node ID:                    ";
+        const ITEM_04: &str = "Network Address:            ";
+        const ITEM_05: &str = "Interactive emoji address:  ";
+        const ITEM_06: &str = "One-sided emoji address:    ";
```
brianp, line 67: If we have a full list, is there benefit in adding the public key? Some base node functions could still use it, as full support for the tari address is still a bit hit or miss.

base_layer/core/src/transactions/key_manager/interface.rs
```diff
@@ -73,26 +76,26 @@ impl TransactionKeyManagerBranch {
...
         match self {
             TransactionKeyManagerBranch::DataEncryption => "data encryption".to_string(),
-            TransactionKeyManagerBranch::Alpha => "alpha".to_string(),
             TransactionKeyManagerBranch::CommitmentMask => "commitment mask".to_string(),
             TransactionKeyManagerBranch::Nonce => "nonce".to_string(),
             TransactionKeyManagerBranch::MetadataEphemeralNonce => "metadata ephemeral nonce".to_string(),
             TransactionKeyManagerBranch::KernelNonce => "kernel nonce".to_string(),
             TransactionKeyManagerBranch::SenderOffset => "sender offset".to_string(),
             TransactionKeyManagerBranch::SenderOffsetLedger => "sender offset ledger".to_string(),
+            TransactionKeyManagerBranch::Spend => COMMS.to_string(),
```
brianp: Why do we leave `COMMS` defined outside? 

Just to avoid circular dependencies?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1250,20 +1233,17 @@ where
...
+                script_input_shares.insert(self.resources.key_manager.get_spend_key().await?.1, self_signature);
 
                 // the order here is important, we need to add the signatures in the same order as public keys where
                 // added to the script originally
                 for key in public_keys {
                     if let Some(signature) = script_input_shares.get(&key) {
                         script_signatures.push(StackItem::Signature(signature.clone()));
                         // our own key should not be added yet, it will be added with the script signing
-                        if &key != self.resources.wallet_identity.address.public_spend_key() {
+                        if key != self.resources.key_manager.get_spend_key().await?.1 {
```
brianp, line 1246: Should we have a custom type for returning a `(P, PK)` instead of using tuples and indexes?

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
```diff
@@ -435,7 +435,7 @@ where
             let completed_transaction = CompletedTransaction::new(
                 self.id,
                 self.source_address.clone(),
-                self.resources.wallet_identity.address.clone(),
+                self.resources.tari_address.clone(),
```
brianp, line 438: Should we define exactly which address this is for clarity? Instead of `tari_address` use a specific `intereactive_address`, `one_sided_address` etc. Or do we not know in this circumstance?

applications/minotari_console_wallet/src/ui/components/receive_tab.rs
SWvheerden: I would rather expand tari address than add legacy stuff here

base_layer/core/src/transactions/key_manager/interface.rs
SWvheerden: yes

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: This is not an idea I want to dismiss out of hand. I remember us looking into it the first time we did this pattern for the key manager, and then decided that for two returns only its perhaps worth it just to return a tuple.

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
SWvheerden: good point, let me change this. These should all be interactive, so lets make it interactive.

base_layer/common_types/src/lib.rs
```diff
@@ -20,6 +20,9 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+// This is the string used to derive the comms/spend key of the wallet
+pub const COMMS: &str = "comms";
```
stringhandler: Probably a better name needed here, otherwise this will be used in multiple places for different use cases
```suggestion
pub const WALLET_COMMS_AND_SPEND_KEY_BRANCH: &str = "comms";
```

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -300,20 +278,144 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
...
-        match tx_label {
-            TransactionKeyManagerLabel::ScriptKey => Ok(DomainSeparatedHasher::<
-                Blake2b<U64>,
-                KeyManagerTransactionsHashDomain,
-            >::new_with_label("script key")),
+    pub(crate) async fn get_private_key(&self, key_id: &TariKeyId) -> Result<PrivateKey, KeyManagerServiceError> {
+        match key_id {
+            KeyId::Managed { branch, index } => {
+                // ledger has special rules here
+                if let WalletType::Ledger(wallet) = &self.wallet_type {
```
stringhandler: nit: It might be more effective here to use a `match` statement so that if a future `WalletType` is added we know to change this code

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -332,6 +434,54 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
         Ok((spend_key_id, spend_public_key, script_key_id, script_public_key))
     }
 
+    pub async fn import_key(&self, private_key: PrivateKey) -> Result<TariKeyId, KeyManagerServiceError> {
+        let public_key = PublicKey::from_secret_key(&private_key);
+        let hex_key = public_key.to_hex();
+        self.db.insert_imported_key(public_key.clone(), private_key)?;
+        trace!(target: LOG_TARGET, "Imported key {}", hex_key);
```
stringhandler: Possible leak of sensitive information here
```suggestion
       
```

base_layer/core/src/transactions/key_manager/interface.rs
```diff
@@ -58,13 +61,13 @@ pub enum TxoStage {
...
 pub enum TransactionKeyManagerBranch {
     DataEncryption = 0x00,
-    Alpha = 0x01,
-    MetadataEphemeralNonce = 0x02,
-    CommitmentMask = 0x03,
-    Nonce = 0x04,
-    KernelNonce = 0x05,
-    SenderOffset = 0x06,
-    SenderOffsetLedger = 0x07,
+    MetadataEphemeralNonce = 0x01,
```
stringhandler, line 64: This looks like a `BREAKING CHANGE` here

base_layer/core/src/transactions/key_manager/interface.rs
stringhandler: It's handy to expose this as a common constant. In https://github.com/tari-project/tari-launchpad/pull/367 I copied the constant to get it working since it wasn't exposed

base_layer/core/src/transactions/key_manager/interface.rs
```diff
@@ -73,26 +76,26 @@ impl TransactionKeyManagerBranch {
...
         match key {
             "data encryption" => TransactionKeyManagerBranch::DataEncryption,
-            "alpha" => TransactionKeyManagerBranch::Alpha,
             "commitment mask" => TransactionKeyManagerBranch::CommitmentMask,
             "metadata ephemeral nonce" => TransactionKeyManagerBranch::MetadataEphemeralNonce,
             "kernel nonce" => TransactionKeyManagerBranch::KernelNonce,
             "sender offset" => TransactionKeyManagerBranch::SenderOffset,
             "sender offset ledger" => TransactionKeyManagerBranch::SenderOffsetLedger,
             "nonce" => TransactionKeyManagerBranch::Nonce,
+            COMMS => TransactionKeyManagerBranch::Spend,
```
stringhandler: As noted above, this would benefit from having a better name

stringhandler (commented):
Not an exhaustive review, but happy with the concept. I think it needs to be changed to be a BREAKING CHANGE, as noted below

base_layer/core/src/transactions/key_manager/interface.rs
SWvheerden: yeah its breaking change for ledger, not software, good catch

base_layer/wallet/src/transaction_service/protocols/transaction_receive_protocol.rs
SWvheerden: Going to fix this in a follow up pr

base_layer/core/src/transactions/key_manager/inner.rs
SWvheerden: it should not, its printing public info, but will remove the log incase

brianp (approved):

PR #6412, opened 2024-07-18
chore: remove dependancy on ledger for common types

SWvheerden:
Description
---
remove dependancy on ledger for common types

Motivation and Context
---
it makes the common type a very basic library without large external dependancies


brianp (approved):
Looks a bit more like before it got split up 👍🏻

PR #6414, opened 2024-07-19
fix: panic in from_base58 fn

SWvheerden:
Description
---
Fixes potential panic in: `from_base58` function
Removes uses of `split_at` and replaces with checked version where user strings are used as input

Motivation and Context
---
Code should not panic


hansieodendaal (approved):
utACK

PR #6415, opened 2024-07-19
chore: fix build and osp

SWvheerden:
Description
---
Fixes build
Fixes original one-sided transaction script persistence
Ensures that spend key is always calculated correctly

PR #6416, opened 2024-07-19
chore: remove unnecessary commitment hashing

AaronFeickert:
Description
---
Removes unnecessary hashing of commitments used in faucet-related signatures.

Motivation and Context
---
Signatures used for faucet functionality unnecessarily hash commitments before passing them as an input message. This may have been done on the assumption that the signing functions often refer to the message as a challenge, despite this not being the case.

This PR removes the unnecessary hashing operations.

It may be worthwhile to refactor the associated signing functions to avoid any possible confusion about the nature of messages and challenges (as suggested in #6418).

How Has This Been Tested?
---
Existing tests pass.

---
Check that the commitment data is being passed into the signing functions correctly, and that those functions do in fact assume the input is an arbitrary message and not a challenge.

### DIFF
--- a/applications/minotari_console_wallet/src/automation/commands.rs
+++ b/applications/minotari_console_wallet/src/automation/commands.rs
@@ -31,9 +31,8 @@ use std::{
     time::{Duration, Instant},
 };
 
-use blake2::Blake2b;
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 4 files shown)


AaronFeickert:
Blocked on the build fix in #6415.

SWvheerden (approved):
This looks good. Just holding off on merging this because if I merge this, the current one will not be spendable.

PR #6419, opened 2024-07-22
chore: sync FFI and Console Wallet payment Id display options

SWvheerden:
Description
---
chore: sync FFI and Console Wallet payment Id display options

Motivation and Context
---
Ensures that the payment id's are displayed in the same way between the console wallet and FFI

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
utACK

PR #6425, opened 2024-07-23
feat: encode max allowable  encrypted bytes into struct

SWvheerden:
Description
---
remove vec<u8> and replace with max_bytes_type

Motivation and Context
---
Does not allow the allocation of more than max bytes


sdbondi (dismissed):
Nice utACK

PR #6426, opened 2024-07-24
feat(wallet): add view key commands

stringhandler:
Description
---
Adds commands for exporting a view key and spend key for an existing wallet, and also commands for creating a read-only wallet that only has access to the view key.

Note this is only possible to the great work by @SWvheerden who enabled this functionality in the wallet.

Motivation and Context
---
This allows you to keep an offline cold wallet that has spend access, and at the same time listen for received funds and act on them. This enables ecommerce and exchange support without the danger of running a hot spending wallet.

How Has This Been Tested?
---
Tested locally

---
You can follow the steps in the [exchange guide](https://github.com/tari-project/tari-dot-com/pull/231) to use all of the additions in this PR

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -75,7 +75,7 @@ use tari_core::{
     blocks::pre_mine::{create_pre_mine_genesis_block_file, get_pre_mine_items, PreMineItem},
     covenants::Covenant,
     transactions::{
-        key_manager::TransactionKeyManagerInterface,
+        key_manager::{SecretTransactionKeyManagerInterface, TransactionKeyManagerInterface},
```
SWvheerden: We should not use this trait:
https://github.com/tari-project/tari/pull/6447

applications/minotari_console_wallet/src/init/mod.rs
```diff
@@ -842,12 +856,47 @@ pub fn prompt_wallet_type(
...
 ) -> Option<WalletType> {
-    if non_interactive {
+    if non_interactive && !matches!(boot_mode, WalletBoot::ViewAndSpendKey) {
         return Some(WalletType::default());
     }
 
     match boot_mode {
+        WalletBoot::ViewAndSpendKey => {
+            let view_key = if let Some(vk) = view_private_key  {
+                match PrivateKey::from_base58(&vk) {
```
SWvheerden: in the command window you use hex, here its base58?

applications/minotari_console_wallet/src/init/mod.rs
```diff
@@ -842,12 +856,47 @@ pub fn prompt_wallet_type(
...
+                        println!("Invalid view key provided");
+                        panic!("Invalid view key provided");
+                    },
+                }
+            } else {
+                prompt_private_key("Enter view key: ").expect("View key provided was invalid")
+            };
+            let spend_key = if
+             let Some(sk) = spend_key {
+                let spend_key = match PublicKey::from_base58(&sk) {
```
SWvheerden: same as other command hex vs base58

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -353,10 +353,7 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
 
                         // If we're trying to access any of the private keys, just say no bueno
                         if &TransactionKeyManagerBranch::Spend.get_branch_key() == branch {
-                            return wallet
-                                .private_spend_key
-                                .clone()
-                                .ok_or(KeyManagerServiceError::ImportedPrivateKeyInaccessible);
+                            return Ok(PrivateKey::default());
```
SWvheerden: I think this will have a conflict anyway, make sure to choose the current code in origin. This error here is fixed.

SWvheerden (commented):
Looking good. Just check the conversion. I don't think we has base58 anywhere expect for the completed tariAddress.

SWvheerden (approved):

hansieodendaal:
I cannot get this menu to be displayed when I run the consol wallet:
```
        loop {
            println!("1. Create a new wallet.");
            println!("2. Recover wallet from seed words or hardware device.");
            println!("3. Create a read-only wallet using a view key.");
```

PR #6428, opened 2024-07-24
fix: wallet address deserialize

SWvheerden:
Description
---
Fixes single address deserialize. 

Motivation and Context
---
Its possible for the single address to be as small as 45 chars. 

How Has This Been Tested?
---
Fuzz testing


base_layer/common_types/src/tari_address/mod.rs
```diff
@@ -246,7 +246,7 @@ impl TariAddress {
 
     /// Construct Tari Address from hex
     pub fn from_base58(hex_str: &str) -> Result<TariAddress, TariAddressError> {
-        if hex_str.len() < 47 {
+        if hex_str.len() < 45 {
```
stringhandler: This would be better with a descriptive const. Perhaps TARI_ADDRESS_BASE58_MIN_LENGTH

SWvheerden: yeah good idea

PR #6431, opened 2024-07-26
feat: backup spend for pre-mine

SWvheerden:
Description
---
Adds backup pre-mine spend functionality.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -257,6 +257,26 @@ where
                 )
                 .await
                 .map(OutputManagerResponse::EncumberAggregateUtxo),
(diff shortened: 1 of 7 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1491,6 +1511,243 @@ where
         ))
     }
 
+    #[allow(clippy::too_many_lines)]
+    pub async fn spend_backup_pre_mine_utxo(
```
hansieodendaal, line 1515: Maybe in another PR (to be prioritized) we can DRY up the code in these functions, as many parts of this function look the same as the one that spends via the m-of-n path.

hansieodendaal (dismissed):
utACK 

I am happy if all the tests and clippy pass.

PR #6433, opened 2024-07-26
feat: check address features on send

SWvheerden:
Description
---
Add checking of addresses on send

Motivation and Context
---
We should not send a transaction to an address not accepting that feature. This will check the network and if the address accepts interactive/one-sided


hansieodendaal (approved):
utACK

brianp (approved):
utAck

PR #6436, opened 2024-07-30
chore: remove domain digest

SWvheerden:
Description
---
removes domain digest from MMR

Motivation and Context
---
MMR should not care what hasher you want to use.


brianp (approved):
utAck

PR #6430, opened 2024-07-26
chore: remove temp code

SWvheerden:
Description
---
Removes temp code as rust has now upgraded to 1.80 stable

PR #6437, opened 2024-07-30
chore: use safer slice splitting

AaronFeickert:
Description
---
Ensures that all slice splitting is properly checked to avoid panics.

Motivation and Context
---
There are a few remaining uses of [`slice::split_at`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) that are used to split slices. Because this can panic on a bad index, this PR moves each use to [`slice::split_at_checked`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_checked), which cannot panic.

How Has This Been Tested?
---
Existing tests pass.

---
Ensure that each split is still being done at the same index, and that the error handling is done properly.

### DIFF
--- a/base_layer/common_types/src/encryption.rs
+++ b/base_layer/common_types/src/encryption.rs
@@ -24,7 +24,6 @@ use std::mem::size_of;
 
 use chacha20poly1305::{
     aead::{Aead, Payload},
-    Tag,
     XChaCha20Poly1305,
     XNonce,
 };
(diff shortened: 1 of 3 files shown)


AaronFeickert:
Note that this does not include the uses already addressed by #6430.

SWvheerden (approved):
I looked at these and found they are not a danger at this moment as they don't take in extra outside input or they check it before hand, 
But the work here is done and that stuff might change in future.

AaronFeickert:
I also didn't identify any particular problems with the existing implementation. As noted, this was to reduce risk going forward in the event of changes.

PR #6440, opened 2024-07-31
feat: add scanned height

SWvheerden:
Description
---
adds a callback to the FFI to get access to the height the wallet scanned blocks to
Adds a UI display to console wallet to view the displayed height

Motivation and Context
---
Just showing the tip height of the block chain does not have that high a meaning, its the scanned height that's important an we should show this.

How Has This Been Tested?
---
Manual + unit tests


stringhandler (approved):

brianp (approved):
How often is the known tip, and scanned height different? I'd usually expect the scanned height to catch up fairly fast. What's the use case / reason we're building for here?

SWvheerden:
Fairly ofter actually. 
If you watch your wallet when mining, it take a few seconds to scan the new mined height, although the base_node already shows height mined.

You see the behavior in when the height shown is at tip, but you don't see your coinbase yet. 

Also when you stop and start the wallet, which is a big thing for mobile. It becomes a big issue.

PR #6441, opened 2024-08-01
fix: lagging code

stringhandler:
Fixes a bug where the node would not reorg to a chain with a lower height, even if it had a higher difficulty


SWvheerden (dismissed):

hansieodendaal (approved):
utACK

PR #6448, opened 2024-08-05
feat: prepare MM backend stuff

SWvheerden:
Description
---
This removes a few panics, and adds in a fake mainnet block.
Duplicates stagenet consensus values for mainnet. 

Motivation and Context
---
As we are getting closer to mainnet, we need to ensure that we can test all of the features. This adds in a temporary mainnet block to use in local testing. This block will not be the final one and will only be a place holder. 

How Has This Been Tested?
---
unit tests


base_layer/wallet/tests/transaction_service_tests/service.rs
```diff
@@ -2103,7 +2103,6 @@ async fn manage_multiple_transactions() {
 
     let database_path = temp_dir.path().to_str().unwrap().to_string();
 
-    // TODO: When using a memory type db connection this test fails at `assert_eq!(tx_reply, 3, "Need 3 replies");`
```
hansieodendaal, line 2107: Has this been fixed?

base_layer/core/src/blocks/pre_mine/mod.rs
```diff
@@ -365,7 +365,7 @@ mod test {
     };
 
     // Only run this when you want to create a new utxo file
-    #[ignore]
+
```
hansieodendaal:
```suggestion
#[ignore]
```

hansieodendaal: Failing test:
```
     Summary [  60.043s] 1012/1298 tests run: 1011 passed, 1 failed, 11 skipped
        FAIL [   2.389s] tari_core blocks::pre_mine::test::print_pre_mine
```

hansieodendaal (commented):
Looks good

base_layer/wallet/tests/transaction_service_tests/service.rs
SWvheerden: No, but the test does not need a memory db, and if that's still an issue, log an issue. Todo's don't belong in code

hansieodendaal: Maybe then convert it to a normal comment for the next maintainer that may struggle with it? Probably not worth an issue.

PR #6452, opened 2024-08-06
fix: comms key managedment

SWvheerden:
Description
---
Separates comms key and spend key into two key types with the same derive value

Motivation and Context
---
The comms key and spend may be the same, or it may be different depending on the type of wallet. This means the key manager needs to handle the key differently depending on what is asked.

How Has This Been Tested?
---
Manual + unit tests

### DIFF
--- a/base_layer/core/src/transactions/key_manager/inner.rs
+++ b/base_layer/core/src/transactions/key_manager/inner.rs
@@ -380,7 +380,7 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
                             // return wallet.private_spend_key.clone().ok_or(
                             //     KeyManagerServiceError::ImportedPrivateKeyInaccessible(key_id.to_string()),
                             // );
-                            return Ok(PrivateKey::default())
+                            return Ok(PrivateKey::default());
                         }
                     },
                 }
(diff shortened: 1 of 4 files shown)


stringhandler (dismissed):
untested ack

applications/minotari_ledger_wallet/common/src/common_types.rs
```diff
@@ -305,6 +307,7 @@ mod test {
             (0x07, Branch::Spend),
             (0x08, Branch::RandomKey),
             (0x09, Branch::PreMine),
+            (0x09, Branch::Comms),
```
delta1:
```suggestion
            (0x10, Branch::Comms),
```

SWvheerden: thanks, but should 0x0A 😃

delta1: oh lol right

base_layer/core/src/transactions/key_manager/inner.rs
```diff
@@ -532,28 +531,13 @@ where TBackend: KeyManagerBackend<PublicKey> + 'static
...
-                    .read()
-                    .await;
-                let key = km.get_private_key(index)?;
-                Ok(key)
-            },
-        }
+        self.get_private_key(&TariKeyId::Managed {
+            branch: branch.clone(),
+            index,
+        })
```
brianp: Something doesn't sit right about this.

Just knowing how we flipped on this implementation before. It just doesn't *seem* correct and I can't put my finger on it.

brianp (changes requested):
Tested, and I can't create a wallet on this branch

image

brianp (dismissed):
Ack

Tested, and it works again.

PR #6457, opened 2024-08-07
feat: check encumbrance later

SWvheerden:
Description
---
move the encumbrance check to later in the send process

Motivation and Context
---
If we have this encumbrance check earlier, the wallet might encounter an error and leave some outputs locked. 

How Has This Been Tested?
---
manual


hansieodendaal (approved):
utACK

PR #6461, opened 2024-08-08
feat: check wallet type on send + ui fixes

SWvheerden:
Description
---
Checks the wallet type on send in the wallet_lib
Adds various fixes for the console UI

Motivation and Context
---
Cleans up the Console wallet UI and makes it easier to use and only display correct info.


hansieodendaal (approved):
utACK

PR #6462, opened 2024-08-09
feat!: ledger metadata signature with receiver adress confirmation

brianp:
Description
---
Send the user spend public key to the ledger for script generation and user validation via the ledger device.

Motivation and Context
---
Initially we though we would need to send all the parts to the ledger, instead we will change the format by hashing the script, and hashing all other fields, then hashing those hashes. 
This allows us to hash the other fields, and send just the hash to the ledger instead of serializing all the independent parts (covenants, encrypted data, etc.) As they have variable lengths and it gets messy quick.

Then we can send just the receiver public spend key to the ledger, and use it to generate the entire TariScript for the message, after the user has verified the address on the ledger screen.

How Has This Been Tested?
---

---

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

<!-- Metadata signature change required a new genesis block


applications/minotari_ledger_wallet/comms/src/accessor_methods.rs
```diff
@@ -559,6 +564,7 @@ pub fn ledger_get_one_sided_metadata_signature(
     value: u64,
     sender_offset_key_index: u64,
     commitment_mask: &PrivateKey,
+    receiver_public_spend_key: &PublicKey,
```
SWvheerden:
```suggestion
    receiver_address: &TariAddress,
```

base_layer/core/src/transactions/key_manager/interface.rs
```diff
@@ -222,8 +222,10 @@ pub trait TransactionKeyManagerInterface: KeyManagerInterface<PublicKey> {
         value: MicroMinotari,
         sender_offset_key_id: &TariKeyId,
         txo_version: &TransactionOutputVersion,
-        metadata_signature_message: &[u8; 32],
+        metadata_signature_message_common: &[u8; 32],
         range_proof_type: RangeProofType,
+        script: &TariScript,
+        receiver_public_spend_key: &PublicKey,
```
SWvheerden: use the address, not the public key

applications/minotari_ledger_wallet/common/src/utils.rs
```diff
@@ -50,3 +38,23 @@ pub fn u16_to_string(number: u16) -> String {
...
+    }
+
+    let hex = identifier.to_owned() + data;
+    let hex = hex.as_str();
+
+    let mut serialized = Vec::new();
+    for i in 0..hex.len() / 2 {
+        let byte = u8::from_str_radix(&hex[i * 2..i * 2 + 2], 16).map_err(|_e| "Invalid hex string".to_string())?;
+        serialized.push(byte);
+    }
```
SWvheerden: doe from_hex to_hex not work here?

applications/minotari_ledger_wallet/wallet/src/handlers/get_one_sided_metadata_signature.rs
```diff
@@ -55,19 +57,49 @@ pub fn handler_get_one_sided_metadata_signature(comm: &mut Comm) -> Result<(), A
...
+    metadata_signature_message_common.clone_from_slice(&data[104..136]);
+
+    let fields = [
+        Field {
+            name: "Amount",
+            value: &format!("{}", value.to_string()),
+        },
+        Field {
+            name: "Receiver",
+            value: &format!("{}", receiver_public_spend_key.to_hex()),
```
SWvheerden: display full address, spend key is not human readable from the address

SWvheerden (changes requested):

SWvheerden (commented):
Looks good

SWvheerden (approved):

PR #6463, opened 2024-08-12
feat: commitment proofs

### LINKED ISSUE #6282: Add commitment opening proofs

AaronFeickert:
Description
---
Adds commitment proofs, which prove knowledge of the opening of a commitment with an optional minimum value assertion.

Closes #6282. Supersedes #6348.

Motivation and Context
---
This PR adds commitment proofs. These proofs show that the prover knows the opening of a given commitment, optionally shows that the commitment binds to at least a specified minimum value, and bind to an arbitrary message to mitigate replay attacks.

Notably, they do _not_ necessarily assert spend authority.

How Has This Been Tested?
---
This needs to be tested manually.

---
Test proof verification succeeds against valid values. Test proof verification fails against different commitments, minimum values, and messages.


applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -1816,6 +1816,76 @@ pub async fn command_runner(
...
+                                println!("Commitment proof: {}", proof.to_hex());
+                            },
+                            Err(e) => eprintln!("CreateCommitmentProof error! {}", e),
+                        }
+                    },
+                    Err(e) => eprintln!("CreateCommitmentProof error! {}", e),
+                }
+            },
+            VerifyCommitmentProof(args) => {
+                match output_service.get_unspent_outputs().await {
```
SWvheerden, line 1855: this will only allow you to verify your own commitments

SWvheerden: If you want to verify any output you probably need a new method that will get that from a base node

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -130,7 +133,7 @@ impl<TBackend, TWalletConnectivity, TKeyManagerInterface>
 where
     TBackend: OutputManagerBackend + 'static,
     TWalletConnectivity: WalletConnectivityInterface,
-    TKeyManagerInterface: TransactionKeyManagerInterface,
+    TKeyManagerInterface: SecretTransactionKeyManagerInterface,
```
SWvheerden, line 136:
```suggestion
    TKeyManagerInterface: TransactionKeyManagerInterface,
```
Dont include this trait. You are not suppose to expose the private keys here

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -468,9 +471,124 @@ where
...
+                .create_commitment_proof(commitment, message, minimum_value)
+                .await
+                .map(OutputManagerResponse::CommitmentProofCreated),
+            OutputManagerRequest::VerifyCommitmentProof(commitment, message, minimum_value, proof) => self
+                .verify_commitment_proof(commitment, message, minimum_value, proof)
+                .map(OutputManagerResponse::CommitmentProofVerified),
         }
     }
 
+    async fn create_commitment_proof(
```
SWvheerden, line 484: ship most of the work to the key manager. 
Its the only place where we should work with private keys.

SWvheerden (changes requested):
This looks good as a proof of view access. Not necessarily proof of ownership. 
We need to move the proof generation to inside the key manager. The services should not work with the private keys. There is a lot of complexity there and it should be kept there. 

The ownership proof will only currently work for your own wallet as it searches only your own wallet for utxos.

AaronFeickert:
@SWvheerden thanks for the review! This was originally based on the design in #6240, but certainly looks like it needs to be changed to be properly useful.

hansieodendaal (commented):
Looks nice, and I agree that we should keep the secret keys in the key manager. 

I initially thought that we should guard against someone just pulling an existing range proof from the blockchain and present it as an ownership proof, but then realized that the transcript is unique.

What is interesting, is that we could create this proof even if the actual output may have a revealed value proof.

AaronFeickert:
> then realized that the transcript is unique

Yep, this was very intentional. It's not possible to collide a commitment proof with a range proof. This is the case even in the "other direction", where an adversary tricks a user into including a malicious message in a commitment proof in an attempt to produce a valid range proof.

SWvheerden:
Going to close this for now, we have the issue still open for this, and this needs to be reworked to make it useful

PR #6464, opened 2024-08-12
fix: stop reuse of nonce keys

SWvheerden:
Description
---
Stops the reuse of nonce keys when a wallet has been recovered.

Motivation and Context
---
Currently, nonce keys are based on incrementing key indexes. This is a problem if the wallet is recovered as the wallet does not keep the index of the nonce keys, but rather starts at 0 again. This change is to make all nonce keys random u64.


hansieodendaal (dismissed):
utACK

PR #6466, opened 2024-08-13
fix: backup spend

SWvheerden:
Description
---
Fixes backup spend to ensure correct keys are used, and correct functions are called

Motivation and Context
---
Functions should be correct in their usage and allow ledger/non ledger keys to be used


hansieodendaal (dismissed):
utACK

Just some clippy issues.

_(Needs to be tested with system-level testing after a new pre-mine has been created)_

PR #6469, opened 2024-08-15
chore:  fix cucumber tests

SWvheerden:
Description
---
Fixes Wallet FFI tests by adding missing callback
Adds panic to places where null pointer is returned


hansieodendaal (approved):
ACK

PR #6470, opened 2024-08-15
chore: add panics to cucumber tests

SWvheerden:
Description
---
Adds panics to cucumber tests to induce a panic in the FFI layer if an error occurs right away rather than getting a null pointer ref later on.


hansieodendaal (approved):
utACK

PR #6472, opened 2024-08-16
feat!: add source address to encrypted data

SWvheerden:
Description
---
This adds the source address to encrypted data

Motivation and Context
---
This allows a one-sided payment to send over its address to the receiving wallet
It also allows a wallet recovering funds to see where the funds came from. 

How Has This Been Tested?
---
unit test and manual


hansieodendaal (approved):
utACK

(Please add an issue to make adding the sender address in encrypted data optional for (a) change, (b) interactive transactions and (c) one-sided transactions. For one-sided transactions the sender should have the ability to decide when creating the transaction. See [#6475](https://github.com/tari-project/tari/issues/6475))

PR #6473, opened 2024-08-16
feat: limit open-ended vectors

hansieodendaal:
Description
---
Limited open-ended vectors to guard against malicious network messages in:
- ProofOfWork::pow_data
- OutputFeatures::coinbase_extra
- BlockTemplateData::aux_chain_hashes
- Controller::current_blob

Motivation and Context
---
This is a _defense-in-depth_ exercise.

How Has This Been Tested?
---
Existing unit tests pass.

---
Code review

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

Serialization will not be compatible


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -109,8 +110,6 @@ pub struct ConsensusConstants {
     permitted_output_types: &'static [OutputType],
     /// The allowlist of range proof types
     permitted_range_proof_types: [(OutputType, &'static [RangeProofType]); 5],
-    /// Coinbase outputs are allowed to have metadata, but it has the following length limit
-    coinbase_output_features_extra_max_length: u32,
```
SWvheerden: Why did you remove this?

base_layer/core/src/proof_of_work/difficulty.rs
```diff
@@ -115,17 +115,20 @@ impl Default for Difficulty {
 
 /// This trait is used to add a type to `CheckedAdd`, which greatly simplifies usage in the code.
 /// It is implemented for `Difficulty` and `u64`.
+#[cfg(feature = "base_node")]
 pub trait CheckedAdd<T> {
     fn checked_add(&self, other: T) -> Option<Self>
     where Self: Sized;
 }
 
+#[cfg(feature = "base_node")]
```
SWvheerden: why are all of these suddenly required?

base_layer/core/src/proof_of_work/proof_of_work.rs
```diff
@@ -25,13 +25,18 @@ use std::fmt::{Display, Error, Formatter};
...
+use tari_common_types::MaxSizeBytes;
 use tari_utilities::hex::Hex;
 
 use crate::proof_of_work::PowAlgorithm;
 
 #[allow(dead_code)]
 pub trait AchievedDifficulty {}
 
+/// This is the maximum size of the proof of work data that can be included in a block header. This is used to prev
+pub const NOT_BEFORE_PROOF_BYTES_SIZE: usize = u16::MAX as usize;
```
SWvheerden, line 37: We need to ensure that the monero header + coinbase data can fit in here

base_layer/core/src/validation/block_body/block_body_internal_validator.rs
```diff
@@ -80,23 +80,18 @@ fn validate_block_specific_checks(
...
 
     Ok(())
 }
 
-fn check_coinbase_output_features(
-    body: &AggregateBody,
-    consensus_constants: &ConsensusConstants,
-) -> Result<(), ValidationError> {
-    let max_coinbase_metadata_size = consensus_constants.coinbase_output_features_extra_max_length();
-    body.check_output_features(max_coinbase_metadata_size)
```
SWvheerden: we should still check this

SWvheerden: You are breaking the consensus pattern here. 
We must have a MAX limit of bytes allowed by the vector. 
BUT that should be independent of the consensus value. The consensus value can allow a lower limit

SWvheerden (changes requested):
This looks good, just two things we need to change here. 
Dont remove the consensus values, they can be the same as the MAX limits, but can be less, they are independent. 
We should strive to have less compiler flags not more. These things cause devs to make mistakes.

base_layer/core/src/proof_of_work/difficulty.rs
hansieodendaal: It is for `cargo check --package minotari_wallet_ffi`

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: `OutputFeatures::coinbase_extra` is now typed as
```rust
pub type CoinBaseExtra = MaxSizeBytes<64>;
...rust
pub struct OutputFeatures {
...
    pub coinbase_extra: CoinBaseExtra
}
```

Previously, we needed to ensure the `coinbase_extra` did not exceed limits with `ConsensusConstants::pub fn coinbase_output_features_extra_max_length(&self)`, now we just call `CoinBaseExtra::default().max_size()`

base_layer/core/src/proof_of_work/proof_of_work.rs
hansieodendaal: Agreed. This is quite large at the moment, `assert_eq!(u16::MAX, 65535);`, much larger than a monero header and coinbase data.

https://github.com/monero-project/monero/blob/b089f9ee69924882c5d14dd1a6991deb05d9d1cd/src/cryptonote_basic/cryptonote_basic.h#L457

```c++
  /************************************************************************/
  /*                                                                      */
  /************************************************************************/
  struct block_header
  {
    uint8_t major_version;
    uint8_t minor_version;  // now used as a voting mechanism, rather than how this particular block is built
    uint64_t timestamp;
    crypto::hash  prev_id;
    uint32_t nonce;

    BEGIN_SERIALIZE()
      VARINT_FIELD(major_version)
      VARINT_FIELD(minor_version)
      VARINT_FIELD(timestamp)
      FIELD(prev_id)
      FIELD(nonce)
    END_SERIALIZE()
  };
```

base_layer/core/src/consensus/consensus_constants.rs
hansieodendaal: Reverting

SWvheerden (approved):

base_layer/core/tests/tests/block_validation.rs
```diff
@@ -871,32 +874,10 @@ async fn test_block_sync_body_validator() {
...
-    };
-    assert!(
-        matches!(
-            err,
-            ValidationError::TransactionError(TransactionError::InvalidOutputFeaturesCoinbaseExtraSize{len, max }) if
-            len == 65 && max == max_len
-        ),
-        "{}",
-        err
-    );
```
SWvheerden: we still need this test

SWvheerden (changes requested):

SWvheerden (approved):

PR #6478, opened 2024-08-19
fix: contact service tari address support

SWvheerden:
Description
---
Contact service still assumes that the node_id and tari address are the same. This changes contact service to check the tari address.

Motivation and Context
---
This allows a user to update the address of a contact without having to delete the contact. It will look at the address given and update the address accordingly. 

How Has This Been Tested?
---
Manual

### DIFF
--- a/base_layer/common_types/src/tari_address/mod.rs
+++ b/base_layer/common_types/src/tari_address/mod.rs
@@ -76,6 +76,10 @@ impl TariAddressFeatures {
     pub fn as_u8(&self) -> u8 {
         self.0
     }
+
+    pub fn combine_features(&self, other: TariAddressFeatures) -> TariAddressFeatures {
+        TariAddressFeatures(self.0 | other.0)
+    }
 }
 
 impl Default for TariAddressFeatures {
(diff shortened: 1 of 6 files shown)


base_layer/common_types/src/tari_address/mod.rs
```diff
@@ -152,6 +158,41 @@ impl TariAddress {
...
+            (TariAddress::Dual(one), _) => Ok(TariAddress::new_dual_address(
+                one.public_view_key().clone(),
+                one.public_spend_key().clone(),
+                one.network(),
+                one.features().combine_features(two.features()),
+            )),
+            (_, TariAddress::Dual(two)) => Ok(TariAddress::new_dual_address(
+                two.public_view_key().clone(),
+                one.public_spend_key().clone(),
+                one.network(),
```
hansieodendaal, line 185: The logic would be more consistent this way:
```suggestion
                two.public_spend_key().clone(),
                two.network(),
```

hansieodendaal (approved):
utACK

PR #6481, opened 2024-08-19
fix: ledger 1-sided spend

SWvheerden:
Description
---
Fix spending of normal 1-sided ledger transactions

Motivation and Context
---
Fixes key usage of ledger

How Has This Been Tested?
---
manual


hansieodendaal (approved):
ACK

PR #6492, opened 2024-08-21
fix!: don't use the ledger unless both keys are for ledger

brianp:
Description
---
Correct the key requests coming from the key manager in the 

Motivation and Context
---

- Correct the `sign_with_nonce_and_challenge` code path to properly request ledger keys only when both keys are to be derived from the ledger, otherwise use the software key manage
- Fix the tari script serialization on the metadata signature to produce proper sigs
- Refactor the function to generate TariScripts on the ledger to be something a bit simpler
- Refactor the metadata signature domain hashing. We were doing `hash(hash(script) + hash(common))` but this can be simplified to `hash(script + hash(common)`. There's no need to hash the script by itself, before hashing it alongside the common hash.

How Has This Been Tested?
---
One sided sends are borked from ledger. TX creation is requesting the wrong keys from ledger.
Metadata signature generation is wrong.

---

Breaking Changes
---

- [ ] None
- [x] Requires hard fork
- [ ] Other - Please specify

The change in hashing function causes the gen block, and faucets to no longer match. This hashing change is a breaking change for the purpose of simplification.

### DIFF
--- a/applications/minotari_ledger_wallet/comms/src/lib.rs
+++ b/applications/minotari_ledger_wallet/comms/src/lib.rs
@@ -29,9 +29,7 @@ mod test {
     use borsh::BorshSerialize;
     use minotari_ledger_wallet_common::{
         get_public_spend_key_bytes_from_tari_dual_address,
-        hex_to_bytes_serialized,
         tari_dual_address_display,
-        PUSH_PUBKEY_IDENTIFIER,
         TARI_DUAL_ADDRESS_SIZE,
     };
     use rand::rngs::OsRng;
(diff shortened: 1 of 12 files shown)


SWvheerden (dismissed):

SWvheerden (dismissed):

applications/minotari_ledger_wallet/comms/src/lib.rs
```diff
@@ -46,6 +44,7 @@ mod test {
     const NOP_IDENTIFIER: &str = "0173";
     const PUSH_ONE_IDENTIFIER: &str = "017c";
     const CHECK_SIG_VERIFY_IDENTIFIER: &str = "21ad";
+    const PUSH_PUBKEY_IDENTIFIER: &str = "217e";
```
hansieodendaal, line 47: @brianp, we should not re-declare this here, rather include `minotari_ledger_wallet_common::PUSH_PUBKEY_IDENTIFIER`

hansieodendaal (commented):
LGTM

PR #6498, opened 2024-08-22
fix(minotari node): add get_sync_progress to allow list if mining is enabled

MCozhusheck:
Description
---
Allows to call GetSyncProgress if node is in mining node.

Motivation and Context
---
Tari Universe V1 needs to know what's the progress of node sync to display it to the user. Right now it's not possible as nodes refuses to response `GetSyncProgress` call and manually adding this through CLI throws exit code 101 (invalid config).

How Has This Been Tested?
---
Run command `cargo test`

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (approved):

hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #6502, opened 2024-08-22
feat!: add custom noise prologue to comms nodes

hansieodendaal:
Description
---
Added a custom noise prologue to the noise protocol to be hashed into the handshake hash value. This enables an application to specify a unique identifier like the genesis block hash as the noise prologue, effectively stopping any communication handshake from succeeding with nodes not on the same genesis block.

_Thanks to @sdbondi for brainstorming and suggesting the way this requirement could be implemented._

Motivation and Context
---
Base nodes and wallets maintained numerous connections from which they could not sync or obtain useful information. This caused base nodes to try and follow other base nodes that were on a higher proof of work but on a different genesis block, and caused wallets to query base nodes that supplied information from a different blockchain.

How Has This Been Tested?
---
Passed all unit tests.
Passed system-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: Communication between base nodes and wallets will be exclusively linked to the genesis block hash once this PR is implemented. They will not be able to communicate to any previous release.

### DIFF
--- a/applications/minotari_node/src/bootstrap.rs
+++ b/applications/minotari_node/src/bootstrap.rs
@@ -25,9 +25,11 @@ use std::{cmp, str::FromStr, sync::Arc};
 use log::*;
 use minotari_app_utilities::{consts, identity_management, identity_management::load_from_json};
 use tari_common::{
-    configuration::bootstrap::ApplicationType,
+    configuration::{bootstrap::ApplicationType, Network},
     exit_codes::{ExitCode, ExitError},
+    get_static_genesis_block_hash,
 };
+use tari_common_types::types::FixedHash;
 use tari_comms::{
     multiaddr::{Error as MultiaddrError, Multiaddr},
     peer_manager::Peer,
@@ -65,6 +67,7 @@ use tari_p2p::{
 };
 use tari_service_framework::{ServiceHandles, StackBuilder};
 use tari_shutdown::ShutdownSignal;
+use tari_utilities::{hex::Hex, ByteArray};
 
 use crate::ApplicationConfig;
 
@@ -125,6 +128,11 @@ where B: BlockchainBackend + 'static
                 base_node_config.network,
                 self.node_identity.clone(),
                 publisher,
+                FixedHash::from_hex(get_static_genesis_block_hash(
+                    Network::get_current_or_user_setting_or_default(),
+                ))
+                .map_err(|e| ExitError::new(ExitCode::ConfigError, e))?
+                .to_vec(),
             ))
             .add_initializer(SoftwareUpdaterService::new(
                 ApplicationType::BaseNode,
--- a/base_layer/contacts/tests/contacts_service.rs
+++ b/base_layer/contacts/tests/contacts_service.rs
@@ -108,7 +108,7 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
             Network::LocalNet,
             node_identity.clone(),
             publisher,
-            b"tari.contats_service.testing.prologue".to_vec(),
+            b"tari.contacts_service.testing.prologue".to_vec(),
         ))
         .add_initializer(LivenessInitializer::new(
             LivenessConfig {
--- a/base_layer/contacts/tests/contacts_service.rs
+++ b/base_layer/contacts/tests/contacts_service.rs
@@ -108,6 +108,7 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
             Network::LocalNet,
             node_identity.clone(),
             publisher,
+            b"tari.contats_service.testing.prologue".to_vec(),
         ))
         .add_initializer(LivenessInitializer::new(
             LivenessConfig {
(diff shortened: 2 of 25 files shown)


AaronFeickert:
Would this handle the case of an intentional chain fork that uses the same genesis block?

hansieodendaal:
> Would this handle the case of an intentional chain fork that uses the same genesis block?

Yes. `P2pInitializer::noise_prologue` is an open byte vector, so one could easily add some consensus constant differentiator to it that will activate at a certain date and time. There could be a glitch in communication for a very short period before and after that time, but it will be rectified once the majority of nodes move past that.

applications/minotari_node/src/bootstrap.rs
```diff
@@ -125,6 +128,11 @@ where B: BlockchainBackend + 'static
                 base_node_config.network,
                 self.node_identity.clone(),
                 publisher,
+                FixedHash::from_hex(get_static_genesis_block_hash(
```
sdbondi, line 131: Perhaps we should add a prefix to ensure the prologue is fully qualified

```rust
    fn block_hash_to_noise_prologue(block_hash: &FixedHash) -> Vec<u8> {
        const PREFIX: &[u8] = b"com.tari.comms.noise.";
        let mut buf = vec![0u8; PREFIX.len() + FixedHash::byte_size()];
        buf[..PREFIX.len()].copy_from_slice(PREFIX);
        buf[PREFIX.len()..].copy_from_slice(&**block_hash);
        buf
    }

**Review by sdbondi — DISMISSED** [review]
utACK

File: applications/minotari_node/src/bootstrap.rs (hunk 1 above)
Comment (hansieodendaal) [reply applications/minotari_node/src/bootstrap.rs:131]: Good idea

File: base_layer/contacts/tests/contacts_service.rs [hunk 2]
```diff
@@ -108,6 +108,7 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
             Network::LocalNet,
             node_identity.clone(),
             publisher,
+            b"tari.contats_service.testing.prologue".to_vec(),
```
brianp:
```suggestion
            b"tari.contacts_service.testing.prologue".to_vec(),
```

brianp (dismissed):

comms/core/src/connectivity/manager.rs
```diff
@@ -602,6 +602,29 @@ impl ConnectivityManagerActor {
         Ok(())
     }
 
+    async fn on_peer_handshake_failure(&mut self, node_id: &NodeId) -> Result<(), ConnectivityError> {
+        let num_failed = self.mark_peer_failed(node_id.clone());
+
+        if (self.peer_manager.find_by_node_id(node_id).await?).is_some() {
```
sdbondi:
```suggestion
        if self.peer_manager.exists_node_id(node_id).await? {
```

sdbondi (commented):
utACK

common/src/lib.rs
```diff
@@ -77,3 +79,18 @@ pub const DEFAULT_MINER_LOG_CONFIG: &str = "config/log4rs_miner.yml";
 pub const DEFAULT_COLLECTIBLES_LOG_CONFIG: &str = "config/log4rs_collectibles.yml";
 
 pub(crate) const LOG_TARGET: &str = "common::config";
+
+/// This is a static function that returns the genesis block hash for the specified network. This is useful for
+/// applications that need to know the genesis block hash for a specific network, but do not have access to the
+/// genesis block in tari_core. Test `fn test_get_static_genesis_block_hash()` in tari_core will fail if these values
+/// are wrong.
+pub fn get_static_genesis_block_hash<'a>(network: Network) -> &'a str {
```
sdbondi:
```suggestion
pub fn get_static_genesis_block_hash(network: Network) -> &'static str {
```

SWvheerden (commented):
I am very much against the idea of using the gen block in the handshake. We have the wire mode, we can use that to stop wallets and base nodes to not talk to each other on different networks. This seems like an over engineered solution to a problem we already have a solution to. 

BUT
My biggest issue is this. When we launch mainnet, we need to be able to compile the wallets ahead of time. Especially mobile wallets that need a minimum of 3 days lead time to go through app approval process. If we merge this pr, we will launch without any mobile wallets for 3 days.

AaronFeickert:
One option could be to do something akin to Zcash's [branch IDs](https://zips.z.cash/zip-0200#specification), which are an opt-in way for intentional chain forks to offer replay protection, and which was also proposed for handshake binding. Forks simply choose an arbitrary but globally-unique brand ID, which is bound to transactions for replay protection, and which can be bound to the handshake for peer interaction.

hansieodendaal:
I am busy changing the PR to be activated by consensus, espescially for mainnet. It will start with a default block hash, and then at any later date, it will be replaced by the actual block hash. Consensus (as `use_handshake_prologue`) will decide if and when the block hash (as the custom handshake noise prologue) will be used. Together with this, I am adding a prologue suffix (as `handshake_prologue_suffix`) to the consensus constants, which when added to the block hash, could be used to differentiate network communication after a soft fork. This approach seems to be somewhat similar to Zcash's branch IDs.

PR #6503, opened 2024-08-23
feat: ledger reconnect

brianp:
Description
---
If the HidApi disconnects, the instance will still exist but can no
longer be used. We need a new instance, but only a single one can
exist at a time. So we're locking it in a mutex to ensure mutable
safety. Then if at some point communication errors because the instance
is stale we will drop the existing instance by re-assignment of the
inner struct field, and create a new working instance.

Motivation and Context
---
Allow someone with an open console wallet to be able to connect and disconnect their ledger at will. Without the need to restart the wallet. If the connection to the ledger fails, it will re-establish the connection (only one attempt) and then proceed.

How Has This Been Tested?
---
Manually with the console wallet.
Using the ledger demo test file it *mostly* works but there's a weird error return on the reconnection of the ledger before opening the Tari app that could also use some more debugging.

---

- Open a console wallet supported by ledger
- Make a transaction successfully
- Unplug the ledger
- Make a failed transaction because the ledger is missing
- Plug the ledger back in
- Make a transaction successfully

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_ledger_wallet/comms/examples/ledger_demo/main.rs
+++ b/applications/minotari_ledger_wallet/comms/examples/ledger_demo/main.rs
@@ -355,7 +355,7 @@ fn main() {
         },
         Err(e) => {
             if e != LedgerDeviceError::Processing(
-                "GetViewKey: Native HID transport error `Ledger device: Io error`".to_string(),
+                "GetViewKey: Native HID transport error `Ledger device not found`".to_string(),
             ) {
                 println!("\nError: Unexpected response ({})\n", e);
                 return;
--- a/applications/minotari_ledger_wallet/comms/src/error.rs
+++ b/applications/minotari_ledger_wallet/comms/src/error.rs
@@ -33,6 +33,9 @@ pub enum LedgerDeviceError {
     /// Native HID transport error
     #[error("Native HID transport error `{0}`")]
     NativeTransport(String),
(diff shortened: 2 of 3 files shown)


hansieodendaal (changes requested):
Hi, tested on Windows. We need some error handling for `Invalid value for AppSW (28161)` returned when the app is not started but the device connected. (For the demo I commented out the `return;` if an unexpected response was received to complete the test.)
```
test: Ledger app not running
✔ Exit the 'MinoTari Wallet' Ledger app and press Enter to continue.. · Ok

Error: Unexpected response (Processing error `Invalid value for AppSW (28161)`)


test: Ledger disconnected
✔ Disconnect the Ledger device and press Enter to continue.. · Ok

test: Ledger reconnected
✔ Reconnect the Ledger device (with password) and press Enter to continue.. · Ok

Error: Unexpected response (Processing error `Invalid value for AppSW (28161)`)


test: Ledger app restart
✔ Start the 'MinoTari Wallet' Ledger app and press Enter to continue.. · Ok
view_key:       64820a542d44b2fcde5cc9dbe2b56f9e486e0a6b7864de3fb1548728f623cb09

Test completed successfully
```

SWvheerden (approved):

PR #6507, opened 2024-08-29
feat: add more verbose comms errors

hansieodendaal:
Description
---
Added more verbose comms errors.

Motivation and Context
---
Trying to hunt down an edge case where a wallet (as RPC client) continuously connects and then disconnects to a base node (as  RPC server) due to no more RPC sessions available, while it seems ample RPC sessions are available when looking at the base node console.

How Has This Been Tested?
---
Pass all unit tests

---
Code review

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


comms/core/src/connectivity/test.rs
```diff
@@ -423,7 +423,7 @@ async fn pool_management() {
         if conn != important_connection {
             assert_eq!(conn.handle_count(), 2);
             // The peer connection mock does not "automatically" publish event to connectivity manager
-            conn.disconnect(Minimized::No).await.unwrap();
+            let _unused = conn.disconnect(Minimized::No).await;
```
SWvheerden: if the disconnect fails, the test will now pass, where as before it would fail. 
Is there a reason for this change?

comms/core/src/proto/rpc.proto
```diff
@@ -52,8 +52,9 @@ message RpcSessionReply {
...
         HANDSHAKE_REJECT_REASON_UNKNOWN = 0;
         HANDSHAKE_REJECT_REASON_UNSUPPORTED_VERSION = 1;
-        HANDSHAKE_REJECT_REASON_NO_SESSIONS_AVAILABLE = 2;
-        HANDSHAKE_REJECT_REASON_PROTOCOL_NOT_SUPPORTED= 3;
+        HANDSHAKE_REJECT_REASON_NO_SERVER_SESSIONS_AVAILABLE = 2;
+        HANDSHAKE_REJECT_REASON_NO_CLIENT_SESSIONS_AVAILABLE = 3;
+        HANDSHAKE_REJECT_REASON_PROTOCOL_NOT_SUPPORTED= 4;
     }
-    HandshakeRejectReason reject_reason = 3;
+    HandshakeRejectReason reject_reason = 4;
```
SWvheerden:
```suggestion
    HandshakeRejectReason reject_reason = 3;
```
Why change this?

hansieodendaal: ~~For sake of consistency. It previously resolved to~~
~~```~~
~~HANDSHAKE_REJECT_REASON_PROTOCOL_NOT_SUPPORTED= 3;~~
~~```~~
~~Now it resolves to~~
~~```~~
~~HANDSHAKE_REJECT_REASON_PROTOCOL_NOT_SUPPORTED= 4;~~
~~```~~

comms/core/src/connectivity/test.rs
hansieodendaal: Hmmm... it seems I chose the easy way out. It turns out that the peer_connection mock's _disconnect_ behaved differently than the actual _disconnect_, which became evident due to the finer grain of error handling. The mock propagated any errors other than the `YamuxControlError::ConnectionClosed` error, whereas the actual implementation silences all errors.

This will be fixed in the new commit.

SWvheerden (approved):

comms/core/src/proto/rpc.proto
hansieodendaal: Fixed

PR #6508, opened 2024-08-29
feat: wip randomx miner

brianp:
Description
---
Provide a randomx miner.

Motivation and Context
---
Don't have it flagged as malware when shipping.

How Has This Been Tested?
---
NA

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


applications/randomx_miner/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "randomx_miner"
+authors = ["The Tari Development Community"]
+description = "The Tari merge mining proxy for xmrig"
+repository = "https://github.com/tari-project/tari"
+license = "BSD-3-Clause"
+version = "1.3.0-pre.0"
```
SWvheerden:
```suggestion
version = "1.4.0-pre.0"
```

applications/randomx_miner/src/run_miner.rs
```diff
@@ -0,0 +1,253 @@
...
+    mut blockhashing_bytes: Vec<u8>,
+    nonce: u32,
+    vm: RandomXVMInstance,
+) -> Result<(Difficulty, Vec<u8>), MiningError> {
+    let nonce_position = 38;
+    blockhashing_bytes[nonce_position..nonce_position + 4].copy_from_slice(&nonce.to_le_bytes());
+
+    let timestamp_position = 8;
+    let timestamp_bytes: [u8; 4] = (EpochTime::now().as_u64() as u32).to_le_bytes();
+    blockhashing_bytes[timestamp_position..timestamp_position + 4].copy_from_slice(&timestamp_bytes);
```
SWvheerden: I would not update the timestamp every cycle, its wasting time

applications/randomx_miner/src/run_miner.rs
```diff
@@ -0,0 +1,253 @@
...
+        let start = Instant::now();
+        loop {
+            if start.elapsed().as_secs() >= config.template_refresh_interval_ms {
+                debug!(
+                    target: LOG_TARGET,
+                    "Thread {} had {}ms pass. Fetching new template to compare",
+                    thread_number, config.template_refresh_interval_ms
+                );
+                let new_block_template =
+                    runtime.block_on(get_block_template(client, node_address, monero_wallet_address))?;
```
SWvheerden: I would not do this here, I would put this on a separate loop, and let that loop check this. If the template changes then we send an interrupt that the mining threads, having each thread check this wastes mining cycles

applications/randomx_miner/src/run_miner.rs
```diff
@@ -0,0 +1,253 @@
...
+                if new_block_template.blocktemplate_blob != block_template.blocktemplate_blob {
+                    info!(
+                        target: LOG_TARGET,
+                        "Thead {} found detected template change. Restarting mining cycle",
+                        thread_number
+                    );
+                    break;
+                }
+            }
+            let (difficulty, hash) = mining_cycle(blockhashing_bytes.clone(), nonce as u32, vm.clone())?;
```
SWvheerden: let it mut burrow rather than clone

applications/randomx_miner/src/run_miner.rs
```diff
@@ -0,0 +1,253 @@
...
+            stats_store.inc_hashed_count();
```
SWvheerden: this will be bad as well, I think we need to cache the results only update every 1000 hashes or so. Or some interval. Not every hash. Every thread will want to update every hash

applications/randomx_miner/src/run_miner.rs
```diff
@@ -0,0 +1,270 @@
...
+    mut blockhashing_bytes: Vec<u8>,
+    nonce: u32,
+    vm: RandomXVMInstance,
+) -> Result<(Difficulty, Vec<u8>), MiningError> {
+    let nonce_position = 38;
+    blockhashing_bytes[nonce_position..nonce_position + 4].copy_from_slice(&nonce.to_le_bytes());
+
+    let timestamp_position = 8;
+    let timestamp_bytes: [u8; 4] = u32::try_from(EpochTime::now().as_u64())?.to_le_bytes();
+    blockhashing_bytes[timestamp_position..timestamp_position + 4].copy_from_slice(&timestamp_bytes);
```
SWvheerden: I would not do this every hash, it wastes CPU cycles

brianp:
Closing this PR in favour of: https://github.com/tari-project/tari/pull/6537/files

and a new repo: https://github.com/tari-project/clythor

PR #6509, opened 2024-08-30
feat(p2pool-randomx-support): added pow algo to GetNewBlockRequest in p2pool proto

ksrichard:
Description
---
Added PoW algo to p2pool client request, so clients can get a SHA-3 or Random-X block to mine on.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_miner/src/run_miner.rs
+++ b/applications/minotari_miner/src/run_miner.rs
@@ -22,24 +22,17 @@
 
 use std::{convert::TryFrom, str::FromStr, sync::Arc, thread, time::Instant};
 
-use crate::{
-    cli::Cli,
-    config::MinerConfig,
-    errors::{err_empty, MinerError},
-    miner::{Miner, MiningReport},
-    stratum::stratum_controller::controller::Controller,
-};
 use futures::stream::StreamExt;
 use log::*;
-use minotari_app_grpc::tari_rpc::pow_algo::PowAlgos;
-use minotari_app_grpc::tari_rpc::PowAlgo;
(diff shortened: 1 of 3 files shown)


applications/minotari_miner/src/run_miner.rs
```diff
@@ -455,8 +457,10 @@ async fn get_new_block_base_node(
 async fn get_new_block_p2pool_node(
     sha_p2pool_client: &mut ShaP2PoolGrpcClient,
 ) -> Result<GetNewBlockResponse, MinerError> {
+    let mut pow_algo = PowAlgo::default();
```
SWvheerden, line 460: Why do this?

SWvheerden: why not just declare immediately with he correct value

ksrichard: because it is simpler to do this as I did not see any method or something that would use `PowAlgos` type to have the right pow algo.

stringhandler: Feels like there is a way, but this is fine for now

stringhandler (approved):

applications/minotari_miner/src/run_miner.rs
ksrichard: Actually I did not find any better yet, since there is no method for that to initialize a `PowAlgo` from `PowAlgos` enum (it is the enum which contains `i32` value for `PowAlgo`), so not sure.

PR #6510, opened 2024-08-30
feat: improve header validation speed

SWvheerden:
Description
---
Improves the header validation speed by caching the gen block hash creation. 

Motivation and Context
---
This is very noticeable during sync, but it takes about 10ms to create the gen block, and if you only want the hash, this can speed up validation a lot if you do this repeatedly. 

After this pr, the validation of 
Sha3x: 10ms -> 10us
RandomX: 20ms -> 10ms

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
utACK

PR #6511, opened 2024-08-30
fix(comms): always send disconnect notification to conn manager

sdbondi:
Description
---
fix(comms): always send disconnect notification to conn manager
fix(comms): take the connection state into account when determining connectedness for tie-breaking

Motivation and Context
---
Previously, if Yamux detected that we were disconnected, the connectivity manager `Disconnected` notification would not be sent. A race condition with disconnects on a connection and the remote disconnecting happen at almost the same time. If the latter happens first, then the notification of the disconnect would not be sent to the connection manager. When a new connection is attempted, the connection manager would think that a previous connection exists.

Ref #6512 
Ref https://github.com/tari-project/tari/issues/6513

How Has This Been Tested?
---
Not explicitly tested, difficult to reproduce.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/core/src/connection_manager/peer_connection.rs
+++ b/comms/core/src/connection_manager/peer_connection.rs
@@ -532,7 +532,6 @@ impl PeerConnectionActor {
                         self.peer_node_id.short_str(),
                         e
                     );
-                    return Ok(());
                 },
                 e => trace!(target: LOG_TARGET, "On disconnect: ({})", e),
             }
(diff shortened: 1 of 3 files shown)


hansieodendaal (dismissed):
utACK

(just the clippy..)

SWvheerden (approved):

PR #6514, opened 2024-08-30
feat: add coinbase extra info to grpc

stringhandler:


applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -319,6 +340,11 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
                         height: current_height,
                         timestamp: current_timestamp.as_u64(),
                         pow_algo: pow_algo.as_u64(),
+                        num_coinbases: coinbases.len() as u64,
+                        first_coinbase_extra: coinbases
+                            .first()
+                            .map(|c| c.features.coinbase_extra.to_vec())
+                            .unwrap_or_default(),
```
hansieodendaal, line 347: - Why only the first coinbase extra?
- If so, how do we know the correct one is returned? Is the ordering not important?

applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -311,6 +311,27 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
...
+                    };
+                    if block.is_none() {
+                        let _network_difficulty_response = tx.send(Err(obscure_error_if_true(
+                            report_error_flag,
+                            Status::internal(format!("Block not found at height {}", current_height)),
+                        )));
+                        return;
+                    }
+                    let block = block.unwrap();
+                    let coinbases = block.block().body.get_coinbase_outputs();
```
hansieodendaal, line 333: If we only need the coinbase extra(s), the function could only return that?

hansieodendaal (commented):
Some comments, questions really.

applications/minotari_node/src/grpc/base_node_grpc_server.rs
stringhandler: I'm not sure what you mean. Would you prefer a different method to return just the coinbase info?

The reason I used it here is that the explorer is calling this method already, and I didn't want to incur the penalty of calling the grpc twice, but it shouldn't be a huge impact

applications/minotari_node/src/grpc/base_node_grpc_server.rs
stringhandler: I think the ordering is important, but it's tricky. We allow multiple coinbases in a block, but that makes it difficult to put in extra merge mining data. It would be easier if the extra was in the header. Perhaps we should mark one coinbase and the primary coinbase, or only allow one coinbase to have an extra data field. @SWvheerden what do you think?

applications/minotari_node/src/grpc/base_node_grpc_server.rs
hansieodendaal: O, I can see how the comment is not clear. I was just wondering why `pub fn get_coinbase_outputs(&self)` could not be made to return a coinbase extra vector instead, seeing that is the only thing used in this function.

stringhandler: I thought it might be useful in future. In fact, some code can be refactored to use this method instead, but I didn't include it in this pr

applications/minotari_node/src/grpc/base_node_grpc_server.rs
SWvheerden: I think perhaps a simpler solution might be a better fit here. 
We have the same problem with transactions and identifying them. They have multiple excess signatures, and we use one to identify them. 
I think here we do something similar. Rather than just returning the first coinbase extra, let's return the first coinbase with extra data in it. I think it will likely only have a single coinbase with extra data in it. There might be some cases where the miner wants to use more than one coinbase put in data, but we cant stop that really...

stringhandler: ah good idea

SWvheerden:
Closing for #6561

PR #6522, opened 2024-09-03
feat: paper wallet cli

SWvheerden:
Description
---
Add paper wallet to console wallet
Add scrape wallet to wallet FFI

Motivation and Context
---
This will allow you to import a paper wallet to your own wallet

How Has This Been Tested?
---
Manual


stringhandler (approved):
lgtm.

Maybe just use a TempFile instead so that it gets deleted on drop

PR #6527, opened 2024-09-04
feat: add resilience to initial wallet connect

hansieodendaal:
Description
---
Added resilience to the initial wallet connection in cases where a custom base node has not been set, where the wallet will try all base nodes or all seeds it knows of and not only one it chose at random.

Motivation and Context
---
Wallets would try indefinitely to establish an initial connection to an offline peer.

How Has This Been Tested?
---
- Added a new unit test.
- System-level testing
  Start a wallet with multiple custom seed nodes where only one is online. Shut down the wallet after the established connection, delete its peer db, and redo. Do this multiple times. Confirm that the wallet establishes a connection every time.

---
Review code changes

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_console_wallet/src/init/mod.rs
+++ b/applications/minotari_console_wallet/src/init/mod.rs
@@ -588,17 +588,12 @@ pub async fn start_wallet(
         .addresses
         .best()
         .ok_or_else(|| ExitError::new(ExitCode::ConfigError, "Configured base node has no address!"))?;
-    let backup_peers = base_nodes
-        .iter()
-        .filter(|&v| v != selected_base_node)
-        .cloned()
-        .collect::<Vec<_>>();
 
     wallet
         .set_base_node_peer(
             selected_base_node.public_key.clone(),
             Some(net_address.address().clone()),
-            Some(backup_peers),
+            Some(base_nodes.to_vec()),
         )
         .await
         .map_err(|e| {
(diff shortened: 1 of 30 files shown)


applications/minotari_console_wallet/src/init/mod.rs
```diff
@@ -570,18 +572,34 @@ fn setup_identity_from_db<D: WalletBackend + 'static>(
...
+        ));
+    }
+    let selected_base_node = base_nodes.choose(&mut OsRng).expect("base_nodes is not empty");
+    let net_address = selected_base_node
         .addresses
         .best()
         .ok_or_else(|| ExitError::new(ExitCode::ConfigError, "Configured base node has no address!"))?;
+    let backup_peers = base_nodes
+        .iter()
+        .filter(|&v| v != selected_base_node)
```
SWvheerden: Why do you remove this one here ?
In automation you did not remove it?

SWvheerden: Looking at the code, I dont think this should be filtered out

SWvheerden (commented):
Looks good, just 1 comment

applications/minotari_console_wallet/src/init/mod.rs
hansieodendaal: `pub async fn set_base_node_peer(` takes in `backup_peers: Option<Vec<Peer>>`, so the selected peer should be removed from the list.

hansieodendaal: But yes, `pub async fn set_base_node_peer(` also filters it out, so I will remove it here.

PR #6541, opened 2024-09-09
feat: ledger version check

### LINKED ISSUE #6439: Minimum version for ledger wallet comms test
Not for this PR. 
But we should probably move this to a min number rather than expected.

_Originally posted by @SWvheerden in https://github.com/tari-project/tari/pull/6432#discussion_r1694923062_

SWvheerden:
Description
---
Changed the ledger app version check from an explicit check to a min version check
Uses semver for checking

Closes: #6439


hansieodendaal (approved):
utACK

PR #6543, opened 2024-09-09
feat: add configurable exclude dialer addresses for universe

hansieodendaal:
Description
---
Added user-configurable communication node addresses that should never be dialled. In Tari Universe the base node and wallet is configured in a listen-only TCP mode with a fake public address and when their addresses are propagated throughout the network, other peers should not try to dial them.

Motivation and Context
---
The previous implementation was hard-coded and did not work properly.

How Has This Been Tested?
---
Added unit tests.
System-level testing.

---
Code review.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/contacts/tests/contacts_service.rs
+++ b/base_layer/contacts/tests/contacts_service.rs
@@ -26,7 +26,7 @@ use rand::rngs::OsRng;
 use tari_common::configuration::{MultiaddrList, Network, StringList};
 use tari_common_sqlite::connection::{DbConnection, DbConnectionUrl};
 use tari_common_types::{tari_address::TariAddress, types::PublicKey};
-use tari_comms::{multiaddr::Multiaddr, peer_manager::PeerFeatures, NodeIdentity};
+use tari_comms::{peer_manager::PeerFeatures, NodeIdentity};
 use tari_comms_dht::{store_forward::SafConfig, DhtConfig};
 use tari_contacts::contacts_service::{
     error::{ContactsServiceError, ContactsServiceStorageError},
--- a/base_layer/contacts/tests/contacts_service.rs
+++ b/base_layer/contacts/tests/contacts_service.rs
@@ -88,6 +88,7 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
                 auto_request: true,
                 ..Default::default()
             },
+            excluded_dial_addresses: vec![],
             ..Default::default()
         },
         allow_test_addresses: true,
@@ -96,10 +97,6 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
         rpc_max_simultaneous_sessions: 0,
         rpc_max_sessions_per_peer: 0,
         listener_self_liveness_check_interval: None,
-        excluded_dial_addresses: vec![
-            "/ip4/172.2.3.4/tcp/18188".parse::<Multiaddr>().expect("will not fail"),
-            "/ip4/172.2.3.4/tcp/18189".parse::<Multiaddr>().expect("will not fail"),
-        ],
     };
     let peer_message_subscription_factory = Arc::new(subscription_factory);
     let shutdown = Shutdown::new();
--- a/base_layer/contacts/tests/contacts_service.rs
+++ b/base_layer/contacts/tests/contacts_service.rs
@@ -26,7 +26,7 @@ use rand::rngs::OsRng;
 use tari_common::configuration::{MultiaddrList, Network, StringList};
 use tari_common_sqlite::connection::{DbConnection, DbConnectionUrl};
 use tari_common_types::{tari_address::TariAddress, types::PublicKey};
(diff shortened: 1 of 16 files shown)


base_layer/contacts/tests/contacts_service.rs
```diff
@@ -96,6 +96,10 @@ pub fn setup_contacts_service<T: ContactsBackend + 'static>(
         rpc_max_simultaneous_sessions: 0,
         rpc_max_sessions_per_peer: 0,
         listener_self_liveness_check_interval: None,
+        excluded_dial_addresses: vec![
+            "/ip4/172.2.3.4/tcp/18188".parse::<Multiaddr>().expect("will not fail"),
+            "/ip4/172.2.3.4/tcp/18189".parse::<Multiaddr>().expect("will not fail"),
```
SWvheerden: We should not do this, we should use `0.0.0.0` and exclude them. 
These are valid addresses and someone might actually have them and try and use them .
https://ipinfo.io/AS7018/172.0.0.0/12-172.2.0.0/22
https://www.techtarget.com/searchnetworking/answer/What-is-the-IP-address-0000-used-for#:~:text=On%20servers.-,IP%20address%200.0.,On%20PCs%20and%20client%20devices.

comms/core/src/connectivity/manager.rs
```diff
@@ -659,6 +659,21 @@ impl ConnectivityManagerActor {
...
+                    msg
+                );
+                self.ban_peer(
+                    node_id,
+                    Duration::from_secs(3 * 24 * 60 * 60), // 3 days
+                    "All peer addresses are excluded (User intervention)".to_string(),
+                )
+                .await?;
+                (node_id, ConnectionStatus::Failed, None)
+            },
```
SWvheerden: I dont think we should ban the node. 
If the node is banned, they cannot dial us. Which they can and should be

SWvheerden (changes requested):
This looks good so far, I have two issues with the implementation:
1: We should not exclude 172.2.3.4 this is a valid address, we should use 0.0.0.0 which is the special case address and exclude that. 
2: We should not ban the node as it will cause issues for that node. We should just exclude it from being dialed. But we should allow that node to dial us.

hansieodendaal:
Currently, setting `public_addresses = ["/ip4/0.0.0.0/tcp/xxxx",]` for base node and `public_addresses = ["/ip4/0.0.0.0/tcp/yyyy",]` for wallet does not work, as the remote peer (i.e. seed node in this case) closes the channel once a connection has been made. 

To allow a peer to advertize a public address of `"/ip4/0.0.0.0/tcp/xxxx"` and accept that it should keep open their inbound connections, but not try to connect to them, is for a next PR.

comms/dht/src/connectivity/mod.rs
```diff
@@ -853,13 +853,47 @@ impl DhtConnectivity {
...
+        if !self.config.excluded_dial_addresses.is_empty() {
+            let mut neighbours = Vec::with_capacity(self.neighbours.len());
+            for peer in &self.neighbours {
+                let addresses = self.peer_manager.get_peer_multi_addresses(peer).await?;
+                if !addresses
+                    .iter()
+                    .any(|addr| self.config.excluded_dial_addresses.contains(addr.address()))
+                {
+                    neighbours.push(peer.clone());
+                }
```
SWvheerden:
```suggestion
                let addresses = self.peer_manager.get_peer_multi_addresses(peer).await?;
                if !addresses
                    .iter()
                    .all(|addr| self.config.excluded_dial_addresses.contains(addr.address()))
                {
                    neighbours.push(peer.clone());
                }
```
Should this not be all?
I don't think we don't want to dial a peer, we only don't want to dial an address. 
If the peer has an address we can dial that's not excluded we should use that

comms/dht/src/connectivity/mod.rs
```diff
@@ -853,13 +853,47 @@ impl DhtConnectivity {
...
+            }
+            self.neighbours = neighbours;
+
+            let mut random_pool = Vec::with_capacity(self.random_pool.len());
+            for peer in &self.random_pool {
+                let addresses = self.peer_manager.get_peer_multi_addresses(peer).await?;
+                if !addresses
+                    .iter()
+                    .any(|addr| self.config.excluded_dial_addresses.contains(addr.address()))
```
SWvheerden:
```suggestion
                    .all(|addr| self.config.excluded_dial_addresses.contains(addr.address()))
```
Same here

SWvheerden (commented):
I think we need to make an distinction between peers and address here. 
A peer might have an invalid, excluded address. But if the peer has other valid addresses, it should not be excluded and still be include. We should just ignore the invalid address.

comms/dht/src/connectivity/mod.rs
hansieodendaal: Correct, I intended that but made a logical error here

SWvheerden (dismissed):

PR #6544, opened 2024-09-09
feat: stop interactive only address from being used in mining

SWvheerden:
Description
---
Block use of interactive only address from being used in coinbase builder


hansieodendaal (approved):
utACK

PR #6545, opened 2024-09-09
feat: display waiting for intial peer data

SWvheerden:
Description
---
Displays waiting for initial peer data on the status string

Motivation and Context
---
better info for the user:
Screenshot 2024-09-09 at 16 59 47


base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
```diff
@@ -230,6 +240,13 @@ impl StateInfo {
             Listening(info) => info.is_synced(),
         }
     }
+
+    pub fn get_initial_connected_peers(&self) -> u64 {
+        match self {
+            StateInfo::Listening(info) => info.initial_delay(),
+            _ => 0,
+        }
+    }
```
hansieodendaal: The naming here is really confusing; I would expect `get_initial_connected_peers` to return `initial_connected_peer`, not `initial_delay`.

hansieodendaal (commented):
One comment, otherwise seems fine

base_layer/core/src/base_node/state_machine_service/states/events_and_states.rs
SWvheerden: good point

PR #6546, opened 2024-09-09
chore: reduce size of proxy logs

stringhandler:
Proxy config would generate 200mb files that are impossible to read in an editor


hansieodendaal (approved):
utACK

PR #6557, opened 2024-09-11
feat!: add multiaddr with range checks for use with universe

hansieodendaal:
Description
---
Added `MultiaddrRange`, which implements range checks for `Multiaddr`, when using IP4 with TCP. This enables specifying a range of IP4 with TCP addresses. As an exmple, any communication node can enable test addresses to connect to them (`allow_test_addresses = true`), but refrain from dialling any test addresses in return (`excluded_dial_addresses = ["/ip4/127.*.*.*/tcp/*"]`). 

With application to universe:
- TCP seed node settings:
   ```
   allow_test_addresses = true
   excluded_dial_addresses = [
         "/ip4/127.*.*.*/tcp/0:18188", 
         "/ip4/127.*.*.*/tcp/18190:65534",
         "/ip4/127.0.0.0/tcp/18189",
         "/ip4/127.1:255.1:255.2:255/tcp/18189"
   ] # Only '/ip4/127.0.0.1/tcp/0:18189' allowed
   ```
- Universe base node settings:
   ```
   type = "tcp"
   public_addresses = ["/ip4/127.0.0.1/tcp/18189"]
   tcp.listener_address = "/ip4/0.0.0.0/tcp/18189"
   allow_test_addresses = true
   #excluded_dial_addresses = []
   ```
- Universe wallet settings:
   ```
   dns_seeds = []
   custom_base_node = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx::/ip4/127.0.0.1/tcp/18189"
   type = "tcp"
   public_addresses = ["/ip4/127.0.0.1/tcp/18188"]
   tcp.listener_address = "/ip4/0.0.0.0/tcp/18188"
   allow_test_addresses = true
   #excluded_dial_addresses = []
   ```

Motivation and Context
---
Currently, Universe base nodes and wallets use `/ip4/172.2.3.4/tcp/18189` and `/ip4/172.2.3.4/tcp/18188` as their public addresses respectively, but any node trying to contact them is not able to. This results in many wasted resources. The Universe wallets also maintain connections with the seed nodes, which is not ideal.

How Has This Been Tested?
---
- Added new unit tests
- System-level testing using the suggested settings^ (simulated seed node, simulated universe base node, simulated universe wallet)

**From the seed node to the universe wallet**
```
>> add-peer 4602fb85883fec887e6b5e5a93cbc9547f19817685e78ff0a9e585826e322b44 /ip4/127.0.0.1/tcp/18188
Peer with node id '9a7764f742bf4e4bae6c5bd4f6' was added to the base node.
>> ☎️  Dialing peer...
☠️ ConnectionFailed: All peer addresses are excluded for peer 9a7764f742bf4e4bae6c5bd4f6
```
**From the seed node to the universe base node**
```
>> add-peer ee9ad9dce31a2d4a9225f4965e50df98ae4f85b58f94b34b9db9cc44f2aa2921 /ip4/127.0.0.1/tcp/18189
Peer with node id '998eb49cf4f2dd3b3d5a394c8e' was added to the base node.
☎️  Dialing peer...
⚡️ Peer connected in 0ms!
Connection: Id: 1, Node ID: 998eb49cf4f2dd3b, Direction: Inbound, Peer Address: /ip4/192.168.5.114/tcp/62398, Age: 1913s, #Substreams: 2, #Refs: 2
```
**From the universe base node to the universe wallet**
```
>> add-peer 4602fb85883fec887e6b5e5a93cbc9547f19817685e78ff0a9e585826e322b44 /ip4/127.0.0.1/tcp/18188
Peer with node id '9a7764f742bf4e4bae6c5bd4f6' was added to the base node.
☎️  Dialing peer...
⚡️ Peer connected in 0ms!
Connection: Id: 8, Node ID: 9a7764f742bf4e4b, Direction: Inbound, Peer Address: /ip4/127.0.0.1/tcp/62412, Age: 2054s, #Substreams: 6, #Refs: 2
```
**From the universe base node to the seed node**
```
>> add-peer 6677c4d401b98f403de671712a98ad8bf2976db27a1d411b08bedfd86751e048 /ip4/192.168.5.114/tcp/9991
Peer with node id '85d605836f02951c65651f99d0' was added to the base node.
☎️  Dialing peer...
>> ⚡️ Peer connected in 0ms!
Connection: Id: 0, Node ID: 85d605836f02951c, Direction: Outbound, Peer Address: /ip4/192.168.5.114/tcp/9991, Age: 2050s, #Substreams: 2, #Refs: 3
```

---
- Code review
- System-level testing

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

**BREAKING CHANGE:** The wallet FFI interface changed (in `fn comms_config_create`)

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -126,6 +126,7 @@ use tari_common_types::{
 };
 use tari_comms::{
     multiaddr::Multiaddr,
+    net_address::IP4_TCP_TEST_ADDR_RANGE,
     peer_manager::{NodeIdentity, PeerQuery},
     transports::MemoryTransport,
     types::CommsPublicKey,
@@ -5326,7 +5327,7 @@ pub unsafe extern "C" fn comms_config_create(
                         minimum_desired_tcpv4_node_ratio: 0.0,
                         ..Default::default()
                     },
-                    excluded_dial_addresses: vec![],
+                    excluded_dial_addresses: vec![IP4_TCP_TEST_ADDR_RANGE.parse().expect("valid address range")],
                     ..Default::default()
                 },
                 allow_test_addresses: true,
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -126,6 +126,7 @@ use tari_common_types::{
 };
 use tari_comms::{
     multiaddr::Multiaddr,
+    net_address::{MultiaddrRange, MultiaddrRangeList, IP4_TCP_TEST_ADDR_RANGE},
     peer_manager::{NodeIdentity, PeerQuery},
     transports::MemoryTransport,
     types::CommsPublicKey,
@@ -5199,6 +5200,7 @@ pub unsafe extern "C" fn transport_config_destroy(transport: *mut TariTransportC
 /// `database_path` - The database path char array pointer which. This is the folder path where the
 /// database files will be created and the application has write access to
 /// `discovery_timeout_in_secs`: specify how long the Discovery Timeout for the wallet is.
+/// `exclude_dial_test_addresses`: exclude dialing of test addresses; this should be 'true' for production wallets
 /// `error_out` - Pointer to an int which will be modified to an error code should one occur, may not be null. Functions
 /// as an out parameter.
 ///
@@ -5217,6 +5219,7 @@ pub unsafe extern "C" fn comms_config_create(
     datastore_path: *const c_char,
     discovery_timeout_in_secs: c_ulonglong,
     saf_message_duration_in_secs: c_ulonglong,
+    exclude_dial_test_addresses: bool,
     error_out: *mut c_int,
 ) -> *mut TariCommsConfig {
     let mut error = 0;
@@ -5294,6 +5297,20 @@ pub unsafe extern "C" fn comms_config_create(
                 MultiaddrList::from(vec![public_address])
             };
 
+            let excluded_dial_addresses = if exclude_dial_test_addresses {
+                let multi_addr_range = match MultiaddrRange::from_str(IP4_TCP_TEST_ADDR_RANGE) {
+                    Ok(val) => val,
+                    Err(e) => {
+                        error = LibWalletError::from(InterfaceError::InternalError(e)).code;
+                        ptr::swap(error_out, &mut error as *mut c_int);
+                        return ptr::null_mut();
+                    },
+                };
+                MultiaddrRangeList::from(vec![multi_addr_range])
+            } else {
+                MultiaddrRangeList::from(vec![])
+            };
+
             let config = TariCommsConfig {
                 override_from: None,
                 public_addresses: addresses,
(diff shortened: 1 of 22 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -5327,7 +5327,8 @@ pub unsafe extern "C" fn comms_config_create(
                         minimum_desired_tcpv4_node_ratio: 0.0,
                         ..Default::default()
                     },
-                    excluded_dial_addresses: vec![],
+                    // FIXME: This should be set to 'IP4_TCP_TEST_ADDR_RANGE', but cucumber tests need to use that range
```
SWvheerden:
```suggestion
```

comms/core/src/net_address/multiaddr_range.rs
```diff
@@ -0,0 +1,540 @@
...
+};
+
+/// A MultiaddrRange for testing purposes that matches any IPv4 address and any port
+pub const IP4_TCP_TEST_ADDR_RANGE: &str = "/ip4/127.*.*.*/tcp/*";
+
+/// ----------------- MultiaddrRange -----------------
+/// A struct containing either an Ipv4AddrRange or a Multiaddr. If a range of IP addresses and/or ports needs to be
+/// specified, the MultiaddrRange can be used, but it only supports IPv4 addresses with the TCP protocol.
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
+pub struct MultiaddrRange {
```
SWvheerden: Why not use an enum here?
It seems as if the code is always `if let Some(xxx)...`
An enum would make this easier to work with as you don't have to if let the

SWvheerden (commented):
This looks good, just one question about the implementation. 
Concept of wild card and range is interesting, and would def work. I would have expected to block network segments, 
as in if you want to block for example 127.0.0.1, you had 127.0.0.1/32
if you want to block 127.0.*.* you block 127.0.0.0/16
They way IP addresses work, I don't think you will ever find a need to block 127.0.*.1

base_layer/wallet_ffi/src/lib.rs
hansieodendaal: Added a parameter to `comms_config_create(` to facilitate this requirement.

hansieodendaal:
> This looks good, just one question about the implementation. Concept of wild card and range is interesting, and would def work. I would have expected to block network segments, as in if you want to block for example 127.0.0.1, you had 127.0.0.1/32 if you want to block 127.0._._ you block 127.0.0.0/16 They way IP addresses work, I don't think you will ever find a need to block 127.0.*.1

I found specifying as below intuitive and flexible.
```
let config_str = r#"something = [
            "/ip4/127.*.100:200.*/tcp/18000:19000",
            "/ip4/127.0.150.1/tcp/18500",
            "/ip4/127.0.0.1/udt/sctp/5678",
            "/ip4/127.*.*.*/tcp/*"
        ]"#;
```

SWvheerden (approved):
What you have done seems intuitive in terms of filtering in general, but its not for networking as that's not how IP addresses work. Thats the point I am making. Networks work in bits and you have your mask (/x at the end) to indicate how many bits are for the network address, and how many bits are for the individual devices on the network. 

For example 127.0.0.1/24 means that the network is 127.0.0 and the 1 is the device id, so 127.0.0.1 and 127.0.0.2 are in the same network while 127.0.1.1 and 127.0.0.1 is not. 
If the IP address is 127.0.0.1/22 then it shares a network with 127.0.0.3.254. They are the same network. the `.` in the IP address mean nothing, they are only there to make it easier to read. 

If you take for example the IP we "spam" a lot 172.2.3.4, its part of 172.2.0.0/22 network, so we cant block that with a single command as that network runs from `172.2.0.1 - 172.2.3.254`

But I will iterate this is fine for now, but we might need to revisit this in future as that's not how IP addresses work.

PR #6558, opened 2024-09-11
fix: upgrade diesel version

SWvheerden:
Description
---
Upgrades diesel version

Motivation and Context
---
Closes security: https://github.com/tari-project/tari/security/dependabot/278

How Has This Been Tested?
---
Manual


hansieodendaal (approved):
utACK

PR #6559, opened 2024-09-12
fix: randomX on mac x86

SWvheerden:
Description
---
Randomx on Mac using x86 fails with the recommended flags and requires fallback to the default flags.


brianp (approved):
Yeah, this is exactly what I thought it would be, but didn't realize we were "recovering" from the error.

PR #6564, opened 2024-09-14
feat: update consensus

SWvheerden:
Description
---
Updates the consensus constants for nextnet, stagenet and mainnet

Motivation and Context
---
Coinbase lock height was updated for mainnet to reflect a day's lock before they can be spent. This is loosely based on btc as they have a 16 our lock height. And we have smaller block times with more re-orgs.

The min difficulty  for sha3x has been upped to be equal to the mining power of roughly 3 4080 gpus


hansieodendaal (approved):
utACK

PR #6567, opened 2024-09-17
feat: add generic config list type and use throughout all config settings

hansieodendaal:
Description
---
Added a generic config list type for use with config files and applied this to config arrays of type 'Vec<T>'. With this PR, all array-type config options can be omitted in the config file and/or passed in via the command line as a config property override with `-p <override>'.

Motivation and Context
---
Using config options where arrays have been used was not flexible or consistent throughout the config implementation. 

How Has This Been Tested?
---
Added unit tests.
System-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [X] None
- [ ] Requires hard fork
- [ ] Other - Please specify


leet4tari (approved):
can't comment on the code, but only some very basic testing on minotari_node with ```base_node.grpc_server_allow_methods``` with example options and ```esmeralda.p2p.seeds.peer_seeds``` as empty.

SWvheerden (approved):

PR #6569, opened 2024-09-18
feat!: add ability to encrypt seed words

SWvheerden:
Description
---
allow the use of encrypted seed words.

### DIFF
--- a/applications/minotari_console_wallet/src/lib.rs
+++ b/applications/minotari_console_wallet/src/lib.rs
@@ -284,7 +284,7 @@ fn get_recovery_seed(
 ) -> Result<Option<CipherSeed>, ExitError> {
     if matches!(boot_mode, WalletBoot::Recovery) && !matches!(wallet_type, Some(WalletType::Ledger(_))) {
         let seed = if let Some(ref seed_words) = cli.seed_words {
-            get_seed_from_seed_words(seed_words)?
+            get_seed_from_seed_words(seed_words, None)?
         } else {
             prompt_private_key_from_seed_words()?
         };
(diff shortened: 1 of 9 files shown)


applications/minotari_console_wallet/src/lib.rs
```diff
@@ -284,7 +284,7 @@ fn get_recovery_seed(
 ) -> Result<Option<CipherSeed>, ExitError> {
     if matches!(boot_mode, WalletBoot::Recovery) && !matches!(wallet_type, Some(WalletType::Ledger(_))) {
         let seed = if let Some(ref seed_words) = cli.seed_words {
-            get_seed_from_seed_words(seed_words)?
+            get_seed_from_seed_words(seed_words, None)?
```
hansieodendaal, line 287: I think the passphrase should be an option here as well, and be included in function `prompt_private_key_from_seed_words()`.

hansieodendaal (commented):
Apart from fixing tests, one functional use comment, otherwise it seems fine.

applications/minotari_console_wallet/src/lib.rs
SWvheerden: This is fine I think, its an internal helper function, console wallet does not yes use the encrypted seed words

hansieodendaal: But what if you want to convert a mobile wallet with encrypted seed words to a console wallet?

hansieodendaal (dismissed):
utACK

applications/minotari_console_wallet/src/lib.rs
SWvheerden: https://github.com/tari-project/tari/issues/6572

PR #6575, opened 2024-09-23
fix: view wallet key usage

SWvheerden:
Description
---
fix view wallet key usage

Motivation and Context
---
View wallet should not try and derive its keys, but rather just return them.


stringhandler (approved):
Tested locally and it's working great

PR #6585, opened 2024-09-25
feat!: add pool and tari difficulties to sha3 miner and merge mining proxy

hansieodendaal:
Description
---
- Added pool and tari difficulties to the p2pool request.
- In p2pool mode, the SHA3 miner and the merge mining proxy will submit the block to the base node if the achieved difficulty equals or exceeds the Tari target difficulty.
- Since the SHA3 miner and the merge mining proxy (_the latter only if enabled in config_) always calculate the achieved difficulty; it is now included in the `SubmitBlockRequest` as an option so that the local p2pool node does not have to recalculate it if provided.

**Edit** 
- For the merge mining proxy, evaluate the best block hash to decide if a new tari template must be requested. In re-orgs where the height will be the same as before the re-org, only checking height is insufficient.
- Added pertinent information to proxy log messages.

Motivation and Context
---
When implemented fully in sha-p2pool:
- Improved efficiency overall in submitting blocks to the base node.
- Improved efficiency for the p2pool node internally when evaluating local blocks.

**Edit** 
- Many orphans were observed in the Tari blockchain when doing merged mining with sha-p2pool.

How Has This Been Tested?
---
Currently, all unit and cucumber tests pass as those run without p2pool mode. A combined system-level test is required when this is implemented.

---
Code review.

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: gRPC `SubmitBlockRequest` interface change that will impact sha-p2pool.

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/proxy.rs
+++ b/applications/minotari_merge_mining_proxy/src/proxy.rs
@@ -187,7 +187,7 @@ impl InnerService {
         let mut base_node_client = self.base_node_client.clone();
         trace!(target: LOG_TARGET, "Successful connection to base node GRPC");
 
-        let result =
+        let tari_tip_info =
             base_node_client
                 .get_tip_info(grpc::Empty {})
                 .await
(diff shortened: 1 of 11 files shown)


applications/minotari_merge_mining_proxy/src/proxy.rs
```diff
@@ -296,49 +296,112 @@ impl InnerService {
...
+                return Err(MmProxyError::LogicalError("No node to submit block to".to_string()));
+            }
+
+            let submit_to_base_node = achieved_target >= block_data.template.tari_target_difficulty;
+            let submit_to_p2pool = p2pool_client.is_some() &&
+                achieved_target >= block_data.template.p2pool_target_difficulty.expect("has value");
+
+            let mut bn_time = Duration::from_secs(0);
+            let mut p2p_time = Duration::from_secs(0);
+            let (resp_bn, resp_p2p) = match (submit_to_base_node, submit_to_p2pool) {
```
SWvheerden: Dont use a match here, use two if's
You are duplicateing the submit to base and submit to p2pool code twice by using a match

applications/minotari_miner/src/run_miner.rs
```diff
@@ -485,35 +488,62 @@ async fn get_new_block_p2pool_node(
...
+    difficulty: u64,
 ) -> Result<SubmitBlockResponse, MinerError> {
-    if config.sha_p2pool_enabled {
-        if let Some(client) = sha_p2pool_client {
-            return Ok(client
-                .submit_block(SubmitBlockRequest {
-                    block: Some(block),
-                    wallet_payment_address: wallet_payment_address.to_hex(),
-                })
+    match (submit_to_base_node, config.sha_p2pool_enabled) {
```
SWvheerden: again here use ifs

PR #6586, opened 2024-09-26
feat: add payment id to notify script

SWvheerden:
Description
---
Adds payment_id to the notify scripts


hansieodendaal (approved):
utACK

PR #6588, opened 2024-09-26
feat: ledger stax + flex support

SWvheerden:
Description
---
Upgrading ledger library for stax and flex support


applications/minotari_ledger_wallet/wallet/src/app_ui/menu.rs
```diff
@@ -45,3 +51,14 @@ pub fn ui_menu_main(comm: &mut Comm) -> Event<Instruction> {
         }
     }
 }
+
+#[cfg(any(target_os = "stax", target_os = "flex"))]
+pub fn ui_menu_main(_: &mut Comm) -> NbglHomeAndSettings {
+    // Load glyph from 64x64 4bpp gif file with include_gif macro. Creates an NBGL compatible glyph.
+    const FERRIS: NbglGlyph = NbglGlyph::from_include(include_gif!("key_64x64.gif", NBGL));
```
brianp:
```suggestion
    const TARI: NbglGlyph = NbglGlyph::from_include(include_gif!("key_64x64.gif", NBGL));
```

applications/minotari_ledger_wallet/wallet/src/app_ui/menu.rs
```diff
@@ -45,3 +51,14 @@ pub fn ui_menu_main(comm: &mut Comm) -> Event<Instruction> {
...
+
+    // Display the home screen.
+    NbglHomeAndSettings::new()
+        .glyph(&FERRIS)
```
brianp:
```suggestion
        .glyph(&TARI)
```

Or anything that is more representative than `FERRIS` :D

applications/minotari_ledger_wallet/wallet/src/handlers/get_one_sided_metadata_signature.rs
```diff
@@ -95,17 +107,35 @@ pub fn handler_get_one_sided_metadata_signature(comm: &mut Comm) -> Result<(), A
...
+            Some(&CROSSMARK),
+        );
+        if !review.show() {
+            return Err(AppSW::UserCancelled);
+        }
+    }
+    #[cfg(any(target_os = "stax", target_os = "flex"))]
+    {
+        // Load glyph from 64x64 4bpp gif file with include_gif macro. Creates an NBGL compatible glyph.
+        const FERRIS: NbglGlyph = NbglGlyph::from_include(include_gif!("key_64x64.gif", NBGL));
```
brianp: Same with `FERRIS`

What's an NBGL campatible glyph anyway? What's NBGL?

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_offset.rs
```diff
@@ -199,6 +202,7 @@ pub fn handler_get_script_offset(
     comm.append(&[RESPONSE_VERSION]); // version
     comm.append(&script_offset.to_vec());
     offset_ctx.reset();
+    offset_ctx.finished = true;
```
brianp: Was this to do with your state issue?
What exactly is the bool solving, I think I'm overlooking its purpose.

applications/minotari_ledger_wallet/wallet/src/main.rs
```diff
@@ -40,62 +36,27 @@ use handlers::{
...
-
-unsafe impl critical_section::Impl for MyCriticalSection {
-    unsafe fn acquire() -> RawRestoreState {
-        // nothing, it's all good, don't worry bout it
-    }
-
-    unsafe fn release(_token: RawRestoreState) {
-        // nothing, it's all good, don't worry bout it
-    }
-}
```
brianp, line 94: Was this tested on the nanos+ ? 

I was under the distinct impression without defining the allocator we lose the ability to well... Allocate anything. Essentially it's memory-less

brianp (commented):
Some changes, and questions.

I don't love that this is how they want us to write apps. I'd almost be more inclined to write two different ones, and modularize all the functional bits so we just wrap it in two different UX's. 

*but* the work is done now, and it is how they suggest doing it so 🤷🏻

applications/minotari_ledger_wallet/wallet/src/main.rs
SWvheerden: yes it was, 
Ledger realised that they need one, so the rust_sdk lib now includes one and it clashes as you can only have one defined.

applications/minotari_ledger_wallet/wallet/src/handlers/get_script_offset.rs
SWvheerden: an attempt at getting the UI working, good catch this needs to be removed

applications/minotari_ledger_wallet/wallet/src/main.rs
brianp: Ahhh neat.

brianp (approved):
Ack

PR #6594, opened 2024-10-02
feat: add use transaction output as provided with encumber_aggregate_utxo

hansieodendaal:
Description
---
Added an option to `fn encumber_aggregate_utxo` to use the transaction output from the blockchain or as provided from file.

Motivation and Context
---
When reading pre-mine from file for genesis block immediate spends, the output manager service should use the provided transaction output and not fetch it from a base node.

How Has This Been Tested?
---
System-level testing

---
Code review

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/wallet/src/output_manager_service/service.rs
+++ b/base_layer/wallet/src/output_manager_service/service.rs
@@ -258,6 +258,7 @@ where
                 dh_shared_secret_shares,
                 recipient_address,
                 original_maturity,
+                use_output,
             } => self
                 .encumber_aggregate_utxo(
                     tx_id,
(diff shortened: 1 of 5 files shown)


base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1255,6 +1257,7 @@ where
         original_maturity: u64,
         range_proof_type: RangeProofType,
         minimum_value_promise: MicroMinotari,
+        use_output: UseOutput,
```
SWvheerden, line 1257: I would remove the output hash and move that into the output type

SWvheerden (dismissed):
Looks good, just one nit

SWvheerden (approved):

PR #6596, opened 2024-10-02
feat: make pre mine spend tx stealth

SWvheerden:
Description
---
Make the pre-mine spending transactions stealth tx


hansieodendaal (commented):
Looking good, I just need to run some system-level tests to verify

hansieodendaal (approved):
ACK (passed system-level testing)

PR #6599, opened 2024-10-03
fix: ffi wallet seed behaviour

SWvheerden:
Description
---
Fixes the wallet FFI behavior so that the wallet will automatically connect to a random seed node and set the rest of the seed nodes as the backup ones. 

Motivation and Context
---
The wallet FFI should not rely on the calling of the set_base_node function to correctly function. That should happen with the start of the wallet. Calling set_base_node should also only limit it to connect to a specific set node and not automatically connect to the seed nodes as well.


hansieodendaal (approved):
utACK

_(this is much better by the way)_

PR #6600, opened 2024-10-03
feat(peerseeds/autoupdate): hickory-dns upgrade, default to system DNS

sdbondi:
Description
---
Upgrade from trust dns to hickory DNS
Allow system default DNS to be used for DNS resolution

Motivation and Context
---
Cloudflare 1.1.1.1 was used by default and this can be blocked. Better UX to use the system default.

How Has This Been Tested?
---
Existing tests updated

---
Seed peers still work

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (dismissed):
pending ci

sdbondi:
Looks like we're using an old version of gcc CI. I've added that to the CI deps install script so let's see.

https://github.com/rustls/rustls/issues/1967

sdbondi:
Guess not - I've removed the `aws-ls-rs` feature from rusttls. Looks like that is working

PR #6602, opened 2024-10-04
fix(core): reduce block messages with many connections

sdbondi:
Description
---
fix(core): reduce block messages with many connections

Motivation and Context
---
When a seed node is connected to many hundreds of base node peers, it propagates blocks on all connections.
This PR changes this to send to 20 random peers. Default propagation_factor is changed from 4 to 20.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (approved):
I like it

SWvheerden (approved):
This is one of those changes we made to make the small networks work better, but its detrimental to bigger ones

PR #6604, opened 2024-10-04
feat: dont sync local only peers

SWvheerden:
Description
---
Dont send peers with local only addresses over the wire

### DIFF
--- a/comms/core/src/peer_validator/helpers.rs
+++ b/comms/core/src/peer_validator/helpers.rs
@@ -95,20 +95,12 @@ fn validate_address(addr: &Multiaddr, allow_test_addrs: bool) -> Result<(), Peer
             expect_end_of_address(addr_iter)
         },
 
-        Protocol::Ip4(addr)
-            if !allow_test_addrs && (addr.is_loopback() || addr.is_link_local() || addr.is_unspecified()) =>
-        {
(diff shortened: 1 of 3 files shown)


comms/core/src/peer_manager/peer.rs
```diff
@@ -177,6 +178,32 @@ impl Peer {
...
+                let proto = a.address().iter().next();
+                match proto {
+                    Some(Protocol::Ip4(addr)) => !addr.is_loopback() && !addr.is_link_local(),
+                    Some(Protocol::Ip6(addr)) => !addr.is_loopback() && !is_unicast_link_local(&addr),
+                    Some(_) => true,
+                    None => false,
+                }
+            })
+            .collect::<Vec<_>>()
+            .is_empty()
```
sdbondi: `any`? + logic correction

```suggestion
        self
            .addresses
            .addresses()
            .iter()
            .flat_map(|a| a.address())
            .any(|a| {
                match a {
                    Protocol::Ip4(addr) => !addr.is_loopback() && !addr.is_link_local(),
                    Protocol::Ip6(addr) => !addr.is_loopback() && !is_unicast_link_local(&addr),
                    _ => true,
                }
            });
```

sdbondi (approved):
utACK

PR #6608, opened 2024-10-08
feat!: add context to ffi callbacks

SWvheerden:
Description
---
Adds c_void context pointer to all FFI callbacks
Splits get address functions into interactive and one-sided

Motivation and Context
---
By adding a c_void pointer to the all callbacks, each callback can be identified by the context to know from what wallet the callback comes from. 
Identifying the callback becomes important if you are running more than one wallet and thus need to identify the calling wallet. 

Allows the FFI to get both the interactive and one-sided addresses from the wallet. 

Breaking Changes
---

Adds c_void context to all callbacks. 
Renames get_wallet_address to get_wallet_interactive_addess and adds get_wallet_one_sided_address

### DIFF
--- a/base_layer/core/src/blocks/genesis_block.rs
+++ b/base_layer/core/src/blocks/genesis_block.rs
@@ -93,7 +93,6 @@ fn print_mr_values(block: &mut Block, print: bool) {
 
     let mut kernel_mmr = KernelMmr::new(Vec::new());
     for k in block.body.kernels() {
-        println!("k: {}", k);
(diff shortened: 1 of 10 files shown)


brianp (dismissed):
On the client side what's being used as the context? Any pointer I assume, but specifically, they've decided to use? The wallet pointer?

SWvheerden:
yeah its any pointer, but I dont think you can choose the wallet pointer in the way its constructed. As you need to provide the pointer before you get the wallet?

PR #6610, opened 2024-10-08
chore: changes mainnet default network

SWvheerden:
Description
---
Changes stagenet dev target to run mainnet network by default and not stagenet


brianp (approved):
Ack

PR #6613, opened 2024-10-08
feat: enable identity grpc method by default

SWvheerden:
Description
---
allow grpc identify method by default as this allows wallets to find the local base node.


hansieodendaal (approved):
utACK

PR #6620, opened 2024-10-10
feat: add FFI function to create seed words from Mbase58 encrypted cipher seed

SWvheerden:
Description
---
add FFI function to create seed words from Mbase58 encrypted cipher seed

Motivation and Context
---
Mobile clients need to get the seed words out so that they can recover the wallet based on an encrypted cipher seed. 

How Has This Been Tested?
---
unit test

### DIFF
--- a/applications/minotari_console_wallet/src/automation/commands.rs
+++ b/applications/minotari_console_wallet/src/automation/commands.rs
@@ -102,6 +102,7 @@ use tari_crypto::{
     ristretto::{pedersen::PedersenCommitment, RistrettoSecretKey},
 };
 use tari_key_manager::{
+    cipher_seed::CipherSeed,
     key_manager_service::{KeyId, KeyManagerInterface},
     SeedWords,
 };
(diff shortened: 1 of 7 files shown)


applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -2485,15 +2486,36 @@ pub async fn command_runner(
...
+                                .map_err(|e| CommandError::General(e.to_string()))?;
+                            CipherSeed::from_enciphered_bytes(&bytes, passphrase)
+                                .map_err(|e| CommandError::General(e.to_string()))?
+                        },
+                        (_, _) => {
+                            return Err(CommandError::General(
+                                "Either seed words or cipher seed must be provided".to_string(),
+                            ))
+                        },
+                    };
```
hansieodendaal: I think the true and false in both cases should be mirrored. For example, if `args.seed_words.as_str().is_empty()` then `SeedWords::from_str(args.seed_words.as_str())` will fail.

hansieodendaal (commented):
One comment

applications/minotari_console_wallet/src/automation/commands.rs
SWvheerden: good catch here

hansieodendaal (approved):
utACK

PR #6621, opened 2024-10-10
feat: upgrade utils

SWvheerden:
Description
---
upgrades tari utilties and tari crypto
upgrades borsh
removes blocking of `127.0.0.1` address by default

Motivation and Context
---
blocking of `127.0.0.1` blocks universe connections between the wallet and base node by default.


hansieodendaal (approved):
utacK

PR #6624, opened 2024-10-11
fix: example mainnet config seed section

SWvheerden:
Description
---
Fixes the mainnet example seed section

Motivation and Context
---
Although there are not any mainnet nodes atm (correctly so), but starting up the application expects something there. This fills in example data just to have the checks be happy


hansieodendaal (commented):
utACK

PR #6636, opened 2024-10-16
feat: change ffi wallet recovery interface

hansieodendaal:
Description
---
Changed the FFI wallet recovery interface to accept a list of base node peers for recovery instead of only a single peer. This will enable the mobile wallet to pass all seed peers to the recovery process instead of only a single peer.

Motivation and Context
---
The recovery process for the mobile wallet should work in the same way that the console wallet does.

How Has This Been Tested?
---
Must be tested with system-level tests with the mobile wallet.

---
Code review.

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: Wallet FFI method `pub unsafe extern "C" fn wallet_start_recovery(` had an interface update.

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -8901,7 +8901,7 @@ pub unsafe extern "C" fn wallet_is_recovery_in_progress(wallet: *mut TariWallet,
 ///
 /// ## Arguments
 /// `wallet` - The TariWallet pointer.
-/// `base_node_public_key` - The TariPublicKey pointer of the Base Node the recovery process will use
+/// `base_node_public_keys` - An optional TariPublicKeys pointer of the Base Nodes the recovery process must use
 /// `recovery_progress_callback` - The callback function pointer that will be used to asynchronously communicate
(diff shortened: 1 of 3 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -8955,7 +8955,7 @@ pub unsafe extern "C" fn wallet_is_recovery_in_progress(wallet: *mut TariWallet,
 #[no_mangle]
 pub unsafe extern "C" fn wallet_start_recovery(
     wallet: *mut TariWallet,
-    base_node_public_key: *mut TariPublicKey,
+    base_node_public_keys: *mut TariPublicKeys,
```
SWvheerden, line 8958: We should not do this, 
We should flip this around. Currently it means the caller aka mobile needs to do the heavy lifting here. 

We should keep this a single base_node key. 
The logic inside the function should be:
```
if TariPublicKey == Null
{
     get all seed peers and use that
}
else{
     use only the specified peer
}
```

SWvheerden (approved):
this is a much better design

PR #6644, opened 2024-10-22
chore!: update feature-dan2

sdbondi:
Description
---
Update feature-dan2 to latest development

Breaking Changes
---

- [ ] None
- [x] Requires hard fork
- [ ] Other - Please specify


leet4tari (dismissed):
utACK - Can only say I don't see anything wrong with the workflow.

PR #6649, opened 2024-10-25
feat: close rpc connections when not in use - without close on dial requests

hansieodendaal:
Description
---
Added the ability to close RPC connections for a given peer:
- The RPC server can request sessions to be dropped.
- RPC session counts on the server are atomically managed with RPC server sessions.
- When a client peer connection (final clone) drops, as a safety precaution, all client RPC connections will be signalled to drop, resulting in the server RPC sessions closing.
- Improved management of wallet connectivity to base nodes.
- **Edit:** Introduced an optional culling of the oldest client RPC connection(s) into an RPC server for a given peer (_default: true_). This will allow all new RPC connections to succeed. If older RPC sessions are dropped by the server while the client is in use, the client can re-request an RPC connection, but many times the clients are not aware of the RPC sessions held on the server.

Motivation and Context
---
Wallets keep on hitting their maximum allowed RPC sessions in the server.

How Has This Been Tested?
---
- Added new unit tests.
- System-level testing.

---
Code review

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/common/config/presets/c_base_node_c.toml
+++ b/common/config/presets/c_base_node_c.toml
@@ -162,7 +162,9 @@ listener_self_liveness_check_interval = 15
 #rpc_max_simultaneous_sessions = 100
 # The maximum comms RPC sessions allowed per peer (default value = 10).
 #rpc_max_sessions_per_peer = 10
-# Cull the oldest peer RPC connection when the maximum number of sessions is reached (default value = true).
+# If true, and the maximum per peer RPC sessions is reached, the RPC server will close an old session and replace it
+# with a new session. If false, the RPC server will reject the new session and preserve the older session.
+# (default value = true).
 #pub cull_oldest_peer_rpc_connection_on_full = true
 
 [base_node.p2p.transport]
--- a/common/config/presets/c_base_node_c.toml
+++ b/common/config/presets/c_base_node_c.toml
@@ -162,6 +162,8 @@ listener_self_liveness_check_interval = 15
 #rpc_max_simultaneous_sessions = 100
 # The maximum comms RPC sessions allowed per peer (default value = 10).
 #rpc_max_sessions_per_peer = 10
+# Cull the oldest peer RPC connection when the maximum number of sessions is reached (default value = true).
+#pub cull_oldest_peer_rpc_connection_on_full = true
 
 [base_node.p2p.transport]
 # -------------- Transport configuration --------------
--- a/comms/core/src/protocol/rpc/client/pool.rs
+++ b/comms/core/src/protocol/rpc/client/pool.rs
@@ -92,19 +92,52 @@ where T: RpcPoolClient + From<RpcClient> + NamedProtocolService + Clone
     pub async fn get_least_used_or_connect(&mut self) -> Result<RpcClientLease<T>, RpcClientPoolError> {
         loop {
             self.check_peer_connection()?;
+            let peer_node_id = self.connection.peer_node_id().clone();
 
             let client = match self.get_next_lease() {
-                Some(c) => c,
+                Some(c) => {
+                    trace!(
+                        target: LOG_TARGET,
+                        "used existing RPC client session for connection '{}'",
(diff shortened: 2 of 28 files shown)


comms/dht/src/store_forward/service.rs
```diff
@@ -264,6 +264,11 @@ impl StoreAndForwardService {
                 warn!(target: LOG_TARGET, "Failed to get the minimize connections threshold: {:?}", err);
                 None
             });
+        if self.ignore_saf_threshold.is_some() {
```
sdbondi, line 269: It would be nice if there was a comment explaining the reasoning for the delay

common/config/presets/c_base_node_c.toml
```diff
@@ -162,6 +162,8 @@ listener_self_liveness_check_interval = 15
 #rpc_max_simultaneous_sessions = 100
 # The maximum comms RPC sessions allowed per peer (default value = 10).
 #rpc_max_sessions_per_peer = 10
+# Cull the oldest peer RPC connection when the maximum number of sessions is reached (default value = true).
```
sdbondi:
```suggestion
# If true, and the maximum per peer RPC sessions is reached, the RPC server will close an old session and replace it with a new session. If false, the RPC server will reject the new session and preserve the older session. (default value = true).
```

comms/core/src/protocol/rpc/client/pool.rs
```diff
@@ -90,34 +90,72 @@ where T: RpcPoolClient + From<RpcClient> + NamedProtocolService + Clone
...
 
             if !client.is_connected() {
+                trace!(
+                    target: LOG_TARGET,
+                    "RPC client for connection '{}' is not connected, pruning",
+                    peer_node_id
+                );
                 self.prune();
-                continue;
+                return Err(RpcClientPoolError::CouldNotObtainRpcConnection);
```
sdbondi, line 154: This appears to return an (rather meaningless) error instead of pruning the disconnected session and returning another session that is connected. What is the reasoning here?

comms/rpc_macros/src/generator.rs
```diff
@@ -198,7 +198,7 @@ impl RpcCodeGenerator {
             pub async fn connect<TSubstream>(framed: #dep_mod::CanonicalFraming<TSubstream>) -> Result<Self, #dep_mod::RpcError>
               where TSubstream: #dep_mod::AsyncRead + #dep_mod::AsyncWrite + Unpin + Send + #dep_mod::StreamId + 'static {
                 use #dep_mod::NamedProtocolService;
-                let inner = #dep_mod::RpcClient::connect(Default::default(), Default::default(), framed, Self::PROTOCOL_NAME.into()).await?;
+                let inner = #dep_mod::RpcClient::connect(Default::default(), Default::default(), framed, Self::PROTOCOL_NAME.into(), Default::default()).await?;
```
sdbondi: nit: for clarity and to avoid potential bugs from future argument changes
```suggestion
                let inner = #dep_mod::RpcClient::connect(Default::default(), Default::default(), framed, Self::PROTOCOL_NAME.into(), None).await?;
```

sdbondi (dismissed):
utACK

comms/core/src/protocol/rpc/client/pool.rs
hansieodendaal: At this stage the client should have been connected, as either an existing client with connection has been assigned, or a new client connection has been established. If it is not connected it is actually an error condition, and my reasoning is rather let the connecting peer decide what to do. This curcumvent a state where the pool tries to establish a connection time after time if the server keeps on rejecting it.

common/config/presets/c_base_node_c.toml
hansieodendaal: I have added some unit tests for this in the new commit

SWvheerden (approved):

PR #6653, opened 2024-10-29
fix(comms): ensure that inbound messaging terminates on disconnect

sdbondi:
Description
---
fix(comms): ensure that inbound messaging terminates on disconnect

Motivation and Context
---
Ensure that the inbound messaging worker terminates when the peer connection is disconnected.
This PR ensures this by using the `PeerConnection::on_disconnect` future. It has not been confirmed 
that the inbound worker would not terminate before however this PR will guarantee it.

How Has This Been Tested?
---
Updated unit tests. 

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7891,6 +7891,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "web-time"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


SWvheerden (approved):

hansieodendaal:
Nice one

Issue #6516, opened 2024-08-31
Base node connectivity starvation - cannot ping active connections or receive metadata


hansieodendaal:
**Node 1 to node 3**

![Image](https://github.com/user-attachments/assets/7b981e74-4776-485e-bf9c-1054546c38dc)

```
>> get-peer a448be3c78c0aecab52b2e677b2a78928c9c9ddb491f20c533ecbd059e401726
Emoji ID: 👢🎢➕🍼🐚💰💉📖💔🍣🍪🏥🐝🥝🐚🐾🐷👘👙🔫🎣🍎🍐💼🥐🚀💩🐋💃🎃🍆🍗🎱
Public Key: a448be3c78c0aecab52b2e677b2a78928c9c9ddb491f20c533ecbd059e401726
NodeId: 57f8de99f1371a7e1f5d0110f2
Addresses:
- /onion3/g4uzpwqgf7biyrfp33vbkyfrfsv6vgnylom6xq2lw5plnus7d5uk2fqd:18141 Score: Some(200)  - Source: FromDiscovery Latency: Some(6.222805ms) - Last Seen: 2024-08-31 05:23:23.196727900 - Last Failure:None
- /ip4/192.168.5.100/tcp/9993 Score: Some(200)  - Source: Config Latency: Some(9.376276ms) - Last Seen: 2024-08-30 15:08:21.470117400 - Last Failure:None
User agent: tari/basenode/1.3.0-pre.0
Features: PeerFeatures(MESSAGE_PROPAGATION | DHT_STORE_FORWARD)
Flags: PeerFlags(SEED)
Supported protocols:
- t/mempool/1
- t/mempool-sync/1
- t/dht/1
- t/bnwallet/1
- t/blksync/1
- t/msg/0.1
Last seen: 2024-08-31 05:23:23.196727900
>>
>>
>> dial-peer a448be3c78c0aecab52b2e677b2a78928c9c9ddb491f20c533ecbd059e401726
>> ☎️  Dialing peer...
⚡️ Peer connected in 0ms!
Connection: Id: 159, Node ID: 57f8de99f1371a7e, Direction: Inbound, Peer Address: /ip4/192.168.5.100/tcp/52844, Age: 177751s, #Substreams: 0, #Refs: 2

>>
>> ping-peer a448be3c78c0aecab52b2e677b2a78928c9c9ddb491f20c533ecbd059e401726
🏓 Pinging peer...
>>
```
```
[...]

PR #6655, opened 2024-10-29
feat: verify active base node peer connections and disconnect if stale

hansieodendaal:
Description
---
Added check connections to the p2p services (`MonitorPeersService`). All active connections are pinged on a set (slowish) interval (10 times slower than the _auto ping metadata interval_). The nodes that do not respond timeously on three consecutive iterations with a corresponding pong are disconnected. This will help keep the list of active connections (lazily) up to date.

**Edit:** 
Fixed an error in the liveness service where misbehaving ping peers were never disconnected. The liveness service and monitor peers service work hand in hand. Liveness selects 8 randomly connected peers to obtain metadata from and will disconnect any of those that misbehave after 1 minute (2x ping intervals). The monitor peers service assesses all connected peers at a much slower pace and disconnects non-responsive peers after 15 minutes (10 x 3 ping intervals).

Motivation and Context
---
See #6516

How Has This Been Tested?
---
Performed system-level testing. From the log below we can see that 5 of 41 active peer connections did not respond with a ping. Peer `e19e1454a1e0519866297960ad ` was disconnected because it did not respond three times in a row,
```
2024-10-29 15:12:07.664466900 [minotari::base_node::monitor_peers] TRACE Found 5 of 41 outbound base node peer 
  connections that did not respond to pings
2024-10-29 15:12:07.664619800 [minotari::base_node::monitor_peers] TRACE Peer e2fa82050c2f7579febafb7e08 
  stats - (iteration, connected, responsive) [(3, true, true), (4, true, true), (5, true, false)]
2024-10-29 15:12:07.664683300 [minotari::base_node::monitor_peers] DEBUG Disconnecting e19e1454a1e0519866297960ad 
  as the peer is no longer responsive - (iteration, connected, responsive) [(2, true, true), (3, true, false), 
  (4, true, false), (5, true, false)]
2024-10-29 15:12:07.665853300 [minotari::base_node::monitor_peers] TRACE Peer 6ea597117476676d5ddcb18153 
  stats - (iteration, connected, responsive) [(1, true, true), (2, true, true), (3, true, true), (4, true, true), 
  (5, true, false)]
2024-10-29 15:12:07.665965500 [minotari::base_node::monitor_peers] TRACE Peer a671f812efe5ab14cbb3c1f9f4 
  stats - (iteration, connected, responsive) [(2, true, true), (3, true, true), (4, true, true), 
  (5, true, false)]
2024-10-29 15:12:07.665997800 [minotari::base_node::monitor_peers] TRACE Peer e336b264e02f611cf4fbf51f22 
  stats - (iteration, connected, responsive) [(2, true, true), (3, true, true), (4, true, true), 
  (5, true, false)]
```

---
- Code review
- System-level testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/p2p/src/services/liveness/service.rs
```diff
@@ -267,9 +273,18 @@ where
...
                 self.state.inc_pings_sent();
-                Ok(LivenessResponse::Ok)
+                Ok(LivenessResponse::Ok(Some(vec![nonce])))
+            },
+            SendPings(node_ids, delay_between_pings) => {
+                let mut nonces = Vec::with_capacity(node_ids.len());
+                for node_id in node_ids {
+                    nonces.push(self.send_ping(node_id).await?);
+                    self.state.inc_pings_sent();
+                    time::sleep(delay_between_pings).await;
```
SWvheerden: this is bad
This will block all liveness requests. If you want to do this you need to move it over to a new thread so that its not blocking here.

SWvheerden (changes requested):
This looks good, 
I have one question regarding why the monitor is in the node application and not in the p2p lib as it seems like that would be the better place to put it. 

And then the sleep needs to come out as it will block the handling of all requests.

applications/minotari_node/src/lib.rs
```diff
@@ -167,6 +168,13 @@ pub async fn run_base_node_with_cli(
             "Force Sync Peers have been set! This node will only sync to the nodes in this set."
         );
     }
+    // // Monitor peers to ensure the reported active connections are still active
+    // task::spawn(monitor_peers::monitor_peers(
+    //     ctx.base_node_comms().clone(),
+    //     ctx.liveness(),
+    //     shutdown,
+    //     config.base_node.metadata_auto_ping_interval,
+    // ));
```
SWvheerden:
```suggestion

```

SWvheerden (dismissed):

SWvheerden (approved):

PR #6658, opened 2024-10-30
feat: validate blockchain fetch from DNS server with local state

MCozhusheck:
Description
---
Add new service which will periodically get checkpoints from a DNS server and validate with it's own state. If we fail this check we will include this information in `get_new_block_template` request so the sender is aware that this node is out of sync.

Motivation and Context
---
This will allow miners including one from Tari Universe to avoid mining on the orphan chain. This implementation is 
modeled after the [Monero Pulse](https://docs.getmonero.org/infrastructure/monero-pulse/#moneropulse)

How Has This Been Tested?
---
Run app on the nextnet and look for logs with `tari_pulse` in the `base_layer.log` file like this:
`tail -f base_layer.log | grep tari_pulse `

Make `get_new_block_template`  and verify if new fields are present.

Create orphan chain (eg. by turning off tor and banning all TCP peers) and check (with above methods) if nodes detects orphan chain.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+use tokio::{net::TcpStream as TokioTcpStream, sync::watch, time};
+
+use super::LocalNodeCommsInterface;
+use crate::base_node::comms_interface::CommsInterfaceError;
+
+const LOG_TARGET: &str = "c::bn::tari_pulse";
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct TariPulseConfig {
+    pub check_interval: Option<Duration>,
```
SWvheerden:
```suggestion
    pub check_interval:Duration,
```
then you can get rid of the unwrap
And implement default, so you can just call default to construct it

applications/minotari_app_grpc/proto/base_node.proto
```diff
@@ -177,6 +178,7 @@ message NewBlockTemplateResponse {
     NewBlockTemplate new_block_template = 1;
     bool initial_sync_achieved = 3;
     MinerData miner_data = 4;
+    MetaData metadata = 5;
```
SWvheerden: why add this?
There is nothing a miner should need for this?
If you want the tip height and tip hash you can `new_block_template.header.height and new_block_template.header.prev_hash`

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        Ok(client)
+    }
+
+    async fn run(
+        &mut self,
+        mut base_node_service: LocalNodeCommsInterface,
+        notify_passed_checkpoints: watch::Sender<bool>,
+    ) {
+        let mut interval = time::interval(self.config.check_interval.unwrap());
```
SWvheerden:
```suggestion
        let mut interval = time::interval(self.config.check_interval);
```
Never unwrap unless in a test

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+pub struct TariPulseService {
+    dns_name: Name,
+    config: TariPulseConfig,
+}
+
+impl TariPulseService {
+    pub async fn new() -> Result<Self, anyhow::Error> {
+        let dns_name = Name::from_str("checkpoints-nextnet.tari.com").unwrap();
+        let config = TariPulseConfig {
+            check_interval: Some(Duration::from_secs(120)),
```
SWvheerden: dont hard code in here, create a default, and pass in the config to the service when created

SWvheerden: ideally read the value from base node config

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+                node_clone.get_header(height).await.and_then(|header| match header {
+                    Some(header) => Ok((header.height(), header.hash().to_hex())),
+                    None => {
+                        error!(target: LOG_TARGET, "Header not found for height: {}", height);
+                        Err(CommsInterfaceError::InternalError("Header not found".to_string()).into())
+                    },
+                })
+            }
+        }))
+        .await?;
```
SWvheerden: This is very weirdly written. 

But you only seem to care about a true or false here if you are on the chain or not. 
So all you have to do is check the last height. if the last height = true, then you are on the same chain.

SWvheerden (commented):
This looks good, a few comments

MCozhusheck:
Support for other chains than nextnet have yet to be implemented after these dns servers are up and running.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: Rewritten logic so we only check for the newest block, reducing numbers of calls to get block header

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: Added field to base node config for this interval

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: Done

applications/minotari_app_grpc/proto/base_node.proto
MCozhusheck: Reverted change

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: Removed unnecesary wrapper

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        Self {
+            check_interval: Duration::from_secs(120),
+            network: Network::default(),
+        }
+    }
+}
+
+fn get_network_dns_name(network: &Network) -> Name {
+    match network {
+        Network::NextNet => Name::from_str("checkpoints-nextnet.tari.com").unwrap(),
```
SWvheerden:
```suggestion
        Network::NextNet => Name::from_str("checkpoints-nextnet.tari.com").expect("infallible"),
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+                    interval_failed.tick().await;
+                    continue;
+                },
+            };
+
+            let max_height_block = dns_checkpoints
+                .iter()
+                .max_by(|a, b| a.0.cmp(&b.0))
+                .ok_or(CommsInterfaceError::InternalError("No checkpoints found".to_string()))
+                .unwrap();
```
SWvheerden:
```suggestion
```
dont use unwraps

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+            failed_checkpoints_notifier: receiver,
+        });
+        let config = TariPulseConfig {
+            check_interval: self.interval.unwrap_or_default(),
+            network: self.network,
+        };
+
+        context.spawn_when_ready(move |handles| async move {
+            let base_node_service = handles.expect_handle::<LocalNodeCommsInterface>();
+            let mut tari_pulse_service = TariPulseService::new(config).await.unwrap();
```
SWvheerden:
```suggestion
            let mut tari_pulse_service = TariPulseService::new(config).await.expect("Should be able to get the service");
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+            check_interval: Duration::from_secs(120),
+            network: Network::default(),
+        }
+    }
+}
+
+fn get_network_dns_name(network: &Network) -> Name {
+    match network {
+        Network::NextNet => Name::from_str("checkpoints-nextnet.tari.com").unwrap(),
+        _ => panic!("Network not supported"),
```
SWvheerden: have you tested this running any network except nextnet?
This panic should not be here. Properly map it out. Even thou the other networks are not supported yet, add their DNS, they can fail connectivity if need be.

applications/minotari_node/src/config.rs
```diff
@@ -142,6 +142,8 @@ pub struct BaseNodeConfig {
     pub state_machine: BaseNodeStateMachineConfig,
     /// Obscure GRPC error responses
     pub report_grpc_error: bool,
+    // Interval to check if the base node is still in sync with the network
+    pub tari_pulse_interval: Option<Duration>,
```
SWvheerden: I dont think serde does handle this option here correct. 
If it does I stand corrected. Please test it, but if it does not, just make it a simple duration
But please add an example here to the config file about the option with the default in: https://github.com/tari-project/tari/blob/development/common/config/presets/c_base_node_c.toml
All the other options are their with their defaults. And they are commented out, but give users an indication what they can set.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    async fn initialize(&mut self, context: ServiceInitializerContext) -> Result<(), ServiceInitializationError> {
+        info!(target: LOG_TARGET, "Initializing Tari Pulse Service");
+        let shutdown_signal = context.get_shutdown_signal();
+        let (sender, receiver) = watch::channel(false);
+        context.register_handle(TariPulseHandle {
+            shutdown_signal: shutdown_signal.clone(),
+            failed_checkpoints_notifier: receiver,
+        });
+        let config = TariPulseConfig {
+            check_interval: self.interval.unwrap_or_default(),
```
SWvheerden: I suspect this will almost always be 0 and not the default of 120 you had it as.

SWvheerden (changes requested):

hansieodendaal:
I am unsure about this approach altogether, as it centralizes authority and appoints the DNS server as an oracle that can influence outcomes.

MCozhusheck:
> I am unsure about this approach altogether, as it centralizes authority and appoints the DNS server as an oracle that can influence outcomes.

This is only used to hint that this node might be not connected to the rest of the network. This was common issue in [Tari Universe](https://github.com/tari-project/universe) where most of the network was connected through Tor but after disabling this  feature in the settings, users where trying to connect through TCP where there no other nodes.
There are some other concerns that required Monero to introduce [Monero Pulse](https://docs.getmonero.org/infrastructure/monero-pulse/#moneropulse-is-dns-based) which solved 1 vulnerability regarding invalid historical block in the chain.

SWvheerden (approved):
LGTM

PR #6660, opened 2024-10-31
feat: trickle validators

ksrichard:
Description
---
We allow here only a configured (in Consensus constants) number of validator nodes to be active in an epoch, so this way we avoid having way too much validators appear at the same time.
All the validator nodes are saved to the blockchain database, when there is a new validator node registration UTXO, but with a different `start epoch` that is calculated from the already saved validator nodes.
There is a new gRPC method called `GetValidatorNodeChanges`, where we can ask base node to give back all the validator node additions and deletions in the given start/end height range.
When Layer-2 scans new blocks and sees that there is a change in the validator node set, simply calls this new grpc method and adds/removes validator nodes. This way base node controls fully how many nodes are registered at which epoch.

Motivation and Context
---
Layer-2 validator nodes are registered all at once when they were added to the chain with a special UTXO, but it allows to register too many nodes at the same time, so consensus can catch up much slower.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/chain_storage/lmdb_db/validator_node_store.rs
```diff
@@ -131,6 +134,61 @@ impl<'a, Txn: Deref<Target = ConstTransaction<'a>>> ValidatorNodeStore<'a, Txn>
...
+                vn_entry.start_epoch <= end_epoch
+            {
+                return Ok(true);
+            }
+        }
+
+        Ok(false)
+    }
+
+    /// Returns validator nodes until a given epoch.
```
hansieodendaal:
```suggestion
    /// Returns validator nodes count until a given epoch.
```

base_layer/core/src/chain_storage/lmdb_db/validator_node_store.rs
```diff
@@ -131,6 +134,61 @@ impl<'a, Txn: Deref<Target = ConstTransaction<'a>>> ValidatorNodeStore<'a, Txn>
...
+        while let Ok(Some((_, value))) = cursor.next::<ValidatorNodeStoreKey>() {
+            if value.sidechain_id == sidechain_id && value.start_epoch <= epoch {
+                result += 1;
+            }
+        }
+
+        Ok(result)
+    }
+
+    /// Returns validator nodes in a given epoch.
```
hansieodendaal:
```suggestion
    /// Returns validator nodes count in a given epoch.
```

hansieodendaal:
Interesting approach.

LGTM, apart form failing clippy and tests.

base_layer/core/src/chain_storage/lmdb_db/validator_node_store.rs
ksrichard: ah yes, makes sense, thanks!

base_layer/core/src/chain_storage/lmdb_db/validator_node_store.rs
ksrichard: good catch, thanks!

sdbondi (approved):

PR #6662, opened 2024-11-01
fix!: explictly close rpc connections when changing base node

sdbondi:
Description
---
fix!: explictly close rpc connections when changing base node
establish relay quickly
fix wallet connectivity logic

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

BREAKING CHANGE: RPC protocol adds the "check bytes" periodically sent by the server. These bytes will cause errors in older clients.


applications/minotari_console_wallet/src/lib.rs
```diff
@@ -106,6 +106,7 @@ pub fn run_wallet(shutdown: &mut Shutdown, runtime: Runtime, config: &mut Applic
         profile_with_tokio_console: false,
         view_private_key: None,
         spend_key: None,
+        reachability: None,
```
hansieodendaal, line 109: Maybe `Some(ReachabilityMode::Private)`  for the wallet?

base_layer/wallet/src/connectivity_service/base_node_peer_manager.rs
```diff
@@ -71,36 +73,47 @@ impl BaseNodePeerManager {
...
     }
 
     /// Set the last connection attempt stats
     pub fn set_last_connection_attempt(&mut self) {
-        self.last_connection_attempt = Some(LastConnectionAttempt {
-            peer_index: self.current_peer_index,
+        let mut lock = self
+            .last_connection_attempt
+            .lock()
+            // In the currently impossible case that a panic occurs while this mutex is unlocked, we'll simply recover to use the previous value before the panic.
```
hansieodendaal:
```suggestion
            // In the currently impossible case that a panic occurs while this mutex is unlocked, we'll simply recover 
            // to use the previous value before the panic.
```

base_layer/wallet/src/connectivity_service/handle.rs
```diff
@@ -61,11 +61,6 @@ impl WalletConnectivityHandle {
 #[async_trait::async_trait]
 impl WalletConnectivityInterface for WalletConnectivityHandle {
     fn set_base_node(&mut self, base_node_peer_manager: BaseNodePeerManager) {
-        if let Some(selected_peer) = self.base_node_watch.borrow().as_ref() {
-            if selected_peer.get_current_peer_id() == base_node_peer_manager.get_current_peer_id() {
-                return;
-            }
-        }
```
hansieodendaal, line 68: Why remove this? Should we not return if the user selects the current peer again?

base_layer/wallet/src/connectivity_service/service.rs
```diff
@@ -271,48 +281,35 @@ impl WalletConnectivityService {
...
         }
-        self.pools.remove(&peer_id);
     }
 
-    async fn setup_base_node_connection(&mut self) {
-        let Some(mut peer_manager) = self.get_base_node_peer_manager() else {
-            debug!(target: LOG_TARGET, "No base node peer manager set");
-            return;
-        };
+    async fn setup_base_node_connection(&mut self, mut peer_manager: BaseNodePeerManager) {
```
hansieodendaal, line 301: Every time this function is called, it should attempt to use the next peer on the list as we have an issue with the peer connection at the current index, except if it is the very first time, in which case it should attempt connecting to the peer at the current peer index.

integration_tests/src/wallet_process.rs
```diff
@@ -226,6 +226,7 @@ pub fn get_default_cli() -> Cli {
         profile_with_tokio_console: false,
         view_private_key: None,
         spend_key: None,
+        reachability: Some(ReachabilityMode::Private),
```
hansieodendaal, line 229: Maybe we need to be more aggressive in establishing a connection for the integration tests? I see many of them are failing.
```suggestion
        reachability: Some(ReachabilityMode::Auto),
```

hansieodendaal (commented):
Hi there @sdbondi, looking good, also running this branch for my testing. Just some comments.

base_layer/wallet/src/connectivity_service/handle.rs
sdbondi: You could also get a different set of peers on the list. It's a little hard to reason about this architecture. We can probably use a watch to notify when a peer ID changes and keep a list of appropriate peers in the connectivity service since that is where we decide to try another peer.

sdbondi: But yes it is a minor issue that we reconnect if we select the same peer.

base_layer/wallet/src/connectivity_service/service.rs
sdbondi: I believe this is what is happening - we move to the next peer after attempting the current

integration_tests/src/wallet_process.rs
sdbondi: Integration tests do not have a public IP so Private is appropriate

applications/minotari_console_wallet/src/lib.rs
sdbondi: Agree, wallets will in general not have a public IP - will change

sdbondi: This is the cli override though, set the default in https://github.com/sdbondi/tari/blob/ff17e520306c5f5993014e455d46a3d44d6de829/base_layer/wallet/src/config.rs

base_layer/wallet/src/connectivity_service/service.rs
hansieodendaal: The essence here is that we should not try the current index if it wasn't due to a base node peer change, but immediately move to the next peer.

sdbondi: :ok_hand: pushed

base_layer/core/src/base_node/service/service.rs
```diff
@@ -304,32 +306,135 @@ where B: BlockchainBackend + 'static
...
-        let source_peer = new_block.origin_or_source();
         let short_ban = self.base_node_config.blockchain_sync_config.short_ban_period;
         let long_ban = self.base_node_config.blockchain_sync_config.ban_period;
         task::spawn(async move {
-            let result = handle_incoming_block(inbound_nch, new_block).await;
+            let new_block = match result {
+                Ok(msg) => msg,
+                Err(err) => {
+                    if let Err(err) = network
+                        .report_gossip_message_validation_result(
```
SWvheerden, line 352: Does this work correctly in that we first report the message as ignore, then we only later report as rejected?
Does the libp2p lib handle this correctly?

base_layer/core/src/base_node/service/service.rs
```diff
@@ -342,7 +447,19 @@ where B: BlockchainBackend + 'static
...
                     }
-                    error!(target: LOG_TARGET, "Failed to handle incoming block message: {}", e)
+                    error!(target: LOG_TARGET, "Failed to handle incoming block message: {}", e);
+                    if let Err(e) = network
+                        .report_gossip_message_validation_result(
+                            message_id,
+                            propagation_source,
+                            // Some other error occurred, we will not propagate, but we'll also not penalise the peer
+                            // for a failure not related to the message contents.
+                            gossipsub::MessageAcceptance::Ignore,
```
SWvheerden, line 457: following the above logic, should we also not reject on a. ban reason?

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -286,8 +286,9 @@ where B: BlockchainBackend
             // lets load the smt into memory
             let mut smt = blockchain_db.smt_write_access()?;
             warn!(target: LOG_TARGET, "Loading SMT into memory from stored db");
+            let timer = Instant::now();
             *smt = blockchain_db.db_write_access()?.calculate_tip_smt()?;
-            warn!(target: LOG_TARGET, "Finished loading SMT into memory from stored db");
+            warn!(target: LOG_TARGET, "Finished loading SMT into memory from stored db in {:.2?}", timer.elapsed());
```
SWvheerden, line 291: haha yeah, we should go back to caching it.
Doing it every block becomes too intensive, but we should have a checkpoint like every 100 blocks or so,

base_layer/core/src/mempool/service/service.rs
```diff
@@ -173,13 +198,36 @@ impl MempoolService {
...
-                error!(
-                    target: LOG_TARGET,
-                    "Failed to handle incoming transaction message: {:?}", e
-                );
+        if let Err(e) = inbound_handlers.handle_transaction(transaction, source_peer_id).await {
+            error!(
+                target: LOG_TARGET,
+                "Failed to handle incoming transaction message: {:?}", e
+            );
+            // NOTE: We assume that all errors are due to a "bad" transaction.
```
SWvheerden, line 206: I feel like this might be a bit broad...
But we can use this for now, just we have to monitor it

base_layer/wallet/src/base_node_service/config.rs
```diff
@@ -41,7 +41,7 @@ impl Default for BaseNodeServiceConfig {
     fn default() -> Self {
         Self {
             base_node_monitor_max_refresh_interval: Duration::from_secs(30),
-            base_node_rpc_pool_size: 10,
+            base_node_rpc_pool_size: 3,
             event_channel_size: 250,
```
SWvheerden, line 48: is this perhaps not too low?
I think a wallet can have the following running at a single time:

- TMS validation
- OMS validation
- base node status check
- base node tip
- UTXO scanning

I am not sure if I miss any here.

network/rpc_framework/src/server/mod.rs
```diff
@@ -512,59 +525,89 @@ where TSvc: Service<Request<Bytes>, Response = Response<Body>, Error = RpcStatus
...
+        match result {
+            Ok(frame) => {
+                #[cfg(feature = "metrics")]
+                metrics::inbound_requests_bytes(&self.peer_id, &self.protocol).observe(frame.len() as f64);
+
+                let start = Instant::now();
+
+                if let Err(err) = self.handle_request(frame.freeze()).await {
+                    if let Err(err) = self.framed.close().await {
+                        // let level = err.io().map(err_to_log_level).unwrap_or(log::Level::Error);
```
SWvheerden:
```suggestion
```

SWvheerden (commented):
Looks good, a few comments and questions

base_layer/core/src/base_node/service/service.rs
sdbondi: Do we? I assume libp2p will error if you report more than once but not 100% sure (just logically inferring it) but in any case I don't think we do that here unless  I'm missing something.

There is a `return` after the above ignore

base_layer/core/src/base_node/service/service.rs
sdbondi: Fair point, though I'm trying to avoid any issues with future gossiping. We handle the validation failure above. So this is not a result of validation failing, we're presumably banning the peer for some other reason which may not be related to the block in the gossip message e.g. IO error when requesting the block.

I would say that reducing their gossip score (which may result in them being removed from the mesh) because we could not connect to their RPC is not correct. However since we were not able to validate the message itself, we cannot Accept it either. Open to thoughts

base_layer/core/src/chain_storage/blockchain_database.rs
sdbondi: lol yeah just curious if the loading of the SMT was the reason for the base node taking forever to start up. Improvement definitely needed here.

base_layer/core/src/mempool/service/service.rs
sdbondi: Yeah could be a problem - let's see.

base_layer/wallet/src/base_node_service/config.rs
sdbondi: RPC is designed to share clients, basically you'll just have to wait for the previous request to complete before yours is started. Since this I've changed the behaviour to drop RPCs once they are not being used so perhaps a good number is the number of services using it - by your count  6. 
Isnt "base node tip" using messaging protocol through the liveness service and not RPC? Will audit this  and come up with a minimal number.

sdbondi: Seems it's used in 8 places, difficult to know how many are required at the same time. In my rather basic testing it never got above 3 and if it is ever a problem in any of the services to wait some bounded time for responses. To be safe I'll set it to 8.

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: Long startup is probably waiting for 5 pings from nodes to decide on a sync peer

sdbondi: no it's before that it takes 20s on our test esme network (from the log above)

base_layer/core/src/base_node/service/service.rs
SWvheerden: a ok, sorry yeah your are correct, we return.

hansieodendaal:
I conducted many system-level stress tests with this branch and feel that the libp2p implementation is pretty solid for almost all use cases, and am happy to approve this PR. Outstanding issues are logged in https://github.com/orgs/tari-project/projects/25/views/1.

hansieodendaal (approved):
Nice.

ACK

base_layer/core/src/chain_storage/blockchain_database.rs
SWvheerden: Is it the smt that takes that long?

PR #6675, opened 2024-11-07
feat(mempool)!: gossip reference to tx and request if needed

sdbondi:
Description
---
- Gossip full transaction if less than 64Kb
- Otherwise, gossip new transaction notification 
- Batch notifications and request transactions using mempool sync in batches every interval
- Raised the message size limit for base node GRPC 

Motivation and Context
---
Transactions routinely are too large for gossipsub. 
A new protocol has been added to mempool sync. Many transactions may be gossiped in quick succession and therefore it is inefficient to request each time a transaction notification is received.  This protocol aims to address this by collecting notifications and periodically requesting many transactions in batched.

Details:

1. NewTransaction gossip message is received, 
2. If it contains the full transaction (it fits into the max message size for gossipsub) the transaction is added as before.
3. Otherwise, the references to new transactions, the propagation peer and message id are collected
4. Every interval the batch is processed, requesting all transactions in a single request
5. Transactions are streamed back and validated. 
6. The result of each of these is collected and the propagation message is either accepted, rejected or ignored depending on the result. This ensures that further propagation of the transaction message only occurs for valid transactions.

Base node GRPC decode and encode limit were raised from 4MiB to 10MiB, this is because some blocks within the legal weight limit were not transferable via GRPC when mining (~6MiB) causing GRPC errors and the inability to mine.

How Has This Been Tested?
---
Manually, submitting transactions and observing logs

---
Submit transactions to one node and observe the mempool of other nodes on the network

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify
BREAKING CHANGE: Changes to the mempool sync protocol are not compatible with previous versions


base_layer/core/src/mempool/config.rs
```diff
@@ -46,12 +48,18 @@ impl SubConfigPath for MempoolConfig {
...
+    /// The maximum number of transactions to sync in a single sync session
     pub initial_sync_max_transactions: usize,
     /// The maximum number of blocks added via sync or re-org to triggering a sync
     pub block_sync_trigger: usize,
+    /// The maximum number of transactions a peer can specifically request.
+    pub max_request_transactions: usize,
+    /// The length of time between checking and, if required, requesting the transactions from the want list.
+    pub request_want_list_interval: Duration,
+    /// Maximum concurrent inbound handler tasks.
+    pub max_concurrent_inbound_tasks: usize,
```
hansieodendaal, line 62: Please add to `c_base_node_c.toml`

sdbondi: I dont think these need to be tweaked by users, generally these should just work. These are here to make it clear what tweaks could be made to the protocol by developers. But welcome to do that in another PR if you feel it will be valuable.

base_layer/core/src/mempool/service/service.rs
```diff
@@ -161,44 +176,151 @@ impl MempoolService {
...
+            return;
+        };
+
+        let transaction = match payload {
+            proto::mempool::new_transaction::Payload::ExcessSig(excess_sig) => {
+                match RistrettoSecretKey::from_canonical_bytes(&excess_sig) {
+                    Ok(excess_sig) => {
+                        let tx_id = MempoolTransactionId::try_from(excess_sig.as_bytes()).unwrap();
+                        match self.inbound_handlers.mempool().has_tx_with_excess_sig(excess_sig).await {
+                            Ok(status) if status.is_new() => {
```
SWvheerden, line 202: dont you need to report [validation](gossipsub::MessageAcceptance::X)  somthing here?
I propose ignore. As I expect us to send a new message out if we added it

base_layer/core/src/mempool/service/service.rs
```diff
@@ -161,44 +176,151 @@ impl MempoolService {
...
+                                }
+                            },
+                            Ok(status) if status.is_stored() => {
+                                // Already stored, we can forward the gossip
+                                if let Err(err) = self
+                                    .network
+                                    .report_gossip_message_validation_result(
+                                        msg.message_id,
+                                        msg.propagation_source,
+                                        gossipsub::MessageAcceptance::Accept,
```
SWvheerden: I think this should be ignore as well. 
The reason for this is, its a message about a tx that we already saw and processed. This means we already notified all peers we are connected to about this tx. So why send them another message about the tx?

base_layer/core/src/mempool/service/service.rs
```diff
@@ -161,44 +176,151 @@ impl MempoolService {
...
+                                    )
+                                    .await
+                                {
+                                    warn!(target: LOG_TARGET, "handle_incoming_tx call network: {err}");
+                                }
+                            },
+                            Ok(status) => {
+                                // Technically unreachable since has_tx_with_excess_sig does not return this
+                                warn!(target: LOG_TARGET, "has_tx_with_excess_sig returned unexpected status {status}");
+                                // Already stored, we can forward the gossip
```
SWvheerden: Comment does not line up with code?

base_layer/core/src/mempool/service/service.rs
```diff
@@ -161,44 +176,151 @@ impl MempoolService {
...
+                                    )
+                                    .await
+                                {
+                                    warn!(target: LOG_TARGET, "handle_incoming_tx call network: {err}");
+                                }
+                            },
+                        }
+                    },
+                    Err(_) => {
+                        debug!(target: LOG_TARGET, "NewTransaction message from gossip contained an invalid excess sig, ignoring");
```
SWvheerden:
```suggestion
                        debug!(target: LOG_TARGET, "NewTransaction message from gossip contained an invalid excess sig, rejecting");
```

SWvheerden (commented):
This looks good, some questions about propagation. 
We have to be careful about what messages we want to send out. This becomes a big issue with large networks.  Libp2p will ensure that a single gossip message is sent to the entire network. So if we will send a message based on that later on, we should not forward the message as this will lead us to send an exponentially increasing number of messages, something akin to (peers ) *(peers-1) for every single ,message if my math is correct here

base_layer/core/src/mempool/service/service.rs
sdbondi: This is passing it onto the mempool sync protocol for resolving, all messages are resolved there as applicable. E.g if we get the transaction from them, then we Accept, if we get an invalid transaction we Reject

base_layer/core/src/mempool/service/service.rs
sdbondi: After discussing this with you offline, I agree with this since this assumes messages were not re-received when in fact gossipsub does not offer perfect deduping (something which would consume unbounded resources)

base_layer/core/src/mempool/service/inbound_handlers.rs
```diff
@@ -67,30 +72,58 @@ impl MempoolInboundHandlers {
...
-                        proto::common::Transaction::try_from(&*tx).map_err(MempoolServiceError::ConversionError)?;
-                    // Gossip the transaction
-                    if let Err(err) = self.gossip_publisher.publish(msg).await {
-                        warn!(
-                            target: LOG_TARGET,
-                            "Error publishing transaction {}: {}.", first_tx_kernel_excess_sig, err
-                        );
+                    let mut transaction_too_large_to_gossip = true;
+                    // TODO: determine more precisely the maximum size of each transaction element
+                    if tx.body.outputs().len() + tx.body.inputs().len() < 4 && tx.body().kernels().len() < 4 {
```
hansieodendaal: Most, almost all, transactions will be small enough to fit within 64 kB, we might as well not do this initial screening to simplify the code.
```suggestion
```

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -317,344 +444,9 @@ impl MempoolSyncProtocol {
...
-        } else {
-            #[cfg(feature = "metrics")]
-            metrics::rejected_inbound_transactions().inc();
-            debug!(
-                target: LOG_TARGET,
-                "Did not store new transaction `{}` in mempool: {}", excess_sig_hex, stored_result
-            )
+        };
+        if self.inbound_tasks.try_push(fut).is_err() {
+            warn!(target: LOG_TARGET, "Rejecting inbound task for peer {peer_id} because we've reached the max_concurrent_inbound_tasks ({})", self.config.max_concurrent_inbound_tasks);
```
hansieodendaal:
```suggestion
            warn!(
                target: LOG_TARGET, 
                "Rejecting inbound task for peer {peer_id} because we've reached the max_concurrent_inbound_tasks ({})",
                self.config.max_concurrent_inbound_tasks
            );
```

base_layer/core/src/mempool/sync_protocol/mod.rs
```diff
@@ -237,23 +364,23 @@ impl MempoolSyncProtocol {
     fn handle_protocol_notification(&mut self, notification: ProtocolNotification<Substream>) {
         match notification.event {
             ProtocolEvent::NewInboundSubstream { peer_id, substream } => {
-                self.spawn_inbound_handler(peer_id, substream);
+                // TODO: we need to limit the number of sessions we handle - switch to using RPC?
```
hansieodendaal: I think this is fine as the number of futures that can be active to handle inbound tasks is defined by `max_concurrent_inbound_tasks`. Also, during stress test conditions, out of 4,050 mempool protocol responders that were started in `start_inbound_handler(&mut self, peer_id: PeerId, substream: Substream)`, only 2 failed due to the peer sending an emoty request. 
```suggestion
```

hansieodendaal (dismissed):
Nice work! Just some small comments noted.

I ran 2x coin-split system-level stress tests containing huge transactions and 1x interactive transaction system-level stress test. Throughout the mempools stayed in sync and all transactions were propagated to mempool peers as expected. Interrogation of the trace logs also unveiled no undue issues.

ACK

base_layer/core/src/mempool/service/inbound_handlers.rs
sdbondi: Hmm yeah I was avoiding the code of `encoded_len` unless needed, but it should be (not evaluated) no more than a few ns of processing for the vast majority so I'll simplify :+1:

base_layer/core/src/mempool/sync_protocol/mod.rs
sdbondi: Thanks, I added a limit to inbound handlers and forgot to delete this comment

PR #6686, opened 2024-11-12
feat!: add new header field

SWvheerden:
Description
---
Add new header field, block output mmr which only tracks outputs in current block

Motivation and Context
---
This is needed in order to validate that all outputs included in the block is correctly captured in the pow of the header. The current strategy is to only use the output smt, which commits to all outputs on the blockchain. This is required for horizon sync and prune mode to verify the blockchain as we sync. 
But someone else wants to verify the block outputs, they have to have the entire outputs set aka entire blockchain in order to verify that the outputs are indeed locked behind pow. 
This addition allows any person to take only the block and know that the outputs contained in that block, is correctly locked behind the pow of the header. 

The output field is constructed as two merkle mountain ranges. 
the primary one used in the header field is constructed as
[coinbase_1, coinbase_2, .., coinbase_x, non_coinbase_ouputs_merkle_root]
with the non_coinbase_ouputs_merkle_root constructed as
[output_1, output_2, .., output_x]

This double mountain range approach allows you to prove the output are in the block header, as well as create a proof for a specific output. But separating them allows p2pool to only need to store a single hash + list of all outputs in order to verify all coin bases are correctly included in the header.  Rather than all outputs or before this pr, the entire block. 

Breaking change
---
Need to resync database


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1372,13 +1377,21 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
...
 
     for kernel in body.kernels() {
         kernel_mmr.push(kernel.hash().to_vec())?;
     }
 
     let mut outputs_to_remove = Vec::new();
+    let mut coinbases = Vec::new();
     for output in body.outputs() {
+        if output.features.is_coinbase() {
+            coinbases.push(output.hash())
```
```suggestion
           block_output_mmr.push(output.hash().to_vec())?;
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -1393,6 +1406,10 @@ pub fn calculate_mmr_roots<T: BlockchainBackend>(
             }
         }
     }
+    for coinbase in coinbases {
+        block_output_mmr.push(coinbase.to_vec())?;
+    }
```
hansieodendaal: Why this intermediate coinbase vector?
```suggestion
```

hansieodendaal (dismissed):
utACK

Looks good, some nits only, apart from CI that should pass

PR #6692, opened 2024-11-15
test!: fix cucumber on libp2p feature branch

hansieodendaal:
Description
---
FIx cucumber tests on libp2p feature branch

_**Notes:**_
- _Instabilities with cucumber when using concurrency other than `--concurrency 1` were detected locally, more so for the wallet FFI suite of tests than the normal tests, but `--concurrency 1`  is recommended for both. This is because addresses are mixed (discoverable) with libp2p when running in the single cucumber process._
- _Outstanding issue #6693 to be addressed in a consecutive PR._

Motivation and Context
---
Some tests were failing

How Has This Been Tested?
---
- `cargo test --release --test cucumber -- --tags "@wallet-ffi and not @broken" --concurrency 1 --retry 2`
- `cargo test --release --all-features --test cucumber -- --tags "@critical and (not @long-running) and (not @wallet-ffi) and (not @chat-ffi) and (not @broken)" --concurrency 1 --retry 2`

---
Code review
Run the cucumber tests

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

BREAKING CHANGE: Wallet FFI interface change on `pub unsafe extern "C" fn comms_config_create`

### DIFF
--- a/integration_tests/src/wallet_process.rs
+++ b/integration_tests/src/wallet_process.rs
@@ -37,7 +37,7 @@ use tari_shutdown::Shutdown;
 use tokio::runtime;
 use tonic::transport::Channel;
 
-use crate::{get_peer_seeds, get_port, wait_for_service, TariWorld};
+use crate::{get_peer_seeds, get_port, wait_for_service, ServiceType, TariWorld};
 
 #[derive(Clone, Debug)]
 pub struct WalletProcess {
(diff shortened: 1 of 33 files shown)


integration_tests/src/wallet_process.rs
```diff
@@ -177,23 +185,25 @@ pub async fn spawn_wallet(
...
         kill_signal: shutdown,
+        udp_port,
     });
 
     tokio::time::sleep(Duration::from_secs(5)).await;
 
-    wait_for_service(port).await;
-    wait_for_service(grpc_port).await;
+    wait_for_service(tcp_port, ServiceType::Tcp, &wallet_name).await;
+    // wait_for_service(udp_port, ServiceType::Udp, &wallet_name).await;
```
SWvheerden:
```suggestion
```

SWvheerden (dismissed):

PR #6696, opened 2024-11-20
fix(tari-pulse): await shutdown signal in main loop

MCozhusheck:
Description
---
Await shutdown signal and reduce error logging when node failed to compare blocks hashes.

Motivation and Context
---
Error logs were spamming on not supported network, greatly reducing readability and shutdown signal wasn't working propely.

How Has This Been Tested?
---
Run local node on esmeralda network (which is not supported yet) and watch for error logs. Also check shutdown behavior.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/src/base_node/tari_pulse_service/mod.rs
+++ b/base_layer/core/src/base_node/tari_pulse_service/mod.rs
@@ -121,25 +121,28 @@ impl TariPulseService {
         notify_passed_checkpoints: watch::Sender<bool>,
     ) {
         let mut interval = time::interval(self.config.check_interval);
-        let mut interval_failed = time::interval(Duration::from_secs(60));
         let mut shutdown_signal = self.shutdown_signal.clone();
 
         loop {
             tokio::select! {
                 _ = interval.tick() => {
                     let passed_checkpoints = match self.passed_checkpoints(&mut base_node_service).await {
-                        Ok(passed) => passed,
+                        Ok(passed) => {
+                            interval = time::interval(self.config.check_interval); // reset interval if back to healthy
+                            interval.tick().await;
+                            passed
+                        },
                         Err(err) => {
-                            error!(target: LOG_TARGET, "Failed to check if node has passed checkpoints: {:?}", err);
-                            let interval_in_secs = interval_failed.period().as_secs();
+                            warn!(target: LOG_TARGET, "Failed to check if node has passed checkpoints: {:?}", err);
+                            let interval_in_secs = interval.period().as_secs();
                             if interval_in_secs > (60 * 30) {
                                 warn!(target: LOG_TARGET, "Reached maximum retry interval of 30 minutes. Exiting");
                                 break;
                             }
-                            error!(target: LOG_TARGET, "Retrying in {} minutes", interval_in_secs/60);
-                            interval_failed.tick().await;
-                            interval_failed.tick().await;
-                            interval_failed = time::interval(Duration::from_secs(interval_in_secs * 2));
+                            info!(target: LOG_TARGET, "Retrying in {} seconds", interval_in_secs);
+                            // increase interval if node repeatly fails to fetch checkpoints
+                            interval = time::interval(Duration::from_secs(interval_in_secs * 2));
+                            interval.tick().await;
                             continue;
                         },
                     };
--- a/base_layer/core/src/base_node/tari_pulse_service/mod.rs
+++ b/base_layer/core/src/base_node/tari_pulse_service/mod.rs
@@ -33,7 +33,7 @@ use hickory_client::{
     rr::{DNSClass, Name, RData, Record, RecordType},
     tcp::TcpClientStream,
 };
-use log::{error, info};
+use log::{error, info, warn};
 use serde::{Deserialize, Serialize};
 use tari_p2p::Network;
 use tari_service_framework::{async_trait, ServiceInitializationError, ServiceInitializer, ServiceInitializerContext};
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -116,21 +121,41 @@ impl TariPulseService {
...
+                                warn!(target: LOG_TARGET, "Reached maximum retry interval of 30 minutes. Exiting");
+                                break;
+                            }
+                            error!(target: LOG_TARGET, "Retrying in {} minutes", interval_in_secs/60);
+                            interval_failed.tick().await;
+                            interval_failed.tick().await;
+                            interval_failed = time::interval(Duration::from_secs(interval_in_secs * 2));
+                            continue;
+                        },
+                    };
```
hansieodendaal: I think this could be simplified, something like
```suggestion
                _ = interval.tick() => {
                    let passed_checkpoints = match self.passed_checkpoints(&mut base_node_service).await {
                        Ok(passed) => {
                            interval_failed = time::interval(Duration::from_secs(0));
                            passed,
                        }
                        Err(err) => {
                            let interval_in_secs = interval_failed.period().as_secs();
                            if interval_in_secs > (60 * 30) {
                                warn!(target: LOG_TARGET, "Reached maximum retry interval of 30 minutes. Exiting");
                                break;
                            }
                            error!(target: LOG_TARGET, "Failed to check if node has passed checkpoints, will retry: {:?}", err);
                            interval_failed += interval;
                            continue;
                        },
                    };
```

hansieodendaal (commented):
Hi, looking good.

Please change
```rust
    // Interval to check if the base node is still in sync with the network
    pub tari_pulse_interval: Duration,
```
to
```rust
    /// Interval to check if the base node is still in sync with the network
    #[serde(with = "serializers::seconds")]
    pub tari_pulse_interval: Duration,
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -116,21 +121,44 @@ impl TariPulseService {
...
-                    interval_failed.tick().await;
-                    continue;
-                },
-            };
+            tokio::select! {
+                _ = interval.tick() => {
+                    let passed_checkpoints = match self.passed_checkpoints(&mut base_node_service).await {
+                        Ok(passed) => {
+                            interval = time::interval(self.config.check_interval); // reset interval if back to healthy
+                            interval.tick().await;
```
hansieodendaal: We do not need another interval tick here, as `_ = interval.tick() => {` is doing it for us
```suggestion
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -116,21 +121,44 @@ impl TariPulseService {
...
+                            let interval_in_secs = interval.period().as_secs();
+                            if interval_in_secs > (60 * 30) {
+                                warn!(target: LOG_TARGET, "Reached maximum retry interval of 30 minutes. Exiting");
+                                break;
+                            }
+                            info!(target: LOG_TARGET, "Retrying in {} seconds", interval_in_secs);
+                            // increase interval if node repeatly fails to fetch checkpoints
+                            interval = time::interval(Duration::from_secs(interval_in_secs * 2));
+                            interval.tick().await;
+                            continue;
```
hansieodendaal: Ditto no need to wait again
```suggestion
                            interval = time::interval(Duration::from_secs(interval_in_secs * 2));
                            continue;
```

hansieodendaal (dismissed):
ACK

Some nits

hansieodendaal:
> Please change
> 
> ```rust
>     // Interval to check if the base node is still in sync with the network
>     pub tari_pulse_interval: Duration,
> ```
> 
> to
> 
> ```rust
>     /// Interval to check if the base node is still in sync with the network
>     #[serde(with = "serializers::seconds")]
>     pub tari_pulse_interval: Duration,
> ```

This must be changed in `common/config/presets/c_base_node_c.toml`

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: First tick is instantaneous according to [docs](https://docs.rs/tokio/latest/tokio/time/fn.interval.html).

Without this await, service quickly reaches limit of 30 minutes.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
MCozhusheck: same as above, we need to `.await` first for next invoke to actually sleep.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
SWvheerden: but thats fine, the first run can be instant, at this moment its waiting for 2 ticks everytime you check it

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -116,21 +121,44 @@ impl TariPulseService {
...
+                            interval = time::interval(self.config.check_interval); // reset interval if back to healthy
+                            interval.tick().await;
+                            passed
+                        },
+                        Err(err) => {
+                            warn!(target: LOG_TARGET, "Failed to check if node has passed checkpoints: {:?}", err);
+                            let interval_in_secs = interval.period().as_secs();
+                            if interval_in_secs > (60 * 30) {
+                                warn!(target: LOG_TARGET, "Reached maximum retry interval of 30 minutes. Exiting");
+                                break;
```
SWvheerden: I would not break here, just leave the interval at max 30 mins

MCozhusheck: Done

MCozhusheck: Also after removing this `tick().await` after assign new interval the loop continues to max value right away.
Screenshot from 2024-11-22 11-59-15

MCozhusheck: Looks like after assigning new interval we have to `tick().await` to make sure that the next iteration of the loop actually awaits

SWvheerden (approved):

PR #6698, opened 2024-11-23
feat: pin hickory-client and hickory-resolver

hansieodendaal:
Description
---
Pined hickory-client and hickory-resolver

Motivation and Context
---
There was an interface change between 
- `hickory-client v0.25.0-alpha.2` and `hickory-client v0.25.0-alpha.3`
- `hickory-resolver v0.25.0-alpha.2` and `hickory-resolver v0.25.0-alpha.3`
and other projects that use `"https://github.com/tari-project/tari.git"` as dependency will resolve to the latter and not the former due to its lock file being generated after that of `"https://github.com/tari-project/tari.git"`.

How Has This Been Tested?
---
All builds succeed

---
Review CI builds

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/base_layer/core/Cargo.toml
+++ b/base_layer/core/Cargo.toml
@@ -93,7 +93,7 @@ tiny-keccak = { package = "tari-tiny-keccak", version = "2.0.2", features = [
     "keccak",
 ] }
 dirs-next = "1.0.2"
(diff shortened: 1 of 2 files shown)


hansieodendaal:
We do not use `hickory-proto` and therefor cannot pin that version as well

PR #6705, opened 2024-11-27
feat: make hickory dns optional

SWvheerden:
Description
---
Hickory dns inclused open ssl which has trouble compiling on all platforms. Making it optional only when needed removes the need to solve this


hansieodendaal (approved):
utACK

leet4tari:
Build test in local fork for ffi, binaries and docker images, looks good.

leet4tari (approved):
Build test in local fork for ffi, binaries and docker images, looks good. ACK

PR #6713, opened 2024-12-05
feat!: add change metadata to encrypted data

hansieodendaal:
Description
---
- Added receiver address, amount, and payment ID to encrypted data so a view-only wallet can identify the receiver when importing a change output.
- Fixed payment ID not propagated for interactive transactions.
- Added payment ID to burn transactions.
- **Edit** Removed `message` from transactions - payment ID is now used consistently throughout

_**Note:** Burn transaction status for change outputs needs to be fixed, see #6726_

Motivation and Context
---
Change data needs to be available for wallet recovery. This is especially useful for view-only wallets.

How Has This Been Tested?
---
System-level testing

**one-sided**

- Sender wallet, one-sided
image

- Receiver wallet, one-sided
image

- View-only wallet, one-sided
image

**interactive**

- Sender wallet, interactive
image

- Receiver wallet, interactive
image

- View-only wallet, interactive
image

**burn**

- Sender wallet, burn
image

- View-only wallet, burn
image

---
Code review
System-level testing

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: 
- Console wallet command line arguments: 
  - `--message` is not supported anymore, only `--payment-id`
- Wallet FFI: 
  - `fn completed_transaction_get_message(` removed
  - `fn pending_outbound_transaction_get_message(` removed
  - `fn pending_outbound_transaction_get_payment_id(` added
  - `fn pending_inbound_transaction_get_message(` removed
  - `fn pending_inbound_transaction_get_payment_id(` added
  - `fn wallet_send_transaction(` has interface change
- Wallet gRPC methods:
  - `message CreateBurnTransactionRequest {` has interface change
  - `message PaymentRecipient {` has interface change
  - `message TransactionInfo {` has interface change
  - `message CoinSplitRequest {` has interface change
  - `message ImportUtxosRequest {` has interface change
  - `message TransactionEvent {` has interface change
  - `message RegisterValidatorNodeRequest {` has interface change


applications/minotari_console_wallet/src/cli.rs
```diff
@@ -265,6 +271,10 @@ pub struct PreMineSpendEncumberAggregateUtxoArgs {
     pub input_file_names: Vec<String>,
     #[clap(long)]
     pub pre_mine_file_path: Option<PathBuf>,
+    #[clap(short, long, default_value = "Spend pre-mine encumber aggregate UTXO")]
+    pub message: String,
```
SWvheerden: Message does not make sense here as these are 1 sided transactions. They don't have messages, they only have payment_id's

applications/minotari_console_wallet/src/cli.rs
```diff
@@ -293,6 +303,10 @@ pub struct PreMineSpendBackupUtxoArgs {
     pub recipient_address: TariAddress,
     #[clap(long)]
     pub pre_mine_file_path: Option<PathBuf>,
+    #[clap(short, long, default_value = "Spend pre-mine backup UTXO")]
+    pub message: String,
```
SWvheerden: same here

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -79,33 +81,65 @@ pub struct EncryptedData {
...
+            PaymentId::U256(v) => {
+                let bytes: &mut [u8] = &mut [0; SIZE_U256];
+                v.to_little_endian(bytes);
+                bytes.to_vec()
+            },
+            PaymentId::Address(v) => v.to_vec(),
             PaymentId::Open(v) => v.clone(),
-            PaymentId::AddressAndData(_v, d) => d.clone(),
+            PaymentId::AddressAndData { data, .. } => data.clone(),
+            PaymentId::ChangeInfo { data, .. } => data.clone(),
```
SWvheerden: This fundamentally changes how its used. 
Paymentid != message
paymentid may include a message
The get_data function is there to only strip out user-supplied data. 
Thats why address etc returns an empta vec and only open returns full data while open and data gets just the data.

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -79,33 +81,65 @@ pub struct EncryptedData {
...
+            } => sender_address.get_size() + id.len(),
+            PaymentId::ChangeInfo {
+                recipient_address,
+                amount: _amount,
+                data: msg_and_id,
+            } => recipient_address.get_size() + SIZE_VALUE + msg_and_id.len(),
         }
     }
 
     pub fn get_data(&self) -> Vec<u8> {
```
SWvheerden: perhaps a better name here will help, like user_data or something in those lines

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -120,62 +154,157 @@ impl PaymentId {
...
+        }
+    }
+
+    /// Helper function to convert a byte slice to a string for the open and data variants
+    pub fn stringify_bytes(bytes: &[u8]) -> String {
+        String::from_utf8_lossy(bytes).to_string()
+    }
+
+    /// Helper function to display the payment id in a compact form
+    pub fn compact_display(&self) -> String {
```
SWvheerden: I think this should only display user data as in actual payment_id, not all the system level detail

base_layer/core/src/transactions/transaction_protocol/proto/transaction_sender.proto
```diff
@@ -40,6 +40,8 @@ message SingleRoundSenderData {
     uint32 kernel_version = 14;
     // the sender address
     string sender_address = 15;
+    // The payment id to receiver
+    bytes payment_id = 16;
```
SWvheerden, line 42:
```suggestion
```
Payment id is not generated by the sender, its generated by the receiver.

base_layer/core/src/transactions/transaction_protocol/sender.rs
```diff
@@ -134,6 +137,8 @@ pub struct SingleRoundSenderData {
     pub metadata: TransactionMetadata,
     /// Plain text message to receiver
     pub message: String,
+    /// A user payment ID for the sender/receiver
+    pub payment_id: PaymentId,
```
SWvheerden, line 137:
```suggestion
```
Sender does not have a payment id, its for the receiver only.

applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -440,16 +440,7 @@ impl TransactionsTab {
...
-                    let bytes = v.get_data();
-                    if bytes.is_empty() {
-                        format!("#{}", v)
-                    } else {
-                        String::from_utf8(bytes)
-                            .unwrap_or_else(|_| "Invalid string".to_string())
-                            .to_string()
-                    }
-                },
+                Some(v) => format!("#{}", v.compact_display()),
```
SWvheerden: users should not see system level detail

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -253,6 +258,54 @@ where KM: TransactionKeyManagerInterface
...
+
+    /// Append message and payment id as bytes
+    pub fn append_message_and_id(&self) -> Option<Vec<u8>> {
+        if let Some(mut message) = self.recipient_text_message.clone() {
+            if let Some(payment_id) = self.payment_id.clone() {
+                message.push_str(" | ");
+                let mut msg_and_id_bytes = message.into_bytes();
+                match payment_id {
+                    PaymentId::Empty => {
+                        let mut empty = "*no payment id*".to_string().as_bytes().to_vec();
```
SWvheerden: if its empty, leave it empty, we should not add bytes to say its empty.

base_layer/core/src/transactions/transaction_protocol/transaction_initializer.rs
```diff
@@ -253,6 +258,54 @@ where KM: TransactionKeyManagerInterface
...
     }
 
+    /// Provide a text message for receiver
+    pub fn with_payment_id(&mut self, payment_id: PaymentId) -> &mut Self {
+        self.payment_id = Some(payment_id);
+        self
+    }
+
+    /// Append message and payment id as bytes
+    pub fn append_message_and_id(&self) -> Option<Vec<u8>> {
```
SWvheerden: We should only include the user data not all data

base_layer/wallet/migrations/2024-12-05-110700-payment_id/up.sql
```diff
@@ -0,0 +1,6 @@
+
+ALTER TABLE inbound_transactions
+    ADD payment_id BLOB NULL;
+
+ALTER TABLE outbound_transactions
+    ADD payment_id BLOB NULL;
```
SWvheerden, line 6: cant see why this is required?
It should be on the outputs already

SWvheerden (changes requested):
Although this PR moves in the right direction there are a few issues here:
1: Payment_id != message. There can be a message inside the payment_id, but its very much user data(message) + system level data. We need to keep separation between the two. This PR completely merges it and makes it all user data.  Looking at the data I suspect there might have been some regression in terms of display here. 

2: On the transmission of the data, the sender should not and can not control what the  receiver pputs in his/her output. So no use in sending anything like that over. That data is sent over, but never actually used in the output. Interactive payments should keep using only the message. Dont add payment id for the receiver there. 

3: On the display, we should not just display the raw data on the transactions. 
For example on the view only wallet. It should not display the incoming data. It should display output going transaction for that amount. Not incoming. It should be more clever how it creates the transaction. It should read the details from the payment id field to create the outgoing tx. It should look almost similar to the spending wallet

4: This is a regression by the looks of it, but we should not display default addresses. We should self, or something like that there.

base_layer/core/src/transactions/transaction_protocol/proto/transaction_sender.proto
hansieodendaal: Agreed to use `payment_id` throughout and to drop `message`

base_layer/core/src/transactions/transaction_protocol/sender.rs
hansieodendaal: Agreed to use `payment_id` throughout and to drop `message`

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
hansieodendaal: Agreed to use payment_id throughout and to drop message

base_layer/wallet/migrations/2024-12-05-110700-payment_id/up.sql
hansieodendaal: To conevert between the different types transactions - inbound, outbound and completed.

applications/minotari_console_wallet/src/cli.rs
hansieodendaal: Agreed to use payment_id throughout and to drop message

applications/minotari_console_wallet/src/cli.rs
hansieodendaal: Agreed to use payment_id throughout and to drop message

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -120,64 +254,165 @@ impl PaymentId {
...
+        String::from_utf8_lossy(bytes).to_string()
+    }
+
+    /// Helper function to display the payment id's user data
+    pub fn user_data_as_string(&self) -> String {
+        match self {
+            PaymentId::Empty => self.to_string(),
+            PaymentId::U64(v) => format!("{}", v),
+            PaymentId::U256(v) => format!("{}", v),
+            PaymentId::Address(v) => v.to_base58(),
```
SWvheerden, line 405: II think this one should probably also display nothing, but its fine for now

SWvheerden (approved):
think is can go in like this for now

PR #6716, opened 2024-12-06
fix: lock version of hickory-client

Misieq01:
### [ Motivation ]

Solves issues in repositories:
`universe`  => https://github.com/tari-project/universe
`tarigpuminer` => https://github.com/stringhandler/tarigpuminer

From what I understand `hickory-client` was implemented on version `0.25.0-alpha.2`. Since that time there was release for version `0.25.0-alpha.3` and `0.25.0-alpha.4` with breaking changes which require code adjustments. Because we didn't lock version to  `0.25.0-alpha.2` here, in mentioned repositories during build when it comes to building tari_core => tari_p2p it tries to use versions higher then `0.25.0-alpha.2` which breaks the build. It also happens in this repository when you delete Cargo.lock file.

To this moment we patched this in mentioned repositories by including these packages with locked versions in Cargo.toml  but it cause other smaller problems


base_layer/core/Cargo.toml
```diff
@@ -94,7 +94,11 @@ tiny-keccak = { package = "tari-tiny-keccak", version = "2.0.2", features = [
     "keccak",
 ] }
 dirs-next = "1.0.2"
-hickory-client = { version = "0.25.0-alpha.2", features = ["dns-over-rustls", "dnssec-openssl"], optional = true }
+hickory-client = { version = "=0.25.0-alpha.2", features = [
+    "dns-over-rustls",
+    "dnssec-openssl",
+], optional = true }
+hickory-proto = { version = "=0.25.0-alpha.2" }
```
SWvheerden, line 101: why is this added in here?

SWvheerden (changes requested):

base_layer/core/Cargo.toml
Misieq01: When I've locked` hickory-client` to  `0.25.0-alpha.2` for some reason this `hickory-proto` still tried to use `alpha 4` version of itself which broke the build so I added entry for this package with also locked version to `0.25.0-alpha.2`

SWvheerden: you are adding a package thats not used by tari. You can see its complaining about it in machete

stringhandler: This is needed because hickory pushed a new version that doesn't compile, so you have to pin it like this.

SWvheerden: please comment why its done there as well, I cant seem to add it via github review

PR #6721, opened 2024-12-10
chore: upgrade pgp and chrono

SWvheerden:
Description
---
upgrades pgp
upgrades chrono to same version used in pgp
Chrono upgrades force datetime change away from naive to datetime<UTC> 

replaces #6717


hansieodendaal (approved):
utACK

PR #6731, opened 2025-01-06
fix: templates

SWvheerden:
Description
---
If the mempool is out of sync, this will give the mempool 10*10ms time to process the latest block to ensure that it is up to sync with the latest blocks.

Motivation and Context
---
Recent changes to templates can cause miners to get invalid templates.  We need to ensure that mempool has processed the latest tip height before asking for a new template as the new template might not be valid

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -257,16 +257,29 @@ where B: BlockchainBackend + 'static
             },
             NodeCommsRequest::GetNewBlockTemplate(request) => {
                 let best_block_header = self.blockchain_db.fetch_tip_header().await?;
-                let last_seen_hash = self.mempool.get_last_seen_hash().await?;
-                let mut is_mempool_synced = true;
-                if last_seen_hash != FixedHash::default() && best_block_header.hash() != &last_seen_hash {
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -257,16 +257,29 @@ where B: BlockchainBackend + 'static
...
-                let last_seen_hash = self.mempool.get_last_seen_hash().await?;
-                let mut is_mempool_synced = true;
-                if last_seen_hash != FixedHash::default() && best_block_header.hash() != &last_seen_hash {
+                let mut last_seen_hash = self.mempool.get_last_seen_hash().await?;
+                let mut is_mempool_synced = false;
+                let mut counter = 0;
+                // this will wait a max of 100ms before returning anyway with a potential broken template
+                // We need to ensure the mempool has seen the latest base node height before we can be confident the
+                // template is correct
+                while !is_mempool_synced && counter < 10 {
```
stringhandler: this could take longer than 100ms if any of the calls use a rwlock. It would be better to use a `tokio::timeout` and create a config variable with the max time to wait

SWvheerden: Changed it to rather than count 10 times, set a max timeout of 150ms that quit

stringhandler (approved):

PR #6732, opened 2025-01-07
feat: cache smt every 1000 blocks

SWvheerden:
Description
---
Caches the smt every 1000 blocks by default or half the prune interval. 

Motivation and Context
---
Loading the smt can take almost 2 mins on nextnet atm. This brings down he loading to 3 secs. 

How Has This Been Tested?
---
Manual tests on nextnet

### DIFF
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -963,7 +963,7 @@ where B: BlockchainBackend
             Ok(v) => v,
             Err(e) => {
                 // some error happend, lets reset the smt to its starting state
-                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db, via calculate root");
+                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db via calculate root");
                 *smt = db.calculate_tip_smt()?;
                 return Err(e);
             },
@@ -2083,7 +2083,7 @@ fn reorganize_chain<T: BlockchainBackend>(
                 );
                 ChainStorageError::AccessError("write lock on smt".into())
             })?;
-            warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db, via reorg");
+            warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db via reorg");
             *write_smt = backend.calculate_tip_smt()?;
             return Err(e);
         }
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -934,6 +934,7 @@ where B: BlockchainBackend
             Ok(v) => v,
             Err(e) => {
                 // some error happend, lets rewind the smt
+                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db via new block prepare");
                 *smt = db.calculate_tip_smt()?;
                 return Err(e);
             },
@@ -962,6 +963,7 @@ where B: BlockchainBackend
             Ok(v) => v,
             Err(e) => {
                 // some error happend, lets reset the smt to its starting state
+                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db, via calculate root");
                 *smt = db.calculate_tip_smt()?;
                 return Err(e);
             },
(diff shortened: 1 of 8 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -962,6 +963,7 @@ where B: BlockchainBackend
             Ok(v) => v,
             Err(e) => {
                 // some error happend, lets reset the smt to its starting state
+                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db, via calculate root");
```
brianp:
```suggestion
                warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db via calculate root");
```

nit for consistency

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2080,6 +2083,7 @@ fn reorganize_chain<T: BlockchainBackend>(
                 );
                 ChainStorageError::AccessError("write lock on smt".into())
             })?;
+            warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db, via reorg");
```
brianp:
```suggestion
            warn!(target: LOG_TARGET, "Reloading SMT into memory from stored db via reorg");
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -261,17 +277,27 @@ pub struct LMDBDatabase {
...
         file_lock: File,
         consensus_manager: ConsensusManager,
+        prune_interval: u64,
     ) -> Result<Self, ChainStorageError> {
         let env = store.env();
+        let smt_cache_period = if prune_interval == 0 {
+            SMT_CACHE_PERIOD
+        } else {
+            prune_interval / 2
+        };
```
brianp, line 300:
```suggestion
        let smt_cache_period = prune_interval.checked_div(2).unwrap_or(SMT_CACHE_PERIOD);
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1341,6 +1384,10 @@ impl LMDBDatabase {
             header.height,
             &BlockAccumulatedData::new(kernel_mmr.get_pruned_hash_set()?, total_kernel_sum),
         )?;
+        allow_smt_change.store(false, Ordering::SeqCst);
```
brianp, line 1387: When does this get reset?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1173,7 +1212,9 @@ impl LMDBDatabase {
         header: &BlockHeader,
         body: AggregateBody,
         smt: Arc<RwLock<OutputSmt>>,
+        allow_smt_change: Arc<AtomicBool>,
     ) -> Result<(), ChainStorageError> {
+        let can_we_change_smt = allow_smt_change.load(Ordering::SeqCst);
```
brianp, line 1217: There is a lot happening in this function outside of whether we can change the SMT or not, and I don't _think_ it matters for most of the function, as we don't end up updating the SMT until the end.

Is this something that we could not do in this function, and somehow test at the caller after this returns, to see fi we also need to call `insert_smt` ?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
SWvheerden: it does not, 
This is to ensure it only ever runs once on smt per transction

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
SWvheerden: This is a weird design detail, and I think the least worse one there is. 
Main issue we have is we have the DB which is the source of truth, now we have the memory cache of the SMT because loading it takes too long. Now we want to create a File copy of the cache as calculating it takes too long. 

So the DB needs to stay the source of truth and always needs to be correct and in sync. 

But all other patterns require multiple rwlock's which might lead to an inconsistent DB state, and or requires manual dev implementation to dave the SMT

brianp (dismissed):

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
SWvheerden: if prune_interval == 0, it use `SMT_CACHE_PERIOD` 
0/2 = 0, not an invalid div

PR #6733, opened 2025-01-07
chore: sort completed transactions

mmrrnn:
## Description

This pull request focuses on refactoring the transaction handling in the wallet service by replacing `HashMap` with `Vec` for storing transactions. This change impacts several parts of the codebase, including transaction service responses, database storage, and transaction handling logic. Additionaly, we sort txs by `mined_timestamp` in desc order.

## Testing
Tested manually using `minotari_console_wallet`

image

### DIFF
--- a/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -1088,15 +1088,17 @@ async fn handle_pending_outbound(
     transaction_service: &mut TransactionServiceHandle,
     sender: &mut Sender<Result<TransactionEventResponse, Status>>,
 ) {
-    match transaction_service.get_pending_outbound_transactions().await {
-        Ok(txs) => {
-            if let Some(tx) = txs.iter().find(|tx| tx.tx_id == tx_id) {
+    use models::WalletTransaction::PendingOutbound;
+    match transaction_service.get_any_transaction(tx_id).await {
+        Ok(tx) => match tx {
+            Some(PendingOutbound(tx)) => {
                 let transaction_event =
                     convert_to_transaction_event(event.to_string(), TransactionWrapper::Outbound(Box::new(tx.clone())));
                 send_transaction_event(transaction_event, sender).await;
-            } else {
+            },
+            _ => {
                 error!(target: LOG_TARGET, "Not found in pending outbound set tx_id: {}", tx_id);
-            }
+            },
         },
         Err(e) => error!(target: LOG_TARGET, "Transaction service error: {}", e),
     }
--- a/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -88,7 +88,7 @@ use minotari_wallet::{
     output_manager_service::{handle::OutputManagerHandle, UtxoSelectionCriteria},
     transaction_service::{
         handle::TransactionServiceHandle,
-        storage::models::{self, CompletedTransaction, WalletTransaction},
+        storage::models::{self, WalletTransaction},
     },
     WalletSqlite,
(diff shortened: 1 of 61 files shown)


SWvheerden (changes requested):
I would not do an order here in rust. 
I would change it where it gets it from the database and use SQL to to do the order. 

Also don't order by timestamp, order by mined_timestamp

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -1089,10 +1089,10 @@ async fn handle_pending_outbound(
     sender: &mut Sender<Result<TransactionEventResponse, Status>>,
 ) {
     match transaction_service.get_pending_outbound_transactions().await {
-        Ok(mut txs) => {
-            if let Some(tx) = txs.remove(&tx_id) {
```
SWvheerden, line 1093: nit: I think it would have been better to rather use the DB. 
Either call get_any_tx with the id, or create a new call to search the db
But I think this is fine for now.

applications/minotari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -730,11 +730,11 @@ impl AppStateInner {
                 .transaction_service
                 .get_pending_outbound_transactions()
                 .await?
-                .values()
+                .iter()
                 .map(|t| CompletedTransaction::from(t.clone()))
                 .collect::<Vec<CompletedTransaction>>(),
         );
-
+        // Question(C): We sort txs by mined_timestamp in sql, here is by timestamp
```
SWvheerden:
```suggestion
```
This is probably much of a muchness. Timestamp = creation date. 
Creation date != mined date. 
Only weird caveat is that mined timestamp could be null, which sqlite will then put at the bottom of the list. If you just created a tx you would want to see it at the top. 

On the view only wallet, in universe, these two values will probably be the same, as the wallet does not talk to other wallets to create transactions

applications/minotari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -747,26 +747,15 @@ impl AppStateInner {
...
                 .transaction_service
                 .get_cancelled_completed_transactions()
-                .await?
-                .values()
-                .cloned()
-                .collect::<Vec<CompletedTransaction>>(),
+                .await?,
         );
-
+        // Question(C): We sort txs by mined_timestamp in sql, here is by timestamp
```
SWvheerden:
```suggestion
```
See above why its better to leave as is here

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1192,6 +1173,8 @@ impl InboundTransactionSql {
     ) -> Result<Vec<InboundTransactionSql>, TransactionStorageError> {
         Ok(inbound_transactions::table
             .filter(inbound_transactions::cancelled.eq(i32::from(cancelled)))
+            // QUESTION(A): Should we order by timestamp or last_send_timestamp?
```
SWvheerden:
```suggestion
```
Think timestamp is fine

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1456,6 +1439,8 @@ impl OutboundTransactionSql {
     ) -> Result<Vec<OutboundTransactionSql>, TransactionStorageError> {
         Ok(outbound_transactions::table
             .filter(outbound_transactions::cancelled.eq(i32::from(cancelled)))
+            // QUESTION(A): Should we order by timestamp or last_send_timestamp?
```
SWvheerden:
```suggestion
```

SWvheerden (commented):
utACK

SWvheerden (approved):

PR #6737, opened 2025-01-09
fix: ledger

SWvheerden:
Description
---
Fixes Ledger CI builds to use spesific image
Fixes ledger builds


leet4tari (approved):
utACK - LGTM

PR #6738, opened 2025-01-10
feat!: remove coinbases from weight calculations

SWvheerden:
Description
---
removes coinbases from the weight calculations making calculations regarding weight simpler. 
Adds max coinbase constant, currently 1000. 

Motivation and Context
---
This will simplify the mining and verification of coinbases and weights, by making coinbases not count towards block weight. 

How Has This Been Tested?
---
unit tests
>

### DIFF
--- a/base_layer/core/src/transactions/aggregated_body.rs
+++ b/base_layer/core/src/transactions/aggregated_body.rs
@@ -294,7 +294,7 @@ impl AggregateBody {
         coinbase_min_maturity: u64,
         factories: &CryptoFactories,
         height: u64,
-        maximum_coinbase_amount: u64,
+        maximum_coinbase_count: u64,
     ) -> Result<(), TransactionError> {
         let mut coinbase_utxo_sum = Commitment::default();
         let mut coinbase_kernel = None;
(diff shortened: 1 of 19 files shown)


base_layer/core/src/transactions/aggregated_body.rs
```diff
@@ -294,6 +294,7 @@ impl AggregateBody {
         coinbase_min_maturity: u64,
         factories: &CryptoFactories,
         height: u64,
+        maximum_coinbase_amount: u64,
```
hansieodendaal: Amount suggests monetary value
```suggestion
        maximum_coinbase_count: u64,
```

base_layer/core/src/transactions/coinbase_builder.rs
```diff
@@ -1050,8 +1056,131 @@ mod test {
...
+        body.sort();
+
+        body.check_coinbase_output(
+            block_reward,
+            rules.consensus_constants(0).coinbase_min_maturity(),
+            &factories,
+            42,
+            1,
+        )
+        .unwrap_err();
```
hansieodendaal, line 1184: I would expand this assertion a bit
```suggestion
        assert!(body.check_coinbase_output(
            block_reward,
            rules.consensus_constants(0).coinbase_min_maturity(),
            &factories,
            42,
            1,
        )
        .is_err());

        assert!(body.check_coinbase_output(
            block_reward,
            rules.consensus_constants(0).coinbase_min_maturity(),
            &factories,
            42,
            2,
        )
        .is_ok());

        assert!(body.check_coinbase_output(
            block_reward,
            rules.consensus_constants(0).coinbase_min_maturity(),
            &factories,
            42,
            3,
        )
        .is_ok());
```

hansieodendaal (changes requested):
I think we need to verify that the maximum block size of 4MB for comms purposes will still be adhered to when we have full blocks with say 1000 coinbases in the block. When we get a new block template for mining this limitation must be taken into account, and not only discovered when we try to transmit such a mined block.

base_layer/core/src/transactions/coinbase_builder.rs
SWvheerden: its already tested above that higher one passes, we want to ensure the lower one fails.

SWvheerden:
Good point about the comms block size, 
I increased this to 6MB to ensure its still safe.  2MB should be more than safe for 1000 coinbases
That 4MB limit is just a value to ensure that no one can keep streaming bytes to the receiver endlessly and fill the memory.

hansieodendaal (dismissed):
utACK

PR #6742, opened 2025-01-15
feat: new grpc method

SWvheerden:
Description
---
Creates a new grpc method to get the reward of the nextblock, the sha + rx estimate hash rates and metadata tip info
Adds caching to grpc calls

Motivation and Context
---
Currently, universe calls get_template to just read the reward which is an expensive operation. 
It also streams up to 100 headers to calculate the estimated hash rate. 

Both of these operations can be made much simpler and faster. This PR provides a call to do just that. 
This Pr also adds cahcing to the calls to only update if an update is required.

### DIFF
--- a/applications/minotari_node/src/grpc/data_cache.rs
+++ b/applications/minotari_node/src/grpc/data_cache.rs
@@ -38,35 +38,35 @@ impl DataCache {
     }
 
     pub async fn get_randomx_estimated_hash_rate(&self, current_tip: &FixedHash) -> Option<u64> {
-        let res = self.inner_data_cache.read().await.randomx_estimated_hash_rate;
-        if res.1 == *current_tip {
-            Some(res.0)
+        let res = &self.inner_data_cache.read().await.randomx_estimated_hash_rate;
+        if res.tip == *current_tip {
+            Some(res.data)
         } else {
             None
         }
     }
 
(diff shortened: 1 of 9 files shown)


hansieodendaal (dismissed):
utACK

Any system-level test comparisons?

applications/minotari_node/src/grpc/data_cache.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    pub async fn set_sha3x_estimated_hash_rate(&self, hash_rate: u64, current_tip: FixedHash) {
+        self.inner_data_cache.write().await.sha3x_estimated_hash_rate = (hash_rate, current_tip);
+    }
+
+    pub async fn get_randomx_new_block_template(&self, current_tip: &FixedHash) -> Option<NewBlockTemplate> {
+        let res = &self.inner_data_cache.read().await.randomx_new_block_template;
+        if res.1 == *current_tip {
+            Some(res.0.clone())
```
brianp: This is just a nit and not a change request.

I always prefer a struct and names values over indexing from a tuple `res.1` or `res.0` as they lose context and meaning in these functions.

brianp (dismissed):
LGTM

One comment, but no required changes.

applications/minotari_node/src/grpc/data_cache.rs
SWvheerden: valid point, let me update

PR #6753, opened 2025-01-20
feat: keep keys in compressed form

SWvheerden:
Description
---
Keeps all crypto public keys in compressed form, and only uncompressing when needed. 

Motivation and Context
---
This should improve speed and reduce the compression of keys. Comparing unit test speeds this PR is roughly the same as before. But unit test almost always do crypto operations to ensure everything is correct. Where this PR increases performance is when only reading something and providing it. For example, just reading the mainnet gen block in its current form  on avg it takes 56 seconds. With this pr this time is taken down to on avg 15 seconds. Thats a 370% speed improvement. 

How Has This Been Tested?
---
unit tests


base_layer/common_types/src/wallet_types.rs
```diff
@@ -60,7 +60,11 @@ pub struct ProvidedKeysWallet {
 impl Display for ProvidedKeysWallet {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         write!(f, "public spend key {}", self.public_spend_key)?;
-        write!(f, "public view key{}", PublicKey::from_secret_key(&self.view_key))?;
+        write!(
```
stringhandler: Actually a problem in the original code. We should now allow heavy code in a `to_string()` method. This should be stored rather

base_layer/common_types/src/types/mod.rs
```diff
@@ -45,18 +46,18 @@ pub use fixed_hash::{FixedHash, FixedHashSizeError};
...
+pub type ComAndPubSignature = CompressedRistrettoComAndPubSig;
 
 /// Define the explicit Commitment implementation for the Tari base layer.
-pub type Commitment = PedersenCommitment;
+pub type Commitment = CompressedPedersenCommitment;
 pub type CommitmentFactory = ExtendedPedersenCommitmentFactory;
 
 /// Define the explicit Public key implementation for the Tari base layer
-pub type PublicKey = RistrettoPublicKey;
+pub type PublicKey = CompressedKey<RistrettoPublicKey>;
```
stringhandler: I think we should change this into `CompressedRistretto`, the others are fine

stringhandler: hmm. Maybe we should add types `Compressed....` so that we can specify in cases where we know they will only be used in a compressed form

comms/core/src/noise/crypto_resolver.rs
```diff
@@ -121,7 +122,8 @@ impl Dh for CommsDiffieHellman {
     }
 
     fn dh(&self, public_key: &[u8], out: &mut [u8]) -> Result<(), snow::Error> {
-        let pk = CommsPublicKey::from_canonical_bytes(&public_key[..self.pub_len()]).map_err(|_| snow::Error::Dh)?;
+        let pk =
```
stringhandler: Just noticed here. Is there a `public_key` that they can pass in that will reveal your secret key? E.g. PublicKey::zero

base_layer/core/src/blocks/genesis_block.rs
```diff
@@ -570,6 +570,26 @@ mod test {
...
 
+    #[test]
+    fn test_speed() {
+        let start = Instant::now();
+        let network = Network::MainNet;
+        for _ in 0..1000 {
+            set_network_by_env_var_or_force_set(network);
+            if !network_matches(network) {
+                panic!("Network could not be set ('mainnet_genesis_sanity_check()')");
+            }
```
hansieodendaal: We only need to set the network once
```suggestion
        set_network_by_env_var_or_force_set(network);
        if !network_matches(network) {
            panic!("Network could not be set ('mainnet_genesis_sanity_check()')");
        }
        for _ in 0..1000 {
```

base_layer/common_types/src/types/mod.rs
hansieodendaal:
```suggestion
pub type PublicKey = RistrettoPublicKey;
pub type CompressedPublicKey = CompressedKey<RistrettoPublicKey>;
```

base_layer/core/src/blocks/pre_mine/mod.rs
```diff
@@ -804,9 +804,12 @@ pub async fn create_pre_mine_genesis_block_info(
         .enumerate()
     {
         let signature_threshold = get_signature_threshold(public_keys.len())?;
-        let total_script_key = public_keys.iter().fold(PublicKey::default(), |acc, x| acc + x);
+        let mut total_script_key = RistrettoPublicKey::default();
```
hansieodendaal: With the proposed change to types ^ we can continue to use `PublicKey`
```suggestion
        let mut total_script_key = PublicKey::default();
```

base_layer/core/src/blocks/pre_mine/mod.rs
```diff
@@ -804,9 +804,12 @@ pub async fn create_pre_mine_genesis_block_info(
...
+        for key in public_keys{
+            total_script_key = total_script_key + key.to_public_key().unwrap();
+        }
         let key_manager = create_memory_db_key_manager().unwrap();
-        let view_key = public_key_to_output_encryption_key(&total_script_key).unwrap();
+        let view_key = public_key_to_output_encryption_key(&PublicKey::new_from_pk(&total_script_key)).unwrap();
```
hansieodendaal:
```suggestion
        let view_key = public_key_to_output_encryption_key(&CompressedPublicKey::new_from_pk(&total_script_key)).unwrap();
```

base_layer/core/src/blocks/pre_mine/mod.rs
SWvheerden: I dont want to use public key at all, as it has a conflict with the trait name

base_layer/common_types/src/wallet_types.rs
SWvheerden: yeah, I removed it from the display as you can do nothing with the public view key, so why display it in the first place

PR #6757, opened 2025-01-23
fix: the smt cache height

SWvheerden:
Description
---
Currently, the smt by default is cached every 25 blocks, which is too much. reduce this down to 500 by default for archival nodes.

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -301,14 +301,10 @@ impl LMDBDatabase {
         pruning_horizon: u64,
     ) -> Result<Self, ChainStorageError> {
         let env = store.env();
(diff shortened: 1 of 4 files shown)


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -291,9 +298,10 @@ impl LMDBDatabase {
         file_lock: File,
         consensus_manager: ConsensusManager,
         prune_interval: u64,
+        pruning_horizon: u64,
     ) -> Result<Self, ChainStorageError> {
         let env = store.env();
-        let smt_cache_period = if prune_interval == 0 {
+        let smt_cache_period = if pruning_horizon == 0 || prune_interval == 0 {
```
stringhandler, line 304: maybe add a comment here what the thinking is, because it's not clear why the interval / 2

stringhandler (dismissed):

PR #6765, opened 2025-01-27
fix: wallet validation

SWvheerden:
Description
---
Fixes wallet validation to always correctly trigger

Motivation and Context
---
Currently when a new block comes, validation is triggered, but if validation is already running, nothing happens. We need to stop the current validation as we need to validate with the newest information. 

How Has This Been Tested?
---
Manual


hansieodendaal (dismissed):
utACK

**Note:** This process may struggle to finish for big wallets if it is restarted every time a block is mined.

SWvheerden:
It will still finish, it wont revalidate outputs its already validated. And when this stops, it does not delete progress on validation it already made.

PR #6767, opened 2025-01-27
feat: fix mempool sync start early

### LINKED ISSUE #6766: Base node initial sync triggers too early
The base node will trigger initial sync and say its synced, then start to do a sync from a node:

To test, run a node and see this in the log line: `Initial sync is done. Starting mempool sync protocol`
After which it will start syncing. 

Related to this:
https://github.com/tari-project/universe/issues/1348

hansieodendaal:
Description
---
Mempool sync can only start when `is_synced` is set to `true`.
- An edge case was witnessed whereby mempool sync can start early if the entire network is generally
slow to respond to pings. The `ChainMetadataEvent::NetworkSilence` event is transmitted if 3x
consecutive ping cycles were started without corresponding pongs received, which sets
`is_synced` to `true`. The moment that peer metadata is then received, block sync and mempool sync
will occur at the same time. 
- Prolonged network silence is now detected in the `Starting` state. At startup, the first successful metadata message received from a peer will signal the transition to `Listening`, otherwise `Listening` will be notified of the network silence, which can act accordingly.
- Additional debug logs were added to track all events that can set the `is_synced` to `true`.
- Removed `log-mdc` from the dependencies as the relevant code is not used.

The edge case warning message must be monitored to see if it is a real issue:
```rust
warn!(
    target: LOG_TARGET,
    "Initial sync achieved based on event 'NetworkSilence'; this may not be true if the entire \
    network in general is slow to respond to pings"
);
```

Fixes #6766

Motivation and Context
---
Mempool sync _"started"_ before block sync was achieved.

How Has This Been Tested?
---
System-level testing

---
Code review
System-level testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -340,6 +340,9 @@ impl From<Waiting> for Listening {
 
 impl From<HeaderSyncState> for Listening {
     fn from(sync: HeaderSyncState) -> Self {
+        if sync.is_synced() {
+            debug!(target: LOG_TARGET, "Initial sync achieved (from HeaderSyncState)");
+        }
```
SWvheerden:
```suggestion
```

base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -349,6 +352,9 @@ impl From<HeaderSyncState> for Listening {
 
 impl From<BlockSync> for Listening {
     fn from(sync: BlockSync) -> Self {
+        if sync.is_synced() {
+            debug!(target: LOG_TARGET, "Initial sync achieved (from BlockSync)");
+        }
```
SWvheerden:
```suggestion
```

base_layer/core/src/base_node/state_machine_service/states/listening.rs
```diff
@@ -358,6 +364,9 @@ impl From<BlockSync> for Listening {
 
 impl From<DecideNextSync> for Listening {
     fn from(sync: DecideNextSync) -> Self {
+        if sync.is_synced() {
+            debug!(target: LOG_TARGET, "Initial sync achieved (from DecideNextSync)");
+        }
```
SWvheerden:
```suggestion
```

SWvheerden (dismissed):

PR #6773, opened 2025-01-30
fix: wallet migration

SWvheerden:
Description
---
Small fix for the key manager table to enable migration of wallets from pre compressed pr to dev


hansieodendaal (approved):
utACK

PR #6784, opened 2025-02-05
fix: validation trigger

SWvheerden:
Description
---
Lets validation triggers not on new block, but rather on scanning complete of new block

Motivation and Context
---
We should not trigger a new validation after a block has been detected, but rather scanned


hansieodendaal (commented):
- The coding logic looks good.
- I conducted system-level tests, wallet recovery as well as mining with coinbase scanning, all 
 went well.
- _Failing unit and cucumber tests._

brianp:
I tested in TU in combination with @mmrrnn's changes and successfully see win/fail animations at the correct times.

PR #6795, opened 2025-02-10
feat: add p2pool tip info and better mmproxy template handling

stringhandler:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new RPC method to retrieve tip information from the P2Pool service.
	- Added a new variable for user agent configuration in the HTTP client.
- **Refactor**
	- Streamlined block template management with a new management structure.
	- Simplified method signatures and reduced parameter complexity in several service methods.
- **Bug Fixes**
	- Improved error handling in the retrieval of node and network state metrics.
- **Chores**
	- Updated project dependencies and made minor formatting adjustments.
	- Renamed modules for better clarity and organization.

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
+++ b/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
@@ -133,12 +133,6 @@ impl BlockTemplateManager<'_> {
             },
         };
 
-        let height = block
-            .block
-            .as_ref()
-            .map(|b| b.header.as_ref().map(|h| h.height).unwrap_or_default())
-            .unwrap_or_default();
-
         let miner_data = block
             .miner_data
             .as_ref()
--- a/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
+++ b/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
@@ -51,7 +51,7 @@ use crate::{
 const LOG_TARGET: &str = "minotari_mm_proxy::proxy::block_template_protocol";
 
 /// Structure holding grpc connections.
-pub struct BlockTemplateManager<'a> {
+pub(crate) struct BlockTemplateManager<'a> {
     config: Arc<MergeMiningProxyConfig>,
     base_node_client: &'a mut BaseNodeGrpcClient,
     p2pool_client: Option<ShaP2PoolGrpcClient>,
@@ -80,7 +80,6 @@ impl<'a> BlockTemplateManager<'a> {
     }
 }
 
-#[allow(clippy::too_many_lines)]
 impl BlockTemplateManager<'_> {
     /// Create [FinalBlockTemplateData] with [MoneroMiningData].
     pub async fn get_next_tari_block_template(
@@ -330,7 +329,7 @@ fn add_monero_data(
 
 /// Private convenience container struct for new template data
 #[derive(Debug, Clone)]
-pub struct NewBlockTemplateData {
+pub(crate) struct NewBlockTemplateData {
     pub template: grpc::NewBlockTemplate,
     pub miner_data: grpc::MinerData,
 }
@@ -342,10 +341,10 @@ impl NewBlockTemplateData {
 }
 
 /// The AuxChainMerkleRoot is a 32 byte hash
-pub type AuxChainMr = MaxSizeBytes<32>;
+pub(crate) type AuxChainMr = MaxSizeBytes<32>;
 /// Final outputs for required for merge mining
 #[derive(Debug, Clone)]
-pub struct FinalBlockTemplateData {
+pub(crate) struct FinalBlockTemplateData {
     pub template: BlockTemplateData,
     pub target_difficulty: Difficulty,
     pub blockhashing_blob: String,
@@ -356,7 +355,7 @@ pub struct FinalBlockTemplateData {
 
 /// Container struct for monero mining data inputs obtained from monerod
 #[derive(Clone)]
-pub struct MoneroMiningData {
+pub(crate) struct MoneroMiningData {
     pub seed_hash: FixedByteArray,
     pub blocktemplate_blob: String,
     pub difficulty: u64,
--- a/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
+++ b/applications/minotari_merge_mining_proxy/src/block_template_manager.rs
@@ -42,7 +42,7 @@ use tari_max_size::MaxSizeBytes;
 use tari_utilities::{hex::Hex, ByteArray};
 
 use crate::{
-    block_template_data::{BlockTemplateData, BlockTemplateDataBuilder, BlockTemplateRepository},
+    block_template_data::{BlockTemplateData, BlockTemplateDataBuilder},
     common::merge_mining,
     config::MergeMiningProxyConfig,
     error::MmProxyError,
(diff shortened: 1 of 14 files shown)


applications/minotari_merge_mining_proxy/src/block_template_manager.rs
```diff
@@ -81,224 +81,115 @@ impl<'a> BlockTemplateProtocol<'a> {
...
-                    .map(|h| h.height)
-                    .unwrap_or_default();
-                debug!(target: LOG_TARGET, "Used existing new block template at height: #{} (try {})", height, loop_count);
-                (template_with_coinbase, height)
-            },
-        };
-        Ok((block_template_with_coinbase, height))
+                        // return Err(MmProxyError::FailedToGetBlockTemplate(
+                        // "mempool not in sync".to_string(),
+                        // ));
```
SWvheerden:
```suggestion
```

applications/minotari_merge_mining_proxy/src/run_merge_miner.rs
```diff
@@ -111,10 +111,13 @@ pub async fn start_merge_miner(cli: Cli) -> Result<(), anyhow::Error> {
...
 
     info!(target: LOG_TARGET, "Configuration: {:?}", config);
+    let agent = concat!("minotari_mm_proxy/", env!("CARGO_PKG_VERSION"));
     let client = reqwest::Client::builder()
         .connect_timeout(Duration::from_secs(5))
         .timeout(Duration::from_secs(10))
-        .pool_max_idle_per_host(25)
+        .user_agent(agent)
+        .tcp_keepalive(Duration::from_secs(60))
+        // .pool_max_idle_per_host(25)
```
SWvheerden:
```suggestion
```

SWvheerden (commented):
Looking good, 
Just needs a green ci

applications/minotari_merge_mining_proxy/src/run_merge_miner.rs
stringhandler: applied, thanks

applications/minotari_merge_mining_proxy/src/proxy/inner.rs
```diff
@@ -112,39 +112,53 @@ impl InnerService {
...
-            .map(|meta| meta.best_block_height)
-            .ok_or(MmProxyError::GrpcResponseMissingField("base node metadata"))?;
-        if result.get_ref().initial_sync_achieved != self.initial_sync_achieved.load(Ordering::SeqCst) {
-            self.initial_sync_achieved
-                .store(result.get_ref().initial_sync_achieved, Ordering::SeqCst);
-            debug!(
-                target: LOG_TARGET,
-                "Minotari base node initial sync status change to {}",
-                result.get_ref().initial_sync_achieved
-            );
```
hansieodendaal, line 139: I think we should keep this

applications/minotari_merge_mining_proxy/src/config.rs
```diff
@@ -177,7 +177,7 @@ impl Default for MergeMiningProxyConfig {
             range_proof_type: RangeProofType::RevealedValue,
             p2pool_enabled: false,
             monerod_fallback: Default::default(),
-            monerod_connection_timeout: Duration::from_secs(2),
+            monerod_connection_timeout: Duration::from_secs(105),
```
hansieodendaal: This seems huge, especially for the default value - I would only go as high as 5s. Where only a single monerod server is assigned in `monerod_url` (_non_default use like in Tari Universe_), this big value could be assigned during start-up.

hansieodendaal (commented):
I agree mostly with the logic, except for some subtleties as noted.

_System-level testing in progress_

applications/minotari_merge_mining_proxy/src/proxy/inner.rs
hansieodendaal: Yes, the merge mining proxy now reports initial sync being achieved when it has not been:
```
2025-02-20 16:01:46.863171000 [minotari_mm_proxy::proxy::inner] DEBUG Initial base node sync achieved. Ready to mine at height #2277
```
vs.
```
16:01 v1.11.5-pre.0 esmeralda State: Syncing blocks: (dbbcb376e13725c5) 2766/7070 (39%)
```

stringhandler: this piece of code merely sets the initial sync achieved atomicbool and it is set in a different place anyway

SWvheerden: I think its fine to remove this. 
We are mining invalid tari blocks yes, but we are not mining invalid monero blocks. We should not block the mining of monero while the tari block syncs.

hansieodendaal: If we do not care about this, the config option to mine only after sync is achieved should be removed. As it currently stands, the logic does not work anymore.

applications/minotari_merge_mining_proxy/src/proxy/inner.rs
```diff
@@ -112,39 +112,53 @@ impl InnerService {
...
+        } else {
+            let mut base_node_client = self.base_node_client.clone();
+            trace!(target: LOG_TARGET, "Successful connection to base node GRPC");
 
-        let result =
-            base_node_client
-                .get_tip_info(tari_rpc::Empty {})
-                .await
-                .map_err(|err| MmProxyError::GrpcRequestError {
+            let result = base_node_client.get_tip_info(tari_rpc::Empty {}).await.map_err(|err| {
```
hansieodendaal, line 134: This result must be used to assign values to `base_node_height` and `base_node_hash` - as a result the info log below displays incorrectly, and these lines are broken:
```rust
        // Add them together. You could multiply them a factor, but xmrig will generally just check if the height or the
        // hash is different.
        let reported_height = p2pool_height + monero_height + base_node_height;

        // As of xmrig 6.22.0, the block hash is stored separately and does not need to match the block template they
        // are mining, but if that changes in future, you might need to return the monero hash.

        let hash: Vec<u8> = Blake2s256::new()
            .chain(&monero_hash)
            .chain(&base_node_hash)
            .chain(&p2pool_hash)
            .finalize()
            .to_vec();
```

applications/minotari_merge_mining_proxy/src/proxy/inner.rs
```diff
@@ -112,39 +112,53 @@ impl InnerService {
...
         info!(
             target: LOG_TARGET,
-            "Monero height = #{}, Minotari base node height = #{}", json["height"], height
+            "Monero height = #{}, Minotari base node height = #{}, P2pool height: #{}", monero_height, base_node_height, p2pool_height
         );
 
-        json["height"] = json!(cmp::max(json["height"].as_i64().unwrap_or_default(), height as i64));
+        // Add them together. You could multiply them a factor, but xmrig will generally just check if the height or the
+        // hash is different.
+        let reported_height = p2pool_height + monero_height + base_node_height;
```
hansieodendaal, line 156: Let's move the info log to below the `reported_height` calculation and include it in the log as well.

hansieodendaal (commented):
I have concluded my system-level testing. It works well for the best part, except for my two main comments:
- The default value for `monerod_connection_timeout` should not be more than 5s.
- When using the base node client the `base_node_height` and `base_node_hash` are not feeding back into XMRig due to those values being unassigned.

hansieodendaal (dismissed):
ACK

SWvheerden (approved):

PR #6850, opened 2025-03-07
fix!: monero seed validation

SWvheerden:
Description
---
Fixes monero seed validation to always check seeds. 

Motivation and Context
---
Block 26320 is  bad block breaking consensus. It has a RandomX VM key of 91ef83186cefaa646dc4c6e950e68e4debab52b4f4a9b7f465891e91fe5f6ce4 
this key was used between height 2729 and 843, which is about what you would expect from the monero consensus but was reused only once on block 26320 with hash 999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e. 

This PR will bypass the hash checking so that we can accept this block in the chain, but only this block. It also fixes the validation bug that caused the RandomX Seeds to not be added the database for tracking. 

How Has This Been Tested?
---
Manual sync

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

Wil whitelist a bad block to be accepted into the chain, and clear out all bad blocks listed

## Summary by CodeRabbit

- **Refactor**
  - Streamlined block processing and chain reorganization by removing legacy consensus parameters.

- **New Features**
  - Enhanced the handling and logging of specialized proof-of-work data in block headers.
  - Updated the migration process to support a new database schema version and to clear legacy records.
  - Introduced a mechanism to allow specific block headers to bypass certain validation checks, ensuring smoother validation flow.

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -19,7 +19,6 @@
 // SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
 use std::{
     cmp::max,
     convert::TryFrom,
@@ -126,6 +125,7 @@ use crate::{
     },
     consensus::{ConsensusConstants, ConsensusManager},
     output_mr_hash_from_smt,
+    proof_of_work::{monero_rx::MoneroPowData, PowAlgorithm},
     transactions::{
         aggregated_body::AggregateBody,
         transaction_components::{
@@ -860,6 +860,22 @@ impl LMDBDatabase {
             )));
         } else {
             // we can continue
+        }
+
+        if header.pow_algo() == PowAlgorithm::RandomX {
+            let monero_header = MoneroPowData::from_header(&header, &self.consensus_manager).map_err(|e| {
+                ChainStorageError::InvalidArguments {
+                    func: "insert_best_block",
+                    arg: "block",
+                    message: format!("block contained invalid or malformed monero PoW data: {}", e),
+                }
+            })?;
+            let vm_key = monero_header.randomx_key.to_vec();
+            trace!(
+                target: LOG_TARGET,
+                "inserting monero vm key: {} for height {}",vm_key.to_hex(), header.height
+            );
+            self.insert_monero_seed_height(&txn, &vm_key, header.height)?;
         }
 
         lmdb_insert(
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -863,7 +863,7 @@ impl LMDBDatabase {
         }
 
         if header.pow_algo() == PowAlgorithm::RandomX {
-            let monero_header = MoneroPowData::from_header(&header, &self.consensus_manager).map_err(|e| {
+            let monero_header = MoneroPowData::from_header(header, &self.consensus_manager).map_err(|e| {
                 ChainStorageError::InvalidArguments {
                     func: "insert_best_block",
                     arg: "block",
@@ -875,7 +875,7 @@ impl LMDBDatabase {
                 target: LOG_TARGET,
                 "inserting monero vm key: {} for height {}",vm_key.to_hex(), header.height
             );
-            self.insert_monero_seed_height(&txn, &vm_key, header.height)?;
+            self.insert_monero_seed_height(txn, &vm_key, header.height)?;
         }
 
         lmdb_insert(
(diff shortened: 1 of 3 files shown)


base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -70,16 +78,19 @@ impl<B: BlockchainBackend> HeaderChainLinkedValidator<B> for HeaderFullValidator
...
-        check_not_bad_block(db, header.hash())?;
-        check_blockchain_version(constants, header.version)?;
-        check_height(header, prev_header)?;
-        check_prev_hash(header, prev_header)?;
+        // dont run these checks for blocks we have whitelisted
+        if !WHITELISTED_HEADERS.contains(&header.hash().to_hex().as_str()) {
+            check_not_bad_block(db, header.hash())?;
+            check_blockchain_version(constants, header.version)?;
+            check_height(header, prev_header)?;
+            check_prev_hash(header, prev_header)?;
```
hansieodendaal: I think we need to be more specific here and not exclude all validation for the whitelisted header.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -862,6 +862,22 @@ impl LMDBDatabase {
             // we can continue
         }
 
+        if header.pow_algo() == PowAlgorithm::RandomX {
```
hansieodendaal, line 865: Yes, it makes sense to do it here

hansieodendaal (changes requested):
The first part looks good

I have some doubts about the second part, though. I do not think we should be so broad when excluding whitelisted header validation, as it could open us up to other attacks, for example slipping in a completely different header and then we just accept it. If we know the header that we are happy with, just do a byte comparison for the full incoming header identified by hash so no other data of that header could be changed.

base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -41,6 +41,14 @@ use crate::{
 
 pub const LOG_TARGET: &str = "c::val::header_full_validator";
 
+// 999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e is a bad block that bypassed validation due to a bug
```
stringhandler:
```suggestion
// 999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e is a bad block on NEXTNET that bypassed validation due to a bug
```

base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -41,6 +41,14 @@ use crate::{
...
+// in saving Monero seeds. The block uses a randomX  VM key of
+// 91ef83186cefaa646dc4c6e950e68e4debab52b4f4a9b7f465891e91fe5f6ce4, this key was used between height 2729 and 843 which
+// is about what you would expect from the Monero consensus. This block reuses this key at height 26320, which is a
+// violation of the Monero consensus rules. But in order to keep the network on the same chain, we whitelist this block
+// to bypass validation as that is the only validation it has failed.
+pub const WHITELISTED_HEADERS: [&str; 1] = ["999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e"];
```
stringhandler:
```suggestion
#[cfg(nextnet)]
pub const WHITELISTED_HEADERS: [&str; 1] = ["999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e"];
```

stringhandler: i.e. let's only add this exception for nextnet. I would also be satisfied with a comment 
```
[
// Nextnet height 22222
"99978..."
]
```

base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -70,16 +78,19 @@ impl<B: BlockchainBackend> HeaderChainLinkedValidator<B> for HeaderFullValidator
...
-        sanity_check_timestamp_count(header, prev_timestamps, constants)?;
-        check_header_timestamp_greater_than_median(header, prev_timestamps)?;
+            sanity_check_timestamp_count(header, prev_timestamps, constants)?;
+            check_header_timestamp_greater_than_median(header, prev_timestamps)?;
 
-        check_timestamp_ftl(header, &self.rules)?;
-        check_pow_data(header, &self.rules, db)?;
+            check_timestamp_ftl(header, &self.rules)?;
+            check_pow_data(header, &self.rules, db)?;
+        }
```
stringhandler, line 107:
```suggestion
        }
        else {
            info!("Whitelisted block .... encountered")
        }
```

base_layer/core/src/validation/header/header_full_validator.rs
stringhandler: There is no point doing any validation if the whole header is whitelisted

base_layer/core/src/validation/header/header_full_validator.rs
```diff
@@ -41,6 +41,14 @@ use crate::{
 
 pub const LOG_TARGET: &str = "c::val::header_full_validator";
 
+// 999785e3bb6a43189c7236de7c9720df27d56f6b9724cd22f5115a097d3f770e is a bad block that bypassed validation due to a bug
+// in saving Monero seeds. The block uses a randomX  VM key of
+// 91ef83186cefaa646dc4c6e950e68e4debab52b4f4a9b7f465891e91fe5f6ce4, this key was used between height 2729 and 843 which
+// is about what you would expect from the Monero consensus. This block reuses this key at height 26320, which is a
+// violation of the Monero consensus rules. But in order to keep the network on the same chain, we whitelist this block
+// to bypass validation as that is the only validation it has failed.
```
stringhandler, line 50:
```suggestion
// to bypass validation as that is the only validation it has failed.
// Remove this block if Nextnet is reset
```

hansieodendaal: Maybe we need to pin it to the specific genesis block hash in play, then it is not important if we omit to remove it.

hansieodendaal: The comment should also talk about _header_ not _block_.

PR #6879, opened 2025-03-14
fix!: median timestamp check

SWvheerden:
Description
---
Fixes the median timestamp check to correctly check the that the timestamp needs to be > than he median

Motivation and Context
---
If the check is only >=, then it allows an attacked to keep the chain at the median timestamp, thus descreasing the minining difficulty 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved block header timestamp validations to ensure headers are appropriately incremented compared to network medians.
	- Enhanced the logic to iteratively adjust timestamps until they meet stricter criteria.
	- Updated warning and error messages to provide clearer feedback when timestamp conditions are not met.


hansieodendaal (approved):
utACK

This might just do the trick, good catch.

PR #6880, opened 2025-03-14
feat: check coinbase count

SWvheerden:
Description
---
Coinbases counts are checked as part of validation. 
They are not checked as part of new block template creation. 
This checks the consensus as part of creation so as to stop miners from create invalid blocks 

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error handling for coinbase entry limits during block template and block creation. Users will receive a clear error message if the number of coinbase entries exceeds the allowed limit, ensuring adherence to consensus rules.


hansieodendaal (commented):
utACK

PR #6908, opened 2025-04-02
perf: remove saf and various other improvements

stringhandler:
1. Removes SAF from inbound pipeline
2. Adds timing and warnings when inbound pipeline takes too long
3. Increase a few channel sizes
4. Decrease the timeout for output pipeline processing. Note: this should be strictly shorter than the inbound pipeline. The inbound can get blocked by waiting on output processing.

## Summary by CodeRabbit

- **New Features**
  - Enhanced message filtering to streamline communication routing.

- **Refactor**
  - Transitioned internal messaging channels to an unbounded design for improved throughput.
  - Upgraded logging and timing checks to better detect processing delays.

- **Chores**
  - Increased default limits for concurrent tasks and adjusted message buffering for enhanced stability.
  - Updated transaction query labels for clearer wallet interactions.

- **Tests**
  - Streamlined test setups and temporarily marked obsolete scenarios as ignored.


hansieodendaal (changes requested):
**Edited**
- The additional logging around pipeline operations is a good addition.
- Removing SAF will be more performant.
~~- Apart from changing limits for maximum concurrent inbound and outbound, the rest may not be needed. In my system-level testing, the seed node utilized up to 99 inbound and 49 outbound concurrent pipelines.~~
~~- I can see why SAF is not needed for a seed node use case, but it is still an integral part of interactive transactions UX.~~
~~- Why do we need to introduce unbounded, e.g. `UnboundedSender`, everywhere?~~
- I have not seen any evidence in testing suggesting that the outbound pipeline influences the inbound pipeline. Sometimes outbound tasks can take more than 1s and finish successfully. The `match time::timeout(Duration::from_secs(1), pipeline.oneshot(msg)).await` timeout is too strict.
- ~~Seeing that we remove SAF in this PR, `pub struct PeerFeatures:` must be changed as well, removing `DHT_STORE_FORWARD`.~~ _PeerFeatures::DHT_STORE_FORWARD (or its bits) must remain intact otherwise the node public key will change; this will effectively stop communication with all other nodes on the network._

hansieodendaal (changes requested):
I performed a long-running system-level test on a seed node and a normal base node. Here are the results:
- Seed node (with 250 concurrent inbound and outbound pipelines):
  - The seed node was started at ~18:30 and at ~19:00 started having issues with the inbound pipeline timing out (set at 5s).
  - All normal/successful inbound pipeline processing stopped at ~19:00.
  (compare regex `Finished inbound pipeline .*? in \d+(\.\d+)?s\b` with regex `Finished inbound pipeline .*? in \d+(\.\d+)?ms\b`)
  - The inbound pipeline timeouts coincided with the base node console stopping to respond with the `Watched command `status` failed: deadline has elapsed` message, and never recovered.
  - From ~19:00 - ~02:20 a total of 3240 inbound pipeline timeouts occurred.
  - The outbound pipeline only timed out once (set at 1s), a short while after startup.
  - Inbound pipeline usage peaked at 38/250 before the first timeout and at 42/250 after the first time-outs.
  - Outbound pipeline usage peaked at 2/250.
- Normal node (with 50 concurrent inbound and outbound pipelines)
  - The node continued to operate normally until ~07:20 when it was shut down.
  - The inbound pipeline timed out 3 times (set at 5s).
  - The outbound pipeline timed out 231 times (set at 1s).
  - Inbound pipeline usage peaked at 50/50 (438 counts).
  - Outbound pipeline usage peaked at 40/50 (1 count).
- General
  - SAF processes are still running.
  - The interlock prevented the seed base node from shutting down normally as the terminal window had to be closed.

Summary of findings:
- We have an interlock that stops the node operating as a seed node after a short while.
- Having 50 concurrent pipelines are not enough for a seed node or a normal node.
- No correlation was found between outbound and inbound timeouts; the timeouts can be set equal.
- A 1s timeout on the outbound pipeline is not feasible.
- SAF should be removed completely from the code.

SWvheerden:
closing for #6936

PR #6916, opened 2025-04-03
chore(ci): build docker with metrics

ns212:
Description
---
Build minotari_node with metrics feature enabled; Remove unused feature flags;

Motivation and Context
---

How Has This Been Tested?
---
Tested on a seed node.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
	- Updated the default container configuration to include an expanded set of capabilities for enhanced performance and monitoring.
	- Improved container startup reliability with a new process management component that ensures smoother signal handling and process reaping.

### DIFF
--- a/.github/workflows/build_dockers_workflow.yml
+++ b/.github/workflows/build_dockers_workflow.yml
@@ -20,7 +20,7 @@ name: Build docker images - workflow_call/on-demand
         default: x86-64
       features:
         type: string
-        default: 'safe,grpc'
+        default: 'safe,libtor,metrics'
       version:
         type: string
         description: 'build tag/version'
--- a/buildtools/docker_rig/tarilabs.Dockerfile
+++ b/buildtools/docker_rig/tarilabs.Dockerfile
@@ -121,6 +121,7 @@ RUN apt-get update && \
       libsqlite3-0 \
       openssl \
       procps \
+      tini \
       lsof && \
     # Docker image reduction
     apt-get clean all && \


.github/workflows/build_dockers_workflow.yml
```diff
@@ -20,7 +20,7 @@ name: Build docker images - workflow_call/on-demand
         default: x86-64
       features:
         type: string
-        default: 'safe,grpc'
+        default: 'safe,libtor,metrics'
```
leet4tari, line 23: Should we not keep grpc?
```libtor``` is the default for nix, but having it in seems more clear.

SWvheerden: I think we do need grpc for the block explorers

ns212: I removed it because I don't see grpc defined as a feature flag anymore

SWvheerden: what are we building here?
I see the base node only has has grpc as a config option not compile feature

leet4tari: Docker images for tari_suite.

Console wallet has the empty feature too. Was it not meant as a none default build feature?

leet4tari: I did local build with and without grpc feature, and my base node still started up, so I think this can go in.

Issue #6907, opened 2025-04-01
Use base node starting mode to do peer sync from seed nodes only


SWvheerden:
We should only do peer sync from the seed nodes, after which we should choose random other peers to sync from and not the seed peers.

SWvheerden:
Not going to implement this now

PR #6951, opened 2025-04-15
feat:  peer health check

### LINKED ISSUE #6944: base node tor health check grpc
Build liveness check for network inside base node and expose via grpc for health check

SWvheerden:
Description
---
This adds in a health check whereby the node checks if it can discover a peer and dial a peer. 
It discover, then dials the seed peers 

fixes: #6944 

## Summary by CodeRabbit

- **New Features**
  - Introduced liveness health checks that monitor and report peer discovery and dial latencies for seed peers.
  - Added detailed liveness metrics to the network state gRPC response, providing enhanced visibility into peer connectivity.

- **Configuration**
  - Added a new configuration option for setting the liveness health check interval, with a default of 10 minutes.

- **Bug Fixes**
  - Improved logging by lowering the log level for certain connection errors to reduce unnecessary warnings.

- **Tests**
  - Added unit tests for seed peer retrieval logic.
  - Marked a specific block creation test to be ignored by default during test runs.

- **Chores**
  - Updated documentation and configuration presets to reflect new liveness health check settings.

### DIFF
--- a/applications/minotari_node/src/config.rs
+++ b/applications/minotari_node/src/config.rs
@@ -144,7 +144,9 @@ pub struct BaseNodeConfig {
     pub report_grpc_error: bool,
     // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
-    pub tari_pulse_interval: Duration,
+    pub tari_pulse_interval: Duration,// Interval to check if the base node is still in sync with the network
+    #[serde(with = "serializers::seconds")]
+    pub tari_pulse_health_check: Duration,
 }
 
 impl Default for BaseNodeConfig {
@@ -184,6 +186,7 @@ impl Default for BaseNodeConfig {
             state_machine: Default::default(),
             report_grpc_error: false,
             tari_pulse_interval: Duration::from_secs(120),
+            tari_pulse_health_check: Duration::from_secs(60 * 10),
         }
     }
 }
--- a/applications/minotari_node/src/config.rs
+++ b/applications/minotari_node/src/config.rs
@@ -144,7 +144,7 @@ pub struct BaseNodeConfig {
     pub report_grpc_error: bool,
     // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
-    pub tari_pulse_interval: Duration,// Interval to check if the base node is still in sync with the network
+    pub tari_pulse_interval: Duration, // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
     pub tari_pulse_health_check: Duration,
 }
--- a/applications/minotari_node/src/config.rs
+++ b/applications/minotari_node/src/config.rs
@@ -144,7 +144,8 @@ pub struct BaseNodeConfig {
     pub report_grpc_error: bool,
     // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
-    pub tari_pulse_interval: Duration, // Interval to check if the base node is still in sync with the network
+    pub tari_pulse_interval: Duration,
+    // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
     pub tari_pulse_health_check: Duration,
 }
--- a/base_layer/core/src/base_node/tari_pulse_service/mod.rs
+++ b/base_layer/core/src/base_node/tari_pulse_service/mod.rs
@@ -58,7 +58,7 @@ pub struct TariPulseConfig {
 pub struct LivenessCheckResult {
     pub peer: NodeId,
     pub discovery_latency: Option<Duration>,
-    pub ping_latency: Option<Duration>,
+    pub dial_latency: Option<Duration>,
 }
 
 impl Default for TariPulseConfig {
@@ -271,11 +271,13 @@ pub struct TariPulseServiceInitializer {
 
 impl TariPulseServiceInitializer {
     pub fn new(dns_interval: Duration, liveness_interval: Duration, network: Network) -> Self {
-        Self { dns_interval, liveness_interval, network }
+        Self {
+            dns_interval,
+            liveness_interval,
+            network,
+        }
     }
 }
-
-
 
 #[async_trait]
 impl ServiceInitializer for TariPulseServiceInitializer {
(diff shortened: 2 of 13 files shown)


base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -244,3 +315,42 @@ impl ServiceInitializer for TariPulseServiceInitializer {
...
+                result.discovery_latency = Some(start.elapsed());
+            }
+            let start2 = Instant::now();
+
+            if let Ok(_) = comms.dial_peer(result.peer.clone()).await {
+                result.dial_latency = Some(start2.elapsed());
+            }
+            (*result_clone).write().await.push(result);
+        }));
+    }
```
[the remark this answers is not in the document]
SWvheerden: There are build in timeouts into the calls.

applications/minotari_node/src/config.rs
```diff
@@ -144,7 +144,9 @@ pub struct BaseNodeConfig {
     pub report_grpc_error: bool,
     // Interval to check if the base node is still in sync with the network
     #[serde(with = "serializers::seconds")]
-    pub tari_pulse_interval: Duration,
+    pub tari_pulse_interval: Duration, // Interval to check if the base node is still in sync with the network
```
hansieodendaal:
```suggestion
    pub tari_pulse_interval: Duration, 
    // Interval to check if the base node can still interact with the network
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -89,20 +117,46 @@ impl TariPulseService {
...
+                            Ok(val) => val,
+                            _ => {
+                                debug!(
+                                    target: LOG_TARGET,
+                                    "Could not acquire lock for health check, skipping this tick"
+                                );
+                                return;
+                            },
+                        };
+                    check_health(comms, discovery, notify_channel).await;
```
hansieodendaal:
```suggestion
                        check_health(comms, discovery, notify_channel).await;
```

hansieodendaal (commented):
Looking good - some observations:
- I think it will be prudent if we can supply custom base nodes to include in the liveness check alongside the seed peers. _Use case: A user runs two base nodes, on the same or different computers. Each base node is configured to report on the other base node. When liveness for each other fails, it is time to restart tor with an empty cache._
- We should swop out `dial-peer` for `ping-peer` in the liveness test. Successful discovery indicates a connected network, whereas successful ping indicates a healthy connected peer. Both metrics will be useful. Dial always succeeds after discovery, but ping does not always succeed after dial, which is inherent in ping.
- We should think about some ways in which to restart `tor`, which will be confirmed by the discovery metrics.

SWvheerden:
You can already add custom "seed" nodes via the config to use

Swapping over to ping-peer, good suggestion, 

But we cannot restart tor from the base node as a lot of cases we run tor completely outside of the base node, in the case of the seed nodes, docker runs on a separate instance, and there is no way the base node application can delete files on another docker application.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -372,6 +373,11 @@ async fn check_health(
                     }
                 }
             }
+            if let Ok(Some(mut conn)) = comms.get_connection(result.peer.clone()).await {
+                if let Err(err) = conn.disconnect(Minimized::No).await {
+                    warn!(target: LOG_TARGET, "Failed to disconnect peer {} ({})", result.peer, err);
+                }
+            }
```
hansieodendaal, line 380: This connection should not be dropped here, as there are other processes running to get rid of unused outbound connections. 

An incoming connection on a seed peer from this enquiring node is zero cost to the seed node after this conversation is finished, but the connection may have been established previously as a legitimate outbound connection. 

Another consideration is that a consecutive discovery request will still be a proper discovery, although it should be faster, thus coming back to this code while the outbound connection is still open does not invalidate the tests.

hansieodendaal (commented):
The previous comment _We should think about some ways in which to restart tor with empty caches, which will be confirmed by the discovery metrics._ still holds, even if it is not done from within the base node. This problem should not only be solved for Tari Universe or the seed nodes.

One additional comment below.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
SWvheerden: Disagree. These connections are primary to the seed peers.  The reason our nodes use upwards of 1000 connections is because we keep them open because we might need them later. 
And it's super important to not keep connections open to the seed peers. You are not suppose to use them as a connection hub. If another process wants to use the connection, they can reopen the connection.

hansieodendaal: If we insist on closing the connection, it should only be done if a new connection was established for this purpose, otherwise, a legitimate connection established by another process will be closed.

SWvheerden:
On the 
`The previous comment We should think about some ways in which to restart tor with empty caches, which will be confirmed by the discovery metrics. still holds, even if it is not done from within the base node. This problem should not only be solved for Tari Universe or the seed nodes.`
I agree, but is beyond the scope of this PR. This PR only proves a health check system to use. You need an external application to check this, if this a CRON job, another application. Its beyond the scope here. 

If someone is watching the application it self, they will detect its broken and needs to restart. But the node it self should not do this.

base_layer/core/src/base_node/tari_pulse_service/mod.rs
SWvheerden: this has the added benefit of doing this as well: #6907 
I don't think this is a big issue to disconnect from the seed peers. They will reopen connections if needed. They are there to seed peers, not seed the network.

hansieodendaal: What if the connection we are closing here is from a custom seed peer we are syncing from? Those should not be disconnected. The code here does not differentiate between DNS seeds and custom seeds. DNS seed connections may be closed, but not custom seeds that are not DNS seeds (one can still add a DNS seed as a custom seed).

hansieodendaal:
> On the `The previous comment We should think about some ways in which to restart tor with empty caches, which will be confirmed by the discovery metrics. still holds, even if it is not done from within the base node. This problem should not only be solved for Tari Universe or the seed nodes.` I agree, but is beyond the scope of this PR. This PR only proves a health check system to use. You need an external application to check this, if this a CRON job, another application. Its beyond the scope here.
> 
> If someone is watching the application it self, they will detect its broken and needs to restart. But the node it self should not do this.

We should create an issue so we can provide tooling to do this for MacOS, Linux and Windows. We cannot only cater to Tari Universe users. This is a known problem.

SWvheerden:
https://github.com/tari-project/tari/issues/6962

PR #6954, opened 2025-04-16
fix: force coinbase fee = 0

SWvheerden:
Description
---
Although not exploitable per say, its better to have a fixed consensus checking that the fee = 0 for the coinbase. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to ensure coinbase kernels have a zero fee, improving transaction correctness and error handling.


hansieodendaal (dismissed):
utACK

PR #6963, opened 2025-04-17
feat: add sqlite peer_db

hansieodendaal:
Description
---
- Added an sqlite implementation of peer_db.
- Removed the lmdb implementation of peer_db.

Motivation and Context
---
The lmdb peer_db was memory and resource intensive.

How Has This Been Tested?
---
New unit tests
Existing unit tests depending on the new peer_db implementation pass
System-level testing passed

---
Code review
System-level testing 

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGES: 
- The existing lmbd peer_db will be discarded (left intact but not used anymore).
- In PeerManager, the method `fn closest_peers` is replaced with `fn closest_n_active_peers` as amore flexible alternative.

## Summary by CodeRabbit

- **New Features**
  - Transitioned peer storage backend from LMDB to SQLite, enhancing reliability and simplifying setup.
  - Added test utilities for creating peers with address stats and recent last-seen timestamps.
  - Extended peer and address data structures with constructors, conversion methods, and reset functions.
- **Bug Fixes**
  - Unified peer add/update operations into `add_or_update_peer` for consistency across modules.
  - Corrected configuration comments to specify SQLite database paths instead of LMDB.
- **Refactor**
  - Removed legacy peer query framework and migration code; replaced with async wrappers over SQLite.
  - Enhanced peer selection, filtering, and banning logic to leverage new database queries.
  - Simplified batch peer retrieval in wallet and FFI layers with direct SQL queries.
  - Updated test setups to use SQLite, replacing in-memory or LMDB dependencies.
  - Replaced direct sync calls with async wrappers in discovery, network, and connection modules.
  - Improved asynchronous handling and batch processing for peer management.
- **Chores**
  - Removed LMDB and related dependencies; added SQLite, Diesel, and migration crates.
  - Created SQL migration scripts for node identity, peers, and addresses schemas.
  - Managed temporary database cleanup in tests and connection pools.
  - Cleaned imports and streamlined peer manager initialization in examples and tests.
- **Documentation**
  - Updated configuration comments to clarify transition from LMDB to SQLite storage paths.


comms/dht/src/proto/mod.rs
```diff
@@ -138,7 +138,8 @@ impl TryFrom<rpc::PeerIdentityClaim> for PeerIdentityClaim {
             .filter_map(|addr| Multiaddr::try_from(addr).ok())
             .collect::<Vec<_>>();
 
-        let features = PeerFeatures::from_bits(value.peer_features).ok_or_else(|| anyhow!("Invalid peer features"))?;
+        let features = PeerFeatures::from_bits_u32_truncate(value.peer_features)
+            .ok_or_else(|| anyhow!("Invalid peer features"))?;
```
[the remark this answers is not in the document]
hansieodendaal:
```
PeerFeatures::from_bits(u8::try_from(value & 0b1111_1111).expect("will not fail"))
```
This method (`PeerFeatures::from_bit`) will return None if any unknown bits are set.

comms/core/src/peer_manager/storage/migrations/2025-04-14-072200_initial/up.sql
```diff
@@ -0,0 +1,43 @@
...
+     avg_latency BIGINT,
+     latency_sample_count INTEGER,
+     last_attempted TIMESTAMP,
+     last_failed_reason TEXT,
+     quality_score INTEGER,
+     source TEXT NOT NULL,
+
+     FOREIGN KEY (peer_id) REFERENCES peers (peer_id) ON DELETE CASCADE,
+     CONSTRAINT unique_address UNIQUE (address)
+);
```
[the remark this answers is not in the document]
hansieodendaal: Agree with missing index, but we do not allow the same address to be associated more than one peer or the same peer more than once.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2285 @@
...
+                                .clone()
+                                .map_or("NULL".to_string(), |meta| format!("x'{}'", hex::to_hex(&meta))),
+                            p.peer.deleted_at.map_or("NULL".to_string(), |dt| format!("'{}'", dt)),
+                        )
+                    })
+                    .collect::<Vec<String>>()
+                    .join(", "),
+            );
+
+            peer_query.push_str(" ON CONFLICT (node_id) DO NOTHING");
```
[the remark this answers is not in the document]
hansieodendaal: SQLite does not support the DEFAULT keyword in INSERT statements, which Diesel uses for batch inserts. Diesel's batch insert API is designed for databases like PostgreSQL that support this feature.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2267 @@
...
+                        ));
+                    }
+                    ids.push(format!("x'{}'", address_update.address));
+                }
+            }
+
+            if !set_clauses.is_empty() {
+                address_query.push_str(&set_clauses.join(", "));
+                address_query.push_str(&format!(" WHERE id IN ({})", ids.join(", ")));
+                conn.batch_execute(&address_query)?;
```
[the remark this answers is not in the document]
hansieodendaal: SQLite does not support the DEFAULT keyword in INSERT statements, which Diesel uses for batch inserts. Diesel's batch insert API is designed for databases like PostgreSQL that support this feature.

comms/core/src/peer_manager/peer_features.rs
```diff
@@ -64,6 +64,11 @@ impl PeerFeatures {
             _ => "unknown",
         }
     }
+
+    /// Returns the bits of the PeerFeatures as an i32
+    pub fn to_i32(&self) -> i32 {
+        i32::try_from(self.bits()).unwrap_or_default()
```
sdbondi: Is this used? Seems kind of dangerous as it will lose data silently. Since both i32 and u32 are 32 bits it is more correct to cast `self.bits() as i32`, which could be negative but is lossless i.e. you can convert it back using `my_i32 as u32`. This is the correct way to convert this when using diesel, or if this is not used, remove it
```suggestion
        i32::try_from(self.bits()).unwrap_or_default()
```

comms/core/src/peer_manager/manager.rs
```diff
@@ -349,30 +421,39 @@ impl PeerManager {
...
-        self.peer_storage.read().await.is_peer_banned(node_id)
+        let peer_manager = self.clone();
+        let node_id = node_id.clone();
+        tokio::task::spawn_blocking(move || peer_manager.peer_storage_sql.is_peer_banned(&node_id)).await?
     }
 
-    pub async fn update_each<F>(&self, mut f: F) -> Result<usize, PeerManagerError>
+    // TODO: This function is still hugely inefficient as it retrieve all peers and then perform an update operation -
+    // TODO: it should be split into smaller targeted queries - investigate caller function use case for this.
+    pub async fn update_each<F>(&self, f: &mut F, features: Option<PeerFeatures>) -> Result<usize, PeerManagerError>
```
stringhandler: I am not sure this method is called. I think I was able to remove it. It's probably better to find where this is used, if at all and then change it to a single update statement

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2304 @@
...
+                    let added_at = p.peer.added_at;
+                    let user_agent = sql_escape(&p.peer.user_agent);
+                    let metadata = p
+                        .peer
+                        .metadata
+                        .clone()
+                        .map_or("NULL".to_string(), |meta| format!("x'{}'", hex::to_hex(&meta)));
+                    let deleted_at = p.peer.deleted_at.map_or("NULL".to_string(), |dt| format!("'{}'", dt));
+
+                    format!(
```
stringhandler: Constructing sql strings is bad practice, especially since the data is retrieved from the network. 

It is better to create a sql statement with parameters and bind them to the parameters, or even better, use diesel's generated structs. 

In this case, rather:
1. Execute Begin Transaction
2. In a for loop, in each iteration, insert a single peer.
3. Commit the transaction

This will eliminate the need for `sql_escape` and also prevent problems where the query becomes too long that the api rejects it

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2304 @@
...
+    ) -> Result<(), StorageError> {
+        let mut conn = self.connection.get_pooled_connection()?;
+        conn.transaction::<_, StorageError, _>(|conn| {
+            // Batch update peers
+            if !peers_with_addresses.is_empty() {
+                let mut peer_query = String::from("UPDATE peers SET ");
+                let mut set_clauses = vec![];
+                let mut node_ids = vec![];
+
+                for update in &peers_with_addresses {
```
stringhandler: This can be done with diesel. Rather use that.

It looks like you have the diesel struct already implemented below, so why not just use that?

comms/core/src/peer_manager/storage/database.rs
stringhandler: This code is already wrapped in a transaction, so just insert one line at a time

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2304 @@
...
+        if n == 0 {
+            return Ok(Vec::new());
+        }
+
+        let mut conn = self.connection.get_pooled_connection()?;
+        let exclude_node_ids = exclude_node_ids.iter().map(|id| id.to_hex()).collect::<Vec<_>>();
+
+        conn.transaction::<_, StorageError, _>(|conn| {
+            // Step 1: Filtered list of node_ids
+            let mut node_ids: Vec<String> = peers::table
```
stringhandler: This reads out the whole db. How about this instead:
Add a column `random_i` to the database and fill it with either a random byte or a deterministic byte from a hash of the peer_id. Then change your query to sort by the random byte and take the top `n`. You can also choose a random offset to skip some nodes

comms/core/src/peer_manager/storage/migrations/2025-04-14-072200_initial/up.sql
```diff
@@ -0,0 +1,45 @@
...
+    node_id TEXT NOT NULL,
+    flags INTEGER NOT NULL,
+    banned_until TIMESTAMP,
+    banned_reason TEXT,
+    features INTEGER NOT NULL,
+    supported_protocols TEXT NOT NULL,
+    added_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    user_agent TEXT NOT NULL,
+    metadata BLOB,
+    deleted_at TEXT,
```
stringhandler:
```suggestion
    deleted_at TIMESTAMP,
```

stringhandler (commented):
This pr could introduce some vulnerabilities if the manual sql crafting is not removed. 

I don't think we need to physically delete any rows from the database, since it is pretty efficient to filter by deleted_at. This was not the case with lmdb though.

comms/core/src/peer_manager/manager.rs
hansieodendaal: New db query `reset_offline_non_wallet_peers` added, and all this code removed.

applications/minotari_console_wallet/src/ui/state/app_state.rs
```diff
@@ -944,8 +944,17 @@ impl AppStateInner {
         let peer_manager = self.wallet.comms.peer_manager();
         let mut peers = Vec::with_capacity(connections.len());
         for c in &connections {
-            if let Ok(Some(p)) = peer_manager.find_by_node_id(c.peer_node_id()).await {
-                peers.push(p);
+            match peer_manager.find_by_node_id(c.peer_node_id()).await {
```
SWvheerden: this can probably be done better with `.find_by_node_ids(Vec<Peer_id>)`

comms/core/src/peer_manager/manager.rs
```diff
@@ -20,93 +20,68 @@
...
-                    });
-                }
-            }
-        }
-        let mut lock = self.peer_storage.write().await;
-        let peer_id = lock.add_peer(peer)?;
+    pub async fn add_or_update_peer(&self, peer: Peer) -> Result<PeerId, PeerManagerError> {
+        let peer_manager = self.clone();
+        let peer_id =
+            tokio::task::spawn_blocking(move || peer_manager.peer_storage_sql.add_or_update_peer(peer)).await??;
```
SWvheerden: why spawn a block task here if we are going to await the results any way?

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2667 @@
...
+        self.this_peer_identity.clone()
+    }
+
+    fn add_this_peer_node_identity_to_db(&self) -> Result<(), StorageError> {
+        let mut conn = self.connection.get_pooled_connection()?;
+
+        let node_identity_indexes = node_identity::table.load::<NewThisPeerIdentitySql>(&mut conn)?;
+        if node_identity_indexes.len() > 1 {
+            return Err(StorageError::UnexpectedResult(format!(
+                "There are multiple node identities fort this peer in the database, expected 1, found {}",
```
SWvheerden:
```suggestion
                "There are multiple node identities for this peer in the database, expected 1, found {}",
```

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2667 @@
...
+                node_identity_indexes.len()
+            )));
+        }
+        if !node_identity_indexes.is_empty() {
+            if self.this_peer_identity.public_key.to_hex() == node_identity_indexes[0].public_key &&
```
SWvheerden: why check this?
These are unique fields on the db?

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -0,0 +1,2667 @@
...
+                        supported_protocols,
+                        added_at,
+                        user_agent,
+                        metadata,
+                        deleted_at
+                    ))
+                })
+                .collect::<Result<Vec<String>, _>>()?;
+
+            let mut peer_query = format!(
```
SWvheerden: why not just go :
```
for peer in peers:
add_peer_to_db
```
quick enough, we are not adding thousands at a time?

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: We have individual queries to insert or update - this was an attempt to do it in batch mode. Diesel with the SQLite backend does not support this.

For now these two queries have been commented out with a note to revisit them again in future, as they are currently still broken.

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: We have individual queries to insert or update - this was an attempt to do it in batch mode. Diesel with the SQLite backend does not support this.

For now these two queries have been commented out with a note to revisit them again in future, as they are currently still broken.

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: Good point, thanks. Investigating this I could do this directly with the query instead:
```rust
                .order_by(diesel::dsl::sql::<diesel::sql_types::Integer>("RANDOM()"))
                .limit(n as i64)
```

comms/core/src/peer_manager/manager.rs
hansieodendaal: As per the DM discussion, we are currently not in agreement on what the best course of action is.

To answer the question for the record:
_Spawn blocking is used here to let Tokio know to start this task on a thread where blocking is acceptable, as is usually used for disk i/o operations. This have advantages for the ret of the Tokio runtime_

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: The peer_db could be copied to a different node with a different node id, or the node id for the node could be changed, then all the distance calculations would be incorrect.

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: Batching inside sql is orders of magnitude quicker when we add hundreds or thousands of peers, which is a valid use case when a node receives its initial peer lists.

Note: This code is currently commented out and is pending further investigation and testing.

comms/core/src/peer_manager/peer_storage_sql.rs
```diff
@@ -0,0 +1,758 @@
...
+            return Ok(NodeDistance::max_distance());
+        }
+
+        // If we have less than `n` matching peers in our threshold group, the threshold should be max
+        if dists.len() < n {
+            return Ok(NodeDistance::max_distance());
+        }
+
+        Ok(dists.pop().expect("dists cannot be empty at this point"))
+    }
```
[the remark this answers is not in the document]
hansieodendaal: The function in database.rs does this before the results are returned:
```rust
filtered_node_ids.sort_by_key(|a| a.distance(&region_node_id));
```

comms/core/src/peer_manager/manager.rs
```diff
@@ -416,73 +447,90 @@ impl fmt::Debug for PeerManager {
...
+            rand::thread_rng().gen_range(1..255),
+            rand::thread_rng().gen_range(1..255),
+            rand::thread_rng().gen_range(1..255),
+            rand::thread_rng().gen_range(5000..9000),
+        ];
+        let net_address = format!("/ip4/{}.{}.{}.{}/tcp/{}", n[0], n[1], n[2], n[3], n[4])
+            .parse::<Multiaddr>()
+            .unwrap();
+        net_addresses.add_address(&net_address, &PeerAddressSource::Config);
+    }
```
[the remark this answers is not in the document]
hansieodendaal: No compiler error here

comms/core/src/peer_manager/manager.rs
hansieodendaal: I concede, my system-level testing showed this is not a good design - the spawn blocking calls do not work well with SQLite in WAL mode.

applications/minotari_console_wallet/src/ui/state/app_state.rs
hansieodendaal: Also applied to all functions that loop over individual peer_db entries

common_sqlite/src/error.rs
```diff
@@ -63,10 +63,30 @@ pub enum StorageError {
...
     fn from(value: MessageFormatError) -> Self {
         StorageError::MessageFormatError(value.to_string())
     }
 }
+
+impl From<HexError> for StorageError {
+    fn from(value: HexError) -> Self {
+        StorageError::HexError(value.to_string())
+    }
+}
```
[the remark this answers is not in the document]
hansieodendaal: 67 |     HexError(#[from] HexError),
   |                ^^^^ method cannot be called on `&HexError` due to unsatisfied trait bounds

comms/core/src/peer_manager/manager.rs
```diff
@@ -20,93 +20,60 @@
...
+    /// Get this peer's identity
+    pub fn this_peer_identity(&self) -> ThisPeerIdentity {
+        self.peer_storage_sql.this_peer_identity()
     }
 
+    /// Get the number of peers in the PeerManager - any error will translate to a size of zero
     pub async fn count(&self) -> usize {
-        self.peer_storage.read().await.count()
+        self.peer_storage_sql.count()
     }
```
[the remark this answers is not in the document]
hansieodendaal: `PeerDatabaseSql` in the backend has been refactored to mitigate this, and the system level has been tested under high load together with unit test `test_concurrent_add_or_update_and_get_closest_peers`.

SWvheerden (approved):

PR #6966, opened 2025-04-17
perf: cache block template

stringhandler:
Cache the block template (before coinbases) if the height and mempool has not changed

## Summary by CodeRabbit

- **New Features**
  - Introduced a caching mechanism for block template generation, improving performance by avoiding redundant computations when the mempool and best block header remain unchanged.

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -35,24 +35,15 @@ use tokio::sync::RwLock;
 use crate::base_node::metrics;
 use crate::{
     base_node::comms_interface::{
-        error::CommsInterfaceError,
-        local_interface::BlockEventSender,
-        FetchMempoolTransactionsResponse,
-        NodeCommsRequest,
-        NodeCommsResponse,
-        OutboundNodeCommsInterface,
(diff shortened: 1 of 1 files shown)


SWvheerden (approved):

base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -267,6 +269,23 @@ where B: BlockchainBackend + 'static
...
                 let best_block_header = self.blockchain_db.fetch_tip_header().await?;
                 let mut last_seen_hash = self.mempool.get_last_seen_hash().await?;
+                {
+                    let read_lock = self.cached_block_template.read().await;
+                    if let Some(cache) = read_lock.as_ref() {
+                        if cache.0 == last_seen_hash && &cache.1.header.prev_hash == best_block_header.hash() {
+                            return Ok(NodeCommsResponse::NewBlockTemplate(cache.1.clone()));
+                        }
+                    }
+                }
```
hansieodendaal, line 279: I would remove these lines - we are not supposed to access this function asynchronously. 
- If one thead has obtained the write lock:
   - the next thread will wait at the read lock.
   - the mempool has changed and the next thread should wait at the read lock.
The lines below do exactly the same thing.

hansieodendaal (commented):
@stringhandler see comment for improvement - the existing code will work fine.

PR #6973, opened 2025-04-21
fix: don't recalc smt on height change

stringhandler:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling during block validation and MMR root calculations, providing clearer warnings and distinguishing expected errors from unexpected ones.
  - Enhanced logging for specific error cases to help users better understand validation failures.

### DIFF
--- a/base_layer/core/src/validation/block_body/block_body_full_validator.rs
+++ b/base_layer/core/src/validation/block_body/block_body_full_validator.rs
@@ -94,7 +94,25 @@ impl BlockBodyFullValidator {
             );
             ChainStorageError::AccessError("write lock on smt".into())
         })?;
-        let mmr_roots = chain_storage::calculate_mmr_roots(backend, &self.consensus_manager, &block, &mut output_smt)?;
+        let mmr_roots =
+            match chain_storage::calculate_mmr_roots(backend, &self.consensus_manager, &block, &mut output_smt) {
+                Ok(mmr_roots) => mmr_roots,
+                Err(e) => {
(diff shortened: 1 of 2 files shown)


base_layer/core/src/validation/block_body/block_body_full_validator.rs
```diff
@@ -94,7 +94,25 @@ impl BlockBodyFullValidator {
...
+                        "Validator could not calculate MMR roots for block {}: {:?}", block.hash().to_hex(), e
+                    );
+                    match e {
+                        ChainStorageError::CannotCalculateNonTipMmr(ref _e) => {
+                            return Err(e.into());
+                        },
+                        _ => {},
+                    }
+                    // Recalculate SMT as it might have been altered.
+                    *output_smt = backend.calculate_tip_smt()?;
```
hansieodendaal, line 109: To re-calculate the smt inside a validator seems out of place - validate should only validate and not fix things

stringhandler: This is due to the bad design of the SMT. It can get corrupted in calculate_mmr_roots and this is to roll it back on error. I have not added logic here, just matched the logic that is **present in other places** but missing here. 

Background: calculate_mmr_roots changes the SMT, then calculates the root, then undoes changes to the SMT. If there is an error during `calculate_mmr_roots` and it exits early we need to revert the SMT.

SWvheerden (approved):

PR #6974, opened 2025-04-22
fix!: payment id

SWvheerden:
Description
---
Enhances the payment_id security to make it more robust against malicious deserialization. 
Removes u64 size bucket to make outputs more the same
Make change and sender data type's serialized size the same. 

## Summary by CodeRabbit

- **Refactor**
  - Simplified and standardized payment ID metadata by removing legacy fields and consolidating amount, fee, and sender flags.
  - Updated serialization format for payment IDs with explicit variant tagging for improved consistency.
  - Modified transaction and output creation to include fee and amount details in payment IDs.
  - Removed DeepSource static analysis tool configuration.
  - Adjusted key derivation and payment ID variant handling to support wider identifiers.
  - Streamlined transaction weight tracking by removing weight and input/output counts from payment ID metadata.
  - Refactored UI to align with updated payment ID metadata structure.

- **Bug Fixes**
  - Enhanced parsing of payment ID variants to prevent misinterpretation.

- **Tests**
  - Updated tests to reflect the revised payment ID structure and serialization format.


applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
```diff
@@ -404,9 +397,9 @@ impl TransactionsTab {
                         tx.direction.clone(),
                         tx.amount,
                         fee,
-                        weight,
-                        inputs_count,
-                        outputs_count,
+                        tx.weight,
+                        tx.inputs_count,
+                        tx.outputs_count,
```
hansieodendaal, line 402: We should make these 3 values options, then this branch can return None for each, as the information from `tx.???` will be wrong. The display can blank out those values as `unknown`.

base_layer/core/src/blocks/pre_mine/mod.rs
```diff
@@ -853,7 +850,7 @@ pub async fn create_pre_mine_genesis_block_info(
                 RangeProofType::RevealedValue,
             ))
             .with_script(script)
-            .encrypt_data_for_recovery(&key_manager, Some(&view_key_id), PaymentId::U64(i.try_into().unwrap()))
+            .encrypt_data_for_recovery(&key_manager, Some(&view_key_id), PaymentId::U256(i.into()))
```
hansieodendaal, line 853: Future pre-mine operations must be able to read the original encrypted `PaymentId::U64`

base_layer/core/src/transactions/transaction_components/encrypted_data.rs
```diff
@@ -159,8 +158,6 @@ pub enum PaymentId {
     /// No payment ID.
     #[default]
     Empty,
-    /// A u64 number.
-    U64(u64),
```
hansieodendaal, line 163: We must still be able to decrypt a `PaymentId::U64` from the genesis block - this is used in pre-mine spending

hansieodendaal (commented):
It looks good, except I am concerned that we will not be able to get the correct payment id when we read the pre-mine outputs in the genesis block. Those have been created with the previous `PaymentId::U64` and when this code parses it the first byte will be lost to `TxType` and the remaining 7 will be returned as data for `PaymentId::Open`. I think we need some code in pre-mine spend to convert the `PaymentId::Open` to `u64`.

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -2012,7 +2018,7 @@ where
         }
 
         let (output, sender_offset_key_id) = self
-            .output_to_self(output_features, amount, covenant, payment_id)
+            .output_to_self(output_features, amount, covenant, payment_id, fee_per_gram)
             .await?;
```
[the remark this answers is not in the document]
hansieodendaal: Agreed, we need a better estimation here

hansieodendaal (commented):
Apart from the failing unit tests, just one comment below, otherwise looking good.

Have you run any system-level tests to verify transaction metadata displayed on the console wallet?
- One-sided in sender wallet
- One-sided in receiver wallet
- Interactive in sender wallet
- Interactive in receiver wallet
- View-only sender wallet change outputs

applications/minotari_console_wallet/src/ui/components/transactions_tab.rs
SWvheerden: defaults are fine

PR #7018, opened 2025-05-03
fix: stop allowing too many connections

fluffypony:
Description
---

We have an outbound connection limiter. We should probably actually use it.

Motivation and Context
---

This has been plaguing the network by overloading nodes with open inbound ports.

How Has This Been Tested?
---

I mean...it's trivial to test.

---

Compile and run. Observe that outbound connections doesn't exceed 32. Celebrate.

Breaking Changes
---

- [x] None

## Summary by CodeRabbit

- **Refactor**
  - Updated the default configuration to enable connection minimization by default, which may affect how excess network connections are managed.


fluffypony:
I have a better idea, new PR to follow

PR #7035, opened 2025-05-08
chore(wallet): allow read only wallet to receive keys args as envs

MCozhusheck:
Description
---
Allow `view_private_key` and `spend_key` CLI args for minotari_console_wallet to be received as envs.

Motivation and Context
---
Some users complained about these value being visible in logs. This is not security issue but just to make it more obvious since `view_private_key` sounds like something that shouldn't be visible as plain text.

## Summary by CodeRabbit

- **New Features**
  - Added support for setting the view private key and spend key via environment variables, in addition to existing command-line options.


hansieodendaal (commented):
Please add `hide_env_values = true` to both env's as suggested by coderabbitai

SWvheerden (dismissed):

PR #7041, opened 2025-05-09
feat: add view key to ffi

SWvheerden:
Description
---
Adds get view method to the FFI
Adds get spend key method to the FFI
Correctly feature gate's the mainnet seeds to mainnet only

## Summary by CodeRabbit

- **New Features**
  - Added the ability to retrieve the private view key and public spend key from the wallet via new interface options.

- **Chores**
  - Updated wallet creation to use a fixed list of peer seeds only when targeting the main network.


hansieodendaal (approved):
utACK

PR #7074, opened 2025-05-21
fix: base node panic

SWvheerden:
Description
---
Fix potential panic in base node

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the calculation method for estimated hash rates of different proof-of-work algorithms, ensuring more accurate and reliable network state reporting.

- **Chores**
  - Updated mainnet configuration to adjust the expected block target times for multiple proof-of-work algorithms.
  - Refined pre-mine token maturity schedules and spendable supply expectations in tests to reflect updated block boundaries.

### DIFF
--- a/applications/minotari_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/minotari_node/src/grpc/base_node_grpc_server.rs
@@ -426,7 +426,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
                         obscure_error_if_true(report_error_flag, Status::internal(e.to_string()))
                     })?;
                 let target_time = constants.pow_target_block_interval(PowAlgorithm::Sha3x);
-                let estimated_hash_rate = target_difficulty.as_u64().checked_sub(target_time).unwrap_or(0);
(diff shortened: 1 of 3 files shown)


applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -426,7 +426,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
                         obscure_error_if_true(report_error_flag, Status::internal(e.to_string()))
                     })?;
                 let target_time = constants.pow_target_block_interval(PowAlgorithm::Sha3x);
-                let estimated_hash_rate = target_difficulty.as_u64() / target_time;
+                let estimated_hash_rate = target_difficulty.as_u64().saturating_sub(target_time);
```
SChernykh, line 429: I don't understand... Why do you subtract time from difficulty? Shouldn't you do a division here?

SWvheerden: you are right, it should be...

PR #7112, opened 2025-05-27
fix: exclude coinbases from fee calc

SWvheerden:
Description
---
dont include the coinbase in the fee calcs

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected block fee calculations to exclude coinbase transactions, ensuring only non-coinbase fees are included.


hansieodendaal (approved):
utACK

PR #7117, opened 2025-05-28
feat!: remove the ability to send completely raw bytes via grpc

SWvheerden:
Description
---
Changes the transfer command, that if you provide the raw bytes, it puts it into the open format. 
This does change legacy behaviour, but most people who use this are not aware that the bytes if provided in raw from need to contain some other details as well.

By providing raw data to the wallet it thinks its i nthe correct form, but its not and this leads to weird looking payment id's 

## Summary by CodeRabbit

- **New Features**
  - Transaction details now include comprehensive input and output commitment data for enhanced transparency.
  - Transaction responses in transfers and batch processes provide enriched metadata for better insight.
- **Refactor**
  - Improved payment ID construction during transfers for more precise transaction identification.

### DIFF
--- a/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -121,6 +121,8 @@ use tari_core::{
             SideChainFeature,
             UnblindedOutput,
         },
+        transaction_key_manager::TransactionKeyManagerInterface,
+        transaction_protocol::recipient::RecipientState,
     },
 };
 use tari_script::script;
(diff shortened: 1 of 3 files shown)


applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -704,7 +750,7 @@ impl wallet_server::Wallet for WalletGrpcServer {
         let mut transfers = Vec::new();
         for (hex_address, address, amount, fee_per_gram, payment_type, user_payment_id, raw_payment_id) in recipients {
             let payment_id = if !raw_payment_id.is_empty() {
-                PaymentId::from_bytes(&raw_payment_id)
+                PaymentId::open(raw_payment_id.to_vec(), TxType::PaymentToOther)
```
hansieodendaal, line 753: It would be better if this option were completely removed from the interface, as it boils down to being the same as `UserPaymentId::user_bytes`. Having a choice between two things that give the same result is confusing. It is also possible to specify both `raw_payment_id` and `user_payment_id` with undefined behaviour for the user, where `raw_payment_id` will always have preference.

hansieodendaal (commented):
Just one comment on raw payment id.
Supplemented transaction onfo looks good.

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
SWvheerden: This is backwards compatibility, by removing this, its a silent fail. 
If we remove the grpc as well, it becomes a silent fail as well.

PR #7128, opened 2025-05-29
feat: disable default dht discovery forwarding

hansieodendaal:
Description
---
- Disabled default forwarding of DHT discovery messages as this is not required for the proper operation of the network. A config option has been added to enable or disable this, but the default is false.
- Adjusted the DHT dedup cache default config to be a lot more conservative before starting new dedup cycle - the amount of time an identical inbound message will be ignored before being processed again.

Motivation and Context
---
Mainnet is being flooded with messages that do not contribute to the blockchain.

How Has This Been Tested?
---
System-level testing shows markedly reduced overheads, inbound pipeline finish times and overall memory usage. The dedup cache also does its job much better now so the base node does not can ignore the endless round robbin of the messages.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration option to enable or disable message forwarding in the network's DHT. This option is disabled by default and can be enabled for communication nodes.

- **Improvements**
  - Increased the default cache capacity for deduplication and extended the default cache trim interval, enhancing overall message handling efficiency.

- **Documentation**
  - Updated configuration files and documentation comments to reflect the new forwarding option and revised cache settings.


SWvheerden (approved):
I am slightly weary that this might cause some unintended breakage down the line

PR #7131, opened 2025-05-30
fix: mismatched tms db state

SWvheerden:
Description
---
Somehow the TMS db for completed transactions loses its mined height, but keeps its mined_confirmed status. The outputs are correctly marked with mined_heights in the OMS. 
This tries to correct the fields and update them 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection and retrieval of transactions with mismatched mined status, ensuring more accurate reporting of unconfirmed transactions in the wallet.


hansieodendaal (approved):
utACK

PR #7141, opened 2025-05-31
fix!: update target time

SWvheerden:
Description
---
This updates the target time inside the lwma if the consensus constants change
During sync, the consensus constants need to be updated on the block height that is currently correct. 
The difficulties for Sha3x and RxM need to be calculated using a target time of 240 secs up to block 14_999, but on block 15_000 they need to ber calculated using a target time of 360. The previous difficulties are still correctly calculated, as they are still targeting a time of 240. 

## Summary by CodeRabbit

- **New Features**
  - Added the ability to update the target block interval for supported algorithms dynamically.
  - Improved validation and error handling when updating target block times.

- **Bug Fixes**
  - Ensured algorithms that are no longer permitted are correctly removed from the active set.


stringhandler (approved):
Tested  locally and working

PR #7171, opened 2025-06-03
fix: reduce threshold for flood ban

stringhandler:
Reduce the threshold for banning to 1000 messages per minute. Previously it was 100_000 per 100 seconds

## Summary by CodeRabbit

- **Chores**
  - Updated default settings to reduce the threshold and timespan for banning peers due to message flooding in the DHT system.


hansieodendaal (approved):
utACK

PR #7184, opened 2025-06-04
fix: remove code for deleting stale peers

stringhandler:
This code is no longer needed since the switch to sqlite

## Summary by CodeRabbit

- **Refactor**
  - Removed the automatic deletion of stale peers from the connectivity and peer management processes. The system will no longer periodically delete inactive or stale peers from the database.


SWvheerden (dismissed):

hansieodendaal:
I do not think we should remove the code, but change it to a soft delete instead, as is used elsewhere in the code base (`async fn on_peer_connection_failure(&mut self, node_id: &NodeId)`).

PR #7185, opened 2025-06-04
feat: improve processing of stale peers

hansieodendaal:
Description
---
- Updated processing of stale peers to not hard delete any peers, but only to mark them as deleted (soft delete).
- Updated the liveness service to also update peer stats when a ping is received, when this can be done with certainty, i.e., if the peer only has one address, and not only when a pong is received.

Motivation and Context
---
- When peers are removed from the database, any future contact with that node is not defined.
- Receiving a ping should also count towards peer stats, even if this node did not initiate the metadata exchange.

How Has This Been Tested?
---
Updated unit test.
System-level testing.

---
Code review.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Peer metadata such as last seen time and latency are now updated and refreshed when Ping and Pong messages are received.

- **Bug Fixes**
  - Improved handling of stale peers: instead of permanently deleting them, stale peers are now marked as deleted, allowing for better tracking and potential recovery.

- **Refactor**
  - Renamed methods and updated logic to reflect the shift from hard deletion to soft deletion of stale peers, ensuring more accurate peer management.

### DIFF
--- a/base_layer/p2p/src/services/liveness/service.rs
+++ b/base_layer/p2p/src/services/liveness/service.rs
@@ -195,6 +195,15 @@ where
 
                 let ping_event = PingPongEvent::new(node_id, None, ping_pong_msg.metadata.into(), ping_pong_msg.nonce);
                 self.publish_event(LivenessEvent::ReceivedPing(Box::new(ping_event)));
+
+                // When we receive a ping we do not know from which of the peer's addresses it was sent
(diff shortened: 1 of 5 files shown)


SWvheerden:
repalced with: #https://github.com/tari-project/tari/pull/7184

PR #7233, opened 2025-06-19
fix: imported transaction directions

SWvheerden:
Description
---
Fix imported transactions. 

Doing a spend to self will result in two inbound transactions for the same amount, they need to be one out and 1 in

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved accuracy in determining transaction direction and destination address for various transaction types, ensuring correct categorization and display in transaction histories.

### DIFF
--- a/base_layer/wallet/src/transaction_service/service.rs
+++ b/base_layer/wallet/src/transaction_service/service.rs
@@ -3830,29 +3830,23 @@ where
         {
             (
                 match tx_type {
-                    TxType::PaymentToOther | TxType::Burn => TransactionDirection::Outbound,
-                    TxType::PaymentToSelf |
-                    TxType::CoinSplit |
-                    TxType::CoinJoin |
-                    TxType::ValidatorNodeRegistration |
+                    TxType::PaymentToOther |
+                    TxType::Burn |
                     TxType::CodeTemplateRegistration |
+                    TxType::ValidatorNodeRegistration |
+                    TxType::CoinSplit |
+                    TxType::PaymentToSelf => TransactionDirection::Outbound,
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3830,29 +3830,23 @@ where
         {
             (
                 match tx_type {
-                    TxType::PaymentToOther | TxType::Burn => TransactionDirection::Outbound,
-                    TxType::PaymentToSelf |
+                    TxType::PaymentToOther |
+                    TxType::Burn |
+                    TxType::CodeTemplateRegistration |
+                    TxType::ValidatorNodeRegistration |
                     TxType::CoinSplit |
```
hansieodendaal, line 3837: `TxType::PaymentToSelf`, `TxType::CoinSplit` and `TxType::CoinJoin` must all be treated similarly, in this case `TransactionDirection::Outbound`

hansieodendaal: For that matter, all change direction metadata except `TxType::Coinbase` should be treated similarly
With `TxType::Coinbase` there can be no change

hansieodendaal (commented):
See comments below

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: the ones that are left, dont contain any change in their transactions.

hansieodendaal: A comment would clear it up

PR #7252, opened 2025-06-23
fix: remove long timeout in interactive_tx till tx is persisted into db

### LINKED ISSUE #7251: Tx should be stored imeaditly, not after broadcast
Transactions should be stored immediately, and not wait for a broadcast attempt before they are submitted to the db

See PR #7234 

Timeouts should not be required here

SWvheerden:
Description
---
fixes: #7251 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction handling to ensure pending outbound transactions are added at the correct stage and with accurate status and fee information.
  - Adjusted transaction size checks to cancel oversized transactions earlier in the process.
  - Reduced the timeout when waiting for transaction persistence, resulting in faster feedback during transfers.
- **Tests**
  - Added a new test scenario to validate handling of multiple consecutive interactive transactions between wallets.
  - Introduced a test step to support sending multiple interactive transactions with specified fees in sequence.

### DIFF
--- a/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -841,7 +841,6 @@ impl wallet_server::Wallet for WalletGrpcServer {
                         .get_wallet_one_sided_address()
                         .await
                         .map_err(|e| Status::internal(format!("{:?}", e)))?;
-                    // The transaction is only persisted after it has been sent via comms, so we need to wait
                     let wallet_tx = timeout(Duration::from_millis(100), async {
                         loop {
                             let tx = self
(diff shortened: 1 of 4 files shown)


hansieodendaal (dismissed):
ACK

This provides a better transaction flow.

I also tested this by removing the wait loop in the grpc server method altogether, like it was before the original PR #7234 changed it, but then it still breaks in cucumber test `Scenario: Wallet imports pre_mine UTXO` although not on the first but at the 2nd interactive transaction not persisted, which means the changed transaction flow improves matters a bit but does not fix it. We still need the wait loop to ensure the transaction is persisted.

hansieodendaal (approved):
ACK

The additional cucumber test `Scenario: Wallet sends many interactive transactions to another wallet` proves that the grpc timeout is adequate for the wallet to handle many interactive transactions.

PR #7259, opened 2025-06-24
feat: auto zero value coinbase reward calculation

MoneroOcean:
Description
---
This adds way to auto use correct (max) coinbase value in case you pass single coinbase with zero value.

Motivation and Context
---
This is hard to guess correct coinbase value to pass its check in GetNewBlockWithCoinbases and GetNewBlockTemplateWithCoinbases GRPC methods.

How Has This Been Tested?
---
Tested as part of RandomX-T pool implementation. Will add new integration tests to this PR but I was not able to run them locally since I can't find seed node setup environment.

---
Run new integration scenario: Verify gprc can create block with zero value coinbase

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of blocks with a single zero-value coinbase, ensuring it is assigned a minimal or full block reward as appropriate.
- **Tests**
  - Added integration tests to verify processing and submission of blocks containing a single zero-value coinbase output.
  - Introduced a test scenario validating block creation with a zero-value coinbase via the gRPC interface.

### DIFF
--- a/applications/minotari_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/minotari_node/src/grpc/base_node_grpc_server.rs
@@ -1128,7 +1128,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         if coinbases.len() == 1 && coinbases[0].value == 0 {
             coinbases[0].value = 1;
         }
-        
+
         let mut total_shares = 0u128;
         for coinbase in &coinbases {
             total_shares += u128::from(coinbase.value);
@@ -1390,7 +1390,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
         if coinbases.len() == 1 && coinbases[0].value == 0 {
             coinbases[0].value = reward.as_u64();
         }
-       
+
         let mut amount = 0u64;
         for coinbase in &coinbases {
             amount += coinbase.value;
--- a/applications/minotari_node/src/grpc/base_node_grpc_server.rs
+++ b/applications/minotari_node/src/grpc/base_node_grpc_server.rs
@@ -1126,7 +1126,7 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
 
         // assume full coinbase reward in case of single coinbase with zero value
         if coinbases.len() == 1 && coinbases[0].value == 0 {
-            coinbases[0].value = reward as u64;
+            coinbases[0].value = 1;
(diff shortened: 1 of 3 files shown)


applications/minotari_node/src/grpc/base_node_grpc_server.rs
```diff
@@ -1123,6 +1123,12 @@ impl tari_rpc::base_node_server::BaseNode for BaseNodeGrpcServer {
                 })?
                 .as_u64(),
         );
+
+        // assume full coinbase reward in case of single coinbase with zero value
+        if coinbases.len() == 1 && coinbases[0].value == 0 {
+            coinbases[0].value = reward as u64;
```
stringhandler:
```suggestion
            coinbases[0].value = 1;
```

This field is badly named. It should be `number_of_shares`. Simply setting it to 1 will get the full reward.

stringhandler (commented):
Thanks for this!

applications/minotari_node/src/grpc/base_node_grpc_server.rs
MoneroOcean: Oh, I see. So technically this PR is only needed for GetNewBlockWithCoinbases. GetNewBlockTemplateWithCoinbases should work with value 1 like you suggested. I am currently using GetNewBlockTemplate/GetNewBlockWithCoinbases way (where it is almost impossible to find out correct coinbase value) but plan to transition to GetNewBlockTemplateWithCoinbases implementation, so probably this PR is not needed at all. Up to you.

hansieodendaal (commented):
Please run
```
cargo +nightly-2025-01-01 fmt --all
```
otherwise looking good!

SWvheerden (dismissed):

metalaureate:
👋 @MoneroOcean — your PR was merged and this bounty is queued for an XTM payout. To receive it, **reply to this PR with your Tari (XTM) payout address** (base58 or emoji ID) on its own line, for example:

`XTM: <your address>`

For your protection we only accept an address posted by **@MoneroOcean** (the merged-PR author). Comments from anyone else are ignored.

MoneroOcean:
XTM: 12FrDe5cUauXdMeCiG1DU3XQZdShjFd9A4p9agxsddVyAwpmz73x4b2Qdy5cPYaGmKNZ6g1fbCASJpPxnjubqjvHDa5

PR #7215, opened 2025-06-13
feat!: full http wallet refactor

stringhandler:
## Summary by CodeRabbit

* **New Features**
  * Introduced new HTTP API endpoints for detailed UTXO and transaction queries.
  * Added JSON-RPC support for transaction submission.
  * Enabled wallet configuration of HTTP client URL for base node communication.
  * Introduced HTTP client factory trait and default implementation.
  * Added HTTP base node mock client for testing.
  * Added cargo aliases to streamline CI checks.

* **Improvements**
  * Enhanced error handling with structured JSON responses in HTTP APIs.
  * Simplified wallet connectivity and base node monitoring using HTTP client abstraction.
  * Improved logging clarity and formatting across services.
  * Updated output manager and transaction service to use generic HTTP client interfaces.
  * Removed legacy base node peer management from wallet UI and state.
  * Improved key recovery method signatures for clarity.
  * Streamlined wallet startup and base node peer configuration flows.
  * Enhanced latency tracking and error handling in HTTP base node wallet client.
  * Refactored transaction broadcast and validation protocols to use generic client interfaces.
  * Simplified event monitoring by removing connectivity and contacts liveness events.
  * Improved error logging by logging error objects directly instead of string conversions.

* **Bug Fixes**
  * Fixed issues related to output key recovery method parameters and output validation logic.

* **Removals**
  * Removed legacy peer management, base node configuration, and connectivity service code.
  * Disabled deprecated CLI commands and gRPC endpoints incompatible with new architecture.
  * Removed obsolete tests and legacy mocks related to RPC and peer management.
  * Removed event handling for base node connectivity and contacts liveness.
  * Removed network tab UI component and related peer selection features.
  * Removed FFI functions related to base node peer setting and transaction confirmations.
  * Removed wallet database chain metadata caching and replaced with dynamic tip height retrieval.
  * Deleted wallet connectivity service implementation and related modules.
  * Removed base node wallet RPC mock and replaced with HTTP client mock in tests.

* **Documentation**
  * Added detailed OpenAPI schemas for new and existing API endpoints.
  * Removed outdated RPC method documentation.
  * Added OpenAPI attributes to HTTP handler query parameter structs and endpoints.

* **Chores**
  * Updated dependencies to support new features, HTTP client functionality, and documentation.
  * Reformatted dependency declarations for clarity.
  * Cleaned up imports and code formatting across multiple modules.
  * Added new trait and struct for HTTP client factory abstraction.
[...]

PR #7278, opened 2025-07-03
fix: correctly validate coinbase transactions for recovered wallets

SWvheerden:
Description
---
This ensures that coinbases are correctly validated when doing recovery. 
Current wallet, when recovery is done, will very likely mark coinbases are not mined, and never revalidate them. 
To fix this we reset those statuses during startup, 
and then we also rerun this after a large recovery. 

What causes the bug is following events:
1:  Wallet detects the coinbases, adds them to the db and marks that as not yet mined. 
2: TMS runs validation, sees them as ummined, 
3: OMS runs validation and sees them as mined, and updates them correctly
4: TMS runs validation again, but height has now changed beyond the scanned limet as in the name of "efficiency", only checks for a certain height, which the coinbases have long since expired, since the scanning the block height jumps a lot while recovering. 
5: The TMS never rechecks them again to ensure its not incorrect. 

## Summary by CodeRabbit

* **New Features**
  * Added tracking of the number of scanned blocks during UTXO scanning, with conditional revalidation of rejected transactions when a threshold is reached.

* **Improvements**
  * Enhanced handling of coinbase transactions by updating their status and clearing related fields under certain conditions.
  * Made the safety height margin constant publicly accessible for broader use.

* **Other Changes**
  * Minor internal adjustment to element access in payment records retrieval (no impact on end-user functionality).


base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -986,6 +986,17 @@ impl TransactionBackend for TransactionServiceSqliteDatabase {
             ))
             .execute(&mut conn)?;
         trace!(target: LOG_TARGET, "rows updated: {:?}", result);
+        // we want to double check unmined coinbases again, so lets set those
```
hansieodendaal, line 988: The comment on top of the function should also change, as we now include coinbases

MCozhusheck, line 988: Shouldn't we also update comment above function definition, specifically `Exclude coinbases` as now unmined coinbases are also checked now?

MCozhusheck (approved):

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -210,7 +222,10 @@ where
     }
 
     #[allow(clippy::too_many_lines)]
-    async fn attempt_sync(&mut self, peer: NodeId) -> Result<(u64, u64, MicroMinotari, Duration), UtxoScannerError> {
+    async fn attempt_sync(
+        &mut self,
+        peer: NodeId,
+    ) -> Result<(u64, u64, u64, MicroMinotari, Duration), UtxoScannerError> {
```
hansieodendaal, line 228: I think this warrants a struct

hansieodendaal (commented):
In general, not specific to this PR, we may need to add an age limit to rejected/other transactions we re-validate, for example, ignore all transactions older than a week. A user action could then force re-validation of everything.

Some commens below.

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
SWvheerden: normally I would agree, but this all changes in #7215 . So this is very much temp code.

PR #7280, opened 2025-07-03
feat: add payref background task

hansieodendaal:
Description
---
Added background task to rebuild payref indexes based on a payref rebuild status in the metadata database, rather than doing it as a migration task at stsrtup.

Motivation and Context
---
Payref migration takes very long time to complete and is not essential to base node operation.

How Has This Been Tested?
---
System-level testing [**TBD**]

---
- Code review
- System-level testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Added background rebuilding of PayRef indexes, improving index availability and tracking rebuild progress.
  * Introduced explicit status tracking for PayRef index rebuilds, with user-facing error reporting if indexes are unavailable.
* **Bug Fixes**
  * Enhanced error messages to provide more context when PayRef indexes are not available.
* **Chores**
  * Improved logging and internal management of PayRef index rebuilds.
  * Updated public exports to include PayRef rebuild status for easier access in other modules.

### DIFF
--- a/base_layer/core/src/chain_storage/async_db.rs
+++ b/base_layer/core/src/chain_storage/async_db.rs
@@ -356,6 +356,11 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self
     }
 
+    pub fn update_payref(&mut self, header_hash: HashOutput, output_hash: HashOutput) -> &mut Self {
+        self.transaction.update_payref(header_hash, output_hash);
+        self
+    }
+
     pub fn prune_output_from_all_dbs(
         &mut self,
         output_hash: HashOutput,
--- a/base_layer/core/src/chain_storage/async_db.rs
+++ b/base_layer/core/src/chain_storage/async_db.rs
@@ -356,11 +356,6 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self
     }
 
-    pub fn update_payref(&mut self, header_hash: HashOutput, output_hash: HashOutput) -> &mut Self {
-        self.transaction.update_payref(header_hash, output_hash);
-        self
-    }
-
     pub fn prune_output_from_all_dbs(
         &mut self,
         output_hash: HashOutput,
--- a/base_layer/core/src/chain_storage/blockchain_backend.rs
+++ b/base_layer/core/src/chain_storage/blockchain_backend.rs
@@ -146,7 +146,13 @@ pub trait BlockchainBackend: Send + Sync + 'static {
     /// Returns the stored payref rebuild status.
     fn fetch_payref_rebuild_status(&self) -> Result<PayrefRebuildStatus, ChainStorageError>;
     /// Updates the payref rebuild status.
-    fn update_payref_metadata_key(&self, status: PayrefRebuildStatus) -> Result<(), ChainStorageError>;
+    fn build_payref_indexes_for_height(
+        &self,
+        height: u64,
+        metadata_at_start: ChainMetadata,
+        initialize_stats: Option<u64>,
+        finalize: bool,
+    ) -> Result<PayrefRebuildStatus, ChainStorageError>;
     /// Returns the UTXO count
     fn utxo_count(&self) -> Result<usize, ChainStorageError>;
     /// Returns the kernel count
--- a/base_layer/core/src/chain_storage/blockchain_backend.rs
+++ b/base_layer/core/src/chain_storage/blockchain_backend.rs
@@ -220,4 +220,6 @@ pub trait BlockchainBackend: Send + Sync + 'static {
     fn set_stats_total_height(&self, total: u64);
     /// Update the current progress step
     fn update_stats_progress(&self, current: u64);
+    /// Provide an interface for clients to resize the lmbd environment size based on some higher level logic
+    fn resize_lmdb_if_required(&self) -> Result<(), ChainStorageError>;
 }
--- a/base_layer/core/src/chain_storage/blockchain_backend.rs
+++ b/base_layer/core/src/chain_storage/blockchain_backend.rs
@@ -220,6 +220,4 @@ pub trait BlockchainBackend: Send + Sync + 'static {
     fn set_stats_total_height(&self, total: u64);
     /// Update the current progress step
     fn update_stats_progress(&self, current: u64);
-    /// Provide an interface for clients to resize the lmbd environment size based on some higher level logic
-    fn resize_lmdb_if_required(&self) -> Result<(), ChainStorageError>;
 }
--- a/base_layer/core/src/chain_storage/blockchain_backend.rs
+++ b/base_layer/core/src/chain_storage/blockchain_backend.rs
@@ -6,7 +6,12 @@ use tari_common_types::{
     types::{BadBlock, CompressedCommitment, CompressedPublicKey, FixedHash, HashOutput, Signature},
 };
 
-use super::{lmdb_db::lmdb_tree_reader::OwnedLmdbTreeReader, MinedInfo, TemplateRegistrationEntry};
+use super::{
+    lmdb_db::lmdb_tree_reader::OwnedLmdbTreeReader,
+    MinedInfo,
+    PayrefRebuildStatus,
+    TemplateRegistrationEntry,
+};
 use crate::{
     blocks::{Block, BlockAccumulatedData, BlockHeader, BlockHeaderAccumulatedData, ChainBlock, ChainHeader},
     chain_storage::{
@@ -34,7 +39,7 @@ use crate::{
 /// us to keep the reading and writing API extremely simple. Extending the types of data that the backends can handle
 /// will entail adding to those enums, and the backends, while this trait can remain unchanged.
 #[allow(clippy::ptr_arg)]
-pub trait BlockchainBackend: Send + Sync {
+pub trait BlockchainBackend: Send + Sync + 'static {
     /// Commit the transaction given to the backend. If there is an error, the transaction must be rolled back, and
     /// the error condition returned. On success, every operation in the transaction will have been committed, and
     /// the function will return `Ok(())`.
@@ -138,6 +143,10 @@ pub trait BlockchainBackend: Send + Sync {
     fn fetch_tip_header(&self) -> Result<ChainHeader, ChainStorageError>;
     /// Returns the stored chain metadata.
     fn fetch_chain_metadata(&self) -> Result<ChainMetadata, ChainStorageError>;
+    /// Returns the stored payref rebuild status.
+    fn fetch_payref_rebuild_status(&self) -> Result<PayrefRebuildStatus, ChainStorageError>;
+    /// Updates the payref rebuild status.
+    fn update_payref_metadata_key(&self, status: PayrefRebuildStatus) -> Result<(), ChainStorageError>;
     /// Returns the UTXO count
     fn utxo_count(&self) -> Result<usize, ChainStorageError>;
     /// Returns the kernel count
--- a/base_layer/core/src/chain_storage/blockchain_backend.rs
+++ b/base_layer/core/src/chain_storage/blockchain_backend.rs
@@ -145,7 +145,7 @@ pub trait BlockchainBackend: Send + Sync + 'static {
     fn fetch_chain_metadata(&self) -> Result<ChainMetadata, ChainStorageError>;
     /// Returns the stored payref rebuild status.
     fn fetch_payref_rebuild_status(&self) -> Result<PayrefRebuildStatus, ChainStorageError>;
-    /// Updates the payref rebuild status.
+    /// Builds the payref indexes for a given block height, with stats.
     fn build_payref_indexes_for_height(
         &self,
         height: u64,
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -387,17 +387,48 @@ where B: BlockchainBackend
                 start_height, metadata_at_start.best_block_height()
             );
 
+            let mut initialize_stats = Some(metadata_at_start.best_block_height());
             for height in start_height..=metadata_at_start.best_block_height() {
+                // This is a safety check when we get close to the end to ensure we don't process block heights that
+                // are not in the database.
+                let finalize = if metadata_at_start.best_block_height() - height < 100 {
+                    match db_rw_lock
+                        .clone()
+                        .read()
+                        .map_err(|_e| ChainStorageError::AccessError("Read lock on blockchain backend failed".into()))
+                    {
+                        Ok(read_txn) => match read_txn.fetch_chain_metadata() {
+                            Ok(current_metadata) => {
+                                height == metadata_at_start.best_block_height() ||
+                                    height == current_metadata.best_block_height()
+                            },
+                            Err(e) => {
+                                error!(
+                                    target: LOG_TARGET,
+                                    "[PayRef] Failed to read chain metadata for height {}: {:?}", height, e
+                                );
+                                break;
+                            },
+                        },
+                        Err(e) => {
+                            error!(
+                                target: LOG_TARGET,
+                                "[PayRef] Failed to acquire read lock on the database for height {}: {:?}", height, e
+                            );
+                            break;
+                        },
+                    }
+                } else {
+                    false
+                };
+
+                // We can safely proceed with the rebuild if the height is less than or equal to the best block height
                 match process_payref_for_height(
                     db_rw_lock.clone(),
                     height,
                     metadata_at_start.clone(),
-                    if height == start_height {
-                        Some(metadata_at_start.best_block_height())
-                    } else {
-                        None
-                    },
-                    height == metadata_at_start.best_block_height(),
+                    initialize_stats,
+                    finalize,
                 ) {
                     Ok(current_status) => {
                         last_status = current_status;
@@ -410,6 +441,17 @@ where B: BlockchainBackend
                         );
                         break;
                     },
+                }
+                if initialize_stats.is_some() {
+                    initialize_stats = None;
+                }
+                if finalize {
+                    debug!(
+                        target: LOG_TARGET,
+                        "[PayRef] Starting index rebuilding completed, Final status: {:?}",
+                        last_status
+                    );
+                    break;
                 }
             }
         });
--- a/base_layer/core/src/chain_storage/blockchain_database.rs
+++ b/base_layer/core/src/chain_storage/blockchain_database.rs
@@ -2747,19 +2747,6 @@ fn process_payref_for_height<B: BlockchainBackend>(
         .write()
         .map_err(|_e| ChainStorageError::AccessError("Write lock on blockchain backend failed".into()))?;
 
-    // The average size added to the db per block for payrefs for the first 16,500 blocks was approximately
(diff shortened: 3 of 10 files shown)


base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,65 @@ where B: BlockchainBackend
...
+        let initial_status = {
+            let db = self.db_read_access()?;
+            db.fetch_payref_rebuild_status()?
+        };
+        if !initial_status.is_rebuilt {
+            let chain_metadata = {
+                let db = self.db_read_access()?;
+                db.fetch_chain_metadata()?
+            };
+            let db_rw_lock = self.db.clone();
```
```suggestion
        if initial_status.is_rebuilt {
        return Ok(())
        }
            let chain_metadata = {
                let db = self.db_read_access()?;
                db.fetch_chain_metadata()?
            };
            let db_rw_lock = self.db.clone();
```
This will remove an entire indent in the code.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,65 @@ where B: BlockchainBackend
...
+                        chain_metadata.clone(),
+                        height == chain_metadata.best_block_height(),
+                    ) {
+                        Ok(current_status) => {
+                            last_status = current_status;
+                            if let Ok(write_txn) = db_rw_lock.write().map_err(|_e| {
+                                ChainStorageError::AccessError("Write lock on blockchain backend failed".into())
+                            }) {
+                                write_txn.update_stats_progress(height);
+                            }
```
```suggestion

```
This should not be done on a new write txn, should be the same one used to update the payrefs for the height.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,65 @@ where B: BlockchainBackend
...
+            db.fetch_payref_rebuild_status()?
+        };
+        if !initial_status.is_rebuilt {
+            let chain_metadata = {
+                let db = self.db_read_access()?;
+                db.fetch_chain_metadata()?
+            };
+            let db_rw_lock = self.db.clone();
+
+            tokio::spawn(async move {
```
```suggestion
            tokio::task::spawn(async move {
```

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2651,6 +2710,57 @@ fn convert_to_option_bounds<T: RangeBounds<u64>>(bounds: T) -> (Option<u64>, Opt
     (start, end)
 }
 
+/// Process a batch of blocks for PayRef migration
+fn process_payref_for_height<B: BlockchainBackend>(
+    db: Arc<RwLock<B>>,
```
SWvheerden, line 2738:
```suggestion
    db: &mut B,
```
This should be a single write lock for this entire tx

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2651,6 +2710,57 @@ fn convert_to_option_bounds<T: RangeBounds<u64>>(bounds: T) -> (Option<u64>, Opt
...
+    let block_hash = header.hash();
+    let outputs = read_txn.fetch_outputs_in_block(&block_hash)?;
+    drop(read_txn);
+
+    let mut write_txn = db
+        .write()
+        .map_err(|_e| ChainStorageError::AccessError("Write lock on blockchain backend failed".into()))?;
+
+    let mut txn = DbTransaction::new();
+    for output in &outputs {
```
SWvheerden: what happens if the block was reorged while this is happening?
This entire fn needs a single write lock

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -2651,6 +2710,57 @@ fn convert_to_option_bounds<T: RangeBounds<u64>>(bounds: T) -> (Option<u64>, Opt
     (start, end)
 }
 
+/// Process a batch of blocks for PayRef migration
+fn process_payref_for_height<B: BlockchainBackend>(
```
SWvheerden, line 2737: this function can move the lmdb_rs

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1968,6 +1990,28 @@ impl LMDBDatabase {
         if let Some(output_hash) = lmdb_get::<_, HashOutput>(txn, &self.payref_to_output_index, payref.as_slice())? {
             self.fetch_mined_info_by_output_hash_in_txn(txn, &output_hash)
```
```suggestion
            return self.fetch_mined_info_by_output_hash_in_txn(txn, &output_hash)
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -1968,6 +1990,28 @@ impl LMDBDatabase {
...
+                        current_height: status.last_rebuild_height.unwrap_or_default(),
+                        start_height: if let Some(metadata) = status.metadata_at_start {
+                            metadata.best_block_height()
+                        } else {
+                            0
+                        },
+                        target_height: self.fetch_chain_metadata()?.best_block_height(),
+                    })
+                };
+            }
```
```suggestion
            if let MetadataValue::PayrefRebuildStatus(status) =
                lmdb_get::<_, MetadataValue>(txn, &self.metadata_db, &MetadataKey::PayrefRebuildStatus.as_u32())?
                    .unwrap_or(MetadataValue::PayrefRebuildStatus(PayrefRebuildStatus::default()))
            {
               if !status.is_rebuilt {
                  
                   trace!(target: LOG_TARGET, "Payref index is not completed yet, current status: {:?}", status);
                    Err(ChainStorageError::PayRefIndexNotAvailable {
                        current_height: status.last_rebuild_height.unwrap_or_default(),
                        start_height: if let Some(metadata) = status.metadata_at_start {
                            metadata.best_block_height()
                        } else {
                            0
                        },
                        target_height: self.fetch_chain_metadata()?.best_block_height(),
                    })
                };
            }
```
soo many nested ifs
this makes it harder to read

base_layer/core/src/chain_storage/async_db.rs
```diff
@@ -356,6 +356,11 @@ impl<'a, B: BlockchainBackend + 'static> AsyncDbTransaction<'a, B> {
         self
     }
 
+    pub fn update_payref(&mut self, header_hash: HashOutput, output_hash: HashOutput) -> &mut Self {
```
SWvheerden:
```suggestion
    pub fn update_payref(&mut self, header_hash: HashOutputt) -> &mut Self {
```
We can have this be a single tx,should not be multiple,

base_layer/core/src/chain_storage/blockchain_backend.rs
```diff
@@ -138,6 +143,10 @@ pub trait BlockchainBackend: Send + Sync {
     fn fetch_tip_header(&self) -> Result<ChainHeader, ChainStorageError>;
     /// Returns the stored chain metadata.
     fn fetch_chain_metadata(&self) -> Result<ChainMetadata, ChainStorageError>;
+    /// Returns the stored payref rebuild status.
+    fn fetch_payref_rebuild_status(&self) -> Result<PayrefRebuildStatus, ChainStorageError>;
+    /// Updates the payref rebuild status.
+    fn update_payref_metadata_key(&self, status: PayrefRebuildStatus) -> Result<(), ChainStorageError>;
```
SWvheerden:
```suggestion
    fn fetch_payref_rebuild_status(&self) -> Result<PayrefRebuildStatus, ChainStorageError>;
```

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -3040,10 +3108,25 @@ impl fmt::Display for MetadataKey {
...
+pub struct PayrefRebuildStatus {
+    /// Whether the payref index has been rebuilt fully - the indexes only need to be rebuilt once
+    /// and up to the current chain height. This will automatically be added to new blocks.
+    pub is_rebuilt: bool,
+    /// The height of the block at which the last rebuild was done
+    pub last_rebuild_height: Option<u64>,
+    /// The hash of the block at which the last rebuild was done
+    pub last_rebuild_block_hash: Option<BlockHash>,
+    /// The chain metadata at the start of the rebuild process
+    pub metadata_at_start: Option<ChainMetadata>,
```
SWvheerden, line 3167: why are we tracking this?
I would expect us to track the height of when the db migration goes to 5, as that's the height we can stop rebuilding payrefs, but this value gets over ridden each time and is never used?

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -3040,10 +3108,25 @@ impl fmt::Display for MetadataKey {
...
+/// Payref rebuild status - for new base nodes or once rebuilt, this will be set to true
+#[derive(Clone, Debug, Serialize, Deserialize, Default, PartialEq)]
+pub struct PayrefRebuildStatus {
+    /// Whether the payref index has been rebuilt fully - the indexes only need to be rebuilt once
+    /// and up to the current chain height. This will automatically be added to new blocks.
+    pub is_rebuilt: bool,
+    /// The height of the block at which the last rebuild was done
+    pub last_rebuild_height: Option<u64>,
+    /// The hash of the block at which the last rebuild was done
+    pub last_rebuild_block_hash: Option<BlockHash>,
```
SWvheerden: also never used

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,65 @@ where B: BlockchainBackend
...
+        };
+        if !initial_status.is_rebuilt {
+            let chain_metadata = {
+                let db = self.db_read_access()?;
+                db.fetch_chain_metadata()?
+            };
+            let db_rw_lock = self.db.clone();
+
+            tokio::spawn(async move {
+                let start_height = initial_status.last_rebuild_height.unwrap_or_default();
```
SWvheerden: you need to compare this to the previous hash

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,65 @@ where B: BlockchainBackend
...
+                    "[PayRef] Starting index rebuilding for heights {} to {}",
+                    start_height, chain_metadata.best_block_height()
+                );
+                if let Ok(write_txn) = db_rw_lock
+                    .write()
+                    .map_err(|_e| ChainStorageError::AccessError("Write lock on blockchain backend failed".into()))
+                {
+                    write_txn.set_stats_total_height(chain_metadata.best_block_height());
+                }
+                for height in start_height..=chain_metadata.best_block_height() {
```
SWvheerden: dont need to run to end, only to the height the chain was when we initially first run this,
The node might have shutdown and added a few blocks

SWvheerden (changes requested):

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: As discussed, height will be sufficient

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,110 @@ where B: BlockchainBackend
...
+                            error!(
+                                target: LOG_TARGET,
+                                "[PayRef] Failed to acquire read lock on the database for height {}: {:?}", height, e
+                            );
+                            break;
+                        },
+                    }
+                } else {
+                    false
+                };
```
```suggestion
```
Two issues with this, 
1: its on a read lock before the write lock, so the data can change again between the two. 
2: you should check if the height exists, what happens if you rewind below the metadata_at_start between iterations of the for loop? This check will never catch that.

base_layer/core/src/chain_storage/lmdb_db/lmdb.rs
```diff
@@ -66,10 +66,11 @@ where
...
         Ok(_) => {
-            trace!(
-                target: LOG_TARGET, "Inserted {} bytes with key '{}' into '{}'",
-                val_buf.len(), to_hex(key.as_lmdb_bytes()), table_name
-            );
+            // Note: Enable when detail trace logs are required for testing - this fills up trace logs pretty quickly
+            // trace!(
+            //     target: LOG_TARGET, "Inserted {} bytes with key '{}' into '{}'",
+            //     val_buf.len(), to_hex(key.as_lmdb_bytes()), table_name
+            // );
```
SWvheerden:
```suggestion
             trace!(
                 target: LOG_TARGET_X, "Inserted {} bytes with key '{}' into '{}'",
                 val_buf.len(), to_hex(key.as_lmdb_bytes()), table_name
             );
```
Either delete this, keep as is or change the target to something wth a postfix, I prefer the later, but does not matter much

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2546,13 +2586,72 @@ impl BlockchainBackend for LMDBDatabase {
...
+        let binding = self.fetch_chain_header_by_height(height)?;
+        let header = binding.header();
+        let block_hash = header.hash();
+        let outputs = self.fetch_outputs_in_block(&block_hash)?;
+
+        // Update the payref index for each output
+        for output in &outputs {
+            let output_hash = output.hash();
+            self.update_payref(&write_txn, &block_hash, &output_hash)?;
+        }
```
```suggestion
        let outputs : Vec<(Vec<u8>, TransactionOutputRowData)> =
        lmdb_fetch_matching_after(&write_txn, &db.utxos_db, block_hash.as_slice())?;

        // Update the payref index for each output
        for output in &outputs {
            self.update_payref(&write_txn, &block_hash, &output.hash)?;
        }
```
Then you don't have to hash them, that data is already in the db

SWvheerden (commented):
small things, but crucially, this is missing resize checks. 
because this is happening in async we can do the check X blocks, we need to check every tx

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,68 @@ where B: BlockchainBackend
...
+        // after this process started will have the correct payref indexes and therefor do not need to be processed.
+        let metadata_at_start = if let Some(metadata) = initial_status.metadata_at_start.clone() {
+            metadata
+        } else {
+            let db = self.db_read_access()?;
+            db.fetch_chain_metadata()?
+        };
+        let db_rw_lock = self.db.clone();
+
+        tokio::task::spawn(async move {
```
hansieodendaal, line 382: This does not stop when the base node try to shut down

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,88 @@ where B: BlockchainBackend
...
+                        "[PayRef] Starting index rebuilding completed, Final status: {:?}",
+                        last_status
+                    );
+                    break;
+                }
+            }
+        });
+
         Ok(())
     }
```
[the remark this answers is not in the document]
hansieodendaal: This version where `tokio::task::spawn(async move {` is used in conjunction with 
```
let res = tokio::task::spawn_blocking(move || {
                    process_payref_for_height(db, height, metadata, initialize_stats, finalize)
                })
                .await;
```
works well. It shuts down with the tokio environment.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2546,13 +2585,84 @@ impl BlockchainBackend for LMDBDatabase {
...
+        if height > best_block_height {
+            return Err(ChainStorageError::InvalidOperation(format!(
+                "Cannot build payref indexes for height {} which is greater than the current best block height {}",
+                height, best_block_height
+            )));
+        }
+
+        if let Some(height) = initialize_stats {
+            self.set_stats_total_height(height);
+        }
```
MCozhusheck, line 2628: `initialize_stats` is `best_block_height` but if we started migrating from  `start_height` that is not 0
```rust
for height in start_height..=metadata_at_start.best_block_height()
```
then stats could be misleading as the percentage progress wouldn't be able to reach 100%.
Not sure if such scenario is possible but it wouldn't cause any bugs anyways since so it's not something to worry about.

MCozhusheck (approved):

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
hansieodendaal: We pick up the initial status from the db if it exist. That means for a clean rebuild we get a target and once we have that target we never change it. A consecutive restart without indexes fully built will pick up the original target height again.

base_layer/core/src/chain_storage/blockchain_database.rs
```diff
@@ -352,6 +352,88 @@ where B: BlockchainBackend
...
+            let mut initialize_stats = Some(metadata_at_start.best_block_height());
+            for height in start_height..=metadata_at_start.best_block_height() {
+                let finalize = height == metadata_at_start.best_block_height();
+                let metadata = metadata_at_start.clone();
+                let db = db_rw_lock.clone();
+                // We use `spawn_blocking` with `.await` here to ensure that the async spawned task will be able to
+                // shut down when base node shutdown is triggered
+                let res = tokio::task::spawn_blocking(move || {
+                    process_payref_for_height(db, height, metadata, initialize_stats, finalize)
+                })
```
MCozhusheck, line 400: On a second thought - If we close `minotari_node` while background migration is running could closing tokio runtime cause database to be in invalid state?

function `process_payref_for_height` have this line:
```rust
    let write_txn = db
        .write()
        .map_err(|_e| ChainStorageError::AccessError("Write lock on blockchain backend failed".into()))?;
```
I believe we should have some cleanup that would release lock and stop any writings attempts while shutting down.

hansieodendaal: The most severe way we can close a base node is by killing the process, then none of our shutdown or cleanup code can run. With proper shutdown signals, we have no issues. Killing the base node process is equally detrimental to any backend operation, but having done so many times in the past, I have not seen this to be a real problem.

SWvheerden: it should not happen becasue the following happens depending on which task is spawned when the tokio runtime is shutdown:
task: will shutdown when doing an `await`
blocking_task: will not shutdown and complete its run time. 

all of the db calls are done in blocking calls. so they should finish then shutdown.

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -2546,13 +2585,83 @@ impl BlockchainBackend for LMDBDatabase {
...
+        &self,
+        height: u64,
+        metadata_at_start: ChainMetadata,
+        initialize_stats: Option<u64>,
+        finalize: bool,
+    ) -> Result<PayrefRebuildStatus, ChainStorageError> {
+        let write_txn = self.write_transaction()?;
+        unsafe {
+            LMDBStore::resize_if_required(&self.env, &self.env_config, None)?;
+        }
```
```suggestion
        unsafe {
            LMDBStore::resize_if_required(&self.env, &self.env_config, None)?;
        }
        let write_txn = self.write_transaction()?;
```

base_layer/core/src/chain_storage/blockchain_database.rs
hansieodendaal: I just confirmed a hard shutdown by killing the base node while the base node is running with payref migrations going in the background did not bother the base node. It restarted successfully and continued payref migration where it left off.

Issue #7279, opened 2025-07-03
Create replace-by-fee function


SWvheerden:
Create two methods in the wallet, 
# 1 replace by fee
 This method should ask you for an unmined tx_id, and will then recreate the tx with a different fee and create that one. It should cancel the other transaction. 

# User pay for fee
This should ask you for a tx_id of an unmined tx and allow you to spend the change/received output.

these should be cli and grpc, it should error out of the transaction is already mined.

SWvheerden:
we should allowed send to self or another users for both

SWvheerden:
if you have the following tx:
inputs = i_1, i_2
outputs = O_1 + O_C

replace by fee should create a new tx with:
inputs = i_1, i_2 (+ I_3 if needed)
outputs = O_2 + O_C2
There by spending all the inputs of the original, plus including a new input if more funds is needed to pay for the larger fee. 
Creates new outputs

User pay for fee
create a new tx with:
inputs: O_1/O_C (depending on if its receiver or sender who does this) + I_3 (if needed)
outputs: O_2 + O_C2
Takes the same unspent unmined outputs from, the original, and creates a new tx. It might need to include additional inputs.

SWvheerden:
testing scenarios
replace by fee:
```
 Given I have a seed node NODE
    When I have wallet WALLET_A connected to all seed nodes
    When I have wallet WALLET_B connected to all seed nodes
    When I have SHA3X mining node MINER connected to base node NODE and wallet WALLET_A
    When mining node MINER mines 15 blocks
    Then all nodes are at height 15
    When I wait for wallet WALLET_A to have at least 55000000000 uT    
    When I send a one-sided transaction of 100000000 uT from wallet WALLET_A to wallet WALLET_B at fee 100
    Then wallet WALLET_A detects all transactions as Pending
    When I send a replace by fee of 100000000 uT from wallet WALLET_A to wallet WALLET_B at fee 200
    Then wallet WALLET_A detects all transactions as Pending
    When mining node MINER mines 1 blocks
    Then wallet WALLET_A detects only 1 transaction as unconfirmed
```
user pay for fee
```
 Given I have a seed node NODE
    When I have wallet WALLET_A connected to all seed nodes
    When I have wallet WALLET_B connected to all seed nodes
    When I have SHA3X mining node MINER connected to base node NODE and wallet WALLET_A
    When mining node MINER mines 15 blocks
    Then all nodes are at height 15
[...]

PR #7284, opened 2025-07-07
feat: replace by fee and user pay for fee commands

MCozhusheck:
Description
---
Implements : #7279 

Add 2 new methods:
- `replace_by_fee` which takes `transaction_id` and `fee_increase` and bids the fee of the unmined transaction.
- `user_pay_for_fee` which also takes `transaction_id` but also `destination` as an adress. This methods takes outputs of unmined transaction and creates new transaction which will sends those outputs to given destination address.

Motivation and Context
---
These methods will be helpful in cases where transaction has been stuck in the mempool and also gives guarantees that the new transaction won't perform double spending.

How Has This Been Tested?
---
Run 2 new integration tests for both methods describes in `WalletTransactions.feature`  by running commands:
`cargo +nightly test --test cucumber -- --name "Adds new transaction to spend pending transaction outputs"`
and
`cargo +nightly test --test cucumber -- --name "Replace transaction with higher fee"`

---
Run 2 commands above for integration tests and do code review for potential oversights.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Added support for Replace-By-Fee (RBF) transactions, allowing users to increase the fee of a pending transaction.
  * Introduced a user-pay-for-fee feature, enabling users to create a transaction that spends outputs from a pending transaction and pays a specified fee.
  * Enhanced CLI and gRPC interfaces with new commands and endpoints for fee management.
  * Improved UTXO selection with a "must include" filter for advanced transaction control.
  * Added detailed transaction fee calculation methods to optimize fee per gram based on total fee.

* **Bug Fixes**
  * Fixed and extended integration and feature tests to cover new transaction fee scenarios.

* **Tests**
  * Added comprehensive tests for the new fee management features and UTXO selection logic.
  * Added unit tests for transaction fee calculation accuracy and boundary conditions.


applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -1635,6 +1639,32 @@ message GetFeePerGramStatsResponse {
...
+  uint64 transaction_id = 1;
+}
+
+message UserPayForFeeRequest {
+  // The transaction id of the transaction to pay for
+  uint64 transaction_id = 1;
+  // Destination address for the payment
+  bytes destination_address = 2;
+  // Amount to pay in microTari
+  uint64 amount = 3;
```
SWvheerden: use the transfer message here

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -2043,6 +2047,48 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+
+    async fn user_pay_for_fee(
+        &self,
+        request: Request<UserPayForFeeRequest>,
+    ) -> Result<Response<UserPayForFeeResponse>, Status> {
+        let request = request.into_inner();
+        let mut transaction_service = self.get_transaction_service();
+        let compressed_public_key = CompressedPublicKey::from_canonical_bytes(&request.destination_address)
+            .map_err(|e| Status::invalid_argument(format!("Invalid public key: {}", e)))?;
+        let destination_address = TariAddress::new_single_address(
```
SWvheerden: this should be replaced by the transfer,but never use the single address, as they cannot receive one-sided payments, they are only there if we don't have the view key, and only the comms key.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3966,6 +3981,168 @@ where
...
+    ) -> Result<TxId, TransactionServiceError> {
+        // Get the original pending outbound transaction
+        let original_transaction = self
+            .resources
+            .db
+            .get_pending_outbound_transaction(tx_id)
+            .map_err(|_| TransactionServiceError::TransactionAlreadyCompleted(tx_id.to_string()))?;
+
+        // Check if transaction is already confirmed
+        if original_transaction.status.is_confirmed() {
```
SWvheerden: you should check is its mined, that fn does not exist yet, but it should look at confirmed + unconfirmed

SWvheerden (commented):
looking good so far

base_layer/wallet/src/output_manager_service/input_selection.rs
```diff
@@ -71,6 +71,15 @@ impl UtxoSelectionCriteria {
             ..Default::default()
         }
     }
+
+    pub fn must_include(commitments: Vec<CompressedCommitment>) -> Self {
+        Self {
+            filter: UtxoSelectionFilter::MustInclude { commitments },
+            ordering: UtxoSelectionOrdering::Default,
+            min_dust: 0,
```
hansieodendaal: This is already default
```suggestion
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -243,6 +243,36 @@ impl OutputSql {
...
+                    // If no commitments specified, fall back to standard behavior
+                    query = query.filter(
+                        outputs::output_type
+                            .eq(i32::from(OutputType::Standard.as_byte()))
+                            .or(outputs::output_type.eq(i32::from(OutputType::Coinbase.as_byte()))),
+                    );
+
+                    if selection_criteria.excluding_onesided {
+                        query = query.filter(outputs::source.ne(OutputSource::OneSided as i32));
+                    }
```
hansieodendaal: Suggested for improved maintenance
```suggestion
                let mut selection_criteria = selection_criteria.clone();
                selection_criteria.filter = UtxoSelectionFilter::Standard;
                return OutputSql::fetch_unspent_outputs_for_spending(
                    &selection_criteria,
                    amount,
                    tip_height,
                    conn,
                );
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -274,6 +304,60 @@ impl OutputSql {
...
+                        .filter(outputs::maturity.le(i64_tip_height));
+                }
+
+                // Filter for the specific commitments
+                must_include_query = match commitments.len() {
+                    1 => must_include_query.filter(outputs::commitment.eq(commitments[0].to_vec())),
+                    _ => must_include_query.filter(
+                        outputs::commitment.eq_any::<Vec<Vec<u8>>>(commitments.iter().map(|c| c.to_vec()).collect()),
+                    ),
+                };
```
hansieodendaal: **NIT:** Maybe simplify this code for maintenance purposes? Any SQL backend optimisation would be marginal.
```suggestion
must_include_query = must_include_query.filter(
    outputs::commitment.eq_any::<Vec<Vec<u8>>>(commitments.iter().map(|c| c.to_vec()).collect())
);
```

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -274,6 +304,60 @@ impl OutputSql {
             },
         };
 
+        // Handle MustInclude case specially
```
hansieodendaal: It seems to me that this code should be integrated into the two match statements above to improve maintenance and reduce duplicate code.

applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -1635,6 +1639,32 @@ message GetFeePerGramStatsResponse {
   repeated FeePerGramStat fee_per_gram_stats = 1;
 }
 
+message ReplaceByFeeRequest {
+  // The transaction id of the transaction to replace
+  uint64 transaction_id = 1;
+  // fee per gram to use for the estimate
+  uint64 fee_per_gram = 2;
```
hansieodendaal: Maybe it would be better UX to let the user choose the minimum amount they want to increase the transaction fee by? Then the code can increase `fee_per_gram` in the estimate until the desired minimum increase in fee is met.
```suggestion
  // fee per gram increase to use for the estimate - the current fee will at least be increased by this amount
  uint64 fee_per_gram_increase = 2;
```

applications/minotari_console_wallet/src/cli.rs
```diff
@@ -222,6 +224,22 @@ pub struct SendMinotariArgs {
     pub payment_id: String,
 }
 
+#[derive(Debug, Args, Clone)]
+pub struct ReplaceByFeeArgs {
+    #[clap(short, long)]
+    pub tx_id: u64,
+    #[clap(short, long)]
+    pub new_fee_per_gram: MicroMinotari,
```
hansieodendaal: See comment in proto file regarding `fee_per_gram` vs. `fee_per_gram_increase`

base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
```diff
@@ -144,6 +144,7 @@ where
     /// Execute the Transaction Send Protocol as an async task.
     pub async fn execute(
         mut self,
+        utxo_selection_criteria: Option<UtxoSelectionCriteria>,
```
hansieodendaal: I think `utxo_selection_criteria: Option<UtxoSelectionCriteria>` should rather be part of `pub struct TransactionSendProtocol` and not passed in as a parameter.
```suggestion
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3966,6 +3981,168 @@ where
...
+        });
+
+        debug!(
+            target: LOG_TARGET,
+            "Replace-by-fee: Cancelled transaction {} for destination {} previous fee: {} new fee: {}",
+            tx_id,
+            destination,
+            original_transaction.fee,
+            fee_per_gram
+        );
```
hansieodendaal: fee_per_gram`original_transaction.fee` is a transaction fee
`fee_per_gram` is not a transaction fee

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3966,6 +3981,168 @@ where
...
+
+        // Create new transaction protocol with updated fee
+        let protocol = TransactionSendProtocol::new(
+            new_tx_id,
+            self.resources.clone(),
+            tx_reply_receiver,
+            cancellation_receiver,
+            destination.clone(),
+            original_amount,
+            fee_per_gram, // This is the new, higher fee
```
hansieodendaal: This is not guaranteed to be higher - see suggestion about providing minimum fee increase vs. fee per gram.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3966,6 +3981,168 @@ where
...
+        >,
+    ) -> Result<TxId, TransactionServiceError> {
+        let original_transaction = self
+            .resources
+            .db
+            .get_pending_outbound_transaction(tx_id)
+            .map_err(TransactionServiceError::TransactionStorageError)?;
+
+        // Check if transaction is already confirmed
+        if original_transaction.status.is_confirmed() {
```
hansieodendaal: As per the comment above, we need to check for unconfirmed and confirmed status, as a transaction with any of those statuses will already be in the blockchain.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3966,6 +3981,168 @@ where
...
+
+        // Create new transaction protocol with new outputs
+        let protocol = TransactionSendProtocol::new(
+            new_tx_id,
+            self.resources.clone(),
+            tx_reply_receiver,
+            cancellation_receiver,
+            destination.clone(),
+            amount,
+            original_transaction.fee,
```
hansieodendaal: The `fee_per_gram: MicroMinotari,` for the new transaction cannot be equal to the entire fee of the original transaction.
```suggestion
            some_other_valur_for_fee_per_gram,
```

hansieodendaal (commented):
Hi there,

Some comments to consider.

Thanks.

applications/minotari_app_grpc/proto/wallet.proto
MCozhusheck: Changed the content to use `PaymentRecipient` same as `TransferRequest` but fields `amount` and `fee_per_gram` are ignored as they will be derived from transaction outputs in case of amount and fee_per_gram will be calculated.

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
MCozhusheck: Done

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Added new function and replace it for both methods.

applications/minotari_app_grpc/proto/wallet.proto
MCozhusheck: Changed `replace_by_fee` method to accept `fee_increase` which is the increase of transaction fee which can't be zero.

applications/minotari_console_wallet/src/cli.rs
MCozhusheck: all issues of `fee_per_gram` and `fee` have been resolved in later commits.

base_layer/wallet/src/transaction_service/protocols/transaction_send_protocol.rs
MCozhusheck: Refactored as suggested.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Added check at the begging of the function to ensure increase is non-zero value now.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Added helper function to calculate it and passed estimated `fee_per_gram` based on target fee.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Done.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Fixed. Now this value is calculated based on original transaction to get it's total weight.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
MCozhusheck: This part is handled in separately since this needs to ensure that besides of given inputs there is enough amounts for new transaction with increased fees. This needs to query more inputs that are different then that were already included.

base_layer/wallet/src/output_manager_service/input_selection.rs
MCozhusheck: The field `min_dust` must  described as it constructing struct `UtxoSelectionCriteria`.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
MCozhusheck: Replaced code as suggested.

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
MCozhusheck: Implemented suggested code.

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -2126,6 +2131,114 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+        &self,
+        request: Request<UserPayForFeeRequest>,
+    ) -> Result<Response<UserPayForFeeResponse>, Status> {
+        let message = request.into_inner();
+        let recipients = message
+            .recipients
+            .into_iter()
+            .enumerate()
+            .map(|(index, transfer_with_id)| -> Result<_, String> {
+                let dest = transfer_with_id.recipient.unwrap();
```
SWvheerden: never use unwrap

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
```diff
@@ -2126,6 +2131,114 @@ impl wallet_server::Wallet for WalletGrpcServer {
...
+        let mut results = Vec::new();
+        for (address, result) in transfer_results {
+            match result {
+                Ok(tx_id) => {
+                    let wallet_address = self
+                        .wallet
+                        .get_wallet_one_sided_address()
+                        .await
+                        .map_err(|e| Status::internal(format!("{:?}", e)))?;
+                    let wallet_tx = timeout(Duration::from_millis(1000), async {
```
SWvheerden:
```suggestion
                    let wallet_tx = timeout(Duration::from_millis(100), async {
```

base_layer/common_types/src/transaction.rs
```diff
@@ -73,6 +73,13 @@ impl TransactionStatus {
         )
     }
 
+    pub fn is_mined(&self) -> bool {
+        matches!(
+            self,
+            TransactionStatus::MinedUnconfirmed | TransactionStatus::MinedConfirmed
```
SWvheerden: stupid system, but you need to catch, coinbase and one-sided as well here

base_layer/wallet/src/output_manager_service/input_selection.rs
SWvheerden: by using `..Default::default()` it means it should use default values for the rest
which should be min_dust = 0

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -243,6 +243,35 @@ impl OutputSql {
                     ),
                 };
             },
+
+            UtxoSelectionFilter::MustInclude { commitments } => {
```
SWvheerden, line 247: I think this is complex enough already, make this a second function, even if you have to copy some stuff over.

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3944,6 +3964,40 @@ where
...
     }
 
+    /// Get transactions that have been broadcast but not yet mined
+    /// Returns transactions with status 'Broadcast' that are not cancelled
+    pub fn get_broadcast_transactions(&self) -> Result<Vec<CompletedTransaction>, TransactionServiceError> {
+        let broadcast_transactions = self.db.get_transactions_to_be_broadcast()?;
+        let filtered_transactions = broadcast_transactions
+            .into_iter()
+            .filter(|tx| tx.status == TransactionStatus::Broadcast)
+            .collect();
```
SWvheerden:
```suggestion
```
dont need to refilter, again. The origin looks at completed and broadcast, you need to look at both. Also never do filtering in rust where the db can do it for you

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3944,6 +3964,40 @@ where
...
+        Ok(filtered_transactions)
+    }
+
+    /// Extract input commitments from a CompletedTransaction
+    /// This is useful when you need to get the input commitments from a broadcast transaction
+    /// Returns only the commitments that have full data (not compact/hash-only inputs)
+    pub fn get_input_commitments_from_completed_transaction(
```
SWvheerden: I would move this to completed transaction, inside of the completed tx imp
l

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4193,390 @@ where
...
+            fee_per_gram, // Calculated from total fee and weight
+            payment_id.clone(),
+            TransactionMetadata::default(),
+            None, // No reply channel for internal call
+            TransactionSendProtocolStage::Initial,
+            None,
+            Some(UtxoSelectionCriteria::must_include(original_inputs)),
+        );
+
+        // Launch the new transaction protocol with the same UTXOs (no need to cancel original)
```
SWvheerden: only make this 1-sided, currently this is interactive

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4193,390 @@ where
...
+    ///
+    /// # Returns
+    /// A tuple of (weight_in_grams, fee_per_gram) where:
+    /// - `weight_in_grams`: The calculated transaction weight in grams
+    /// - `fee_per_gram`: The fee per gram calculated from total_fee / weight_in_grams
+    ///
+    /// # Notes
+    /// Calculates actual features and scripts sizes from the original transaction outputs,
+    /// then converts total fee to fee_per_gram using floating point division to avoid truncation
+    fn calculate_fee_per_gram_from_total_fee(
```
SWvheerden: move this to completed tx

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4193,390 @@ where
...
+            tx_reply_receiver,
+            cancellation_receiver,
+            destination.clone(),
+            calculated_amount,
+            fee_per_gram,
+            original_transaction.payment_id,
+            TransactionMetadata::default(),
+            Some(reply_channel),
+            TransactionSendProtocolStage::Initial,
+            None,
```
SWvheerden: one-sided

SWvheerden (commented):
this is looking good, a few things, major thing is to use 1-sided and not interactive transactions

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -1375,6 +1375,18 @@ pub async fn command_runner(
                 );
                 println!();
             },
+            ReplaceByFee(args) => {
+                transaction_service
+                    .replace_by_fee(args.tx_id.into(), args.new_fee_per_gram)
```
hansieodendaal:
```suggestion
                    .replace_by_fee(args.tx_id.into(), args.fee_increase)
```

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -1375,6 +1375,18 @@
...
+                    .await
+                    .unwrap();
+            },
```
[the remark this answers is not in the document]
hansieodendaal: Agree, unwrap should be handled.

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -1375,6 +1375,18 @@ pub async fn command_runner(
...
+            UserPayForFee(args) => {
+                transaction_service
+                    .user_pay_for_fee(args.tx_id.into(), args.destination, args.amount)
+                    .await
+                    .unwrap();
+            },
```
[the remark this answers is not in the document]
hansieodendaal: Agree, unwrap should be handled.

applications/minotari_console_wallet/src/cli.rs
```diff
@@ -218,6 +220,22 @@ pub struct SendMinotariArgs {
...
```
hansieodendaal:
```suggestion
    pub fee_increase: MicroMinotari,
```

applications/minotari_console_wallet/src/cli.rs
```diff
@@ -218,6 +220,22 @@ pub struct SendMinotariArgs {
...
+}
+
+#[derive(Debug, Args, Clone)]
+pub struct UserPayForFeeArgs {
+    #[clap(short, long)]
+    pub tx_id: u64,
+    pub destination: TariAddress,
+    pub amount: MicroMinotari,
```
hansieodendaal: Why are the input arguments different for the command line than what it is for grpc, apart from the vec of values? Here for the 3rd argument we only have 
```rust
pub amount: MicroMinotari,
```
but for grpc we have
```rust
message TxOutputsToSpendTransfer {
    // Id of the transaction to spend outputs from.
    uint64 tx_id = 1;
    // Fee to pay for the transaction in microTari.
    uint64 fee = 2;
    // Amount and fee_per_gram are ignored. These values will be calculated automatically.
    PaymentRecipient recipient = 3;
}

// A recipient for a transfer, including address, amount, fee, and optional payment ID.
message PaymentRecipient {
  // Base58 Tari address of the recipient.
  string address = 1;
  // Amount to send in microTari (1 T = 1_000_000 µT).
  uint64 amount = 2;
  // Fee rate per gram.
  uint64 fee_per_gram = 3;
  enum PaymentType {
    // Default Mimblewimble-style transaction.
    STANDARD_MIMBLEWIMBLE = 0 [deprecated = true] ;
    // One-sided transaction (receiver not required to participate).
    ONE_SIDED = 1 [deprecated = true] ;
    // One-sided stealth address (adds privacy by hiding destination).
    ONE_SIDED_TO_STEALTH_ADDRESS = 2;
  }
  // The type of payment to perform.
  PaymentType payment_type = 5;
  // raw payment id
  bytes raw_payment_id = 6;
  // Optional user encrypted payment ID for reference (max 256 bytes).
  UserPaymentId user_payment_id = 7;
}
```

base_layer/common_types/src/transaction.rs
```diff
@@ -73,6 +73,13 @@ impl TransactionStatus {
...
+        )
+    }
```
hansieodendaal:
```suggestion
    pub fn is_mined(&self) -> bool {
        matches!(
            self,
            TransactionStatus::MinedUnconfirmed |
                TransactionStatus::MinedConfirmed |
                TransactionStatus::OneSidedUnconfirmed |
                TransactionStatus::OneSidedConfirmed |
                TransactionStatus::CoinbaseUnconfirmed |
                TransactionStatus::CoinbaseConfirmed
        )
    }
```

applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
MCozhusheck: Done

base_layer/common_types/src/transaction.rs
MCozhusheck: Added coinbase, one sided unconfirmed and one sided confirmed

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
MCozhusheck: Move it to separate function that deals with it separately.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: This was leftover I didn't clean up but I added new query that looks for broadcast and completed transactions
https://github.com/tari-project/tari/pull/7284/files#diff-721d99ed44a6cc982ea83c52ab1f074c763c4a761db123fc336ad925d9e30f8cR535

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Done

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: Changed both operations to use one sided please review them too:

https://github.com/tari-project/tari/pull/7284/files#diff-7d136ae4e0d6e98e6c4b29cde8f9151dc56e8c26137f99d8afe6b4adae20e3d2R4216

https://github.com/tari-project/tari/pull/7284/files#diff-7d136ae4e0d6e98e6c4b29cde8f9151dc56e8c26137f99d8afe6b4adae20e3d2R4400

base_layer/wallet/src/output_manager_service/input_selection.rs
MCozhusheck: My bad, there is indeed there is slice pattern that allows not to specify every field. I have also updated `pub fn specific(commitments: Vec<CompressedCommitment>) -> Self {...}` to also use default value.

applications/minotari_console_wallet/src/cli.rs
hansieodendaal: I see from `message PaymentRecipient`, we only use the recipient address, so the proto can be simplified.

base_layer/common_types/src/transaction.rs
```diff
@@ -73,6 +73,17 @@ impl TransactionStatus {
...
+    pub fn is_mined(&self) -> bool {
+        matches!(
+            self,
+            TransactionStatus::MinedUnconfirmed |
+                TransactionStatus::MinedConfirmed |
+                TransactionStatus::Coinbase |
+                TransactionStatus::OneSidedConfirmed |
+                TransactionStatus::OneSidedUnconfirmed
+        )
+    }
```
hansieodendaal:
```suggestion
    pub fn is_mined(&self) -> bool {
        matches!(
            self,
            TransactionStatus::MinedUnconfirmed |
                TransactionStatus::MinedConfirmed |
                TransactionStatus::CoinbaseUnconfirmed |
                TransactionStatus::CoinbaseConfirmed |
                TransactionStatus::OneSidedConfirmed |
                TransactionStatus::OneSidedUnconfirmed
        )
    }
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4156,270 @@ where
...
+                    },
+                }
+            } else {
+                warn!(
+                    target: LOG_TARGET,
+                    "user-pay-for-fee: Could not find transaction output for commitment {:?}",
+                    commitment
+                );
+            }
+        }
```
hansieodendaal: Consider this simplification, as only outputs that can be decrypted are owned by this wallet.
```suggestion
let all_outputs = original_transaction
    .transaction
    .body
    .outputs()
    .iter()
    .collect::<Vec<_>>();

let mut spendable_outputs = Vec::new();
let mut total_amount = MicroMinotari::zero();

let view_key = self.resources.transaction_key_manager_service.get_private_view_key().await?;

for output in all_outputs {
    match EncryptedData::decrypt_data(&view_key, output.commitment(), output.encrypted_data()) {
        Ok((amount, _, _)) => {
            total_amount += amount;
            spendable_outputs.push(output.commitment().clone());
        },
        Err(_) => {
            debug!(target: LOG_TARGET, "Output {:?} not decryptable; skipping.", output.commitment());
        }
    }
}
```
Then later on
```rust
UtxoSelectionCriteria::must_include(spendable_outputs)
```

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4156,270 @@ where
...
+
+        let calculated_amount = total_amount;
```
hansieodendaal: Why use `calculated_amount` here if it is never mutated and just equal to `total_amount`?

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +244,117 @@ impl CompletedTransaction {
...
+                    e
+                ))
+            })?;
+            let default_covenant_size = Covenant::new().get_serialized_size().map_err(|e| {
+                TransactionStorageError::FailedToCalculateTransactionFee(format!("Failed to serialize Covenant: {}", e))
+            })?;
+            default_output_features_size + default_script_size + default_covenant_size
+        } else {
+            total_features_and_scripts_size / original_outputs.len()
+        };
```
hansieodendaal: The original transaction cannot have zero outputs, and zero length can be checked above as suggested, then we can simplify this.
```suggestion
        // Calculate average size per output from original transaction
        let avg_features_and_scripts_size_per_output = total_features_and_scripts_size / original_outputs.len();
```

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +244,117 @@ impl CompletedTransaction {
...
+    /// # Notes
+    /// Calculates actual features and scripts sizes from the original transaction outputs,
+    /// then converts total fee to fee_per_gram using floating point division to avoid truncation
+    pub fn calculate_fee_per_gram_from_total_fee(
+        &self,
+        total_fee: MicroMinotari,
+        consensus_manager: &ConsensusManager,
+        num_inputs: usize,
+        num_outputs: usize,
+    ) -> Result<(u64, MicroMinotari), TransactionStorageError> {
```
hansieodendaal: This function should error if `total_fee`, `num_inputs` ,  `num_outputs` or `self.transaction.body.outputs().len()` is zero, as that is not allowed.

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +244,117 @@ impl CompletedTransaction {
...
+            features_and_scripts_size,
+        );
+
+        let fee_per_gram = if weight_in_grams > 0 {
+            let fee_per_gram_f64 = total_fee.0 as f64 / weight_in_grams as f64;
+            #[allow(clippy::cast_possible_truncation)]
+            let fee_per_gram_u64 = fee_per_gram_f64.round() as u64;
+            MicroMinotari::from(fee_per_gram_u64)
+        } else {
+            MicroMinotari::zero()
```
hansieodendaal:
```suggestion
            MicroMinotari::from(1)
```

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +244,117 @@ impl CompletedTransaction {
...
+        };
+
+        Ok((weight_in_grams, fee_per_gram))
```
hansieodendaal, line 344: Consider adding a sanity check to ensure the new fee request is met, for example.
```suggestion
        let mut fee_per_gram = MicroMinotari::from(fee_per_gram);
        loop {
            let calculated_fee = fee_calculator.calculate(
                fee_per_gram,
                1, // num_kernels = 1
                num_inputs,
                num_outputs,
                features_and_scripts_size,
            );

            if calculated_fee >= total_fee {
                break;
            }

            fee_per_gram += MicroMinotari::from(1);
        }

        Ok((weight_in_grams, fee_per_gram))
```

hansieodendaal (commented):
Hi there, looking good. Some comments to consider.

base_layer/wallet/src/transaction_service/storage/models.rs
MCozhusheck: I am bit reluctant to implement this - if calculation where indeed invalid than we would get an error and be aware that there is something off with a fee calculations. But with this sanity check we would cover up internal calculation bugs.

base_layer/wallet/src/transaction_service/storage/models.rs
MCozhusheck: Fair point. Added validation at the top of the function body for these cases.

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: This is a great way for getting spendable outputs. Change code as suggested to simplify logic.

applications/minotari_console_wallet/src/cli.rs
MCozhusheck: The idea was to use `Transfer` message to make it more similar to normal transfer. But indeed it doesn't make sense for `TxOutputsToSpendTransfer` to use whole `PaymentRecipient`.
Changed it to just take destionation address but left response to be `TransferResult` unchanged.

base_layer/wallet/src/transaction_service/storage/models.rs
MCozhusheck: Changed it as suggested.

base_layer/wallet/src/transaction_service/storage/models.rs
MCozhusheck: Added loop as final validation of whole calculation to ensure the calculated fee is higher or equal to that given as function argument.

hansieodendaal (dismissed):
Looking good.

utACK

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -4139,6 +4156,198 @@ where
...
+        if original_transaction.status.is_mined() {
+            return Err(TransactionServiceError::TransactionAlreadyMined(tx_id.to_string()));
+        }
+
+        let all_outputs = original_transaction
+            .transaction
+            .body
+            .outputs()
+            .iter()
+            .collect::<Vec<_>>();
```
SWvheerden, line 4280: what happens if you dont have all outputs in your db? Does the selection graph handle that?
If you are the recipient, it could happen that you only have the change in your db, but you have the full output here,  If you are the sender, you will only have the change.

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +241,145 @@ impl CompletedTransaction {
...
+        }
+
+        let original_outputs = self.transaction.body.outputs();
+        if original_outputs.is_empty() {
+            return Err(TransactionStorageError::FailedToCalculateTransactionFee(
+                "Transaction must have at least one output".to_string(),
+            ));
+        }
+
+        let consensus_constants = consensus_manager.consensus_constants(0);
```
SWvheerden: pass in the correct tip height here, don't use 0

SWvheerden: or easier, change the function to take in a consensus contact and not the manager

base_layer/wallet/src/transaction_service/storage/models.rs
SWvheerden: I think remove the loop, its unnessary complex and will cause a bit of loops. 
Even if we have the perfect fee per gram here, we are not going to end up with the perfect fee on the other end, keep this best effort

base_layer/wallet/src/transaction_service/storage/models.rs
```diff
@@ -237,6 +241,145 @@ impl CompletedTransaction {
...
+        let mut fee_per_gram = if weight_in_grams > 0 {
+            // Use ceiling division to ensure we never underestimate the fee
+            let fee_per_gram_u64 = total_fee.0.div_ceil(weight_in_grams);
+            // Ensure minimum of 1 (though ceiling division should handle this for positive values)
+            MicroMinotari::from(fee_per_gram_u64.max(1))
+        } else {
+            MicroMinotari::from(1)
+        };
+
+        loop {
```
SWvheerden: take out, see below

integration_tests/tests/features/WalletTransactions.feature
```diff
@@ -267,7 +261,7 @@ Feature: Wallet Transactions
   #   When I mine 6 blocks on NODE_C
   #   Then all nodes are at height 16
 
-   @missing-steps
+  @missing-steps @missing-steps
```
SWvheerden:
```suggestion
   @missing-steps
```

SWvheerden (commented):
Looking good 1 question and a small change

base_layer/wallet/src/transaction_service/service.rs
MCozhusheck: `all_outputs` are later on filtered to contain only those belonging to the user, this is done by attempting to decrypt every output and those that succeeded are later on used as outputs.
If somehow this ended up with no outputs or the total amount of outputs was zero then fee calculation would return error and propagate it back to the user.
I am no sure if given case is possible, if there is transaction id in a wallet then it should contain outputs that can be decrypted by the user.

base_layer/wallet/src/transaction_service/storage/models.rs
MCozhusheck: changed it as suggested.

SWvheerden (approved):

PR #7298, opened 2025-07-09
fix: dont start second utxo scanner for recovery

SWvheerden:
Description
---
Dont start a new utxo scanning service, simply listen to its feedback. 

## Summary by CodeRabbit

* **New Features**
  * Simplified the wallet recovery process to use the existing UTXO scanner service for monitoring progress.
* **Refactor**
  * Streamlined the recovery workflow by removing redundant parameters and tasks, resulting in a cleaner and more efficient recovery operation.
  * Removed unused message settings from the UTXO scanner configuration.
* **Chores**
  * Updated function signatures and field visibility for improved integration and maintainability.
  * Cleaned up related test setups and documentation to reflect the simplified recovery interface.

### DIFF
--- a/base_layer/wallet_ffi/src/lib.rs
+++ b/base_layer/wallet_ffi/src/lib.rs
@@ -10768,7 +10768,10 @@ mod test {
     use std::{ffi::c_void, str::from_utf8, sync::Mutex};
 
     use minotari_wallet::{
-        storage::sqlite_utilities::run_migration_and_create_sqlite_connection,
+        storage::{
+            sqlite_db::wallet::WalletSqliteDatabase,
+            sqlite_utilities::run_migration_and_create_sqlite_connection,
+        },
         transaction_service::handle::TransactionSendStatus,
     };
     use once_cell::sync::Lazy;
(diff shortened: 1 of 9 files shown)


base_layer/wallet_ffi/src/lib.rs
```diff
@@ -10133,68 +10123,11 @@ pub unsafe extern "C" fn wallet_start_recovery(
...
-        Ok(v) => v,
-        Err(e) => {
-            *error_out = LibWalletError::from(WalletError::ServiceInitializationError(e)).code;
-            return false;
-        },
-    };
-
-    let event_stream = recovery_task.get_event_receiver();
-    let recovery_join_handle = (*wallet).runtime.spawn(recovery_task.run());
+    let event_stream = (*wallet).wallet.utxo_scanner_service.clone().get_event_receiver();
```
MCozhusheck, line 10126: Is it guaranteed to exist?  As far as I understand we rely that `recovery_event_monitoring` is called after recovery was started but if it's not the case then broadcast receiver will be closed and it will end up in this match arm:
```rust
            Err(broadcast::error::RecvError::Closed) => {
                break;
            },
```
and we won't be able to monitor recovery progress.
We could verify if channel is closed first with `is_closed()` with some timeout?

SWvheerden: yes, that channel is created with the handle, which is created by the utxo scanner service. That always runs as part of the wallet

PR #7308, opened 2025-07-11
perf: dont update confirmations of tx every single time

SWvheerden:
Description
---
We keep writing to the confirmation colomn of the transaction table, and that value never gets read. 
Only update the tx if needed. 

## Summary by CodeRabbit

* **Refactor**
  * Streamlined transaction confirmation logic by removing explicit confirmation count handling.
  * Updated transaction storage methods to no longer track or require the number of confirmations.
  * Simplified logging and event publishing related to transaction confirmations.

* **Tests**
  * Adjusted tests to align with the updated transaction storage method signatures.


hansieodendaal (approved):
Very sensible.

utACK

PR #7313, opened 2025-07-14
fix: view wallet scan height

SWvheerden:
Description
---
if a view wallet does not have birthday as a start, it should start at 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing wallet birthday information to ensure a default value is used when necessary, enhancing reliability during scanning operations.

### DIFF
--- a/base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
+++ b/base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
@@ -216,7 +216,7 @@ where
             // wallet birthday
             self.resources.db.clear_scanned_blocks()?;
             let wallet_birthday = match self.resources.db.get_wallet_type()? {
-                Some(WalletType::ProvidedKeys(wallet)) => wallet.birthday,
+                Some(WalletType::ProvidedKeys(wallet)) => Some(wallet.birthday.unwrap_or_default()),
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -216,7 +216,7 @@ where
             // wallet birthday
             self.resources.db.clear_scanned_blocks()?;
             let wallet_birthday = match self.resources.db.get_wallet_type()? {
-                Some(WalletType::ProvidedKeys(wallet)) => wallet.birthday,
+                Some(WalletType::ProvidedKeys(wallet)) => Some(wallet.birthday.unwrap_or_default()),
                 _ => None,
```
MCozhusheck, line 220: Shouldn't this also return birthday as 0?

MCozhusheck (approved):

hansieodendaal (approved):
utACK

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
SWvheerden: no if this is none, it means the next function to read from database aka from cipher seed

PR #7314, opened 2025-07-14
fix: increase http server limit

SWvheerden:
Description
---
fixes http limits

## Summary by CodeRabbit

* **New Features**
  * Introduced a maximum request body size limit of 4096 bytes for the POST `/json_rpc` endpoint to enhance request handling.

### DIFF
--- a/applications/minotari_node/src/http/server.rs
+++ b/applications/minotari_node/src/http/server.rs
@@ -4,6 +4,7 @@
 use std::sync::Arc;
 
 use axum::{
+    extract::DefaultBodyLimit,
     routing::{get, post},
     Extension,
     Router,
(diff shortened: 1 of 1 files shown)


MCozhusheck (approved):
Could this limit be not enough in cases of paginated transaction list from gRPC method `get_all_completed_transactions`?
Also is this limit for whole message or just one chunk if response is a stream?

applications/minotari_node/src/http/server.rs
```diff
@@ -74,7 +75,10 @@ impl<S: BaseNodeWalletQueryService> Server<S> {
             .route("/transactions", get(handler::transaction_query::handle::<B>))
             .route("/sync_utxos_by_block", get(handler::sync_utxos_by_block::handle::<B>))
             .route("/get_utxos_by_block", get(handler::get_utxos_by_block::handle::<B>))
-            .route("/json_rpc", post(handler::json_rpc::handle::<B>))
+            .route(
+                "/json_rpc",
+                post(handler::json_rpc::handle::<B>).layer(DefaultBodyLimit::max(4096)),
```
hansieodendaal, line 80: Let's up it to what grpc uses
```suggestion
                post(handler::json_rpc::handle::<B>).layer(DefaultBodyLimit::max(8192)),
```

applications/minotari_node/src/http/server.rs
```diff
@@ -4,6 +4,7 @@
 use std::sync::Arc;
 
 use axum::{
+    extract::DefaultBodyLimit,
```
hansieodendaal, line 7: We can also add the same increased size limit to these routes:
- `"/get_utxos_mined_info",`
- `"/get_utxos_deleted_info"`
- `"/sync_utxos_by_block"`
- `"/get_utxos_by_block"`

hansieodendaal (commented):
utACK

I would up more size limits form the default, and increase it from 4MiB.

applications/minotari_node/src/http/server.rs
SWvheerden: I dont think thats needed, and thats set on the client, not server

SWvheerden:
So this only applies to the post of the transaction to the base node, higher is not needed, this needs to fit into a block still

PR #7325, opened 2025-07-16
feat: enable caching of http requests

SWvheerden:
Description
---
enables caching of http data

## Summary by CodeRabbit

* **New Features**
  * Added HTTP caching headers to multiple API endpoints, enabling improved response caching for clients.

* **Refactor**
  * Updated several API endpoints to return full HTTP responses, allowing for enhanced control over response headers and caching behavior. Endpoints affected include those for block headers, tip info, UTXOs, transaction queries, and related data.

### DIFF
--- a/applications/minotari_node/src/http/handler/get_header_by_height.rs
+++ b/applications/minotari_node/src/http/handler/get_header_by_height.rs
@@ -3,13 +3,20 @@
 
 use std::sync::Arc;
 
-use axum::{extract::Query, http::StatusCode, Extension, Json};
+use axum::{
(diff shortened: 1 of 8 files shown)


stringhandler (approved):
looks good, but should we not add in a config or env var for this

MCozhusheck (commented):
We could introduce helper function to add these headers something like `enrich_response_with_cache_policy` or something like that so cache policy is kept in a single place.

SWvheerden:
yeah: https://github.com/tari-project/tari/issues/7326

Issue #7304, opened 2025-07-11
Improve timeouts and error handling during Seedstrap


SWvheerden:
Some errors in the process are detected and handled timeously (as per #7294), for example not being able to obtain a connection or when a connection is closed while RPC streaming is taking place, but not all errors in the chain of events are handled like this. One example is when a connection is made successfully but the RPC session times out - this will result in a force kill of the process after 300s. All of this can be improved.

```rust
2025-07-09 17:39:47.438701200 [comms::dht::network_discovery::seed_strap] DEBUG Attempting to discover peers via seed nodes (SeedStrap).

'cb84ab7690c6ddd234ea7aea07' is not responsive, failed to dial

2025-07-09 17:39:47.439100300 [comms::dht::network_discovery::seed_strap] DEBUG SeedStrap: Attempt 2/5: Attempting to connect to seed peer 'cb84ab7690c6ddd234ea7aea07' to get their peer list
2025-07-09 17:40:20.462689000 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: Attempt 2/5: Failed to dial seed peer 'cb84ab7690c6ddd234ea7aea07': ConnectionFailed: Noise handshake error: timed out.

'9b748deefd1b1e1a8433027865' is the next seed peer to try, can be dialed, but has problem establishing the RPC session

2025-07-09 17:40:20.462993100 [comms::dht::network_discovery::seed_strap] DEBUG SeedStrap: Attempt 2/5: Attempting to connect to seed peer '9b748deefd1b1e1a8433027865' to get their peer list
2025-07-09 17:40:20.463680900 [comms::dht::network_discovery::seed_strap] DEBUG SeedStrap: Connected to seed peer '9b748deefd1b1e1a8433027865'. Requesting peer list.
2025-07-09 17:40:20.463689100 [comms::dht::network_discovery::seed_strap] DEBUG SeedStrap: Beginning RPC client connection to seed peer '9b748deefd1b1e1a8433027865'
2025-07-09 17:41:50.475017200 [comms::dht::network_discovery::seed_strap] ERROR SeedStrap: Failed to connect RPC client to seed peer 9b748deefd1b1e1a8433027865: Handshake error: IO Error: abb7ab8d/3: connection is closed
[...]

PR #7331, opened 2025-07-16
feat: add timeouts to all seedstrap comms calls

hansieodendaal:
Description
---
Added timeouts to all seedstrap internal comms calls. This is to guard against adverse network conditions affecting seedstrap duration, as a minimum amount of successful seed peer downloads is desired, and waiting for the state machine logic to retry seedstrap is not optimal.

Closes ##7304.

Motivation and Context
---
See #7304.

How Has This Been Tested?
---
System-level testing (I could not invoke the bad behaviour witnessed in the linked issue at this time, as it is dependent on seeds being bad/unresponsive)

---
Code review
System-level testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Added detailed timeout handling for various network discovery operations, including dialing peers, establishing RPC connections, and fetching peer streams.
  * Introduced a new timeout error message to provide clearer feedback when operations exceed expected durations.
  * Added configurable timeout settings for network discovery bootstrap with default values.

* **Bug Fixes**
  * Improved error handling to ensure peers are not incorrectly banned due to timeout errors during network discovery.


MCozhusheck (approved):
This looks great, by aren't those timeouts too strict?
```rust
// Timeout for individual operations during getting peers info from seed nodes
const DIAL_TIMEOUT: Duration = Duration::from_secs(30);
const RPC_CONNECT_TIMEOUT: Duration = Duration::from_secs(15);
const RPC_GET_PEERS_STREAM_TIMEOUT: Duration = Duration::from_secs(10);
const RPC_STREAM_ITEM_TIMEOUT: Duration = Duration::from_secs(10);
```
I am worried that many nodes will experience nodes failures in Tari Universe since there are additional points of failure.

comms/dht/src/network_discovery/seed_strap.rs
```diff
@@ -46,8 +47,11 @@ use crate::{
 // Use a reasonable value based on the existing configuration
 const DHT_RPC_MAX_PEERS_PER_REQUEST: u32 = 500;
 
-// Define a timeout for individual stream items
-const STREAM_ITEM_TIMEOUT: Duration = Duration::from_secs(10);
+// Timeout for individual operations during getting peers info from seed nodes
+const DIAL_TIMEOUT: Duration = Duration::from_secs(30);
+const RPC_CONNECT_TIMEOUT: Duration = Duration::from_secs(15);
+const RPC_GET_PEERS_STREAM_TIMEOUT: Duration = Duration::from_secs(10);
+const RPC_STREAM_ITEM_TIMEOUT: Duration = Duration::from_secs(10);
```
SWvheerden: We need to make these config options

SWvheerden (approved):

SWvheerden (changes requested):
we need the timings  these to configs

hansieodendaal:
> This looks great, by aren't those timeouts too strict?

System-level testing showed that the entire seedstrap process will complete within a couple of seconds, well withing 10s, These help the bootstrap process to make progress (choose a next seed peer candidate) and not be stagnant when it hits a glitch.

stringhandler (commented):
These timeouts are all too long and open us to abuse. They should be in the order of 2-3 seconds

SWvheerden (dismissed):

SWvheerden:
We can make the timeouts stricter, as in probably 5 secs max if, we have a latency of 1-2 seconds, that cant be the timeout. 

But more stricter timeouts should just be more dials, if they timeout, this should have 0 impact upstream to TU

hansieodendaal:
In consideration of the timeout comments, I propose these defaults. Typically dialing a peer and obtaining the RPC connection can take longer.
```
            bootstrap_dial_peer_timeout: Duration::from_secs(30),
            bootstrap_rpc_connect_timeout: Duration::from_secs(10),
            bootstrap_rpc_get_peers_stream_timeout: Duration::from_secs(5),
            bootstrap_rpc_streaming_timeout: Duration::from_secs(5),
```
Previously we only had the `bootstrap_rpc_streaming_timeout = 10s` and any other connection issue, like the peer connection being closed by some other process while streaming, or the resulted in a 5 min (300s) timeout. Peer dial and obtain RPC connection were prone to their own configured timeouts.

Currently in a seedstrap session, we contact 5 seed peers concurrently, and if any one of them has an issue like going outside these timeout boundaries, we try to sync from the next available seed peer. When all seed peers are exhausted the current seedtrap session ends, then we wait out the 300s state machine timeout before a new session starts, or when 300s have elapsed and the current session is not complete we scratch it and start a new session.

If these proposed timeouts are not acceptable, I will change all of them to 5s, but I advise against it.

SWvheerden (approved):
lets try these out

PR #7341, opened 2025-07-17
feat!: add initial validation flag to wallet state

SWvheerden:
Description
---
Sets a flag if the wallet has completed a validation after its finished scanning

Breaking
---
changes grpc proto files by adding new field

## Summary by CodeRabbit

* **New Features**
  * Added a new indicator to show whether the wallet has completed initial validation of all outputs after scanning since startup.

* **Bug Fixes**
  * Improved event handling and state tracking for wallet validation and scanning processes.

* **Tests**
  * Updated tests to reflect changes in event structure for transaction validation state.

* **Style**
  * Refined pattern matching for transaction validation events across the wallet interface.

### DIFF
--- a/applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
@@ -190,10 +190,11 @@ impl WalletDebouncer {
                                 },
                                 TransactionEvent::TransactionValidationStateChanged{faux, ..} => {
                                     self.set_refresh_needed(true).await;
+                                    #[allow(clippy::collapsible_if)]
                                     if faux {
                                         let mut intial = self.initial_validation_done.lock().await;
                                         if !(*intial) {
-                                            if self.intial_scanning_done.lock().await{
+                                            if *self.intial_scanning_done.lock().await{
                                                 // we should only set this after we completed initial scanning and then completed faux tx validation
                                                 *intial = true;
                                             }
--- a/applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
@@ -192,7 +192,12 @@ impl WalletDebouncer {
                                     self.set_refresh_needed(true).await;
                                     if faux {
                                         let mut intial = self.initial_validation_done.lock().await;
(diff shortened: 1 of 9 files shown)


applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -177,9 +185,15 @@ impl WalletDebouncer {
...
-                                TransactionEvent::TransactionValidationStateChanged(..) => {
+                                TransactionEvent::TransactionImported(_)  => {
+                                    self.set_refresh_needed(true).await;
+                                },
+                                TransactionEvent::TransactionValidationStateChanged{faux, ..} => {
                                     self.set_refresh_needed(true).await;
+                                    if faux {
+                                        let mut intial = self.initial_validation_done.lock().await;
+                                        *intial = true;
+                                    }
```
hansieodendaal: I would use the first `TransactionEvent::TransactionValidationCompleted(` to set 
```
 let mut intial = self.initial_validation_done.lock().await;
*intial = true;
```
if it is false, and then we do not have to set the `faux` flag. The `faux` flag in `TransactionEvent::TransactionValidationStateChanged` is not intuitive.

hansieodendaal (commented):
Hi, some comments.

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -48,6 +48,8 @@ pub struct WalletDebouncer {
     balance: Arc<Mutex<Balance>>,
     scanned_height: Arc<Mutex<u64>>,
     refresh_needed: Arc<Mutex<bool>>,
+    intial_scanning_done: Arc<Mutex<bool>>,
+    initial_validation_done: Arc<Mutex<bool>>,
```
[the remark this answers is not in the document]
hansieodendaal: Agree, typo

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
SWvheerden: thats not going to catch all cases, we the faux handling happens last of the tx validation, and we need to ensure this is run after scanning completed

PR #7344, opened 2025-07-18
feat!: only scan in 10 block sections

SWvheerden:
Description
---
we need to ensure the wallets scan in 10 block sections so that we can cache the results

## Summary by CodeRabbit

* **New Features**
  * Improved UTXO scanning with dynamic pagination, enabling smoother and more efficient wallet synchronization to the blockchain tip.
  * Added progress reporting and enhanced block caching during UTXO scanning.
  * Introduced a new pagination token to enhance synchronization continuity.

* **Refactor**
  * Simplified UTXO sync APIs by removing the need for an explicit end header, streamlining integration and usage for clients.
  * Updated scanning logic to iteratively sync to the tip, improving reliability and resource management.

### DIFF
--- a/base_layer/core/src/base_node/rpc/query_service.rs
+++ b/base_layer/core/src/base_node/rpc/query_service.rs
@@ -293,7 +293,7 @@ impl<B: BlockchainBackend + 'static> Service<B> {
             }
         }
 
-        let has_next_page = (end_height - current_header.height) > 0;
+        let has_next_page = (end_height.saturating_sub(current_header.height)) > 0;
 
         Ok(SyncUtxosByBlockResponse {
             blocks: utxos,
--- a/base_layer/core/src/base_node/rpc/query_service.rs
+++ b/base_layer/core/src/base_node/rpc/query_service.rs
@@ -1,8 +1,6 @@
 // Copyright 2025 The Tari Project
 // SPDX-License-Identifier: BSD-3-Clause
 
-use std::cmp;
-
 use log::trace;
 use serde_valid::{validation, Validate};
 use tari_common_types::{types, types::FixedHashSizeError};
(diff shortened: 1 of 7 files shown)


base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -192,17 +193,18 @@ impl<B: BlockchainBackend + 'static> Service<B> {
                 height: start_header_height,
             })?;
 
-        if start_header.height > end_header.height {
+        if start_header.height > tip_header.header().height {
```
stringhandler, line 196: This check doesn't seem right.

base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -269,7 +271,12 @@ impl<B: BlockchainBackend + 'static> Service<B> {
 
             fetched_utxos += 1;
 
-            if current_header.height >= end_header.height || fetched_utxos >= request.limit {
+            if current_header.height >= tip_header.header().height {
+                next_header_to_request = vec![];
```
stringhandler, line 275:
```suggestion
                next_header_to_request = None;
```

base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -280,13 +287,18 @@ impl<B: BlockchainBackend + 'static> Service<B> {
...
                         height: current_header.height + 1,
                     })?;
+            if current_header.height == end_height {
+                next_header_to_request = current_header.hash().to_vec();
+                break; // Stop if we reach the end height}
+            }
         }
 
-        let has_next_page = (end_header.height - current_header.height) > 0;
+        let has_next_page = (end_height - current_header.height) > 0;
```
stringhandler: This doesn't look right. Higher up you used number of utxos to determine if there was another page

base_layer/core/src/base_node/rpc/query_service.rs
SWvheerden: its an vec not an option type, I need to do vec<vec<u8>> if I want to do that

PR #7359, opened 2025-07-22
feat: prevent sync peers sending local addresses

hansieodendaal:
Description
---
This PR prevents peers from sending local addresses during peer sync when peer info is requested. The rationale here is that peer sync is performed by a local node to learn about the network, and then for the local node to attempt to contact those peers. 

Motivation and Context
---
- A local node can't contact a remote peer on its internal address.
- Many peer entries in the peer db consist of peers with internal addresses only.

How Has This Been Tested?
---
System-level testing for migrations.
Unit tests have been added to verify that only peers with their external addresses are supplied when requested as such.

---
Code review.
System-level test.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying and filtering peers with external network addresses.
  * Introduced database and schema changes to track whether peer addresses are external.
  * Added new test utilities for creating peers with internal or external addresses.

* **Bug Fixes**
  * Corrected a parameter name in peer info RPC to improve consistency.

* **Documentation**
  * Updated documentation to clarify address filtering and correct event value descriptions.

* **Style**
  * Improved error messages for better clarity.

* **Tests**
  * Enhanced test coverage for peer address filtering and address types.


comms/core/src/peer_manager/peer.rs
```diff
@@ -324,6 +324,34 @@ impl Peer {
     pub fn to_short_string(&self) -> String {
         format!("{}::{}", self.public_key, self.addresses)
     }
+
+    /// Returns the peer as optional without any internal addresses - None if all addresses are internal
+    #[allow(dead_code)]
```
SWvheerden: why dead code here?
Is this unused?

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1505,10 +1515,36 @@ impl PeerDatabaseSql {
...
+                external_addresses_only,
+                Some(n),
+                conn,
+            )?;
 
-            self.get_peers_by_node_ids_str(&node_ids_hex, conn)
+            let peers = self.get_peers_by_node_ids_str(&node_ids_hex, conn)?;
+            let mut peers_result = Vec::with_capacity(peers.len());
+            if external_addresses_only {
+                for peer in &peers {
```
SWvheerden: why do this again?
You asked your own db to only return external addresses, why check it again.

SWvheerden (changes requested):

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: The 1st db query returns node_ids for peers that have, amongst others, at least one external address. The 2nd query returns all peer data for the selected node ids. Then we filter out all internal addresses by converting the peer to a peer with external addresses only.

comms/core/src/peer_manager/peer.rs
hansieodendaal: No, it is used, my mistake.

comms/core/src/peer_manager/storage/database.rs
SWvheerden: cant we do this in the sql?

hansieodendaal: Most of it is already in SQL, just the last filtering of addresses in Rust when the selected peers are retrieved, no back and forth between SQL and Rust. I would think this is acceptable?

comms/core/src/peer_manager/storage/migrations/2025-07-19-085200_external_flag/down.sql
```diff
@@ -0,0 +1,2 @@
+-- Remove the external flag from the table
+ALTER TABLE multi_addresses DROP is_external;
```
SWvheerden:
```suggestion
ALTER TABLE multi_addresses DROP COLUMN is_external;
```

comms/core/src/peer_manager/storage/database.rs
SWvheerden: no, never should we filter in rust when we can filter in rust. 

1st If I call `get_active_peer_node_ids` with only external addresses set to true, I should not be expected to know I should do additional filtering. 
2nd, we can filter to this on sql level, which is multiple times faster, and does not need to serialize clone and drop stuff again.

SWvheerden (changes requested):

SWvheerden (approved):
Def would help that peers wont send local addresses. We can probably think about adding a check lcoally for this as well, as this wont stop peers sending sending locall addresses to your node. But its a step forward,

Issue #7306, opened 2025-07-11
Dont accept invalid peer data


SWvheerden:
Base nodes and wallets should also clean up their peer db.

Seedstrap phase
```rust
2025-07-10 13:11:34.102486000 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: (From Seed 'd4b68aa41844d331149939e1da', Peer Candidate 10/100): Invalid peer data for be8563b53bfa3af6721d9f6f7b received from seed peer 'd4b68aa41844d331149939e1da': Peer signature was invalid for peer 'be8563b53bfa3af6721d9f6f7b'
2025-07-10 13:11:34.117168300 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: (From Seed 'd4b68aa41844d331149939e1da', Peer Candidate 52/100): Invalid peer data for 21c8cf3c900503c165ae0d572e received from seed peer 'd4b68aa41844d331149939e1da': Peer signature was invalid for peer '21c8cf3c900503c165ae0d572e'
2025-07-10 13:11:34.193485300 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: (From Seed 'd4b68aa41844d331149939e1da', Peer Candidate 97/100): Invalid peer data for 56bd53636182e8d90e08a462ea received from seed peer 'd4b68aa41844d331149939e1da': Peer signature was invalid for peer '56bd53636182e8d90e08a462ea'
2025-07-10 13:11:34.944181400 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: (From Seed 'cb84ab7690c6ddd234ea7aea07', Peer Candidate 14/100): Invalid peer data for 0cbefbc0b35bb5a18c2bfc07fa received from seed peer 'cb84ab7690c6ddd234ea7aea07': Peer signature was invalid for peer '0cbefbc0b35bb5a18c2bfc07fa'
2025-07-10 13:11:34.953097700 [comms::dht::network_discovery::seed_strap] WARN  SeedStrap: (From Seed 'cb84ab7690c6ddd234ea7aea07', Peer Candidate 43/100): Invalid peer data for fbe55cc8c69294824c2d195db4 received from seed peer 'cb84ab7690c6ddd234ea7aea07': Peer signature was invalid for peer 'fbe55cc8c69294824c2d195db4'
[...]

PR #7361, opened 2025-07-22
feat: add tolerance for invalid peer info while busy with peer sync

hansieodendaal:
Description
---
~~Added tolerance for banable offences while busy with peer sync. A sync session will be interrupted after a configurable number of banable offences has been committed. When the sync session ends due to this threshold or otherwise, the peer will be banned for the most severe ban offence. This scheme enables a local node to accumulate good peer info sets optimally.~~

**Update:** _All sync peers emit invalid peer info sets due to random embedded invalid peer signature errors in the peer identity claims. This PR is an interim measure to combat large scale banning of peers and to collect good peer info sets from peers that send up to a threshold number of bad peer identity claims. Those sync peers will not be banned during peer sync._
_(See linked issue #7306.)_

**Note:** This does not apply to the seedstrap phase as seed peer ban offences are ignored then.

Motivation and Context
---
Peer sync efficiency can be improved if good peer info sets are accepted from sync peers in the midst of having committed ~~a~~ _this specific_ banable offence.

How Has This Been Tested?
---
System-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Added a configurable threshold (default 100) for the number of offenses tolerated during peer syncing before banning a peer.
* **Bug Fixes**
  * Improved timeout handling, error logging, and peer validation to accumulate multiple offenses before banning during peer sync.
* **Refactor**
  * Enhanced offense severity comparison with ordering and equality traits.
  * Unified and simplified peer banning logic and error handling for better robustness and clarity.

### DIFF
--- a/comms/dht/src/actor.rs
+++ b/comms/dht/src/actor.rs
@@ -136,17 +136,6 @@ pub enum OffenceSeverity {
     High,
 }
 
-impl OffenceSeverity {
-    /// Compare two OffenceSeverity values and return the maximum severity
-    pub fn max(self, other: OffenceSeverity) -> OffenceSeverity {
-        match (self, other) {
-            (OffenceSeverity::High, _) | (_, OffenceSeverity::High) => OffenceSeverity::High,
-            (OffenceSeverity::Medium, _) | (_, OffenceSeverity::Medium) => OffenceSeverity::Medium,
-            _ => OffenceSeverity::Low,
-        }
-    }
-}
-
 impl Display for DhtRequest {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         #[allow(clippy::enum_glob_use)]
--- a/comms/dht/src/actor.rs
+++ b/comms/dht/src/actor.rs
@@ -129,11 +129,22 @@ pub enum DhtRequest {
     },
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
 pub enum OffenceSeverity {
     Low,
     Medium,
     High,
+}
+
+impl OffenceSeverity {
+    /// Compare two OffenceSeverity values and return the maximum severity
+    pub fn max(self, other: OffenceSeverity) -> OffenceSeverity {
+        match (self, other) {
+            (OffenceSeverity::High, _) | (_, OffenceSeverity::High) => OffenceSeverity::High,
+            (OffenceSeverity::Medium, _) | (_, OffenceSeverity::Medium) => OffenceSeverity::Medium,
+            _ => OffenceSeverity::Low,
+        }
+    }
 }
 
 impl Display for DhtRequest {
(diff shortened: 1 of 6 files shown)


SWvheerden (changes requested):
this is way overly complicated, the issues says only count invalid peers,

hansieodendaal:
> this is way overly complicated, the issues says only count invalid peers,

`We can add a ban if the peer sends more than x amount of bad peers in a batch as a precaution to flooding bad peers.`

I will make a new one

comms/dht/src/network_discovery/discovering.rs
```diff
@@ -188,44 +214,170 @@ impl Discovering {
...
+            let GetPeersResponse { peer } = match resp {
+                Ok(val) => val,
+                Err(err) => {
+                    warn!(
+                        target: LOG_TARGET,
+                        "Discovering: Sync peer `{}` sent invalid response: {:?}",
+                        sync_peer, err
+                    );
+                    return Err(NetworkDiscoveryError::from(err));
+                },
```
SWvheerden, line 318:
```suggestion
            let GetPeersResponse { peer } = resp.inspect_err(|e|
                    warn!(
                        target: LOG_TARGET,
                        "Discovering: Sync peer `{}` sent invalid response: {:?}",
                        sync_peer, err
                    ))?
```

comms/dht/src/actor.rs
```diff
@@ -129,13 +129,24 @@ pub enum DhtRequest {
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
 pub enum OffenceSeverity {
     Low,
     Medium,
     High,
 }
 
+impl OffenceSeverity {
+    /// Compare two OffenceSeverity values and return the maximum severity
+    pub fn max(self, other: OffenceSeverity) -> OffenceSeverity {
```
SWvheerden:
```suggestion
```
not needed?

SWvheerden (changes requested):
Replace all the match errors just to log them with `inspect_err`

comms/dht/src/network_discovery/discovering.rs
hansieodendaal: Maybe I am missing something here, but the compiler does not allow it because the `inspect_err` method is designed for inspecting errors without consuming the Result, and in each case I need to return on error?

SWvheerden: hence the `?` at the end?
Doing the folloing is the same:
```
let res = match function(){
Ok(v) => v, 
Err(e) = {
                log (stuff)
                return e
                }
```
vs 
```
let res = function().inspect_err(|e| log(stuff))?;
```

stringhandler:
I disagree with the approach here. We should not be adding complicated ban logic, we will never be able to test this properly

comms/dht/src/network_discovery/discovering.rs
```diff
@@ -149,37 +144,68 @@ impl Discovering {
...
+            "Discovering: Established RPC connection to sync peer `{}`", peer_node_id
         );
         let result = self.request_peers(peer_node_id, client).await;
         self.ban_on_offence(peer_node_id.clone(), result).await?;
 
         Ok(())
     }
 
-    async fn request_peers(
+    async fn get_stream(
```
MCozhusheck, line 193:
```suggestion
    async fn get_peer_stream(
```
for more readable code

PR #7362, opened 2025-07-22
feat: only update state if changed

SWvheerden:
Description
---
Only publish a state change event if really needed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved monitoring to avoid redundant state updates when no new block is detected.
  * Enhanced trace logging to provide clearer information about block height, sync status, and latency.

### DIFF
--- a/base_layer/wallet/src/base_node_service/monitor.rs
+++ b/base_layer/wallet/src/base_node_service/monitor.rs
@@ -90,6 +90,7 @@ where TWalletConnectivity: WalletConnectivityInterface
     async fn monitor_node(&mut self, mut shutdown_signal: ShutdownSignal) -> Result<(), BaseNodeMonitorError> {
         let mut interval = interval(Duration::from_secs(10));
         interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        let mut last_checked_height = 0;
 
         loop {
             select! {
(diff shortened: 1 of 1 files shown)


base_layer/wallet/src/base_node_service/monitor.rs
```diff
@@ -126,6 +127,19 @@ where TWalletConnectivity: WalletConnectivityInterface
...
                     let best_block_height = chain_metadata.best_block_height();
+                    trace!(
+                        target: LOG_TARGET,
+                        "Base node Tip: {} ({}) Latency: {} ms",
+                        best_block_height,
+                        if is_synced { "Synced" } else { "Syncing..." },
+                        latency.as_millis()
+                    );
+
+                    if best_block_height == last_checked_height {
```
hansieodendaal: Maybe rather check best block hash as a new block can come in at the same height

hansieodendaal (commented):
Hi, please see comments below.

base_layer/wallet/src/base_node_service/monitor.rs
MCozhusheck: When comparing just height we might miss block reorg. I agree that block hash would be better suited here.

SWvheerden: yeah, changed it to hash

PR #7372, opened 2025-07-25
feat: enhance wallet connectivity monitoring

### LINKED ISSUE #7360: fix online calls for wallet
We been to ensure the grpc calls for online are working, 
In the grpc call, add a check that if the wallet has scanned a block it goes online, 

In the debouncer we can add a check that if it does not scan a block in 5 minutes, it tries to get the tip info, if that works it stays offline, if that fails it goes offline.

MCozhusheck:
Description
---
Closes #7360 

Enhance wallet connectivity check to proactively check it during wallet scanning.

Previously grpc call `check_connectivity` falsely returned `Offline` status while syncing. This was caused because it relied on `is_online` method in `https.rs` which check if last http call was done in less than a minute ago but initially no http calls were made and returned incorrect value.
Now this grpc method relies on pro-actively updated value checked every 5 minutes which verifies if tip of the scanned block has changed since the previous check.

How Has This Been Tested?
---

Replaced wallet in TU and check `get_state` and `check_connectivity` methods calls results while scanning wallet. In both cases I received `Online` status.

```
❯ grpcurl -plaintext -import-path tari/applications/minotari_app_grpc/proto -proto wallet.proto -d '{}' 127.0.0.1:42027 tari.rpc.Wallet/CheckConnectivity
{
  "status": "Online"
}
```

```
14:07:05 INFO  Wallet status: GetStateResponse { scanned_height: 54930, balance: Some(GetBalanceResponse { available_balance: 3107794619, pending_incoming_balance: 190000, pending_outgoing_balance: 2111889, timelocked_balance: 0 }), network: Some(NetworkStatusResponse { status: Online, avg_latency_ms: 18, num_node_connections: 11 }) }
14:07:05 INFO  Initial wallet scanning: 90.81590476977763% (54930/60485)
```

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Improved wallet connectivity detection with periodic background checks for more reliable online/offline status reporting.
  * Wallet status now prioritizes wallet-specific connectivity when reporting overall network status.

* **Refactor**
  * Enhanced internal synchronization for wallet state and scanning activity, resulting in more responsive and efficient wallet operations.

* **Bug Fixes**
  * Corrected naming of transaction output validation requests for consistency across the wallet and its test mocks.

* **Other**
  * Online status checks for the base node wallet client now use live server queries instead of cached timestamps for improved accuracy.


applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -2183,7 +2183,7 @@ pub async fn command_runner(
                 }
                 println!("Starting validation process");
                 let mut oms = wallet.output_manager_service.clone();
-                oms.validate_txos().await?;
+                oms.validate_utxos().await?;
```
hansieodendaal: The original name is correct, as the underlying function validates spent and unspent outputs, thus _txos_.
The enum `OutputManagerRequest::ValidateUtxos` is incorrect, and could be renamed to `OutputManagerRequest::ValidateTxos`

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -2309,7 +2309,7 @@ pub async fn command_runner(
                         .map_err(|e| CommandError::General(e.to_string()))?;
                     print!("Wallet recovery completed");
                     let mut oms = new_wallet.output_manager_service.clone();
-                    oms.validate_txos().await?;
+                    oms.validate_utxos().await?;
```
hansieodendaal: Ditto name

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -257,4 +282,65 @@ impl WalletDebouncer {
...
+            }
+        }
+    }
+
+    /// Check if height has changed since last scan
+    /// if not then get tip info if we get response we got online otherwise offline
+    async fn check_connectivity(&self) {
+        let scanned_height = self.scanned_height.load(Ordering::SeqCst);
+        let last_scan_activity = self.last_scan_activity.load(Ordering::SeqCst);
+        let mut connectivity = self.wallet.wallet_connectivity.clone();
```
hansieodendaal: We should remember and check best block hash for this test, not best block height, as there could have been a reorg to the same height.

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -257,4 +282,65 @@ impl WalletDebouncer {
...
+        }
+    }
+
+    /// Returns true if the wallet can communicate with the base node, false otherwise.
+    async fn test_base_node_connectivity(&self) -> bool {
+        let mut connectivity = self.wallet.wallet_connectivity.clone();
+        let client = connectivity.obtain_base_node_wallet_rpc_client().await;
+        let tip_info = client.get_tip_info().await;
+        tip_info.is_ok()
+    }
```
hansieodendaal: We are going to remove all this legacy code soon, so we should rather use some HTTP call to the base node for this check.

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -20,41 +20,58 @@
...
 
 /// This struct is used to get the balance of the wallet, implementing a debouncer. When the `get_balance` method is
 /// called the first time, the balance will be fetched from the backend after starting a task to monitor wallet events
 /// that could change the balance. When these wallet events are received, a flag will be set to indicate that the
 /// balance needs to be updated. When ever a client requests the balance, it will be fetched from the backend if the
 /// flag is set and clear the flag, otherwise the cached balance will be returned.
+///
+/// Additionally, this struct implements enhanced online status detection:
+/// - Tracks wallet scanning activity as an indicator of connectivity
+/// - Proactively checks connectivity when no scanning occurs for 5 minutes
```
hansieodendaal, line 58: The 5 minutes may be an issue here, as the blockchain status can legitimately remain unchanged for 20+ minutes. Maybe we need some robot-type thing status (_not literally_), where green means regular changes, orange we might have a problem and red we have one. This is some block chain timings pulled from the base node:
```
>> block-timing 10
Timing for blocks #60325 - #60316
Max block time: 134
Min block time: 7
Avg block time: 66.66666666666667
>>
>>
>> block-timing 100
Timing for blocks #60325 - #60226
Max block time: 910
Min block time: 1
Avg block time: 130.43434343434345
>>
>>
>> block-timing 1000
Timing for blocks #60325 - #59326
Max block time: 1382
Min block time: 1
Avg block time: 119.42942942942943
>
```

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
```diff
@@ -20,41 +20,58 @@
...
     transaction_service::handle::{TransactionEvent, TransactionServiceHandle},
     utxo_scanner_service::handle::{UtxoScannerEvent, UtxoScannerHandle},
+    WalletSqlite,
 };
 use tari_shutdown::ShutdownSignal;
 use tokio::sync::Mutex;
 use tonic::Status;
 
 const LOG_TARGET: &str = "wallet::ui::grpc::get_balance_debounced";
+const CONNECTIVITY_CHECK_INTERVAL: Duration = Duration::from_secs(300); // 5 minutes
```
hansieodendaal, line 48: Ditto rethink interval

hansieodendaal (commented):
Nice addition. Please see comments below.

base_layer/wallet/src/connectivity_service/interface.rs
```diff
@@ -39,5 +39,7 @@ pub trait WalletConnectivityInterface: Clone + Send + Sync + 'static {
 
     async fn get_connectivity_status(&self) -> OnlineStatus;
 
+    async fn change_connectivity_status(&mut self, status: OnlineStatus);
```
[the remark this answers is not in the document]
MCozhusheck: Added log if sending message fails.

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
MCozhusheck: In HTTP client there is [is_online](https://github.com/tari-project/tari/blob/development/clients/rust/base_node_wallet_client/src/client/http.rs#L123) that have simple but flawed logic - it would return `false` even if wallet is connected only because other methods were not called within last minute. We could make it instead call other method to the base node - let's say `get_tip_info` and if it works we could mark wallet as `online` but we also should monitor progress of wallet syncing as requested in this [issue](https://github.com/tari-project/tari/issues/7360).

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
MCozhusheck: 5 minutes was proposed in the [issue](https://github.com/tari-project/tari/issues/7360) but if state doesn't change in the given interval we also make `get_tip_info` which then decides if wallet is in `Online` state.
```rust
        if last_scan_activity == scanned_height {
            // No new blocks scanned - test if we can reach the base node
            let new_status = if self.test_base_node_connectivity().await {
                trace!(target: LOG_TARGET, "No new blocks to scan, but connectivity check successful");
                OnlineStatus::Online
            } else {
                warn!(target: LOG_TARGET, "No scanning activity and connectivity check failed");
                OnlineStatus::Offline
            };
            connectivity.change_connectivity_status(new_status).await;
        } else {
            // Recent scanning activity indicates we're online
            trace!(target: LOG_TARGET, "Recent scanning activity detected - wallet is online");
            connectivity.change_connectivity_status(OnlineStatus::Online).await;
        }
 ```

applications/minotari_console_wallet/src/automation/commands.rs
MCozhusheck: Reverted this change and renamed `OutputManagerRequest::ValidateUtxos` to `OutputManagerRequest::ValidateTxos`

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
SWvheerden: 5 min is fine, 5 min is just a lets ask the base node for a check

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
SWvheerden: yeah this is fine

SWvheerden (changes requested):
we need to update the `CheckConnectivity` grpc call as well

applications/minotari_console_wallet/src/grpc/wallet_debouncer.rs
MCozhusheck: I left it with block height for now since while scanning reorgs are not expect till it gets close to the tip height.

SWvheerden (approved):

PR #7373, opened 2025-07-25
feat: remove index tracking from commitment mask index

SWvheerden:
Description
---
Removes index tracking from commitment mask index

Motivation and Context
---
This used to be very important before we started using the key manager and we needed to keep track of more than one spend wallet on the same cipher seed. Trying to keep the indexes in sync causes massive performance issues accorss view wallets, offline singing etc. We can jut decryt the key and save it when we detect it. 

## Summary by CodeRabbit

* **Refactor**
  * Simplified key index management by removing key chain searching and index updating logic.
  * Restricted special handling in key generation to only the PRE_MINE branch; all others now use random indices.
  * Recovered keys are now directly imported without searching for existing key indices.

* **Bug Fixes**
  * Removed outdated tests and methods related to key index searching and updating.

* **Tests**
  * Removed tests covering key index finding, updating, and cross-branch index consistency. 

No user-facing features were added or removed.


hansieodendaal (approved):
utACK

Nice

PR #7374, opened 2025-07-28
fix: corruption of peer db when updating peers

SWvheerden:
Description
---
Only update peer claims for the correct node

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy when updating peer addresses to ensure changes only affect the intended peer, preventing updates from impacting multiple peers with shared addresses.


hansieodendaal (commented):
Thanks for this, I will do a system-level test and revert.

hansieodendaal (approved):
ACK

Nice catch. I did multiple restarts to accumulate well over 4,000 node peers for a test node and test wallet each in mainnet, and each time the peer_db validated all peers successfully.

PR #7382, opened 2025-07-30
feat: refresh the peer_db

hansieodendaal:
Description
---
The database needs to be refreshed to get rid of corrupted data, which was fixed by PR#7374.

Motivation and Context
---
See above

How Has This Been Tested?
---
System-level testing running migrations

---
Code review
System-level testing

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **Bug Fixes**
  * Cleared corrupt data from peer and multiaddress records to improve data integrity.
  * Reset database counters to ensure consistent data indexing.


SWvheerden (changes requested):
Wont this cause a problem where ther are not address in the peer db?

we should dump the entire db, not just one the one table, do it from rust

hansieodendaal:
> Wont this cause a problem where ther are not address in the peer db?
> 
> we should dump the entire db, not just one the one table, do it from rust

The migration dumps both the  `peers` and `multi_addresses` tables.

I will do a test with an empty database to confirm if the migration works, - I do not see any reason why it will not work. 

I have previously tested:
- New peer_db running all migrations
- Migrate from previous version where the peer_db has existing data in both the `peers` and `multi_addresses` tables

I have now confirmed - this works:
- Migrate from previous version where both the `peers` and `multi_addresses` tables are empty
- Migrate from previous version where only the `peers` table is empty
- Migrate from previous version where only the `multi_addresses` table is empty

PR #7397, opened 2025-08-06
feat: add seed peer exclusion to the dht

hansieodendaal:
Description
---
Added seed peer exclusion to the DHT. This will force peer selection to select other peers to be added to the managed DHT pools for dialing.

Motivation and Context
---
Seed peers are known entities; they have been dialled during initial seed_strap, and a well-connected network should try to learn about and connect to other peers as well.

How Has This Been Tested?
---
System-level testing.

---
- Code review.
- System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **Bug Fixes**
  * Seed peers are no longer selected as neighbours or random peers, improving routing stability and reducing unintended propagation paths.
  * Updated default peer selection counts to 6 neighbouring and 6 random peers for more balanced connectivity.

* **Documentation**
  * Updated configuration documentation to reflect corrected defaults and clearer wording for peer selection settings.

* **Chores**
  * Removed obsolete, commented-out Store-and-Forward (SAF) and DHT selection options from preset configuration files to reduce clutter.


comms/dht/src/config.rs
```diff
@@ -167,8 +167,8 @@ impl Default for DhtConfig {
         // NB: please remember to update field comments to reflect these defaults
         Self {
             protocol_version: DhtProtocolVersion::latest(),
-            num_neighbouring_nodes: 6,
-            num_random_nodes: 6,
+            num_neighbouring_nodes: 8,
+            num_random_nodes: 16,
```
SWvheerden: why make this more?

comms/dht/src/actor.rs
```diff
@@ -835,8 +904,12 @@ impl DhtActor {
                 .chain(candidates.iter())
                 .cloned()
                 .collect::<Vec<_>>();
-            // If we don't have enough connections, let's select some more disconnected peers (at least 2)
-            let n = cmp::max(config.broadcast_factor.saturating_sub(candidates.len()), 2);
+            // If we don't have enough connections, let's select some more disconnected peers (at least 2, but account
+            // for connected seed peers)
+            let n = cmp::max(
+                config.broadcast_factor.saturating_sub(candidates.len()) + connected_seeds_len,
```
SWvheerden: why are we now including the seed peers here, are they not counted above?

SWvheerden (commented):
Some questions. 
I dont get why need every progegate etc to care about out seed peer connections or not.

comms/dht/src/actor.rs
hansieodendaal: No, they are not counted. The `candidates` considers any current connections, then when we branch into `if !closest_request.connected_only {` we want to disregard/not acknowledge seed connections. As an example, if we want to broadcast to 8 peers, and we have 7 connections, but 5 of those connections are to seeds, we want to select 5 additional peers to broadcast to in addition to the 1 we are missing. This is to ensure that we are not overly reliant on seed peers to have a well-connected network.

comms/dht/src/config.rs
hansieodendaal: This is to increase the network awareness in the peer_db. As an example, running a base node for a couple of hours, it only tried to dial 23 of 894 peers on 47 of their 1063 collective multi-addresses. The DHT will compile the neighbour and random lists, then try to dial them, then with its next major cycle, refresh those and repeat the process.

comms/dht/src/actor.rs
SWvheerden: I want to make the argument that at this point in the code, we should not care. The seed nodes are still nodes, they also need to receive and will broadcast stuff out to the network. If we have 5 seed nodes at this point in time, then they are part of the 8  connections. 

The only place where we care about seed nodes is, that if we want to sync, we do not burden the seed nodes with sync. And this, by extension, comes from the liveness service. Which means we need to ensure that the liveness service won't select seed nodes to connect to when it gets other nodes.

comms/dht/src/config.rs
SWvheerden: We dont care about that. 
If we selected a good 6/6 nodes with healthy connections, we are happy. We do not need more connections.

PR #7412, opened 2025-08-13
chore: refactor transaction builder to build one sided transaction and remove interactive transaction

### LINKED ISSUE #7351: Create simplified tx builder
Currently all transactions are builded throught he stp (sender transaction protocol). This is a massive large over complicated piece of code that assumes interactive transactions which we dont use if at all atm.

We need to create a simplified one to replace all 1-sided transactions. This should only depend on core components such as outputs etc, It can allow the unblinded outputs etc. But this code should not depend on any storage etc. 
We should try and ensure this code should be safe to compile to other targets such as as wasm.

MCozhusheck:
Description
---
Fixes: #7351 

Wallet no longer support interactive transaction so we can remove it from `transaction_protocol`. With that in mind we can simplify transaction builder. Previously building one-sided transaction required:
- Create new send transaction protocol from transaction builder
- Send transaction protocol (stp) was a state machine that needed correct methods to move it next state. All state handling was done in runtime so invalid function call resulted in failed state from which stp couldn't recover.

Creating transaction should be limit to just a single building pattern with no state machine. 

How Has This Been Tested?
---
Integration and unit tests.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


MCozhusheck:
closing in favor of chore: [simplified tx builder](https://github.com/tari-project/tari/pull/7425)

PR #7434, opened 2025-08-22
feat!: update c29 fork block

hansieodendaal:
Description
---

Updated the c29 fork block to be at height `95180` and approximately `target_time = Tue, 16 Aug 2025 10:00:00 +0000`. This has been calculated as follows:

- Get block timings from the base node for the last 10,000 blocks.

```
11:24 v5.0.0-pre.4 mainnet State: Listening Tip: 77442 (Fri, 22 Aug 2025 09:22:28 +0000) ...
>>
>> block-timing 10000
Timing for blocks #77442 - #67443
Max block time: 1443
Min block time: 1
Avg block time: 121.8988898889889
>>
>>
>> watch status
11:24 v5.0.0-pre.4 mainnet State: Listening Tip: 77442 (Fri, 22 Aug 2025 09:22:28 +0000) ...
```

Our start time is the time published by the base node:
- `start_time = Fri, 22 Aug 2025 09:22:28 +0000`

Our start block is the block published by the base node:
- `start_block = 77442`

Approximate time difference in seconds: `target_time - start_time = 2162147 seconds`. Using average block time, we get `77442 + 2162147 / 121.899 = 95179.2` then rounded up to `95180`.

Motivation and Context
---
The fork block height was wrong.

How Has This Been Tested?
---
See the calculation above.

---
Code review and review the calculation in the PR description.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

BREAKING CHANGE: Consensus constant change to `con_5.effective_from_height = 95_180;`.

## Summary by CodeRabbit

* **Chores**
  * Postponed the mainnet V2 network upgrade to block height 95,000 (previously 85,000 / earlier note of 95,180).
  * Consensus and proof-of-work configuration changes will now activate at height 95,000.
  * No immediate user action required; existing nodes operate normally until the new activation height.


base_layer/core/src/consensus/consensus_constants.rs
```diff
@@ -933,7 +933,7 @@ impl ConsensusConstants {
         con_4.proof_of_work = algos;
 
         let mut con_5 = con_4.clone();
-        con_5.effective_from_height = 85_000;
+        con_5.effective_from_height = 95_180;
```
SWvheerden:
```suggestion
        con_5.effective_from_height = 95_000;
```

SWvheerden (dismissed):

PR #7435, opened 2025-08-22
feat: make key manager backend async

SWvheerden:
Description
---
this allows more proper handling of the backend writing. 
replaces memory sql db key manager in miners

## Summary by CodeRabbit

* **New Features**
  * Shared RPC models added: fee-per-gram stats and BlockHeader now include a hash.

* **Refactor**
  * Key manager initialization and related flows converted to async for non-blocking startup and tests.
  * RPC model imports consolidated so mempool/fee stats and various RPC responses come from a single source.

* **Breaking Changes**
  * Multiple public helpers/constructors are now async (require .await).
  * Validation error enums have been narrowed (some variants removed).


hansieodendaal (commented):
Looking good, some comments:
- CI shows two failing unit tests to be fixed.
- I agree with the gist of CodeRabbitAI's comments to avoid race conditions in the SQL memory db backend, highlighted by the async trait. Here is my view. In `TransactionKeyManagerBackend`:
  - Wrap the read - compute - write train in `fn increment_key_index` and `fn set_key_index` in a `conn.immediate_transaction::<`
  - In `fn insert_imported_key` we get a pooled connection, then get another one when `self.get_imported_key(&public_key).is_ok()` is called (read db), then we write to the db with the original pooled connection. Wrap the  the read - compute - write train in `fn increment_key_index` and avoid acquiring two pooled connections for this singular function. 
  - Could there be any implications of `acquire_read_lock!(self.cipher)` inside awaiting on the wrapped db methods?
  
Thanks

hansieodendaal (approved):
utACK

**Note:** We might need a high-throughput concurrency unit test, similar to `async fn test_concurrent_add_or_update_and_get_closest_peers()`

PR #7438, opened 2025-08-25
fix: tcp node peers discovery

MCozhusheck:
Description
---
Fixes base node connectivity issue with other peers when tor is disabled.

Filters out peers addresses which are unsupported on host machine - this includes:
- `onion` addresses if user has run `minotari_node` without tor.
- IPv6 addresses if host machine has no IPv6 address

This reduce amount of failed dial attempts to other peers and improves connectivity overall.

Motivation and Context
---
Some users might not want to run tor with base node so we must ensure it's still working without it.

How Has This Been Tested?
---
Run test cases in `peer_manager::storage::database::tests` which validates peer filtering.
Build `minotari_node` and replaced one in the Tari Universe and I have removed my node data directories including peer_db. I was able to pass bootstrap and enter header sync with 6 peers on mainnet.

---
Code review, running test and same as above - run node and wait until it get some connections.

Additional check `network.log` for base node to verify if `minotari_node` was attempting to peers that had only onion address. It should dial only if address is reachable.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- New Features
  - Introduced TransportProtocol to enumerate network address schemes and configure supported protocols during startup.
  - Builder option to set transport protocols; transports now advertise supported protocols and auto-detect IPv6.
  - Peer selection, discovery and dialing honor protocol reachability for more reliable connections.
- Documentation
  - Clarified that peer propagation considers address reachability.
- Tests
  - Updated tests to support protocol-aware APIs.
- Chores
  - Added runtime dependency to detect IPv6.


comms/core/src/types.rs
```diff
@@ -54,4 +56,72 @@
...
+        match addr.iter().next() {
+            Some(Protocol::Ip4(_)) => &TransportProtocol::Ipv4,
+            Some(Protocol::Ip6(_)) => &TransportProtocol::Ipv6,
+            Some(Protocol::Onion(_, _)) => &TransportProtocol::Onion,
+            Some(Protocol::Onion3(_)) => &TransportProtocol::Onion,
+            Some(Protocol::Memory(_)) => &TransportProtocol::Memory,
+            _ => &TransportProtocol::Ipv4,
+        }
+    }
+}
```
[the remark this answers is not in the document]
MCozhusheck: They have lifetime of `'static` so in this case it's not invalid.

comms/dht/src/config.rs
```diff
@@ -114,6 +114,10 @@ pub struct DhtConfig {
     /// Enables the DHT to forward messages to other nodes in the network - communication nodes only
     /// Default: false
     pub enable_forwarding: bool,
+    /// Transport protocols that should be used by the DHT
+    /// Includes protocols such as IPv4, IPv6, Tor and Memory
+    /// This list could be altered at runtime if detected to be unavailable (eg. IPv6 if host has no IPv6 address)
+    pub transport_protocols: Vec<TransportProtocol>,
 }
```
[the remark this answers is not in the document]
MCozhusheck: there is implementation of: `impl Default for DhtConfig` at line 204 in the same file.

comms/core/src/peer_manager/manager.rs
```diff
@@ -146,15 +146,16 @@ impl PeerManager {
...
     }
 
-    /// Get available dial candidates that are communication nodes, not banned, not deleted,
+    /// Get available dial candidates that are communication nodes, not banned, not deleted, reachable
     /// and not in the excluded node IDs list
     pub async fn get_available_dial_candidates(
         &self,
         exclude_node_ids: &[NodeId],
         limit: Option<usize>,
+        transport_protocols: &[TransportProtocol],
```
SWvheerden:
```suggestion
        transport_protocols: &[TransportProtocol],
```
Cant we make this a self varialbe, then we dont need to ask for it each time from the callee?

hansieodendaal: I do not think `transport_protocols` in this context is something that `PeerManager` should own - it is determined by the caller.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1569,6 +1617,23 @@ impl PeerDatabaseSql {
...
+
+        for prefix in &protocols_prefixes {
+            let like_pattern = format!("{}%", prefix);
+            let condition = multi_addresses::address.like(like_pattern);
+
+            addr_filter = match addr_filter {
+                None => Some(Box::new(condition)),
+                Some(existing) => Some(Box::new(existing.or(condition))),
+            };
+        }
```
hansieodendaal: I think these can all go into a helper function? Opportunity to DRY.

hansieodendaal (commented):
Hi, looking good. Apart from the one DRY comment, it bothers me that we have to use the config struct to pass information about exclude_ipv6. It feels to me that the check for supported protocols could also determine whether a specific protocol is available.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1633,6 +1655,21 @@ impl PeerDatabaseSql {
...
+            return None;
+        }
+
+        let conditions: Vec<String> = transport_protocols
+            .iter()
+            .map(|protocol| format!("multi_addresses.address LIKE '{}%'", sql_escape(protocol.get_prefix())))
+            .collect();
+
+        Some(conditions.join(" OR "))
+    }
```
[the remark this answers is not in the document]
MCozhusheck: I don't understand how is `Some(format!("({})", conditions.join(" OR ")))` different from `Some(conditions.join(" OR "))`?

comms/core/src/builder/mod.rs
```diff
@@ -310,10 +312,18 @@ impl CommsBuilder {
         self
     }
 
+    /// Set the transport protocols to use for communication
+    /// if not provided, defaults to IP4, IP6 and TOR
```
hansieodendaal: The comment _if not provided, defaults to IP4, IP6 and TOR_ seems out of date

comms/core/src/connectivity/proactive_dialer.rs
```diff
@@ -200,7 +200,7 @@ impl ProactiveDialer {
         // Get available dial candidates using SQL-based filtering
         let candidates = self
             .peer_manager
-            .get_available_dial_candidates(&managed, Some(count * 3)) // Get 3x more for health scoring
+            .get_available_dial_candidates(&managed, Some(count * 3))
```
hansieodendaal: I think we can leave this original comment in place
```suggestion
            .get_available_dial_candidates(&managed, Some(count * 3)) // Get 3x more for health scoring
```

comms/core/src/peer_manager/manager.rs
```diff
@@ -48,14 +48,21 @@ use crate::{
 pub struct PeerManager {
     // yo dawg, I heard you like wrappers, so I wrapped your wrapper in a wrapper so you can wrap while you wrap
     peer_storage_sql: PeerStorageSql,
+    transport_protocols: Vec<TransportProtocol>,
```
hansieodendaal, line 51: I still think `transport_protocols` is out of place in the `PeerManager` - the peer manager should not know or care about transport protocols.

comms/core/src/peer_manager/storage/database.rs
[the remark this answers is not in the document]
hansieodendaal: Agreed, it should be wrapped

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -2009,6 +2047,54 @@ mod tests {
...
+    fn test_filtering_peers_by_transport_protocols() {
+        let db_connection = DbConnection::connect_temp_file_and_migrate(MIGRATIONS).unwrap();
+        let peers_db = PeerDatabaseSql::new(
+            db_connection,
+            &create_test_peer(false, PeerFeatures::COMMUNICATION_NODE),
+        )
+        .unwrap();
+        let transport_protocols = vec![TransportProtocol::Ipv4, TransportProtocol::Ipv6];
+
+        // Create new node and wallet peers with internal and external addresses
```
hansieodendaal, line 2060: Maybe create test peers with both internal and external addresses and then some peers with internal addresses only (`create_test_peer_add_internal_addresses`).
Then we need test peers with ip6 addresses as well.

hansieodendaal (commented):
Hi looking good. Some more comments below.

comms/core/src/builder/mod.rs
MCozhusheck: Updated comment since it's default value for builder is to include all protocols:
```rust
impl Default for CommsBuilder {
    fn default() -> Self {
        Self {
            ...
            transport_protocols: TransportProtocol::get_all(),
        }
    }
}

impl CommsBuilder {
    /// Create a new `CommsBuilder` instance
    pub fn new() -> Self {
        Default::default()
    }
    ...
}
```

comms/core/src/peer_manager/manager.rs
MCozhusheck: It's convenient to have `transport_protocols` in `PeerManager` since:
- it's won't change during the life time of program (changing from TCP to TOR and vice versa requires to restart `minotari_node`) 
- if it was up to caller to decide which `transport_protocols` to include in functions calls then other struct would need to add `transport_protocols` to it's structs (DhtConfig and CommsNode would have to keep that vec and pass it everytime they call peer_manager)

SWvheerden: I dont think this is out of place at all, apart from the main reason we did this, the peer manager should know what it can speak so that it selects peers that it can talk to.

SWvheerden:
Testing this, it seems this is wrong. 
I have the node set to tcp only, but calling whoami only prints tor addresses:
Screenshot 2025-09-03 at 11 52 37

SWvheerden:
It also seems as if its still making tor connections:
Screenshot 2025-09-03 at 11 57 21
The bottom 2 addresses are inbound dials via tor

SWvheerden (approved):
This can go in as is, and we can fix the others issues in a next pr

PR #7457, opened 2025-09-01
fix: set validator registration value to 1000uT

sdbondi:
Description
---
fix: set validator registration value to 1000uT for localnet and igor

Motivation and Context
---
The transaction builder checks that you aren't sending less funds than the fee cost. This causes a
registration sending 0 to fail. Send 1000 to cover for this.

How Has This Been Tested?
---
Tested registration in tari-ootle PR  (https://github.com/tari-project/tari-ootle/pull/1560)

## Summary by CodeRabbit

* **Chores**
  * Raised the minimum validator node registration deposit to 1000 MicroMinotari on localnet and Igor.
  * Users registering validator nodes on these networks must now provide at least 1000 MicroMinotari as a deposit.
  * No changes to public APIs or user interfaces; impact is limited to validator registration requirements.


sdbondi:
Just found the setting to turn that check off in the wallet, however it may be better to use a non-zero registration amount but no rush to merge

PR #7460, opened 2025-09-01
fix(wallet): use minimum value promise in VN reg output

sdbondi:
Description
---
fix(wallet): use minimum value promise in VN reg output

Motivation and Context
---
Base node validation for VN reg requires the minimum value promise to be >= deposit amount

Ref https://github.com/tari-project/tari-ootle/pull/1560

How Has This Been Tested?
---
In PR https://github.com/tari-project/tari-ootle/pull/1560

## Summary by CodeRabbit

* **New Features**
  * Added an optional "minimum value promise" for pay-to-self transactions and UTXO aggregation.
  * Propagated across wallet transaction flows and advanced transaction options so it's available where outputs are constructed.
  * Defaults to zero when not specified, preserving existing behavior and workflows.

### DIFF
--- a/base_layer/wallet/src/output_manager_service/handle.rs
+++ b/base_layer/wallet/src/output_manager_service/handle.rs
@@ -100,6 +100,7 @@ pub enum OutputManagerRequest {
         fee_per_gram: MicroMinotari,
         lock_height: Option<u64>,
         payment_id: MemoField,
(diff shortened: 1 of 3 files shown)


hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #7471, opened 2025-09-04
chore: improvements needed in new wallet 

SWvheerden:
Description
---
Moved historical blocks out of tari core so that wallet can access them without needing tari core
Created legacy transaction status for current wallet to use, with new simplified transaction status for new wallet
Changed `try_output_key_recovery` top use private key and not key id
Add new wallet output constructor

## Summary by CodeRabbit

- New Features
  - Dual support for legacy and new transaction statuses across gRPC, wallet, and FFI.
  - Wallet seed birthday accessor added.
  - Wallet can import existing outputs via a new imported-output constructor.
  - Exposed MAX_ENCRYPTED_DATA_SIZE and added Borsh serialization for MemoField.

- Refactor
  - Block-related types moved to a shared component and a builder-based API introduced for accumulated header data.
  - Broad import-path and module surface cleanups.

- Breaking Changes
  - Wallet and related APIs now use LegacyTransactionStatus/LegacyImportStatus.
  - Key-recovery APIs now accept an owned PrivateKey instead of a key ID.


base_layer/transaction_components/src/transaction_components/wallet_output.rs
```diff
@@ -157,6 +166,40 @@ impl WalletOutput {
...
+    /// This will create a new wallet output and try and calculate the required script key and input stack to spend this
+    /// output, will return None if it cannot calculate the script key or input stack
+    #[allow(clippy::too_many_arguments)]
+    pub async fn new_imported<KM: TransactionKeyManagerInterface>(
+        value: MicroMinotari,
+        commitment_mask_key_id: TariKeyId,
+        memo: MemoField,
+        output: TransactionOutput,
+        key_manager: &KM,
+    ) -> Option<Self> {
```
MCozhusheck: Is there a specific reason we’re using Option here instead of Result? Returning a Result might help us keep the error context.

base_layer/transaction_components/src/transaction_components/wallet_output.rs
```diff
@@ -157,6 +166,40 @@ impl WalletOutput {
...
+            output.sender_offset_public_key,
+            output.metadata_signature,
+            0,
+            output.covenant,
+            output.encrypted_data,
+            output.minimum_value_promise,
+            output.proof,
+            memo,
+        );
+        let (input_data, script_key) = output.get_script_private_key_id(key_manager).await.ok()??;
```
MCozhusheck: here by calling `ok()` we are discarding error but that could make debugging easier to find out why it failed.

base_layer/transaction_components/src/transaction_components/memo_field.rs
```diff
@@ -1337,6 +1342,36 @@ impl Display for MemoField {
     }
 }
 
+impl BorshSerialize for MemoField {
```
hansieodendaal, line 1345: I commented this out, and the code compiled - suggest removing; otherwise, it has to be applied somewhere with unit tests.

base_layer/transaction_components/src/transaction_components/memo_field.rs
```diff
@@ -1337,6 +1342,36 @@ impl Display for MemoField {
...
+        let bytes = self.to_bytes();
+        writer.write_varint(bytes.len())?;
+        for b in &bytes {
+            BorshSerialize::serialize(&b, writer)?;
+        }
+        Ok(())
+    }
+}
+
+impl BorshDeserialize for MemoField {
```
hansieodendaal, line 1356: ditto remove or apply with unit tests; not used

base_layer/common_types/src/transaction.rs
```diff
@@ -130,6 +132,81 @@ pub struct TransactionConversionError {
...
+            LegacyTransactionStatus::CoinbaseUnconfirmed => write!(f, "Coinbase Unconfirmed"),
+            LegacyTransactionStatus::CoinbaseConfirmed => write!(f, "Coinbase Confirmed"),
+            LegacyTransactionStatus::CoinbaseNotInBlockChain => write!(f, "Coinbase not mined"),
+            LegacyTransactionStatus::Queued => write!(f, "Queued"),
+        }
+    }
+}
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum TransactionStatus {
```
hansieodendaal, line 182: What about coinbase types? Will it be picked up in the kernel?

hansieodendaal (changes requested):
Looking good. Some comments:
- Please remove unused Borsh ser/des code in MemoField.
- CodeRabbitAi pick up many unused old code layout references - please confirm.

base_layer/common_types/src/transaction.rs
SWvheerden: yes, it will, no need to duplicate tx types, or the output

base_layer/transaction_components/src/transaction_components/memo_field.rs
SWvheerden: its used by the new wallet, and no I am not going to add additional tests, from_bytes is already tested

base_layer/transaction_components/src/transaction_components/memo_field.rs
SWvheerden: Used by the new wallet, and no I am not going to add tests for to_bytes, its already tested

base_layer/transaction_components/src/transaction_components/wallet_output.rs
SWvheerden: good point, its used in scanning, none means no new key, but might help beter in other circumstances.

PR #7472, opened 2025-09-04
feat: add prometheus difficulty gauges for expanded metrics export

hansieodendaal:
Description
---
This PR adds new Prometheus gauges to expose difficulty metrics for display in Grafana.

Because Prometheus gauges are limited to f64 and i64, we cannot transmit the full 512-bit accumulated difficulty (U512) directly. Instead, we expose it in two forms:

1. Log_2 representation (f64) of the log_2(U512) value.

2. Binary scientific notation parts (2 × i64):
   - `exp2`: `floor(log₂(U512 value))`
   - `sig53`: top 53 bits of the U512 value, normalised to [1, 2)
   - Together, these allow reconstructing an approximate U512 value with the first ~16 significant decimal digits exact: [reconstruction: `(sig53 / 2^52) * 2^exp2`].

For target difficulty (u64), we expose:

1. Log_2 representation (f64) of the log_2(u64) value.

2. The raw 64-bit target difficulty value (i64).

Additionally, we publish the block height at which these values are sampled, lagging ~25 blocks behind the tip, to ensure stability across potential short reorgs.

**Notes:** 
- When using these metrics to identify chain divergences or forks, the published height must always be used as the X-axis reference to ensure correctness.
- Prometheus stores all metric values as f64 (64-bit floating point). Splitting a 512-bit number into multiple gauges and expecting Grafana or Prometheus to recombine them with full precision will not work, because each gauge is still limited to f64 precision. This is why the code extracts the top 53 bits (the significand that fits in an f64) and the exponent, allowing approximate reconstruction within the limits of f64 accuracy.

Motivation and Context
---
Not all pertinent accumulated difficulty values are currently available in metrics.

How Has This Been Tested?
---
Added unit tests.

---
Code review.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Added difficulty indicator metrics (target, accumulated, indicator height, derived indicators including exp2/sig53 and f64 approximation). Metrics publish automatically on new blocks and reorganizations; active only when metrics are enabled.
  - Added high-precision large-integer utilities and conversions to support those metrics.

- **Tests**
  - Added tests for metric calculations, edge cases, and precision/reconstruction.

### DIFF
--- a/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
+++ b/base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
@@ -57,7 +57,7 @@ use crate::{
             NodeCommsResponse,
             OutboundNodeCommsInterface,
         },
-        metrics::{log2_u128, log2_u512, milli_bits, u512_exp2_sig53},
+        metrics::{approximate_u512_with_f64, log2_u128, log2_u512, milli_bits, u512_exp2_sig53},
     },
     blocks::ChainBlock,
     chain_storage::{async_db::AsyncBlockchainDb, BlockAddResult, BlockchainBackend, ChainStorageError},
@@ -1102,6 +1102,8 @@ where B: BlockchainBackend + 'static
             .unwrap_or(0);
         let (acc_diff_exp2, acc_diff_sig53) =
             u512_exp2_sig53(&chain_header.accumulated_data().total_accumulated_difficulty).unwrap_or((0, 0));
+        let acc_diff_as_f64 =
+            approximate_u512_with_f64(&chain_header.accumulated_data().total_accumulated_difficulty).unwrap_or(0.0);
 
         // Publish
         metrics::accumulated_difficulty_indicator().set(acc_diff_milli_bits);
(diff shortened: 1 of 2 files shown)


base_layer/core/src/base_node/comms_interface/inbound_handlers.rs
```diff
@@ -1069,6 +1076,46 @@ where B: BlockchainBackend + 'static
...
+        let (acc_diff_exp2, acc_diff_sig53) =
+            u512_exp2_sig53(&chain_header.accumulated_data().total_accumulated_difficulty).unwrap_or((0, 0));
+
+        // Publish
+        metrics::accumulated_difficulty_indicator().set(acc_diff_milli_bits);
+        metrics::target_difficulty_indicator().set(target_diff_milli_bits);
+        #[allow(clippy::cast_possible_wrap)]
+        metrics::difficulty_indicator_height().set(height as i64);
+        #[allow(clippy::cast_possible_wrap)]
+        metrics::target_difficulty_indicator().set(chain_header.accumulated_data().target_difficulty.as_u64() as i64);
```
MCozhusheck: here we set `metrics::target_difficulty_indicator().set(...)` twice. Is it intended or should one of them be `target_difficulty().set(...);`?

hansieodendaal: Good pickup

SWvheerden (commented):
This is soo overly complicated, and needs expert knowledge of what to look for, but for now this is probably good enough

PR #7473, opened 2025-09-04
feat: claim burn kernel merkle proof

sdbondi:
Description
---
feat(node): adds database call to generate kernel merkle proofs
feat(node): adds http call to generate kernel merkle proofs
feat(wallet): adds task to automatically fetch merkle proofs after burnt output is confirmed
feat(wallet): ownership proof changes to a Schnorr signature and uses `k.G + v.H - v.H` to sign
fix(wallet): remove unnecessary lock on cipher
fix(wallet): bug if user changes confirmations to anything other than the default
chore: remove deprecated tari_async helper
test: new unit test for kernel proof generation and verification

Motivation and Context
---
Allow for "stateless" burn claims from L1 to L2. 

How Has This Been Tested?
---
Manually. New unit test that tests kernel merkle proof generation

Tested the wallet on a previous (4.9.0) mainnet wallet database and successfully synced UTXOs from height ~65000.

---

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

BREAKING CHANGE: previous burn proofs in the wallet will be deleted, breaking protobuf changes for calls to `create_burn_transaction` 

## Summary by CodeRabbit

- New Features
  - Retrieve full Burn Claim Proofs from the wallet and export confirmed proofs to JSON.
  - Node HTTP endpoint to generate kernel Merkle proofs; wallet background sync stores proofs for confirmed burns.
  - Console wallet UI: added Sidechain Key input and an improved Burn Proofs list showing confirmation status and timestamp.

- Changes
  - Ownership-proof format changed and range-proof removed; burn endpoints may return a proof optionally.
  - Burn CLI flow (burn-minotari) and related MakeItRain burn option removed.

- Bug Fixes
  - Minor log and user-facing message refinements.


applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -2060,3 +2060,29 @@ message HttpPeer {
   // Online status of the peer
   bool is_online = 3;
 }
+
+message GetBurnClaimProofRequest {
+  bytes commitment = 1;
+}
```
hansieodendaal, line 2066: `commitment` is not guaranteed to be persisted/unique in the blockchain when outputs are spent - will this only be applicable for unspent outputs?

hansieodendaal: O never mind, a burnt output will always be unspent (cannot be spent), so it will be persisted.

SWvheerden: burns will alwys be spent, it counts as a spent output, but it should be counted as unique

base_layer/wallet/src/transaction_service/handle.rs
```diff
@@ -1065,7 +1065,9 @@ impl TransactionServiceHandle {
         }
     }
 
-    /// Burns the given amount of Tari from the wallet
+    /// Burns the given amount of Tari from the wallet>
+    /// If a claim_public_key is provided, a BurnClaimProof will be returned that can be used to claim the
+    /// equivalent amount of tokens on a sidechain
```
hansieodendaal, line 1070: I think we should explain the optional `sidechain_deployment_key` as well.

hansieodendaal: What use cases are there for not claiming on  a sidechain?

base_layer/wallet/src/storage/sqlite_db/models.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub output_hash: Vec<u8>,
+    pub commitment: Vec<u8>,
+    pub burn_proof: Vec<u8>,
+    pub kernel: Vec<u8>,
+    pub kernel_merkle_proof: Option<Vec<u8>>,
+    pub created_at: chrono::NaiveDateTime,
+    pub updated_at: chrono::NaiveDateTime,
+}
+
+fn encrypt_domain(commitment: &[u8], field_name: &'static str) -> Vec<u8> {
```
hansieodendaal: This functionname suggest something is being encrypted while it is just preparing data, maybe `domain_to_encrypt`?
```suggestion
fn encrypt_domain(commitment: &[u8], field_name: &'static str) -> Vec<u8> {
```

base_layer/wallet/src/storage/sqlite_db/models.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl Encryptable<XChaCha20Poly1305> for BurntProofSql {
+    fn domain(&self, field_name: &'static str) -> Vec<u8> {
+        encrypt_domain(&self.commitment, field_name)
+    }
+
+    fn encrypt(mut self, cipher: &XChaCha20Poly1305) -> Result<Self, String> {
+        self.burn_proof =
+            encrypt_bytes_integral_nonce(cipher, self.domain("encoded_burn_proof"), Hidden::hide(self.burn_proof))?;
```
hansieodendaal, line 74: It feels like `"encoded_burn_proof"` should be a constant.
```suggestion
            encrypt_bytes_integral_nonce(cipher, self.domain("encoded_burn_proof"), Hidden::hide(self.burn_proof))?;
```

base_layer/wallet/src/storage/serializers.rs
```diff
@@ -0,0 +1,70 @@
...
+        .get(CompressedPublicKey::key_length()..)
+        .expect("signature length checked above");
+    let signature = PrivateKey::from_canonical_bytes(signature_bytes)
+        .map_err(|e| WalletStorageError::ConversionError(format!("Failed to decode signature: {}", e)))?;
+    Ok(CompressedSignature::new(pub_nonce, signature))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
hansieodendaal, line 57: Should we not have a test where we compare `encode_signature` to a known result in bytes and then `decode_signature` from that known bytes, starting with hex version of `signature`?

This way we can pick up regression in `bincode::serialize` and `bincode::deserialize` that could affect our code.

applications/minotari_console_wallet/src/automation/commands.rs
```diff
@@ -711,10 +711,15 @@ pub async fn command_runner(
                         println!("Burnt {} Minotari in tx_id: {}", args.amount, tx_id);
                         println!("The following can be used to claim the burnt funds:");
                         println!();
-                        println!("claim_public_key: {}", proof.reciprocal_claim_public_key);
-                        println!("commitment: {}", proof.commitment.to_public_key()?);
-                        println!("ownership_proof: {:?}", proof.ownership_proof);
-                        println!("ownership_proof: {:?}", proof.range_proof);
+                        if let Some(proof) = proof {
+                            println!("claim_public_key: {}", proof.reciprocal_claim_public_key);
+                            println!("commitment: {}", proof.commitment.to_public_key()?);
```
hansieodendaal: 1. `BurnMinotari(args) => {` executes `match burn_tari(` which in turn calls onto `pub async fn burn_tari(`. In the match it short-curcuits `sidechain_deployment_key: Option<PrivateKey>,` to `None`.
2. In turn, `pub async fn burn_tari(` executes `wallet_transaction_service.burn_tari(`, but it short-curcuits `claim_public_key: Option<CompressedPublicKey>,` to `None`.
3. In it's current form, burning XTM via the command line, those funds will be lost forever.
4. `pub struct BurnMinotariArgs {` must make provision for `claim_public_key` and `sidechain_deployment_key` and those should be wired in.

```suggestion
                        if let Some(proof) = proof {
                            println!("claim_public_key: {}", proof.reciprocal_claim_public_key);
                            println!("commitment: {}", proof.commitment.to_public_key()?);
```

applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -1576,10 +1576,10 @@ message CreateBurnTransactionResponse{
   uint64 transaction_id = 1;
   bool is_success = 2;
   string failure_message = 3;
+  // Only provided if a burn claim key is provided
   bytes commitment = 4;
-  CommitmentSignature ownership_proof = 5;
-  bytes range_proof = 6;
-  bytes reciprocal_claim_public_key = 7;
```
hansieodendaal, line 1582: I think we should deperecate the fileds not in use anymore and make new entries for new fields.

hansieodendaal (commented):
Looks good. I have some general questions (maybe not all for this PR) and some comments.

- Can burn proofs be recovered/rebuilt when we recover a burnt output? The wallet could be lost just after a burn transaction has been mined.
- Do we have any old-style burn commitments on the blockchain? What if someone burns XTM using a previous version of the code? Do we not need to support those as well, from the wallet's perspective, but also using it in the Ootle?

applications/minotari_app_grpc/proto/wallet.proto
sdbondi: Not necessary at this stage, nothing is using this except L2

base_layer/wallet/src/transaction_service/handle.rs
sdbondi: None == Ootle - we apparently want to support multiple side chains

base_layer/wallet/src/storage/sqlite_db/models.rs
sdbondi: yeah can do, I'm just copying the code that is in the rest of the wallet - remember this wallet is being rewritten

base_layer/wallet/src/storage/sqlite_db/models.rs
sdbondi: Changed to `get_encryption_domain`

base_layer/wallet/src/storage/serializers.rs
sdbondi: Changes to bincode affect the blockchain db, so that would be picked up there

base_layer/wallet/src/output_manager_service/storage/output_status.rs
```diff
@@ -40,6 +40,26 @@ pub enum OutputStatus {
     ShortTermEncumberedToBeSpent,
     SpentMinedUnconfirmed,
     NotStored,
+    Burnt,
```
SWvheerden:
```suggestion
```
I dont think this is a good idea, 
Burnt can have all of the above as well.
We should track it using the features not statuements

sdbondi: Ok will take a look - it might be that the status was Unspent before and there's no way in the DB to query for burnt, and this was the easiest way around it (also doing the least work because of the wallet rewrite), but I agree in principle, so will see what needs to be done

sdbondi: Might need to add a column in this case to allow us to pick out burnt UTXOs - but let's see

EDIT: Oh I see OutputType::Burn should be ok

sdbondi: Removed the Burnt variant

applications/minotari_console_wallet/src/automation/commands.rs
sdbondi: Going to remove this feature as it's not used (discussed with @SWvheerden)

SWvheerden (approved):

PR #7482, opened 2025-09-08
feat: add timeouts to health check

### LINKED ISSUE #7380: Fix seed peer health check lock not being released
This log entry will be encountered when the lock does not release, and it has been observed frequently in esmeralda and mainnet:
```
[c::bn::tari_pulse] DEBUG Could not acquire lock for health check, skipping this tick
```
The probable cause is that pongs are never returned from the liveness service once pings are sent.

Suggested approach:
- Add tokio timeout guards around each `.await` block inside `async fn check_health(` and add trace/debug logs to indicate which call got stuck - discovery, pin + pong, disconnect
- The health check fires every 10 minutes - each spawned async health check in `async fn check_health(` should complete within <=3 minutes.

hansieodendaal:
Description
---
- This PR fixes an interlock in the health check whereby either discover-peer or ping-peer locks up and does not return. This culminated in the health check being interlocked for the remainder of the base node's uptime.
- The discover-peer and ping-peer health checks are now run in parallel.
- The ping-peer health check is now preceded by a dial-peer to ensure we have a connection before we proceed.
- Changed the gRPC `liveness_results` in the `GetNetworkState` method to return `ms` instead of `s` and documented it as such; this was previously undocumented. When the network is well connected, discovery and ping latencies would be less than `1 s`, resulting in zero values to be returned.

Fixes #7380.

Motivation and Context
---
See #7380.

How Has This Been Tested?
---
- System-level testing  (`c::bn::tari_pulse` logged at least trace level)
- When running the gRPC `GetNetworkState` method, `liveness_results` should not be an empty array if seed peers are contactable.

---
Code review.
System-level testing  (to verify in the log files change `c::bn::tari_pulse` to at least debug level).

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: Changed the gRPC `liveness_results` in the `GetNetworkState` method to return `ms` instead of `s` and documented it as such; this was previously undocumented.

## Summary by CodeRabbit

* **New Features**
  * Health-check interval is now optional and disabled by default.

* **Bug Fixes**
  * More reliable per-peer health checks with clearer timeouts, improved connection cleanup, and fewer hangs/false negatives.
  * Service startup now waits for the first liveness event to ensure initial health state.

* **Refactor**
  * Discovery and ping checks run in parallel for faster reporting.
  * Latency measurements standardized to milliseconds.

* **Documentation**
  * Config comment updated to indicate health checks default to disabled.


base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -329,9 +332,22 @@ async fn check_health(
     notify_comms_health: watch::Sender<Vec<LivenessCheckResult>>,
 ) {
     let results = Arc::new(RwLock::new(Vec::new()));
-    let peers = node_comms.get_seeds().await.unwrap_or_else(|_| vec![]);
+
+    let peers = match tokio::time::timeout(Duration::from_secs(10), node_comms.get_seeds()).await {
+        Ok(Ok(seeds)) => seeds,
+        Ok(Err(err)) => {
+            warn!(target: LOG_TARGET, "check_health: failed to get seed peers: {}", err);
+            return;
```
MCozhusheck, line 386: previously on fail we returned empty array of peers `unwrap_or_else(|_| vec![]);`
but here we are early exiting the function. We keep the old behavior.
```suggestion
            vec![]
 ```

File: base_layer/core/src/base_node/tari_pulse_service/mod.rs [hunk 2]
```diff
@@ -329,9 +332,22 @@ async fn check_health(
...
+        },
+        Err(_) => {
+            warn!(target: LOG_TARGET, "check_health: timeout getting seed peers");
+            return;
```
MCozhusheck, line 390: Same case here
```suggestion
            vec![]
```

base_layer/core/src/base_node/tari_pulse_service/mod.rs
SWvheerden: I think this is fine, lets not suppress the error

SWvheerden (commented):
Can we add a config option to disable this and make it disabled by default

base_layer/core/src/base_node/tari_pulse_service/mod.rs
```diff
@@ -123,21 +124,27 @@ impl TariPulseService {
         notify_passed_checkpoints: watch::Sender<bool>,
         notify_comms_health: watch::Sender<Vec<LivenessCheckResult>>,
     ) {
+        tokio::time::sleep(Duration::from_secs(30)).await; // Wait for the node to start up properly
```
SWvheerden: why is this required?

hansieodendaal: The health check starts runnig before the node is properly bootsrapped, resulting in unneccessary warnings, and we want to give preference to seedstrap to start.

SWvheerden: cant we do something else, like listen for seedstrap done event?

hansieodendaal: I added listening for the first liveness event instead of the sleep.

SWvheerden (approved):

PR #7486, opened 2025-09-09
fix: filter node identity onion address when running TCP node

### LINKED ISSUE #7467: Base node keeps old tor address eventhou it only has a single new address
The base node was started up and it generated a tor id, the tor file was deleted and and a new tor identity was created. The base node now reports two addresses under whoami. 
This is incorrect

MCozhusheck:
Description
---
Fixes: #7467

TCP only node would create or load from config file tor address even if it was unreachable causing failed dialed attempts from other peers.

>> whoami
Public Key: 20c62d010f073410950a00b404b21f6c91d69595ff6e19dba884b067dd100a37
Node ID: e9fa93f2ae080380a9593ec1d7
Public Addresses: /ip4/127.0.0.1/tcp/36707, /onion3/fdsqzg3pcloty5ud3glwe7v6kvbcnl3pubiq6ufnsvwvq2gaifovlsad:18141
Features: PeerFeatures(MESSAGE_PROPAGATION | DHT_STORE_FORWARD)

This PR filter out onion address and also internal tcp addresses when loading identity config.

How Has This Been Tested?
---
Build `minotari_node` binary and run it with TCP only and check if `whoami` command returns only a single IP address. Check it also for TOR  node if `whoami` return both onio and IP address. 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Transport-aware identity setup that filters advertised public addresses according to selected transport, with clearer debug logging.
  * Improved defaults for TCP by excluding onion and localhost addresses.

* **Bug Fixes**
  * Prevents nodes from advertising unreachable or local addresses when using TCP.

* **Chores**
  * Dependency configured to use workspace resolution for the p2p component.


MCozhusheck:
@coderabbit review

applications/minotari_app_utilities/src/identity_management.rs
```diff
@@ -121,11 +136,37 @@ pub fn setup_node_identity<P: AsRef<Path>>(
...
+            "After filtering for TCP transport, {} addresses remain: {:?}",
+            filtered_addresses.len(),
+            filtered_addresses
+        );
+        id.set_public_addresses(filtered_addresses);
+        save_as_json(&path, &id)?;
+        id
+    } else {
+        id
+    };
```
hansieodendaal: Let's assume we start up tcp from a previous mixed tcp/tor type id fle, then we change the type to tcp and save the file, then we start again as tor, then the id file does not contain tor info anymore? How will the tor info be refreshed?

hansieodendaal (commented):
Hi, looking good. Please answer the question below?

applications/minotari_app_utilities/src/identity_management.rs
SWvheerden: I think we need to change the update section to always update, tcp/tor. It should be fixed to latest always.

MCozhusheck: Applied suggestion. Now JSON file will be update in every cases so it's always up to date.

hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #7488, opened 2025-09-10
fix: discovery syncing regression bug

hansieodendaal:
Description
---
- Fixed regression bug due to a recent PR. When responding to a discovery syncing request, a node's own transport capabilities should not dictate which peers are returned. The call originates in `async fn get_peers(&self, request: Request<GetPeersRequest>) -> Result<Streaming<GetPeersResponse>, RpcStatus> {`.
- Also fixed discovery syncing requests to not request clients anymore, as old wallet addresses should not be going around.

Motivation and Context
---
See above.

How Has This Been Tested?
---
Not tested - just reasoned about.

---
Code review.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- Refactor
  - Simplified peer discovery: syncing no longer uses transport-protocol filtering, streamlining peer selection.
- Bug Fixes
  - Discovery requests adjusted to exclude client-only peers, reducing irrelevant results and improving relevance of discovered nodes.

### DIFF
--- a/comms/core/src/peer_manager/manager.rs
+++ b/comms/core/src/peer_manager/manager.rs
@@ -177,12 +177,8 @@ impl PeerManager {
         features: Option<PeerFeatures>,
         external_addresses_only: bool,
     ) -> Result<Vec<Peer>, PeerManagerError> {
-        self.peer_storage_sql.discovery_syncing(
-            n,
-            excluded_peers,
-            features,
-            external_addresses_only,
-        )
+        self.peer_storage_sql
+            .discovery_syncing(n, excluded_peers, features, external_addresses_only)
     }
 
     /// Adds or updates a peer and sets the last connection as successful.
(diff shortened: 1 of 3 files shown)


MCozhusheck (approved):

SWvheerden (commented):
I have a question why this is needed?
So discovery_syncing is used to provide peers to remote peers asking for peers. 
So we "want" to send all peers to the remote peer. 
But if we are talking to the remote peer, it means we are talking the some transport protocols. 

For example why do we need send peers who can only talk tor, to a node that only talks tcp?

hansieodendaal:
> I have a question why this is needed? So discovery_syncing is used to provide peers to remote peers asking for peers. So we 

We do not want to limit the network _footprint/map_ shared with any base node. That same base node that now only talkes tcp may be restarted using tor. A tor/tcp base node can talk to a tcp only node, and would definately want all the peer info, for example, we may decide to have a tcp only seed, then it should not be limiting during seed strap.

SWvheerden (approved):

PR #7500, opened 2025-09-16
fix: always cancel transactions

SWvheerden:
Description
---
Cancel tx and outputs. Don't assume they exist or not. Sending 1-sided offline transactions doesn't have an existing transaction to cancel 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of canceling pending transactions: missing or already-processed pending records and unlock failures are now logged as warnings and do not abort cancellation.
  * Reduces spurious errors during cancellation while preserving normal cleanup, signals, and event emission for more stable wallet behavior.

### DIFF
--- a/base_layer/wallet/src/transaction_service/service.rs
+++ b/base_layer/wallet/src/transaction_service/service.rs
@@ -3169,12 +3169,12 @@ where
             };
         };
 
-        self.db.cancel_pending_transaction(tx_id).inspect_err(|e| {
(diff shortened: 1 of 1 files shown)


hansieodendaal (commented):
I agree with coderabbit. The statement `self.resources.output_manager_service.cancel_transaction(tx_id).await?;` just below should get the same treatment and not return an error if the outputs are not found, and it could also do with the addition of an `inspect_err`.

PR #7501, opened 2025-09-16
feat: allow key manager to work without storage

SWvheerden:
Description
---
This pr removes storage requirement from the key manager, allowing it to be stateless. 

Motivation and Context
---
This removes the requirement of keeping some synced state in the keymanager.

Breaking Changes
---
This adds two new key types in the key manager. This means that a wallet using the new version is not compatible in using a previous version when doing offline signing, as both wallets need to be upgraded. 

## Summary by CodeRabbit

- **New Features**
  - DH-based key identifiers for commitment masks and encrypted data (TariKeyId variants) and per-key encrypted-key creation API.
  - Output recovery uses the output’s sender-offset public key and now returns an optional recovery result.
  - Offline signing supports multiple versions; 2.0.0 is latest.

- **Refactor**
  - Key manager exposes legacy and non-legacy constructors and consistently accepts a wallet type.
  - Key import API accepts an optional encryption/context parameter; call sites and tests updated.


base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -564,11 +649,17 @@ where TBackend: TransactionKeyManagerBackend + 'static
...
-    pub async fn import_key(&self, private_key: PrivateKey) -> Result<TariKeyId, KeyManagerServiceError> {
-        let public_key = CompressedPublicKey::from_secret_key(&private_key);
-        self.db.insert_imported_key(public_key.clone(), private_key).await?;
-        let key_id = TariKeyId::Imported { key: public_key };
-        Ok(key_id)
+    pub async fn import_key(
+        &self,
+        private_key: PrivateKey,
+        encryption_key: Option<TariKeyId>,
+    ) -> Result<TariKeyId, KeyManagerServiceError> {
```
hansieodendaal, line 655: This wrapper around `async fn created_encrypted_key(` can be flattened - we do not need it as it makes the code harder to read and hides what is being done.

 `async fn created_encrypted_key(`  can change to a public method and accept `encryption_key: Option<TariKeyId>,` instead of `encryption_key: TariKeyId,`.

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -503,9 +529,68 @@ where TBackend: TransactionKeyManagerBackend + 'static
...
+                let private_key = Box::pin(self.decrypt_encrypted_key(encrypted, key)).await?;
+                Ok(private_key)
+            },
         }
     }
 
+    async fn created_encrypted_key(
+        &self,
+        private_key: PrivateKey,
+        encryption_key: TariKeyId,
```
hansieodendaal, line 564:
```suggestion
        encryption_key: Option<TariKeyId>,
```

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -503,9 +529,68 @@ where TBackend: TransactionKeyManagerBackend + 'static
...
+    ) -> Result<TariKeyId, KeyManagerServiceError> {
+        let pvt_bytes = private_key.to_vec();
```
hansieodendaal, line 566:
```suggestion
        let encryption_key = match encryption_key {
            Some(key) => key,
            None => self.get_view_key().await?.key_id,
        };

                let pvt_bytes = private_key.to_vec();
```

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -503,9 +529,68 @@ where TBackend: TransactionKeyManagerBackend + 'static
...
+            TariKeyId::Encrypted { encrypted, key } => {
+                let key = TariKeyId::from_str(key.to_string().as_str())
+                    .map_err(|_| KeyManagerServiceError::KeySerializationError)?;
+                let private_key = Box::pin(self.decrypt_encrypted_key(encrypted, key)).await?;
+                Ok(private_key)
+            },
         }
     }
 
+    async fn created_encrypted_key(
```
hansieodendaal, line 561: See comments regarding `pub async fn import_key(`
```suggestion
    pub async fn created_encrypted_key(
```

base_layer/transaction_components/src/key_manager/interface.rs
```diff
@@ -96,6 +99,18 @@ pub enum TariKeyId {
...
+        private_key: SerializedKeyString,
+    },
+    DHEncryptedData {
+        public_key: CompressedPublicKey,
+        private_key: SerializedKeyString,
+    },
+    Encrypted {
+        encrypted: Vec<u8>,
+        key: SerializedKeyString,
+    },
```
hansieodendaal, line 144: We would need unit tests to cover these additional entries for `fn from_str(id: &str) -> Result<Self, Self::Err> {` and full round‑trip serialization.

base_layer/transaction_components/src/key_manager/memory_key_manager.rs
```diff
@@ -61,81 +60,49 @@ pub async fn create_memory_key_manager_from_seed(
...
     }
 }
 
 #[async_trait::async_trait]
 impl TransactionKeyManagerBackend for MemoryKeyManagerBackend {
-    async fn get_key_manager(&self, branch: &str) -> Result<Option<KeyManagerState>, KeyManagerStorageError> {
-        let key_manager_states = self.key_manger_states.read().await;
-        Ok(key_manager_states.get(branch).cloned())
+    async fn get_key_manager(&self, _branch: &str) -> Result<Option<KeyManagerState>, KeyManagerStorageError> {
+        unimplemented!()
```
hansieodendaal: This `impl TransactionKeyManagerBackend for MemoryKeyManagerBackend {` can be removed?

hansieodendaal (changes requested):
Hi there, looking good, some comments:
- All unit tests must pass.
- Will not using the key manager with legacy storage not make it significantly slower?

Some more comments below.

stringhandler:
Please add why this pr is a breaking change. It currently says "Breaking changes: None"

base_layer/transaction_components/src/key_manager/inner.rs
SWvheerden: I disagree, prefer smaller functions, the compiler will inline them anyway. Smaller functions makes it easier to graps what happens, it does not hide way stuff.

base_layer/transaction_components/src/key_manager/memory_key_manager.rs
SWvheerden: Not really, its going to complicte the code a lot if we want to keep the same piece of code to run the current key manager, still support currently imported keys, while removing this. If we stop using the console wallet, we can drop this

SWvheerden:
Yeah nothing has been filled in as the PR was not done yet, but filled in all now, also explaining why its technically a breaking change

PR #7508, opened 2025-09-18
fix: ledger birthday

SWvheerden:
Description
---
Fixes the birthday for ledger wallets. 
Ledger wallets dont have birthdays so we need to force it to 0, so that it will corectly pick up all transactions and not today

## Summary by CodeRabbit

* **Bug Fixes**
  * Ledger wallets now scan from the genesis block, ensuring complete transaction history and accurate UTXO detection.
  * Resolves issues where missing or undefined wallet birthdays could lead to missed outputs.
  * Initial sync for Ledger wallets may take longer but is more reliable and predictable.
  * Users may see more complete balances after syncing; no action required.


hansieodendaal (approved):
utACK

PR #7509, opened 2025-09-21
chorce: concise peermetadata implementation

saa938:
Description
 Use the simple serialize convenience function rather than computing sizes and serializing into a preallocated buffer.

Motivation and Context
make it less error prone and clearer

How Has This Been Tested?
Everything still works on my computer.

Make sure everything works fine.

Breaking Changes
[x]

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (commented):
using a pre computed sized vec is faster as you don't have to resize

PR #7512, opened 2025-09-22
feat!: add burnt commitments index

hansieodendaal:
Description
---
- Added a burnt commitments index table to the blockchain database to ensure no duplicate burns can be made.
- Added a blocking migration task to rebuild the burnt commitments index table.
- Fixed a bug in `TransactionBuilder` where not all output types were evaluated for burnt commitments.
- Added a method to retrieve burnt commitments info from the blockchain database for all blocks or an optional inclusive range.
- Wired in the retrieve burnt commitments info method to the command line.
- Added burnt commitment validation methods.
- Added an effective height at which burn commitments are allowed to optimize the migration.
- Changed mainnet consensus to disallow burnt commitments from height 106_000 - no outputs havr been burned up to date onmainnet.

**Note:** This will cause a hard fork on mainnet, as we switch off burn commitments for the time being.

Motivation and Context
---
Burn commitments must be unique across the blockchain.

How Has This Been Tested?
---
- Added unit tests:
  - `it_fails_if_multiple_burnt_outputs_with_single_burn_kernel`
  - `it_fetches_all_burnt_commitments_info`
- System-level testing:
  - Tested burn transactions while the burn commitments index was being rebuild in the background.
  - Successfully upgraded databases (background migration task) for versions v4.0.0, v4.7.0 and v5.0.0.

---
- Code review
- Repeat some system-level testing

Breaking Changes
---

- [] None
- [ ] Requires hard fork
- [X] Other - Please specify

BREAKING CHANGE: Changed mainnet consensus to disallow burnt commitments from height 106_000 - this will cuase a hard fork.

## Summary by CodeRabbit

- New Features
  - Added CLI command to list burnt commitments with height range filtering and optional CSV export.
  - Node API now returns burnt commitment information.

- Performance/Infrastructure
  - Introduced a burn-commitments index with automatic rebuild/migration for faster queries.
  - Reduced connection pool log noise; logs only on slow acquisitions.

- Validation
  - Enforced duplicate/invalid burn commitment checks and activation-height rules.
  - Clearer errors for mismatched burn transaction settings.

- Reliability
  - Improved gRPC error handling in miner and merge mining proxy for more robust error propagation.


base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -784,7 +791,7 @@ impl LMDBDatabase {
         Ok(())
     }
 
-    fn all_dbs(&self) -> [(&'static str, &DatabaseRef); 35] {
+    fn all_dbs(&self) -> [(&'static str, &DatabaseRef); 36] {
```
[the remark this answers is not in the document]
hansieodendaal: There are 36 db's

applications/minotari_node/src/commands/command/get_burnt_commitments.rs
```diff
@@ -0,0 +1,187 @@
...
+            path.join(file_name)
+        } else if fs::create_dir_all(&path).is_ok() {
+            path.join(file_name)
+        } else {
+            PathBuf::from(file_name)
+        }
+    } else {
+        PathBuf::from(file_name)
+    };
+    let _unused = fs::remove_file(&file_path);
```
SWvheerden, line 166: I don't think we should auto delete, we should ask/ error the command

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
```diff
@@ -931,6 +939,66 @@ impl LMDBDatabase {
...
+                    txn,
+                    header_hash,
+                    burn_commitment,
+                    last_rebuild_height,
+                )? {
+                    return Err(ChainStorageError::KeyExists {
+                        table_name: "kernel_burn_commitment_index",
+                        key: burn_commitment.to_hex(),
+                    });
+                }
```
SWvheerden, line 948: why do this?
why not just do `lmdb_get(..)`?

base_layer/transaction_components/src/transaction_builder/builder.rs
```diff
@@ -707,29 +721,24 @@ where KM: TransactionKeyManagerInterface
...
             core_tx_builder.add_input(input.output.to_transaction_input(&self.key_manager).await?.clone());
         }
-        for output in &self.custom_outputs {
+        for output in &self.custom_outputs.clone() {
+            self.verify_burn(&output.output)?;
             core_tx_builder.add_output(output.output.to_transaction_output()?);
         }
         let mut sent_outputs = Vec::new();
-        for recipient in &self.recipient_outputs {
+        for recipient in self.recipient_outputs.clone() {
```
SWvheerden: why clone here?
Cloning this is expensive

base_layer/transaction_components/src/transaction_builder/builder.rs
```diff
@@ -707,29 +721,24 @@ where KM: TransactionKeyManagerInterface
...
+            self.verify_burn(&recipient.output.output)?;
```
SWvheerden: you should call this when adding outputs, not here

base_layer/transaction_components/src/validation/aggregate_body/aggregate_body_internal_validator.rs
```diff
@@ -641,6 +641,45 @@ mod test {
...
+
+        // Create an aggregate body with both burnt outputs and one burn kernel
+        let body = AggregateBody::new(Vec::new(), vec![output1, output2], vec![kernel]);
+
+        // The check_total_burned function should fail
+        println!("Validate: {:?}", check_total_burned(&body));
+        assert!(check_total_burned(&body)
+            .unwrap_err()
+            .to_string()
+            .contains("Burned output has no matching burned kernel"));
```
SWvheerden: match the actual error here,
and then I think this is misleading, as I think the erorr should not return this perhaps if the order is changed, the error will be different?
This should be something along the line of only one burned output per tx allowed

SWvheerden (changes requested):
I am missing a validation check for unique kernels? It should be a validation check not a db write failure

base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
hansieodendaal: At this stage, it does not exist in the blockchain yet, as the kernel burn commitment indexes have not been created fully yet - we have to ensure that a current kernel will not create a future duplicate burn, as older kernels that contain burn commitments have precedence. 

Let's say the kernel_1111 contains burn_ommitment_102 and now kernel_20345 contains the same burn_ommitment_102 and wants to be added to the database, then it should be stopped.

If the index for burn_ommitment_102 has not been created yet, we have to search all kernels to verify the new kernel does not contain a double. If it has been created, we can just look it up.

SWvheerden: then we need to rewrite the db to do a full rebuild of this index on startup, we cannot do it async. 
What you are doing here is even worse, you rebuild the index on insertion of every kernel, without saving the results of the index rebuilding.

hansieodendaal:
This PR is not for now, maybe required at a later stage when Ootle is ready.

PR #7525, opened 2025-09-29
chore: update soft fork height

SWvheerden:
Description
---
updates fork height to be Wednesday afternoon 22 Oct

## Summary by CodeRabbit

* **Chores**
  * Adjusted Nextnet consensus activation height for con_6 from 140,000 to 121,000. The associated network changes will now take effect at block height 121,000. No other runtime behavior or user-facing functionality is altered, and no public APIs are affected. Users should be aware of the updated activation timing if monitoring block-height–based events.


hansieodendaal (approved):
utACK

PR #7529, opened 2025-10-01
feat: remove tcp public address requirement

### LINKED ISSUE #7526: Node should not have to provide public address to run tcp only
We should allow nodes to run tcp only without requiring nodes to provide non-dialable addresses.
Atm a node wont accept a connection from another tcp node unless you provide them an address. 
To get around this we need to set a junk non-dialable address, and then set the config of the node to not dial local addresses. 

If a node tells you it does not have a dialable public address, you should accept that rather then forcing the node to provide you with a fake one.

hansieodendaal:
Description
---
Removed the requirement that a tcp node must provide a public address.

Minimum config required for a TCP node:
```
base_node.p2p.transport.type = "tcp"
```
Note: `base_node.p2p.public_addresses` is default empty.

Closes #7526.

Motivation and Context
---
See #7526.

How Has This Been Tested?
---
Added a unit test in the peer db to verify backend operation.
Performed system-level testing on esmeralda between local TCP nodes - one with a public address and two without. Later, the node with the public address was connected to the TCP seed peers and all nodes synced to and followed the tip.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **Bug Fixes**
  * Peers without public addresses are now accepted (no validation error) and preserved in listings.
  * Empty addresses are skipped when adding/updating peer records; peers can exist with zero addresses.
  * Empty addresses are treated as non-external.
  * Added debug logging for empty-address cases to aid diagnostics.

* **Tests**
  * Added tests verifying handling, storage, and retrieval of peers with no addresses.

### DIFF
--- a/comms/core/src/peer_manager/storage/database.rs
+++ b/comms/core/src/peer_manager/storage/database.rs
@@ -3132,7 +3132,7 @@ mod tests {
 
         // Assert that the addresses list contains a single empty entry
         assert_eq!(peer_from_db.addresses.addresses().len(), 1);
-        // Assert that all other va;ues are empty or zero
+        // Assert that all other values are empty or zero
         assert_eq!(peer_from_db.addresses.addresses()[0].address(), &Multiaddr::empty());
         assert_eq!(peer_from_db.addresses.addresses()[0].last_seen(), None,);
         assert_eq!(peer_from_db.addresses.addresses()[0].connection_attempts(), 0);
--- a/comms/core/src/peer_manager/storage/database.rs
+++ b/comms/core/src/peer_manager/storage/database.rs
@@ -3188,9 +3188,7 @@ mod tests {
 
         // Verify 'peers::table' has an entry but 'multi_addresses::table' not
         let mut conn = peers_db.connection.get_pooled_connection().unwrap();
-        let peer_query: Vec<NewPeerSql> = peers::table
-            .load::<NewPeerSql>(&mut conn)
-            .unwrap();
+        let peer_query: Vec<NewPeerSql> = peers::table.load::<NewPeerSql>(&mut conn).unwrap();
         assert_eq!(peer_query.len(), 1);
         let addresses_query: Vec<NewMultiaddrWithStatsSql> = multi_addresses::table
             .load::<NewMultiaddrWithStatsSql>(&mut conn)
--- a/comms/core/src/peer_manager/storage/database.rs
+++ b/comms/core/src/peer_manager/storage/database.rs
@@ -625,7 +625,7 @@ impl PeerDatabaseSql {
                 last_attempted: None,
                 last_failed_reason: None,
                 quality_score: None,
-                source: "".to_string(),
+                source: "null".to_string(),
             });
         }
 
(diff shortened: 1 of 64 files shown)


hansieodendaal:
This approach of having an empty `multi_addresses` table entry for peers with no shared public address does not work. 

The table is important for normal operations. We will have to introduce a lot of code to make it work without, and it will completely break the patterns we implemented. Having the table entry there without a public address only means the peer will not be selected to dial, similar to a peer that shared a local TCP address as its public address, which has a `multi_addresses` table entry.

The no-address peer can dial and establish a connection to a remote peer; it can send pings to the remote peer, but the remote peer does not respond with pongs.

hansieodendaal:
The current solution works fine in system-level tests - when a `Peer` is popolated from the database it gets an empty `MultiaddrWithStats` if no corresponding entry in the `multi_addresses` table can be found for the peer.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1074,24 +1098,27 @@ impl PeerDatabaseSql {
         let mut results;
         if start_bytes.len() > NodeId::byte_size() {
             results = peers::table
-                .inner_join(multi_addresses::table.on(multi_addresses::peer_id.eq(peers::peer_id)))
+                .left_outer_join(multi_addresses::table.on(multi_addresses::peer_id.eq(peers::peer_id)))
```
SWvheerden, line 1101: these should be left_join
Not outer join. 
outer join returns all matches in both tables.

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1074,24 +1098,27 @@ impl PeerDatabaseSql {
...
+                .select((peers::all_columns, multi_addresses::all_columns.nullable()))
+                .load::<(NewPeerSql, Option<NewMultiaddrWithStatsSql>)>(&mut conn)?;
 
             let mut public_key_match = peers::table
-                .inner_join(multi_addresses::table.on(multi_addresses::peer_id.eq(peers::peer_id)))
+                .left_outer_join(multi_addresses::table.on(multi_addresses::peer_id.eq(peers::peer_id)))
                 .filter(peers::public_key.like(format!("{partial_key}%")))
-                .load::<(NewPeerSql, NewMultiaddrWithStatsSql)>(&mut conn)?;
+                .select((peers::all_columns, multi_addresses::all_columns.nullable()))
+                .load::<(NewPeerSql, Option<NewMultiaddrWithStatsSql>)>(&mut conn)?;
```
SWvheerden, line 1116: why write two querys?
Why not write one with and `or`

SWvheerden (approved):
One comment, I know this is probably out of scope of this pr, and I will merge it as is, but the queries could be combined

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: Yes it could probably be done

PR #7530, opened 2025-10-01
chore: remove RwLock from TransactionKeyManagerWrapper

martinserts:
Description
---

This change removes the `RwLock` from the `TransactionKeyManagerWrapper`.

Motivation and Context
---

The key manager is a central component for many parts of the system. Previously, its internal implementation used an `RwLock` on the entire object. Consequently, every operation had to wait for the lock to be released before the next one could begin (if one of them was a write operation).

Following recent changes, the database is no longer part of the key manager. This modification makes it possible to clone the key manager and use a separate copy in each thread.

This change also makes it explicit which operations mutate the key manager. This leads to multiple changes in the codebase to specify mutability, especially in tests.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Refactor**
  - Key management now uses mutable access across mining, wallet, node, and transaction flows, streamlining internal state handling.

- **Performance**
  - Reduced internal locking around key management, lowering contention and improving responsiveness for mining and signing.

- **Tests**
  - Unit, integration, and helper tests updated to the new key-manager access patterns; no user-facing behavior changes.

### DIFF
--- a/base_layer/transaction_components/src/key_manager/inner.rs
+++ b/base_layer/transaction_components/src/key_manager/inner.rs
@@ -110,9 +110,10 @@ use crate::{
     MicroMinotari,
 };
 
+#[derive(Clone)]
 pub struct TransactionKeyManagerInner<TBackend> {
-    key_managers: HashMap<String, RwLock<TariKeyManager<KeyDigest>>>,
-    db: Option<TBackend>,
+    key_managers: HashMap<String, TariKeyManager<KeyDigest>>,
+    db: Option<Arc<RwLock<TBackend>>>,
     master_seed: CipherSeed,
     crypto_factories: CryptoFactories,
     wallet_type: Arc<WalletType>,
@@ -129,7 +130,7 @@ where TBackend: TransactionKeyManagerBackend + 'static
     ) -> Result<Self, KeyManagerServiceError> {
         let mut km = TransactionKeyManagerInner {
             key_managers: HashMap::new(),
-            db,
+            db: db.map(|db| Arc::new(RwLock::new(db))),
             master_seed,
             crypto_factories,
             wallet_type,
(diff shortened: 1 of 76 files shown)


base_layer/transaction_components/src/validation/helpers.rs
```diff
@@ -263,21 +263,21 @@ mod test {
         #[tokio::test]
         async fn it_succeeds_for_valid_coinbase() {
             let height = 1;
-            let key_manager = create_memory_key_manager().await.unwrap();
-            let test_params = TestParams::new(&key_manager).await;
+            let mut key_manager = create_memory_key_manager().await.unwrap();
+            let test_params = TestParams::new(&mut key_manager).await;
             let rules = test_helpers::create_consensus_manager();
-            let key_manager = create_memory_key_manager().await.unwrap();
+            let mut key_manager = create_memory_key_manager().await.unwrap();
```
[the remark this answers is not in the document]
martinserts: Removed - [4ea2e3a](https://github.com/tari-project/tari/pull/7530/commits/4ea2e3a714eaa9a4d025f4d62009ed39ca2efb00)

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -110,9 +110,10 @@ use crate::{
     MicroMinotari,
 };
 
+#[derive(Clone)]
 pub struct TransactionKeyManagerInner<TBackend> {
-    key_managers: HashMap<String, RwLock<TariKeyManager<KeyDigest>>>,
-    db: Option<TBackend>,
+    key_managers: HashMap<String, TariKeyManager<KeyDigest>>,
+    db: Option<Arc<RwLock<TBackend>>>,
```
hansieodendaal, line 116: Since TBackend is an r2d2 SQLite WAL mode DB connection, it is already thread-safe and designed for concurrent access via its connection pool. We can safely remove the Arc<RwLock<...>> wrapper and use Option<TBackend> directly.
```suggestion
    db: Option<TBackend>,
```

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -129,7 +130,7 @@ where TBackend: TransactionKeyManagerBackend + 'static
     ) -> Result<Self, KeyManagerServiceError> {
         let mut km = TransactionKeyManagerInner {
             key_managers: HashMap::new(),
-            db,
+            db: db.map(|db| Arc::new(RwLock::new(db))),
```
hansieodendaal, line 133:
```suggestion
            db,
```

base_layer/transaction_components/src/key_manager/inner.rs
```diff
@@ -426,7 +419,7 @@ where TBackend: TransactionKeyManagerBackend + 'static
             TariKeyId::Zero => Ok(PrivateKey::default()),
             TariKeyId::Imported { key } => {
                 if let Some(db) = &self.db {
-                    let pvt_key = db.get_imported_key(key).await?;
+                    let pvt_key = db.write().await.get_imported_key(key).await?;
```
hansieodendaal, line 422:
```suggestion
                    let pvt_key = db.get_imported_key(key).await?;

hansieodendaal (changes requested):
LGTM, just the one comment to address please.

base_layer/transaction_components/src/key_manager/inner.rs
SWvheerden: TBackend is a trait, not an object, it can be anything

SWvheerden (approved):

Issue #7506, opened 2025-09-18
Payment_id is not preserved for outbound transactions with wallet recovery


hansieodendaal:
## With open wallets

**wallet 1** 

<img width="625" height="139" alt="Image" src="https://github.com/user-attachments/assets/7d179da1-b288-4eff-b18d-32227ca5c4e9" />

**wallet 2**

<img width="622" height="141" alt="Image" src="https://github.com/user-attachments/assets/cd6fe64d-0e8c-4c78-a8fe-94be3984269a" />

## Recovered **wallet 2**

<img width="558" height="136" alt="Image" src="https://github.com/user-attachments/assets/195e96ba-e17e-46b9-8624-8648688255e8" />

## Note

Data displayed ditrectly from the database with
```
SELECT
    CASE
        WHEN tx_id < 0 THEN printf('%llu', CAST((tx_id + 18446744073709551616) AS UNSIGNED))
        ELSE tx_id
    END AS tx_id_u64,
    amount,
    direction,
    user_payment_id,
    timestamp
FROM
    completed_transactions
WHERE
    timestamp LIKE '%2025-09-18%'
ORDER BY
    timestamp DESC
```

PR #7532, opened 2025-10-02
fix: payment_id is not preserved for outbound transactions with wallet recovery

martinserts:
Description
---

Fixes an issue  #7506, that recovered wallet does not have `payment_id`.

Motivation and Context
---

Outbound transactions in recovered wallet are missing `payment_id`.
When a payment is sent out, its change contains `payment_id` of type **TransactionInfo**, that contains recipient address, amount, fee and ORIGINAL enclosing `payment_id`.

The problem is, that the enclosed `payment_id` of the change was missing.
This PR fixes it.

How Has This Been Tested?
---

Sent out a payment from console wallet, and then recovered it. The outgoing payment was recovered and its payment ID was there.

Screenshot 2025-10-02 at 17 29 32

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* New Features
  * Ensures payment memos (payment IDs) are attached to transactions across wallet and offline-signing flows.
* Bug Fixes
  * Fixes missing/late memo attachment so memos are reliably included in one-sided, multisig deposit, and multisig withdrawal transactions.
* Improvements
  * Improves consistency and traceability by carrying memos throughout the transaction-building process, leading to clearer recipient/payment identification in finalized transactions.

### DIFF
--- a/base_layer/transaction_components/src/offline_signing/offline_signer.rs
+++ b/base_layer/transaction_components/src/offline_signing/offline_signer.rs
@@ -70,7 +70,7 @@ where TKeyManagerInterface: TransactionKeyManagerInterface
         payment_id: MemoField,
         sender_address: TariAddress,
     ) -> Result<PrepareOneSidedTransactionForSigningResult, TransactionBuilderError> {
-        tx_builder.with_memo(payment_id.clone());
+        tx_builder.with_memo(payment_id.clone())
         // we do this to ensure the fee is calculated correctly
         tx_builder
             .add_stealth_recipient(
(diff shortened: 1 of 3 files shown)


hansieodendaal (changes requested):
Looks good.

I am missing the addition in:
- `pub async fn create_deposit_multisig_transaction(`
- `pub async fn spend_multisig_utxo(`
- `async fn build_transaction(`
- `pub async fn send_sha_atomic_swap_transaction(`
- `async fn scrape_wallet(`
- `pub async fn send_many_one_sided_transactions(`
- `pub async fn burn_tari(`

I am doing a recovery test as well, will revert.

hansieodendaal:
The recovery test for the normal transactions I sent worked perfectly well.

SWvheerden (approved):

PR #7535, opened 2025-10-04
fix: multi_address_with_stats merging to only retain the newest valid claims in the peer_db

### LINKED ISSUE #7491: MultiaddressesWithStats merging is wrong
Currently the addresses are just merged together, and new addresses are added. We should be more clever with this, and only allow the newest signed advertised addresses. 
We should compare the date and only use new valid addresses. Currently we accumulate old invalid addresses from peers.

hansieodendaal:
Description
---
The peer_db now only retains the latest claims, and peer multi-address merges cleans out old data.

Closes #7491.

Motivation and Context
---
See #7491.

How Has This Been Tested?
---
Added new unit tests:
- MultiAddressWithStats
  - `fn it_merges_addresses_with_stats_correctly()`
  - `fn it_merges_when_config_overlaps()`
  - `fn it_prefers_advertised_addresses_over_config_on_tie()`
- PeerManager
  - `async fn it_correctly_merges_old_and_new_address_claims()`

System-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- New Features
  - Claim-aware peer address merging that prefers newer signed claims, preserves configured addresses, enforces capacity limits, and deterministically breaks ties in favor of advertised addresses.
  - Batch-style address updates with automatic pruning of stale addresses to keep peers’ address lists current.

- Bug Fixes
  - Eliminates inconsistent ordering, duplicates, and outdated addresses after updates; preserves and propagates usage/latency stats correctly.

- Documentation
  - Clarified merge, replacement, and tie-breaking behavior.

- Tests
  - Added tests for claim timestamps, merge scenarios, tie-breaking, config overlap, and database consistency.

### DIFF
--- a/comms/core/src/net_address/mutliaddresses_with_stats.rs
+++ b/comms/core/src/net_address/mutliaddresses_with_stats.rs
@@ -85,44 +85,16 @@ impl MultiaddressesWithStats {
             .and_then(|a| a.last_attempted())
     }
 
-    /// Adds a new net address to the peer. This function will not add a duplicate if the address
-    /// already exists.
-    pub fn add_address(&mut self, net_address: &Multiaddr, source: &PeerAddressSource) {
-        if let Some(addr_mut) = self.addresses.iter_mut().find(|x| x.address() == net_address) {
-            addr_mut.update_source_if_better(source);
-        } else {
-            self.addresses
-                .push(MultiaddrWithStats::new(net_address.clone(), source.clone()));
-        }
-
-        // Ensure that the addresses are sorted by quality
-        self.sort_addresses();
-    }
-
     pub fn contains(&self, net_address: &Multiaddr) -> bool {
         self.addresses.iter().any(|x| x.address() == net_address)
     }
 
-    /// Compares the existing set of addresses to the provided address set and remove missing addresses and
-    /// add new addresses without discarding the usage stats of the existing and remaining addresses.
-    pub fn update_addresses(&mut self, addresses: &[Multiaddr], source: &PeerAddressSource) {
-        for address in addresses {
-            if let Some(addr) = self.addresses.iter_mut().find(|a| a.address() == address) {
-                addr.update_source_if_better(source);
-            }
-        }
-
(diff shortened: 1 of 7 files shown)


comms/core/src/net_address/mutliaddresses_with_stats.rs
```diff
@@ -129,8 +109,65 @@ impl MultiaddressesWithStats {
...
-        for addr in &other.addresses {
+    /// Merges another set of addresses into this set, preserving usage stats where possible.
+    ///
+    /// Merge logic:
+    /// - If the incoming set has a strictly newer signed claim, all non-config addresses are replaced by the incoming
+    ///   set.
+    /// - If the incoming set is older, only usage stats for known addresses are updated.
+    /// - If neither set has a signed claim or timestamps are equal, addresses are merged individually.
+    ///
+    /// After merging, addresses are sorted by quality and truncated to the maximum allowed.
```
SWvheerden, line 122: 2 questions here, quick one whould we ever get non-signed addresses? And is this not a security issue?

And then the other one, why are we filtering on the config also here?
Zooming out, I get its a manually inserted address, but if the node does not even advertise that addresses a connectable address, why keep it in there if its not dialable?

hansieodendaal: I re-organized the last branch a bit and added better explanatory comments. 

A non-signed address source can only come from config; all other sources contain a claim with a signature that is validated and accepted or rejected. This function caters for all permutations and combinations:
- Incoming is newer -> replace but do not throw away addresses from config
- Incoming is older  -> only update stats for known addresses
- Incoming has the same time stamp -> address-by-address merge
- Either one or both is from config -> address-by-address merge

We keep the config addresses around until we do the actual merge, then the best source is chosen - an identity claim will trump a config source.

SWvheerden (approved):

PR #7536, opened 2025-10-05
fix: proactive dialler not selecting new peers to dial

### LINKED ISSUE #7510: The base node should try to dial all peers, not only a select few
get_available_dial_candidates always returns the same data, which I think is incorrect

it should exclude peers that we have tried and seen that have failed

you can confirm this in the proactive_dialer.rs

```rust
        debug!(
            target: LOG_TARGET,
            "({task_id}) Success rate: {success_rate:.2}, will dial {dial_count} peers for {needed} needed connections",

        );

        // Select healthy peers for dialing
        let mut candidates = self
            .select_healthy_dial_candidates(pool, connection_stats, excluded_peers, dial_count, task_id)
            .await?;
        if candidates.is_empty() {
            candidates = self
                .select_healthy_dial_candidates(pool, connection_stats, &[], dial_count, task_id)
                .await?;
        }
```
if you search for "Success Rate", you will see it picking the same candidates over and over again

to combat this, you can sort by last_attempted

using nulls first

but open to other suggestions

but this leads to us never breaking out of the few that we try

hansieodendaal:
Description
---
Fixed the proactive dialler in that it never selected new peers to dial, improved the initial peer selection logic and increased the MAX_CONCURRENT_DIALS to enable covering the peer database quicker to find connections.

Fixed a bug in `PeerDatabaseSql::get_available_dial_candidates`, whereby it sometimes returned less than the requested number of peers.

Fixes #7510.

Motivation and Context
---
The proactive dialler did not select new peers to dial if it could not find any connections; it always selected the same peers.
To test, re-start a base node, but delete its peer_db so it has to create it from scratch, then when it connects to the seed peers, wait till the seed strap completes, then ban all the seed peers. Monitor active connections in the base node console and all "proactive" entries in the network log file. 

How Has This Been Tested?
---
System-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- Performance Improvements
  - Increased parallel proactive-dial cap to 30 for faster connection attempts.

- Bug Fixes
  - Dialing now respects "exclude failed" and returns early with a clear warning when no candidates.
  - Candidate selection refined to avoid repeated failed dials and improve health-aware ranking.

- Enhancements
  - Candidate pool can be augmented via optional randomized selection when insufficient.

- Tests
  - Unit tests updated to cover new limits, exclusion, and randomization behaviors.

- Chores
  - Updated warning and log messages to reflect the new selection flow.

### DIFF
--- a/comms/core/src/connectivity/proactive_dialer.rs
+++ b/comms/core/src/connectivity/proactive_dialer.rs
@@ -35,7 +35,7 @@ use crate::{
 };
 
 const LOG_TARGET: &str = "comms::connectivity::proactive_dialer";
-const MAX_CONCURRENT_DIALS: usize = 20;
+const MAX_CONCURRENT_DIALS: usize = 30;
 
 /// Proactive peer dialing logic for maintaining target connection counts
 pub struct ProactiveDialer {
--- a/comms/core/src/connectivity/proactive_dialer.rs
+++ b/comms/core/src/connectivity/proactive_dialer.rs
@@ -74,24 +74,29 @@ impl ProactiveDialer {
             return Ok(0);
         }
 
-        let current_connections = pool.count_connected_nodes();
+        let eligible_connections = pool
+            .get_connected_node_ids()
+            .iter()
+            .filter(|p| !excluded_peers.contains(p))
+            .count();
         let target = self.config.target_connection_count;
 
         // Update metrics
 
-        if current_connections >= target {
+        if eligible_connections >= target {
             debug!(
                 target: LOG_TARGET,
-                "({task_id}) Current connections ({current_connections}) meet or exceed target ({target}), no proactive dialing needed",
+                "({task_id}) Eligible connections ({eligible_connections}) meet or exceed target ({target}), no \
+                proactive dialing needed",
             );
 
             return Ok(0);
         }
 
-        let needed = target.saturating_sub(current_connections);
+        let needed = target.saturating_sub(eligible_connections);
         debug!(
             target: LOG_TARGET,
-            "({task_id}) Proactive dialing: need {needed} more connections ({current_connections}/{target})",
+            "({task_id}) Proactive dialing: need {needed} more connections ({eligible_connections}/{target})",
 
         );
 
--- a/comms/core/src/connectivity/proactive_dialer.rs
+++ b/comms/core/src/connectivity/proactive_dialer.rs
@@ -193,18 +193,19 @@ impl ProactiveDialer {
         let mut managed_and_excluded = managed.clone();
         managed_and_excluded.append(&mut excluded_peers.to_vec());
 
-        // Get available dial candidates using SQL-based filtering
+        // Get available dial candidates
         let mut candidates = self
             .peer_manager
-            .get_available_dial_candidates(&managed_and_excluded, Some(count * 3)) // Get 3x more for health scoring
+            .get_available_dial_candidates(&managed_and_excluded, Some(count * 3), true, true) // Get 3x more for health scoring
             .await?;
         if candidates.len() < count * 3 {
             // Only exclude managed and current selected nodes to get more candidates (thus include 'excluded_peers')
             let mut to_be_excluded = candidates.iter().map(|p| p.node_id.clone()).collect::<Vec<_>>();
             to_be_excluded.append(&mut managed);
+            // Now also allow selection from previously failed peers
             let mut random = self
                 .peer_manager
-                .random_peers(count * 3 - candidates.len(), &to_be_excluded, None)
+                .get_available_dial_candidates(&to_be_excluded, Some(count * 3 - candidates.len()), false, true)
                 .await?;
             candidates.append(&mut random);
         }
--- a/comms/core/src/connectivity/proactive_dialer.rs
+++ b/comms/core/src/connectivity/proactive_dialer.rs
@@ -74,29 +74,24 @@ impl ProactiveDialer {
             return Ok(0);
         }
 
-        let eligible_connections = pool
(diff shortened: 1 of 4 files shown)


comms/core/src/connectivity/proactive_dialer.rs
```diff
@@ -74,24 +74,29 @@ impl ProactiveDialer {
...
+            .get_connected_node_ids()
+            .iter()
+            .filter(|p| !excluded_peers.contains(p))
+            .count();
         let target = self.config.target_connection_count;
 
         // Update metrics
 
-        if current_connections >= target {
+        if eligible_connections >= target {
```
SWvheerden: this is wrong. 
Should stay as it was, 
we need x connections, if some are seeds, why might have a different reason why we are connected to them, but thats fine and when we disconnect we will find another connection. We dont wants X + seeds connections as the connections

comms/core/src/peer_manager/storage/database.rs
```diff
@@ -1192,6 +1165,7 @@ impl PeerDatabaseSql {
                     .or(peers::banned_until.lt(chrono::Utc::now().naive_utc())),
             )
             .filter(peers::deleted_at.is_null())
+            .filter(multi_addresses::last_failed_reason.is_null())
```
SWvheerden: this is bad as well. We cant just filter on null. If a peer failed, and we dont have anyother addresses, we need to go back end try ones which are null.

SWvheerden (changes requested):
This will still cause issues, we need to return failed nodes as well, becasue if for some reason there is a network failure locally on your node, all dials will fail, they will be mark them as failed, and never try them again, here. 

We need to prioritise non failed ones yes, but if we dont find enough, we need failed ones as wel.

comms/core/src/connectivity/proactive_dialer.rs
hansieodendaal: This is a simple math issue. The default connections are 8, and when we have six seeds, nodes easily fixate on all or most of the seeds and only add two more nodes (as verified with system-level testing). For a new node seeds are always contacted first, and then they will be selected for pro-active dialing. The purpose of pro-active dialing is to discover more peers that could be dialed, not less.

comms/core/src/peer_manager/storage/database.rs
hansieodendaal: Yeah, I considered something like this. I will add code to select random failed nodes to try as well when we select nodes, and when the list has been exhausted and all addresses have failed, we can just try them at random.

From practical testing, using this PR where we advance through the list and ignore failed ones, ~530 peers were tried out of ~16,000 peers discovered via peer exchange.

comms/core/src/connectivity/proactive_dialer.rs
SWvheerden: Pro-active dialing is not to discover peers, we have peer discovery for that. 
Pro-active dialing is if we are not connected to enough peers, then we should add more peers. 

Take this edge case, 4 seed peers have decided that your node is their connected node, so they are inbound connections to your node. And the seed peers should make connections and be connected to other nodes. Now you node ignores you 8 connections, and make 12 rather? 

I dont get how the nodes fixate on the seed peers? We cull their connections if we made them. And if we select them via pro-active dialing, it means there is not enough other nodes on the network?

comms/core/src/connectivity/proactive_dialer.rs
```diff
@@ -195,13 +190,24 @@ impl ProactiveDialer {
...
+        if candidates.len() < count * 3 {
+            // Only exclude managed and current selected nodes to get more candidates (thus include 'excluded_peers')
+            let mut to_be_excluded = candidates.iter().map(|p| p.node_id.clone()).collect::<Vec<_>>();
+            to_be_excluded.append(&mut managed);
+            let mut random = self
+                .peer_manager
+                .random_peers(count * 3 - candidates.len(), &to_be_excluded, None)
+                .await?;
+            candidates.append(&mut random);
+        }
```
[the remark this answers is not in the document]
hansieodendaal: The aim is to ignore the 'excluded_peers' list if we cannot find enough cancidates, as per the comment:
```
// Only exclude managed and current selected nodes to get more candidates (thus include 'excluded_peers')
```

comms/core/src/peer_manager/storage/database.rs
SWvheerden: Please add this in, this will be an issue with smaller networks like esme and igor

hansieodendaal: Changes made in the latest commit:
- We first try to select non-failed (3 x count as per the selection algo), honouring the exclusion list passed in.
- If we cannot select the target amount of non-failed peers, we ignore the exclusion list and choose from the remaining peers at random.
After the initial selection, the algo will choose the best `count` number of peers from those selected for proactive dialling.

SWvheerden: we cant use random, it excludes the peers
`.filter(multi_addresses::last_seen.is_not_null())`
Just set a bypass for this filter when we need to exlcude it:
`.filter(multi_addresses::last_failed_reason.is_null())`

hansieodendaal: Notes: 
- I implemented randomness and bypass failed in `PeerDatabaseSql::get_available_dial_candidates`.
- There was also a bug in `PeerDatabaseSql::get_available_dial_candidates`, whereby it did not return the selected number of peers.

PR #7541, opened 2025-10-09
feat: add deterministic transaction id

### LINKED ISSUE #7485: calculate tx_id deterministically
Currently, all tx_id's are random. 
We need to calculate them deterministically. 
End goal for this should be if a wallet rescans the database, it needs to recalculate the same tx_id. 
Sender and receiver does not need to be the same. 

And idea could be to hash the change for the sender, and output for the receiver. 
We could fall back to random if needed,

hansieodendaal:
Description
---
Added deterministic transaction IDs, which are an 8-byte (u64) hash based on the transaction output hash in question and the wallet view key. 
- Any scanned or recovered wallet output will have the same transaction ID across view or spend wallets. 
- Sender wallets will be able to calculate the transaction ID for receiver wallets if they need to, for that specific output.
- Sender wallets will use their change output as the determining output hash for the transaction; this will result in the same transaction ID being allocated upon wallet recovery. In the case of no change output, the hash of the first ordered output will be used for the transaction ID.
- For coin split transactions, the hash of the first ordered output will be used for the transaction ID.

Fixed the issue with the Windows test build target link:
```
: error LNK2019: unresolved external symbol __imp_InitializeSecurityDescriptor referenced in function mdb_env_setup_locks
: error LNK2019: unresolved external symbol __imp_SetSecurityDescriptorDacl referenced in function mdb_env_setup_lock
```

Fixes #7485.

Motivation and Context
---
See #7485.

How Has This Been Tested?
---
Added unit tests.
Performed system-level testing.

---
Code review.
System-level testing.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **New Features**
  * Deterministic transaction ID generation for consistent, reproducible TX identifiers
  * Ability to replace/update transaction IDs across wallet outputs

* **Enhancements**
  * Transaction creation and finalization flows now return and propagate transaction IDs directly
  * Scanning and recovery simplified to use deterministic per-output IDs and reduced payloads
  * Storage and confirmation paths support optional TXID replacement for atomic updates


.cargo/config.toml
```diff
@@ -4,6 +4,9 @@ linker = "aarch64-linux-gnu-gcc"
 [target.riscv64gc-unknown-linux-gnu]
 linker = "riscv64-linux-gnu-gcc"
 
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-ladvapi32"]
```
SWvheerden, line 8: why is this suddently needed?
Its not metntioned in the pr notes, or anything?
It also does not have anything to do with the tx id?

base_layer/common_types/src/tx_id.rs
```diff
@@ -24,19 +24,62 @@ use std::{
...
+        );
+        let mut mac = mac_result
+            .expect("MAC should be initialized - call TxId::init_mac(&view_key)")
+            .clone();
+        mac.update(out_hash32);
+        let out = mac.finalize_fixed();
+        let mut buffer = [0u8; MAC_SIZE];
+        buffer.copy_from_slice(&out);
+        TxId(u64::from_le_bytes(buffer))
+    }
```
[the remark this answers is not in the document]
SWvheerden: yeah this is bad...
Its going to limit the runtime to a single wallet and single wallet only. Its very bad leaky abstruction.

base_layer/common_types/src/tx_id.rs
```diff
@@ -24,19 +24,62 @@ use std::{
...
+    pub fn init_mac(view_key: &[u8]) {
+        let _unused = TX_ID_MAC.set(Self::blake2b_tx_id_mac(view_key, TAG_TX_ID));
+        debug_assert_eq!(Blake2bTxIdMac::output_size(), MAC_SIZE);
+    }
+
+    fn blake2b_tx_id_mac(key: &[u8], domain: &str) -> Blake2bTxIdMac {
+        Blake2bTxIdMac::new_with_salt_and_personal(
+            key,               // Unique key
+            b"",               // Salt is not used
+            domain.as_bytes(), // Domain separation string
```
SWvheerden: why not use a domain hash, inside the domain crate?
its going to give us the same thing

base_layer/transaction_components/src/multisig/session.rs
```diff
@@ -173,12 +176,16 @@ where TKeyManagerInterface: TransactionKeyManagerInterface
 
         let finalized_builder = tx_builder.build().await?;
 
-        let (change_hashes, change) = match finalized_builder.change {
+        let (change_hashes, change, tx_id) = match finalized_builder.change {
             Some(change_output) => {
                 let hash = change_output.output_hash();
-                (vec![hash], Some(vec![change_output]))
+                let tx_id = TxId::new_deterministic(&hash);
```
SWvheerden: why not do all this inside the builder, than duplicate this in all functions?

base_layer/transaction_components/src/transaction_components/wallet_output.rs
```diff
@@ -426,6 +429,11 @@ impl WalletOutput {
         Ok(input)
     }
 
+    /// Get the deterministic TxId for this output
+    pub fn to_tx_id(&self) -> TxId {
```
SWvheerden:
```suggestion
    pub fn calculate_tx_id(&self) -> TxId {
```
to implies conversion

base_layer/transaction_components/src/format_currency.rs
```diff
@@ -24,7 +24,7 @@
 /// # Examples
 ///
 /// ```rust
-/// use tari_core::transactions::format_currency;
+/// use transaction_components::format_currency;
```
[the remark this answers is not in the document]
SWvheerden: ditto

base_layer/transaction_components/src/tari_amount.rs
```diff
@@ -81,8 +81,8 @@ impl From<DecimalConvertError> for MicroMinotariError {
     }
 }
 /// A convenience constant that makes it easier to define Tari amounts.
-/// ```edition2018
-/// use tari_core::transactions::tari_amount::{uT, MicroMinotari, T};
+/// ```edition2021
+/// use transaction_components::tari_amount::{uT, MicroMinotari, T};
 /// assert_eq!(MicroMinotari::from(42), 42 * uT);
 /// assert_eq!(1 * T, 1_000_000.into());
```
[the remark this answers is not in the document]
SWvheerden: ditto

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
```diff
@@ -130,48 +115,17 @@ where
...
+        let mut recovered_outputs: Vec<RecoveredOutput> = Vec::new();
+        for (output, has_known_script, hash) in &mut rewound_outputs {
             let db_output = DbWalletOutput::from_wallet_output(
                 output.clone(),
                 None,
                 Self::output_source(output, *has_known_script),
                 None,
                 None,
             );
-            let tx_id = match tx_id {
```
SWvheerden, line 142: this is here for TU and offline signing, just make sure offline signing still works correctly in terms of the tx_ids

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -471,11 +476,13 @@ where
                         self.resources.one_sided_tari_address.clone(),
                     )
                     .await?;
+
+                self.replace_tx_id_in_outputs(temp_tx_id, res.tx_id).await?;
```
SWvheerden: no....
why create a temporaty one, only to replace this. 
See my previous comment. If the txbuilder can create this or creates this, we dont need this stupid work around..

SWvheerden (changes requested):

base_layer/common_types/src/tx_id.rs
```diff
@@ -34,7 +34,8 @@ use serde::{Deserialize, Serialize};
 type Blake2bTxIdMac = Blake2bMac<U8>; // 8-byte keyed BLAKE2b
 const MAC_SIZE: usize = 8;
 const TAG_TX_ID: &str = "tari/tx_id_64";
-static TX_ID_MAC: OnceLock<Blake2bTxIdMac> = OnceLock::new();
+
+static TX_ID_MAC: RwLock<Option<Blake2bTxIdMac>> = RwLock::new(None);
```
SWvheerden: this is just going to cause issues with multiple wallets, you need to know you need to reint this, just change this be some domain hash with the view key + output hash

.cargo/config.toml
hansieodendaal: The PR notes says: _Fixed the issue with the Windows test build target link:_

I could not run the release mode tests in Windows, and have not tried that for a while. Something somewhere in the lmdb dependencies omitted the rustflags.

base_layer/wallet/src/transaction_service/service.rs
hansieodendaal: The outputs are encumbered to be sent in the outputs db with a temp tx_id as handle to retrieve them later, then the tx builder builds and finalizes the transaction, then the tx_id it is replaced in the outputs database to the actual tx_id.

base_layer/common_types/src/tx_id.rs
hansieodendaal: I changed it, but it's a lesser implementation as we create a 32-byte hash, then use the first 8 bytes for the u64.
The trait bounds of `Mac::<Blake2b<OutSize>>::generate` do not allow an 8-byte output.

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
hansieodendaal: Re-confirmation still outstanding

.cargo/config.toml
SWvheerden: please make it a seperate pr

base_layer/wallet/src/output_manager_service/recovery/standard_outputs_recoverer.rs
hansieodendaal: Confirmed - all transactions created for imported outputs are assigned the deterministic transaction id, irrespective of the code path used.

.cargo/config.toml
hansieodendaal: See #7559

base_layer/wallet/src/output_manager_service/storage/database/backend.rs
```diff
@@ -82,6 +82,9 @@ pub trait OutputManagerBackend: Send + Sync + Clone {
     fn cancel_pending_transaction(&self, tx_id: TxId) -> Result<(), OutputManagerStorageError>;
     /// This method will update an output's metadata signature, akin to 'finalize output'
     fn update_output_metadata_signature(&self, output: &TransactionOutput) -> Result<(), OutputManagerStorageError>;
+    /// Replace the TxId for all outputs that match the old_tx_id with the new_tx_id in `spent_in_tx_id` or
+    /// `received_in_tx_id` fields.
+    fn replace_tx_id_in_outputs(&self, tx_id_old: TxId, tx_id_new: TxId) -> Result<(), OutputManagerStorageError>;
```
SWvheerden: can this not be removed now?

hansieodendaal: No, unfortunately, not without a major wallet refactoring, which is happening with wallet 2.0

The outputs are encumbered to be sent in the outputs db with a temp tx_id as handle to retrieve them later (output manager), then the tx builder builds and finalizes the transaction (transaction manager), then the temp tx_id is replaced in the outputs database with the deterministic tx_id.

SWvheerden: You can make a single call, to change it over to the outputs from shorterm encumbered, to encumbered and update the tx_id
Currently it needs to two channel calls, two db calls. This can all be a single update

base_layer/common_types/src/tx_id.rs
```diff
@@ -26,17 +26,32 @@ use std::{
...
 impl TxId {
+    /// Create a new random TxId. Only for temporary use.
     pub fn new_random() -> Self {
+        use rand::{rngs::OsRng, RngCore};
         TxId(OsRng.next_u64())
     }
 
+    /// Create a new TxId deterministically from the given 32-byte output hash and MAC key.
+    pub fn new_deterministic(mac_key: &[u8], output_hash: &[u8; 32]) -> Self {
+        let hash = Mac::<Blake2b<U32>>::generate(mac_key, output_hash, "tari/tx_id_64");
```
stringhandler, line 46: This is different from all the other hashing we do in the repo. Is it not better to have it the same?

stringhandler: I see that it is a succinct way of using tari crypto's hashing though. I suppose it is ok. You are not actually generating a MAC though.

base_layer/transaction_components/src/lib.rs
```diff
@@ -64,3 +67,12 @@ impl BanReason {
         self.ban_duration
     }
 }
+
+// Helper function to derive a TxId from the first ransaction output, or a random TxId if there are no outputs
+pub(crate) fn tx_outputs_to_tx_id(mac_key: &[u8], outputs: &[TransactionOutput]) -> TxId {
+    if let Some(first_output) = outputs.first() {
+        TxId::new_deterministic(mac_key, &first_output.hash())
+    } else {
+        TxId::new_random()
```
stringhandler, line 76: You should create each output's hash, sort and hash them together.

base_layer/transaction_components/src/transaction_builder/builder.rs
```diff
@@ -508,7 +514,29 @@ where KM: TransactionKeyManagerInterface
...
+                            .first()
+                            .map(|c| c.output.calculate_tx_id(view_key.clone().as_bytes()))
+                    })
+                    .unwrap_or_else(TxId::new_random),
+            )
+        } else {
+            None
+        };
+
+        Ok((fee, change, tx_id))
```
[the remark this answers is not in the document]
stringhandler: What is the reason for allowing a new tx_id here?

base_layer/transaction_components/src/transaction_components/wallet_output.rs
```diff
@@ -426,6 +429,11 @@ impl WalletOutput {
         Ok(input)
     }
 
+    /// Calculate the deterministic TxId for this output
+    pub fn calculate_tx_id(&self, mac_key: &[u8]) -> TxId {
```
stringhandler: mac_key is not a very helpful name here.

base_layer/wallet/src/lib.rs
```diff
@@ -58,3 +60,12 @@ pub type WalletSqlite = Wallet<
 >;
 
 pub type WalletKeyManager = TransactionKeyManagerWrapper<TransactionKeyManagerSqliteDatabase<WalletDbConnection>>;
+
+// Helper function to derive a TxId from the first ransaction output, or a random TxId if there are no outputs
+pub(crate) fn tx_outputs_to_tx_id(mac_key: &[u8], outputs: &[TransactionOutput]) -> TxId {
+    if let Some(first_output) = outputs.first() {
+        TxId::new_deterministic(mac_key, &first_output.hash())
+    } else {
+        TxId::new_random()
```
stringhandler: duplicate code

base_layer/wallet/src/output_manager_service/handle.rs
```diff
@@ -1038,6 +1051,23 @@ where KM: TransactionKeyManagerInterface
         }
     }
 
+    pub async fn replace_tx_id_in_outputs(
```
stringhandler: This seems like a bad idea. Exchanges and other users might have external references to this tx_id. I know it's not reliable, but we can't silently change these ids. 

A suggestion for as better approach:
add a version to TX_id or add a version in the database for tx_id. V1 (or v0) tx_ids stay the same, V2 tx_ids use the new method

stringhandler (commented):
This is a very dangerous PR in it's current state. I don't like the idea of updating existing tx_ids, because they are stored and used on exchanges (even though they shouldn't)

I have read the description and issue, but I don't understand the need for the view key? It doesn't seem to have any relevance. An output hash is unique enough. If anything, you should include the inputs in the hash, because the only way to produce a duplicate output, but different data is to have different inputs.

It was not clear to me how recovered tx_ids would differ from created tx_ids. Please let me know where I can look

base_layer/common_types/src/tx_id.rs
hansieodendaal: Yes, that is true. The initial implementation used `Blake2bMac` as Tari crypto does not strictly implement a MAC, which was similar to this example, but the review request was to change it to use Tari crypto.
```rust
type Blake2bTxIdMac = Blake2bMac<U8>;
const MAC_SIZE: usize = 8;
const TAG_TX_ID: &str = "tari/tx_id_64";

    /// Create a new TxId deterministically from the given 32-byte output hash.
    pub fn new_deterministic(mac_key: &[u8], out_hash32: &[u8; 32]) -> Self {
        let mut mac = Blake2bTxIdMac::new_with_salt_and_personal(
            mac_key,               // Unique key
            b"",               // Salt is not used
            TAG_TX_ID.as_bytes(), // Domain separation string
        )
            .expect("key length ok");
        mac.update(out_hash32);
        let out = mac.finalize_fixed();
        let mut buffer = [0u8; MAC_SIZE];
        buffer.copy_from_slice(&out);
        TxId(u64::from_le_bytes(buffer))
    }
```

base_layer/wallet/src/output_manager_service/handle.rs
hansieodendaal: This is only used when creating new transactions - no change to existing transactions.

We have to _encumber_outputs_to_be_sent_ using a temporary tx_id as handle, then we can finalise the transaction, after which we will know if we have a change output or not, which is primarily used for the deterministic tx_id. Then, when encumberance is confirmed, we have to change to temporary tx_id with the deterministic tx_id.

hansieodendaal:
> What is the reason for allowing a new tx_id here? (https://github.com/tari-project/tari/pull/7541#discussion_r2498033536)

The call path `pub async fn get_pre_build_change_output(` -> `async fn add_change_if_required(` from the offline signer does not have access to finalize transaction, hence the need to calculate the transaction id at this level.

hansieodendaal:
> I have read the description and issue, but I don't understand the need for the view key? 

The view key is used to make the tx_id unique to a wallet, just so the same scanned, recovered or imported output will result in a different tx_id in a different wallet.

hansieodendaal:
> It was not clear to me how recovered tx_ids would differ from created tx_ids. Please let me know where I can look

This call path `async fn scan_for_outputs(` -> [ `pub async fn scan_outputs_for_one_sided_payments(`  or  `pub async fn scan_outputs_for_multisig(` or `pub async fn scan_for_recoverable_outputs(`] -> `fn import_onesided_outputs(` never assigned the optional tx_id and used to do 
```rust
let tx_id = tx_id.unwrap_or(TxId::new_random());
```
now it is
```rust
let tx_id = output.calculate_tx_id(view_key.as_bytes());
```
each scanned/recovered output is assigned its own deterministic tx_id.

All newly created transactions would primarily use the change output for the deterministic tx_id, or otherwise, with no change, the first ordered output.

base_layer/transaction_components/src/lib.rs
hansieodendaal: With no change, the transaction is either perfectly matched (inputs + fees = outputs) or a transaction to self. If self is not involved, there would be nothing to recover with wallet recovery, but if it was a transaction to self (coin-join, coin-split or 1-1), the tx_id assigned to the first recovered output of that transaction would always correspond to the original tx_id. If we change the scheme, a tx_id for transaction to self would not correlate to any original tx_id.

base_layer/transaction_components/src/transaction_builder/builder.rs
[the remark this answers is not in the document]
hansieodendaal: The call path `pub async fn get_pre_build_change_output(` -> `async fn add_change_if_required(` from the offline signer does not have access to finalize transaction, hence the need to calculate the transaction id at this level. 

Custom outputs are generally used as outputs to self, and in that case it would be better to choose one of those for the deterministic tx_id to aid in tx_id correlation with wallet recovery.

base_layer/transaction_components/src/lib.rs
SWvheerden: The order when scanning or the order here, is not the same, but also there is no order guarantee when scanning

hansieodendaal: Agreed,
 _the tx_id assigned to the first recovered output of that transaction would always correspond to the original tx_id_ 
 should read 
 _when doing recovery and hitting the first ordered output of the original transaction, the tx_id assigned to that recovered output would always correspond to the original tx_id_

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1664,6 +1832,16 @@ where
...
+                let tx_id = tx_outputs_to_tx_id(
+                    self.resources
+                        .transaction_key_manager_service
+                        .get_view_key()
+                        .await?
+                        .pub_key
+                        .as_bytes(),
+                    transaction.body.outputs(),
+                );
+                self.replace_tx_id_in_outputs(temp_tx_id, tx_id).await?;
```
SWvheerden: why not use confirm encumberance  inside the oms

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -1725,6 +1903,16 @@ where
...
```
SWvheerden: same here

base_layer/wallet/src/output_manager_service/storage/database/backend.rs
SWvheerden: Please remove this, its not needed

SWvheerden (changes requested):

PR #7546, opened 2025-10-14
fix: change how pagination works on utxo scanning

SWvheerden:
Description
---
Properly handle the special edge case, where the request has reached the page limit, but we are also not done with the block. We also dont want to split up the block over two requests. So we need to ensure that we remove the partial block we added so that it can be requested fully in the next request. We also dont want to get in a loop where the block cannot fit into the page limit, so if the block is the same as the first one, we just send it as it, partial. If net we remove it and let it be sent in the next request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed edge cases in paginated UTXO queries when a page ends mid-block to prevent partial or duplicated results.
  * Ensures correct continuation across block boundaries and early exit when the first block is also the last that can be sent.
  * Removes trailing blocks or partial entries so subsequent pages request the correct data.
  * Improves error reporting for rare pagination inconsistencies.
  * Preserves behavior when limits aren’t reached or the chain end is encountered.

### DIFF
--- a/base_layer/core/src/base_node/rpc/query_service.rs
+++ b/base_layer/core/src/base_node/rpc/query_service.rs
@@ -284,11 +284,11 @@ impl<B: BlockchainBackend + 'static> Service<B> {
             }
             if fetched_utxos >= request.limit {
                 next_header_to_request = current_header.hash().to_vec();
-                // this is a special edge case, our request has reached the page limit, but we are also not done with
+                // This is a special edge case, our request has reached the page limit, but we are also not done with
                 // the block. We also dont want to split up the block over two requests. So we need to ensure that we
                 // remove the partial block we added so that it can be requested fully in the next request. We also dont
                 // want to get in a loop where the block cannot fit into the page limit, so if the block is the same as
-                // the first one, we just send it as it, partial. If net we remove it and let it be sent in the next
+                // the first one, we just send it as is, partial. If not we remove it and let it be sent in the next
                 // request.
                 if utxos.first().ok_or(Error::General(anyhow::anyhow!("No utxos founds")))? // should never happen as we always add at least one block
                     .header_hash ==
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -284,6 +284,28 @@ impl<B: BlockchainBackend + 'static> Service<B> {
             }
             if fetched_utxos >= request.limit {
                 next_header_to_request = current_header.hash().to_vec();
+                // this is a special edge case, our request has reached the page limit, but we are also not done with
```
hansieodendaal:
```suggestion
                //This is a special edge case, our request has reached the page limit, but we are also not done with
```

base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -284,6 +284,28 @@ impl<B: BlockchainBackend + 'static> Service<B> {
...
+                // the block. We also dont want to split up the block over two requests. So we need to ensure that we
+                // remove the partial block we added so that it can be requested fully in the next request. We also dont
+                // want to get in a loop where the block cannot fit into the page limit, so if the block is the same as
+                // the first one, we just send it as it, partial. If net we remove it and let it be sent in the next
```
hansieodendaal:
```suggestion
                // the first one, we just send it as is, partial. If not we remove it and let it be sent in the next
```

base_layer/core/src/base_node/rpc/query_service.rs
hansieodendaal: What happens if we have more than 20,000 inputs in a block, or many consecutive blocks with 10,557 inputs?

It appears that if we have more than 20,000 inputs in a block, the current sync logic will terminate prematurely and fail to deliver all UTXOs to the wallet. The wallet limit is currently set to 10 pages, and the chunk size here is 2,000 UTXOs per page. Once the limit is reached, in the next round the wallet will request outputs from the next block, leaving the remainder of UTXOs from the large block unsynced.

With the current mainnet block weight of 90,000 we may not be able to reach 20,000 inputs but we can come close. 10,557 inputs are certainly possible.

hansieodendaal (commented):
It certainly addresses the edge case, but I have some more questions please.

base_layer/core/src/base_node/rpc/query_service.rs
SWvheerden: 10_557 is just more than half of the limit, pr #7549 makes this even larger. So I dont see the issue here?

PR #7553, opened 2025-10-24
fix: broadcast and propagate

SWvheerden:
Description
---
Peers should now properly propagate/broadcast blocks and transactions out to the network 

## Summary by CodeRabbit

* **Refactor**
  * Peer selection logic updated so more broadcast and propagation paths can include already-connected peers even when address-based exclusions exist.
  * This change may improve message delivery reliability and reach by allowing connected peers to be chosen in scenarios that were previously blocked.

### DIFF
--- a/comms/dht/src/actor.rs
+++ b/comms/dht/src/actor.rs
@@ -567,31 +567,31 @@ impl DhtActor {
         trace!(target: LOG_TARGET, "Select peers broadcast strategy: {broadcast_strategy}" );
         #[allow(clippy::enum_glob_use)]
         use BroadcastStrategy::*;
-        let peers = match broadcast_strategy {
+        let (peers, only_connected_nodes) = match broadcast_strategy {
             DirectNodeId(node_id) => {
                 // Send to a particular peer matching the given node ID
-                peer_manager
+                (peer_manager
                     .direct_identity_node_id(&node_id)
                     .await
                     .map(|peer| peer.map(|p| vec![p.node_id]).unwrap_or_default())
-                    .map_err(Into::<DhtActorError>::into)?
+                    .map_err(Into::<DhtActorError>::into)?, false)
             },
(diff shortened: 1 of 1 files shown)


martinserts (dismissed):

comms/dht/src/actor.rs
```diff
@@ -567,31 +567,31 @@ impl DhtActor {
         trace!(target: LOG_TARGET, "Select peers broadcast strategy: {broadcast_strategy}" );
         #[allow(clippy::enum_glob_use)]
         use BroadcastStrategy::*;
-        let peers = match broadcast_strategy {
+        let (peers, only_connected_nodes) = match broadcast_strategy {
```
hansieodendaal, line 570: The actual logic is to bypass excluded address filtering after nodes have been selected, not only connected nodes.
```suggestion
        let (peers, bypass_excluded) = match broadcast_strategy {
```

hansieodendaal (changes requested):
This PR breaks the contract in the DHT config and is not adequately described. The default exclusion list is empty so that it will have no effect for almost all base nodes, but for users who specifically want to exclude certain addresses, the expected behaviour will no longer be guaranteed.
```rust
    /// Addresses that should never be dialed (default value = []). This can be a specific address or an IPv4/TCP
    /// range. Example: When used in conjunction with `allow_test_addresses = true` (but it could be any other
    /// range)   `excluded_dial_addresses = ["/ip4/127.*.0:49.*/tcp/*", "/ip4/127.*.101:255.*/tcp/*"]`
    ///                or
    ///   `excluded_dial_addresses = ["/ip4/127.0:0.1/tcp/122", "/ip4/127.0:0.1/tcp/1000:2000"]`
    pub excluded_dial_addresses: MultiaddrRangeList,
```
This PR should not be merged.

hansieodendaal (dismissed):
Let's test it

PR #7557, opened 2025-10-28
feat: add monitored peers

SWvheerden:
Description
---
Add an option to keep a list of monitored peers so that it always keeps a connection to those peers

## Summary by CodeRabbit

* **New Features**
  * Added "monitored_peers" configuration option enabling node operators to designate specific peers that maintain persistent connections and are protected from automatic disconnection policies.

### DIFF
--- a/applications/minotari_node/src/bootstrap.rs
+++ b/applications/minotari_node/src/bootstrap.rs
@@ -30,7 +30,7 @@ use tari_common::{
 };
 use tari_comms::{
     multiaddr::{Error as MultiaddrError, Multiaddr},
-    peer_manager::Peer,
+    peer_manager::{NodeId, Peer},
     protocol::rpc::RpcServer,
     tor::TorIdentity,
     NodeIdentity,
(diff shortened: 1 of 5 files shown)


common/config/presets/c_base_node_c.toml
```diff
@@ -29,6 +29,9 @@
 # couple of nodes that you always want to have in sync. If set this node will only sync to the nodes in this set.
 # force_sync_peers = ["public_key1::address1", "public_key2::address2",... ]
 
+# This allowlist provides a method to keep a connection open to the spesified nodes in this list.
```
hansieodendaal, line 32:
```suggestion
# This allow-list provides a method to keep a connection open to the specified nodes in this list.
```

applications/minotari_node/src/bootstrap.rs
```diff
@@ -106,17 +106,29 @@ where B: BlockchainBackend + 'static
...
+            .map(|s| SeedPeer::from_str(s))
+            .map(|r| r.map(Peer::from))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| ExitError::new(ExitCode::ConfigError, e))?;
+        let mut monitored_node_ids: Vec<NodeId> = monitored_peers.clone().into_iter().map(|p| p.node_id).collect();
+        let mut total_monitored_ids = force_sync_node_ids.clone();
+        total_monitored_ids.append(&mut monitored_node_ids);
 
-        debug!(target: LOG_TARGET, "{} sync peer(s) configured", sync_peers.len());
+        base_node_config.state_machine.blockchain_sync_config.forced_sync_peers = force_sync_node_ids.clone();
```
hansieodendaal, line 129: The base node config should not be changed here - force sync peers have a very specific meaning and are very different to peers you want to monitor. If force sync peers are configured, those will be the **only** peers used for sync - not additional to other peers.  When monitored peers are configured and allowed to influence the force sync peers config, which are usually empty, the config contract will be changed.
```suggestion
```

applications/minotari_node/src/bootstrap.rs
```diff
@@ -106,17 +106,29 @@ where B: BlockchainBackend + 'static
...
+
+        debug!(target: LOG_TARGET, "{} sync peer(s) configured", force_sync_node_ids.len());
```
hansieodendaal, line 131:
```suggestion
        debug!(target: LOG_TARGET, "{} force sync peer(s) and {} monitored peers configured", force_sync_peers.len(), monitored_peers.len());
```

hansieodendaal (changes requested):
The idea is sound, just one critical error to be fixed, please.

applications/minotari_node/src/bootstrap.rs
SWvheerden: umm this is just a rename?

hansieodendaal: Ok yes, my bad, all good here.

hansieodendaal (approved):
utACK

PR #7563, opened 2025-11-03
fix: validation deadlock

SWvheerden:
Description
---
Fixes a potential validation deadlock where the wallet never finishes the validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet stability by capping consecutive validation resets. On certain synchronization events, the wallet now allows an initial reset but prevents immediate repeated resets, reducing excessive validation cycles and improving reliability during base node height changes and scanner completions.


hansieodendaal (approved):
utACK

_Looks good after the sign change -:)_

PR #7590, opened 2025-11-19
fix: validation trigger

SWvheerden:
Description
---
fixes the validation trigger to only trigger once per, and not run multiple validation checks at the same time. 

## Summary by CodeRabbit

* **Refactor**
  * Simplified transaction service architecture by removing unused wallet storage dependencies.
  * Updated transaction validation workflow to use a protocol-based approach for improved consistency with event handling.
  * Streamlined service initialization by removing redundant type parameters.


hansieodendaal (approved):
utACK

PR #7593, opened 2025-11-20
feat: harden wallet scan resiliency

### LINKED ISSUE #7574: Add checks in wallet scanning to ensure it does not miss any scanned data
The wallet should ensure that it receives all blocks as part of scanning, If the server does not provide that block as part of the scanning, it should ask for that block. 

At the moment the wallet just processes all data received from teh server, without actually checking what it received.

hansieodendaal:
Description
---
This PR adds strict validation to the UTXO streaming scan. The wallet now detects and aborts scans on:
- missing block heights;
- backwards height sequences;
- conflicting headers during paged block responses.

On failure, the scan round is retried from the last valid scanned block to ensure no outputs can be missed.

Fixes #7574.

Motivation and Context
---
See #7574.

How Has This Been Tested?
---
System-level testing.

---
Code review.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened UTXO scan to enforce block-order integrity (detect non-monotonic, non-consecutive, or conflicting blocks).
  * Added sorting of received blocks to ensure consistent height-based processing.
  * Improved handling when the network tip differs from the requested end state, with clearer diagnostics.
  * Ensured persisted scan state reliably reflects the last saved block header to avoid stale progress.


base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -449,6 +456,36 @@ where
                 let response = response?;
                 #[allow(clippy::cast_possible_wrap)]
                 for response in response.blocks {
+                    if let Some(previous_block) = &prev_scanned_block {
+                        if response.height < previous_block.height {
```
SWvheerden, line 462: There is no guarantee that the data is sorted; it should be, but there is no guarantee of this, perhaps sort it before we start the batch,

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
```diff
@@ -449,6 +456,36 @@ where
...
+                        if response.height > previous_block.height + 1 {
+                            // Missing block(s) detected between previous_block and response height - fork block
+                            // re-validation forced.
+                            return Err(anyhow!(
+                                "Non-consecutive block heights received during UTXO scan: {} followed by {}",
+                                previous_block.height,
+                                response.height
+                            ));
+                        }
+                        if response.height == previous_block.height &&
```
SWvheerden, line 479: this is possible and happens. 
the base node will split the block into seperate blocks if the block is too big, more than 2000 inputs/outputs.

SWvheerden (changes requested):

base_layer/wallet/src/utxo_scanner_service/utxo_scanner_task.rs
hansieodendaal: Yes, we check here that if the height is the same, the hashes may not be different.

SWvheerden: aa okay, cool, yeah missed the `!`

SWvheerden (approved):

PR #7606, opened 2025-11-26
fix: scanning page count

SWvheerden:
Description
---
Fixes the counting of pages when streaming utxos to the base node. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pagination accuracy when fetching data with request limits
  * Fixed edge-case handling for data retrieval at pagination boundaries
  * Enhanced chunking logic to properly respect configured limits during requests

### DIFF
--- a/base_layer/core/src/base_node/rpc/query_service.rs
+++ b/base_layer/core/src/base_node/rpc/query_service.rs
@@ -213,7 +213,7 @@ impl<B: BlockchainBackend + 'static> Service<B> {
         // fetch utxos
         let mut utxos = vec![];
         let mut current_header = start_header;
-        let mut fetched_utxos = 0;
+        let mut fetched_chunks = 0;
         let spending_end_header_hash = self
             .db
             .fetch_header(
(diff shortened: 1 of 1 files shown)


base_layer/core/src/base_node/rpc/query_service.rs
```diff
@@ -298,16 +299,15 @@ impl<B: BlockchainBackend + 'static> Service<B> {
...
 
-            fetched_utxos += 1;
-
             if current_header.height >= tip_header.header().height {
                 next_header_to_request = vec![];
                 has_next_page = (end_height.saturating_sub(current_header.height)) > 0;
                 break;
             }
-            if fetched_utxos >= request.limit {
+            if fetched_chunks > request.limit {
```
hansieodendaal, line 310: This does not feel right either. Surely the `request.limit` must mean UTXO count, not chunk count or block count as per previous code?

The wallet requests a maximum of 2000 UTXOs, now we will send up to 2000 chunks of UTXOs.

SWvheerden: no its chunk count, always has been

PR #7611, opened 2025-11-27
fix: cache age for old blocks

SWvheerden:
Description
---
Fixes the cache of older blocks for the seed nodes

## Summary by CodeRabbit

* **Performance**
  * Increased CDN/cache durations for very large tip-delta ranges (now longer s-maxage values, including 1-day and 30-day tiers) to reduce load and improve response latency.
  * Adjusted cache boundary behavior for varied delta sizes.

* **Tests**
  * Updated tests to expect the new longer-cache durations for large deltas.

* **Chores**
  * No changes to public APIs or user-facing wallet behavior.


stringhandler (commented):
These seem very cautious. Why would we cache 5 day old blocks for 1 hour instead of 1 day?

hansieodendaal (dismissed):
utACK

PR #7612, opened 2025-11-27
fix: ledger tx flow

SWvheerden:
Description
---
Fixes ledger tx flow to correctly only ask verification once

## Summary by CodeRabbit

* **New Features**
  * Enhanced ledger device support with explicit ledger key branch selection.
  * New user-verified metadata signature flow for stronger ledger-backed signing.

* **Bug Fixes**
  * Refined transaction signing and nonce/offset generation to reduce edge-case mismatches across wallet types.

* **Refactor**
  * Unified send flow across wallet types (simpler, consistent interactive behavior).
  * Modernized key-management and metadata-signature interfaces for clearer ledger/software paths.

### DIFF
--- a/base_layer/transaction_components/src/key_manager/interface.rs
+++ b/base_layer/transaction_components/src/key_manager/interface.rs
@@ -20,6 +20,7 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use minotari_ledger_wallet_common::common_types::LedgerKeyBranch;
 use tari_common_types::{
     tari_address::TariAddress,
     types::{
@@ -62,7 +63,7 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
     fn get_random_key(
         &self,
         encryption_key: Option<TariKeyId>,
-        ledger_key: bool,
+        ledger_key: Option<LedgerKeyBranch>,
     ) -> Result<TariKeyAndId, KeyManagerError>;
 
     /// Gets the key id at the specified index
--- a/base_layer/transaction_components/src/key_manager/interface.rs
+++ b/base_layer/transaction_components/src/key_manager/interface.rs
@@ -199,7 +199,7 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
         range_proof_type: RangeProofType,
     ) -> Result<ComAndPubSignature, KeyManagerError>;
 
-    // creates a metadata signature for the output requiring manual user verification on a ledger device
+    // Creates a metadata signature for the output, requiring manual user verification if on a ledger device
     fn get_metadata_signature_user_verified(
         &self,
         commitment_mask_key_id: &TariKeyId,
--- a/base_layer/transaction_components/src/key_manager/interface.rs
+++ b/base_layer/transaction_components/src/key_manager/interface.rs
@@ -188,7 +188,7 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
         sender_offset_key_ids: &[TariKeyId],
     ) -> Result<PrivateKey, KeyManagerError>;
 
-    // creates a metadata signature for the output without requiring manual user verification on a ledger device
+    // Creates a metadata signature for the output, without requiring manual user verification if on a ledger device
     fn get_metadata_signature(
         &self,
         commitment_mask_key_id: &TariKeyId,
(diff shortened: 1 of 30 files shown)


base_layer/transaction_components/src/key_manager/interface.rs
```diff
@@ -187,8 +188,7 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
         sender_offset_key_ids: &[TariKeyId],
     ) -> Result<PrivateKey, KeyManagerError>;
 
-    // Look into perhaps removing all nonce here, if the signer and receiver are the same it should not be required to
-    // share or pre calc the nonces
+    // creates a metadata signature for the output without requiring manual user verification on a ledger device
```
hansieodendaal: Please capitalise all comments - we actually need a clippy rule enforcing sentence case for comments.

hansieodendaal:
```suggestion
    // Creates a metadata signature for the output, without requiring manual user verification if on a ledger device
```

base_layer/transaction_components/src/key_manager/interface.rs
```diff
@@ -199,7 +199,8 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
         range_proof_type: RangeProofType,
     ) -> Result<ComAndPubSignature, KeyManagerError>;
 
-    fn get_one_sided_metadata_signature(
+    // creates a metadata signature for the output requiring manual user verification on a ledger device
```
hansieodendaal:
```suggestion
    // Creates a metadata signature for the output, requiring manual user verification if on a ledger device
```

base_layer/transaction_components/src/key_manager/interface.rs
```diff
@@ -199,7 +199,8 @@ pub trait TransactionKeyManagerInterface: Clone + Send + Sync + 'static {
...
+    fn get_metadata_signature_user_verified(
```
hansieodendaal, line 203: Have we added `_user_verified` to all methods where user verification is required?

base_layer/transaction_components/src/offline_signing/one_sided_signer.rs
```diff
@@ -172,7 +173,7 @@ fn build_multisig_output<KM: TransactionKeyManagerInterface>(
     }
     let recipient = &info.recipients.first().ok_or(TransactionBuilderError::NoRecipients)?;
 
-    let sender_offset_key = key_manager.get_random_key(None, true)?;
+    let sender_offset_key = key_manager.get_random_key(None, Some(LedgerKeyBranch::OneSidedSenderOffset))?;
```
hansieodendaal: ditto ledger exclusive function - why?

base_layer/transaction_components/src/multisig/session.rs
```diff
@@ -106,7 +107,9 @@ where TKeyManagerInterface: TransactionKeyManagerInterface
             MemoField::new_address_and_data(recipient.clone(), fee_estimate, true, TxType::PaymentToOther, user_data)
                 .map_err(|e| TransactionError::BuilderError(format!("Failed to create MemoField: {}", e)))?;
 
-        let sender_offset_key = self.key_manager.get_random_key(None, true)?;
+        let sender_offset_key = self
+            .key_manager
+            .get_random_key(None, Some(LedgerKeyBranch::OneSidedSenderOffset))?;
```
hansieodendaal, line 112: It is not intuitive to know that `Some(LedgerKeyBranch::OneSidedSenderOffset)` will only have an effect if the wallet type is ledger without diving a couple of levels deeper into the code. The code will be clearer and more maintainable if we differentiate here, if ledger wallet pass `Some(LedgerKeyBranch::OneSidedSenderOffset)`, otherwise pass `None`.

Also applicable in all other places where we call any `Some(LedgerKeyBranch::XXX)`.

base_layer/transaction_components/src/transaction_components/wallet_output_builder.rs
```diff
@@ -258,7 +259,7 @@ impl WalletOutputBuilder {
         let aggregate_sender_offset_public_key = aggregated_sender_offset_public_key_shares.to_public_key()? +
             &sender_offset_public_key_self.to_public_key()?;
 
-        let ephemeral_pubkey_self = key_manager.get_random_key(None, true)?;
+        let ephemeral_pubkey_self = key_manager.get_random_key(None, Some(LedgerKeyBranch::MetadataEphemeralNonce))?;
```
hansieodendaal, line 262: Ditto differentiation - also all other similar code

hansieodendaal (commented):
Looks like an improvement. Some comments to address, please.

base_layer/transaction_components/src/multisig/session.rs
SWvheerden: I disagree and agree, I dont like having to pass in the ledger key here,  `Some(LedgerKeyBranch::OneSidedSenderOffset)`, but for the ledger to be backwords compaitible, we need to identitify the key.  

But I think having to add `if ledger` here, will be worse. Currently all ledger handling is limited to the key manager. And I think its better that way, the transaction builder should not care if you use a ledger or not, you only care about the ledger if you create a signature.

base_layer/transaction_components/src/key_manager/interface.rs
SWvheerden: yes

PR #7621, opened 2025-12-03
fix: grpc re-validation trigger

SWvheerden:
Description
---
Fixes the ability to trigger re-validation via grpc

## Summary by CodeRabbit

* **New Features**
  * Revalidation request now accepts transaction and output modes to control behavior.
  * Option to revalidate only previously rejected transactions or revalidate all.
  * Optional revalidation of outputs alongside transaction revalidation.

* **Configuration**
  * Wallet HTTP bind option renamed from local_ip to listen_ip (defaults/examples updated).

* **Documentation**
  * Config examples and docs updated to use listen_ip.

### DIFF
--- a/applications/minotari_app_grpc/proto/wallet.proto
+++ b/applications/minotari_app_grpc/proto/wallet.proto
@@ -2344,8 +2344,18 @@ message CancelTransactionResponse {
 }
 
 message RevalidateRequest{
-  bool only_rejected = 1;
-  bool revalidate_outputs = 2;
+  TransactionValidationMode transaction_mode =1;
+  OutputValidationMode output_mode = 2;
+}
+
+enum TransactionValidationMode {
+    RevalidateAll = 0;
+    RejectedOnly = 1;
+}
+
+enum OutputValidationMode {
+    Ignore = 0;
+    Revalidate = 1;
 }
 
 message RevalidateResponse{}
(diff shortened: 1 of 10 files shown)


martinserts (dismissed):

applications/minotari_app_grpc/proto/wallet.proto
```diff
@@ -2343,7 +2343,10 @@ message CancelTransactionResponse {
   string failure_message = 2;
 }
 
-message RevalidateRequest{}
+message RevalidateRequest{
```
stringhandler, line 2346: nit: this is a little unclear. Perhaps it would be better with an enum. Because `only_rejected` is independent from `revalidate_outputs` it appears that you can revalidate only rejected outputs, which it does not do from the code below. 

Perhaps it could be done as 

```rust
{
TransactionValidationMode transaction_mode =1;
OutputValidationMode output_mode = 2;
}

enum TransactionValidationMode {
  RevalidateAll,
  RejectedOnly
}

enum OutputValidationMode {
  Ignore,
  RevalidateAll
}
```

stringhandler (dismissed):
Included a nit for clarity, but otherwise ok

applications/minotari_app_grpc/proto/wallet.proto
SWvheerden: mmm yeah, was trying to keep backward compatibility, but the method did not work, so perhaps its work making this a breaking change and doing this

PR #7638, opened 2025-12-11
fix: scanning paganation and wallet request limits

SWvheerden:
Description
---
Fixes the pagination to correctly page and break on the correct limits.  Adds tests to enforce these
Fixes wallet request to be on avg less than 25mb


brianp (commented):
LGTM 👀 

Integration tests are failing. Are these intermittent failures or specifically related?

PR #7639, opened 2025-12-14
fix: conflicting short options in console wallet

canardleteer:
The console wallet would have a runtime panic when started, becuase of the conflicting `-p` declarations.

Since `payment-id` is more commonly used as `-p` than `pre-image`, it was chosen for alignment with the other subcommands.

Description
---

On `development`, running:
```shell
$ cargo run --bin minotari_console_wallet -- --some-bad-option
```

Exposed a runtime panic:

```shell
Running `target/debug/minotari_console_wallet --some-bad-option`
...
Command finalise-sha-atomic-swap: Short option names must be unique for each argument, but '-p' is in use by both 'pre-image' and 'payment-id'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Motivation and Context
---

I prefer CLI tools to not crash on startup?

How Has This Been Tested?
---
Locally only, and I expect that there is no CI testing in place to catch these kinds of things, or it wouldn't have gotten merged.

---

Run the following command on the PR branch & `development`:
```shell
$ cargo run --bin minotari_console_wallet -- --some-bad-option
```

The difference should be evident.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

- I personally do consider CLI option changes to be breaking, but since it never worked in the first place, it's an "unbreaking change?"


canardleteer:
/gemini review and update prior review with updated commit.

SWvheerden (approved):

PR #7646, opened 2025-12-18
chore: upgrade reqwest from v0.11 to v0.12

leet4tari:
Description
upgrade ```reqwest``` from v0.11 to v0.12
upgrade ```hyper``` from v0.14 to v1.6.0 - need for changes in ```reqwest``` v0.12

Motivation and Context
Avoid RUSTSEC-2025-0134 and assist with https://github.com/tari-project/tari-ootle/pull/1672

How Has This Been Tested?
Builds all binaries in locally and in fork

Check the audit report that reqwest is now v0.12 or above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


sdbondi (commented):
utACK

applications/minotari_merge_mining_proxy/src/run_merge_miner.rs
```diff
@@ -117,28 +115,58 @@ pub async fn start_merge_miner(cli: Cli) -> Result<(), anyhow::Error> {
...
+            println!("'[pools][self-select]' config setting that can be found in 'config/xmrig_config_***.json' or");
             println!("'<xmrig folder>/config.json'.");
             println!();
             Err(err.into())
         },
     }
 }
 
+async fn serve(listener: TcpListener, service: MergeMiningProxyService) -> Result<(), MmProxyError> {
+    loop {
```
SWvheerden: we need to add a shutdown here, you can use TariShutdown if you want, but this is an endless loop here

sdbondi: Shutdown here isn't necessary; the calling code will stop polling the future returned from `serve` on interrupt/shutdown/ctrl+c  and will drop it at the end of the function.

```rust
  let mut shutdown = Box::pin(tokio::signal::ctrl_c());
            let mut serve_fut = Box::pin(serve(listener, randomx_service));
            tokio::select! {
                _ = &mut shutdown => {
                    info!(target: LOG_TARGET, "Ctrl-C received, shutting down merge mining proxy...");
                    println!("Ctrl-C: shutting down merge mining proxy...");
                }
                result = &mut serve_fut => {
                    if let Err(e) = result {
                        error!(target: LOG_TARGET, "Error in merge mining proxy service: {}", e);
                    }
                }
            }
            Ok(())
        },
```

SWvheerden: 100% correct
But if someone else calls this function without calling it from some tokio loop with a select, its a going to end up in an endless loop. 

I would rather we combine to two have an loop with the tokio select loop awaiting `listener.accept()` and the shutdown. 

The way `async fn serve(` is written is assumed its called from a service that will drop it somehow, but you would never know that just by looking at the rust function header or place from where its called.

sdbondi (approved):
utACK

SWvheerden:
test, works as intended

PR #7651, opened 2026-01-09
feat: add branch and bound as utxo selection option

SWvheerden:
Description
---
Add Branch and bound utxo selction algorithim


base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+}
+
+#[derive(Clone, Debug)]
+struct SelectionState<T> {
+    available_utxos: Arc<Vec<T>>,
+    selected_utxos: Vec<usize>,
+    current_value: MicroMinotari,
+    final_target: MicroMinotari,
+    parms: UtxoSectionParams,
+    waste: MicroMinotari,
```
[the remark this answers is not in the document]
SWvheerden: this is named waste in terms of the algo as it refers to waste.

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+        index_to_add: usize,
+        current_iterations: usize,
+        max_iterations: usize,
+        best_result: &mut Option<SelectionState<T>>,
+    ) -> usize {
+        self.add_utxo_sorted(index_to_add);
+        self.current_value += self
+            .available_utxos
+            .get(index_to_add)
+            .expect("utxo_index out of bounds")
```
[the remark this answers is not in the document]
SWvheerden: Its an internal function, it should not be getting a index that does not exist

martinserts (dismissed):

base_layer/transaction_components/src/utxo_selection/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn value(&self) -> MicroMinotari;
+}
+
+impl UtxoValue for MicroMinotari {
+    fn value(&self) -> MicroMinotari {
+        *self
+    }
+}
+
+impl UtxoValue for UnblindedOutput {
```
sdbondi, line 41: I suspect you should implement this for `&UnblindedOutput` as well, so that when you use the algorithm, you don't need to copy all the outputs in and out. Guess we'll know more when we attempt to make use of the BnB in the wallet.

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+    #[test]
+    fn test_single_large_utxo_among_small_ones() {
+        let mut utxos: Vec<MicroMinotari> = vec![MicroMinotari(5); 10];
+        utxos.push(MicroMinotari(1000));
+        let params = section_params(100, 0, 10, 5, 10);
+        let selector = BranchAndBoundUtxoSelector::new(utxos, params, default_params());
+        let result = selector.search().unwrap();
+        // Should select the large UTXO
+        assert_eq!(result.selected_utxos[0].value().as_u64(), 1000);
+        assert!(result.selected_utxos.len() == 1);
```
sdbondi: nit: 
```suggestion
        assert_eq!(result.selected_utxos.len(), 1);
```

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+            if iterations >= max_iterations {
+                break;
+            }
+            if (self.selected_utxos.len() >= self.params.input_limit) ||
+                (self.current_value >= self.params.total_target())
+            {
+                // tx is now at the limit, dont search further or we have the target
+                break;
+            }
+            if self.selected_utxos.contains(&i) {
```
sdbondi, line 148: `O(n)` but probably not an issue with the "normal" amount of UTXOs.

Maybe `self.selected_utxos.binary_search(&i).is_ok()` is faster (`O(log(n)`), or using a `BTreeSet`  but I don't think it's worth making this optimisation unless proven to be needed.

sdbondi (dismissed):
Looks great, thanks for all the tests

base_layer/transaction_components/src/utxo_selection/mod.rs
SWvheerden: The method is `&T`, so there is no benefit in doing this for  `&UnblindedOutput`, unless we expand the trait to include and use `T`

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+        best_result: &mut Option<SelectionState<T>>,
+    ) -> usize {
+        self.add_utxo_sorted(index_to_add);
+        self.current_value += self
+            .available_utxos
+            .get(index_to_add)
+            .expect("utxo_index out of bounds")
+            .value();
+        self.waste += self.params.fee_per_input;
+        self.params.target_amount += self.params.fee_per_input;
```
brianp, line 171: Won't this mean the search requirement is changing the further along we are in the recursion. So previously parsed outputs that may now meet the requirement would be missed, because the didn't meet the requirement at the time they were processed in the loop?

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -0,0 +1,840 @@
...
+            // tx is now at the limit, dont search further
+            return current_iterations;
+        }
+        if let Some(best) = best_result {
+            // we are already worse off than the best here, stop right here
+            if self.waste + self.params.fee_per_input >= best.waste {
+                // no need to continue searching this branch
+                return current_iterations;
+            }
+        }
```
brianp, line 190: At the limit or over the limit? 

If we're at the limit wouldn't it mean we have an exact match. If we're over the limit then well... We probably should have stopped the recursion earlier.

brianp (commented):
Are there no weight considerations? We have limits for number of utxo but what if the first one it picks is huge?

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
SWvheerden: Branch and bound selects the solution with the least waste, which is in effect the  fee paid. 
So if our current branch + its possible solutions are at this point in time equal or worste than the best solution if we have to add another output, this whole branch cannot provide a better solution, so we stop searching this branch and continue searching a new branch.

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
SWvheerden: The target amount is the amount we need to fund this transaction, meaning fee + amount we want to send. 
So every time we add an output, we need to ensure that we increase our target so that we can fund this transaction.

SWvheerden:
The weight is simplified to try to make it easier; we can expand this if needed in the future. Our algorithm will, before we pass it in here, use the weight to calculate the fee per input and output. Most if not all, utxos in the chain use a push_pubkey script. This simplifies it and assumes that the weight will be identical for most

PR #7658, opened 2026-01-16
fix: burn claim flow

stringhandler:
Reopening this pr because the previous one had al ot of AI review noise


base_layer/transaction_key_manager/src/legacy_key_manager/inner.rs
```diff
@@ -644,6 +646,23 @@ where TBackend: TransactionKeyManagerBackend + 'static
...
+        payment_id: MemoField,
+    ) -> Result<EncryptedData, KeyManagerError> {
+        let mask = self.get_private_key(commitment_mask_key_id)?;
+        let shared_secret = self.get_diffie_hellman_shared_secret(sender_offset_key_id, claim_public_key)?;
+        let encryption_key = one_sided::public_key_to_output_encryption_key(&shared_secret)?;
+        let commitment = self.get_commitment(commitment_mask_key_id, &PrivateKey::from(value))?;
+
+        EncryptedData::encrypt_data(&encryption_key, &commitment, value.into(), &mask, payment_id)
+            .map_err(|e| KeyManagerError::UnexpectedError(e.to_string()))
+    }
```
SWvheerden:
```suggestion
     self.encrypt_burn_data_with_dh(commitment_mask_key_id,sender_offset_key_id,claim_public_key,value,payment_id)
    }
```

the legacy key manager is a wrapper around the key manager just for the console wallet so that we can handle old legacy keys. No need to reimplement the function again.

base_layer/wallet/migrations/2025-12-19-050127_burnt_proofs_add_encrypted_data_and_value/down.sql
```diff
@@ -0,0 +1,30 @@
+-- Remove encrypted_data and value columns from burn_proofs table
```
SWvheerden, line 1: Cant we combine these two migrations, they touch the same table, and do the same thing, its just different colomns?

base_layer/wallet/src/transaction_service/service.rs
```diff
@@ -3000,24 +3000,42 @@ where
...
+                payment_id.clone(),
+            )?;
+        } else {
+            output_builder = output_builder.encrypt_data_for_recovery(
+                &self.resources.transaction_key_manager_service,
+                Some(&recovery_key_id),
+                payment_id.clone(),
+            )?;
+        }
+
```
SWvheerden: the tx builder will change the encrypted data, it needs to be updated with this logic as well. 
The tx builder updates the fee inside the encrypted data to match the correct fee. at this point in time we dont yet know the final fee.

base_layer/transaction_components/src/key_manager/manager.rs
```diff
@@ -1310,6 +1310,23 @@ impl TransactionKeyManagerInterface for KeyManager {
...
+    ) -> Result<EncryptedData, KeyManagerError> {
+        let mask = self.get_private_key(commitment_mask_key_id)?;
+        let shared_secret = self.get_diffie_hellman_shared_secret(sender_offset_key_id, claim_public_key)?;
+        let encryption_key = public_key_to_output_encryption_key(&shared_secret)?;
+        let commitment = self.get_commitment(commitment_mask_key_id, &PrivateKey::from(value))?;
+
+        EncryptedData::encrypt_data(&encryption_key, &commitment, value.into(), &mask, payment_id)
+            .map_err(|e| KeyManagerError::UnexpectedError(e.to_string()))
+    }
+
```
SWvheerden: why do we need this function?
the burn data just uses a different public key in the DH?
Normal encrypted data is DH with the sender_off + recipient_view_key
this is DH with sender_off + claim_key?

This can just be replaced with this:
```
let encryption_key = TariKeyId::DHEncryptedData {
            private_key: sender_offset_private_key.key_id.clone().into(),
            public_key: claim_public_key,
        };
 output_builder = output_builder.encrypt_data_for_recovery(
                &self.resources.transaction_key_manager_service,
                Some(&encryption_key),
                payment_id.clone(),
            )?;
```

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: but see my above comment, I dont think we need this special logic

SWvheerden (changes requested):

base_layer/wallet/migrations/2025-12-19-050127_burnt_proofs_add_encrypted_data_and_value/down.sql
stringhandler: Why? What purpose is there in redoing this work?

base_layer/wallet/src/transaction_service/service.rs
stringhandler: The process is this:
The second layer wallet provides a claim_public_key, the base layer encrypts the data needed for spending the commitment but using this CPK as a public view key. 

It is not the same as a standard encrypted data that the base layer will recover. 

Not doing it this way would require either the base layer wallet knowing the second layer's private view key or the second layer wallet knowing the base layer's private view key.

The fact that the encrypted data is changed is tbh a bad design and lost me a lot of time

base_layer/transaction_components/src/key_manager/manager.rs
stringhandler: Ok I will try that. Just a note that this KeyId pattern is not obvious and I only found it after a few times of going through the code.

base_layer/wallet/migrations/2025-12-19-050127_burnt_proofs_add_encrypted_data_and_value/down.sql
SWvheerden: its one pr adding two migrations?

base_layer/wallet/src/transaction_service/service.rs
SWvheerden: yeah not ideal, its best of the bad designs we have so far. All its doing is updating the fee, but to update the fee, we need to update the encrypted data, as the data is already stored in encrypted form.

base_layer/wallet/migrations/2025-12-19-050127_burnt_proofs_add_encrypted_data_and_value/down.sql
stringhandler: ok I will fix it, retest, rereview

base_layer/transaction_components/src/key_manager/manager.rs
stringhandler: removed the problem method

SWvheerden (approved):

PR #7660, opened 2026-01-17
fix(mm_proxy): update reserved_offset after inserting merge mining tag

HashVault:
Description
---
The merge mining proxy inserts a merge mining tag at the beginning of the Monero coinbase transaction's extra field, which shifts all subsequent data by 35 bytes. However, the `reserved_offset` field in the response was not being updated to reflect this shift.

This caused mining pools to have to calculate the correct nonce position themselves instead of using the provided `reserved_offset` value.

This fix adjusts `reserved_offset` by adding the merge mining tag size (35 bytes) after modifying the block template.

Motivation and Context
---
Mining pools that use the `reserved_offset` value from the proxy response were receiving an incorrect offset (e.g., 131) when the actual position in the modified blob was different (e.g., 166). This forced pools to calculate the correct position themselves by searching for the nonce pattern in the blob.

How Has This Been Tested?
---
Empirical testing the offset mismatch: daemon reported 131, actual position was 166, difference = 35 bytes.

---
1. Start the merge mining proxy
2. Request a block template via getblocktemplate RPC
3. Verify that reserved_offset in the response equals the original monerod reserved_offset + 35

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/minotari_merge_mining_proxy/src/proxy/inner.rs
+++ b/applications/minotari_merge_mining_proxy/src/proxy/inner.rs
@@ -67,6 +67,11 @@ const LOG_TARGET: &str = "minotari_mm_proxy::proxy::inner";
 /// The identifier used to identify the tari aux chain data
 const TARI_CHAIN_ID: &str = "xtr";
 const BUSY_QUALIFYING: &str = "BusyQualifyingMonerodUrl";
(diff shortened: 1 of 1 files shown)


SWvheerden:
I believe this has been fixed? Please reopen if need be

PR #7662, opened 2026-01-20
fix: user pay for fee and replace by fee

SWvheerden:
Description
---
Fixes user pay for fee and replace by fee wallet calls
Fixes base node settings
Fixes some cucumber tests


base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -475,7 +475,6 @@ impl OutputSql {
         // First, get the must-include outputs
         let mut must_include_query = outputs::table
             .into_boxed()
```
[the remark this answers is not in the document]
SWvheerden: We specifically use this call to double spend outputs via replace by fee where we want to replace the transaction with a higher fee version, so we need to select the same output again.

martinserts (dismissed):

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
```diff
@@ -502,7 +501,7 @@ impl OutputSql {
         let i64_amount = i64::try_from(amount).unwrap_or(i64::MAX);
 
         // If must-include outputs are sufficient, return only them
-        if must_include_total >= i64_amount {
+        if must_include_total > i64_amount {
```
brianp, line 505: What does this change achieve. Why was equality considered wrong?

brianp (dismissed):

base_layer/wallet/src/output_manager_service/storage/sqlite_db/output_sql.rs
SWvheerden: Yes, if it chooses he exact amount, there is no funds left to pay the fee increase

SWvheerden: added a comment clarifying.

PR #7671, opened 2026-01-30
feat: adds branch and bound to be used by console wallet

SWvheerden:
Description
---
links in the branch and bound to the console wallet. 
Adjust branch and bound to prioritise the highest change value if the waste is equal between solutions. As higher change utxos will be more usefull in future.

### DIFF
--- a/base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
+++ b/base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
@@ -59,6 +59,23 @@ where T: UtxoValue
 
         best_result.map(|state| state.to_selection_result())
     }
+
+    /// Search for the best UTXO selection while ensuring the specified UTXOs are always included.
+    /// The `must_select` UTXOs will be pre-selected before the search begins, guaranteeing they
+    /// are part of the final selection even if they wouldn't be the optimal choice.
+    pub fn search_with_must_select(&self, must_select: Vec<T>) -> Option<SelectionResult<T>> {
+        let mut initial_state = SelectionState::new_with_selected_utxos(
+            self.available_utxos.clone(),
+            Arc::new(must_select),
+            self.search_params.clone(),
+            self.params.allow_dust_waste,
+        );
+        let mut best_result: Option<SelectionState<T>> = None;
+
+        initial_state.start_search(self.params.max_search_iterations, &mut best_result);
+
+        best_result.map(|state| state.to_selection_result())
+    }
 }
 
 pub struct BranchAndBoundUtxoSelectorParams {
@@ -116,8 +133,12 @@ impl<T> SelectionState<T>
 where T: UtxoValue
 {
     fn new_blank(available_utxos: Arc<Vec<T>>, params: UtxoSectionParams, allow_dust_waste: bool) -> Self {
+        // Sort UTXOs by value in descending order (biggest first)
+        let mut sorted_utxos = (*available_utxos).clone();
+        sorted_utxos.sort_by_key(|b| std::cmp::Reverse(b.value()));
+
         Self {
-            available_utxos,
+            available_utxos: Arc::new(sorted_utxos),
             selected_utxos: Vec::new(),
             current_value: MicroMinotari::from(0),
             waste: MicroMinotari::from(0),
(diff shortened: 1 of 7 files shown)


base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -694,10 +806,26 @@ mod tests {
...
+        // total_target() = 110 + 20 = 130
+        // 200 > 130 + 25 = 155 ✓ change is viable
+        // final_fee starts at 20, then += 10 (input) = 30, then += 15 (change) = 45?
+        // But test shows 50... let me check if there's something with how final_fee is modified
+        // Actually looking at the code, when adding input, params.target_amount is modified
+        // So for 1 input: final_fee = 20 + 10 = 30, then change_fee = 15, total = 45
+        // But getting 50 means there's 5 extra somewhere...
+        // Oh wait, the comparison logic adds change_cost as extra_waste, but change_cost includes fee_per_input
+        // Let me verify the actual value
+        assert_eq!(result.final_fee, MicroMinotari::from(50));
```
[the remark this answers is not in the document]
SWvheerden: No the test is correct, the comments are wrong, fixed them explaining whats wrong

base_layer/wallet/tests/output_manager_service_tests/service.rs
```diff
@@ -362,14 +361,8 @@ async fn test_utxo_selection_no_chain_metadata() {
...
         .unwrap();
-    // test that lowest 2 maturities were encumbered
     let utxos = oms.get_unspent_outputs().await.unwrap();
-    assert_eq!(utxos.len(), 8);
-    for (index, utxo) in utxos.iter().enumerate() {
-        let i = index as u64 + 3;
-        assert_eq!(utxo.wallet_output.features().maturity, i);
-        assert_eq!(utxo.wallet_output.value(), i * amount);
-    }
+    assert_eq!(utxos.len(), 9);
```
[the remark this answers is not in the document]
SWvheerden: The previous logic simply test that that the utxo's where selected in order of age, not a very good selection logic. We dont select via age anymore, so we can't test this. There are multiple additional tests to ensure that the selection algorithim does select optimally.

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -107,41 +120,77 @@ struct SelectionState<T> {
...
+        must_select: Vec<T>,
+        params: UtxoSectionParams,
+        allow_dust_waste: bool,
+    ) -> Self {
+        let mut new_blank = Self::new_blank(available_utxos, params, allow_dust_waste);
+        for utxo in &must_select {
+            let index = new_blank
+                .available_utxos
+                .iter()
+                .position(|u| u == utxo)
```
brianp, line 162: This seems like it could be a performance issue for large (pool) wallets etc. 

Would it be better to convert it to a btree or map?

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
```diff
@@ -107,41 +120,77 @@ struct SelectionState<T> {
...
+                .expect("must_select UTXO not found in available_utxos");
```
brianp: `expect` should probably be an `Err` return. I see this doesn't return a Result. Are we sure that's the best path?

base_layer/wallet/src/output_manager_service/service.rs
```diff
@@ -1790,74 +1803,86 @@ where
...
+        let output_fee = fee_calc.calculate(
+            fee_per_gram,
+            0,
+            0,
+            num_outputs,
+            total_output_features_and_scripts_byte_size,
+        );
+        let input_fee = fee_calc.calculate(fee_per_gram, 0, 1, 0, 0);
+        let bnb = BranchAndBoundUtxoSelectionBuilder::new(uo)
+            .with_target_amount(amount + kernel_fee)
```
brianp, line 1817: kernel_fee was already added to the amount before this function was called at line https://github.com/tari-project/tari/pull/7671/files#diff-d6798df19e4bcb4a252285295d195488f3e6dd3e65f5d42ce9012d15971ab691R1875

While building the `UtxoSelectionCriteria`.

Are we sure this is accurate?

brianp (commented):
A couple comments, but is mostly looks good.

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
SWvheerden: I dont think it would, this will only really be slow if you have: 1 a very large set of utxos you are selecting from, and 2 a very large set of utxos you need to use.  
In both cases this function only gets called if you want to spend spesific utxos such as aggregation or user_pay_for_fee etc.

base_layer/wallet/src/output_manager_service/service.rs
SWvheerden: Yes, the amount is the target amount you want to send. You need to, but you need to start off by targeting the actual target amount plus the kernel fee. The BnB does not use or count the kernel fee. The BnB will add add output fees, and input fees to the target amount, but becasue it does not work or touch the kernel, we need to ensure that the target amount is high enough to cover the kernel. 

It will start off by saying you want to send 100T, and your kernel fee is 1T, so the actual target is 101T,
and for every input it uses the target goes up by 1T etc. So this keeps moving up as it searches for the ideal set. 

But the BnB keeps track of all fee's paid, but again, it does not know or count the kernel fee, thats why for reporting purposes, we need to add it in in ln 1875

base_layer/transaction_components/src/utxo_selection/branch_and_bound/branch_and_bound_selector.rs
SWvheerden: yeah perhaps, lets make it and error, that can fail it requires user input

PR #7679, opened 2026-02-10
chore: prepare easier crates releases

SWvheerden:
Description
---
makes a few changes to do easier crates.io releases


SWvheerden:
replaced by #7690

PR #7726, opened 2026-03-18
fix(wallet): save complete burn proof in file

sdbondi:
Description
---
fix(wallet): save complete burn proof in file

Motivation and Context
---
Use a shared file location for burn proofs, this allows L2 wallets to pick up burns for claiming

   ## Summary
   - Automatically writes a complete burn proof JSON file to a configurable directory after a burn transaction is confirmed and the merkle proof is fetched from the base node, removing the need for users to
   manually specify a file path in the TUI
   - Introduces `CompleteClaimBurnProof` and `BurnClaimProof` types in `tari_sidechain` that include all data needed by an L2 wallet daemon (merkle proof, abridged kernel, encrypted data, value)
   - Renames the existing `BurnClaimProof` to `PartialBurnClaimProof` to clarify it lacks the merkle proof and other fields only available after confirmation
   - Adds a new `TransactionBurnConfirmed` event emitted when the proof file is written
   - Adds `--burn-proof-out` CLI flag and `burn_proofs_dir` config option (defaults to platform data dir e.g. `~/.local/share/tari/burn_proofs`)
   - Removes the manual "burn proof filepath" input field from the console wallet TUI burn tab

How Has This Been Tested?
---
Manually

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


base_layer/wallet/src/transaction_service/protocols/fetch_claim_burn_merkle_proofs.rs
```diff
@@ -140,3 +191,42 @@ where
 
     Ok(())
 }
+
+async fn write_burn_proof_to_file<P: AsRef<Path>>(burn_proofs_dir: P, proof: DbBurnProof) -> anyhow::Result<()> {
```
SWvheerden: I'm not a fan of writing a file deep in the library. For now, this is probably fine as this wallet is for now only going to serve the console wallet, which has a range of weird dependancies. But we should avoid this for the other crates

SWvheerden (dismissed):

base_layer/wallet/src/transaction_service/protocols/fetch_claim_burn_merkle_proofs.rs
sdbondi: Yup I agree, I initially added the new event to handle this, however the event monitor doesnt run for grpc only/headless mode. So would need to implement something else that monitors the events. Instead I kept it simple and hopefully architectural issues can be resolved in the new wallet

sdbondi: @SWvheerden I was kind of forced to implement a separate handler for the file writing in the wallet app level because I realised that we should store the proof per network e.g. ~/.local/share/tari/esmeralda/burn_proofs and passing the network around into the various layers of the wallet became ... just bad...

SWvheerden (approved):

PR #7728, opened 2026-03-23
feat: change logic in node pool management

SWvheerden:
Description
---
Completely redo and change pool logic in the node. 
Node will every 10 min by default change 10% of its pool to make sure that islands dont form. 
Node will refresh entire pool if there are no connections. 
Fixes a bad migration on the sql.

### DIFF
--- a/comms/dht/src/connectivity/mod.rs
+++ b/comms/dht/src/connectivity/mod.rs
@@ -425,7 +425,7 @@ impl DhtConnectivity {
             .drain(..)
             .partition::<Vec<_>, _>(|n| disconnected.iter().any(|c| c.peer_node_id() == n));
         // we want add and at most 10% new peers or at least 1 peer
-        let keep_size = max(pool_size / 10, 1);
+        let keep_size = pool_size - max(pool_size * self.config.connectivity.churn_rate / 100, 1);
         while connected.len() > keep_size {
             // we remove a random peer so as not to keep swapping the same peer each time.
             let mut rng = thread_rng();
--- a/comms/dht/src/connectivity/mod.rs
+++ b/comms/dht/src/connectivity/mod.rs
@@ -38,23 +38,17 @@ use std::{
     sync::Arc,
     time::{Duration, Instant},
 };
-
+use std::cmp::max;
 use log::*;
+use rand::{thread_rng, Rng};
 pub use metrics::{MetricsCollector, MetricsCollectorHandle};
-use tari_comms::{
-    Minimized,
-    PeerConnection,
-    PeerManager,
-    connectivity::{
-        ConnectivityError,
-        ConnectivityEvent,
-        ConnectivityEventRx,
-        ConnectivityRequester,
-        ConnectivitySelection,
-    },
-    multiaddr,
-    peer_manager::{NodeId, Peer, PeerManagerError},
-};
+use tari_comms::{Minimized, PeerConnection, PeerManager, connectivity::{
+    ConnectivityError,
+    ConnectivityEvent,
+    ConnectivityEventRx,
+    ConnectivityRequester,
+    ConnectivitySelection,
+}, multiaddr, peer_manager::{NodeId, Peer, PeerManagerError}, PeerConnectionError};
 use tari_shutdown::ShutdownSignal;
 use thiserror::Error;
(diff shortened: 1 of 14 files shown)


comms/dht/src/connectivity/mod.rs
```diff
@@ -320,40 +325,56 @@ impl DhtConnectivity {
...
+            new_peers.len(),
+        );
+        let old_peers = self.random_pool.drain(..).collect::<Vec<_>>();
+        for peer in &new_peers {
+            self.insert_random_peer(peer.clone());
         }
-
+        // Make sure we drop any connection handles that removed from the pool
+        for old_peer in &old_peers {
+            self.remove_connection_handle(old_peer).await?;
```
sdbondi, line 373: When you drop a connection handle, the reaper inside the connection manager will disconnect any peers what have no handle. So this should cause a lot of connection churn (depending on how often the reaping happens etc)

The previous code would look for disconnected peers in the random pool, and only connect new peers to get the number of connected peers back up to the target value. TARGET - NUM_DISCONNECTED = NUM_NEW_PEERS_TO_CONNECT so if we have good connections to TARGET peers then the function is a no-op.

I suspect you want to preserve (something like) this behaviour rather than potentially cause churn?

comms/dht/src/connectivity/mod.rs
```diff
@@ -320,40 +325,56 @@ impl DhtConnectivity {
...
         if try_revive_connections {
-            self.redial_pool_peers_as_required().await?;
+            let should_refresh = self
+                .random_pool_last_refresh
+                .map(|instant| instant.elapsed() >= self.config.connectivity.random_pool_refresh_interval)
+                .unwrap_or(true);
+            if should_refresh {
+                self.refresh_entire_pool().await?;
+            } else {
+                debug!(
```
sdbondi, line 339: With the call to `refresh_random_pool` moved into an else (`!refresh_entire_pool`) this will never "revive" the connections, so the debug message doesnt do what it says it does

comms/dht/src/connectivity/mod.rs
```diff
@@ -320,40 +325,56 @@ impl DhtConnectivity {
...
             self.random_pool.len(),
         );
 
         if try_revive_connections {
-            self.redial_pool_peers_as_required().await?;
+            let should_refresh = self
+                .random_pool_last_refresh
+                .map(|instant| instant.elapsed() >= self.config.connectivity.random_pool_refresh_interval)
+                .unwrap_or(true);
+            if should_refresh {
```
sdbondi, line 336: Is there a reason to churn connections after an interval?

comms/dht/src/connectivity/mod.rs
```diff
@@ -503,24 +557,29 @@ impl DhtConnectivity {
...
+        Ok(())
     }
 
-    fn remove_connection_handle(&mut self, node_id: &NodeId) {
+    async fn remove_connection_handle(&mut self, node_id: &NodeId) -> Result<(), PeerConnectionError> {
         if let Some(idx) = self.connection_handles.iter().position(|c| c.peer_node_id() == node_id) {
-            let conn = self.connection_handles.swap_remove(idx);
+            let mut conn = self.connection_handles.swap_remove(idx);
+            if conn.is_connected() {
+                conn.disconnect(Minimized::No, "DhtConnectivty").await?;
```
sdbondi, line 584: Ah I see that the connection is immediately disconnected, not reaped. If some other service is using this connection, it will be interrupted (the reaper will not disconnect until its actually unused). This may be mostly fine (it's more of an issue with wallets using P2P, which they dont now), but I still wonder why we keep on disconnecting and reestablishing connections that were previously stable.

comms/dht/src/connectivity/mod.rs
SWvheerden: this should only trigger if there are no connections, but we also dont want to keep triggering this, we want give time to make new connections, hence the wait

comms/dht/src/connectivity/mod.rs
SWvheerden: this is so we can force some churn and stop islands from forming in the network

sdbondi (dismissed):

PR #7734, opened 2026-04-01
feat: payref tracking

SWvheerden:
Description
---
Add historical payref tracking

### DIFF
--- a/base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
+++ b/base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
@@ -29,11 +29,14 @@ use chacha20poly1305::XChaCha20Poly1305;
 use chrono::{DateTime, NaiveDateTime, Utc};
 use diesel::{prelude::*, result::Error as DieselError};
 use log::*;
-use tari_common_sqlite::{sqlite_connection_pool::PooledDbConnection, util::diesel_ext::ExpectedRowsExtension};
+use tari_common_sqlite::{
+    sqlite_connection_pool::PooledDbConnection,
+    util::{diesel_ext::ExpectedRowsExtension, retry::retry_db},
+};
 use tari_common_types::{
     burn_proof::{EncodedMerkleProof, PartialBurnClaimProof},
     encryption::{Encryptable, decrypt_bytes_integral_nonce, encrypt_bytes_integral_nonce},
-    payment_reference::generate_payment_reference,
+    payment_reference::{PaymentReference, generate_payment_reference},
     tari_address::TariAddress,
     transaction::{
         LegacyTransactionStatus,
@@ -55,7 +58,14 @@ use zeroize::Zeroize;
 
 use crate::{
     schema,
-    schema::{completed_transactions, inbound_transactions, outbound_transactions, payrefs, scanned_blocks},
+    schema::{
+        completed_transactions,
+        inbound_transactions,
+        outbound_transactions,
+        payref_history,
+        payrefs,
+        scanned_blocks,
+    },
     storage::{
         serializers,
         serializers::bincode_decode,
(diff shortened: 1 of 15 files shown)


brianp (dismissed):
Looks good. Clean refactoring with `completed_tx_to_transaction_info`, migration and indexes are solid, tests cover the key paths.

A few things worth considering:

1. **`retry_db` uses `thread::sleep`** which blocks the tokio runtime since these DB methods are called from async contexts without `spawn_blocking`. With 10 retries and up to 5s backoff, worst case stalls all async tasks on that thread for a while. Might want to cap it tighter or move callers to a blocking thread.

2. **Payref archival in `process_reorg` isn't wrapped in a transaction.** If it fails mid-loop, some payrefs get archived and others don't. A crash + retry could also duplicate history entries. Worth wrapping in `conn.transaction()`.

3. **`as` casts between `i64`/`u64` for `tx_id`** silently wrap on negative values. `TryFrom` would be safer for the new code paths.

None of these are blockers but 1 and 2 could bite in production under load or during actual reorgs.

common_sqlite/src/util/retry.rs
```diff
@@ -0,0 +1,94 @@
...
+                        );
+                    }
+                    return Err(e);
+                }
+                warn!(
+                    target: LOG_TARGET,
+                    "{}: database is locked (attempt {}/{}), retrying in {:?}",
+                    op_name,
+                    attempt + 1,
+                    DEFAULT_MAX_RETRIES,
```
brianp, line 85: `thread::sleep` here blocks the tokio runtime since callers like `update_mined_height` and `short_term_encumber_outputs` run in async contexts without `spawn_blocking`. With 10 retries and 5s max backoff, this could stall all async tasks on that thread.

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1430,7 +1486,16 @@ impl TransactionBackend for TransactionServiceSqliteDatabase {
...
             .load::<i64>(&mut conn)?;
         for tx_id in &tx_ids {
-            CompletedTransactionSql::set_as_unmined((*tx_id as u64).into(), &mut conn)?;
+            // Archive existing payrefs to history before unmining (for reorg tracking).
+            // This is done here rather than in set_as_unmined because set_as_unmined is also
+            // called during normal validation (not just reorgs), and the extra DB operations
+            // would cause lock contention under high load.
+            let tx_id_typed = TxId::from(*tx_id as u64);
+            let existing_payrefs = PayrefSql::find_all_by_tx_id(tx_id_typed, &mut conn)?;
+            for pr in &existing_payrefs {
```
brianp: This archive + unmined loop isn't wrapped in a `conn.transaction()`. If it fails partway through, some payrefs get archived and others don't. A retry would also duplicate history entries since `archive_from_payref` has no conflict handling.

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
```diff
@@ -1430,7 +1486,16 @@ impl TransactionBackend for TransactionServiceSqliteDatabase {
             .filter(completed_transactions::mined_height.ge(reorg_height as i64))
             .load::<i64>(&mut conn)?;
         for tx_id in &tx_ids {
-            CompletedTransactionSql::set_as_unmined((*tx_id as u64).into(), &mut conn)?;
+            // Archive existing payrefs to history before unmining (for reorg tracking).
+            // This is done here rather than in set_as_unmined because set_as_unmined is also
+            // called during normal validation (not just reorgs), and the extra DB operations
```
brianp: `*tx_id as u64` silently wraps negative values. `TryFrom` would be safer here.

brianp (dismissed):

base_layer/wallet/src/transaction_service/storage/sqlite_db.rs
SWvheerden: Its definatly happening but thats fine, otherwise we would get invalid tx_ids, as we do wrap them into negative. 
But we also unwrap them again before using, so it goes negative i64 to positive u64 again

PR #7768, opened 2026-04-17
fix: bnb edge case

SWvheerden:
Description
---
Fixes an edge case of branch and bound to correctly adjust for memofield size


martinserts (dismissed):

sdbondi (dismissed):
utACK

PR #7774, opened 2026-04-20
fix: better sync

SWvheerden:
Description
---
Makes sure we keep the sync peer initial state, and listens for a termination signal for a closed connection.


martinserts (approved):

sdbondi (approved):
Comms changes make sense

PR #7781, opened 2026-04-23
fix: sync peer swapping

SWvheerden:
Description
---
Its possible for a peer to be disconnected, banned, etc during peer sync. 
If this happens a the next sync peer should be used, but its possible for the peer manager to reap that connection. 
This causes the sync to get stuck trying to await a connection thats not made. 
This fixes the issue where it awaits a connection to a disconnected peer, 
and it adds a sync peer list to the peer manager telling it that these peers are being used, dont reap their connections


sdbondi (approved):
utACK As a quick fix - but we should look at addressing the root cause of the disconnected peer connection right after dialing

PR #7790, opened 2026-04-27
fix(deps): upgrade diesel to 2.3.8 for RUSTSEC-2026-0111

### LINKED ISSUE #7787: RUSTSEC-2026-0111: Possible UTF-8 corruption in Diesels SQLite backend
> Possible UTF-8 corruption in Diesels SQLite backend

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unsound                |
| Package             | `diesel`                      |
| Version             | `2.2.10`                   |
| URL                 | [https://github.com/diesel-rs/diesel/pull/5042](https://github.com/diesel-rs/diesel/pull/5042) |
| Date                | 2026-04-24                         |

Diesel uses the `sqlite3_value_text` function to receive strings from SQLite while deserializing query results. We misinterpreted the corresponding [SQLite](https://sqlite.org/c3ref/value_blob.html) documentation that this function always returns a UTF-8 encoded string values as `*const c_char`. Based on that we used `str::from_utf8_unchecked` to construct a Rust string slice without any additional UTF-8 checks in place. It turned out that this function doesn&#39;t always return correct UTF-8 strings. For field of the SQLite side storage type `BLOB` this pointer can contain arbitrary bytes, which makes the usage of `str::from_utf8_unchecked` unsound as this violates the safety contract of `str` to only contain valid UTF-8 encoded Strings.

## Mitigation

The preferred mitigation to the outlined problem is to update to a Diesel version 2.3.8 or newer, which includes fixes for the problem.

## Resolution

Diesel now correctly checks whether the provides byte buffer is actually valid UTF-8, instead of relying on SQLite&#39;s documentation. This fix is included in the `2.3.8` release.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2026-0111.html) for additional details.

samrusani:
Description
---
Upgrade Diesel dependencies from 2.2.10 to 2.3.8 across Tari crates that use the SQLite backend, and refresh diesel_migrations to 2.3.

This addresses the RUSTSEC-2026-0111 advisory (possible UTF-8 corruption unsoundness in Diesel's SQLite backend).

Motivation and Context
---
Issue #7787 reports that the workspace lockfile includes vulnerable diesel 2.2.10.

This patch keeps the scope limited to dependency updates in crates that directly depend on Diesel:
- common_sqlite
- comms/core
- comms/dht
- base_layer/wallet
- base_layer/transaction_key_manager

How Has This Been Tested?
---
- cargo check --locked --ignore-rust-version -p tari_common_sqlite -p tari_comms -p tari_comms_dht -p tari_transaction_key_manager -p minotari_wallet

Note: local toolchain in this environment is rustc 1.92.0 while workspace requires 1.93.0; therefore --ignore-rust-version was used for local verification.

---
1. Confirm Cargo.toml Diesel version bumps in the five crates above.
2. Confirm Cargo.lock no longer contains diesel 2.2.10.
3. Run:
   cargo check --locked -p tari_common_sqlite -p tari_comms -p tari_comms_dht -p tari_transaction_key_manager -p minotari_wallet

Breaking Changes
---
- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

Closes #7787.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -785,10 +785,11 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.16"
+version = "1.2.61"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be714c154be609ec7f5dad223a33bf1482fff90472de28f7362806e6d4832b8c"
+checksum = "d16d90359e986641506914ba71350897565610e87ce0ad9e6f28569db3dd5c6d"
 dependencies = [
+ "find-msvc-tools",
  "jobserver",
  "libc",
  "shlex",
@@ -1568,6 +1569,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "darling"
+version = "0.21.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9cdf337090841a411e2a7f3deb9187445851f91b309c0c0a29e05f74a00a48c0"
+dependencies = [
+ "darling_core 0.21.3",
+ "darling_macro 0.21.3",
+]
+
+[[package]]
 name = "darling_core"
 version = "0.14.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


samrusani:
Pushed an updated commit signed via SSH (new head: e75e1bd30).

CI is showing “action_required” with 0 jobs, so I suspect it’s just waiting for maintainer approval to run workflows on this PR/fork.

SWvheerden (approved):

Issue #7795, opened 2026-04-29
Create better ledger installer


SWvheerden:
## Bounty: Create Better Ledger Installer

**Tier:** S — 15,000 XTM

### Description

Replace the existing per-OS/per-model shell scripts for installing the Minotari app on Ledger hardware wallets with a single cross-platform installer. The installer should auto-detect the connected Ledger model (Nano S, Nano S Plus, Nano X, Stax), download the correct firmware file, and install it — all in one step. Support macOS, Windows, and Linux.

### Acceptance Criteria

- [ ] A single installer executable or script works for macOS, Windows, and Linux
- [ ] The installer detects the connected Ledger model automatically without user input
- [ ] The correct Minotari app binary is downloaded and installed for the detected model
- [ ] The installer reports clear success/failure status and handles errors (no device connected, unsupported model) gracefully

### Context

- Original issue: https://github.com/tari-project/tari/issues/7795
- Current scripts live in the repo, one per OS/ledger type
- Ledger model detection can use the Ledger USB HID interface or `ledgerctl` tooling

### How It Works

1. Comment on this issue to signal intent (courtesy, not a lock)
2. Fork the repo and do the work
3. Submit a PR that references the issue with a closing keyword (e.g. `Fixes #7795`) so GitHub links your PR to the bounty automatically
4. Bounties are competitive. Multiple PRs may be submitted for the same bounty. Review comments are public, and contributors may borrow ideas from each other's work. The maintainer selects the PR that best solves the problem, considering code quality, completeness, and contributor behavior. In cases where multiple contributors made significant contributions, the maintainer may split the bounty. The maintainer's decision is final.
5. Earn XTM (processing may take up to a week; reach out to @possum on Discord with your Tari address)

### Notes

- AI-assisted development is expected and encouraged
- If you get stuck, ask in Discord

metalaureate:
<details><summary>Original issue description (preserved before bounty rewrite)</summary>

We have scripts for installing the minotari app on the wallet, one for each os/ledger type. 
Create an installer for mac and winodws to do the same thing, automatically deciding which ledger to download depending on the connected type. 

We should also include linux here

## Effort 
small

## Acceptance criteria
* Have a one click installer that runs detects the correct ledger, and downloads the correct file and installs it.

</details>

taiwofelix35:
ok

Sp1kyss:
Hey, submitted a PR for this — https://github.com/tari-project/tari/pull/7805

[...]

PR #7805, opened 2026-05-05
feat: unified cross-platform Minotari Ledger installer with auto-detection

Sp1kyss:
Closes #7795

## What this does

Replaces the per-model/per-OS scripts with a single installer that works on **macOS, Windows, and Linux** and auto-detects which Ledger is connected — no user input needed.

## How to run

**macOS / Linux:**
```bash
chmod +x applications/minotari_ledger_wallet/wallet/install_scripts/install_minotari_ledger.sh
./applications/minotari_ledger_wallet/wallet/install_scripts/install_minotari_ledger.sh
```

**Windows (PowerShell):**
```powershell
.\applications\minotari_ledger_wallet\wallet\install_scripts\install_minotari_ledger.ps1
```

## How detection works

1. Enumerates USB HID devices filtered by Ledger's vendor ID (`0x2C97`)
2. Maps the product ID to a model slug (`flex`, `stax`, `nanosplus`, `nanox`, `nanos`)
3. Falls back to `ledgerctl info` output parsing if the `hid` library isn't available

All known PID variants per model are covered (including alternate transport PIDs like `0x5000` for Nano X).

## Error handling

- No device detected → clear message with checklist (USB connected, PIN entered, Developer Mode on)
- Release asset not found → lists available assets so the user can report it
- Reports clear success/failure status for each step

## Files changed

| File | Description |
|---|---|
| `install_scripts/install_minotari_ledger.py` | Core installer (Python 3.8+, stdlib only + pip deps) |
| `install_scripts/install_minotari_ledger.sh` | macOS/Linux launcher (creates isolated venv, handles libusb) |
| `install_scripts/install_minotari_ledger.ps1` | Windows launcher (creates isolated venv) |
| `install_scripts/README.md` | Usage docs + troubleshooting |

The existing per-model scripts are untouched — this adds the unified option alongside them.

### DIFF
--- a/applications/minotari_ledger_wallet/wallet/install_scripts/README.md
+++ b/applications/minotari_ledger_wallet/wallet/install_scripts/README.md
@@ -4,13 +4,14 @@ A single cross-platform installer that automatically detects your connected Ledg
 
 ## Supported Ledger Models
 
-| Model | Detected via |
-|---|---|
-| Nano S | USB HID (PID 0x0001 / 0x1000) |
-| Nano S Plus | USB HID (PID 0x0004 / 0x4000) |
-| Nano X | USB HID (PID 0x0005 / 0x5000) |
-| Stax | USB HID (PID 0x0006 / 0x6000) |
-| Flex | USB HID (PID 0x0007 / 0x7000) |
+> Minotari does not ship an app for the original Nano S, only the Nano S Plus. The original Nano S is intentionally **not** detected by this installer.
+
+| Model | Slug | Detected via |
+|---|---|---|
+| Nano S Plus | `nanos+` | USB HID (PID 0x0004 / 0x4000) |
+| Nano X | `nanox` | USB HID (PID 0x0005 / 0x5000) |
+| Stax | `stax` | USB HID (PID 0x0006 / 0x6000) |
+| Flex | `flex` | USB HID (PID 0x0007 / 0x7000) |
 
 ## Requirements
 
@@ -40,11 +41,11 @@ python3 install_minotari_ledger.py
 
 ## What it does
 
-1. Installs Python dependencies (`ledgerctl`, `hid`, etc.) into an isolated virtual environment
-2. Detects the connected Ledger model via USB HID (with `ledgerctl info` as fallback)
-3. Fetches the latest Minotari release from GitHub
-4. Downloads and extracts the correct firmware zip for your model
-5. Installs the app onto the Ledger via `ledgerctl install`
+1. Installs Python dependencies (`ledgerwallet`, `protobuf`, `ecdsa`) into an isolated virtual environment. `hidapi` is pulled in transitively by `ledgerwallet`.
+2. Detects the connected Ledger model via USB HID (with `python -m ledgerwallet info` as fallback).
+3. Fetches the latest Minotari release from GitHub.
+4. Downloads and extracts the correct firmware zip for your model.
+5. Installs the app onto the Ledger via `python -m ledgerwallet install`.
 
 ## Troubleshooting
 
--- a/applications/minotari_ledger_wallet/wallet/install_scripts/README.md
+++ b/applications/minotari_ledger_wallet/wallet/install_scripts/README.md
@@ -1,0 +1,61 @@
+# Minotari Ledger Wallet — Unified Installer
+
+A single cross-platform installer that automatically detects your connected Ledger hardware wallet model and installs the correct Minotari app — no manual model selection needed.
+
+## Supported Ledger Models
+
+| Model | Detected via |
+|---|---|
+| Nano S | USB HID (PID 0x0001 / 0x1000) |
+| Nano S Plus | USB HID (PID 0x0004 / 0x4000) |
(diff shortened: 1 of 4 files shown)


SWvheerden (commented):
please check the review commands, and we dont support the nano, only the nanosplus

SWvheerden:
this looks fine, I just want to test this on a clean install machine

SWvheerden (changes requested):
I run this through claude, and there are some issues I verified. 

  1. Nano S Plus ↔ Nano X product IDs are swapped

  install_minotari_ledger.py:43-56. I verified against Ledger's authoritative device table (@ledgerhq/devices, vendor 0x2c97):

  ┌─────────────┬────────────┬─────────────┐
  │   Device    │ legacy PID │ modern (MM) │
  ├─────────────┼────────────┼─────────────┤
  │ Nano X      │ 0x0004     │ 0x40xx      │
  ├─────────────┼────────────┼─────────────┤
  │ Nano S Plus │ 0x0005     │ 0x50xx      │
  └─────────────┴────────────┴─────────────┘

  The PR has these reversed:
  # Nano S Plus
  0x0004: "nanos+",   # ← that's Nano X
  0x4000: "nanos+",   # ← that's Nano X
  # Nano X
  0x0005: "nanox",    # ← that's Nano S Plus
  0x5000: "nanox",    # ← that's Nano S Plus
  A connected Nano S Plus is detected as Nano X and vice-versa — the exact mis-classification the design set out to prevent. The
  README table (lines for Nano S Plus / Nano X PIDs) repeats the same swap.

  Secondary point: modern Ledger PIDs are MMii where ii is an interface bitfield (commonly 0x01, 0x11, 0x15, …), so exact-value
  matching misses the most common real-world PIDs (e.g. 0x4011, 0x5011) and silently falls through to the ledgerctl path. Matching
  on the high byte is far more robust:
  model = LEDGER_MM.get(pid >> 8) or LEDGER_LEGACY.get(pid)
  with {0x40: "nanox", 0x50: "nanosplus", 0x60: "stax", 0x70: "flex"}.

  2. Wrong slug for Nano S Plus — release asset will never be found

  install_minotari_ledger.py:36, 173. The published assets and in-zip manifests use nanosplus, confirmed by the live release
  (minotari_ledger_wallet-nanosplus-v5.4.0-pre.2-….zip) and every existing per-model script (app_nanosplus.json). The new code uses
  slug nanos+:
  pattern = f"minotari_ledger_wallet-{model}"   # → "minotari_ledger_wallet-nanos+"
  No asset contains nanos+, so fetch_asset_url raises "No release asset found for model 'nanos+'". The code comment and README claim
  nanos+ "match[es] the asset names" — that's false. ledger_app.toml's devices = [..., "nanos+", ...] is the build-target name, not
  the release-asset/manifest slug; the PR conflated the two. (The app_<model>.json lookup happens to survive via the app*.json
  fallback, but asset fetch fails first.) Fix: use nanosplus as the slug for nano s plus throughout. nanox/stax/flex slugs are
  correct.


  3. Hardcoded releases/latest excludes pre-releases — currently no ledger assets at all

  install_minotari_ledger.py:60. Ledger bundles currently ship only in pre-releases (v5.4.0-pre.2); the actual latest (v5.3.1) has
  none. GitHub's /releases/latest endpoint excludes pre-releases, so right now the installer fails for every model with "Available
  assets: [non-ledger…]". The old scripts mitigated this with a -t/--tag option.

  4. No tag override — @args/"$@" are forwarded but never read

  The launchers pass @args (ps1) / "$@" (sh) into the Python script, but main() ignores sys.argv. This is a feature regression vs.
  the existing per-model scripts (which support -t TAG for pre-releases) and compounds #3. Add argparse with --tag and build the API
  URL accordingly (/releases/tags/<tag>).
  
  5. No integrity verification of downloaded firmware

  Each release ships a .zip.sha256, but the installer downloads and flashes onto a hardware wallet without verifying it. For a
  security-sensitive device installer this is worth adding (download the .sha256, compare). The old scripts didn't either, but a
  from-scratch unified installer is the right place to fix it.

  
  - Emoji escapes don't render in the launchers. install_minotari_ledger.sh:9 etc. use echo "\U0001f527…" — bash echo prints the
  literal \U0001f527. Same in install_minotari_ledger.ps1 (Write-Host "\U0001f680…"); PowerShell uses backtick escapes, not \U. Use
  literal UTF-8 emoji, or printf/`u{…} respectively. (The .py file is fine — those are real Python string escapes.)
  - README.md "Direct Python (any platform)" runs pip install into the user's global interpreter, which hits PEP 668
  "externally-managed-environment" on modern macOS/Linux. The launchers handle this via venv; consider dropping/qualifying the
  direct-python path.
  - run([... "pip", "install", "--upgrade", "pip"], capture_output=True) (:122) swallows output on failure (check=True raises with
  no visible stderr). Drop capture_output or surface it on error.
  - Dependency reinstall every run — ensure_dependencies always upgrades pip and reinstalls; harmless but slow.

SWvheerden (changes requested):
Look at the new scripts, using json and ledgerctl does not work anymore

SWvheerden:
Closing in favour of #7864

PR #7811, opened 2026-05-05
fix: propagate mempool rejection details to wallet via HTTP API

chengyixu:
## Description

When a base node's mempool rejects a wallet transaction, the wallet currently receives only a generic "Validation Failed" message. This PR extends the HTTP JSON-RPC submit_transaction response to propagate the specific rejection reason so the wallet can display actionable feedback.

## Changes

1. **`TxStorageResponse` enum** (`mempool/mod.rs`): Added `Option<String>` detail field to `NotStored` and `NotStoredConsensus` variants. The detail carries the exact `ValidationError` message from the mempool's validation path.

2. **`TxSubmissionResponse` struct** (`tx_submission_response.rs`): New `details: Option<String>` field with `#[serde(skip_serializing_if = "Option::is_none")]` for backward compatibility — older wallet clients that don't know about this field will simply ignore it.

3. **Mempool storage** (`mempool_storage.rs`): Updated `insert()` to pass the `ValidationError` message into the `TxStorageResponse` variants instead of discarding it after logging.

4. **HTTP handler** (`submit_transaction.rs`): Threads the `details` through from `TxStorageResponse` into `TxSubmissionResponse`.

5. **Pattern match updates**: Updated all `match` arms across gRPC server, query service, RPC service, proto conversion, and tests to match the new variant signatures.

## Acceptance Criteria

- [x] Wallet receives a specific, human-readable reason when the mempool rejects (e.g., "Contains an unknown input", "Transaction contains already spent inputs")
- [x] The rejection reason identifies the exact failure from the mempool validator
- [x] Older wallet versions continue to work (new field is optional and skipped when None)
- [x] All 12 mempool tests pass, all 8 node_service tests pass, all transaction_components tests pass

## Testing

```
cargo test -p tari_core -- mempool    # 12 passed
cargo test -p tari_core -- node_service  # 8 passed
cargo test -p tari_transaction_components  # all passed
```

Closes #7776

PR #7816, opened 2026-05-05
feat: surface detailed rejection reasons in HTTP transaction submission

### LINKED ISSUE #7776: Better feedback of rejected transactions
## Bounty: Better Feedback of Rejected Transactions

**Tier:** M — 60,000 XTM

### Description

When a base node's mempool rejects a wallet transaction, the wallet currently receives only a generic "validation failure" message. Extend the HTTP interface to propagate the specific rejection reason (e.g., double-spend of commitment X, insufficient fee, invalid range proof) so the wallet can display actionable feedback to the user. The change must be backward-compatible — older wallets that don't understand the new fields must continue to function normally.

### Acceptance Criteria

- [ ] The wallet receives a specific, human-readable reason when the mempool rejects a transaction (not just a generic validation error)
- [ ] The rejection reason identifies the exact failure (e.g., which commitment was double-spent, which rule was violated)
- [ ] Older wallet versions continue to work correctly against a base node with this change (backward-compatible HTTP response)
- [ ] At least one integration or cucumber test verifies that a rejected transaction surfaces the detailed reason to the wallet

### Context

- Original issue: https://github.com/tari-project/tari/issues/7776
- The mempool validation logic already knows the specific reason; this is about surfacing it through the HTTP API
- Must only modify the HTTP interface (not gRPC) per the original issue constraints

### How to Claim

### Notes

ryriigh:
## Summary

When the mempool rejects a wallet transaction via the HTTP API, the response now includes a `rejection_detail` field with a human-readable explanation identifying the specific failure. This surfaces information that was previously only logged internally.

### Changes

- Added `rejection_detail: Option<String>` to `TxSubmissionResponse` (skipped when null for backward compatibility)
- Extended `MempoolStorage::insert` to return rejection detail strings from validation errors
- Added `Mempool::insert_detailed()` method that preserves the detail
- Added `SubmitTransactionDetailed` request/response variants in the mempool service layer
- HTTP handler now calls the detailed path and populates rejection reasons

### Example responses

Accepted:
```json
{"accepted": true, "rejection_reason": "None", "is_synced": true}
```

Rejected (new `rejection_detail` field):
```json
{"accepted": false, "rejection_reason": "ValidationFailed", "is_synced": true, "rejection_detail": "Consensus rule violated: script must be exactly one OP_PUSHPUBKEY"}
```

### Backward Compatibility

- `rejection_detail` is `Option<String>` with `#[serde(skip_serializing_if = "Option::is_none")]` — older wallets never see the field
- Existing `MempoolHandle::submit_transaction()` is unchanged — gRPC path unaffected
- All existing tests pass without modification

### Acceptance Criteria

- [x] Wallet receives specific, human-readable reason on rejection
- [x] Identifies exact failure (which rule violated, fee amounts, etc.)
- [x] Backward-compatible (older wallets continue working)
- [ ] Integration test (cucumber/integration test to be added if required)

Closes #7776

### DIFF
--- a/base_layer/transaction_components/src/rpc/models/tx_submission_response.rs
+++ b/base_layer/transaction_components/src/rpc/models/tx_submission_response.rs
@@ -29,6 +29,8 @@ pub struct TxSubmissionResponse {
     pub accepted: bool,
     pub rejection_reason: TxSubmissionRejectionReason,
     pub is_synced: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rejection_detail: Option<String>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
--- a/applications/minotari_node/src/http/handler/json_rpc/submit_transaction.rs
+++ b/applications/minotari_node/src/http/handler/json_rpc/submit_transaction.rs
@@ -48,35 +48,40 @@ pub async fn handle<T: BlockchainBackend + 'static>(
             anyhow::anyhow!("Failed to get tip info: {e}")
         })?
         .is_synced;
-    let res = match mempool_service.submit_transaction(transaction).await {
-        Ok(response) => {
+    let res = match mempool_service.submit_transaction_detailed(transaction).await {
+        Ok((response, detail)) => {
             debug!(target: LOG_TARGET, "Transaction submitted successfully: {response:?}");
             match response {
                 TxStorageResponse::UnconfirmedPool => TxSubmissionResponse {
                     accepted: true,
                     rejection_reason: TxSubmissionRejectionReason::None,
                     is_synced,
+                    rejection_detail: None,
                 },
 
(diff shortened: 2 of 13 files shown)


base_layer/transaction_components/src/rpc/models/tx_submission_response.rs
```diff
@@ -29,6 +29,8 @@ pub struct TxSubmissionResponse {
     pub accepted: bool,
     pub rejection_reason: TxSubmissionRejectionReason,
     pub is_synced: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rejection_detail: Option<String>,
```
SWvheerden, line 33: wont work with older wallets, look at how the other https request handle version changes

SWvheerden (changes requested):
we should not add a new path, just update the http request with a new version to include the extra detail

SWvheerden:
closing this in favour of #7811

PR #7821, opened 2026-05-10
fix: propagate mempool rejection details to wallet and gRPC (closes #7776)

### LINKED ISSUE #7776: Better feedback of rejected transactions

watcharaponthod-code:
Closes #7776

When the mempool rejects a transaction, wallets previously only received a generic cancellation reason with no actionable information. This change threads the specific rejection message end-to-end: from the mempool through the HTTP JSON-RPC layer, through the wallet broadcast protocol, persisted in the wallet database, and surfaced in the gRPC `TransactionInfo` response.

## What changed

**Core – `TxStorageResponse` enum**

`NotStoredConsensus` and `NotStored` now carry an `Option<String>` with the exact rejection reason (e.g. "double-spend of commitment X", "invalid range proof"). The `Display` impl includes the reason when present so log output is immediately useful.

**Mempool storage**

Validation and consensus errors are forwarded into the new payload rather than being swallowed.

**HTTP JSON-RPC handler (`minotari_node`)**

`TxSubmissionResponse` gains a `rejection_detail: Option<String>` field. It is annotated `#[serde(skip_serializing_if = "Option::is_none")]` so the field is absent from the JSON for older wallets — fully backward-compatible. Six unit tests cover the detail propagation and the serialisation behaviour.

**Wallet broadcast protocol**

The `cancel_pending_transaction_with_detail` helper carries the `rejection_detail` string from the HTTP response into storage when the node rejects a submission.

**Wallet database / SQLite**

`CompletedTransaction` gains a `rejection_detail: Option<String>` field. A Diesel migration (`2026-05-10-120000_add_rejection_detail`) adds a nullable `rejection_detail` column to `completed_transactions` with a matching down migration that recreates the table without the column (required for SQLite compatibility).

`reject_completed_transaction` is updated throughout the storage layer to accept and persist the detail string.

**gRPC wallet server**

`TransactionInfo.rejected_reason` now uses the specific `rejection_detail` string when available, falling back to the cancellation reason enum as before.

**Integration test**

`test_rejection_detail_stored_on_mempool_rejection` funds a wallet with a UTXO, configures the HTTP base-node mock to return a rejection carrying `rejection_detail: Some("double-spend of commitment abc123")`, submits a coin-split transaction, waits for a `TransactionCancelled` event, and asserts the detail string is persisted on the `CompletedTransaction` record.

## Acceptance criteria

- [x] Wallet receives the specific rejection reason from the node
- [x] Reason identifies the exact failure (not just a generic code)
- [x] Backward-compatible — `rejection_detail` absent from JSON when `None`
- [x] At least one integration test covering the full path

### DIFF
--- a/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
+++ b/applications/minotari_console_wallet/src/grpc/wallet_grpc_server.rs
@@ -1694,7 +1694,9 @@ impl wallet_server::Wallet for WalletGrpcServer {
                             .into_iter()
                             .map(|pr| pr.to_vec())
                             .collect(),
-                        rejected_reason: txn.cancelled.map(|r| r.to_string()).unwrap_or_default(),
+                        rejected_reason: txn.rejection_detail.clone()
+                            .or_else(|| txn.cancelled.map(|r| r.to_string()))
+                            .unwrap_or_default(),
                         lock_height: txn.lock_height,
                     }),
                 };
@@ -1997,7 +1999,9 @@ impl wallet_server::Wallet for WalletGrpcServer {
                             .into_iter()
                             .map(|pr| pr.to_vec())
                             .collect(),
-                        rejected_reason: txn.cancelled.map(|r| r.to_string()).unwrap_or_default(),
+                        rejected_reason: txn.rejection_detail.clone()
+                            .or_else(|| txn.cancelled.map(|r| r.to_string()))
+                            .unwrap_or_default(),
                         lock_height: txn.lock_height,
                     };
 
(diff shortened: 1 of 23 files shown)


SWvheerden:
closing in favour of #7811

PR #7865, opened 2026-05-29
feat: change connection handling

SWvheerden:
Description
---
Changes connection handling to use weak and strong ref so that we dont disconnect used peers

### DIFF
--- a/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
+++ b/base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
@@ -28,7 +28,13 @@ use std::{
 
 use futures::StreamExt;
 use log::*;
-use tari_comms::{PeerConnection, connectivity::ConnectivityRequester, peer_manager::NodeId, protocol::rpc::RpcClient};
+use tari_comms::{
+    PeerConnection,
+    RefKind,
+    connectivity::ConnectivityRequester,
+    peer_manager::NodeId,
+    protocol::rpc::RpcClient,
+};
 use tari_node_components::blocks::{Block, ChainBlock};
 use tari_transaction_components::{BanPeriod, aggregated_body::AggregateBody};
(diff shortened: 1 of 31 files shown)


comms/dht/src/connectivity/mod.rs
```diff
@@ -416,10 +416,21 @@ impl DhtConnectivity {
...
+                debug!(
+                    target: LOG_TARGET,
+                    "Skipping disconnect of peer '{}' because it is strongly held",
+                    connected[index].short_str()
+                );
+                continue;
+            }
             let disconnect_peer = connected.swap_remove(index);
             disconnected_peers.push(disconnect_peer);
         }
```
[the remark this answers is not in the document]
SWvheerden: this is not required to disconnect peers, this is purely so we eventually swap peers and make sure we dont form islands

base_layer/core/src/base_node/sync/block_sync/synchronizer.rs
```diff
@@ -94,26 +100,42 @@ impl<'a, B: BlockchainBackend + 'static> BlockSynchronizer<'a, B> {
...
+        for peer in self.sync_peers.iter_mut() {
+            if let Some(conn) = peer.connection() &&
+                !conn.is_connected()
+            {
+                peer.clear_connection();
+            }
+            if peer.ensure_strong_connection() {
+                continue;
+            }
+            match self
```
sdbondi, line 117: Why not always return a strong connection and then the caller can `.downgrade` as needed instead of passing another concept into dial peer?

sdbondi: And we should leverage the type system, like weak pointers for Arc/Rc do

```rust
let strong = dial(..).await?;
let weak: PeerConectionWeak = strong.downgrade();
```

sdbondi (approved):

PR #7867, opened 2026-06-02
feat: add synced gRPC healthcheck

ns212:
This pull request adds support for gRPC health checking to the Minotari node. The main changes introduce a health reporter that tracks the node's state via the state machine and updates the gRPC health status accordingly. This allows external systems to check the node's readiness and health using the standard gRPC health check protocol.

**gRPC Health Checking Integration:**

* Added the `tonic-health` dependency to `Cargo.toml` for gRPC health checking support.
* Updated the `run_grpc` function in `lib.rs` to optionally accept a `StateMachineHandle`, set up a `HealthReporter`, and add the gRPC health service to the server. The health status is updated based on the node's state. [[1]](diffhunk://#diff-16fa446fb226d326cfbacc0938b31c552672cb25fb555d2e48b93b5c8ea2f3f6R346) [[2]](diffhunk://#diff-16fa446fb226d326cfbacc0938b31c552672cb25fb555d2e48b93b5c8ea2f3f6R366-R383)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ permissions: {}
 env:
   toolchain: stable
   toolchain_msrv_reject_older: 1.93.0
-  nightly_toolchain: nightly-2025-06-24
+  nightly_toolchain: nightly-2026-06-01
   CARGO_HTTP_MULTIPLEXING: false
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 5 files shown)


sdbondi (approved):
LGTM - just to check, a failing health check does not:

- stop serving grpc
- kill the node

It is solely for clients to check the sync status of the node (and these clients will expect the health check to fail intermittently from time to time?)

PR #7870, opened 2026-06-03
fix(base_node): reject duplicate validator node registrations within a block

sdbondi:
Description
---
Adds a chain-linked validation rule (`verify_no_duplicate_validator_node_registrations`) that rejects a block body containing more than one validator node registration for the same `(sidechain id, validator node public key)`, with a new `DuplicateValidatorNodeRegistration` validation error.

Motivation and Context
---
A validator node registration UTXO is only checked against the validator node set as of the **parent** block (`check_validator_node_registration` → `validator_node_exists`). That check cannot see other registrations in the **same** block body, and `contains_duplicated_outputs` only compares whole-output equality. So two outputs registering the same validator node public key (with different commitments) both pass validation.

They then collide at commit time: the validator node set is keyed by sidechain id + public key and inserted with LMDB `NO_OVERWRITE` (`ValidatorNodeStore::insert`), so the second insert returns `KeyExists` and the block is aborted during the DB write (`reorganize_chain` restores the chain and returns an error — see the "this should typically never happen unless there is a bug in the validator" comment there) instead of being rejected cleanly during validation.

This makes the invariant ("at most one active registration per validator node") an explicit validation rule rather than an accidental side effect of the storage layer. The new error is classified as a long ban, so an offending block is recorded as bad rather than silently retried. The check runs before commit on both the block path (`BlockBodyFullValidator::validate`) and the mempool transaction path (`TransactionChainLinkedValidator`).

Note: this does not change the canonical chain — such blocks were already uncommittable; it converts a commit-time abort into a clean validation rejection.

How Has This Been Tested?
---
- New unit tests: distinct registrations pass; two registrations for the same validator node are rejected with `DuplicateValidatorNodeRegistration`.
- `cargo test -p tari_core --lib validation::` — 47 passed, 0 failed.
- `cargo clippy -p tari_core --lib` — no new warnings.

---
Run `cargo test -p tari_core --lib validation::aggregate_body::aggregate_body_chain_validator`. To verify the gap manually, construct an `AggregateBody` with two `OutputType::ValidatorNodeRegistration` outputs that share a validator node public key but have different commitments, and confirm `verify_no_duplicate_validator_node_registrations` returns `DuplicateValidatorNodeRegistration`.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/base_layer/core/src/validation/aggregate_body/aggregate_body_chain_validator.rs
+++ b/base_layer/core/src/validation/aggregate_body/aggregate_body_chain_validator.rs
@@ -104,6 +104,7 @@ impl AggregateBodyChainLinkedValidator {
         check_inputs_are_spendable(db, constants, header.height, &body)?;
         check_outputs(db, constants, &body, header.height)?;
         verify_no_duplicated_inputs_outputs(&body)?;
+        verify_no_duplicate_validator_node_registrations(&body)?;
         check_total_burned(&body)?;
         verify_timelocks(&body, header.height)?;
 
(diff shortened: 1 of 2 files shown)


SWvheerden (dismissed):
looks good, just needs a clippy

PR #7864, opened 2026-05-29
feat(installer): add unified Minotari Ledger installer

tzmWW:
Fixes #7795

## Summary
- Add one cross-platform Python installer for the Minotari Ledger Wallet app.
- Detect supported Ledger devices with `ledgerwallet` target IDs and report the original Nano S as unsupported.
- Resolve a matching Tari release artifact, require the `.zip.sha256` sidecar, verify it, and safely extract the archive.
- Install current `minotari_ledger_wallet.apdu` artifacts through `ledgerblue.runScript --scp`; legacy manifest fallback is intentionally not supported.
- Keep the existing per-model scripts as thin compatibility wrappers around the unified auto-detecting installer.
- Stage the unified installer and wrappers as release assets so downloaded wrapper assets remain runnable.

## Correctness Note
The current stable Tari release may not include Ledger wallet assets, while pre-releases do. The default lookup scans recent non-draft releases, including pre-releases, and only selects a Ledger artifact when both the model zip and matching checksum sidecar are present.

## Testing
- `uv run python applications/minotari_ledger_wallet/wallet/install_scripts/test_install_minotari_ledger.py`
- `uv run python applications/minotari_ledger_wallet/wallet/install_scripts/install_minotari_ledger.py --help`
- `uv run python -m py_compile applications/minotari_ledger_wallet/wallet/install_scripts/install_minotari_ledger.py applications/minotari_ledger_wallet/wallet/install_scripts/test_install_minotari_ledger.py`
- `uvx ruff check applications/minotari_ledger_wallet/wallet/install_scripts/install_minotari_ledger.py applications/minotari_ledger_wallet/wallet/install_scripts/test_install_minotari_ledger.py`
- Repo-layout wrapper `--help` smoke test
- Flat release-asset wrapper `--help` smoke test
- `git diff --check`
- Real Nano S Plus hardware install succeeded after updating device firmware to a compatible Ledger OS version.

PR #7872, opened 2026-06-06
Add unified Minotari Ledger installer

Peter7896:
Fixes #7795

## Summary
- add a unified cross-platform Minotari Ledger installer for Nano S Plus, Nano X, Stax, and Flex
- auto-detect a single connected Ledger device over USB, with explicit --model override for multi-device or manual flows
- resolve the newest non-draft Tari release that contains the matching Ledger wallet asset, verify the .sha256 sidecar, and safely extract the archive
- install the downloaded .apdu script through ledgerblue.runScript --scp, with model target IDs when no ELF file is present
- keep the legacy model-specific shell and PowerShell scripts as thin compatibility wrappers

## Validation
- python -m py_compile applications\minotari_ledger_wallet\wallet\install_scripts\install_minotari_ledger.py applications\minotari_ledger_wallet\wallet\install_scripts\test_install_minotari_ledger.py
- python applications\minotari_ledger_wallet\wallet\install_scripts\test_install_minotari_ledger.py
- python applications\minotari_ledger_wallet\wallet\install_scripts\install_minotari_ledger.py --help
- powershell -NoProfile -ExecutionPolicy Bypass -File applications\minotari_ledger_wallet\wallet\install_scripts\nanosplus\install_ledger_win.ps1 --help
- python applications\minotari_ledger_wallet\wallet\install_scripts\install_minotari_ledger.py --model nanosplus --dry-run --no-bootstrap --download-dir .ledger-installer-dryrun
- git diff --check


Peter7896:
Updated the branch after review:

- skipped bootstrap pip work when `ledgerblue` is already importable in the installer venv
- cleaned up implicit temporary download directories after install and dry-run flows
- added tests for both paths
- refreshed the commit as a verified SSH-signed commit; signed-commits and GitGuardian checks are passing

Validation:
- `python -m py_compile applications\minotari_ledger_wallet\wallet\install_scripts\install_minotari_ledger.py applications\minotari_ledger_wallet\wallet\install_scripts\test_install_minotari_ledger.py`
- `python applications\minotari_ledger_wallet\wallet\install_scripts\test_install_minotari_ledger.py`, 15 passed
- `git diff --check`

SWvheerden:
🔴 1. USB auto-detection won't fire for a device on the dashboard (the primary, documented flow)

  The PID table stores the wrong kind of value:

  "nanosplus": LedgerModel("nanosplus", "Ledger Nano S Plus", "0x33100004", (0x50, 0x0005)),
  "nanox":     LedgerModel("nanox",     "Ledger Nano X",      "0x33000004", (0x40, 0x0004)),
  "stax":      LedgerModel("stax",      "Ledger Stax",        "0x33200004", (0x60, 0x0006)),
  "flex":      LedgerModel("flex",      "Ledger Flex",        "0x33300004", (0x70, 0x0007)),

  Detection matches the device's full 16-bit product ID against these values (PID_TO_MODEL.get(pid)). But Ledger's application-mode
  PID scheme is (model_tag << 8) | interface_flags, where the model tag is the high byte: 0x50=Nano S Plus, 0x40=Nano X, 0x60=Stax,
  0x70=Flex. A real device sitting on the dashboard reports e.g. 0x5011, not 0x0050. So:

  - 0x50 (=0x0050) only matches a device literally reporting PID 0x0050 — no standard Ledger does.
  - 0x0005 etc. are the bootloader/recovery PIDs — only present when the device is not on the dashboard.

  The README instructs the user to "leave it on the dashboard," which is exactly the state where lsusb/PnP/system_profiler report
  0x50xx/0x40xx/… and model_from_pid() returns None → select_model(None) raises "Could not auto-detect." This affects all three OS
  paths.

  The tests bake in the wrong values (2c97:0050, PID_0007, product_id 0x0060), so they pass while masking the gap.

  Fix: compare the high byte for the new scheme, keep the legacy PIDs for bootloader mode. Roughly:
  def model_from_pid(value):
      pid = normalize_pid(value)
      if pid is None: return None
      return PID_TO_MODEL.get(pid) or PID_TO_MODEL.get(pid >> 8)  # 0x5011 -> 0x50
  and store model tags as 0x50/0x40/0x60/0x70 plus legacy 0x0005/0x0004/0x0006/0x0007.

  Mitigation: the --model override works, and every legacy wrapper passes --model explicitly — so wrapper-invoked installs are
  unaffected. Only the bare python3 install_minotari_ledger.py path (the one the new top-level README leads with) is broken. Please
  verify against a real device before merge.

  🟠 2. --apdu flag dropped from the runScript invocation

  Every legacy script ran:
  ledgerblue.runScript --targetId 0x… --fileName …apdu --apdu --scp
  build_install_command emits --scp --fileName … (--targetId|--elfFile) but no --apdu. If --apdu affected how runScript interprets the
  file/secure-channel (not just logging), this is a silent install regression on real hardware. Please confirm the removal is
  intentional and state why in the PR — this is the one step the unit tests can't exercise.

  🟡 3. Missing .sha256 sidecar hard-fails instead of falling back

  resolve_release_asset accepts the first release whose .zip matches, even if find_asset_pair found no checksum (checksum_url=None).
  install() then raises Missing checksum sidecar. A release with the zip but no sidecar aborts the whole run rather than continuing to
  an older release that has both. Consider requiring the checksum in the selection predicate so the loop skips such releases.

  🟡 4. macOS/Windows PID strings with trailing text silently drop

  normalize_pid does int(value, 16) after stripping 0x. system_profiler sometimes emits vendor_id = "0x2c97 (Ledger)"; that (and any
  decorated product_id) fails to parse → None → device skipped. Ext

SWvheerden:
Closing in favour of #7864

PR #7875, opened 2026-06-09
feat: unified uv installer to replace all os-specific scripts (fixes #7795)

Ishant5436:
## Summary

Replaces OS-specific ledger installation scripts with a unified cross-platform installer using `uv`. Removes the requirement for local Python/pip or virtualenv setup by executing in an isolated ephemeral environment.

## Changes

- `scripts/install_ledger.sh`: added unified installer using `uv run`
- Removed legacy OS-specific installation folders and scripts
- Added automatic detection and support for Stax, Flex, Nano X, and Nano S+ devices

## Testing

- Validated `uv run` execution path on Windows and Unix-based environments
- Confirmed ephemeral environment isolation without system Python dependencies

Closes #7795


Ishant5436:
I have updated the PR to fix the `nanos+` mapping mismatch and added timeouts/cleanups as suggested. I also signed the commits with my SSH key. Please let me know if anything else is needed!

Ishant5436:
My Tari Emoji Address for bounty payout: 🍗📟🍬🔫🐞🍊🐙🎢🐢👑🚗🍯🍯🎸🐘👙🐬🦂🎲🐭🎼🍵👙💄📿🍾🎂🍞🔦💡👠🤠👍🎢🏰🍬🔒💊🚓💻🎋🐪😇🦋🔫➕🔮🎩💩😱🚪🐘🎩👞🦁🎾🎁💩😱🌰🐷💤💩👻🏥🍀👘

SWvheerden:
🔴 Blocking issues
  
  1. ledgerctl is not a real PyPI package — dependency resolution will fail

  The PEP 723 header declares:
  # dependencies = [ "ledgerwallet", "ledgerctl", "requests" ]
  ledgerctl does not exist on PyPI (https://pypi.org/pypi/ledgerctl/json → 404). The ledgerctl command is the console-script entry
  point provided by the ledgerwallet package itself. So uv run will fail to resolve dependencies and the installer never starts. The
  old scripts' pip install ledgerctl line was already wrong/redundant; this faithfully reproduces that mistake.
  Fix: drop "ledgerctl" from the dependency list — ledgerwallet already installs the ledgerctl binary. 
  
  2. Ledger USB product-ID map swaps Nano X and Nano S Plus

  Verified against LedgerHQ's official device table (productIdMM high byte / legacyUsbProductId):

  ┌─────────────┬────────────────┬────────────────┐
  │   Device    │ Real high byte │ Real legacy ID │
  ├─────────────┼────────────────┼────────────────┤
  │ Nano X      │ 0x40           │ 0x0004         │
  ├─────────────┼────────────────┼────────────────┤
  │ Nano S Plus │ 0x50           │ 0x0005         │
  ├─────────────┼────────────────┼────────────────┤
  │ Stax        │ 0x60           │ 0x0006         │
  ├─────────────┼────────────────┼────────────────┤
  │ Flex        │ 0x70           │ 0x0007         │
  └─────────────┴────────────────┴────────────────┘

  The script maps these incorrectly and inconsistently:
  0x0004: "nanosplus",   # WRONG — 0x0004 is Nano X
  0x4000: "nanosplus",   # WRONG — 0x40xx is Nano X
  0x5011: "nanosplus",   # ok    — 0x50xx is Nano S Plus
  0x0005: "nanox",       # WRONG — 0x0005 is Nano S Plus
  0x5000: "nanox",       # WRONG — 0x50xx is Nano S Plus
  0x5015: "nanox",       # WRONG — 0x50xx is Nano S Plus
  0x4005: "nanox",       # ok    — 0x40xx is Nano X
  For the two most common devices, this downloads and attempts to flash the wrong firmware (-nanox.zip onto a Nano S Plus and
  vice-versa). Stax/Flex happen to be correct.
  
  Recommended fix: don't hand-maintain this table at all. Derive the model from the high byte and use the library's own model enum,
  e.g.:
  model_by_high_byte = {0x40: "nanox", 0x50: "nanosplus", 0x60: "stax", 0x70: "flex"}
  slug = model_by_high_byte.get(device.product_id >> 12)  # or >> 8 & 0xF0 logic
  or better, read the model name directly off the ledgerwallet device object instead of re-deriving it. Either way, please verify
  against a physical Nano X and Nano S Plus before merge.

  ---
  🟠 Functional regressions & correctness
  
  - --tag / pre-release support dropped. Every old script accepted -t/--tag to install a specific release (explicitly "including
  pre-releases"). The new script hardcodes releases/latest with no override. This matters here: releases/latest returns the latest
  non-prerelease GitHub release, but Tari ships mostly pre-releases (current dev tag is v5.4.0-pre.3). Users may silently get stale
  firmware, or the lookup may not find a matching asset at all. Please restore a tag/pre-release option.
  - install.sh sources the wrong env file. After installing uv it does . "$HOME/.cargo/env". uv's standalone installer installs to
  ~/.local/bin and writes ~/.local/bin/env, not ~/.cargo/env (that's rustup's). So on a machine without uv, the freshly-installed uv
  won't be on PATH and the subsequent uv run fails. Source "$HOME/.local/bin/env" instead.
  - Loose app_*.json matching. download_and_extract walks the archive for any file where "app_" in file and file.endswith(".json")
  and takes the first hit, rather than the model-specific app_{slug}.json the old scripts targeted. The per-slug zip likely contains
  only one, but the loose match + arbitrary os.walk ordering is fragile; match f"app_{slug}.json" explicitly.
  - device.vendor_id / device.product_id attribute access is unverified. ledgerwallet.transport.enumerate_devices() returns
  transport objects; please confirm they actually expose .vendor_id/.product_id (HID vs. these attributes vary by backend) or the
  script will AttributeError immediately. Worth an actual run on hardware.
  - Multiple-device handling: silently picks devices[0]. Minor, but a one-line warning when len(devices) > 1 would help.

  🟡 Minor / style

  - time.sleep(1) before install is pointless — drop it.
  - install.ps1 rebuilds $env:Path from Machine+User registry only, discarding the current process PATH. Usually fine right after a
  uv install, but it's a footgun.
  - No integrity/signature check on the downloaded zip. Not a regression (old scripts were the same), and Ledger requires on-device
  confirmation + app signing, but worth a note in the README's removed "Security" section, which this PR dropped entirely.
  - README claims the script "securely flashes" firmware — consider toning down given there's no checksum verification.

  Testing

  The PR body says "tested locally and works beautifully across OSes," but issues #1 and #2 above mean it could not have succeeded
  as written for a Nano X/S Plus, and the ledgerctl dep would block uv run on a clean machine. I'd ask the author which device/OS
  combination was actually exercised, ideally with a terminal capture.

Ishant5436:
hey, sorry about that, i accidentally closed this while cleaning up my branches! everything is actually fully working locally. re-opening to finish this up and address any feedback.

Ishant5436:
just pushed an update addressing all the feedback!
- dropped `ledgerctl` from the uv dependencies since `ledgerwallet` handles it.
- completely rewrote the device mapping to use the high byte (0x40/0x50/0x60/0x70) as requested so Nano X and Nano S Plus are mapped accurately without maintaining a massive list.
- added back `-t/--tag` support. if not provided, it fetches the absolute latest (which includes pre-releases).
- `install.sh` now sources `~/.local/bin/env` properly.
- `install.ps1` safely appends to the process PATH instead of rebuilding from the registry to avoid footguns.
- fixed the loose `app_*.json` match to explicitly look for `app_{slug}.json`.
- added a warning if multiple ledger devices are connected.
- removed the unnecessary sleep.

thanks for the incredibly detailed review! let me know if anything else catches your eye.

SWvheerden (changes requested):
Look at the new scripts, using json and ledgerctl does not work anymore

Ishant5436:
Ah, I see! Ledger moved to `.apdu` scripts. I have completely refactored the unified Python installer to support this modern format:
- Added `ledgerblue` to dependencies.
- It now properly extracts `minotari_ledger_wallet.apdu` from the zip.
- Maps devices to their secure channel `targetId` (Nano X: `0x33000004`, Nano S Plus: `0x33100004`, etc.).
- Executes `ledgerctl delete` for cleanup, followed by `ledgerblue.runScript` to flash the new firmware securely.

Ready for another look!

Ishant5436:
Hi there! Just checking in on this—is there anything else needed from my side before this can be reviewed and merged? Thanks!

SWvheerden:
I do like the uv approach of this, and I will cherry-pick this out of here, but #7864 is the more complete option

SWvheerden:
https://github.com/tari-project/tari/pull/7922

SWvheerden:
@Ishant5436 can you provide us a tari address for a bounty

PR #7888, opened 2026-06-18
feat: harden block add

SWvheerden:
Description
---
harden block add with additional safety checks

### DIFF
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -2068,7 +2068,9 @@ impl LMDBDatabase {
                 block_hash.to_hex()
             )));
         }
-        if header.height > 0 {
+        // We make this 1 to circumvent networks thats dont have genesis funds, as they will only have a jmt root from
+        // height 1
+        if header.height > 1 {
             let k = MetadataKey::JMTVersion;
             let current_jmt_version = match lmdb_get(txn, &self.metadata_db, &k.as_u32())? {
                 Some(MetadataValue::JMTVersion(v)) => v,
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -2074,15 +2074,18 @@ impl LMDBDatabase {
                 Some(MetadataValue::JMTVersion(v)) => v,
                 _ => 0u64,
             };
-            let root = output_smt.get_root_hash(current_jmt_version).map_err(ChainStorageError::JellyfishMerkleTreeError)?;
+            let root = output_smt
+                .get_root_hash(current_jmt_version)
+                .map_err(ChainStorageError::JellyfishMerkleTreeError)?;
             let prev_header = lmdb_get::<_, BlockHeader>(txn, &self.headers_db, &(header.height - 1)).or_not_found(
                 "BlockHeader",
                 "height",
-                (header.height-1).to_string(),
+                (header.height - 1).to_string(),
             )?;
             if prev_header.output_mr.as_slice() != root.0.as_slice() {
                 return Err(ChainStorageError::InvalidOperation(format!(
-                    "The output merkle root of the current tip header at height {} does not match the stored JMT root. Header: {}, calculated: {}",
+                    "The output merkle root of the current tip header at height {} does not match the stored JMT \
+                     root. Header: {}, calculated: {}",
                     prev_header.height,
                     prev_header.output_mr.to_hex(),
                     hex::encode(root.0.as_slice())
@@ -5856,6 +5859,7 @@ fn num_to_key(n: u32) -> Option<MetadataKey> {
         10 => Some(MetadataKey::AccumulatedDataCheckStatus),
         11 => Some(MetadataKey::BlockchainConsistencyCheckStatus),
         12 => Some(MetadataKey::HorizonSyncOutputCheckpoint),
+        13 => Some(MetadataKey::JMTVersion),
         _ => None,
     }
 }
--- a/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
+++ b/base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs
@@ -2068,6 +2068,27 @@ impl LMDBDatabase {
                 block_hash.to_hex()
             )));
         }
+        if header.height > 0 {
+            let k = MetadataKey::JMTVersion;
+            let current_jmt_version = match lmdb_get(txn, &self.metadata_db, &k.as_u32())? {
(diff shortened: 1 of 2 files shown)


sdbondi (approved):

sdbondi (changes requested):
I get this error when running a fresh db

Failed to apply DB transaction: Root node not found for version 0.

sdbondi (commented):
## Review — `feat: harden block add`

**What it does:** (1) a pre-flight check in `insert_tip_block_body` (height > 1) that recomputes the JMT root at the stored `JMTVersion` and compares it to the previous header's `output_mr`; (2) a rewrite of `write_node_batch` that derives the saved `JMTVersion` from the batch's max node version (skipping empty batches) and warns on desync; plus registering `JMTVersion` in `num_to_key` and a new test.

**Overall:** The new pre-flight check is sound and a useful fail-fast guard — it stays consistent across reorgs, horizon sync, pruning, and genesis-rebuild because it anchors to the *actual* stored version rather than deriving it from height. I don't see a production-reachable correctness regression for healthy nodes. But the writer rewrite doesn't do what its comments claim, the new test doesn't exercise its named path, and there's a latent false-rejection edge. Details below, most significant first.

> Verification note: I read the `jmt-0.11.0` internals (`tree_cache::freeze`, `tree::put_value_set`/`get_root_hash`) and ran the new test to ground these claims rather than reasoning from the diff alone.

---

### 1. The `write_node_batch` rewrite is functionally inert — it does not fix the desync it documents
`lmdb_tree_writer.rs:114-138`

The new code persists `written_version = node_batch.nodes().keys().map(version).max()` only `if let Some(...)`. Tracing through `jmt-0.11.0`:

- `put_value_set` → `put_value_sets` → `TreeCache::freeze`. `freeze` **always emits a root node** at the target version: it copies the previous root forward when nothing changed (`tree_cache.rs:316-323`) and seeds a null genesis root for version 0 (`tree_cache.rs:170`). So `node_batch.nodes()` is **never empty** and `.max()` is **always `Some`**.
- Every `LmdbTreeWriter::write_node_batch` caller (`lmdb_db.rs:2226`, `:1926`, `:2649`) passes `put_value_set` output where all new nodes carry version `current+1`. So `written_version == current+1 ==` the old code's `val` in **every reachable case**.

Net effect: the persisted `JMTVersion` is identical to before; the only functional addition is the warning. The `if let Some(written_version)` skip-on-empty-batch branch is **dead code**, and the comments ("an empty batch does NOT bump the version") describe a path that never executes. If a genuine desync exists in production, this writer change neither prevents nor repairs it — all actual protection comes from the new check in #3, which *errors* on a desync rather than recovering. Could you confirm what real-world desync motivated this, and whether the inert writer change is intended?

### 2. The new test doesn't exercise the path it's named for
`lmdb_tree_writer.rs:267-319`

`test_jmt_version_noop_batch_desync` deletes a non-existent key, which is **not** a no-op — it re-roots the tree. Running it:

```
no-op batch produced 1 node(s), 1 value(s)
after NO-OP batch, JMTVersion = 1
get_root_hash(JMTVersion=1) = Ok(...)
```

It produces 1 node, bumps `JMTVersion` 0→1, and `get_root_hash(1)` succeeds — **identically under the old code** (old `val = current+1 = 1`, root at 1 exists). The assertion `root_at_v1.is_ok()` passes regardless of the fix, so the test guards nothing. Exercising the skip branch would require an empty / `PutResult::NotChanged` batch, which per #1 `put_value_set` can't produce.

### 3. Latent false-rejection: a missing `JMTVersion` defaults to 0, then `get_root_hash(0)` hard-errors
`lmdb_db.rs:2075-2081`

```rust
let current_jmt_version = match lmdb_get(...) { Some(JMTVersion(v)) => v, _ => 0u64 };
let root = output_smt.get_root_hash(current_jmt_version).map_err(...)?;
```

`jmt::get_root_hash` returns a hard `MissingRootError` (`tree.rs:1395-1407`) when no root node exists at that exact version. On a long-running node, version 0's root has long been stale-pruned by `cleanup_stale`. So if `JMTVersion` is ever absent, this defaults to 0 and `get_root_hash(0)` errors → a **valid block is rejected** and tip import stalls. The corrupt-metadata recovery classifies `JMTVersion` as non-essential and auto-deletes it (`lmdb_db.rs:5809-5814`), yet this check makes it load-bearing for block insertion — a contradiction. Reachability is low (deletion only on a deserialize failure of that u64), but the safe fix is to **skip the check when the version is absent** (as it already does for `height <= 1`) rather than default to 0 and fail.

### 4. Misleading error message identity
`lmdb_db.rs:2088-2095`

The message says "The output merkle root of the **current tip header** at height {}" but interpolates `prev_header.height` (= `header.height - 1`) and `prev_header.output_mr`. Everything printed describes the **previous** header, not the block being inserted, which will misdirect diagnosis. Suggest wording like "previous header (height N)".

---

### Minor / notes

- **Warn-and-continue on detected desync** (`lmdb_tree_writer.rs:124-130`): the writer detects `written_version != expected` but only `warn!`s and commits anyway. Moot today (branch is inert per #1), but if it ever fired it would persist a known-mismatched state and surface later as an opaque root mismatch. Consider erroring, or at least not advancing the version.
- **Duplication** (`lmdb_db.rs:1896, 2075, 2193, 2640, 2978, 3877` + `lmdb_tree_writer.rs:119`): the `match lmdb_get(...JMTVersion...) { Some(JMTVersion(v)) => v(+1), _ => 0 }` idiom is now repeated at 7 sites with no `fetch_jmt_version(txn)` helper; the +1-vs-no-bump rule lives nowhere shared.
- **`num_to_key` addition is correct**: enum discriminant ordering makes `as_u32(JMTVersion) == 13 == num_to_key(13)`. Before this, a `JMTVersion` key fell through to the "(unknown) → auto-delete" branch, so registering it is a real improvement.
- **Altitude**: the check is a special-case guard bolted onto the already-`#[allow(too_many_lines)]` `insert_tip_block_body`; the `height > 1` carve-out for unfunded-genesis networks is a network-shape constant. The "persisted-tree-vs-header" invariant could live in a shared validation layer reused by the sync/reorg/horizon paths that have no equivalent guard.
- **Latent, not production-reachable**: if the output SMT ever fully empties, `freeze` produces a `Node::Null` root which `write_node_batch` *deletes* (lines 78-81) while still bumping `written_version` to that version → `get_root_hash` would then error. Can't happen on a real chain (always-unspent coinbase outputs), and the old code behaved the same; noting for completeness.

**Bottom line:** Correctness looks safe for healthy nodes and the new check is a good guard, but #1 and #2 are worth addressing before merge — the writer change and its test don't accomplish what they're written to do, which undermines the "harden" intent. #3 and #4 are easy, worthwhile hardening of the new check itself.

<sub>🤖 Review assisted by Claude Code (high-recall multi-pass), grounded by reading the `jmt` source and running the added test.</sub>

PR #7915, opened 2026-07-04
chore(deps): update cucumber junit path

samrusani:
Description
---
Updates the integration-test Cucumber/JUnit dependency path without keeping a global `junit-report` patch override.

This bumps `cucumber` from `0.22.1` to `0.23.0` and lets Cargo resolve crates.io `junit-report 0.9.0`. That moves the JUnit output path from:

```text
cucumber 0.22.1 -> junit-report 0.8.3 -> quick-xml 0.31.0
```

to:

```text
cucumber 0.23.0 -> junit-report 0.9.0 -> quick-xml 0.39.4
```

Refs #7912.
Refs #7914.

Motivation and Context
---
Issues #7912 and #7914 report RustSec advisories against `quick-xml 0.31.0`. In this workspace that dependency is pulled only through the integration-test JUnit output path:

```text
cucumber -> junit-report -> quick-xml
```

The integration workflow still uploads `integration_tests/cucumber-output-junit.xml`, so this keeps the `output-junit` feature instead of removing JUnit reporting.

Review note: crates.io `junit-report 0.9.0` still resolves `quick-xml 0.39.4`, which is below the advisory fixed range of `quick-xml >=0.41.0`. A previous revision temporarily patched `junit-report` to an upstream commit that used `quick-xml 0.41.0`, but that global override was removed after review because it would keep forcing `junit-report` for all dependants until a crates.io release catches up.

How Has This Been Tested?
---
- `cargo tree -i quick-xml --locked`
- `cargo tree -i crossbeam-epoch --locked`
- `cargo +1.93.0 test --no-run --locked --all-features --package tari_integration_tests --test cucumber`
- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `cargo audit`

Audit note: `cargo audit` still reports `RUSTSEC-2026-0194` and `RUSTSEC-2026-0195` for `quick-xml 0.39.4`. That residual finding is expected with the normal crates.io `junit-report 0.9.0` path and should be revisited when `junit-report` publishes a release that depends on `quick-xml >=0.41.0`.

---
- Run `cargo tree -i quick-xml --locked` and confirm the path resolves through crates.io `junit-report 0.9.0`.
- Run `cargo audit` and confirm the remaining quick-xml findings match the expected residual advisories above.
- Compile the Cucumber integration target with `cargo +1.93.0 test --no-run --locked --all-features --package tari_integration_tests --test cucumber`.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,3 +117,6 @@ debug = false
 
 [patch.crates-io]
 liblmdb-sys = { git = "https://github.com/tari-project/lmdb-rs", tag = "0.7.6-tari.1" }
+# crates.io junit-report 0.9.0 still depends on quick-xml 0.39, which is below
+# the RustSec-patched quick-xml 0.41 line used by upstream junit-report master.
+junit-report = { git = "https://github.com/bachp/junit-report-rs", rev = "d4eb4902e245a64d08110891069d3024f9db9d04" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,6 +117,3 @@ debug = false
 
 [patch.crates-io]
 liblmdb-sys = { git = "https://github.com/tari-project/lmdb-rs", tag = "0.7.6-tari.1" }
-# crates.io junit-report 0.9.0 still depends on quick-xml 0.39, which is below
-# the RustSec-patched quick-xml 0.41 line used by upstream junit-report master.
-junit-report = { git = "https://github.com/bachp/junit-report-rs", rev = "d4eb4902e245a64d08110891069d3024f9db9d04" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1501,9 +1501,9 @@ dependencies = [
 
 [[package]]
 name = "cucumber"
-version = "0.22.1"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "16cbb27bc2064274afa3a3d8bc9a0e71333589850573aa632ec4520e4af14d94"
+checksum = "96a87e18d925b19ebe0fd47ea45316abd216d81ec0879c2448c3f9a0e9da62be"
 dependencies = [
  "anyhow",
  "clap",
@@ -1531,9 +1531,9 @@ dependencies = [
 
 [[package]]
 name = "cucumber-codegen"
-version = "0.22.1"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a1afaf9c422380861111c6be56f39b324e351fd9efc07a1486268798bf79cfd"
+checksum = "ed2fc8a8bbb73af3230db699e8690c5c786655f75eb89e5f18d76055fa1a9a4d"
 dependencies = [
  "cucumber-expressions",
  "inflections",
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -117,3 +117,6 @@ debug = false
 
 [patch.crates-io]
 liblmdb-sys = { git = "https://github.com/tari-project/lmdb-rs", tag = "0.7.6-tari.1" }
+# crates.io junit-report 0.9.0 still depends on quick-xml 0.39, which is below
+# the RustSec-patched quick-xml 0.41 line used by upstream junit-report master.
+junit-report = { git = "https://github.com/bachp/junit-report-rs", rev = "d4eb4902e245a64d08110891069d3024f9db9d04" }
```
SWvheerden: This should not be necessary, the lock file should be good. 
This will keep overriding it, downgrade included.

SWvheerden (changes requested):

Cargo.toml
samrusani: Thanks, I checked this path again.

Removing the `junit-report` patch and regenerating the lockfile makes Cargo resolve crates.io `junit-report 0.9.0`, which still pulls `quick-xml 0.39.4`. With the current RustSec DB, that reintroduces both quick-xml advisories this PR is trying to close:

- `RUSTSEC-2026-0194`: fixed in `quick-xml >=0.41.0`
- `RUSTSEC-2026-0195`: fixed in `quick-xml >=0.41.0`

I also checked the normal release path: `junit-report 0.9.0` is still the latest crates.io release, and `cucumber 0.23.0` is still the latest `cucumber` release. The integration workflow also uploads `integration_tests/cucumber-output-junit.xml`, so dropping the `output-junit` feature would be a broader CI/reporting behavior change.

So I left the branch unchanged for now. The patch is currently the narrowest way I found to keep JUnit output while resolving quick-xml to `0.41.0`. If you prefer, I can instead remove JUnit output from this PR, but that changes the integration-test artifact path rather than just fixing the vulnerable dependency.

SWvheerden: I got you. 

But the actual vulnarability, where it sits in code, I think we can leave it unpatched for now. 
To exploit this, you would need to open up a pr with bad code, the code needs to get reviewed, and only then do we run CI, which would upload the report. 

When junit-report has a new version out, we can upgrade it, upgrading to cucumber we can still do,

samrusani: Updated this per your direction.

Changes pushed in `bb5b2466f`:
- removed the global `junit-report` `[patch.crates-io]` override
- kept the `cucumber 0.23.0` upgrade
- regenerated the lockfile through crates.io `junit-report 0.9.0`, which resolves `quick-xml 0.39.4`
- rebased on current `development`, so the merged `crossbeam-epoch 0.9.20` update is included in the branch base
- updated the PR title/body from `Fixes` to `Refs` for #7912/#7914 and documented the expected residual `quick-xml` audit findings

Validation:
- `cargo tree -i quick-xml --locked` -> `cucumber 0.23.0 -> junit-report 0.9.0 -> quick-xml 0.39.4`
- `cargo tree -i crossbeam-epoch --locked` -> `crossbeam-epoch 0.9.20`
- `cargo +1.93.0 test --no-run --locked --all-features --package tari_integration_tests --test cucumber`
- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `cargo audit` still reports the expected residual `quick-xml` advisories `RUSTSEC-2026-0194` and `RUSTSEC-2026-0195` on `quick-xml 0.39.4`.

SWvheerden (approved):

PR #7946, opened 2026-07-28
fix(comms): filter private DHT peer addresses

### LINKED ISSUE #7935: tari_comms_dht accepts loopback and private peer addresses when allow_test_addresses=false
## Bounty: Console Wallet Key Migration

**Tier:** S — 15,000 XTM

### Description

Thanks to @biran01 for finding this. 

tari_comms_dht accepts signed peer claims that advertise loopback, private RFC1918, link-local, and DNS-style internal addresses even when shipped defaults keep allow_test_addresses = false. 

### Acceptance Criteria

- If allow_test_addresses = false, node should reject all loopback and private RFC1918 addresses. 
- A peer advertising those addresses should still be accepted if they have valid other addresses, but only the loopback ones should be rejected
- Unit tests proving the filtering.

### How It Works

3. Submit a PR that references the issue with a closing keyword (e.g. `Fixes #7829`) so GitHub links your PR to the bounty automatically
4. Bounties are competitive. Multiple PRs may be submitted for the same bounty. Review comments are public, and contributors may borrow ideas from each other's work. The maintainer selects the PR that best solves the problem, considering code quality, completeness, and contributor behavior. In cases where multiple contributors made significant contributions, the maintainer may split the bounty. The maintainer's decision is final.
5. Earn XTM (processing may take up to a week; reach out to @possum on Discord with your Tari address)

### Notes

kamexETH:
Description
---

Filter disallowed local/test addresses out of signed peer identity claims after verifying the original claim, while retaining that claim for signature verification and address provenance.

- Apply the filtering path to connection handshakes as well as DHT peer validation.
- Accept mixed public/local claims and use only their permitted addresses for storage and dialing.
- Drop new claims with no usable addresses without issuing a ban.
- Preserve an existing peer's known usable addresses when an all-local update arrives.
- Reject IPv4 unspecified, loopback, RFC1918, link-local, multicast, broadcast, documentation, `0.0.0.0/8`, and reserved `240.0.0.0/4` addresses.
- Reject IPv6 unspecified, loopback, unique-local, link-local, multicast, documentation, and IPv4-mapped or IPv4-compatible local addresses.
- Reject DNS-style internal names such as `localhost`, `.local`, `.internal`, `.lan`, `.localdomain`, and `home.arpa`.
- Reuse the same classifier for validation and the peer database's external-address flag.
- Preserve the existing opt-in behavior when test addresses are allowed.

Fixes #7935

Motivation and Context
---

A valid signature proves that a peer authored its address claim, but it does not make loopback or private addresses useful to another node. Filtering therefore happens only after verifying the original, unmodified signed claim. This avoids storing or dialing unusable local addresses without discarding or banning a peer that also advertises a valid route.

Persistence and propagation scope
---

`is_external` is persisted in the peer database. This patch does not eagerly rewrite existing rows; their classification converges when an address is next inserted or updated.

The original signed claim remains unmodified and may continue to be gossiped because its signature covers the complete claim. Recipients filter the claim before storing or dialing its addresses.

How Has This Been Tested?
---

- `cargo test -p tari_comms connection_manager::common::test --lib`
  - 2 passed
- `cargo test -p tari_comms net_address::multiaddr_with_stats::test --lib`
  - 5 passed
- `cargo test -p tari_comms peer_validator::helpers::test --lib`
  - 4 passed
- `cargo test -p tari_comms_dht peer_validator::tests --lib`
  - 7 passed
- Peer-database classification tests:
  - `test_filtering_peers_by_transport_protocols`
  - both `discovery_syncing_peers_with_external_addresses_only` implementations
  - all passed
- `cargo clippy -p tari_comms -p tari_comms_dht --lib --tests -- -D warnings -A clippy::result_large_err`
  - passed; the allow is for a pre-existing warning in `comms/core/src/test_utils/mocks/connection_manager.rs`

---

Run the commands above. The tests cover valid signatures with mixed claims on both handshake and DHT paths, all-local claims for new and existing peers, non-ban semantics, the explicit `allow_test_addresses = true` path, invalid signatures, address-class boundaries, DNS false positives, trailing-dot FQDNs, and structural-error ordering.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/comms/core/src/connection_manager/common.rs
+++ b/comms/core/src/connection_manager/common.rs
@@ -36,7 +36,7 @@ use crate::{
     multiaddr::Multiaddr,
     net_address::{MultiaddressesWithStats, PeerAddressSource},
     peer_manager::{NodeId, NodeIdentity, Peer, PeerFeatures, PeerFlags, PeerIdentityClaim, PeerManagerError},
-    peer_validator::{PeerValidatorConfig, PeerValidatorError, validate_peer_identity_claim},
+    peer_validator::{PeerValidatorConfig, PeerValidatorError, validate_and_filter_peer_identity_claim_addresses},
     proto::identity::PeerIdentityMsg,
     protocol,
     protocol::{NodeNetworkInfo, ProtocolId},
@@ -48,6 +48,11 @@ const LOG_TARGET: &str = "comms::connection_manager::common";
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ValidatedPeerIdentityExchange {
     pub claim: PeerIdentityClaim,
+    /// Addresses from `claim` that this node is permitted to store and dial.
+    ///
+    /// The unmodified claim is retained separately because its signature and
+    /// provenance cover the original address list.
+    pub permitted_addresses: Vec<Multiaddr>,
     pub metadata: PeerIdentityMetadata,
 }
 
@@ -204,10 +209,12 @@ pub(super) fn validate_peer_identity_message(
             .try_into()?,
     };
 
-    validate_peer_identity_claim(config, authenticated_public_key, &peer_identity_claim)?;
+    let permitted_addresses =
+        validate_and_filter_peer_identity_claim_addresses(config, authenticated_public_key, &peer_identity_claim)?;
 
     Ok(ValidatedPeerIdentityExchange {
         claim: peer_identity_claim,
+        permitted_addresses,
         metadata: PeerIdentityMetadata {
             user_agent,
             supported_protocols,
@@ -242,7 +249,7 @@ pub(super) fn create_or_update_peer_from_validated_peer_identity(
                 peer.node_id.short_str()
             );
             peer.addresses.add_or_update_addresses(
-                &peer_identity.claim.addresses,
+                &peer_identity.permitted_addresses,
                 &PeerAddressSource::FromPeerConnection {
                     peer_identity_claim: peer_identity.claim.clone(),
                 },
@@ -250,7 +257,7 @@ pub(super) fn create_or_update_peer_from_validated_peer_identity(
 
             // For inbound connections we cannot distinguish between the peer's addresses, so we mark all as seen
             peer.addresses
-                .mark_all_addresses_as_last_seen_now_with_latency(&peer_identity.claim.addresses, latency);
+                .mark_all_addresses_as_last_seen_now_with_latency(&peer_identity.permitted_addresses, latency);
 
             peer.features = peer_identity.claim.features;
             peer.supported_protocols = peer_identity.metadata.supported_protocols.clone();
@@ -265,13 +272,13 @@ pub(super) fn create_or_update_peer_from_validated_peer_identity(
                 peer_node_id.short_str()
             );
             let mut addresses = MultiaddressesWithStats::from_addresses_with_source(
-                peer_identity.claim.addresses.clone(),
+                peer_identity.permitted_addresses.clone(),
                 &PeerAddressSource::FromPeerConnection {
                     peer_identity_claim: peer_identity.claim.clone(),
                 },
             );
             // For inbound connections we cannot distinguish between the peer's addresses, so we mark all as seen
-            addresses.mark_all_addresses_as_last_seen_now_with_latency(&peer_identity.claim.addresses, latency);
+            addresses.mark_all_addresses_as_last_seen_now_with_latency(&peer_identity.permitted_addresses, latency);
             Peer::new(
                 authenticated_public_key,
                 peer_node_id,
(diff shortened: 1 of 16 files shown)


kamexETH:
Resolved: the branch now points to GitHub-verified commit `b0487d1`. Its tree hash is identical to the locally tested commit, and the signed-commits check is green.

SWvheerden:
Nice work — the core insight is right: verify the signature over the unmodified claim, then filter. The DHT-side change is well tested. Two things that need to be resolved before merge, plus some smaller notes.

  ## 1. The fix isn't applied on the handshake path, and that path bans

  `validate_addresses` is also used by the connection handshake (`connection_manager/common.rs:207`, reached from `listener.rs:400` and `dialer.rs:458`), and both call sites wrap it in
  `common::ban_on_offence`. `PeerValidatorError::as_ban_duration()` (`peer_validator/error.rs:51`) returns `BAN_DURATION_LONG` (2h) unconditionally.

  So after this PR, a peer whose claim contains **any** private/loopback/link-local address — *even alongside a perfectly good public address* — fails the handshake and is banned for 2 hours. That's the
  exact outcome the PR description argues against ("without discarding a peer that also advertises a valid route"), just on the path this PR doesn't touch. Before the PR only `0.0.0.0`/`::` triggered it,
  so this is new.

  Same shape on the DHT join/discovery paths: `inbound/dht_handler/task.rs:224` and `discovery/service.rs:268` wrap `validate_peer` in `ban_offence` with `OffenceSeverity::High`, so a peer that filters
  down to zero addresses is now banned rather than just dropped.

  Realistic trigger: an operator sets `p2p.public_addresses` to a LAN address (private cluster, docker, NAT) without `allow_test_addresses`. That works today; after this PR the node gets banned
  network-wide.

  (The peer-*sync* paths are fine — `network_discovery/discovering.rs` lists `PeerValidationError` under "no banning needed", and `on_connect.rs`/`seed_strap.rs` handle failures per peer. So a sync peer
  isn't punished for relaying a bad claim. 👍)

  Preferred fix: use the new filtering variant in `validate_peer_identity_message` too. Failing that, make the `InvalidMultiaddr("Non-global IP addresses are invalid")` case non-bannable — misconfiguration
  isn't an attack.

  ## 2. `is_external` is persisted; stale rows aren't rewritten

  `is_external` is a column in `multi_addresses` (`peer_manager/storage/schema.rs:36`), written on insert/update (`database.rs:259`, `:353`, `:1476`) and queried by `discovery_syncing(..., 
  external_addresses_only)`, which `dht/src/rpc/service.rs:114` drives from `!allow_test_addresses`.

  Redefining the classifier doesn't rewrite existing rows, so peers already stored with `is_external = true` for a `192.168.x` / `fe80::` / `.local` address keep that flag and keep being served until their
  row is next updated. The PR ticks "Breaking Changes: None" — either add a migration that recomputes the column, or document the convergence delay.

  Related scope note worth adding to the PR body: since `dht/src/rpc/peer_info.rs:52` builds outbound `UnvalidatedPeerInfo` from `addr.source().peer_identity_claim()`, the unfiltered claim keeps
  propagating (unavoidable — the signature covers it). This PR stops nodes *dialling/storing* private addresses; it doesn't stop them being gossiped.

  ## Smaller points

  - **Existing peers get banned on an all-local update.** In `dht/src/peer_validator.rs`, when `existing_peer` is `Some` and already has usable addresses, an all-local update returns `PeerHasNoAddresses` →
  High-severity ban of a peer we can already reach. Returning `Ok(peer)` with the existing addresses untouched seems safer. `it_rejects_an_all_local_update_for_an_existing_peer` currently locks in the ban
  behaviour.
  - **Missing IPv4 classes** in `is_internal_ipv4`: multicast (224/4), broadcast, reserved (240/4), documentation, and `0.0.0.0/8` beyond the exact `0.0.0.0`.
  `is_multicast()`/`is_broadcast()`/`is_documentation()` are all stable — these are never valid peers.
  - **Missing IPv6 classes**: multicast (`ff00::/8`) and documentation (`2001:db8::/32`). Also `to_ipv4_mapped()` misses the deprecated IPv4-compatible `::a.b.c.d`; `to_ipv4()` covers both forms.
  - **No direct tests for `is_external_address`.** It's now `pub` and feeds the persisted DB column, but is only exercised transitively via `peer_validator::helpers::test`. The suffix matcher looks
  correct, but the valuable cases are the negative ones: `mylocal.com`, `internal-node.example.com`, `locallake.io`, `LOCALHOST.example.com`, and a trailing-dot FQDN like `example.com.`.
  - **Stale doc comment** on `is_external()` (`multiaddr_with_stats.rs:147`) — still says "not a loopback, unspecified or private IP address" but the classifier now covers link-local, ULA, mapped and DNS.
  - `it_rejects_an_all_local_update_for_an_existing_peer` re-implements `make_unvalidated_peer` inline — call the helper and build `existing_peer` separately.
  - `it_filters_local_addresses_but_keeps_the_peer` hardcodes `max_permitted_peer_addresses_per_claim = 7` to match the address count; derive it from the vec length so the test doesn't start failing on
  `PeerIdentityTooManyAddresses` when a case is added.
  - Minor/cosmetic: the new DNS guard arm sits before the TCP-port check, so `/dns4/localhost` (no port) now reports "Non-global DNS addresses are invalid" instead of "does not include a TCP port".

kamexETH:
Addressed the review in `df8ed1c`:

- The handshake path now verifies the original signed claim and uses only its permitted addresses, so mixed claims are accepted without banning the peer.
- New all-local claims are dropped without a ban, while an existing reachable peer retains its known usable addresses.
- Added the missing IPv4/IPv6 classes, IPv4-compatible handling, direct classifier tests, DNS negative cases, structural-error ordering, and the requested test refactors/doc update.
- Documented persisted-row convergence and why the original unfiltered signed claim may continue to be gossiped.

The targeted handshake, classifier, validator, DHT, and peer-database tests pass, as does strict Clippy for `tari_comms` and `tari_comms_dht`. The commit is GitHub-verified.

SWvheerden:
Looking good, 
This needs addressing, though.
 1: Peer-sync rounds now abort on the first all-local peer (should fix)

  comms/dht/src/network_discovery/discovering.rs:332-337 — validate_and_add_peer(...).await...? propagates out of the while let Some(resp) loop, so one peer whose claim filters to zero addresses (and that we don't already know via an external address) discards the remainder of the streamed 
  round. PeerValidationError isn't a ban offence, so the sync peer keeps serving the same list and the next round truncates at the same point.

2: Free signature verification with no penalty (low-medium)

  The structural pre-pass in validate_and_filter_peer_identity_claim_addresses now passes allow_test_addrs = true, so all-local claims reach Ristretto signature verification instead of being rejected cheaply before it — and is_ban_offence() returns false, so there's no cost to repeating them.
  Keeping the non-ban semantics is right for NAT'd peers; just be aware the spam bound now rests entirely on DHT message rate limiting.
  
3: is_ban_offence defaults new variants to "ban" — the exhaustive matches in discovery/service.rs and dht_handler/task.rs were replaced by !matches!(...), so a future DhtPeerValidatorError variant becomes a High/Medium ban with no compiler nudge. An exhaustive match self in the helper
  preserves that.

4: Misleading error text — "Peer '{peer}' has no address claims" is now emitted when the peer had claims but every address was filtered. There's already an unused InvalidPeerAddresses variant; a dedicated PeerHasNoUsableAddresses would read correctly in operator logs.
  
5:  Missing ranges (not security-critical): v4 192.88.99.0/24, 198.18.0.0/15, 100.64.0.0/10; v6 fec0::/10, 2002::/16, 2001::/32, 100::/64.

6: Divergent public API — validate_peer_identity_claim and validate_addresses now have zero callers outside helpers.rs, and their reject-the-whole-claim semantics differ from the new function. Worth removing or documenting which one new code should use.

kamexETH:
Addressed the follow-up review in `d65e438` (rebased onto current `development`):

- Peer-sync now skips an invalid relayed claim and continues consuming the streamed round; peer-manager/RPC failures still abort.
- Added `PeerHasNoUsableAddresses` so genuinely empty claims and policy-filtered claims produce accurate operator errors.
- Ban classification is exhaustive at both the core and DHT error layers; empty/all-local claims remain non-bannable, while future variants require an explicit decision.
- Added CGNAT, deprecated 6to4 relay, benchmarking, IPv6 discard-only, Teredo, 6to4, and deprecated site-local ranges plus classifier coverage.
- Documented the strict all-or-nothing validator API versus the filtering ingestion API.
- Kept signature verification for all-local signed claims intentionally: handshake/existing-peer paths may retain authenticated identity metadata even with an empty permitted address set, so skipping verification would create an authentication bypass. DHT rate limiting remains the cost bound.

Local verification after rebase:
- 13 targeted core/DHT tests pass.
- nightly rustfmt check passes for all five touched files.
- `cargo clippy -p tari_comms -p tari_comms_dht --lib --no-deps -- -D warnings -A clippy::result_large_err` passes. The single allow is an existing lint in `comms/core/src/test_utils/mocks/connection_manager.rs`, unrelated to this PR.
- Both commits are SSH-signed and GitHub-verifiable.

SWvheerden (approved):

metalaureate:
👋 @kamexETH — your PR was merged and this bounty is queued for an XTM payout. To receive it, **reply to this PR with your Tari (XTM) payout address** (base58 or emoji ID) on its own line, for example:

`XTM: <your address>`

For your protection we only accept an address posted by **@kamexETH** (the merged-PR author). Comments from anyone else are ignored.

PR #7956, opened 2026-08-18
fix: ledger secuerity

SWvheerden:
Description
---
fixes a bug in the how the ledger handles its script offset


SWvheerden:
repalced by #8000

